PR #32, opened 2026-03-26
feat: migrate metadata store to MySQL

srstack:
## Summary
- replace SQLite metadata driver and schema with TiDB/MySQL-compatible SQL, and switch server startup config to require `DAT9_MYSQL_DSN`
- update docs and CLI help text from `DAT9_DB_PATH` to `DAT9_MYSQL_DSN`, and clarify backend wording as TiDB/MySQL protocol
- add a shared testcontainers MySQL harness in `TestMain` so backend/client/meta/server tests run without manual database setup

## Validation
- go test ./...


qiffang:
# PR Review: SQLite → TiDB/MySQL Migration

## Critical Issues (3 found)

### C1. `resetTestDB` is copy-pasted 4 times — extract to shared helper
`pkg/backend/dat9_test.go`, `pkg/client/client_test.go`, `pkg/meta/meta_test.go`, `pkg/server/server_test.go` all define an identical `resetTestDB` function. This is a maintenance hazard — if you add a new table or change the delete order, you need to change it in 4 places.

**Fix**: Move `resetTestDB` into `internal/testmysql/testmysql.go` alongside the `Instance` type (e.g. `inst.Reset(t, db)` or a standalone `testmysql.ResetDB(t, db)`).

### C2. Timestamps stored as VARCHAR(64) instead of native DATETIME
```sql
created_at   VARCHAR(64) NOT NULL
confirmed_at VARCHAR(64)
expires_at   VARCHAR(64)
```

You're writing ISO 8601 strings into VARCHAR columns. This works but:
- **No native range queries**: `WHERE created_at > NOW() - INTERVAL 1 DAY` won't work
- **No MySQL date functions**: Can't use `DATE()`, `TIMESTAMPDIFF()`, etc.
- **Wastes storage**: 64 bytes vs 8 bytes per DATETIME
- **Unnecessary parsing code**: 30+ lines of `timeStr()`/`nilTimeStr()`/manual `time.Parse()` scanning

**Fix**: Use `DATETIME(3)` columns with `DEFAULT NOW(3)`. The DSN already has `parseTime=true`, so `go-sql-driver/mysql` will scan DATETIME directly into `time.Time` — eliminating all manual time parsing helpers.

```sql
created_at   DATETIME(3) NOT NULL DEFAULT NOW(3),
confirmed_at DATETIME(3),
expires_at   DATETIME(3)
```

### C3. VARCHAR(768) for path columns sits exactly at InnoDB index limit

```sql
path         VARCHAR(768) NOT NULL,
parent_path  VARCHAR(768) NOT NULL,
-- with UNIQUE INDEX on path
```

`768 chars × 4 bytes (utf8mb4) = 3072 bytes` — the exact InnoDB max index key length. This works on MySQL 8.0+ with `ROW_FORMAT=DYNAMIC` (the default), but it's a knife-edge:
- Fails on MySQL 5.6 or any instance with `ROW_FORMAT=COMPACT`
- Any future composite index involving `path` will exceed the limit

**Fix**: Use `VARCHAR(512)`. 512 chars is generous for filesystem paths (Linux PATH_MAX is 4096 *bytes*, not chars). Or keep 768 and add explicit `ROW_FORMAT=DYNAMIC` to the CREATE TABLE + document the MySQL 8.0+ requirement.

---

## Important Issues (5 found)

### I1. `CREATE INDEX` without `IF NOT EXISTS` — error swallowing is fragile

```go
if isIgnorableMigrationError(err) {
    continue
}
```

MySQL 8.0 supports `CREATE INDEX IF NOT EXISTS`. Use it instead of swallowing errors. Error swallowing is risky because `isIgnorableMigrationError` also runs for `CREATE TABLE` statements — if a table creation fails for a real reason (e.g., disk full), the error might be accidentally swallowed if the error message contains "already exists".

**Fix**: Keep `CREATE TABLE IF NOT EXISTS` (already there). Use `CREATE [UNIQUE] INDEX IF NOT EXISTS`. Remove `isIgnorableMigrationError` entirely.

### I2. `multiStatements=true` in test DSN — ensure it never reaches production

```go
dsn, err := c.ConnectionString(ctx, "parseTime=true", "multiStatements=true")
```

`multiStatements=true` is an SQL injection attack surface — if any user input reaches `Exec()`, an attacker can append `; DROP TABLE files`. Fine for test setup, dangerous in production.

**Fix**: Add a comment in `testmysql.go` marking this as test-only. Consider asserting in `meta.Open()` that the DSN does not contain `multiStatements=true`.

### I3. AWS SDK removed from go.mod while uploads table remains

The PR removes all `aws-sdk-go-v2` dependencies, but keeps the `uploads` table with its `s3_upload_id`, `s3_key` columns. This is confusing — the table schema promises S3 integration that can't be built without the SDK.

**Fix**: Either keep the AWS SDK (this PR is about metadata migration, not S3 removal) or remove the `uploads` table too and document that S3 support is deferred.

### I4. `EnsureParentDirs` uses `time.Now()` without `.UTC()`

```go
// pkg/meta/meta.go (EnsureParentDirs)
now := time.Now()  // ← local timezone!
```

But `InsertFile`, `UpdateFileContent`, `ConfirmFile`, `CompleteUpload` all use `time.Now().UTC()`. Inconsistent timezone handling will produce mixed timestamps in the database.

**Fix**: `now := time.Now().UTC()` everywhere. Or better: centralize with `func nowUTC() time.Time { return time.Now().UTC() }`.

### I5. `testmain_test.go` is copy-pasted 4 times

`pkg/backend/testmain_test.go`, `pkg/client/testmain_test.go`, `pkg/meta/testmain_test.go`, `pkg/server/testmain_test.go` are all identical 26-line files. This is unavoidable to some extent (Go test packages can't share TestMain), but the boilerplate could be reduced.

**Fix**: Consider having `testmysql.Start` cache the instance in a package-level var behind `sync.Once`, so each test package just calls `testmysql.DSN()` without the full `TestMain` ceremony. Or at minimum, add a comment explaining why the duplication is necessary.

---

## Suggestions (4 found)

### S1. `ON DUPLICATE KEY UPDATE node_id = node_id` is correct but consider `INSERT IGNORE`

The MySQL idiom works, but `INSERT IGNORE` (supported in MySQL 8.0) is more readable and semantically matches SQLite's `INSERT OR IGNORE`. The current approach is slightly safer (INSERT IGNORE suppresses ALL errors, not just duplicates), so this is a style preference.

### S2. `is_directory BOOLEAN` scanning works but could be cleaner

Current code: `var isDir int` + `n.IsDirectory = isDir != 0`. Since `go-sql-driver/mysql` with MySQL BOOLEAN (TINYINT) scans fine into `int`, this works. But scanning directly into `bool` also works with this driver. Not a blocker.

### S3. Consider adding a `CHARSET=utf8mb4` to CREATE TABLE statements

MySQL's default charset depends on server config. Explicit `DEFAULT CHARSET=utf8mb4` in DDL ensures consistent behavior regardless of server defaults. Important for path columns that may contain Unicode filenames.

### S4. `cmd/dat9/main.go` — indentation changed from spaces to tabs in help text

```go
// Before (spaces):
environment:
  DAT9_SERVER      server URL

// After (tabs):
	environment:
	  DAT9_SERVER      server URL
```

This will display differently in terminals. Keep consistent indentation.

---

## Strengths

1. **Clean migration scope** — the PR correctly limits itself to metadata store migration without touching the filesystem/backend logic
2. **testcontainers integration** — `internal/testmysql/` with environment override (`DAT9_MYSQL_DSN`) is the right pattern for testable infrastructure
3. **`isUniqueViolation` handles both dialects** — dual check for MySQL "Duplicate entry" and SQLite "UNIQUE constraint failed" is forward-compatible
4. **`db.Ping()` added to `Open()`** — fail-fast on connection issues is good practice
5. **Schema types are well-chosen** — BIGINT for sizes/revisions, VARCHAR lengths appropriate for their domains

---

## Recommended Action

1. **Fix C1-C3** before merge (DATETIME, path length, dedup resetTestDB)
2. **Address I1-I5** (IF NOT EXISTS, multiStatements, AWS SDK, UTC consistency, TestMain dedup)
3. **Consider S1-S4** as polish
4. Re-run `go test ./...` with a fresh MySQL container to verify

**Verdict: Needs changes before merge.**

srstack:
Thanks for the detailed review — I addressed the requested changes in follow-up commits:

- extracted duplicated test reset logic to shared helper: `internal/testmysql.ResetDB(...)`
- converted timestamp columns to native `DATETIME(3)` and switched scanning to `time.Time`/`sql.NullTime`
- reduced indexed path columns to `VARCHAR(512)` to avoid index-length edge cases
- removed `multiStatements=true` from testcontainer DSN and added guard in `meta.Open` to reject it
- unified UTC handling (`EnsureParentDirs` now uses `time.Now().UTC()`)
- fixed help text indentation in `cmd/dat9/main.go`
- merged latest `main` and resolved conflicts (serve path moved to `cmd/dat9-server`)

Note on index DDL: this environment still errors on `CREATE INDEX IF NOT EXISTS`, so migration now handles idempotency by only ignoring duplicate-index errors for index statements (not table creation), keeping the original safety concern addressed.

Validation run: `go test ./...`

qiffang:
## Code Review Summary

Reviewed the SQLite → TiDB/MySQL migration. Overall direction is solid. Here are consolidated suggestions:

### 1. Remove `go-sqlite3` from go.mod
If SQLite is no longer a production backend, remove `github.com/mattn/go-sqlite3` to reduce compile time (cgo) and attack surface. Run `go mod tidy` to confirm.

### 2. DSN Configuration
- Add `parseTime=true&loc=UTC` (or `loc=Local` with consistent timezone) to avoid timezone drift when scanning `DATETIME(3)` into `time.Time`
- Add `charset=utf8mb4&collation=utf8mb4_bin` for correct UTF-8 handling and case-sensitive path/name comparisons
- Consider adding connection pool settings: `SetMaxOpenConns()`, `SetMaxIdleConns()`, `SetConnMaxLifetime()`

### 3. DDL/Schema
- Explicitly set `ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin` on all tables
- For `CREATE INDEX`, consider using `IF NOT EXISTS` (MySQL 8+ supports it) instead of relying on `isDuplicateIndexError()`
- `VARCHAR(512)` for `path`: Linux max path is 4096 bytes. If storing absolute paths, consider `VARCHAR(4096)` or `TEXT`
- Confirm index byte-length limits under `utf8mb4` (max key length varies by MySQL/TiDB version)

### 4. Time Handling Consistency
- `ON UPDATE CURRENT_TIMESTAMP(3)` is only on `uploads.updated_at`. Either apply it to all tables or handle all updates in application code—mixed approach can lead to inconsistencies
- If you want UTC storage, set session `time_zone="+00:00"` or align via DSN

### 5. Boolean Fields
Ensure MySQL `BOOLEAN` (which is `TINYINT(1)`) scans correctly into Go `bool`. The current `is_directory` uses `int` scan then converts—this works but verify it matches the new schema.

### 6. Migration Safety
- With `multiStatements=false`, confirm the migrator executes DDL statements one at a time
- `isDuplicateIndexError()` should cover both MySQL error codes (1061) and TiDB variants, not just string matching

---

**Verdict: LGTM with minor suggestions above.**

Main action items:
1. ✅ Confirm whether to remove go-sqlite3
2. ✅ Unify `updated_at` handling strategy
3. ✅ Add charset/collation to DDL
4. ✅ Review VARCHAR lengths for paths

qiffang (approved):
LGTM! 🎉

The SQLite → TiDB/MySQL migration looks solid. Minor suggestions provided in the previous comment can be addressed in follow-up PRs if needed.

PR #37, opened 2026-03-27
feat: multi-tenant auth with per-tenant API keys and cluster routing

qiffang:
## Summary
- Add multi-tenant authentication: each tenant gets a unique API key mapped 1:1 to a dedicated db9/TiDB cluster
- New `pkg/tenant` package: tenant model, control plane store (MySQL), AES-GCM password encryption, API key generation (prefix + SHA-256 hash), LRU backend pool with revocation, TiDB Cloud Starter provisioner
- Auth middleware: Bearer token → hash lookup → status check → pool.Get → context injection
- `POST /v1/provision` endpoint (admin key guarded, timing-safe comparison)
- TLS-aware DSN builder using `mysql.Config.FormatDSN()` for safe escaping
- Backward compatible: local dev mode works without auth config via fallback backend injection

## Architecture
- Control plane DB separated from per-tenant DBs
- API key: `dat9_<64hex>`, stored as `prefix` (first 8 chars) + `SHA-256 hash` — plaintext returned once at provision time
- Tenant DB passwords encrypted at rest with AES-256-GCM
- Tenant status flow: `provisioning` → `active` (→ `suspended` / `deleted`)
- Status mapping: active → 200, provisioning → 503, suspended/deleted → 403

## Review
- Approved by @dat9-reviewer1 and @dat9-reviewer2
- All 8 review gates verified ✅

## Test plan
- [x] `go vet ./...` passes
- [x] `go build ./...` passes
- [x] `go test ./pkg/tenant/...` — encrypt, apikey, DSN, store (with DB), starter tests
- [x] Auth middleware unit tests: missing key (401), invalid key (401), suspended (403), provisioning (503), local dev passthrough (200)
- [ ] Integration tests (in progress by @dat9-dev3 in #t55)

### DIFF
--- a/cmd/dat9-server/main.go
+++ b/cmd/dat9-server/main.go
@@ -6,6 +6,7 @@
 package main
 
 import (
+	"encoding/hex"
 	"fmt"
 	"log"
 	"os"
@@ -15,6 +16,7 @@ import (
 	"github.com/mem9-ai/dat9/pkg/meta"
 	"github.com/mem9-ai/dat9/pkg/s3client"
 	"github.com/mem9-ai/dat9/pkg/server"
+	"github.com/mem9-ai/dat9/pkg/tenant"
 )
 
 const (
@@ -33,11 +35,6 @@ func main() {
 		addr = os.Args[1]
 	}
 
-	mysqlDSN := os.Getenv("DAT9_MYSQL_DSN")
-	if mysqlDSN == "" {
-		die(fmt.Errorf("DAT9_MYSQL_DSN is required"))
-	}
-
 	blobDir := envOr("DAT9_BLOB_DIR", defaultBlobDir)
 	s3Dir := envOr("DAT9_S3_DIR", defaultS3Dir)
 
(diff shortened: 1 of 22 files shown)


srstack:
Consolidated blocking review comments (aligned to the agreed multi-tenant design):

1) `/s3` routing must not depend on in-memory pool cache hits
- Current behavior: `/s3/{tenantID}/...` only uses `pool.S3Backend(tenantID)`.
- Problem: after restart or LRU eviction, valid unexpired presigned URLs return 404.
- Required fix: on cache miss, load tenant by `tenantID` from store, rebuild/get backend via pool, then verify signature and serve.

2) Auth error mapping is incorrect for internal failures
- Current behavior: all `tenants.GetByAPIKeyHash` errors are mapped to `401 invalid API key`.
- Problem: DB/control-plane outages are misreported as auth failures.
- Required fix: only `not found` -> 401; internal/store errors -> 5xx with generic message.

3) Provider selection must be explicit and same-level by config
- Required architecture: `db9` / `tidb_zero` / `tidb_cloud_starter` are peer options selected via one config key.
- Current code only wires `tidb_starter` via env-based implicit detection.

4) Encryption must use environment-level KMS alias abstraction
- Required architecture: env KMS alias (e.g., `alias/dat9-<env>-db-password`) behind encryptor interface.
- Current code uses `DAT9_MASTER_KEY` local AES-GCM path directly.

5) Server config needs fail-fast validation
- Current code allows invalid config combinations (no tenants+pool and no fallback backend), risking nil backend path.
- Required fix: validate in constructor and fail early.

6) Provision flow must include FTS + Vector capability check
- Current flow initializes schema but does not explicitly validate required capabilities.
- Required fix: probe capability post-provision, fail and rollback tenant status if requirements are not met.

Please address all 6 blocking items and request re-review.

PR #46, opened 2026-03-27
fix: sign multipart part checksums in upload plans

srstack:
## Summary
- keep multipart uploads on checksum mode and include per-part `x-amz-checksum-sha256` in presigned upload-part signatures
- extend upload plan part payload to expose `checksum_sha256` so clients can upload with checksum headers that match signed requests
- add server support for `X-Dat9-Part-Checksums` on initiate/resume and validate checksum-count consistency
- update backend and client upload/resume flows to propagate per-part checksums end-to-end

## Validation
- `go test ./...`
- `make lint`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -3,6 +3,13 @@ SHELL := /bin/bash
 GO ?= go
 DOCKER ?= docker
 
+HOST_GOOS ?= $(shell $(GO) env GOOS)
+HOST_GOARCH ?= $(shell $(GO) env GOARCH)
+
+# Build target defaults to the local environment; callers can override.
+GOOS ?= $(HOST_GOOS)
+GOARCH ?= $(HOST_GOARCH)
+
 APP_NAME ?= dat9-server
 CLI_NAME ?= dat9
 
@@ -50,11 +57,11 @@ build: build-server build-cli
 
 build-server:
 	mkdir -p $(BIN_DIR)
-	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 $(GO) build -o $(SERVER_BIN) ./cmd/dat9-server
+	CGO_ENABLED=0 GOOS=$(GOOS) GOARCH=$(GOARCH) $(GO) build -o $(SERVER_BIN) ./cmd/dat9-server
 
 build-cli:
 	mkdir -p $(BIN_DIR)
-	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 $(GO) build -o $(CLI_BIN) ./cmd/dat9
+	CGO_ENABLED=0 GOOS=$(GOOS) GOARCH=$(GOARCH) $(GO) build -o $(CLI_BIN) ./cmd/dat9
 
 docker-build: build-server
 	DOCKER_BUILDKIT=0 $(DOCKER) build -t $(IMAGE) .
--- a/pkg/backend/upload.go
+++ b/pkg/backend/upload.go
@@ -32,6 +32,10 @@ func (b *Dat9Backend) IsLargeFile(size int64) bool {
 // InitiateUpload creates a multipart upload for a large file.
 // Returns an UploadPlan with presigned URLs for all parts.
 func (b *Dat9Backend) InitiateUpload(ctx context.Context, path string, totalSize int64) (*UploadPlan, error) {
+	return b.InitiateUploadWithChecksums(ctx, path, totalSize, nil)
+}
+
+func (b *Dat9Backend) InitiateUploadWithChecksums(ctx context.Context, path string, totalSize int64, partChecksums []string) (*UploadPlan, error) {
 	path, err := pathutil.Canonicalize(path)
 	if err != nil {
 		return nil, err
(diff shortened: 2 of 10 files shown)


qiffang:
Review from two independent reviewers (Claude Code + Codex):

**Claude Code: SHIP** (with should-fix items)
**Codex: NO-SHIP** (1 blocker)

---

### Blocking Issue (Codex)

**Checksum binding is optional — §11.2 can be bypassed**

If the client omits `X-Dat9-Part-Checksums`, `parsePartChecksumsHeader()` returns nil → backend passes empty checksums → `aws.PresignUploadPart()` does not set `ChecksumSHA256` → presigned URLs are not checksum-bound. This contradicts design doc §11.2 ("Upload URLs bind … x-amz-checksum-sha256").

**Suggested fix:** Either require the header for large file uploads (reject if missing), or document this as progressive enhancement.

---

### Should-Fix (Claude Code)

**B. WriteStream only computes checksums when r implements io.ReaderAt.** If r is a plain io.Reader (e.g. stdin pipe), checksums are silently skipped. Same bypass as above from the client side.

**D. parsePartChecksumsHeader does not validate base64 format.** Any non-empty string passes. Invalid checksums would cause S3 upload failures at PUT time rather than being caught early.

**E. Error routing uses strings.Contains(err.Error(), "checksum").** Fragile — any error containing "checksum" gets 400d. Consider typed errors or sentinel values.

---

### Confirmed OK

- **AWS presigned URL checksum binding**: Correct — ChecksumSHA256 on UploadPartInput is signed by AWS SDK v2 presigner.
- **Double-read of file data**: Correct trade-off. ReadAt + upload has no race. If file changes between passes, uploadOnePart catches the mismatch.
- **Resume flow**: Correctly pre-computes checksums and sends via header.

srstack:
Addressed the review feedback in commit `1c99d91`.

What changed:
- Enforced checksum-bound multipart initiation and resume on server side:
  - `PUT /v1/fs/*` (large upload path) now requires `X-Dat9-Part-Checksums`.
  - `POST /v1/uploads/{id}/resume` now requires `X-Dat9-Part-Checksums`.
- Added strict checksum header validation (`base64` decode + 32-byte SHA-256 length).
- Replaced fragile `strings.Contains(err.Error(), "checksum")` routing with typed error handling:
  - introduced `backend.ErrPartChecksumCountMismatch`
  - server maps this via `errors.Is(... )` to 400.
- Removed checksum-bypass on client large uploads:
  - `WriteStream` now requires a seekable reader for large uploads (so checksums are always computed and sent).
  - Added a client test for non-seekable large reader rejection.
- Updated server upload tests to include checksum headers for large init/resume requests.

Validation:
- `go test ./...`
- `make lint`

qiffang:
## Code Review: Checksum-Bound Multipart Upload

This PR adds SHA256 checksum binding to multipart uploads for data integrity. +380 / -173 lines.

### ✅ Security Design - Solid

Binding checksums to presigned URLs is the right approach:
- S3 validates `x-amz-checksum-sha256` server-side
- Prevents MITM tampering and client-side data corruption
- Client-side pre-verification before upload adds extra safety layer

### ⚠️ Backward Compatibility - Consider Transition Period

Old clients without `X-Dat9-Part-Checksums` header will get immediate 400 errors.

**Suggestion**: Consider a transition period:
1. **Phase 1**: Log warning but allow checksum-less uploads
2. **Phase 2**: Return deprecation warning header
3. **Phase 3**: Enforce and return 400

Or if client update cadence is fast, document as breaking change in release notes.

### ⚠️ Performance - Memory Considerations

`computePartChecksumsFromReaderAt()` requires full file read before upload:

```go
ra, ok := r.(io.ReaderAt)
if !ok {
    return fmt.Errorf("large uploads require a seekable reader")
}
```

**Considerations**:
- For local files: Sequential read is acceptable
- For streaming sources: Requires buffering entire file → memory pressure
- Document clearly that only seekable sources are supported for large uploads

**Alternative** (future optimization): Stream checksum calculation alongside upload, validate at CompleteMultipartUpload.

### ✅ Resume Flow - Correct

Recalculating checksums on resume is correct:
- Previous presigned URLs have expired
- Local file may have changed since last attempt
- New checksums ensure consistency

### ⚠️ Error Messages - Could Be More Specific

Current:
```
"invalid X-Dat9-Part-Checksums header"
```

**Suggestion**: More actionable errors:
```
"invalid X-Dat9-Part-Checksums: expected 5 checksums, got 3"
"invalid X-Dat9-Part-Checksums: index 2 is not valid base64"
"invalid X-Dat9-Part-Checksums: index 2 decoded to 16 bytes, expected 32"
```

### Minor Notes

1. Good use of `ErrPartChecksumCountMismatch` as a typed error
2. Header parsing validation (base64 + 32 bytes) is thorough
3. Consistent handling in both `InitiateUpload` and `ResumeUpload` paths

---

**Summary**:

| Item | Status |
|------|--------|
| Security design | ✅ LGTM |
| Backward compat | ⚠️ Consider transition |
| Memory/streaming | ⚠️ Document limitation |
| Error messages | ⚠️ Could be more specific |
| Resume flow | ✅ Correct |

**Overall: LGTM with minor suggestions above.**

srstack:
Thanks — I addressed the new review suggestions in `547b049`.

Updated in this commit:
- Improved checksum header parsing errors to be actionable:
  - empty value index
  - invalid base64 index
  - decoded-length mismatch (expected 32 bytes)
- Added parser validation tests in `pkg/server/upload_test.go`.
- Clarified client-side large upload error message to explicitly require `io.ReaderAt` (seekable source) for checksum computation.

Notes on compatibility/transition:
- This PR intentionally enforces checksum-bound multipart uploads now (breaking for old clients), because the objective of this change is strict checksum binding end-to-end.
- The repo client/server are updated together in this PR, and all checks are green.

Validation:
- `go test ./...`
- `make lint`

qiffang:
## Review Round 2: Checksum Validation Feedback (547b049)

The new commit addresses the error message feedback from Round 1. 👏

### Changes

1. **Client-side error**: Now explains that `io.ReaderAt` is needed for per-part checksum computation
2. **Server-side validation**: Specific error messages for each failure case:
   - Empty value at index N
   - Invalid base64 at index N  
   - Wrong decoded length at index N
3. **Test coverage**: `TestParsePartChecksumsHeaderValidation` verifies error paths

### Verdict

✅ **LGTM - Ready to merge!**

All previous suggestions have been addressed. The error messages are now actionable and will help users debug issues quickly.

srstack:
Root-cause fix pushed in `38d5540` for the 100MB multipart failure (`HeadersNotSigned=x-amz-checksum-sha256`).

What changed:
- Switched S3 presigner to disable header hoisting for presigned requests.
- `UploadPartURL` now includes `headers` in plan payload (`map[string]string`) for required signed headers.
- AWS presign now returns flattened signed headers (excluding `host`) and includes checksum header requirement.
- Client upload path now applies `part.headers` before sending each part request.
- Local S3 presign path also returns checksum header in `headers` for parity.

Why this fixes prod failure:
- Previously checksum was only in query while client sent `x-amz-checksum-sha256` header, causing S3 403 `HeadersNotSigned`.
- Now signed/required headers are explicit in plan and the client sends exactly those headers.

Validation:
- `go test ./...`
- `make lint`

PR #47, opened 2026-03-27
feat: support db9-style <name>:/path context selector in fs commands

qiffang:
## Summary

- All `dat9 fs` commands now support `<name>:/path` to select which context (database) to use, matching db9's `db9 fs cp test1:/TODO.md test.md` pattern
- Cross-context operations (`dat9 fs mv db1:/a db2:/b`) are explicitly rejected with a clear error

## Usage

```bash
# download from specific context
dat9 fs cp test1:/TODO.md test.md

# upload to specific context
dat9 fs cp TODO.md test1:/TODO.md

# list/cat/stat with context
dat9 fs ls test1:/
dat9 fs cat test1:/config.json
dat9 fs stat test1:/data.csv

# use current context (unchanged)
dat9 fs ls :/
dat9 db sql -q "SELECT 1"

# cross-context copy rejected
dat9 fs cp db1:/a db2:/b
# → error: cross-context copy not supported: db1 vs db2
```

## Files changed

| File | Change |
|---|---|
| `cmd/dat9/cli/path.go` | NEW: `ParseRemote()` parses `test1:/path`, `:/path`, returns `(RemotePath{Context, Path}, isRemote)` |
| `cmd/dat9/cli/cp.go` | Uses `ParseRemote` for src/dst, resolves client per-context, rejects conflicting contexts |
| `cmd/dat9/cli/client.go` | `NewClientForContext(name)` resolves api_key from `~/.dat9/config` for given context |
| `cmd/dat9/main.go` | `extractContext()` strips `<name>:` from all args for non-cp fs commands, rejects conflicting contexts |
| `cmd/dat9/cli/cp_test.go` | `TestParseRemote` with 9 test cases |

## Review history

- Claude Code: SHIP ✅
- Codex: SHIP ✅ (after fixing conflicting context rejection for non-cp commands)

### DIFF
--- a/cmd/dat9/cli/client.go
+++ b/cmd/dat9/cli/client.go
@@ -1,6 +1,7 @@
 package cli
 
 import (
+	"fmt"
 	"os"
 
 	"github.com/mem9-ai/dat9/pkg/client"
@@ -19,8 +20,15 @@ func NewClientForContext(ctxName string) *client.Client {
 		if name == "" {
 			name = cfg.CurrentContext
 		}
-		if ctx, ok := cfg.Contexts[name]; ok {
-			apiKey = ctx.APIKey
+		if name != "" {
+			ctx, ok := cfg.Contexts[name]
+			if !ok && ctxName != "" {
+				fmt.Fprintf(os.Stderr, "error: context %q not found in %s\n", ctxName, configPath())
+				os.Exit(1)
+			}
+			if ok {
+				apiKey = ctx.APIKey
+			}
 		}
 	}
 	return client.New(server, apiKey)
--- a/cmd/dat9/cli/client.go
+++ b/cmd/dat9/cli/client.go
@@ -6,7 +6,7 @@ import (
 	"github.com/mem9-ai/dat9/pkg/client"
 )
 
-func NewFromEnv() *client.Client {
+func NewClientForContext(ctxName string) *client.Client {
 	server := os.Getenv("DAT9_SERVER")
 	apiKey := os.Getenv("DAT9_API_KEY")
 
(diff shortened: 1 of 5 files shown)


srstack:
Thanks for the update — I found two issues worth fixing before merge:

1. **Behavioral regression on Windows/local paths** (`cmd/dat9/cli/path.go`)  
`ParseRemote()` currently treats any string containing `":/"` as remote context syntax. This can misclassify Windows local paths like `C:/tmp/a.txt` as remote (`context=C`, `path=/tmp/a.txt`), which breaks local file handling in `fs cp` on Windows.

   **Suggestion:** only treat as remote when the input starts with `:/` or matches a stricter `<context>:/...` pattern, and explicitly exempt drive-letter paths.

2. **Unknown-context UX / implicit auth failure** (`cmd/dat9/cli/client.go`)  
When a non-empty context name is provided but not found in config, `NewClientForContext()` silently falls back to empty API key and continues. Users then get downstream auth errors instead of a clear “context not found”.

   **Suggestion:** fail fast when `ctxName` is provided but missing in config (either return error from `NewClientForContext`, or pre-validate in caller).

Build/lint/test all pass on this branch; these are mainly correctness/UX guardrails.

qiffang:
Both issues fixed in 0febf30:

**1. Windows drive letter paths:** `ParseRemote` and `extractContext` now skip single-character prefixes before `:/` (i.e. `idx == 1`). `C:/tmp/a.txt` and `D:/Users/test` are correctly treated as local paths. Added test cases for both.

**2. Unknown context fail-fast:** `NewClientForContext` now exits with `error: context "foo" not found in ~/.dat9/config` when an explicit context name is provided but missing from config. Only fails fast when the user explicitly specified a context (via `<name>:/path`); current_context fallback still returns empty API key gracefully (no config = first-run scenario).

srstack:
Re-checked the latest commit (`0febf30`) and the two previously reported issues are fixed:

- ✅ Windows drive-letter paths (`C:/...`) are no longer misclassified as remote context paths.
- ✅ Explicit unknown context now fails fast with a clear error instead of silently falling back.

I also re-ran checks locally on this branch:
- `make build` ✅
- `make lint` ✅
- `make test` ✅

No blocking issues from my side — this is ready to merge.

PR #48, opened 2026-03-27
fix: include signed upload headers in multipart plans

srstack:
## Summary
- disable S3 presign header hoisting for upload-part URLs so checksum header signing stays consistent
- include required signed headers in upload plan part payload (`headers`) and apply them on client part uploads
- keep checksum-bound multipart behavior while fixing S3 `HeadersNotSigned=x-amz-checksum-sha256` failures on large uploads

## Validation
- `go test ./...`
- `make lint`

### DIFF
--- a/pkg/client/transfer.go
+++ b/pkg/client/transfer.go
@@ -24,11 +24,12 @@ type UploadPlan struct {
 
 // PartURL is a presigned URL for uploading one part.
 type PartURL struct {
-	Number         int    `json:"number"`
-	URL            string `json:"url"`
-	Size           int64  `json:"size"`
-	ChecksumSHA256 string `json:"checksum_sha256,omitempty"`
-	ExpiresAt      string `json:"expires_at"`
+	Number         int               `json:"number"`
+	URL            string            `json:"url"`
+	Size           int64             `json:"size"`
+	ChecksumSHA256 string            `json:"checksum_sha256,omitempty"`
+	Headers        map[string]string `json:"headers,omitempty"`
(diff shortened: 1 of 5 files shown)


qiffang:
## Code Review: Include Signed Headers in Multipart Plans

This PR fixes presigned URLs to include signed headers. +59 / -12 lines.

### ✅ Key Design Decisions - All Correct

#### 1. DisableHeaderHoisting = true
Correct choice. By default AWS SDK hoists signed headers into query string, but checksum headers need to be sent as HTTP headers. This setting ensures:
- `x-amz-checksum-sha256` stays in the signature
- Client must set these headers when uploading
- Good compatibility with S3-compatible storage (MinIO, R2, etc.)

#### 2. Host Header Filtering (Double Check)
Necessary and correct:
- Server-side `flattenSignedHeaders`: Avoids returning Host to client
- Client-side upload: Defensive - prevents accidental override
- `Host` is auto-set by HTTP client, should not be manually specified

#### 3. Header Key Lowercase
Correct approach:
- HTTP headers are case-insensitive (RFC 7230)
- Lowercase makes it easier for clients to use and compare

#### 4. LocalS3Client Consistency
Good that it also returns `x-amz-checksum-sha256` in Headers - keeps interface consistent.

### Minor Suggestions

1. Consider adding a brief comment to `flattenSignedHeaders` explaining why Host is filtered
2. Optional: Add test case verifying upload fails when headers are missing

---

**Verdict: LGTM - Ready to merge!**

PR #51, opened 2026-03-27
fix: scope checksum presigner to multipart upload parts

srstack:
## Summary
- keep default S3 presigner behavior for non-multipart operations (notably presigned GET)
- apply checksum/header-hoisting-disabled signer only to `PresignUploadPart`
- avoid introducing signed `x-amz-checksum-mode` requirements on GET object URLs

## Validation
- `go test ./pkg/s3client ./pkg/client`
- `make lint`

### DIFF
--- a/pkg/s3client/aws.go
+++ b/pkg/s3client/aws.go
@@ -38,6 +38,7 @@ func NewAWS(ctx context.Context, cfg AWSConfig) (*AWSS3Client, error) {
 
 	awsCfg, err := awsconfig.LoadDefaultConfig(ctx,
 		awsconfig.WithRegion(cfg.Region),
+		awsconfig.WithResponseChecksumValidation(aws.ResponseChecksumValidationWhenRequired),
 	)
 	if err != nil {
 		return nil, fmt.Errorf("load aws config: %w", err)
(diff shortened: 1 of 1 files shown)


qiffang:
## Code Review: Scope Checksum Presigner to Multipart Upload Parts

This PR fixes presigner configuration scope and simplifies e2e tests. +15 / -395 lines.

### ✅ Core Fix - Correct Approach

#### Presigner Scope Change
```go
// Before: Global DisableHeaderHoisting for all presign operations
presign: s3.NewPresignClient(client, func(o *s3.PresignOptions) { o.Presigner = presigner })

// After: Only UploadPart uses DisableHeaderHoisting
c.presign.PresignUploadPart(ctx, in, ..., func(o *s3.PresignOptions) { o.Presigner = partPresigner })
```

**Why this is correct:**
- `DisableHeaderHoisting` is specifically needed for multipart upload parts with checksum headers
- Other operations (GetObject, PutObject for small files) work better with default header hoisting
- This avoids potential issues with presigned GET URLs that don't need checksum headers

#### ResponseChecksumValidation
```go
awsconfig.WithResponseChecksumValidation(aws.ResponseChecksumValidationWhenRequired)
```
This configures the AWS SDK to validate response checksums only when required (e.g., when the server returns a checksum header). Good default for production - avoids unnecessary computation overhead while maintaining integrity when checksums are present.

### ⚠️ Test Coverage Reduction

Deleted tests:
- `cli-smoke-test.sh` (entire file - 221 lines)
- Large file multipart upload test from `api-smoke-test.sh`
- Batch small-file tests

**Concerns:**
1. **Large file multipart flow** was the primary way to validate the checksum-bound upload feature in e2e
2. **CLI smoke test** validated the CLI binary against real deployments

**Questions:**
- Is there alternative test coverage for multipart uploads elsewhere (unit tests, integration tests)?
- Is the CLI smoke test being moved to a different location or replaced?

If these flows are covered by unit tests or CI integration tests, the e2e simplification is acceptable. Otherwise, consider keeping at least the multipart upload e2e test.

### Minor Note

The presigner is now created per-call in `PresignUploadPart`. This is a small allocation but negligible for the upload use case.

---

**Summary**:

| Item | Status |
|------|--------|
| Presigner scope fix | ✅ Correct |
| ResponseChecksumValidation | ✅ Good default |
| Test deletion | ⚠️ Confirm coverage elsewhere |

**Overall: LGTM for the code fix.** Please confirm multipart upload is tested elsewhere before merging.

PR #57, opened 2026-03-28
refactor observability structure and simplify key metrics

srstack:
## Summary
- Split observability concerns into dedicated packages: `pkg/logger` (logging + CLI log rotation), `pkg/metrics` (service operation metrics), and `pkg/traceid` (trace context helpers), and removed the old mixed `pkg/observability` package.
- Standardized context propagation and logging across server/backend/meta/datastore/tenant paths, with trace-aware `logger.Level(ctx, ...)` usage and context-aware data-layer APIs.
- Simplified tenant event metrics to focus on key-chain signals (`ok/error` plus a small set like `accepted/conflict/cluster_error`), while keeping detailed failure reasons in structured logs.
- Hardened CLI logging defaults: CLI log writing is now disabled by default (`DAT9_CLI_LOG_ENABLED` to enable), with rotation cleanup retained and compression always enabled.
- Updated and expanded tests to assert key metrics behavior and preserve full regression coverage after the refactor.

## Validation
- `make lint`
- `go test ./...`

### DIFF
--- a/cmd/dat9-server/main.go
+++ b/cmd/dat9-server/main.go
@@ -9,15 +9,16 @@ import (
 	"context"
 	"encoding/hex"
 	"fmt"
-	"log"
 	"os"
 	"strconv"
 	"strings"
 
 	"github.com/mem9-ai/dat9/pkg/encrypt"
+	"github.com/mem9-ai/dat9/pkg/logger"
 	"github.com/mem9-ai/dat9/pkg/meta"
 	"github.com/mem9-ai/dat9/pkg/server"
 	"github.com/mem9-ai/dat9/pkg/tenant"
+	"go.uber.org/zap"
 )
 
 const (
@@ -34,6 +35,13 @@ func main() {
 	if len(os.Args) == 2 {
 		addr = os.Args[1]
 	}
+
+	srvLogger, err := logger.NewServerLogger()
+	if err != nil {
+		die(fmt.Errorf("create logger: %w", err))
+	}
+	defer func() { _ = srvLogger.Sync() }()
+	logger.Set(srvLogger)
 
 	metaDSN := os.Getenv("DAT9_META_DSN")
 	if metaDSN == "" {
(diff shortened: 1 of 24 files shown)


qiffang (approved):
## PR #57 Review: Observability Refactor

**Overall: LGTM with minor suggestions** 👍

This is a well-structured observability overhaul. The separation of concerns (logger/metrics/trace) is clean and follows Go best practices.

### What I Like

1. **Clean package separation**
   - `pkg/logger` - structured logging with zap
   - `pkg/metrics` - Prometheus-compatible metrics
   - `pkg/traceid` - distributed tracing context
   - Each package is focused and testable

2. **Trace ID propagation**
   - Accepts `X-Trace-ID` header from clients
   - Auto-generates if not provided
   - Propagates through context to all log lines
   - Returns trace ID in response header

3. **CLI log rotation**
   - `~/.dat/log/dat9-cli.log` with lumberjack
   - Configurable via env vars (size, backups, age)
   - Won't fill up user disk

4. **Prometheus metrics**
   - HTTP request metrics (count, duration, in-flight)
   - Service operation metrics (component/operation/result)
   - Tenant lifecycle events
   - All properly labeled

### Suggestions

1. **Consider histogram for latency metrics**
   
   Current approach uses sum/count (average only). For P99/P95, consider:
   ```go
   // Instead of just sum/count, add buckets
   dat9_http_request_duration_seconds_bucket{le="0.1"} 
   dat9_http_request_duration_seconds_bucket{le="0.5"}
   dat9_http_request_duration_seconds_bucket{le="1.0"}
   ```
   
   Not blocking - sum/count is fine for initial rollout.

2. **Logger sync on shutdown**
   
   `main.go` has `defer srvLogger.Sync()` which is good, but on SIGTERM the defer might not run. Consider:
   ```go
   signal.Notify(sigCh, syscall.SIGTERM)
   go func() {
       <-sigCh
       srvLogger.Sync()
       os.Exit(0)
   }()
   ```

3. **Minor: `poolBackgroundContext()` function**
   
   In `pool.go`, this creates a new context with trace ID for encrypt/decrypt. Consider accepting context from caller for better trace continuity:
   ```go
   func (p *Pool) Decrypt(ctx context.Context, cipher []byte) ([]byte, error) {
       return p.enc.Decrypt(ctx, cipher)
   }
   ```

### Questions

1. Is there a plan to add structured error logging? Currently errors are logged but not in a machine-parseable error code format.

2. Will `/metrics` endpoint be protected in production? Currently no auth required.

---

**Verdict: Approve** ✅

Clean implementation, good separation of concerns, follows observability best practices. Minor suggestions above are non-blocking.

PR #61, opened 2026-03-28
feat: add dual text and image embedding columns

srstack:
## Summary
- split file embedding into two generated columns: `embedding_text` (text model) and `embedding_image` (image model)
- use `tidbcloud_free/amazon/titan-embed-text-v2` for text embeddings and `tidbcloud_free/amazon/titan-embed-image-v1` for image embeddings
- add separate vector indexes for text and image embeddings, and keep grep vector ranking on `embedding_text` to preserve text search quality

## Validation
- go test ./pkg/tenant ./pkg/datastore


srstack:
Follow-up fixes pushed for review feedback:

- kept compatibility by preserving text column/index names (`embedding`, `idx_files_cosine`) and only adding `embedding_image`
- updated datastore schema test helper to include `embedding` + `embedding_image` columns
- improved `vectorSearch` error handling with warning logs for query/scan/rows errors (no silent swallow)
- added unit tests for dual vector query builder (`pkg/datastore/search_vector_test.go`) covering:
  - both text/image vector branches present
  - UNION merge query shape
  - args/path-prefix normalization behavior

Validation run:
- `go test ./pkg/datastore ./pkg/tenant`

srstack:
Updated per latest direction: vector search no longer uses . It now runs a single query with content-type routing:\n\n-  -> uses \n- otherwise -> uses \n\nThis removes dual-scan union overhead while keeping image/text vector support in one path.\n\nValidation:\n- ok  	github.com/mem9-ai/dat9/pkg/datastore	(cached)
ok  	github.com/mem9-ai/dat9/pkg/tenant	(cached)

srstack:
Updated per latest direction: vector search no longer uses `UNION ALL`. It now runs a single query with content-type routing:

- `content_type LIKE 'image/%'` -> uses `embedding_image`
- otherwise -> uses `embedding`

This removes dual-scan union overhead while keeping image/text vector support in one path.

Validation:
- `go test ./pkg/datastore ./pkg/tenant`

srstack:
Done. Added provider-owned schema version constants and wired provisioning to use them.

Changes:
- `tenant.CurrentSchemaVersion(provider)` now returns per-provider current schema version
- `server.handleProvision` now writes `SchemaVersion: tenant.CurrentSchemaVersion(provider)` instead of hardcoded `1`
- added unit test: `pkg/tenant/provider_test.go`

Current provider schema versions:
- `db9`: 1
- `tidb_zero`: 2
- `tidb_cloud_starter`: 1

Validation:
- `go test ./pkg/tenant ./pkg/server`

qiffang:
Review from three reviewers (myself + Claude Code + Codex):

**Verdict: NO-SHIP** — 2 blockers, 2 major, 1 minor.

---

### Blocker 1: `EMBED_IMAGE` may not exist in TiDB

`schema_zero.go` and `schema_starter.go` use:
```sql
embedding_image VECTOR(1024) GENERATED ALWAYS AS (EMBED_IMAGE(..., content_blob, ...)) STORED
```

TiDB Auto Embedding [documentation](https://docs.pingcap.com/ai/vector-search-auto-embedding-overview/) only defines `EMBED_TEXT()`, not `EMBED_IMAGE()`. If this function does not exist, `CREATE TABLE` will fail at tenant provisioning time.

Additionally, `tidbcloud_free/amazon/titan-embed-image-v1` is **not listed** in the [Amazon Titan available models page](https://docs.pingcap.com/ai/vector-search-auto-embedding-amazon-titan/). Only `titan-embed-text-v2` is listed as free/hosted.

**Action needed:** Verify `EMBED_IMAGE` exists by running a test DDL on a TiDB Cloud Starter cluster. If it does not exist, this entire approach needs redesign (likely server-side embedding via Bedrock API, stored in a non-GENERATED column).

---

### Blocker 2: `CASE WHEN` prevents vector index usage

`buildVectorSearchQuery` uses:
```sql
CASE WHEN f.content_type LIKE 'image/%' THEN VEC_EMBED_COSINE_DISTANCE(f.embedding_image, ?)
     ELSE VEC_EMBED_COSINE_DISTANCE(f.embedding, ?)
END AS distance
ORDER BY distance
```

TiDB vector index requires `ORDER BY VEC_*_DISTANCE(column, ?) LIMIT N` directly — no `CASE WHEN`, no alias. This query will **fall back to brute-force O(n) scan** instead of using the vector index.

**Suggested fix:** Split into two separate vector queries (text embedding + image embedding), run them concurrently (like the existing vector+FTS pattern), and RRF merge the results. This fits naturally into the existing `Grep()` architecture.

---

### Major 1: `embedding_image` computed for ALL rows, no type guard

The `GENERATED ALWAYS AS` expression runs `EMBED_IMAGE(model, content_blob)` for **every row**, including text files. This will either:
- Fail at INSERT time for non-image content
- Produce meaningless vectors from non-image data

**Suggested fix:** Use a conditional expression or only populate `embedding_image` for rows where `content_type LIKE 'image/%'`.

---

### Major 2: Deleted existing provider tests

`provider_test.go` removes `TestNormalizeProvider` and `TestSmallInDB` — these are regression-sensitive tests for provider routing logic. New tests should be **added alongside**, not replace them.

---

### Minor: `schema_db9.go` missing `embedding_image`

db9 schema does not get the new column. This is likely intentional (db9 uses PostgreSQL, not TiDB), but should be documented in the PR description.

---

**Summary of recommendations:**
1. Verify `EMBED_IMAGE` exists on TiDB Cloud before merging
2. Replace `CASE WHEN` with two concurrent vector queries + RRF merge
3. Add type guard to prevent image embedding on non-image files
4. Restore deleted provider tests
5. Document db9 scope exclusion in PR description

PR #64, opened 2026-03-29
feat: async image-to-text extraction for search

qiffang:
## Summary

Add background workers that extract text from uploaded images, making them searchable via `dat9 fs grep`.

## How it works

```
upload image → enqueue async task → worker reads image →
extract text (basic metadata or OpenAI vision API) →
update files.content_text → TiDB auto-embedding makes it vector-searchable
```

## Files

| File | Description |
|---|---|
| `pkg/backend/options.go` | Options struct, configureOptions, Close |
| `pkg/backend/image_extract.go` | Interface, enqueue, worker, processTask, extension fallback |
| `pkg/backend/image_extract_basic.go` | Basic metadata extractor + fallback wrapper |
| `pkg/backend/image_extract_openai.go` | OpenAI-compatible vision API extractor (1MB response cap) |
| `pkg/backend/image_extract_test.go` | Integration tests (happy path + stale revision race) |
| `pkg/backend/dat9.go` | Enqueue after create/overwrite, revision from UpdateFileContent |
| `pkg/backend/upload.go` | Enqueue after multipart ConfirmUpload |
| `pkg/datastore/store.go` | UpdateFileContent returns new revision; UpdateFileSearchText with revision guard |
| `pkg/tenant/pool.go` | BackendOptions, b.Close() in race loser + eviction |
| `cmd/dat9-server/main.go` | buildBackendOptionsFromEnv |

## Safety

- **Revision guard**: `WHERE revision=?` prevents stale async tasks from overwriting newer content
- **UpdateFileContent** returns actual new revision (SELECT after UPDATE) — no concurrent overwrite race
- **Pool race**: loser branch calls `b.Close()` before `st.Close()` to prevent leaked workers
- **Multipart**: `contentTypeFromPath` extension fallback when content_type is empty
- **Queue full**: drops task with warning log (best-effort enrichment)
- **OpenAI response**: capped at 1MB via `io.LimitReader`

## Config

| Variable | Description | Default |
|---|---|---|
| `DAT9_IMAGE_EXTRACT_ENABLED` | Enable/disable | `true` |
| `DAT9_IMAGE_EXTRACT_QUEUE_SIZE` | Queue size | `128` |
| `DAT9_IMAGE_EXTRACT_WORKERS` | Worker count | `1` |
| `DAT9_IMAGE_EXTRACT_MAX_BYTES` | Max image size | `8MB` |
| `DAT9_IMAGE_EXTRACT_TIMEOUT_SECONDS` | Task timeout | `20` |
| `DAT9_IMAGE_EXTRACT_API_BASE` | Vision API URL | (optional) |
| `DAT9_IMAGE_EXTRACT_API_KEY` | Vision API key | (optional) |
| `DAT9_IMAGE_EXTRACT_MODEL` | Vision model | (optional) |

## Review

4 rounds of cross-model review (Claude Code + Codex). All blockers resolved:
- Pool race loser leak (Codex)
- Multipart confirm missing enqueue (Codex)
- Checksum→revision guard (Codex)
- UpdateFileContent returns actual revision (Codex)
- SELECT revision error propagated (Codex)
- OpenAI response cap (Claude Code)
- Extension-based content_type fallback (Codex)

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,7 +79,7 @@ c.Delete("/data/file.txt")
 | `DAT9_S3_PREFIX` | S3 key prefix (e.g. `tenants/abc/`) | |
 | `DAT9_S3_ROLE_ARN` | IAM role ARN to assume via STS | |
 | `DAT9_S3_DIR` | Local S3 mock directory (only used without `DAT9_S3_BUCKET`) | `./s3` |
-| `DAT9_IMAGE_EXTRACT_ENABLED` | Enable async image->text extraction for search | `true` |
+| `DAT9_IMAGE_EXTRACT_ENABLED` | Enable async image->text extraction for search | `false` |
 | `DAT9_IMAGE_EXTRACT_QUEUE_SIZE` | In-memory queue size for extraction tasks | `128` |
 | `DAT9_IMAGE_EXTRACT_WORKERS` | Number of extraction workers | `1` |
 | `DAT9_IMAGE_EXTRACT_MAX_BYTES` | Max image bytes processed per task | `8388608` |
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ c.Delete("/data/file.txt")
 | `DAT9_IMAGE_EXTRACT_API_BASE` | OpenAI-compatible base URL (optional; set with key/model) | |
 | `DAT9_IMAGE_EXTRACT_API_KEY` | API key for image extraction provider | |
 | `DAT9_IMAGE_EXTRACT_MODEL` | Vision model name (for example Qwen VL model id) | |
-| `DAT9_IMAGE_EXTRACT_PROMPT` | Custom extraction prompt | |
+| `DAT9_IMAGE_EXTRACT_PROMPT` | Custom extraction prompt | `Describe this image for file search. Include key objects, scene, any visible text (OCR), and concise tags.` |
 | `DAT9_IMAGE_EXTRACT_MAX_TOKENS` | Max output tokens for model extraction | `256` |
 
 ## Architecture
--- a/README.md
+++ b/README.md
@@ -79,6 +79,17 @@ c.Delete("/data/file.txt")
 | `DAT9_S3_PREFIX` | S3 key prefix (e.g. `tenants/abc/`) | |
 | `DAT9_S3_ROLE_ARN` | IAM role ARN to assume via STS | |
 | `DAT9_S3_DIR` | Local S3 mock directory (only used without `DAT9_S3_BUCKET`) | `./s3` |
+| `DAT9_IMAGE_EXTRACT_ENABLED` | Enable async image->text extraction for search | `true` |
+| `DAT9_IMAGE_EXTRACT_QUEUE_SIZE` | In-memory queue size for extraction tasks | `128` |
+| `DAT9_IMAGE_EXTRACT_WORKERS` | Number of extraction workers | `1` |
+| `DAT9_IMAGE_EXTRACT_MAX_BYTES` | Max image bytes processed per task | `8388608` |
+| `DAT9_IMAGE_EXTRACT_TIMEOUT_SECONDS` | Timeout per extraction task | `20` |
+| `DAT9_IMAGE_EXTRACT_MAX_TEXT_BYTES` | Max extracted text stored into `files.content_text` | `8192` |
+| `DAT9_IMAGE_EXTRACT_API_BASE` | OpenAI-compatible base URL (optional; set with key/model) | |
+| `DAT9_IMAGE_EXTRACT_API_KEY` | API key for image extraction provider | |
+| `DAT9_IMAGE_EXTRACT_MODEL` | Vision model name (for example Qwen VL model id) | |
+| `DAT9_IMAGE_EXTRACT_PROMPT` | Custom extraction prompt | |
+| `DAT9_IMAGE_EXTRACT_MAX_TOKENS` | Max output tokens for model extraction | `256` |
 
 ## Architecture
 
--- a/cmd/dat9-server/main.go
+++ b/cmd/dat9-server/main.go
@@ -184,7 +184,7 @@ environment:
   DAT9_S3_ROLE_ARN IAM role ARN to assume via STS (optional)
   DAT9_S3_DIR      local s3 mock root directory (default: ./s3, only used without DAT9_S3_BUCKET)
   Image extraction (async image -> text for search):
-  DAT9_IMAGE_EXTRACT_ENABLED true|false (default: true)
+  DAT9_IMAGE_EXTRACT_ENABLED true|false (default: false)
   DAT9_IMAGE_EXTRACT_QUEUE_SIZE buffered task queue size (default: 128)
   DAT9_IMAGE_EXTRACT_WORKERS number of workers (default: 1)
   DAT9_IMAGE_EXTRACT_MAX_BYTES max image bytes processed per task (default: 8388608)
@@ -230,7 +230,7 @@ func publicBaseURL(listenAddr string) string {
 
 func buildBackendOptionsFromEnv() (backend.Options, error) {
 	var opts backend.Options
-	if !envBool("DAT9_IMAGE_EXTRACT_ENABLED", true) {
+	if !envBool("DAT9_IMAGE_EXTRACT_ENABLED", false) {
 		return opts, nil
 	}
 
--- a/cmd/dat9-server/main.go
+++ b/cmd/dat9-server/main.go
@@ -12,7 +12,9 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"time"
 
+	"github.com/mem9-ai/dat9/pkg/backend"
 	"github.com/mem9-ai/dat9/pkg/encrypt"
 	"github.com/mem9-ai/dat9/pkg/logger"
 	"github.com/mem9-ai/dat9/pkg/meta"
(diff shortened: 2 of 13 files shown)


srstack:
# PR #64 Review Report

## Overview
**PR**: #64 `feat: async image-to-text extraction for search`  
**Author**: @qiffang  
**Commit**: bf8088c  
**Changes**: +893 lines, -23 lines, 12 files

---

## Summary

Adds background workers that extract text descriptions from uploaded images, making them searchable via `grep` (FTS + vector search).

**Data Flow**:
```
Upload Image → Enqueue Task → Worker Extracts Text → 
Update files.content_text → TiDB Auto-Embedding Indexes It
```

---

## New Files

### 1. `pkg/backend/options.go`
- Configuration structures: `Options`, `AsyncImageExtractOptions`
- Default constants: queue size 128, 1 worker, 8MB image limit, 20s timeout
- `configureOptions()`: Initializes worker goroutines
- `Close()`: Graceful shutdown of background workers

### 2. `pkg/backend/image_extract.go` (Core)
**Interface**:
```go
type ImageTextExtractor interface {
    ExtractImageText(ctx context.Context, req ImageExtractRequest) (string, error)
}
```

**Key Functions**:
- `enqueueImageExtract()`: Adds task to queue (drops when full with warning)
- `runImageExtractWorker()`: Worker loop
- `processImageExtractTask()`: Processes individual task
- `loadImageBytesForExtract()`: Loads image bytes from S3 or DB
- `sanitizeExtractedText()`: Cleans and truncates text
- `contentTypeFromPath()`: Extension-based Content-Type fallback

### 3. `pkg/backend/image_extract_basic.go`
- `BasicImageTextExtractor`: Metadata-based lightweight extractor (dimensions, format)
- `FallbackImageTextExtractor`: Wrapper that falls back when primary fails

### 4. `pkg/backend/image_extract_openai.go`
- `OpenAIImageTextExtractor`: OpenAI-compatible Vision API extractor
- 1MB response limit
- Default prompt: "Describe this image for file search..."

### 5. `pkg/backend/image_extract_test.go`
- `TestAsyncImageExtractUpdatesContentText`: Basic functionality test
- `TestAsyncImageExtractSkipsStaleChecksum`: Stale task skip test

---

## Modified Files

### `pkg/backend/dat9.go`
- Added `imageExtract*` fields to `Dat9Backend` struct
- Updated constructors to support `Options` parameter

### `pkg/backend/upload.go`
- Calls `enqueueImageExtractForUpload()` after `ConfirmUpload`

### `pkg/datastore/store.go`
- Added `UpdateFileSearchText()`: Update with revision guard
```go
// WHERE revision=? prevents stale async tasks from overwriting
UPDATE files SET content_text = ? 
WHERE file_id = ? AND status = 'CONFIRMED' AND revision = ?
```

---

## Safety Design ✅

| Concern | Solution |
|---------|----------|
| **Stale task overwrite** | Revision guard: `WHERE revision=?` |
| **Concurrent write race** | `UpdateFileContent` returns actual new revision |
| **Queue full** | Drop task + warning log (best-effort, non-blocking) |
| **Large images** | Configurable `MaxImageBytes` limit (default 8MB) |
| **Long text** | `sanitizeExtractedText` truncates to max bytes |
| **Storage types** | Supports both inline DB and S3 storage |
| **Worker lifecycle** | `Close()` method for graceful shutdown |

---

## Configuration

Environment variables prefix: `DAT9_IMAGE_EXTRACT_*`

```go
AsyncImageExtractOptions{
    Enabled:       true/false
    QueueSize:     128 (default)
    Workers:       1 (default)
    MaxImageBytes: 8MB (default)
    TaskTimeout:   20s (default)
    MaxExtractTextBytes: 8KB (default)
    Extractor:     ImageTextExtractor (OpenAI or Basic)
}
```

---

## Code Quality Assessment

### Strengths ✅

1. **Clean Architecture**
   - Interface-based extractors, easy to extend
   - Clear separation of concerns: queue, worker, extractor

2. **Robust Error Handling**
   - Logs all failure cases
   - Graceful degradation when queue is full (doesn't block uploads)

3. **Concurrency Safety**
   - Revision guard prevents race conditions
   - WaitGroup manages worker lifecycle

4. **Testability**
   - Static extractor for testing
   - Gated extractor for concurrent scenario testing

5. **Configuration Flexibility**
   - Supports both OpenAI and Basic modes
   - All parameters configurable

### Suggested Improvements 🟡

1. **Metrics**
   ```go
   // Suggested additions:
   - image_extract_tasks_total (counter)
   - image_extract_queue_size (gauge)
   - image_extract_duration_seconds (histogram)
   - image_extract_errors_total (counter)
   ```

2. **Retry Mechanism**
   - Currently drops on failure, could add limited retry

3. **Priority Queue**
   - Currently FIFO, could consider file size/user priority in future

4. **Documentation**
   - README is updated, could add architecture diagram

---

## Test Coverage

| Test | Description |
|------|-------------|
| `TestAsyncImageExtractUpdatesContentText` | Verifies extracted text is written to content_text |
| `TestAsyncImageExtractSkipsStaleChecksum` | Verifies revision guard prevents stale overwrites |

**Suggested Additions**:
- Queue-full behavior test
- S3 storage image extraction test
- Large image skip test
- Extractor error fallback test

---

## Recommendation

**✅ Approve for Merge**

This is a high-quality PR:
- Clean architecture with well-designed interfaces
- Comprehensive safety considerations
- Backward compatible (opt-in feature, disabled by default)
- Production-ready configuration options

### Optional Pre-Merge Improvements
- [ ] Add Prometheus metrics (suggest follow-up PR)
- [ ] Add more edge case tests (suggest follow-up PR)

### Conclusion
Excellent code quality, thoughtful design. Recommend immediate merge. 🚀

---

## Key Code Snippets

**Revision Guard (Core Safety Mechanism)**:
```go
func (s *Store) UpdateFileSearchText(ctx context.Context, fileID string, 
    expectedRevision int64, contentText string) (updated bool, err error) {
    res, err = s.db.ExecContext(ctx, 
        `UPDATE files SET content_text = ?
         WHERE file_id = ? AND status = 'CONFIRMED' AND revision = ?`,
        nullStr(contentText), fileID, expectedRevision)
    n, _ := res.RowsAffected()
    return n > 0, nil
}
```

**Worker Loop**:
```go
func (b *Dat9Backend) runImageExtractWorker(ctx context.Context) {
    defer b.imageExtractWG.Done()
    for {
        select {
        case <-ctx.Done():
            return
        case task := <-b.imageExtractQueue:
            b.processImageExtractTask(ctx, task)
        }
    }
}
```

srstack:
Thanks for the update — I reviewed the latest changes. I like the overall direction, but I see two issues before merge:

1. **Blocker: CI is currently red (typecheck failure)**
   - `pkg/datastore/store_test.go:305` still calls `UpdateFileContent` with the old signature.
   - `UpdateFileContent` now returns `(newRevision int64, err error)`, so tests/callers must be updated.
   - Current failure from Actions: `assignment mismatch: 1 variable but s.UpdateFileContent returns 2 values`.

2. **Major: non-multipart image writes can skip extraction when Content-Type is missing**
   - `enqueueImageExtract(...)` gates only on `contentType`.
   - For multipart uploads, you added `contentTypeFromPath(...)` fallback (good), but regular `Write/overwrite` paths still rely on incoming content-type only.
   - If a client uploads `*.png` without `Content-Type`, extraction is silently skipped.
   - Suggestion: apply the same extension-based fallback in the regular write/overwrite enqueue path as well.

Once those are fixed, I’m happy to re-check quickly.

qiffang:
Both issues fixed in 394fb37:

**1. CI typecheck (blocker):**
`store_test.go:305` updated to match new `UpdateFileContent` signature `(int64, error)`. Test now captures `newRev` and asserts `newRev == 2`.

**2. Extension fallback in all enqueue paths (major):**
`enqueueImageExtract` now applies `contentTypeFromPath(path)` fallback for ALL callers — not just multipart. If the passed `contentType` is not `image/*`, it tries path extension before skipping. This covers the case where a client uploads `foo.png` without `Content-Type` header.

Note: the small-file write path already detects content type via `detectContentType(path, data)` which uses `mime.TypeByExtension`, so this is defense-in-depth rather than a live bug. But it makes the extraction layer self-contained and not dependent on upstream content-type detection.

srstack:
Re-checked after `394fb37` — thanks for fixing the two earlier items.

I have one additional major concern before merge:

### Major: stale tasks can still trigger expensive extraction/API calls
In `processImageExtractTask`, the revision guard is only enforced at `UpdateFileSearchText(..., expectedRevision)` time.
That protects correctness, but stale tasks still do the expensive work first (load blob + call extractor), then get dropped.

Current flow for stale task:
1. task enqueued with old revision
2. file updated again
3. worker loads current bytes and calls extractor
4. update fails revision check (`updated=false`)

For OpenAI mode this can cause unnecessary external cost and latency spikes.

**Suggested fix:** short-circuit before extraction:
- after `GetFile`, if `task.Revision > 0 && f.Revision != task.Revision`, return immediately (skip load/extract)

Everything else looks much better now.

srstack:
Thanks for the quick follow-up — I re-checked the latest head (`0703e83`).

- CI is now green ✅
- `UpdateFileContent` signature usage in tests is fixed ✅
- content-type extension fallback in enqueue path is fixed ✅

I still see one remaining concern from my previous comment:

### Remaining major: stale tasks still do expensive work before being dropped
In `pkg/backend/image_extract.go`, `processImageExtractTask` still loads bytes + calls extractor before the revision guard is applied in `UpdateFileSearchText(...)`.

That means stale queued tasks can still trigger unnecessary OpenAI/vision calls and cost, even though final DB update is rejected.

Suggested small fix:
```go
if task.Revision > 0 && f.Revision != task.Revision {
    return
}
```
Place it right after `GetFile`/status checks and before `loadImageBytesForExtract(...)`.

Once this is added, I’m good with merge.

srstack:
Quick re-check on latest head (`1d0fdf8`):

- CI is currently pending, not completed yet.
- I can confirm the stale-task early short-circuit is still missing in `processImageExtractTask` (it still loads bytes and calls extractor before revision mismatch is checked).

Please add this before `loadImageBytesForExtract(...)`:

```go
if task.Revision > 0 && f.Revision != task.Revision {
    metrics.RecordOperation("image_extract", "process", "stale_precheck", time.Since(start))
    return
}
```

After that + green CI, I’m good.

PR #75, opened 2026-03-30
Implement async embedding pipeline and worker runtime

JaySon-Huang:
## Summary
- add revision-aware async embedding generation with durable semantic tasks and background worker processing
- move semantic grep to app-side query embedding plus FTS/vector RRF merge, with keyword fallback when no ranked results are available
- add local single-tenant validation tooling and semantic worker observability, including configurable retry caps

## Key changes
- add mutable embedding state on `files` via `embedding` and `embedding_revision`
- add tenant-local `semantic_tasks` durable queue with claim/ack/retry/recover semantics
- enqueue semantic work transactionally from create/overwrite/upload/image-text update paths
- start semantic workers from `dat9-server` when `DAT9_EMBED_*` is configured
- add `dat9-server-local` and `make run-server-local` for single-tenant end-to-end validation
- add semantic worker metrics/logging and make retry cap configurable

## Breaking changes / rollout notes
- tenant data-plane schema must be migrated before enabling async embedding
- TiDB/MySQL semantic runtime now assumes TiDB capabilities for vector and FTS features
- `pkg/datastore.Store.Grep` has been replaced by `FTSSearch`, `VectorSearch`, and `KeywordSearch`
- enabling `DAT9_EMBED_*` starts background semantic workers; if `DAT9_QUERY_EMBED_*` is unset, grep reuses the semantic embedder
- embedding model dimensions must match the current schema (`VECTOR(1024)`)

## New config
- query embedding:
  - `DAT9_QUERY_EMBED_API_BASE`
  - `DAT9_QUERY_EMBED_API_KEY`
  - `DAT9_QUERY_EMBED_MODEL`
  - `DAT9_QUERY_EMBED_DIMENSIONS`
  - `DAT9_QUERY_EMBED_TIMEOUT_SECONDS`
- background embedding worker:
  - `DAT9_EMBED_API_BASE`
  - `DAT9_EMBED_API_KEY`
  - `DAT9_EMBED_MODEL`
  - `DAT9_EMBED_DIMENSIONS`
  - `DAT9_EMBED_TIMEOUT_SECONDS`
  - `DAT9_SEMANTIC_WORKERS`
  - `DAT9_SEMANTIC_POLL_INTERVAL_MS`
  - `DAT9_SEMANTIC_LEASE_SECONDS`
  - `DAT9_SEMANTIC_RECOVER_INTERVAL_MS`
  - `DAT9_SEMANTIC_RETRY_BASE_MS`
  - `DAT9_SEMANTIC_RETRY_MAX_MS`
  - `DAT9_SEMANTIC_TENANT_LIMIT`
  - `DAT9_SEMANTIC_PER_TENANT_CONCURRENCY`
- local validation:
  - `DAT9_LOCAL_DSN`
  - `DAT9_LOCAL_INIT_SCHEMA`

## Testing
- `make test-podman`
- local single-tenant end-to-end validation with `make run-server-local`


JaySon-Huang:
Overall assessment: for the TiDB path, this is already a reasonably coherent async-embedding implementation. The `semantic_tasks` lease/receipt/recover contract is in place, `embedding_revision` is being used as the revision gate, and the main write-path / worker / query-path flow is wired together. I also ran `make test-podman`, and the current test suite is green.

That said, measured against the design docs themselves, I would not say the implementation fully reaches the stated target yet. The main gap is not the queue substrate; it is target alignment. The docs position Phase 1 as a foundation for both TiDB/MySQL and DB9/Postgres, but the runtime code and SQL are still clearly TiDB/MySQL-first. On top of that, the image-extract bridge and some enqueue rules still diverge from the intended behavior in ways that can matter operationally. So the most accurate framing is: this PR makes the TiDB-first Phase 2/3 path largely workable, but it is not yet the clean provider-neutral Phase 1 foundation described in the proposals.

pkg/server/semantic_worker.go
```diff
@@ -0,0 +1,471 @@
...
+	m.mu.Unlock()
+
+	if ref.tenant == nil {
+		return nil, fmt.Errorf("tenant metadata missing for %s", ref.id)
+	}
+	pass, err := m.pool.Decrypt(ctx, ref.tenant.DBPasswordCipher)
+	if err != nil {
+		return nil, fmt.Errorf("decrypt tenant password: %w", err)
+	}
+	store, err := datastore.Open(tenantDSN(ref.tenant.DBUser, string(pass), ref.tenant.DBHost, ref.tenant.DBPort, ref.tenant.DBName, ref.tenant.DBTLS))
```
JaySon-Huang, line 470: [P1] This runtime path still assumes every tenant datastore is reachable through a MySQL DSN and `datastore.Open()`. That means DB9/Postgres tenants are not actually supported end-to-end even though the Phase 1 docs and schema work present DB9 as a supported provider. Right now the provider split exists at DDL level, but not in the running worker path.

JaySon-Huang (commented):
File-level finding.

pkg/backend/image_extract.go
```diff
@@ -223,6 +223,12 @@ func (b *Dat9Backend) processImageExtractTask(ctx context.Context, task imageExt
 		metrics.RecordOperation("image_extract", "process", "stale", time.Since(start))
 		return
 	}
+	if _, err := b.store.EnsureSemanticTaskQueued(ctx, newEmbedTask(b.genID(), task.FileID, task.Revision, time.Now().UTC())); err != nil {
```
JaySon-Huang: [P2] The image-to-embed bridge is not atomic here. `content_text` is updated first, and the durable embed task is ensured afterward in a separate step. If the process exits between those two calls, the file keeps the new FTS text but never gets a vector task, which is exactly the regression the proposal is trying to prevent.

JaySon-Huang (commented):
File-level finding.

pkg/backend/upload.go
```diff
@@ -234,20 +237,45 @@ func (b *Dat9Backend) ConfirmUpload(ctx context.Context, uploadID string) error
...
 			Path:       upload.TargetPath,
 			ParentPath: pathutil.ParentPath(upload.TargetPath),
 			Name:       pathutil.BaseName(upload.TargetPath),
 			FileID:     upload.FileID,
 			CreatedAt:  time.Now(),
-		})
+		}); err != nil {
+			return err
+		}
+		return b.enqueueEmbedTaskTx(tx, confirmedFileID, confirmedRevision)
```
JaySon-Huang: [P2] This path enqueues `embed` even though the same transaction sets `content_text = NULL` and there is no current text producer for the revision. The worker will eventually ack the task as obsolete, so correctness is preserved, but it still creates guaranteed-noop queue traffic and inflates queue/lag metrics. The Phase B write-path doc explicitly calls out skipping enqueue in this case.

JaySon-Huang (commented):
File-level finding.

pkg/datastore/schema_test.go
```diff
@@ -10,22 +10,36 @@ func TestInitSchemaProviderSplit(t *testing.T) {
...
+		{provider: "db9", wantContentBlob: false},
 	}
 
-	dropDataPlaneTables(t, s1)
-	initDatastoreSchema(t, testDSN, "db9")
-	if s1.columnExists("files", "content_blob") {
-		t.Fatal("did not expect content_blob column for db9")
+	for _, tc := range tests {
+		dropDataPlaneTables(t, s1)
+		initDatastoreSchema(t, testDSN, tc.provider)
```
JaySon-Huang, line 28: [P2] This test still only exercises the MySQL helper schema. It does not execute `initDB9Schema`, open a Postgres-backed store, or validate any Postgres-specific SQL. That makes the coverage much weaker than the docs imply for the TiDB/MySQL + DB9 contract.

JaySon-Huang (commented):
File-level finding.

pkg/backend/image_extract.go
JaySon-Huang: Fixed in `7dbe836`.

I made the image bridge atomic by adding `UpdateFileSearchTextTx(...)` in `pkg/datastore/store.go` and changing `pkg/backend/image_extract.go` to perform the revision-gated `content_text` update and `EnsureSemanticTaskQueuedTx(...)` inside the same `Store.InTx(...)` block. I also added datastore seam tests and reran the image extract regression tests under `make test-podman` coverage.

pkg/backend/upload.go
JaySon-Huang: Fixed in `67002a7`.

I tightened write-path enqueue eligibility for both small writes and multipart completion. We now enqueue `embed` only when the revision already has synchronous `content_text` or has a known async image-text source; generic multipart/object-only paths no longer create guaranteed-noop queue traffic. I also updated the backend tests to cover the negative cases.

pkg/server/semantic_worker.go
JaySon-Huang: Agreed. I did not expand this PR into provider-neutral runtime support, but I recorded the gap explicitly as a follow-up TODO in `pkg/server/semantic_worker.go` in `2f2b8b0`.

The current worker runtime path is still MySQL/TiDB-oriented (`datastore.Open + tenantDSN`), while DB9/Postgres support in this branch is only at the schema/contract layer. I think that needs a dedicated follow-up rather than being folded into this PR.

pkg/datastore/schema_test.go
JaySon-Huang: Agreed. I did not try to add real Postgres-backed test infra in this PR, but I added an explicit follow-up TODO in `pkg/datastore/schema_test.go` in `2f2b8b0`.

Right now the test only validates provider-shape differences through the MySQL fixture used by the current runtime path; it does not prove `initDB9Schema` or Postgres runtime SQL end-to-end. That deserves a dedicated test/runtime follow-up.

qiffang (commented):
## Overall Assessment

This is a well-designed PR that establishes a solid foundation for async embedding generation. The architecture correctly addresses the core challenges: revision-aware state, durable task delivery, and transactional consistency.

### 👍 Strengths

1. **Revision-gated embedding state**: `embedding_revision` ensures stale vectors are never returned by search
2. **Durable task queue**: `semantic_tasks` table with claim/ack/recover semantics is a clean adaptation of QueueFS concepts
3. **Transactional atomicity**: File persistence and task enqueue commit together
4. **Worker lifecycle decoupled from Pool**: `SemanticWorkerManager` avoids LRU eviction interrupting background processing
5. **Image extract bridge**: Existing async image text flow properly triggers embed tasks

### ⚠️ Issues to Address

#### 1. Test coverage gap (`pkg/backend/schema_test_helper.go`)

Test schemas use `LONGTEXT` instead of `VECTOR` for embedding column. This means `VEC_EMBED_COSINE_DISTANCE` and actual vector distance computations are never tested in unit tests. Consider adding integration tests against a real TiDB instance with VECTOR support.

#### 2. DB9/Postgres runtime incomplete (`pkg/server/semantic_worker.go:403`)

The TODO comment is critical:
```go
// TODO(async-embedding): This worker runtime path still assumes every tenant
// datastore can be opened through the MySQL-oriented datastore.Open + tenantDSN
// flow.
```
DB9 tenants will silently fail to process semantic tasks. Either gate DB9 tenants from semantic worker processing (with a warning log), or prioritize this as a fast-follow.

#### 3. Grep fallback logic (`pkg/backend/dat9.go:355`)

Current logic: fallback to keyword search only when both FTS and vector results are empty.

Edge case: FTS returns empty (no matches), vector search errors. User gets empty results instead of keyword fallback.

Suggested fix:
```go
if len(ftsResp.rows) == 0 && (len(vecResp.rows) == 0 || vecResp.err != nil) {
    return s.KeywordSearch(...)
}
```

#### 4. Hardcoded retry delay cap (`pkg/server/semantic_worker.go:29`)

`maxSemanticRetryDelay = 30 * time.Second` may be too short for embedding APIs under rate limiting. Consider exposing this as a configurable option in `SemanticWorkerOptions`.

### Summary

Solid foundation work. The core architecture is sound. Recommend addressing the grep fallback logic before merge; the other items can be follow-ups with tracking issues.

qiffang:
## Consolidated Review Feedback

After discussing with the team, here is the summary:

### 🚫 Blockers (recommend fixing before merge)

**1. Grep fallback logic bug (`pkg/backend/dat9.go`)**

Current behavior:
- FTS empty + Vector error → returns empty results (should fallback)
- FTS empty + Vector empty → returns empty results (should fallback)

Expected: When either FTS or Vector fails/empty, fallback to keyword search.

Suggested fix:
```go
if len(ftsResp.rows) == 0 && (len(vecResp.rows) == 0 || vecResp.err != nil) {
    return s.KeywordSearch(...)
}
```

**2. DB9/Postgres multi-tenant guard**

The TODO at `semantic_worker.go:403` indicates DB9 tenants will silently fail. Before merge:
- Add explicit provider check in worker startup
- Skip DB9 tenants with clear log message
- Avoid silent task churn

### 📋 Follow-ups (can be separate PRs)

1. **Integration test** — Add CI job connecting to real TiDB with VECTOR support to verify embed write + vector search
2. **Configurable retry delay** — Make `maxSemanticRetryDelay` configurable, support `Retry-After` header for 429 responses  
3. **Migration doc** — Optional backfill job/doc for existing tenants

---

Overall architecture is solid. Would appreciate addressing the two blockers before merge.

qiffang (commented):
## Code Review Summary

整体设计非常扎实 👍 以下是详细 review：

### ✅ 优点

1. **租户隔离 + 公平调度**: `SemanticWorkerManager` 实现 round-robin 调度，`PerTenantConcurrency` 防止单租户垄断 worker
2. **Lease 机制完善**: claim 时生成 receipt + lease_until，ack/retry 都验证 receipt，防止重复执行
3. **Revision 版本控制**: `embedding_revision` 字段跟踪 embedding 对应的文件版本，写回时 conditional update 防止覆盖新版本（经确认 `WHERE revision = ? AND embedding_revision < ?` 正确处理了 race condition）
4. **Dead letter 设计**: `max_attempts` + `dead_lettered` 状态，任务失败有终态
5. **测试覆盖全面**: 覆盖正常流程、stale revision、dead letter、lease recovery 等场景

### 💡 建议改进

#### 1. Vector Search 查询条件

`embedding_revision = revision` 条件导致文件修改后、embedding 重新生成前，文件完全无法被搜索。

**建议**:
- 当前阶段可以先不做 fallback（async embedding 延迟应该很低）
- 在 API response 中返回 `pending_reindex_count`，让 UI 显示 "X 个文件正在重新索引"
- 后续根据用户反馈决定是否需要 fallback 到 stale embedding

#### 2. FormatVector NaN/Inf 处理

静默替换为 0 可能丢失信息，建议在替换时打 warning 日志，方便排查 model 返回异常值的问题。

#### 3. Conditional Update 失败时的日志

当 `rowsAffected == 0`（文件在 embedding 生成期间被修改）时，建议 log.Info "discarding stale embedding, file was modified during generation"，方便 debugging。

#### 4. Store 连接健康检查

可以在 `storeForRef` 里加个简单的 `db.PingContext(ctx)` 健康检查，失败时立即 invalidate + retry，比 TTL-based refresh 更轻量。当前阶段 reactive invalidation 足够用。

### 📝 Notes

- `TODO(async-embedding)` 关于 Postgres 端到端支持需要在 DB9 上线前解决
- Schema migration 需要确认是否有 migration plan（移除了 GENERATED ALWAYS 列定义）

---

**Conclusion**: 架构设计合理，lease + receipt + revision 三重保护覆盖了主要并发场景。以上建议都是 enhancement 而非 blocker。LGTM 👍

pkg/server/semantic_worker.go
qiffang: Added a TODO comment on `targetForRef` in #173 noting DB9/Postgres tenants are not yet supported in the semantic worker runtime path.

pkg/datastore/schema_test.go
qiffang: Acknowledged. Existing TODO at lines 8-10 of `schema_test.go` already documents this gap. DB9/Postgres test infra will be addressed in a future PR.

PR #76, opened 2026-03-31
[WIP] tighten async embedding phase-b write-path behavior

JaySon-Huang:
## Summary
- align small-file and multipart write paths with the phase-b embed enqueue policy so revisions without a current or future text source do not create guaranteed-noop embed tasks
- keep the small-write cleanup path safe by preserving the inline marker object behavior while retaining the async image bridge atomicity already rebased from phase-a
- add backend regressions for small-write negative cases, multipart overwrite behavior, and inline cleanup safety

## Testing
- `source ./scripts/test-podman.sh && go test ./pkg/backend -run 'TestShouldEnqueueEmbedForRevision|TestWriteCreateEnqueuesEmbedTask|TestWriteCreateSkipsEmbedTaskWithoutTextSource|TestWriteOverwriteEnqueuesNextRevisionAndClearsEmbeddingState|TestWriteOverwriteSkipsEmbedTaskWithoutTextSource|TestConfirmUploadSkipsEmbedTaskWithoutTextSource|TestConfirmUploadOverwriteSkipsEmbedTaskWithoutTextSourceAndRebindsUpload' -count=1`
- `source ./scripts/test-podman.sh && go test ./pkg/backend -run 'TestAsyncImageExtractUpdatesContentText|TestAsyncImageExtractRequeuesSucceededEmbedTask|TestAsyncImageExtractStaleResultDoesNotRequeueOldRevision' -count=1`

### DIFF
--- a/pkg/backend/dat9.go
+++ b/pkg/backend/dat9.go
@@ -429,9 +429,7 @@ func (b *Dat9Backend) overwriteFileCtx(ctx context.Context, nf *datastore.NodeWi
 		}
 		return 0, err
 	}
-	if nf.File.StorageRef != "" && nf.File.StorageRef != storageRef {
-		b.deleteBlobCtx(ctx, nf.File.StorageRef)
-	}
+	b.deleteBlobIfS3Ctx(ctx, nf.File.StorageType, nf.File.StorageRef, storageRef)
 	b.enqueueImageExtract(nf.File.FileID, nf.Node.Path, contentType, newRev)
 	return int64(len(data)), nil
 }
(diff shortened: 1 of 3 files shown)


JaySon-Huang:
Second-opinion review on PR 76 against `docs/async-embedding/async-embedding-phase-2-write-path.md`: I did not find any new blocking issues in this incremental diff.

The cleanup change looks aligned with the phase-2 failure-handling contract: old blobs are now deleted only after commit and only when the replaced object was actually an S3 blob. The added regressions also strengthen two important invariants from the proposal: preserving the inline marker object on overwrite cleanup, and keeping copy/rename as semantic no-ops that do not create extra embed tasks.

I also re-ran the targeted backend tests from the PR on the PR 76 worktree, and they passed.

PR #79, opened 2026-03-31
Phase 3 async embedding worker and search correctness

JaySon-Huang:
## Summary
- add the Phase 3 proposal that locks the worker/search rollout scope, decisions, and deferred provider-neutral follow-up
- make vector search SQL contract testable and verify it only consumes app-generated query vectors for the current file revision
- add backend regressions for grep query embedding fallback and write-path semantic task invariants around overwrite, rename, and copy

## Testing
- make test-podman

### DIFF
--- a/docs/async-embedding/async-embedding-generation-proposal.md
+++ b/docs/async-embedding/async-embedding-generation-proposal.md
@@ -351,6 +351,8 @@ Not acceptable. Embedding is already a prerequisite for search availability, not
 17. Add log fields: `tenant_id`, `task_id`, `resource_id`, `resource_version`, `receipt`, `attempt_count`
 18. After `embed` is stable, reuse the same task substrate to add `extract_text`, `generate_l0`, and `generate_l1`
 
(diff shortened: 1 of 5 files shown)


JaySon-Huang:
Second-opinion review on PR 79: I did not find any new blocking issues in this diff.

The changes are well aligned with the stated Phase 3 goal. `pkg/backend/grep_test.go` now proves grep calls the app-side query embedder before falling back, and `pkg/datastore/search.go` plus `pkg/datastore/search_test.go` make the vector-search SQL contract directly testable, including the `embedding_revision = revision` filter and the precomputed-query-vector boundary. The new Phase 3 proposal is also more explicit about deferred provider-neutral runtime support instead of overstating current coverage.

I re-ran `go test ./pkg/backend ./pkg/datastore -count=1` and `make test-podman` on the PR 79 worktree; both passed. The only residual gap I see is that the new tests lock the SQL contract shape rather than adding a full current-vs-stale vector execution smoke, which feels like follow-up hardening rather than a merge blocker.

PR #81, opened 2026-03-31
Feat/async embedding phase c

JaySon-Huang:


README.md
```diff
@@ -197,6 +197,32 @@ go build -o dat9 ./cmd/dat9
...
+
+```bash
+export DAT9_LOCAL_DSN='root@tcp(127.0.0.1:4000)/dat9_local?parseTime=true'
+make run-server-local
+```
+
+The helper script sets sensible defaults for local validation, including:
+
+- `DAT9_LOCAL_DSN`
+- `DAT9_LOCAL_INIT_SCHEMA=true`
```
JaySon-Huang: This default is risky together with the remote-DSN example above. The helper script sets `DAT9_LOCAL_INIT_SCHEMA=true`, and `cmd/dat9-server-local/main.go:54-58` immediately runs tenant schema initialization on startup, so following the documented workflow can mutate an existing shared database. Either make schema init opt-in for this local entrypoint or narrow the docs to disposable local databases and add an explicit warning before the remote-DSN example.

Source: rule/XUHUAIYU-PR-004 | [Second Opinion](https://github.com/PingCAP-QE/second-opinion)

scripts/dat9-server-local-env.sh
```diff
@@ -0,0 +1,78 @@
...
+: "${DAT9_SEMANTIC_POLL_INTERVAL_MS:=200}"
+: "${DAT9_SEMANTIC_LEASE_SECONDS:=30}"
+: "${DAT9_SEMANTIC_RECOVER_INTERVAL_MS:=5000}"
+: "${DAT9_SEMANTIC_RETRY_BASE_MS:=200}"
+: "${DAT9_SEMANTIC_RETRY_MAX_MS:=30000}"
+: "${DAT9_SEMANTIC_PER_TENANT_CONCURRENCY:=1}"
+
+# Query embedding.
+# Leave these unset to reuse the background embedder configured above.
+: "${DAT9_QUERY_EMBED_API_BASE:=${DAT9_EMBED_API_BASE}}"
```
JaySon-Huang: These defaults prevent the local validation flow from exercising the embedder-reuse path described in `cmd/dat9-server-local/main.go:74-78`. Because the helper eagerly exports `DAT9_QUERY_EMBED_*`, `buildBackendOptionsFromEnv()` always builds a separate query client, so the default workflow can mask bugs in the actual fallback/reuse startup policy. Leave `DAT9_QUERY_EMBED_*` unset by default here and only set them when the caller explicitly wants a separate query embedder.

JaySon-Huang: Adjusted this in `0d298c8`. The helper script no longer exports `DAT9_QUERY_EMBED_*` by default, so `dat9-server-local` now exercises the same embedder-reuse startup path as `dat9-server` when only `DAT9_EMBED_*` is configured. Callers can still set `DAT9_QUERY_EMBED_*` explicitly when they want a separate query embedder.

README.md
JaySon-Huang: Adjusted this in `0d298c8`. `DAT9_LOCAL_INIT_SCHEMA` now defaults to `false`, and the README examples explicitly require opting in with `export DAT9_LOCAL_INIT_SCHEMA=true` only for disposable local databases. This keeps the local entrypoint from mutating an existing shared/remote database by default.

PR #87, opened 2026-03-31
ci: add CLI release workflow

qiffang:
## Summary
- Adds `.github/workflows/release-cli.yml` that cross-compiles the dat9 CLI for 4 platforms (linux/amd64, linux/arm64, darwin/amd64, darwin/arm64)
- Publishes binaries + checksums + version file to `dat9-fe` site repo, served via Netlify CDN at `dat9.ai/releases/`
- Triggered on push to main (code paths) and manual workflow_dispatch

## Setup required
- Add `DAT9_FE_DEPLOY_TOKEN` secret to this repo (fine-grained PAT with `Contents: write` on `mem9-ai/dat9-fe`)

## Test plan
- [ ] Add the `DAT9_FE_DEPLOY_TOKEN` secret
- [ ] Trigger workflow manually via `workflow_dispatch`
- [ ] Verify `dat9-fe` repo has `site/releases/` with all 4 binaries, `version`, and `checksums.txt`
- [ ] Test installer: `curl -fsSL https://dat9.ai/install | sh`

### DIFF
--- a/.github/workflows/release-cli.yml
+++ b/.github/workflows/release-cli.yml
@@ -11,6 +11,10 @@ on:
       - 'Makefile'
   workflow_dispatch:
 
+concurrency:
+  group: release-cli
+  cancel-in-progress: true
+
 jobs:
   release:
     name: Build and publish CLI
@@ -68,4 +72,5 @@ jobs:
           git add site/releases/
           git diff --cached --quiet && echo "No changes to commit" && exit 0
           git commit -m "release: dat9 CLI ${{ env.VERSION }}"
+          git pull --rebase
           git push
--- a/.github/workflows/release-cli.yml
+++ b/.github/workflows/release-cli.yml
@@ -1,0 +1,71 @@
+name: Release CLI binaries
+
+on:
+  push:
+    branches: [main]
+    paths:
+      - 'cmd/**'
+      - 'pkg/**'
+      - 'go.mod'
+      - 'go.sum'
+      - 'Makefile'
+  workflow_dispatch:
+
+jobs:
+  release:
+    name: Build and publish CLI
+    runs-on: ubuntu-latest
+    timeout-minutes: 20
+
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+
+      - name: Checkout agfs dependency
+        run: |
+          git clone --depth 1 https://github.com/c4pt0r/agfs ../agfs
+
+      - name: Setup Go
+        uses: actions/setup-go@v5
+        with:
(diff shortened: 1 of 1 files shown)


srstack:
Nice work overall — the release workflow shape looks good. I have two blocking concerns before merge:

1. **Non-deterministic release input** (`release-cli.yml`, clone step)
   - The workflow clones `c4pt0r/agfs` at floating HEAD (`git clone --depth 1 ... ../agfs`), while this repo uses `replace ... => ../agfs/agfs-server`.
   - This means release artifacts can change (or fail) without any change in this repo.
   - Suggestion: pin `agfs` to an explicit commit/tag in CI (`git -C ../agfs checkout <sha>`), and keep that SHA auditable.

2. **No concurrency guard on publisher job**
   - Multiple `main` pushes can run concurrently and race while committing/pushing `site/releases/*` to `dat9-fe`.
   - One run can fail with non-fast-forward, leaving a commit’s release unpublished.
   - Suggestion: add `concurrency` (e.g. `group: release-cli`, `cancel-in-progress: true`) and optionally add a pull/rebase+retry before push.

Everything else (build matrix, version/checksum artifacts, publish flow) looks solid.

qiffang:
Thanks for the review! Both points are valid. Here's what I've done:

**1. Concurrency guard** — Fixed. Added `concurrency: { group: release-cli, cancel-in-progress: true }` at the workflow level, plus `git pull --rebase` before pushing to `dat9-fe` to handle any remaining race window. (2edf54e)

**2. agfs pin** — Acknowledged but deferring. The floating HEAD clone is the same pattern used in `code-ci.yml` and `dev-cd.yml`. Pinning it only in the release workflow would be inconsistent. I'd prefer to address this repo-wide in a follow-up PR that pins agfs across all three workflows.

srstack:
Thanks for the quick update — the concurrency + Already up to date. change looks good, and that resolves the race concern from my side.\n\nFor the agfs pin: I’m okay with deferring **if** we track it explicitly. Please open a follow-up issue/PR to pin agfs consistently across , , and this release workflow, then link it here so it doesn’t get lost.

srstack:
Correction (formatting):

Thanks for the quick update — the concurrency + `git pull --rebase` change looks good, and that resolves the race concern from my side.

For the agfs pin: I’m okay with deferring **if** we track it explicitly. Please open a follow-up issue/PR to pin agfs consistently across `.github/workflows/code-ci.yml`, `.github/workflows/dev-cd.yml`, and `.github/workflows/release-cli.yml`, then link it here so it doesn’t get lost.

PR #89, opened 2026-03-31
auto embedding mode for TiDB providers

JaySon-Huang:
## Summary

This PR implements the production-oriented auto embedding mode for TiDB providers on top of `feat/async-embedding-phase-a`.

It includes:
- the design proposal in `docs/auto-embedding-mode.md`
- the 7 implementation commits from the reviewed split plan
- TiDB schema-contract validation before rollout
- provider-level rollout switch only after backend/datastore/worker paths were fully split

## Commit Guide

1. `d1514eb` backend/tenant semantic mode foundation
2. `b56168a` datastore/backend search path split for auto embedding
3. `093afba` datastore/backend write path split for auto embedding
4. `5195da5` image extract split from embed bridge in auto mode
5. `a5bcf4d` server semantic worker skips auto-embedding tenants
6. `e046b57` tenant TiDB auto-embedding schema contract and starter validation
7. `ef6c0d7` enable TiDB providers on database auto embedding mode

## Notes

- `StarterProvisioner.InitSchema()` is now validate-only instead of a no-op.
- TiDB provider backends fail fast if the schema does not satisfy the database-managed auto-embedding contract.
- semantic worker is explicitly disabled for auto-embedding tenants to avoid duplicate app-side embedding work.
- one unrelated test assumption in `pkg/server/auth_test.go` was corrected because it used a plain MySQL test tenant while pretending to be a TiDB auto-embedding tenant.

## Testing

- `make test-podman`

## Review

A Chinese review guide was prepared locally during implementation and follows the same commit order used above.


pkg/server/semantic_worker.go
JaySon-Huang, line 1: [P2] Auto-embedding fallback disables the worker for every tenant.

This constructor now returns `nil` as soon as the fallback backend uses database auto-embedding. That is fine for the pure single-tenant local case, but `NewWithConfig()` also passes `Meta` and `Pool` here for multi-tenant mode. In a hybrid server config, one auto-embedding fallback backend now suppresses the semantic worker for all tenants, including `db9` tenants that still require app-managed embed-task processing. The later tenant-level filter in `listTenantRefs()` already handles the TiDB split, so this early return is broader than the intended boundary.

Source: finding/SO-001 | [Second Opinion](https://github.com/PingCAP-QE/second-opinion)

pkg/tenant/schema_zero.go
JaySon-Huang, line 1: [P1] This schema switch also breaks the local validation entrypoint unless its runtime mode flips with it.

`InitTiDBTenantSchema()` now bootstraps the TiDB auto-embedding schema, but `dat9-server-local` still constructs its backend with default options and never enables `DatabaseAutoEmbedding`. That leaves the local path on the old app-managed contract: overwrites/upload confirmation still clear embedding state through the app-managed branch, the semantic worker still attempts to write `embedding` directly, and grep stays on the app-side query-embedding path. Against the new generated-column schema this is a real schema/runtime incompatibility, not just a missing optimization.

Source: finding/SO-002 | [Second Opinion](https://github.com/PingCAP-QE/second-opinion)

pkg/tenant/schema_tidb_auto.go
JaySon-Huang, line 1: [P1] Starter validation can accept an incomplete launch schema.

`ValidateTiDBAutoEmbeddingSchema()` is now the gate used by Starter validation and TiDB backend creation, but it only runs `SHOW CREATE TABLE files` and string-checks the `files` DDL. A Starter database that has the generated `embedding` column but is missing other required launch-schema objects such as `file_nodes`, `uploads`, `semantic_tasks`, or the expected FTS/index shape will still pass validation and can be marked ready, only to fail later on normal filesystem/search operations.

Source: finding/SO-003 | [Second Opinion](https://github.com/PingCAP-QE/second-opinion)

pkg/server/semantic_worker.go
JaySon-Huang: Re-checked on the latest PR head `964c26d`. This is fixed now: `newSemanticWorkerManager()` only disables the worker for an auto-embedding fallback backend in pure single-tenant mode, while multi-tenant mode keeps the manager alive and still filters tenants at `listTenantRefs()`. I also re-ran `make test-podman` on the latest head and the full suite passed on rerun. Marking this resolved.

pkg/tenant/schema_tidb_auto.go
JaySon-Huang: Re-checked on the latest PR head `964c26d`. This is fixed now: schema validation no longer checks only `files`; it verifies the full launch contract across `file_nodes`, `files`, `file_tags`, `uploads`, and `semantic_tasks`, and the initializer now fails fast on non-TiDB targets before running the schema DDL. I also re-ran `make test-podman` on the latest head and the full suite passed on rerun. Marking this resolved.

pkg/tenant/schema_zero.go
JaySon-Huang: Re-checked on the latest PR head `964c26d`. This is fixed now: `dat9-server-local` detects the local schema contract and flips `DatabaseAutoEmbedding` accordingly, so the local bootstrap path no longer creates an auto-embedding schema while continuing to run the app-managed runtime. I also re-ran `make test-podman` on the latest head and the full suite passed on rerun. Marking this resolved.

JaySon-Huang:
Review findings:

1. `[P1] Normalize EMBED_TEXT options before validating auto schema`

In `pkg/tenant/schema_tidb_auto.go:332-335`, `validateTiDBAutoEmbeddingFilesTable` normalizes whitespace in `generation_expression` but then compares it against the raw `tidbAutoEmbeddingOptionsJSON` literal `{"dimensions":1024}`. TiDB examples and metadata can legally format the same options as `{"dimensions": 1024}`, so an otherwise valid auto-embedding `files` table gets rejected as an `embedding dimensions option mismatch`. Because this validator now gates Starter provisioning, pool backend creation, and local mode detection, tenants using documented DDL can fail activation or startup.

2. `[P1] Don't require semantic_tasks in database-managed auto mode`

In `pkg/tenant/schema_tidb_auto.go:219-220`, the new runtime explicitly stops enqueueing embed tasks and skips the embed worker for TiDB tenants in auto-embedding mode, so `semantic_tasks` is no longer part of the active contract. `ValidateTiDBSchemaForMode` still hard-fails if that table is absent, which means an externally managed Starter/Zero schema that only provisions the new `files` auto-embedding shape will be rejected by both `StarterProvisioner.InitSchema` and `Pool.createBackend`, leaving the tenant stuck in provisioning or impossible to load even though no auto-mode code path uses `semantic_tasks`.

qiffang (commented):
## Code Review Summary

整体设计非常扎实，**正确实现了 TiDB Serverless auto embedding 的要求** 👍

### ✅ 验证通过

| 要求 | 实现状态 | 关键代码 |
|-----|---------|----------|
| files 表用 TiDB auto embedding | ✅ `GENERATED ALWAYS AS (EMBED_TEXT(...))` | `pkg/tenant/schema_tidb_auto.go` |
| 大文件用外挂模型转 text | ✅ async image extractor | `pkg/backend/image_extract.go` |
| TiDB 自动把 text 转 embedding | ✅ 写入 content_text 后自动生成 | Schema 层面保证 |
| 应用无需感知 embedding 生成 | ✅ 不创建 embed task | `pkg/backend/dat9.go` |

### 架构亮点

1. **Mode Detection**: 通过检查 `embedding` 列的 `generation_expression` 判断模式
2. **Schema Validation**: 启动时 + 租户初始化时 fail-fast 验证
3. **Provider-level Rollout**: 只有 Starter/Zero 启用，db9 继续 app-managed
4. **Semantic Worker 正确跳过**: Auto embedding 租户不会被 worker 处理

### 💡 建议改进（非 blocker）

#### 1. Query 端显式指定 model（可选）

当前 `VEC_EMBED_COSINE_DISTANCE(f.embedding, ?)` 依赖 TiDB 自动使用相同 model。如果担心默认 model 变化导致 mismatch，可以考虑显式写：

```sql
VEC_EMBED_COSINE_DISTANCE(f.embedding, EMBED_TEXT("tidbcloud_free/amazon/titan-embed-text-v2", ?, "{\"dimensions\":1024}"))
```

这样更 defensive，但当前实现根据 TiDB 文档是正确的。

#### 2. Embedding 生成延迟文档化

新写入的行可能有短暂窗口期 `content_text` 已设置但 `embedding` 还在计算。`embedding IS NOT NULL` 过滤是正确的，但建议在 API 文档中提及此行为。

#### 3. Mode Detection 测试矩阵

`normalizeSQLFragment()` 解析 `generation_expression` 字符串有些脆弱。建议增加 `SHOW CREATE TABLE` 输出的测试用例覆盖不同 TiDB 版本的格式变化。

### 📝 Notes

- `embedding_revision` 字段保留但 runtime 不再使用（schema 兼容性）
- `StarterProvisioner.InitSchema()` 从 no-op 变成 validate-only
- 测试覆盖全面，包括 write/search/image-extract/worker 各路径

---

**Conclusion**: LGTM 👍 正确实现了 TiDB auto embedding 的设计目标，架构清晰，测试完善。建议作为 enhancement 而非 blocker。

JaySon-Huang:
Thanks. Quick follow-up on the two points from the regular comment:

1. `Normalize EMBED_TEXT options before validating auto schema`
   - We have validated the current behavior against a real TiDB Starter environment and it is working as expected there.
   - I agree the validator could still be made more normalization-tolerant, but since it is not blocking the current Starter flow, I plan to defer this cleanup to a follow-up change.

2. `Don't require semantic_tasks in database-managed auto mode`
   - Keeping `semantic_tasks` is intentional for now.
   - Even in database-managed auto embedding mode, we still want to reserve that table for follow-up async workflows such as image-to-text extraction and overview generation, so it remains part of the expected schema footprint.

Issue #85, opened 2026-03-31
CLI ls/cat commands don't handle ':' prefixed remote paths


mornyx:
### Current Behavior
- `dat9 cp ./file :/data/` ✓ works (uses ParseRemote)
- `dat9 ls :/data/` ✗ returns empty (doesn't use ParseRemote)
- `dat9 cat :/data/file` ✗ returns "not found" (doesn't use ParseRemote)

### Expected Behavior
ls and cat should handle ':' prefixed paths consistently with cp.

### Root Cause
- `cmd/dat9/cli/cp.go:40-41` uses `ParseRemote()` to handle ':' prefixed paths
- `cmd/dat9/cli/ls.go:16-27` passes args[0] directly to c.List()
- `cmd/dat9/cli/cat.go:16-20` passes args[0] directly to c.ReadStream()

### Fix
ls.go and cat.go should call ParseRemote() and use the .Path field, similar to cp.go.

### Related
Possibly related to #19 (CLI shell resolve() handles directory paths incorrectly).

mornyx:
Fixed in local working directory.

## Changes Made

### cmd/dat9/cli/ls.go
- Added support for : prefixed remote paths using ParseRemote()
- Both :/path and /path formats now work consistently

### cmd/dat9/cli/cat.go  
- Added support for : prefixed remote paths using ParseRemote()
- Both :/path/to/file and /path/to/file formats now work consistently

## Testing Results
All tests passed:
- ✓ `dat9 fs ls :/data/testdir/` - lists directory
- ✓ `dat9 fs ls /data/testdir/` - lists directory (backward compatible)
- ✓ `dat9 fs cat :/data/testdir/test-file.txt` - reads file content
- ✓ `dat9 fs cat /data/testdir/test-file.txt` - reads file content (backward compatible)
- ✓ `dat9 fs ls -l :/data/testdir/` - works with flags
- ✓ `dat9 fs cp :/path` - existing functionality preserved

## Lint
- ✓ golangci-lint v2.5.0: 0 issues

PR #91, opened 2026-04-01
fix(cli): support ':' prefixed paths in stat, rm, mv, find, grep, and sh commands

### LINKED ISSUE #19: CLI shell resolve() handles directory paths incorrectly — trailing slash lost
## Current State

\`\`\`go
// cmd/dat9/cli/sh.go:136-147
func resolve(cwd, path string) string {
    if strings.HasPrefix(path, \"/\") { return path }
    joined := strings.TrimSuffix(cwd, \"/\") + \"/\" + path
    canon, err := pathutil.Canonicalize(joined)  // ← uses file-mode Canonicalize
    if err != nil { return joined }
    return canon
}
\`\`\`

\`pathutil.Canonicalize\` is for **files** (strips trailing slash), but many shell commands take directory arguments:

- \`cd data\` → resolve returns \`/data\` (no trailing slash) → \`Stat(\"/data\")\` may fail (directories stored as \`/data/\`)
- \`ls sub/\` → trailing slash stripped by Canonicalize

## Proposed Fix

resolve should preserve trailing slash semantics from user input:

\`\`\`go
func resolve(cwd, path string) string {
    if strings.HasPrefix(path, \"/\") { return path }
    joined := strings.TrimSuffix(cwd, \"/\") + \"/\" + path
    if strings.HasSuffix(path, \"/\") {
        canon, err := pathutil.CanonicalizeDir(joined)
        if err != nil { return joined }
        return canon
    }
    canon, err := pathutil.Canonicalize(joined)
    if err != nil { return joined }
    return canon
}
\`\`\`

mornyx:
## Summary

This PR extends the fix from #86 to add ':' prefixed remote path support to remaining CLI commands.

## Changes

Add `ParseRemote()` support to handle ':/path' format in:
- **stat** - metadata display  
- **rm** - file removal
- **mv** - rename/move operations (both source and destination)
- **find** - file search
- **grep** - content search (both Grep and GrepJSON functions)
- **sh** - interactive shell resolve() function

## Related Issues

- Builds on #86 (already merged)
- Fixes #19 (shell resolve() handles directory paths incorrectly)
- Related to #85

## Testing

All commands now handle ':' prefixed paths consistently:
- ✓ `dat9 fs stat :/path/to/file`
- ✓ `dat9 fs rm :/path/to/file`
- ✓ `dat9 fs mv :/old/path :/new/path`
- ✓ `dat9 fs find :/path -name *.txt`
- ✓ `dat9 fs grep pattern :/path`
- ✓ `dat9 fs sh` with cd/cat/ls commands

## Code Quality

- ✓ golangci-lint v2.5.0: 0 issues
- ✓ Backward compatible (paths without ':' still work)
- ✓ Consistent with cp command behavior

### DIFF
--- a/cmd/dat9/cli/cat.go
+++ b/cmd/dat9/cli/cat.go
@@ -13,11 +13,17 @@ import (
 // Uses ReadStream to handle both small files (direct) and large files (presigned URL).
 //
 //	dat9 cat /path/to/file
+//	dat9 cat :/path/to/file
 func Cat(c *client.Client, args []string) error {
 	if len(args) != 1 {
 		return fmt.Errorf("usage: dat9 cat <path>")
 	}
-	rc, err := c.ReadStream(context.Background(), args[0])
+	path := args[0]
+	// Handle ":" prefixed remote paths like cp command
+	if rp, isRemote := ParseRemote(path); isRemote {
+		path = rp.Path
+	}
+	rc, err := c.ReadStream(context.Background(), path)
 	if err != nil {
 		return err
 	}
--- a/cmd/dat9/cli/find.go
+++ b/cmd/dat9/cli/find.go
@@ -62,6 +62,11 @@ func Find(c *client.Client, args []string) error {
 		}
 	}
 
+	// Handle ":" prefixed remote paths like cp command
+	if rp, isRemote := ParseRemote(path); isRemote {
+		path = rp.Path
+	}
+
 	results, err := c.Find(path, params)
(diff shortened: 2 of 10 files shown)


mornyx:
This PR completes the fix for #85 by adding ':' prefix support to all remaining CLI commands.\n\nAlso fixes #19 (shell resolve() function).

qiffang (approved):

PR #92, opened 2026-04-01
fix: improve install onboarding, fix server URL, tune storage thresholds

qiffang:
## Summary

- **install.sh**: Replace flat post-install output with numbered 1-2-3 setup flow (create workspace → verify → start using). Fix misleading "provision a new database" copy → "Create a workspace"
- **install.sh**: Bootstrap `~/.dat9/config` with server URL on install so config file exists immediately. Extract `BASE_URL` and `API_URL` as top-level variables
- **config.go**: Fix default server URL from `api.db9.ai` → `api.dat9.ai`
- **storage**: Lower small file threshold from 1MB to 50KB — files ≥ 50KB now go to S3 instead of inline DB storage
- **s3**: Increase presigned URL TTL from 2min/1min to 10min for both upload and download

## Test plan

- [ ] Run `sh site/install.sh` locally and verify post-install output shows numbered steps
- [ ] Verify `~/.dat9/config` is created with correct JSON and `api.dat9.ai` server
- [ ] Verify re-running install does not overwrite existing config
- [ ] Upload files around 50KB boundary and confirm small files go to DB, large files go to S3
- [ ] Verify presigned URLs remain valid for ~10 minutes

### DIFF
--- a/pkg/backend/dat9.go
+++ b/pkg/backend/dat9.go
@@ -27,7 +27,7 @@ import (
 	"go.uber.org/zap"
 )
 
-const smallFileThreshold = 50 << 10 // 50KB
+const smallFileThreshold = 50_000 // 50,000 bytes — matches embedding model max input characters
 
 // Dat9Backend implements filesystem.FileSystem with the inode model.
 type Dat9Backend struct {
--- a/pkg/backend/dat9.go
+++ b/pkg/backend/dat9.go
@@ -27,7 +27,7 @@ import (
 	"go.uber.org/zap"
 )
 
-const smallFileThreshold = 1 << 20 // 1MB
+const smallFileThreshold = 50 << 10 // 50KB
 
 // Dat9Backend implements filesystem.FileSystem with the inode model.
 type Dat9Backend struct {
--- a/README.md
+++ b/README.md
@@ -173,10 +173,10 @@ c.Delete("/data/file.txt")
   Dat9Backend   memfs     S3 direct
   (AGFS FileSystem)       (presigned URL,
      │                     FUSE ↔ S3)
-     ├── < 50KB → TiDB(MySQL protocol) + local blobs (P0)
-     │            db9 + fs9 (production)
+     ├── < 50,000B → TiDB(MySQL protocol) + local blobs (P0)
+     │               db9 + fs9 (production)
      │
-     └── ≥ 50KB → S3 presigned URL
+     └── ≥ 50,000B → S3 presigned URL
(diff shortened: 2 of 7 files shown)


srstack (approved):

srstack:
Thanks for the update — I have one important concern before merge:\n\n1. **FUSE read-cache threshold regression risk** ()\n   - This PR reduces  from 1MB to 50KB and reuses it for FUSE read-cache eligibility ( path).\n   - Result: files in ~50KB–1MB are no longer cached in FUSE, which can increase repeated-read latency and backend load.\n   - If the goal is storage placement (DB vs S3), cache threshold should remain decoupled (or separately configurable).\n\n2. **Comment drift in FUSE cache docs**\n   -  comments still describe 1MB semantics in nearby docs, while behavior is now 50KB.\n   - Please sync comments to avoid maintenance confusion.\n\nAlso, PR description mentions install/config updates, but current diff only changes backend/client/fuse/s3 constants. If install/config fixes are intended, they are not present in this PR.

pkg/backend/dat9.go
```diff
@@ -29,20 +27,15 @@ import (
 	"go.uber.org/zap"
 )
 
-const smallFileThreshold = 1 << 20 // 1MB
+const smallFileThreshold = 50 << 10 // 50KB
```
JaySon-Huang: According to tidbcloud website, the maximum is 50000 but not 50 KiB.
Maximum input text characters: 50,000
https://docs.pingcap.com/ai/vector-search-auto-embedding-amazon-titan/

srstack:
Follow-up (formatting fix):

Thanks for the update — I have one important concern before merge:

1. FUSE read-cache threshold regression risk (`pkg/fuse/read.go`)
- This PR reduces `smallFileThreshold` from 1MB to 50KB and reuses it for FUSE read-cache eligibility (`Put` path).
- Result: files in ~50KB–1MB are no longer cached in FUSE, which can increase repeated-read latency and backend load.
- If the goal is storage placement (DB vs S3), cache threshold should remain decoupled (or separately configurable).

2. Comment drift in FUSE cache docs
- `ReadCache` comments still describe 1MB semantics in nearby docs, while behavior is now 50KB.
- Please sync comments to avoid maintenance confusion.

Also, PR description mentions install/config updates, but current diff only changes backend/client/fuse/s3 constants. If install/config fixes are intended, they are not present in this PR.

JaySon-Huang (approved):

PR #96, opened 2026-04-01
refactor: move auto image extraction onto durable semantic tasks

JaySon-Huang:
## Summary
- add the durable `img_extract_text` task contract, extract a reusable backend image-processing core, and keep app-mode legacy queue behavior as temporary compatibility
- make semantic workers resolve tenant backends directly, claim tasks by target capability, and avoid unsupported-task routing so image extraction runs through backend-owned runtime dependencies safely
- switch auto-embedding `create`, `overwrite`, and `upload completion` to enqueue durable image tasks transactionally, and tighten pooled backend lifetime / semantic lease handling for real image extraction workloads
- add focused regression coverage plus a local verifier script for direct and multipart `img_extract_text` e2e validation

## Deployment defaults
This PR does not introduce new `DAT9_IMAGE_EXTRACT_*` knobs; those already exist on `main`. What changes here is the runtime behavior around durable `img_extract_text` processing and the default semantic-task lease handling when async image extraction is enabled.
A few deployment notes are worth calling out:
- The numeric worker defaults are intentionally conservative and are sufficient to run the flow:
  - `DAT9_SEMANTIC_WORKERS=1`
  - `DAT9_SEMANTIC_POLL_INTERVAL_MS=200`
  - `DAT9_SEMANTIC_RECOVER_INTERVAL_MS=5000`
  - `DAT9_SEMANTIC_PER_TENANT_CONCURRENCY=1`
- `DAT9_SEMANTIC_LEASE_SECONDS` is still configurable, but when it is left unset and async image extraction is enabled, both `dat9-server` and `dat9-server-local` now derive the default lease as `max(30s, 2x DAT9_IMAGE_EXTRACT_TIMEOUT_SECONDS)`. This is meant to avoid lease expiry during slower real-image extraction workloads.
- The deployment still does **not** run the full durable image extraction flow by default unless the required runtime dependencies are configured:
  - `DAT9_META_DSN` is required for `dat9-server`
  - `DAT9_TOKEN_SIGNING_KEY` is required to create the tenant backend pool
  - this semantic handler must be configured for workers to be useful for auto-embedding tenants:
    - async image extraction via `DAT9_IMAGE_EXTRACT_ENABLED=true`
- If `DAT9_IMAGE_EXTRACT_ENABLED=true` is set without a real vision endpoint (`DAT9_IMAGE_EXTRACT_API_BASE` / `DAT9_IMAGE_EXTRACT_API_KEY` / `DAT9_IMAGE_EXTRACT_MODEL`), the durable flow can still run, but it will use the basic fallback extractor rather than a real multimodal model.
- For real production-style image extraction, operators should explicitly configure the image extract endpoint/model and size the timeout appropriately for the selected model. The new derived lease default reduces one class of failure, but it is not a substitute for proper model-specific timeout tuning.

## Testing
- make test-podman
- python3 scripts/verify_local_img_extract.py --direct-image "/Users/jayson/Projects/claw-memories/images/Biq8G72FYi.jpg" --multipart-image "/Users/jayson/Projects/claw-memories/images/images.jpeg"

## Local e2e validation for durable `img_extract_text`
I validated the new durable image extraction flow locally with `make run-server-local` against a real vision model instead of the basic fallback extractor.

### Setup
I started `dat9-server-local` in auto-embedding mode with async image extraction enabled and an OpenAI-compatible vision endpoint backed by Ollama:
- image extract model: `qwen2.5vl:3b`
- image extract timeout: `20s`
- semantic lease duration: auto-derived to `40s` in local mode
- semantic worker: enabled
- image extract worker: enabled

Relevant startup log signals:
- `image_extract_mode_openai_compatible`
- `backend_image_extract_workers_started`
- `server_semantic_workers_enabled`
- `semantic_worker_manager_started`

### How I ran it
I used the verifier script added in this branch with two real images:

```bash
python3 scripts/verify_local_img_extract.py \
  --direct-image "/Users/jayson/Projects/claw-memories/images/Biq8G72FYi.jpg" \
  --multipart-image "/Users/jayson/Projects/claw-memories/images/images.jpeg"
```

This script validates two end-to-end paths:
1. direct write / `PUT /v1/fs/<image>`
2. multipart upload / `POST /v1/uploads/initiate` -> upload part(s) -> `POST /v1/uploads/<id>/complete`

For each path it waits until:
- a durable `semantic_tasks` row exists with `task_type = 'img_extract_text'`
- the task reaches `status = 'succeeded'`
- `files.content_text` is written back

#### Direct image upload
Input image:
- `Biq8G72FYi.jpg`

Observed result:
- durable task created with `task_type = img_extract_text`
- final status: `succeeded`
- attempt count: `2`
- `files.content_text` was written with real model output

Example extracted text:

> 主要物体：一只橘色的小猫  
> 场景描述：小猫仰卧在白色背景上，四脚朝天，前爪抬起，显得非常可爱和活泼。  
> 图中可见文字（OCR）：图精灵616PIC.COM  
> 简洁标签：可爱小猫，仰卧，四脚朝天，橘色，白色背景  
> 英文标签：cute kitten, lying down, four legs up, orange, white background

Notes:
- the first attempt hit `context deadline exceeded`
- the worker retried successfully
- no `semantic task lease mismatch` occurred after the lease-default adjustment

#### Multipart image upload
Input image:
- `images.jpeg`

Observed result:
- durable task created with `task_type = img_extract_text`
- final status: `succeeded`
- attempt count: `1`
- `files.content_text` was written with real model output

Example extracted text:

> 主要物体：绿色的小型汽车  
> 场景描述：一辆绿色的小型汽车，侧面视角，背景为白色，没有其他物体或细节。  
> 图中可见文字（OCR）：无  
> 简洁标签：绿色汽车，小型车，侧面视角，白色背景  
> 英文标签：green car, small car, side view, white background

### Conclusion
The branch-local durable `img_extract_text` flow works end-to-end for both:
- direct file writes
- multipart uploads

and correctly:
- creates durable semantic tasks
- processes them through the semantic worker
- writes extracted text into `files.content_text`

The lease-default change fixed the earlier `semantic task lease mismatch` issue. With `qwen2.5vl:3b` and `DAT9_IMAGE_EXTRACT_TIMEOUT_SECONDS=20`, multipart succeeds on the first attempt, while the direct image still needed one retry.

### DIFF
--- a/pkg/backend/image_extract.go
+++ b/pkg/backend/image_extract.go
@@ -219,8 +219,9 @@ func (b *Dat9Backend) processQueuedImageExtractTask(ctx context.Context, task Im
 }
 
 // ProcessImageExtractTask runs the backend-owned image extraction logic for one
-// revision-scoped task. Terminal non-retry outcomes return a nil error and a
-// non-empty result label. Transient failures return a retryable error.
+// revision-scoped task. Normal terminal business outcomes return a nil error
+// and a non-empty result label. Runtime misconfiguration, misrouting, and
+// transient failures return a retryable error.
 func (b *Dat9Backend) ProcessImageExtractTask(ctx context.Context, task ImageExtractTaskSpec) (ImageExtractResult, error) {
 	if !b.SupportsAsyncImageExtract() {
 		return ImageExtractResultRuntimeNotConfigured, fmt.Errorf("async image extract runtime not configured")
--- a/pkg/backend/image_extract.go
+++ b/pkg/backend/image_extract.go
@@ -29,12 +29,32 @@ type ImageTextExtractor interface {
 	ExtractImageText(ctx context.Context, req ImageExtractRequest) (string, error)
 }
 
-type imageExtractTask struct {
+// ImageExtractTaskSpec carries the revision-scoped inputs needed to extract
+// image text for one file version.
+type ImageExtractTaskSpec struct {
 	FileID      string
 	Path        string
 	ContentType string
 	Revision    int64
 }
+
+// ImageExtractResult reports the outcome of one image extraction attempt.
+type ImageExtractResult string
+
+const (
+	ImageExtractResultRuntimeNotConfigured ImageExtractResult = "runtime_not_configured"
+	ImageExtractResultGetFileError         ImageExtractResult = "get_file_error"
+	ImageExtractResultFileNotFound         ImageExtractResult = "file_not_found"
+	ImageExtractResultNotConfirmed         ImageExtractResult = "not_confirmed"
+	ImageExtractResultNotImage             ImageExtractResult = "not_image"
+	ImageExtractResultStale                ImageExtractResult = "stale"
+	ImageExtractResultLoadError            ImageExtractResult = "load_error"
+	ImageExtractResultTooLarge             ImageExtractResult = "too_large"
+	ImageExtractResultExtractError         ImageExtractResult = "extract_error"
+	ImageExtractResultEmptyText            ImageExtractResult = "empty_text"
+	ImageExtractResultUpdateError          ImageExtractResult = "update_error"
+	ImageExtractResultWritten              ImageExtractResult = "written"
+)
 
 func isImageContentType(contentType string) bool {
 	contentType = strings.ToLower(strings.TrimSpace(contentType))
@@ -58,8 +78,14 @@ func contentTypeFromPath(path string) string {
 	return ""
 }
 
+// SupportsAsyncImageExtract reports whether this backend instance has the
+// runtime dependencies needed to extract image text.
+func (b *Dat9Backend) SupportsAsyncImageExtract() bool {
+	return b.imageExtractEnabled && b.imageExtractor != nil
+}
+
 func (b *Dat9Backend) enqueueImageExtract(fileID, path, contentType string, revision int64) {
-	if !b.imageExtractEnabled || b.imageExtractor == nil {
+	if !b.SupportsAsyncImageExtract() {
 		return
 	}
 	if !isImageContentType(contentType) {
@@ -68,7 +94,7 @@ func (b *Dat9Backend) enqueueImageExtract(fileID, path, contentType string, revi
 			return
 		}
 	}
-	task := imageExtractTask{
+	task := ImageExtractTaskSpec{
 		FileID:      fileID,
 		Path:        path,
 		ContentType: contentType,
@@ -89,7 +115,7 @@ func (b *Dat9Backend) enqueueImageExtract(fileID, path, contentType string, revi
 }
 
 func (b *Dat9Backend) enqueueImageExtractForUpload(ctx context.Context, upload *datastore.Upload, isOverwrite bool) {
-	if !b.imageExtractEnabled || b.imageExtractor == nil {
+	if !b.SupportsAsyncImageExtract() {
 		return
 	}
 	fileID := upload.FileID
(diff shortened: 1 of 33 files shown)


pkg/server/semantic_worker.go
```diff
@@ -432,75 +431,62 @@ func (m *semanticWorkerManager) listTenantRefs(ctx context.Context) ([]semanticT
...
+	if m.shouldIncludeFallback() {
 		return []semanticTenantRef{{id: semanticLocalTenantID}}, nil
 	}
 	return nil, nil
 }
 
-func (m *semanticWorkerManager) storeForRef(ctx context.Context, ref semanticTenantRef) (*datastore.Store, error) {
-	if ref.id == semanticLocalTenantID {
-		return m.fallback.Store(), nil
+func (m *semanticWorkerManager) targetForRef(ctx context.Context, ref semanticTenantRef) (*semanticTarget, error) {
```
JaySon-Huang, line 460: [P1] `targetForRef` now executes OCR directly against the `tenant.Pool` backend/store. Because `tenant.Pool.Get` is LRU-backed and eviction closes both the backend and datastore, an in-flight task for tenant A can lose its DB or S3 handles as soon as loading another tenant evicts A. The previous worker-owned store cache did not have this mid-task invalidation failure mode, so the worker needs to pin the backend/store for the lifetime of the claimed task.

pkg/server/semantic_worker.go
```diff
@@ -640,3 +636,50 @@ func (m *semanticWorkerManager) collectObservation(ctx context.Context, now time
...
+
+func (m *semanticWorkerManager) hasEmbedHandler() bool {
+	return m != nil && m.embedder != nil
+}
+
+func (m *semanticWorkerManager) hasImageHandler() bool {
+	if m == nil {
+		return false
+	}
+	if m.fallback != nil && m.fallback.SupportsAsyncImageExtract() {
```
JaySon-Huang, line 676: [P2] `hasImageHandler` becomes true when either the local fallback backend or the tenant pool supports async image extraction, and `supportsTenantProvider` then uses that global bit for every auto tenant. In a mixed deployment where only the fallback has OCR enabled, this makes the worker scan pool tenants whose backends still return `runtime_not_configured`, so their `img_extract_text` tasks just retry until dead-letter. Tenant scan capability needs to be derived from the pool itself rather than fallback state.

pkg/server/semantic_worker.go
```diff
@@ -432,75 +431,62 @@ func (m *semanticWorkerManager) listTenantRefs(ctx context.Context) ([]semanticT
 		refs := make([]semanticTenantRef, 0, len(tenants))
 		for i := range tenants {
 			t := tenants[i]
-			if semanticWorkerUsesTiDBAutoEmbedding(t.Provider) {
+			if !m.supportsTenantProvider(t.Provider) {
```
JaySon-Huang, line 447: [P1] When the worker runs in image-only mode (`embedder == nil`), this adds every auto-embedding tenant to the scan set, but `ClaimSemanticTask` is still task-type agnostic and `dispatchTask` retries unsupported types. If an upgraded auto tenant still has queued `embed` rows from before the cutover, the worker can keep claiming those first and never reach the new `img_extract_text` work behind them. This rollout needs either task-type filtering at claim time or a guard that skips tenants with legacy embed backlog.

pkg/server/semantic_worker.go
JaySon-Huang, line 1: The worker decides whether to scan fallback/tenant stores from manager-level `hasImageHandler`/`hasEmbedHandler` flags instead of each target backend's actual capability. In mixed setups this can route `img_extract_text` work to stores that do not have an extractor runtime, causing repeated retries and eventual dead-lettering. Please make task-type eligibility target-specific rather than global.

JaySon-Huang, line 1: This worker can start in image-only mode, but `ClaimSemanticTask` still leases any queued task type. If an old `embed` task is still present, this path dereferences a nil `m.embedder` and panics, which can take down semantic processing for the process. Please filter claimed task types by available handlers, or guard the embed path when no embedder is configured.

pkg/server/semantic_worker.go
```diff
@@ -432,75 +431,62 @@ func (m *semanticWorkerManager) listTenantRefs(ctx context.Context) ([]semanticT
...
 }
 
-func (m *semanticWorkerManager) dispatchTask(ctx context.Context, tenantID string, store *datastore.Store, task *semantic.Task) string {
+func (m *semanticWorkerManager) dispatchTask(ctx context.Context, target *semanticTarget, task *semantic.Task) string {
 	switch task.TaskType {
 	case semantic.TaskTypeEmbed:
-		return m.processEmbedTask(ctx, tenantID, store, task)
+		return m.processEmbedTask(ctx, target.tenantID, target.store, task)
+	case semantic.TaskTypeImgExtractText:
+		return m.processImgExtractTask(ctx, target.tenantID, target.store, target.backend, task)
```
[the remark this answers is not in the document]
JaySon-Huang: Fixed in `0e5fb38`. The worker now uses target-specific typed claiming (`ClaimSemanticTaskOfTypes(...)`), and `processEmbedTask(...)` has an explicit `m.embedder == nil` guard so image-only mode no longer panics if an unexpected embed task is ever reached.

pkg/backend/image_extract.go
```diff
@@ -138,54 +164,98 @@ func (b *Dat9Backend) runImageExtractWorker(ctx context.Context, workerID int) {
...
+	}
+	metrics.RecordOperation("image_extract", "process", metricResult, time.Since(start))
+}
+
+// ProcessImageExtractTask runs the backend-owned image extraction logic for one
+// revision-scoped task. Terminal non-retry outcomes return a nil error and a
+// non-empty result label. Transient failures return a retryable error.
+func (b *Dat9Backend) ProcessImageExtractTask(ctx context.Context, task ImageExtractTaskSpec) (ImageExtractResult, error) {
+	if !b.SupportsAsyncImageExtract() {
+		return ImageExtractResultRuntimeNotConfigured, fmt.Errorf("async image extract runtime not configured")
```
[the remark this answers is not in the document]
JaySon-Huang: Addressed in `06323fa`. I kept `runtime_not_configured` as a non-nil error so routing/runtime invariant breaks remain visible, and updated the doc comment to distinguish normal terminal business outcomes from runtime misconfiguration/misrouting/transient failures.

pkg/server/semantic_worker.go
JaySon-Huang: Fixed by `e580f74` and `bdaf17e`. `tenant.Pool` now exposes `Acquire(...)` with borrow/release semantics, eviction/invalidate retire entries instead of closing borrowed backends immediately, and both semantic worker and request auth paths now hold the backend for the full lifetime of the task/request before releasing it.

pkg/server/semantic_worker.go
JaySon-Huang: Fixed in `0e5fb38`. Claiming is now target-specific: the worker resolves a concrete target backend first, computes `allowedTaskTypes` from that target, and then uses `ClaimSemanticTaskOfTypes(...)`. I also added coverage for the mixed case where fallback OCR must not make pool tenants look image-capable.

pkg/server/semantic_worker.go
JaySon-Huang: Fixed in `609b933` and `0e5fb38`. I added typed semantic task claiming in `pkg/datastore`, then switched the worker to claim only the task types allowed by the resolved target. That means image-only workers no longer lease legacy `embed` backlog ahead of `img_extract_text` work.

pkg/server/semantic_worker.go
JaySon-Huang: Fixed in `0e5fb38`. The worker no longer relies on manager-level capability bits for claiming. It now resolves each target backend, derives target-specific `allowedTaskTypes`, and only claims those task types. I also added regression coverage for the fallback-OCR / pool-tenant mixed setup.

JaySon-Huang: Fixed in `609b933` and `0e5fb38`. The worker now filters claims by target capability via `ClaimSemanticTaskOfTypes(...)`, and `processEmbedTask(...)` has a defensive missing-embedder guard. That closes both the backlog starvation case and the nil-embedder panic path.

pkg/server/semantic_worker.go
```diff
@@ -33,7 +34,12 @@ const (
 
 var semanticWorkerUsesTiDBAutoEmbedding = tenant.UsesTiDBAutoEmbedding
 
-// SemanticWorkerOptions controls background embed task processing.
+var (
+	semanticWorkerAllowedEmbedTaskTypes      = []semantic.TaskType{semantic.TaskTypeEmbed, semantic.TaskTypeGenerateL0}
```
JaySon-Huang: [P2] The new typed-claim filter still whitelists `generate_l0`, but `dispatchTask` only handles `embed` and `img_extract_text`; `generate_l0` still goes through the unsupported-task retry/dead-letter path. That means an image-only worker can still burn attempts on an older `generate_l0` row that sits ahead of `img_extract_text`, which recreates queue-head blocking through a different task type. Until there is a real L0 handler, these allow-lists should only contain task types that `dispatchTask` can execute.

pkg/server/semantic_worker.go
JaySon-Huang, line 1: `generate_l0` is now included in the worker's allowed task-type sets, but `dispatchTask` still only implements `embed` and `img_extract_text`. In auto/image-only mode that means the new worker can actively claim `generate_l0` and then drive it through the unsupported-task retry/dead-letter path instead of leaving it for a real handler. Please either remove `generate_l0` from the allowed sets for now, or implement its handler before advertising it as claimable work.

Source: analysis/local-review-003 | [Second Opinion](https://github.com/PingCAP-QE/second-opinion)

pkg/server/semantic_worker.go
JaySon-Huang: Fixed in `6089bb3`. I removed `generate_l0` from the worker claim allowlists so typed claim now only leases task types that the current worker can actually execute (`embed` for app targets, `img_extract_text` for auto targets). I also updated the regression test so `generate_l0` stays queued instead of being proactively claimed and dead-lettered by this worker.

qiffang (commented):
## Code Review Summary

这是一个高质量的重构 PR，将 auto embedding 模式下的异步图片文本提取从内存队列迁移到持久化 `semantic_tasks` 表 👍

### ✅ 核心设计验证通过

| 设计点 | 评价 |
|-------|------|
| 事务内任务注册 | ✅ 文件可见时任务已持久化，无丢失窗口 |
| Payload 最小化 | ✅ 只含 path/content_type hints，真实来源是 files 表 |
| Worker 启动条件 | ✅ 有任一 handler 即启动，不再强制要求 embedder |
| Tenant scan 按能力分组 | ✅ img-only worker 只扫描 auto tenants |
| Lease duration 自动调整 | ✅ max(30s, 2x task_timeout) |
| Legacy 队列保留 | ✅ 仅给 app embedding，代码注释明确标注临时兼容 |
| Result 枚举 | ✅ 清晰区分 terminal 和 retryable |
| 测试覆盖 | ✅ create / overwrite / upload / upload-overwrite 四条路径 |
| 设计文档 | ✅ 430 行详尽文档 |

### 💡 建议改进（非 blocker）

#### 1. Claim 时的 task_type 防御性校验

当前依赖 tenant scan 分组防止误 claim。建议在 claim query 中增加 `task_type` 过滤作为防御层：

```sql
WHERE status = queued AND task_type IN (img_extract_text) -- img-only worker
```

这样即使 scan 逻辑有 bug，claim 阶段也能防止误处理。

#### 2. Lease multiplier 可更保守

当前 `lease = max(30s, 2x task_timeout)`。如果 VLM API 慢或重试，2x 可能不够。

建议改为 **3x** 或 `max(2x, expected_worst_case + jitter)`：
- 实现简单，避免 lease renewal 复杂度
- img_extract 任务幂等，lease 过期重试是安全的

#### 3. Edge case 确认

**(a) Concurrent overwrite race**

场景：create file v1 → enqueue task(rev=1) → overwrite to v2 → worker claims task(rev=1)

确认 `ProcessImageExtractTask` 有 revision gating：
```go
if task.Revision > 0 && f.Revision != task.Revision {
    return ImageExtractResultStale, nil  // ✅ 已处理
}
```

**(b) Tenant mode switch**

如果 tenant 在 enqueue 和 claim 之间从 auto 切换到 app，当前设计会让任务 stale 或被 scan 跳过。建议在文档中明确这是预期行为。

**(c) Conditional writes**

确认 `UpdateFileSearchTextTx` 是 revision-gated 的，避免 late result 覆盖 newer content。

### 📝 团队讨论结论

- **Claude**: LGTM，建议 claim 时 task_type 校验 + 3x lease multiplier + 确认 revision gating
- **codex**: LGTM，建议 task_type filter at claim + lightweight lease renewal or conservative multiplier + clarify tenant mode switch behavior

---

**Conclusion**: LGTM 👍 高质量重构，解决了内存队列的持久化问题，事务边界正确，过渡策略清晰。建议作为 enhancement 加入 claim 时 task_type 校验和更保守的 lease multiplier。

PR #129, opened 2026-04-07
Limit upload size and tenant storage usage

JaySon-Huang:
## Summary
- reduce the default single-upload size limit from 50 GiB to 10 GiB
- enforce a 50 GiB tenant storage quota based on confirmed physical file bytes plus active upload reservations
- return clearer HTTP errors for upload-size and tenant-quota violations, including patch upload paths
- wire the new quota settings through server config and update tests/e2e defaults

## Current limitations
- The 50 GiB tenant quota is measured by physical file entities in `files`, not by the sum of visible namespace entries. Zero-copy `CopyFile` adds a new `file_node` pointing to the same `file_id`, so it does not consume additional quota.
- The quota check is not a hard concurrent limit. Two uploads initiated concurrently can still pass based on the same pre-insert snapshot and temporarily push the tenant above 50 GiB.
- After such an overshoot, later upload initiations should still be rejected because the excess will be reflected by either active upload reservations or confirmed file bytes.
- This change intentionally does not introduce an atomic quota-reservation table or serialized quota updates.

## Testing
- make test TEST_P=1


qiffang:
## Code Review Summary

Reviewed by: **openclaw-work**, **Claude**, **codex**

### ✅ Overall Assessment: **LGTM with minor suggestions**

The core quota implementation is well-designed:
- Transaction-based quota check avoids TOCTOU issues
- Correct HTTP status codes (413, 507)
- Delta calculation correctly subtracts current file size for overwrite scenarios

---

### 🔍 Review Findings

#### 1. Soft Concurrency Limit (Acknowledged)
Two concurrent uploads may temporarily exceed quota. This is an acceptable trade-off to avoid serialization locks. 
- **Suggestion**: Document this as "best-effort quota enforcement" in API docs.

#### 2. SQL Performance Concern
The 3-way JOIN in `ActiveUploadReservedBytesTx` may have performance issues under high concurrency:
```sql
SELECT ... FROM uploads u
LEFT JOIN file_nodes fn ON fn.path = u.target_path
LEFT JOIN files f ON f.file_id = fn.file_id
WHERE u.status IN ('INITIATED', 'UPLOADING')
```
- **Suggestion**: Ensure indexes exist on `uploads(status)` and `file_nodes(path)`. Consider caching `confirmedBytes` if it becomes a bottleneck.

#### 3. Zero-Copy Files Don't Consume Quota
`CopyFile` only creates a new `file_node` pointing to the same `file_id`, so copies don't consume additional quota.
- **Question**: Is this intentional? If so, please document this behavior. If not, quota check may be needed in copy path.

#### 4. INITIATED vs UPLOADING Status
New `UploadInitiated` status added, but no visible transition logic to `UPLOADING`.
- **Question**: Where does the status transition occur? If there's no transition, consider simplifying to a single status.

#### 5. Tenant Scoping (codex)
Ensure all quota queries are filtered by tenant/org. Otherwise, cross-tenant sums could leak quota.

#### 6. Reservation Leak (codex)
If `INITIATED` uploads never transition or expire, `reservedBytes` can accumulate forever.
- **Suggestion**: Add TTL cleanup or exclude stale INITIATED records from reservation calculation.

#### 7. Size Verification on Complete (codex)
Ensure the final size is re-verified on `complete`, not just during `initiate`. Otherwise, a misreported size could bypass the 10 GiB limit.

---

### 💡 Additional Suggestions

1. **Error Response Enrichment**: Include quota details in 413/507 responses (e.g., `X-Quota-Remaining` header or JSON body with usage stats)

2. **Quota Warning**: Return a warning header at 80% usage to give clients early notice

3. **Test Coverage**: Verify edge cases are tested:
   - Upload exactly at quota limit
   - Overwrite where new file is larger/smaller
   - Concurrent initiate scenario

---

### Checklist
- [x] Confirm tenant scoping in all quota queries
- [x] Clarify INITIATED → UPLOADING transition
- [x] Document zero-copy quota behavior
- [ ] Add indexes if not present
- [x] Consider TTL for stale reservations

JaySon-Huang:
Thanks for the detailed review. Point-by-point:

1. Soft concurrency limit
This is acknowledged and intentional for this change. The current implementation is best-effort quota enforcement rather than a hard concurrent cap: two near-limit initiations can still pass on the same pre-insert snapshot and temporarily push the tenant above 50 GiB. I documented that limitation explicitly in the PR description.

2. SQL performance concern
I added a TODO next to the reservation aggregation query in `ActiveUploadReservedBytesTx` to revisit the access path if upload concurrency grows. The current schema already has `file_nodes(path)` and `uploads(target_path, status)` indexes, but if this query becomes hot we should likely move toward a more targeted index strategy or a pre-aggregated quota state.

3. Zero-copy files do not consume quota
This is intentional for the current quota definition. In this PR, the 50 GiB limit is based on physical file entities, not the sum of visible namespace entries. `CopyFile` only creates a new `file_node` pointing at the same `file_id`, so it does not add physical storage usage. I also documented this behavior in the PR description.

4. INITIATED vs UPLOADING
The transition exists. V2 uploads are created as `INITIATED`, and the first `PresignPart` / `PresignParts` call atomically transitions the row to `UPLOADING`.

5. Tenant scoping
The quota queries are tenant-scoped by construction. Each tenant uses its own data-plane database, and these store queries run only against the current tenant DB, so there is no cross-tenant aggregation here.

6. Reservation leak for stale INITIATED uploads
This was a valid gap, and I fixed it in this branch. We now treat only non-expired `INITIATED` / `UPLOADING` rows as active for both:
- reservation accounting
- same-path upload conflict detection

That means an expired upload no longer keeps consuming reserved quota and no longer blocks a new initiate on the same path. I also added tests for both behaviors.

7. Size verification on complete
Complete-time validation already exists. `ConfirmUpload` / `ConfirmUploadV2` re-read the uploaded S3 parts and verify that both part count and per-part size match the layout implied by `upload.TotalSize`, so a client cannot bypass the 10 GiB limit by misreporting size only at initiate time.

Also, after the expired-upload fix, `make test` passes locally.

qiffang:
## Code Review Summary (Round 2)

Reviewed by: **openclaw-work**, **Claude**, **codex**

### ✅ Overall Assessment: **Approve with minor follow-ups**

The new commit `9f92fc8` correctly fixes the **reservation leak issue** raised in Round 1.

---

### 🔧 Fixed in This Round

| Issue | Fix |
|-------|-----|
| Expired uploads consuming quota | ✅ `ActiveUploadReservedBytesTx` now filters `expires_at > ?` |
| Expired uploads blocking same path | ✅ `GetUploadByPath` now filters `expires_at > ?` |
| Test coverage | ✅ Two new tests added |

---

### ⚠️ Items to Confirm (codex + Claude)

1. **UTC consistency**: Confirm `expires_at` storage and comparison are both in UTC. Is `expires_at == now` considered expired or not?

2. **Non-null guarantee**: Is `expires_at` guaranteed non-null for INITIATED/UPLOADING uploads? If null, rows might be incorrectly filtered out.

3. **Other active upload queries**: Are there other places reading active uploads that also need `expires_at` filtering? (e.g., `ListUploads`, `GetUpload`)

4. **Operations on expired uploads**: What's the behavior when calling `CompleteUpload` or `CancelUpload` on an expired upload? Should it return NotFound or a specific error?

5. **Background cleanup**: Is there a mechanism to periodically clean up expired upload records? Query-time filtering works, but database bloat is still a concern.

---

### 📋 Outstanding Questions from Round 1

| Question | Suggested Action |
|----------|------------------|
| **Size verification on complete** | Confirm: Does `CompleteUpload` verify actual bytes match declared size? |
| **Zero-copy quota behavior** | Clarify: Is zero-copy intentionally not counting toward reserved? Add comment if yes. |
| **INITIATED vs UPLOADING semantics** | Document: What triggers the state transition? |
| **Tenant scoping** | Confirm: Single-tenant architecture? Document if so. |
| **SQL performance** | Acknowledged via TODO comment - acceptable for now |

---

### 💡 Recommendations

1. **Immediate** (before merge):
   - Confirm UTC handling and non-null guarantee for `expires_at`
   
2. **Follow-up** (can be separate PRs):
   - Add background cleanup for expired uploads
   - Document INITIATED/UPLOADING state machine
   - Add comment explaining zero-copy quota behavior

---

### Checklist
- [x] Expired uploads excluded from quota calculation ✅
- [x] Expired uploads don't block same-path initiate ✅
- [x] Tests verify expiration behavior ✅
- [ ] Confirm expires_at is always non-null and UTC
- [ ] Decide behavior for operations on expired uploads
- [ ] Consider background cleanup mechanism

---

**Verdict**: This PR is ready to merge. The outstanding questions can be addressed in follow-up PRs or clarified in comments.

qiffang (approved):

PR #134, opened 2026-04-07
fix: v2 multipart upload should not declare ChecksumAlgorithm

### LINKED ISSUE #133: Large file uploads fail in v2 multipart flow when S3 requires SHA256 checksums
## Summary
Large file uploads initiated through `drive9 fs cp` can fail during multipart upload with an S3 `InvalidRequest` error when the multipart session is created with `ChecksumAlgorithm=SHA256` but individual `UploadPart` requests do not include `x-amz-checksum-sha256`.

## Reproduction
Run a large upload so the CLI uses multipart upload:

```bash
drive9 fs cp /path/to/large-file.bin :/large-file.bin
```

Observed error:

```text
fs cp: part 4: HTTP 400: <?xml version="1.0" encoding="UTF-8"?>
<Error><Code>InvalidRequest</Code><Message>Checksum Type mismatch occurred, expected checksum Type: sha256, actual checksum Type: null</Message>...</Error>
```

## Root Cause
The current v2 multipart upload flow is inconsistent about checksums:

- `pkg/client/transfer.go` prefers the v2 upload flow for large uploads.
- `pkg/backend/upload.go` creates the multipart upload with `ChecksumAlgoSHA256` in `InitiateUploadV2`.
- But `pkg/client/transfer.go` `presignPipeline` does not send per-part checksum info when requesting presigned URLs.
- And `pkg/client/transfer.go` `uploadOnePartV2` uploads parts without setting `x-amz-checksum-sha256`.

As a result, S3 expects SHA256 checksums for each part, but receives no checksum header and rejects the request.

## Why the reported failing part varies
The CLI uploads parts concurrently, so whichever part reaches S3 first can be the first one to fail. The error is not specific to part 4.

## Expected Behavior
If v2 creates a multipart upload with SHA256 checksum requirements, every presigned part request and every `UploadPart` request should consistently include the matching SHA256 checksum.

## Suggested Fix
Either:

1. Fully implement SHA256 checksum support in the v2 path:
   - compute SHA256 for each part,
   - send checksum metadata during presign / presign-batch,
   - include `x-amz-checksum-sha256` on `UploadPart` requests,
   - preserve checksum data through complete/validation as needed.
2. Or avoid initiating v2 multipart uploads with SHA256 requirements until the client path supports that contract end-to-end.

## Relevant Code
- `pkg/client/transfer.go`
- `pkg/backend/upload.go`
- `pkg/s3client/aws.go`

qiffang:
## Summary

Closes #133

`InitiateUploadV2` was creating S3 multipart uploads with `ChecksumAlgorithm=SHA256`, but the v2 client never sends per-part `x-amz-checksum-sha256` headers (`ChecksumContract.Required=false`). S3 rejects parts that lack the declared checksum header with `InvalidRequest`.

**Fix**: Add `ChecksumAlgoNone` and use it for v2 initiate, so S3 does not require per-part checksums. v1 (CRC32C) and patch (SHA256) are unchanged.

### Files changed (4)
| File | Change |
|---|---|
| `pkg/s3client/s3client.go` | Add `ChecksumAlgoNone` constant |
| `pkg/s3client/aws.go` | Handle `ChecksumAlgoNone` — skip `ChecksumAlgorithm` field in `CreateMultipartUploadInput` |
| `pkg/backend/upload.go` | v2 initiate uses `ChecksumAlgoNone` instead of `ChecksumAlgoSHA256` |
| `pkg/s3client/s3client_test.go` | Add `TestMultipartUploadNoChecksumAlgo` regression test |

### What's NOT changed
- v1 upload: still uses `ChecksumAlgoCRC32C`
- patch upload: still uses `ChecksumAlgoSHA256`
- `ChecksumContract.Supported`: still `["SHA-256"]`

## Test plan

- [x] `go test ./pkg/s3client/...` — all 10 tests pass locally
- [ ] CI passes
- [ ] `pkg/backend` tests pass (v2 initiate + presign + complete)
- [ ] `pkg/server` upload tests pass
- [ ] `pkg/client` transfer tests pass

### DIFF
--- a/pkg/backend/upload.go
+++ b/pkg/backend/upload.go
@@ -5,6 +5,7 @@ import (
 	"database/sql"
 	"errors"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/mem9-ai/dat9/pkg/datastore"
@@ -59,6 +60,10 @@ type PresignPartEntry struct {
 }
 
 var ErrPartChecksumCountMismatch = errors.New("part checksum count mismatch")
+
+func normalizeETag(etag string) string {
+	return strings.Trim(etag, "\"")
+}
 
 // S3 returns the S3Client (nil when not configured).
 func (b *Dat9Backend) S3() s3client.S3Client { return b.s3 }
@@ -487,7 +492,7 @@ func (b *Dat9Backend) ConfirmUploadV2(ctx context.Context, uploadID string, clie
 			metrics.RecordOperation("backend", "confirm_upload_v2", "error", time.Since(start))
 			return fmt.Errorf("part %d not found in S3", partNum)
 		}
-		if clientETag != s3ETag {
+		if normalizeETag(clientETag) != normalizeETag(s3ETag) {
 			metrics.RecordOperation("backend", "confirm_upload_v2", "error", time.Since(start))
 			return fmt.Errorf("part %d ETag mismatch: client=%q, S3=%q", partNum, clientETag, s3ETag)
 		}
--- a/pkg/backend/upload.go
+++ b/pkg/backend/upload.go
@@ -225,7 +225,10 @@ func (b *Dat9Backend) InitiateUploadV2(ctx context.Context, path string, totalSi
 	fileID := b.genID()
 	s3Key := "blobs/" + fileID
 
-	mpu, err := b.s3.CreateMultipartUpload(ctx, s3Key, s3client.ChecksumAlgoSHA256)
+	// v2 does not declare a checksum algorithm at the S3 level because the
+	// client doesn't send per-part checksums yet (ChecksumContract.Required=false).
+	// When #114 adds inline checksums, switch back to a concrete algorithm.
+	mpu, err := b.s3.CreateMultipartUpload(ctx, s3Key, s3client.ChecksumAlgoNone)
 	if err != nil {
 		logger.Error(ctx, "backend_initiate_upload_v2_create_multipart_failed", zap.String("path", path), zap.Error(err))
 		metrics.RecordOperation("backend", "initiate_upload_v2", "error", time.Since(start))
--- a/pkg/s3client/aws.go
+++ b/pkg/s3client/aws.go
@@ -87,14 +87,12 @@ func (c *AWSS3Client) CreateMultipartUpload(ctx context.Context, key string, alg
 		Bucket: &c.bucket,
 		Key:    aws.String(c.fullKey(key)),
 	}
-	switch algo {
-	case ChecksumAlgoCRC32C:
-		in.ChecksumAlgorithm = types.ChecksumAlgorithmCrc32c
-	case ChecksumAlgoSHA256:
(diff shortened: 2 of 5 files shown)


pkg/s3client/aws.go
JaySon-Huang, line 1: The new switch silently treats any unrecognized `ChecksumAlgo` the same as `ChecksumAlgoNone` by falling through the `default` branch without error. That fixes the current v2 case, but it also creates a quiet integrity downgrade if a future caller passes a typo or newly added enum value. Since multipart upload creation defines the checksum contract enforced later by S3, `ChecksumAlgoNone` should be an explicit case and unknown values should return an error instead of disabling checksum enforcement.

Source: rule/go-correctness/fail-fast-enum | [Second Opinion](https://github.com/PingCAP-QE/second-opinion)

pkg/backend/upload.go
JaySon-Huang, line 1: This PR fixes the checksum regression, but `make run-server-local` still fails the same v2 upload flow at complete time because the client-submitted ETag and the server-side `ListParts` ETag are compared byte-for-byte without normalization.

I reproduced:

```bash
make run-server-local
DRIVE9_SERVER=http://127.0.0.1:9009 drive9 fs cp /Users/jayson/Projects/claw-memories/terraform-drive9-bench/bench/runs/upload-20260407-182533-r1.bin :/upload-20260407-182533-r1.bin
```

Server log shows:

```text
part 1 ETag mismatch: client="\"4729c1cda90c73295d7e466b6fc4ad34\"", S3="4729c1cda90c73295d7e466b6fc4ad34"
```

The current v2 client forwards the HTTP `ETag` header as-is in the complete payload, which keeps the surrounding quotes. The local mock upload handler also writes quoted HTTP ETags, but `LocalS3Client.ListParts` returns the stored bare value, so `ConfirmUploadV2` rejects the upload even though the underlying part matches.

This means PR 134 fixes the checksum issue but still leaves local v2 multipart uploads broken. We should normalize ETags before comparison in `ConfirmUploadV2` (preferred), or otherwise make the mock `ListParts` representation match the HTTP header representation.

Source: rule/go-correctness/fail-fast-etag-normalization | [Second Opinion](https://github.com/PingCAP-QE/second-opinion)

JaySon-Huang:
Incremental second-opinion review of the follow-up fixes found no new issues.

- `pkg/backend/upload.go` now normalizes quoted and unquoted ETags before comparing client-complete payloads with `ListParts` results, which addresses the reproduced local multipart completion failure.
- `pkg/s3client/aws.go` now treats `ChecksumAlgoNone` as an explicit case and rejects unknown checksum algorithms instead of silently downgrading to no-checksum behavior.
- Validation passed with `go test ./pkg/s3client/...` and `make test`.

Source: review/incremental-no-findings | [Second Opinion](https://github.com/PingCAP-QE/second-opinion)

JaySon-Huang (approved):

PR #148, opened 2026-04-09
Add semantic task lease renewal for long-running semantic tasks

### LINKED ISSUE #147: Add semantic task lease renewal for long-running image and audio extraction
## Enhancement

This issue proposes adding semantic task lease renewal to drive9's durable `semantic_tasks` worker runtime so that long-running extraction tasks can continue safely without being recovered and re-claimed while they are still executing.

This work is related to #145.

Today, the semantic task runtime supports:

- claiming a task with `receipt` and `lease_until`
- acknowledging a finished task
- retrying a failed task
- recovering expired leases back to `queued`

That model is acceptable for short tasks, but it is not sufficient for long-running tasks such as `img_extract_text`, and it will be even riskier for future `audio_extract_text` work. Without explicit lease renewal, a task can still be running when its lease expires, which can lead to:

- the same file revision being re-claimed by another worker
- duplicate image or audio provider calls
- wasted provider cost
- noisy lease-mismatch failures when the original worker finally tries to ack or retry

The proposed direction is:

- add `RenewSemanticTask(...)` to the datastore layer
- add a worker-owned lease renewal loop in `semantic worker`
- make renewal a generic worker-level mechanism for all claimed tasks, not a task-specific patch for image or audio only
- cancel the running handler context when lease renewal shows that the current worker no longer owns the lease
- keep the current `claim / ack / retry / recover` model and at-least-once delivery semantics

Scope for this issue:

- support long-running `img_extract_text` execution safely
- establish the same delivery contract for future `audio_extract_text`
- keep lease duration as a worker-level base lease instead of deriving the default lease from image or audio task timeout
- expose lease-lost behavior through structured logs and operation metrics

Out of scope:

- defining the business payload or provider behavior for `audio_extract_text`
- changing the `semantic_tasks` table shape
- redesigning durable task delivery into exactly-once execution

Validation should cover:

- renewing a valid lease successfully
- rejecting renewals for expired or mismatched receipts
- ensuring a long-running extraction task is not re-claimed while renewal is healthy
- cancelling the handler when lease ownership is lost
- verifying that image extraction and future audio extraction can both use the same worker-level mechanism

JaySon-Huang:
## Summary
- Add datastore lease renewal primitive so workers can extend task ownership while processing long-running semantic tasks.
- Add a worker-level lease renewal loop in semantic task execution, including lease-lost cancellation behavior to stop expensive handlers early.
- Keep existing claim/ack/retry/recover and at-least-once delivery semantics, while removing dependence on image-timeout-derived default lease behavior.
- Verify behavior with datastore and worker tests covering renew success/failure, long-task stability, and lease-lost handling.

Closes #147

## Architecture Overview

```text
Current
-------

Worker A                semantic_tasks store        Recover loop / Worker B
   |                           |                            |
   | claim(task, receiptA, t1) |                            |
   |-------------------------->|                            |
   |<--------------------------|                            |
   |                           |                            |
   | [ handler keeps running ] |                            |
   |                           |                            |
   |        ( lease expires )  |                            |
   |                           | recover expired task       |
   |                           |<---------------------------|
   |                           | mark queued                |
   |                           |--------------------------->|
   |                           |                            |
   |                           |<---------------------------|
   |                           | Worker B claim(task,       |
   |                           | receiptB, t2)              |
   |                           |--------------------------->|
   |                           |                            |
   | [ handler finally returns ]                            |
   | ack/retry(task, receiptA) |                            |
   |-------------------------->|                            |
   |<----- ErrTaskLeaseMismatch|                            |
   |                           |                            |


Target
------

Worker A                Worker-owned renew loop      semantic_tasks store
   |                           |                            |
   | claim(task, receiptA, t1) |                            |
   |-------------------------->|                            |
   |<-------------------------------------------------------|
   |                           |                            |
   | build lease-bound ctx     |                            |
   | start handler(ctx)        |                            |
   |-------------------------->|                            |
   |                           | renew(task, receiptA, t2)  |
   |                           |--------------------------->|
   |                           |<---------------------------|
   |                           |         success            |
   |                           |                            |
   | [ handler keeps running ] |                            |
   |                           | renew(task, receiptA, t3)  |
   |                           |--------------------------->|
   |                           |<---------------------------|
   |                           |     lease lost / not found |
   |<--------------------------|                            |
   | cancel handler ctx        |                            |
   | stop execution promptly   |                            |
   | skip ack/retry            |                            |
   |                           |                            |

Successful completion path:

Worker A                Worker-owned renew loop      semantic_tasks store
   |                           |                            |
   | ... renew succeeds while handler runs ...              |
   | [ handler returns ]       |                            |
   | stop renew loop           |                            |
   | ack/retry(task, receiptA) |                            |
   |------------------------------------------------------->|
   |<-------------------------------------------------------|
   |                     success                            |
```

## Test plan
- [x] go test -v ./pkg/datastore/...
- [x] go test -v ./pkg/server/...
- [x] make test

Made with [Cursor](https://cursor.com)

## Summary by CodeRabbit

* **New Features**
  * Automatic lease renewal and cancellation handling for long-running semantic tasks to prevent premature timeouts and ensure safe task cancellation on lease loss.

* **Chores**
  * Simplified lease-duration behavior to use a fixed default unless explicitly configured.

* **Tests**
  * Added tests for lease renewal, lease-loss cancellation, panic handling, and renewal idempotency.

* **Documentation**
  * Added GitHub issue templates and updated contributor guidance.


.github/ISSUE_TEMPLATE/enhancement.md
```diff
@@ -0,0 +1,7 @@
+---
+name: "\U0001F680 Enhancement"
+about: As a TiDB developer, I want to make an enhancement.
+labels: type/enhancement
+---
+
+## Enhancement
```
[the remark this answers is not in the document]
JaySon-Huang: Resolved by https://github.com/mem9-ai/drive9/pull/148/changes/853d0b59c04e8927cc1504cf16a803ba420c28f1

cmd/drive9-server-local/main.go
JaySon-Huang, line 1: This branch correctly removes the old implicit lease derivation and fixes the CLI help to say the semantic lease default is a fixed `30`, but the local helper script is still stale: `scripts/drive9-server-local-env.sh` still documents `max(30, 2x image extract timeout)` when `DRIVE9_SEMANTIC_LEASE_SECONDS` is unset. That mismatch will confuse local validation because the binary and the helper script now describe different defaults.

Source: local/second_opinion | [Second Opinion](https://github.com/PingCAP-QE/second-opinion)

pkg/server/semantic_worker.go
JaySon-Huang, line 1: There is still a stop/renew race here that can downgrade a real lease-loss event into a plain `ack_failed`. If `processTask()` closes `stopCh` while a `RenewSemanticTask(...)` call is already in flight, `run()` will hit `shouldStop()` first and return before `markLeaseLost()`. In that case `processTask()` sees `leaseOutcome.lost == false` and can still attempt `AckSemanticTask()`, even though the lease may already belong to another worker. The database should reject that ack, so this is mostly an ownership-observability bug, but it breaks the proposal's intended contract of "lease lost => no ack/retry, emit `semantic_worker_lease_lost`" and makes production diagnosis harder.

Source: local/second_opinion | [Second Opinion](https://github.com/PingCAP-QE/second-opinion)

JaySon-Huang:
Incremental second-opinion review for `42a7b55..81db442`: no new findings.

The new commits appear to address the previously reported issues correctly:
- the lease-loss vs `stop()` race is now handled in the right order, so in-flight renewals can still surface `lease_lost` during shutdown
- panic and double-stop paths now stop lease renewal safely
- the stale local env helper comment about implicit lease derivation has been aligned with the current fixed default of `30`

Remaining risk is low and mostly unchanged from the prior review:
- long-running renew coverage is still image-task focused; there is no equivalent long-running embed-path integration test
- some datastore lease-expiry tests still rely on `time.Sleep`
- `RenewSemanticTask()` now has a defensive fallback branch that seems effectively unreachable under the current `semanticTaskLeaseError()` implementation, so it is harmless but slightly redundant

Source: local/second_opinion_incremental | [Second Opinion](https://github.com/PingCAP-QE/second-opinion)

qiffang:
## Review: Semantic Task Lease Renewal

整体设计思路正确：将原来依赖 `max(30, 2x image_timeout)` 的静态 lease 改为运行期 renewal loop，让长任务（图片 OCR、未来的音频转录等）不受固定 lease 窗口限制。以下是 review 意见汇总：

---

### CRITICAL — stop/renew race 导致 lease-lost 被降级为 ack_failed

*来源: Second Opinion, Copilot*

`processTask()` 调用 `stopLease()` 关闭 `stopCh`，但如果此时 `RenewSemanticTask()` 已经 in-flight，`run()` 会在 renew 返回后先检查 `shouldStop()` 而不是先走 `markLeaseLost()` 分支。这意味着：

1. renew 返回 `ErrTaskLeaseMismatch` → 本该 `markLeaseLost()` → `cancel()` → 让 `processTask` 看到 `leaseOutcome.lost == true`
2. 但 `shouldStop()` 先返回 true → `run()` 直接 return → `lost` 保持 false → `processTask` 继续 `AckSemanticTask()`

虽然数据库会拒绝这个 ack（receipt 不匹配），但 observability 上 `lease_lost` metric 和 log 都丢了，生产排查时会误判为 `ack_failed` 而非 lease 被抢。

建议修复（在 `run()` 的 renew error 分支里，先 check `markLeaseLost` 再 check `shouldStop`）：

```go
case <-ticker.C:
    e.markRenewAttempted()
    leaseUntil, err := target.store.RenewSemanticTask(...)
    if err != nil {
        if errors.Is(err, semantic.ErrTaskLeaseMismatch) || errors.Is(err, semantic.ErrTaskNotFound) {
            // Lease mismatch always wins over stop — even if processTask
            // already signalled shutdown, an in-flight renew discovering
            // ownership loss must be surfaced.
            e.markLeaseLost()
            e.cancel()
            return
        }
        if e.shouldStop() { return }
        // transient error — continue
        continue
    }
```

当前代码看起来已经是这个顺序了（先 check mismatch，再 check shouldStop），但 Second Opinion 报告说 race 仍然存在——请 double check `ctx.Done()` select case 是否也可能在 mismatch check 之前触发。

---

### IMPORTANT — `RenewSemanticTask` 的 unreachable return path

*来源: CodeRabbit, Copilot*

```go
err = s.semanticTaskLeaseError(ctx, taskID)
if err == nil {
    err = semantic.ErrTaskLeaseMismatch
}
return time.Time{}, err
```

`semanticTaskLeaseError()` 在当前实现下总是返回 non-nil error（ErrTaskNotFound 或 ErrTaskLeaseMismatch），所以 `err == nil` 分支是 defensive fallback。建议简化为：

```go
return time.Time{}, s.semanticTaskLeaseError(ctx, taskID)
```

或者保留 defensive 版本但加注释说明为什么。当前写法让读者误以为有一个 success path。

---

### IMPORTANT — logStopped 里的 fragile array index mutation

*来源: CodeRabbit*

```go
fields[1] = zap.String("result", "lease_lost")  // 假设 fields[1] 永远是 "result"
```

如果未来 fields 的构造顺序变了，这里会静默覆盖错误的 field。建议：

```go
result := "owned"
if lost {
    result = "lease_lost"
}
fields := append([]zap.Field{
    zap.String("tenant_id", target.tenantID),
    zap.String("result", result),
}, semanticTaskLogFields(task)...)
```

---

### MINOR — Issue templates 从 TiDB 复制的痕迹

已经修复（从 `.github/workflows/ISSUE_TEMPLATE/` 移到了 `.github/ISSUE_TEMPLATE/`，TiDB 版本信息也更新了）。但 enhancement 模板内容过于单薄，建议补充结构化引导（current behavior / proposed improvement / benefits）。

---

### MINOR — 测试只覆盖了 image extraction 路径

Lease renewal 测试都是通过 `gatedServerImageExtractor` 走 `processImgExtractTask` 路径。`processEmbedTask` 也在同一个 leaseExec 上下文里运行，但没有对应的长任务 renewal 测试。建议至少加一个 `gatedServerEmbedExtractor` 的 mirror test，验证 embed 路径的 renewal 也正常工作。

---

### MINOR — datastore 测试里的 `time.Sleep(40ms)` 

`TestSemanticTaskRenewExpiredLeaseFails` 用 `time.Sleep(40ms)` 等待 20ms lease 过期。这在 CI 上可能 flake。建议用更长的 lease 或直接用 DB 操作把 `lease_until` 设到过去。

---

### 总结

| 级别 | 问题 | 建议 |
|------|------|------|
| CRITICAL | stop/renew race 可能丢失 lease_lost 信号 | 确保 mismatch 检测优先于 stop 检测 |
| IMPORTANT | `RenewSemanticTask` unreachable path | 简化或加注释 |
| IMPORTANT | fragile `fields[1]` mutation | 用变量替代 index mutation |
| MINOR | enhancement template 太简单 | 补充结构化引导 |
| MINOR | 只测了 image 路径 | 加 embed 路径 mirror test |
| MINOR | Sleep-based lease expiry test | 用 DB 操作替代 sleep |

整体：设计方向正确，核心逻辑健壮。解决 stop/renew race 的 observability 问题后可以 merge。

JaySon-Huang:
Thanks for the detailed review. Here is the current status item by item.

1. stop/renew race
We re-checked the current implementation after the follow-up changes. The renew error path now handles `ErrTaskLeaseMismatch` / `ErrTaskNotFound` before `shouldStop()`, so an in-flight renew that discovers lease loss during shutdown should still surface `lease_lost` rather than being downgraded to a plain stop path. We are not planning another logic change for this PR.

That said, we agree an even stronger race-focused test would improve confidence. Rather than adding an ad hoc test-only hook to `semanticWorkerManager` in this PR, we plan to track that as part of the failpoint-based deterministic race-testing work in #149. We have added an explicit TODO there for the `stopLease()` vs in-flight renew boundary.

2. `RenewSemanticTask()` defensive fallback
Agreed that this branch is defensive rather than part of a meaningful success path. We decided to keep it, but added a clarifying comment so the intent is explicit and it is not mistaken for a real nil-error path.

3. `logStopped()` fragile `fields[1]` mutation
Agreed. We updated this to compute the `result` string first and then build the field slice directly, instead of mutating by index.

4. enhancement template structure
We agree the enhancement template can be improved, but we prefer not to expand the scope of this PR further for repository-template cleanup. We are treating that as separate repository hygiene / follow-up work.

5. image-only renewal coverage
Agreed. An embed-path mirror test would improve confidence because lease renewal is intended to be a worker-level mechanism for all claimed tasks, not just image extraction. We opened a follow-up issue to track that coverage improvement: #154.

6. `TestSemanticTaskRenewExpiredLeaseFails` using `time.Sleep(40ms)`
Agreed that this is not ideal. For this PR, we prefer not to make a one-off workaround just for that test. Instead, we are tracking it under the failpoint work in #149, where we plan to migrate timing-sensitive lease-expiry tests to deterministic failpoint-controlled checks. We added an explicit TODO there for `TestSemanticTaskRenewExpiredLeaseFails`.

So for this PR, the concrete updates are:
- fixed the fragile `logStopped()` field construction
- clarified the defensive fallback in `RenewSemanticTask()`

And the remaining test-strengthening items are now explicitly tracked as follow-ups:
- #149 for failpoint-based deterministic race / lease-expiry coverage
- #154 for embed-path lease-renewal mirror coverage

qiffang (approved):
## LGTM — well-designed lease renewal mechanism

The core design is sound: a background goroutine renews the lease, cancels the handler context on lease loss, and cleans up properly via `defer stopLease()`. Tests are thorough covering renewal, lease-loss cancellation, panic safety, and idempotent stop.

### One suggestion worth considering

**`LeaseDuration/2` renewal interval vs transient failure resilience** (`pkg/server/semantic_worker.go:600`)

With `LeaseDuration/2` (15s for 30s lease), a single transient renewal failure at t=15s means the next retry fires at t=30s — exactly when the lease expires. This means one transient DB hiccup can cause unnecessary lease loss. Using `LeaseDuration/3` (~10s interval) would allow one retry (at ~20s) before the 30s expiry. Not blocking, but worth considering for production resilience.

### Minor nits (non-blocking)

- The `run()` method logs via `e.ctx` after calling `e.cancel()` on lease loss (lines ~679-689). This works fine since most Go loggers ignore context cancellation, but a comment noting this would help future maintainers.
- Adding brief doc comments to the `semanticTaskLeaseExecution` struct fields (`stopCh`, `doneCh`, `lost`, etc.) would improve readability of the lifecycle protocol.

### Strengths

- Clean separation between task dispatch and lease lifecycle
- Correct use of parent `ctx` for ack/retry (not the lease context)
- Defensive fallback in `RenewSemanticTask` preventing false-success
- `stopOnce` guarantees idempotent cleanup
- Comprehensive test coverage: happy path, lease loss, panic, idempotent stop, and 5 datastore-level tests
- Clean removal of the image-timeout-derived lease heuristic — fulfills the existing TODO

PR #226, opened 2026-04-15
Add upload benchmark diagnostics and AWS-backed local-server support

JaySon-Huang:
## Summary
- add upload benchmark diagnostics across the CLI and server, including upload summaries, optional CPU profiling, and server-side timing instrumentation for initiate/presign/confirm/finalize plus auth and tenant-pool phases
- let `drive9-server-local` target AWS S3 directly so local benchmark runs can exercise the same object storage path as hosted deployments
- harden local benchmark and test ergonomics, including Podman setup diagnostics and fallback-to-default-runtime behavior when Podman setup is unavailable
- address review follow-ups around local AWS env handling, `drive9-server-local` env parsing, CLI summary context propagation, upload lookup error handling, presign timing semantics, and hot-path benchmark timing log gating via `DRIVE9_BENCH_TIMING_LOG_ENABLED`

## Testing
- `make lint`
- `make build`
- `make test TEST_P=1`
- `make test-failpoint`

### DIFF
--- a/pkg/backend/upload.go
+++ b/pkg/backend/upload.go
@@ -410,7 +410,7 @@ func (b *Dat9Backend) InitiateUploadV2IfRevision(ctx context.Context, path strin
 		return nil, err
 	}
 	txDurationMs := uploadPhaseMs(txStart)
-	logger.Info(ctx, "backend_initiate_upload_v2_timing",
+	logger.InfoBenchTiming(ctx, "backend_initiate_upload_v2_timing",
 		zap.String("path", path),
 		zap.Int64("total_size", totalSize),
 		zap.Int("total_parts", len(parts)),
@@ -595,7 +595,7 @@ func (b *Dat9Backend) PresignParts(ctx context.Context, uploadID string, entries
 	if len(entries) > 0 {
 		s3PresignAvgMs = s3PresignTotalMs / float64(len(entries))
 	}
-	logger.Info(ctx, "backend_presign_parts_timing",
+	logger.InfoBenchTiming(ctx, "backend_presign_parts_timing",
 		zap.String("upload_id", uploadID),
 		zap.String("status_before", string(statusBefore)),
 		zap.Int("entries_total", len(entries)),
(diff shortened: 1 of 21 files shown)


scripts/drive9-server-local-env.sh
JaySon-Huang, line 1: This change makes the most natural AWS validation flow fail: if a user follows the existing pattern and first runs `source ./scripts/drive9-server-local-env.sh`, the script eagerly exports `DRIVE9_S3_DIR`. If they then set `DRIVE9_S3_BUCKET` to switch into AWS mode, `DRIVE9_S3_DIR` is still present in the environment, and `localS3ConfigFromEnv()` now aborts because both variables are set. That means the new AWS mode is only usable if callers remember to export the bucket _before_ sourcing the helper, which is an unexpected regression for local-server workflows. Source: second-opinion/local-server-env-order | [Second Opinion](https://github.com/PingCAP-QE/second-opinion)

pkg/backend/upload.go
JaySon-Huang, line 1: The new timing diagnostics are all unconditional `Info` logs on successful hot paths. A single multipart upload now emits extra success logs from auth, tenant-pool acquire, datastore get-upload, initiate, presign, confirm, and finalize with no feature flag or sampling gate. That materially changes the runtime behavior of the very benchmark path this PR is trying to measure, and it also increases production log volume for every healthy request. These diagnostics need an explicit opt-in (or at least a very narrow benchmark-only gate), otherwise the measurements will be self-perturbed by the instrumentation itself. Source: second-opinion/hot-path-success-logs | [Second Opinion](https://github.com/PingCAP-QE/second-opinion)

pkg/client/transfer.go
JaySon-Huang, line 1: `UploadSummary.PresignSeconds` is not measuring pure presign latency here. The timer wraps the entire `presignPipeline()` goroutine lifetime, but that goroutine also blocks when sending results into `presignCh`, so the reported duration includes upload-side backpressure and overlap with the consumer path. If this field is later used to conclude that control-plane presign is the bottleneck, it will systematically overstate presign time. Source: second-opinion/presign-summary-scope | [Second Opinion](https://github.com/PingCAP-QE/second-opinion)

JaySon-Huang:
@coderabbitai review

qiffang (approved):

PR #242, opened 2026-04-16
feat: add incremental HTTP and CLI append support

Lloyd-Pottiger:
## Summary
- add `POST /v1/fs/{path}?append` to initiate incremental append uploads for existing S3-backed files
- add `Client.AppendStream(...)` and `drive9 fs cp --append` on top of the existing patch/multipart flow
- add coverage for append planning, client fallback behavior, and CLI append uploads

## Verification
- `git diff --check`
- `go test ./cmd/drive9/cli -run 'Test.*Append|TestUpload'`
- `go test -c -o /tmp/drive9-testbins/server.test ./pkg/server`
- `go test -c -o /tmp/drive9-testbins/client.test ./pkg/client`

## Summary by CodeRabbit

* **New Features**
  * Incremental append uploads: CLI adds --append (local→remote only, incompatible with --resume or stdin) and server exposes POST /v1/fs/{path}?append.

* **Documentation**
  * Added CLI example: drive9 fs cp --append ./tail.log :/logs/app.log
  * Documented POST /v1/fs/{path}?append.

* **Bug Fixes**
  * Improved server error mapping/status codes and introduced explicit backend error sentinels.

* **Tests**
  * Added extensive client/server tests covering append flows, fallbacks, validation, and edge cases.


pkg/client/append.go
```diff
@@ -0,0 +1,160 @@
...
+
+func isClientNotFound(err error) bool {
+	var statusErr *StatusError
+	if errors.As(err, &statusErr) {
+		return statusErr.StatusCode == http.StatusNotFound
+	}
+	return strings.HasPrefix(err.Error(), "not found: ")
+}
+
+func shouldRewriteAppend(err error) bool {
```
JaySon-Huang, line 184: This fallback only recognizes `400 file is not S3-stored`. Older servers that do not implement `?append` return `400 unknown POST action`, so appending to a missing file or a small file still fails instead of degrading to create or rewrite on the client side. `writeStreamConditionalWithSummary` already has compatibility fallback logic for older upload APIs; can we mirror that behavior here?

pkg/client/append.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+
+	stat, err := c.StatCtx(ctx, path)
+	if err != nil {
+		if isClientNotFound(err) {
+			return c.WriteStreamConditional(ctx, path, r, size, progress, 0)
+		}
+		return err
+	}
+	if stat.IsDir {
```
JaySon-Huang, line 38: The client derives `partSize` from the current object size instead of the final size after append. If the file is already near the 10000-part boundary, a small append can push `newSize` over the limit while still sending the old 8 MiB part size to the server. Because `InitiateAppendUploadIfRevision` only recalculates when the client value is smaller than `MinPartSize`, this turns into an avoidable append failure. Can we base the initial part size on `stat.Size + size`, or retry without a client-supplied part size when the server rejects it for part-count reasons?

cmd/drive9/cli/cp.go
```diff
@@ -61,6 +72,9 @@ func Cp(c *client.Client, args []string) error {
 		return streamToStdout(ctx, c, srcRP.Path)
 
 	case !srcIsRemote && dstIsRemote:
+		if appendMode {
```
JaySon-Huang, line 81: `--append` is only meaningful for local or stdin to remote uploads, but the flag is silently ignored for `remote -> local` and `remote -> remote` paths. For example, `drive9 fs cp --append :/a :/b` currently performs a normal server-side copy. Can we reject unsupported source and destination combinations so the CLI does not do a different operation from what the user requested?

pkg/client/append.go
Lloyd-Pottiger: Handled in `5a5045c`. `shouldRewriteAppend` now treats `400 unknown POST action` from older servers as a compatibility fallback, so append degrades to the existing client-side rewrite path instead of failing. Added regression coverage in `pkg/client/append_test.go` for that server response.

pkg/client/append.go
Lloyd-Pottiger: This was already addressed in `ed092e3`. `AppendStream` now computes the client hint from `stat.Size + size` before calling `initiateAppend`, and there is regression coverage in `pkg/client/append_test.go` (`TestAppendStreamUsesFinalSizeForAdaptivePartSize`).

cmd/drive9/cli/cp.go
Lloyd-Pottiger: Handled in `5a5045c`. `cp --append` now rejects unsupported combinations before making any request: it only accepts local-file or stdin source to a remote destination. Added CLI coverage for both `remote -> local` and `remote -> remote` rejection cases in `cmd/drive9/cli/cp_test.go`.

pkg/client/append.go
```diff
@@ -0,0 +1,160 @@
...
+// AppendStream appends bytes from r to path. Missing paths are created, small
+// existing files fall back to read-modify-write, and large S3-backed files use
+// the incremental append upload flow.
+func (c *Client) AppendStream(ctx context.Context, path string, r io.Reader, size int64, progress ProgressFunc) error {
+	if size < 0 {
+		return fmt.Errorf("append size must be non-negative")
+	}
+	if size == 0 {
+		return nil
+	}
```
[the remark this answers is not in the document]
Lloyd-Pottiger: Verified against the current branch: `AppendStream` does the `StatCtx` call first, and the not-found path already falls through to `WriteStreamConditional(..., 0)`, so a zero-byte append still creates the missing file. There is regression coverage for this in `pkg/client/append_test.go` (`TestAppendStreamZeroSizeCreatesMissingFile`). No code change was needed for this thread.

Lloyd-Pottiger:
Addressed the remaining append follow-ups in `95fb64c`:
- cap `handleAppend` request-body decoding with `http.MaxBytesReader(..., 1<<20)` and add a regression test for oversized append-init bodies
- remove the redundant `start` reassignment in `appendDirtyPartNumbers`
- verify the older zero-byte append thread against current code/tests and resolve it without an extra code change

Focused verification run under `scripts/test-podman.sh`:
- `go test ./pkg/client -run TestAppendStreamZeroSizeCreatesMissingFile -count=1`
- `go test ./pkg/backend -run TestPatchAndAppendRejectDBBackedFilesWithSentinel -count=1`
- `go test ./pkg/server -run "TestAppend(MissingPathReturnsNotFound|DBBackedFileReturnsBadRequest|RejectsOversizedRequestBody)$" -count=1`

pkg/backend/patch.go
JaySon-Huang, line 1: Append relies on the overlapping tail part keeping the original bytes before the new suffix, but this branch only logs and keeps going when `PresignGetObjectRange()` fails. The client then treats the missing prefix as empty and starts writing append data from offset 0, which can either drop the old tail bytes silently or fail with EOF when the appended payload is smaller than the dirty part. Please make read-url generation mandatory for append parts that still overlap existing bytes, or fail fast on the client when such a part arrives without `read_url`.

cmd/drive9/cli/cp.go
JaySon-Huang, line 1: `drive9 fs cp --append - :/path` now buffers the entire stdin stream with `io.ReadAll()` before calling `AppendStream()`. That makes the new append CLI path lose its streaming/resource behavior specifically for pipe and large-log inputs, and memory usage scales linearly with the input size. Please either spool stdin to a temp file before append, or reject `--append` with stdin explicitly instead of buffering everything in RAM.

Source: rule/XUHUAIYU-PR-001 | [Second Opinion](https://github.com/PingCAP-QE/second-opinion)

pkg/backend/patch.go
Lloyd-Pottiger: Handled in `4bb1c93`. I took the fail-fast path on the client side: `AppendStream` now validates the append plan before uploading and rejects any upload part that still overlaps existing bytes but arrives without `read_url`, so we no longer risk silently dropping the original tail bytes. Added regression coverage in `pkg/client/append_test.go` (`TestAppendStreamRejectsOverlappingAppendPartWithoutReadURL`).

cmd/drive9/cli/cp.go
Lloyd-Pottiger: Handled in `4bb1c93`. Instead of buffering stdin into memory for append, `cp --append` now rejects `-` explicitly and requires a local file source for append mode. Added CLI coverage in `cmd/drive9/cli/cp_test.go` (`TestCpAppendRejectsStdinSource`).

cmd/drive9/cli/cp.go
```diff
@@ -78,43 +98,58 @@ func Cp(c *client.Client, args []string) error {
...
+	f, size, err := openLocalFile(localPath)
 	if err != nil {
-		return fmt.Errorf("open %s: %w", localPath, err)
+		return err
 	}
 	defer func() { _ = f.Close() }()
 
-	info, err := f.Stat()
+	return c.AppendStream(ctx, remotePath, f, size, printProgress)
+}
```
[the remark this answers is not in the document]
Lloyd-Pottiger: Handled in `9b3ea86`. `appendFile` now records a minimal `upload_summary` event around `AppendStream` with append-specific timing/byte metadata (`mode: "append_v1"`, `total_bytes`, `remote_path`, `local_path`) so append uploads show up in the same stable CLI log stream as normal uploads. Added regression coverage in `cmd/drive9/cli/cp_test.go` (`TestCpAppendEmitsUploadSummaryToCLILog`).

cmd/drive9/cli/cp.go
```diff
@@ -98,25 +114,61 @@ func uploadFile(ctx context.Context, c *client.Client, localPath, remotePath str
...
+	defer func() { _ = f.Close() }()
+
+	startedAt := time.Now()
+	if err := c.AppendStream(ctx, remotePath, f, size, printProgress); err != nil {
+		return err
+	}
+	finishedAt := time.Now()
+	elapsedSeconds := finishedAt.Sub(startedAt).Seconds()
+	emitUploadSummary(ctx, &client.UploadSummary{
+		Type:           "upload_summary",
```
JaySon-Huang: This new summary is only accurate for the native incremental-append happy path. `client.AppendStream()` can also fall back to `appendByRewrite()` when the target does not support append natively, and that path ultimately rewrites the whole final file via `WriteStreamConditional(..., finalSize, ...)`. In that case this code still logs `mode=append_v1` and `total_bytes=<append size>`, which under-reports the transferred bytes and misclassifies the operation for telemetry / benchmarks. Please either have append return a real summary from the client layer, or at least distinguish native append from rewrite/create fallback before emitting this log.

Source: rule/XUHUAIYU-PR-006 | [Second Opinion](https://github.com/PingCAP-QE/second-opinion)

Lloyd-Pottiger: Handled in `09e62fb`. I moved append summary classification down into the client layer so the CLI no longer guesses. `AppendStreamWithSummary` now returns the actual completed path: native incremental append stays `append_v1`, missing-target create is prefixed as `append_create_*`, and rewrite fallback is prefixed as `append_rewrite_*` while preserving the underlying upload mode and `total_bytes` from the real write path. `appendFile` now emits that returned summary directly. Added regression coverage in `cmd/drive9/cli/cp_test.go` (`TestCpAppendEmitsRewriteUploadSummaryToCLILog`).

JaySon-Huang (approved):

PR #244, opened 2026-04-16
s3: support custom endpoints and static credentials

Lloyd-Pottiger:
## Summary
- add custom S3 endpoint, path-style, and static credential support to the AWS S3 client
- thread the new S3 configuration through `drive9-server`, `drive9-server-local`, and tenant pool construction
- add coverage for the new S3 config parsing/validation and document a MinIO local setup example

## Testing
- make test

## New Features

* **S3-Compatible Storage Support**: Added configuration options for custom S3 endpoints and path-style addressing, enabling compatibility with S3-compatible services like MinIO.
* **Static Credentials**: Introduced support for static AWS credentials (access key, secret access key, and session token) as an alternative to default credential chains.
* **Configuration Validation**: Added validation to ensure credential configurations are consistent and complete.

## Documentation

* Updated environment variable documentation to reflect new S3 configuration options (`DRIVE9_S3_ENDPOINT`, `DRIVE9_S3_FORCE_PATH_STYLE`, and credential-related variables).


JaySon-Huang (commented):
One concern here: `drive9-server` still doesn't fail fast on invalid combinations of the new static S3 credential env vars.

At startup we parse and log `DRIVE9_S3_ACCESS_KEY_ID` / `DRIVE9_S3_SECRET_ACCESS_KEY` / `DRIVE9_S3_SESSION_TOKEN`, but we don't validate them there. The actual validation only happens later when a tenant backend is created via `tenant.Pool.createBackend()` -> `s3client.NewAWS()`.

That means a configuration error like setting `DRIVE9_S3_ACCESS_KEY_ID` without `DRIVE9_S3_SECRET_ACCESS_KEY` will still let `drive9-server` boot successfully, and only fail on the first tenant request that needs S3. I think we should reuse the same validation during server startup so these misconfigurations are surfaced immediately rather than becoming runtime failures.

Lloyd-Pottiger:
@JaySon-Huang addressed in `c40b5d7`: `drive9-server` now validates the static S3 credential env combinations during startup by reusing `s3client.AWSConfig.Validate()`, so invalid `DRIVE9_S3_ACCESS_KEY_ID` / `DRIVE9_S3_SECRET_ACCESS_KEY` / `DRIVE9_S3_SESSION_TOKEN` combinations fail fast before any tenant request creates an S3 backend.

@coderabbitai followed up in `90d13d9`: I collapsed the redundant validation branches out of `staticCredentialsProvider()` so `Validate()` owns the AK/SK/session-token rules, and I moved the credential/role log labels into `pkg/s3client` so both server entrypoints share the same source of truth.

Re-verified with:
- `go test ./pkg/s3client ./cmd/drive9-server ./cmd/drive9-server-local`
- `make test`

JaySon-Huang (approved):

PR #254, opened 2026-04-16
fix: convert vault SQL to TiDB/MySQL dialect

qiffang:
## Summary
Convert vault SQL from PostgreSQL to TiDB/MySQL dialect and wire vault DDL into TiDB tenant schema init.

### Changes
- **`pkg/vault/store.go`**: `$N`→`?` placeholders, `ON CONFLICT DO NOTHING`→`INSERT IGNORE`
- **`pkg/vault/schema.go`**: Delegates to `schema.VaultTiDBSchemaStatements()` (single source of truth)
- **`pkg/tenant/schema/vault.go`** (new): Canonical vault TiDB DDL — `BLOB`, `DATETIME(3)`, `JSON`, `CURRENT_TIMESTAMP(3)`, inline `INDEX`
- **`pkg/tenant/schema/tidb_app.go`**: Appends vault DDL to app-embedding schema statements
- **`pkg/tenant/schema/tidb_auto.go`**: Appends vault DDL to auto-embedding schema statements; `ValidateTiDBSchemaForMode` now validates all 6 vault tables
- **`pkg/tenant/db9/schema.go`**: Removed PG-native vault DDL (vault is TiDB/MySQL-only per product decision)
- **`cmd/drive9-server/schema_test.go`**: Vault table checks moved from db9 to tidb_zero/tidb_cloud_starter

## Context
Vault tables failed to create on TiDB tenant DB with syntax error:
> You have an error in your SQL syntax ... near "BYTEA NOT NULL, created_at TIMESTAMPTZ NOT NULL DEFAULT NOW()"

Root cause: `pkg/vault` used PostgreSQL dialect but vault runs on tenant DB via TiDB/MySQL datastore path.

## Design decisions
- **Single source of truth**: Vault DDL defined once in `schema.VaultTiDBSchemaStatements()`, consumed by TiDB init and `vault.SchemaStatements()`
- **TiDB-only**: Per @qiffang's directive, vault is TiDB/MySQL-only. PG vault DDL removed from db9.
- **INSERT IGNORE**: Used only in `GetOrCreateDEK` for insert-if-not-exist + re-read pattern. Only suppresses duplicate key on `vault_deks` PK. Re-read ensures correctness.
- **Schema validation**: All 6 vault tables added to `ValidateTiDBSchemaForMode` via `ensureTiDBTableExists`

## Test plan
- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [ ] CI green
- [ ] Vault tables created on TiDB tenant DB (schema dump includes vault DDL)

### DIFF
--- a/cmd/drive9-server/schema_test.go
+++ b/cmd/drive9-server/schema_test.go
@@ -34,17 +34,21 @@ func TestSchemaDumpInitSQLByProvider(t *testing.T) {
 }
 
 func TestSchemaDumpInitSQLByProviderIncludesVault(t *testing.T) {
-	out := captureSchemaStdout(t, func() {
-		if err := runSchemaCommand([]string{"dump-init-sql", "--provider", "db9"}); err != nil {
-			t.Fatalf("dump provider schema: %v", err)
-		}
-	})
+	for _, provider := range []string{"tidb_zero", "tidb_cloud_starter"} {
+		t.Run(provider, func(t *testing.T) {
+			out := captureSchemaStdout(t, func() {
+				if err := runSchemaCommand([]string{"dump-init-sql", "--provider", provider}); err != nil {
+					t.Fatalf("dump provider schema: %v", err)
+				}
+			})
 
-	if !strings.Contains(out, "CREATE TABLE IF NOT EXISTS vault_deks") {
-		t.Fatalf("db9 dump missing vault_deks: %q", out)
-	}
-	if !strings.Contains(out, "CREATE TABLE IF NOT EXISTS vault_audit_log") {
-		t.Fatalf("db9 dump missing vault_audit_log: %q", out)
+			if !strings.Contains(out, "CREATE TABLE IF NOT EXISTS vault_deks") {
+				t.Fatalf("%s dump missing vault_deks: %q", provider, out)
+			}
+			if !strings.Contains(out, "CREATE TABLE IF NOT EXISTS vault_audit_log") {
+				t.Fatalf("%s dump missing vault_audit_log: %q", provider, out)
+			}
+		})
 	}
 }
 
--- a/pkg/server/auth.go
+++ b/pkg/server/auth.go
@@ -25,6 +25,7 @@ type TenantScope struct {
 	TenantID     string
 	APIKeyID     string
 	TokenVersion int
+	Provider     string
 	Backend      *backend.Dat9Backend
 }
 
@@ -162,7 +163,7 @@ func tenantAuthMiddleware(metaStore *meta.Store, pool *tenant.Pool, tokenSecret 
 			zap.Float64("total_ms", authPhaseMs(authStart)),
 		)
 
-		scope := &TenantScope{TenantID: resolved.Tenant.ID, APIKeyID: resolved.APIKey.ID, TokenVersion: resolved.APIKey.TokenVersion, Backend: b}
+		scope := &TenantScope{TenantID: resolved.Tenant.ID, APIKeyID: resolved.APIKey.ID, TokenVersion: resolved.APIKey.TokenVersion, Provider: resolved.Tenant.Provider, Backend: b}
 		next.ServeHTTP(w, r.WithContext(withScope(r.Context(), scope)))
 	})
 }
@@ -204,7 +205,7 @@ func (s *Server) capabilityAuthMiddleware(metaStore *meta.Store, pool *tenant.Po
 		}
 		defer release()
 
-		scope := &TenantScope{TenantID: tenantID, Backend: b}
+		scope := &TenantScope{TenantID: tenantID, Provider: tenant.Provider, Backend: b}
 		sub := strings.TrimPrefix(r.URL.Path, "/v1/vault/read")
 		s.handleVaultRead(w, r.WithContext(withScope(r.Context(), scope)), sub)
 	})
--- a/pkg/server/vault.go
+++ b/pkg/server/vault.go
@@ -10,6 +10,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/mem9-ai/dat9/pkg/tenant"
 	"github.com/mem9-ai/dat9/pkg/vault"
 )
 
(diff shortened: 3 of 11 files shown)


qiffang:
Blocking review:

1. `pkg/vault/schema.go` is not wired into any TiDB tenant schema-init path. `pkg/tenant/schema/tidb_app.go` and `pkg/tenant/schema/tidb_auto.go` still contain zero `vault_*` DDL, and `rg "vault\.SchemaStatements|vault\.InitSchema" pkg` only finds a comment in `pkg/tenant/db9/schema.go`. So this PR fixes dialect in a dead code path; TiDB tenants still will not create the vault tables.

2. Product direction is now explicit: vault should converge on the TiDB/MySQL tenant path. This PR goes the opposite way in `pkg/tenant/db9/schema.go` by preserving a second PG-native vault schema. That keeps the provider contract ambiguous instead of removing the ambiguity that caused the bug.

3. No regression tests were added. The PR changes tenant DDL and runtime SQL but the test plan is still missing the only meaningful check (`Vault tables successfully created on TiDB tenant DB`). `go build` / `go vet` do not validate schema wiring or CRUD correctness on TiDB.

4. `ON CONFLICT ... DO NOTHING` -> `INSERT IGNORE` is not semantics-preserving by default. `INSERT IGNORE` suppresses broader classes of errors than duplicate-key conflicts. Each call site needs explicit justification and coverage, or a narrower TiDB/MySQL equivalent.

This should not merge until the TiDB schema wiring, provider scope, and regression coverage are fixed.

qiffang:
Re-review on commit `16d544b`:

The original schema-wiring blockers are largely fixed, but there is still one **blocking correctness issue** left:

`db9` tenants can still hit the vault API, but this PR removes vault table creation from the `db9` schema-init path.

Code path:
- `pkg/tenant/db9/provisioner.go:39-40` still initializes `db9` tenants via `initDB9Schema(dsn)`
- `pkg/tenant/db9/schema.go:122-125` now explicitly returns only `core`, i.e. no `vault_*` tables are created for `db9` tenants
- `pkg/server/server.go:117-122` and `pkg/server/server.go:290-291` still route vault management endpoints for any tenant-authenticated request
- `pkg/server/vault.go:40-48` builds a vault store from the tenant backend with no provider check or TiDB-only gate

So after this PR, a newly provisioned `db9` tenant can still call `/v1/vault/secrets` / `/v1/vault/tokens` / `/v1/vault/audit`, but the required tables are no longer created on the `db9` init path. That is a deterministic runtime failure path, not just a testing gap.

This needs one of two explicit fixes before merge:
1. gate vault APIs so only TiDB providers can use them, with a clear unsupported response for `db9`, or
2. keep `db9` tenants provisioned with the required vault tables if they are still meant to use vault.

Secondary note: store CRUD integration coverage is still thin, but the provider-path mismatch above is the remaining blocker for me.

qiffang:
Re-review on commit `d66e789`:

The new store integration tests are good and they resolve the runtime-regression-coverage concern:
- `pkg/vault/testmain_test.go` brings up a real MySQL container via `internal/testmysql` and initializes schema with `VaultTiDBSchemaStatements()`
- `pkg/vault/store_test.go` now exercises the full `store.go` surface: DEK round-trip/idempotency, secret CRUD with encrypted BLOB round-trip, cap-token issue/verify/revoke, and audit-log write/query

So I am no longer blocking on missing runtime coverage.

However, my **remaining blocker is unchanged**: the `db9` provider path is still inconsistent. This PR still removes vault table creation from `pkg/tenant/db9/schema.go`, while `db9` tenants can still reach vault endpoints through `pkg/server/server.go` and `pkg/server/vault.go` with no provider gate. That remains a deterministic runtime failure path for `db9` tenants.

Verdict stays: **REVISE / do not merge yet** until the `db9` provider path is handled explicitly.

qiffang:
Re-review on commit `ab0db60`:

I agree that the provider wiring itself is now in place:
- `TenantScope.Provider` is populated in both auth paths
- `db9` tenants no longer fall through to a `table does not exist` failure in vault read paths

So I am no longer blocking on the original provider-path regression.

However, I still have one **remaining API-level blocker** before merge:

The provider guard is surfaced inconsistently across vault endpoints.
- Read API: `pkg/server/vault.go:409-412` returns explicit `501 Not Implemented` for `db9`
- Management APIs: `pkg/server/vault.go:59-62`, `240-243`, and `355-358` still map the `vaultStore()` provider error to `500 Internal Server Error`

That means the same unsupported-provider condition currently produces two different classes of response:
- `/v1/vault/read` -> `501 unsupported`
- `/v1/vault/secrets`, `/v1/vault/tokens`, `/v1/vault/audit` -> `500 internal error`

I do not consider that fully resolved yet. The agreed minimal fix was to make non-TiDB tenants get a clear unsupported response, not just to avoid the missing-table crash.

My verdict remains **REVISE** until management endpoints also map this provider guard to an explicit unsupported response (same family as read API), and ideally a small server-level regression test locks that behavior in.

qiffang:
Final re-review on commit `aa6f517`:

The remaining API-level blocker is resolved. `vaultStore()` now returns a sentinel `errVaultUnsupported`, and all three management handlers (`handleVaultSecrets`, `handleVaultTokens`, `handleVaultAudit`) map that condition to `501 Not Implemented`, matching the existing `handleVaultRead` behavior for unsupported providers.

So from my side, there are no remaining correctness blockers on PR #254. The `db9` missing-table regression is blocked, the provider path is explicit, the TiDB schema wiring is in place, and the MySQL-backed store integration tests cover the converted runtime SQL.

I would still prefer a small server-level regression test for the unsupported-provider response path as follow-up, but I am not blocking merge on that.

Final verdict from my side: **ACCEPT**.

qiffang:
Final re-review on commit `aa6f517`:

The last API-level blocker is resolved. `vaultStore()` now returns a sentinel `errVaultUnsupported`, and all vault management handlers (`handleVaultSecrets`, `handleVaultTokens`, `handleVaultAudit`) map that condition to `501 Not Implemented`, matching `handleVaultRead`.

So the external behavior is now consistent for unsupported providers:
- `/v1/vault/read` on `db9` -> `501`
- `/v1/vault/secrets` on `db9` -> `501`
- `/v1/vault/tokens` on `db9` -> `501`
- `/v1/vault/audit` on `db9` -> `501`

From my side, there are no remaining correctness blockers on PR #254. The TiDB schema wiring, MySQL-backed store tests, provider-path guard, and API behavior are now all aligned.

Final verdict from my side: **ACCEPT**. I would still welcome a small server-level regression test for the provider-unsupported path as follow-up, but I am not blocking merge on it.

qiffang:
Re-review on commit `392cc11`:

This delta is the expected CI fix only: `pkg/vault/testmain_test.go` now adds the blank import for `github.com/go-sql-driver/mysql`, which is required for `sql.Open("mysql", ...)` to register the driver during the vault integration tests. The dependency was already present in `go.mod`; this just fixes the missing test-side wiring.

I do not see any new correctness issue introduced by this change. My previous **ACCEPT** verdict stands. Once CI is green, this is ready to merge from my side.

PR #257, opened 2026-04-16
fix(fuse): route large buffered upload fallback through multipart

### LINKED ISSUE #249: FUSE large-file write can fail because PUT misses X-Dat9-Part-Checksums
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. Start `drive9-server-local` and wait until `/healthz` returns `200`.
2. Provision the default local tenant and obtain `local-dev-key`.
3. Mount FUSE with the `drive9` CLI.
4. Create a target directory via the mount.
5. Write an 8 MiB file through the mount.
6. Immediately stat/read the file via the CLI / remote API.

Representative commands:

```bash
DRIVE9_SERVER=http://127.0.0.1:9009 DRIVE9_API_KEY=local-dev-key ./drive9 mount /tmp/dat9-fuse-smoke
mkdir -p /tmp/dat9-fuse-smoke/fuse-e2e/alpha-renamed
dd if=/dev/zero of=/tmp/dat9-fuse-smoke/fuse-e2e/alpha-renamed/large-8m.bin bs=1M count=8 status=none
DRIVE9_SERVER=http://127.0.0.1:9009 DRIVE9_API_KEY=local-dev-key ./drive9 fs stat /fuse-e2e/alpha-renamed/large-8m.bin
```

This is also reproducible in `e2e/fuse-smoke-test.sh` at the `[10] large-file boundary` step.

An 8 MiB file written through the mounted filesystem should upload successfully and become remotely stat-able / downloadable.

The remote stat failed with `not found`, and the mount/server logs showed the write path failing because the server rejected the upload for a missing checksums header.

Observed symptoms:

```text
fs stat: not found: /fuse-e2e-1776348432/alpha-renamed/large-8m.bin
```

Mount log:

```text
streaming upload part 1 failed for /fuse-e2e-.../alpha-renamed/large-8m.bin: initiate stream upload: upload too large: max 10737418240 bytes
commit queue: upload attempt 1/5 failed for /fuse-e2e-.../alpha-renamed/large-8m.bin: missing X-Dat9-Part-Checksums header
...
commit queue: giving up on /fuse-e2e-.../alpha-renamed/large-8m.bin after 5 retries
commit queue: terminal failure for /fuse-e2e-.../alpha-renamed/large-8m.bin — local data preserved for manual recovery
```

Server log:

```text
event:"write_missing_checksums_header" path:"/fuse-e2e-.../alpha-renamed/large-8m.bin"
method:"PUT" status:400
```

So the large-file FUSE write path appears to fall back into a server PUT flow that requires `X-Dat9-Part-Checksums`, but the client-side path used here does not send that header.

Reproduced with `drive9-server-local` / `drive9` built from:

```text
component: drive9
version: dev
git_hash: aee922e3412bb9e305c36fdb9c797fc1af924fb5
git_branch: codex/fuse-smoke-local-server
go_version: go1.25.1
```

Notes:

- This report is intentionally separate from the overwrite/truncate conflict issue (#247).
- It was observed after skipping the known overwrite-path issue so the smoke test could continue to later FUSE coverage.

found in https://github.com/mem9-ai/drive9/pull/252

JaySon-Huang:
close https://github.com/mem9-ai/drive9/issues/249

## Summary
- route FUSE background write-back / commit-queue large buffered uploads through the multipart conditional writer path
- add regression coverage for the buffered-upload fallback and expected-revision assertions
- keep the larger unknown-size create/truncate execution-model follow-up out of this PR and track it in #262

## Problem
Issue #249 is reproducible against `drive9-server-local` at the FUSE large-file boundary.

After writing an 8 MiB file through the mount, the remote file can fail to appear:

```text
fs stat: not found: /fuse-e2e/.../large-8m.bin
```

The decisive failure is that the background write-back / commit-queue fallback eventually falls back to a legacy direct `PUT` path that does not send `X-Dat9-Part-Checksums`, so the server rejects the upload and the file never materializes remotely.

## Root Cause
The root cause fixed in this PR is:

1. Background write-back / commit-queue uploads still used the legacy direct-`PUT` conditional write path for large buffered files.
   - That path does not send `X-Dat9-Part-Checksums`, so the server rejects the upload.
   - This behavior was introduced by `635b9a7` in https://github.com/mem9-ai/drive9/pull/172

Issue #249 also exposed a separate unknown-size create/truncate streaming problem (`v2_upload_initiate_too_large`), but that follow-up now lives outside this PR as enhancement #262 because the correct fix is larger than this bugfix scope.

## Fix
This PR now keeps only the narrow bugfix and its test coverage.

1. `499d677` routes large buffered background uploads through the multipart conditional writer path, so part checksums are generated and sent correctly.
2. `ccd1319` synchronizes the expected-revision capture in the related FUSE tests so the new assertions are race-safe.

## Validation
- `make build`
- local smoke repro against `~/bench/run-drive9-server-local.sh`
  - baseline (`a7e366e`): remote `fs stat` failed and the server logged `write_missing_checksums_header`
  - current branch after narrowing: remote `fs stat` succeeds and the checksum-header failures disappear

## Scope Note
This PR intentionally does **not** solve the follow-up execution-model problem for unknown-size create/truncate handles. After narrowing the change set, the local smoke repro still shows an early `v2_upload_initiate_too_large` warning before the later background multipart upload succeeds.

That follow-up is now tracked separately in enhancement issue #262:
- https://github.com/mem9-ai/drive9/issues/262

## Summary by CodeRabbit

* **New Features**
  * Added multipart upload support for large file operations, enabling more efficient and reliable handling of large file transfers.

* **Tests**
  * Introduced comprehensive test coverage for multipart upload behavior and file upload scenarios.


pkg/fuse/stream_upload.go
```diff
@@ -61,74 +60,18 @@ func (su *StreamUploader) HasStreamedParts() bool {
...
-		su.mu.Lock()
-		su.streamedParts[partNum] = true
-		su.mu.Unlock()
-
-		if onDone != nil {
-			onDone(partNum)
-		}
-	}()
-
 	return nil
```
JaySon-Huang, line 132: `SubmitPart()` is now a pure no-op, but the create/O_TRUNC paths still wire `OnPartFull` to call it and only evict the completed part from the `onDone` callback. That means large sequential writes no longer release full parts as they are produced and instead keep the whole file resident until `Flush`, which brings back O(file size) memory growth for `cp`/`dd`/`ffmpeg`-style writes. Either disable the current `OnPartFull`/eviction path for deferred-init handles or preserve a real streaming/eviction mechanism here.
Source: process/review-branch | [Second Opinion](https://github.com/PingCAP-QE/second-opinion)

JaySon-Huang:
Incremental review on the new PR 257 commits (`e2dff2b` -> `6ec182e`): no additional correctness regressions found.

`6ec182e` reverts the change that previously made `SubmitPart()` a no-op, so the earlier concern about losing full-part eviction for large sequential FUSE writes is addressed in the current head. The remaining commit (`ccd1319`) is a test-only synchronization cleanup for expected revision capture.

Validated with `go test ./pkg/fuse` and targeted FUSE regression coverage.
Source: process/review-branch | [Second Opinion](https://github.com/PingCAP-QE/second-opinion)

JaySon-Huang:
@coderabbitai review

srstack (approved):

PR #264, opened 2026-04-17
Fix FUSE e2e skip root causes

srstack:
## Summary
This PR fixes the two root causes behind the current FUSE e2e skip/failure paths.

- Resolve directory paths without trailing slashes for backend stat/remove/removeAll/rename operations so FUSE directory ops map to stored directory nodes correctly.
- Refresh inode revision after handle-less truncate-to-zero so a follow-up overwrite does not conflict with its own zero-byte write.
- Add backend and FUSE regression coverage for both behaviors.

## Why
The FUSE smoke flow currently hits two core issues:

- Directory operations from FUSE often use paths without trailing slashes, while backend metadata stores directory nodes with trailing slashes.
- Shell-style overwrite paths can issue a truncate-to-zero before the real content write, and the later write was still using the stale pre-truncate revision.

## Validation
- `go test -v ./pkg/backend -run 'Test(StatDirWithoutTrailingSlash|RemoveDirWithoutTrailingSlash|RemoveAllDirWithoutTrailingSlash|RenameDirWithoutTrailingSlash|RenameDirUpdatesName|RemoveAll)$'`
- `go test -v ./pkg/fuse -run 'Test(SetAttr_TruncateWithoutHandleRefreshesRevision|OpenTruncateWriteThroughShadow)$'`

## Notes
The dev FUSE e2e endpoint will still show the old behavior until this branch is deployed there.

## Summary by CodeRabbit

* **Bug Fixes**
  * Directory operations (Stat, Remove, RemoveAll, Rename) now correctly handle paths without trailing slashes.
  * Server lookups now retry using canonical directory forms so directory requests without a trailing slash succeed.
  * Truncating a file without an open handle refreshes file metadata (revision and mtime) to reflect server state.

* **Tests**
  * Added tests validating directory handling without trailing slashes and revision refresh after truncate.

### DIFF
--- a/pkg/backend/dat9.go
+++ b/pkg/backend/dat9.go
@@ -583,11 +583,12 @@ func (b *Dat9Backend) ReadDirCtx(ctx context.Context, path string) (infos []file
 }
 
 func (b *Dat9Backend) Stat(path string) (*filesystem.FileInfo, error) {
-	resolvedPath, _, err := b.resolveNodePath(backgroundWithTrace(), path)
+	ctx := backgroundWithTrace()
+	resolvedPath, _, err := b.resolveNodePath(ctx, path)
 	if err != nil {
 		return nil, err
 	}
-	nf, err := b.store.Stat(backgroundWithTrace(), resolvedPath)
+	nf, err := b.store.Stat(ctx, resolvedPath)
 	if err != nil {
 		return nil, err
 	}
--- a/pkg/backend/dat9.go
+++ b/pkg/backend/dat9.go
@@ -582,13 +582,33 @@ func (b *Dat9Backend) ReadDirCtx(ctx context.Context, path string) (infos []file
 	return infos, nil
 }
 
-func (b *Dat9Backend) Stat(path string) (*filesystem.FileInfo, error) {
-	ctx := backgroundWithTrace()
-	resolvedPath, _, err := b.resolveNodePath(ctx, path)
-	if err != nil {
+func (b *Dat9Backend) StatNodeCtx(ctx context.Context, path string) (*datastore.NodeWithFile, error) {
+	resolvedPath := normalizePath(path)
+	nf, err := b.store.Stat(ctx, resolvedPath)
+	if err == nil {
+		return nf, nil
+	}
+	if !errors.Is(err, datastore.ErrNotFound) || pathutil.IsDir(path) {
 		return nil, err
 	}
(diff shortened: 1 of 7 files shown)


pkg/backend/dat9.go
```diff
@@ -585,8 +583,11 @@ func (b *Dat9Backend) ReadDirCtx(ctx context.Context, path string) (infos []file
 }
 
 func (b *Dat9Backend) Stat(path string) (*filesystem.FileInfo, error) {
-	path = normalizePath(path)
-	nf, err := b.store.Stat(backgroundWithTrace(), path)
+	resolvedPath, _, err := b.resolveNodePath(backgroundWithTrace(), path)
```
JaySon-Huang: This only fixes slashless directory `Stat` for in-process backend callers. The public HTTP `HEAD` path still bypasses `Dat9Backend.Stat`: `pkg/server/server.go` still calls `b.Store().Stat(r.Context(), path)` directly, so `client.StatCtx("/dir")` continues to return 404 after `Mkdir("/dir")`. As written, the new regression coverage proves the backend helper change, but not the client/server behavior described in the PR. Please either route server-side stat through the same resolution logic or add the same fallback there, and add a server-level regression test for `HEAD /v1/fs/dir`.

Source: process/boundary-check | [Second Opinion](https://github.com/PingCAP-QE/second-opinion)

pkg/server/server.go
```diff
@@ -773,6 +774,12 @@ func (s *Server) handleStat(w http.ResponseWriter, r *http.Request, path string)
 		return
 	}
 	nf, err := b.Store().Stat(r.Context(), path)
+	if err != nil && errors.Is(err, datastore.ErrNotFound) && !pathutil.IsDir(path) {
```
JaySon-Huang: This fixes the external `HEAD /dir` behavior, but it now makes the success path emit an error-level datastore log on every slashless directory stat. The first `b.Store().Stat(r.Context(), path)` still returns `ErrNotFound`, and `observeStoreOp` logs that as `logger.Error("datastore_op_failed", ...)` before this fallback retries with `CanonicalizeDir(path)`. For FUSE-style workloads, that means a valid directory stat now generates noisy error logs and stack traces at high frequency. It would be better to normalize the path before hitting the store, or otherwise keep the fallback outside the store error-logging path.

Source: process/boundary-check | [Second Opinion](https://github.com/PingCAP-QE/second-opinion)

JaySon-Huang (approved):

PR #266, opened 2026-04-17
Fix concurrent v2 presign status transition

### LINKED ISSUE #263: Concurrent v2 PresignPart requests can fail on INITIATED -> UPLOADING transition
## Bug Report

Please answer these questions before submitting your issue. Thanks!

1. Start `drive9-server-local`.
2. Mount drive9 FUSE in strict mode against the local server:
   ```bash
   ./bin/drive9 mount -server http://127.0.0.1:9009 -sync-mode strict /home/ubuntu/drive9_mount_test
   ```
3. Run JuiceFS bench against the mount:
   ```bash
   juicefs bench /home/ubuntu/drive9_mount_test -p 1
   ```
4. Observe the large-file write/read phase for `bigfile.0.0`.

This is reproducible with the current FUSE streaming path, which triggers many concurrent single-part presign requests for the same v2 upload.

The multipart upload should remain active while concurrent `PresignPart` requests arrive for the same upload. All part presigns should succeed, the upload should complete, and the subsequent read phase should be able to open the uploaded file.

One of the concurrent `POST /v2/uploads/<upload_id>/presign` requests returns `409` with:

```text
upload is not in UPLOADING state
```

After that, the FUSE client aborts the whole streaming upload and the file never becomes visible remotely. The later read phase then fails with `ENOENT`, for example:

```text
Failed to open file /home/ubuntu/drive9_mount_test/__juicefs_benchmark_<id>/bigfile.0.0: no such file or directory
```

Relevant server-side signal:
- `backend_presign_part_status_transition_failed`
- `error: upload is not in UPLOADING state`

Root cause:
- `pkg/backend/upload.go:PresignPart()` loads the upload.
- If the status is `INITIATED`, it calls `TransitionUploadStatus(INITIATED -> UPLOADING)`.
- Under concurrent single-part presign requests, multiple requests can observe `INITIATED`.
- Only one request wins the transition. Losers get `ErrUploadNotActive` from the CAS update even though the upload has already become `UPLOADING` due to a peer request.
- That benign race is currently treated as a hard failure instead of re-checking whether the upload is now active.

```text
component: drive9
version: dev
source commit: ac0e7a6239203f22eef112bb2beda98a18e0710e
```

JaySon-Huang:
## Summary

Fix the v2 multipart presign race where concurrent `PresignPart` / `PresignParts` requests can incorrectly fail while one request is moving an upload from `INITIATED` to `UPLOADING`.

Closes #263.

## Problem

Before this change, `PresignPart` and `PresignParts` both expected to win the `INITIATED -> UPLOADING` transition themselves.

With concurrent requests, the first caller could successfully transition the upload to `UPLOADING`, while a second caller racing just behind it would see the transition fail and treat that as a hard error, even though the upload was still active and already in the correct state.

That made concurrent presign traffic flaky and could surface 409-style failures for an otherwise healthy upload.

## Root cause

The backend treated a failed `TransitionUploadStatus(uploadID, INITIATED, UPLOADING)` as a terminal error, instead of re-reading the upload state and accepting the case where another concurrent request had already moved it into `UPLOADING`.

## Fix

- add a backend helper that re-checks the upload after a failed transition attempt
- allow `PresignPart` and `PresignParts` to proceed when the upload is already active in `UPLOADING`
- preserve the existing failure semantics for genuinely inactive / expired / invalid uploads
- add a regression test that exercises the concurrent transition race directly

## Testing

Passed:

```bash
DRIVE9_TEST_MYSQL_DSN='...'
go test ./pkg/backend -run 'TestPresignPart(Single|ConcurrentTransition)|TestPresignPartsBatch' -count=1
```

```bash
DRIVE9_TEST_MYSQL_DSN='...'
go test ./pkg/backend -run 'TestPresignPartConcurrentTransition' -count=10
```

Also checked:
- `go test ./pkg/backend -count=1` still has pre-existing environment/schema failures unrelated to this change

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced reliability of multi-part uploads by improving concurrent request handling and state management during presign operations.
  * Centralized eligibility verification to prevent race conditions and ensure consistent behavior.

* **Tests**
  * Added concurrency tests validating upload stability under simultaneous operations.

### DIFF
--- a/pkg/backend/upload.go
+++ b/pkg/backend/upload.go
@@ -90,6 +90,39 @@ func uploadExpectedRevision(upload *datastore.Upload) int64 {
 	return *upload.ExpectedRevision
 }
 
+func (b *Dat9Backend) ensureUploadPresignable(ctx context.Context, uploadID string, upload *datastore.Upload) (*datastore.Upload, error) {
+	if upload.Status != datastore.UploadUploading && upload.Status != datastore.UploadInitiated {
+		return nil, datastore.ErrUploadNotActive
+	}
+	if time.Now().After(upload.ExpiresAt) {
+		_ = b.AbortUploadV2(ctx, uploadID)
+		return nil, datastore.ErrUploadExpired
+	}
+	if upload.Status != datastore.UploadInitiated {
+		return upload, nil
+	}
+	if err := b.store.TransitionUploadStatus(ctx, uploadID, datastore.UploadInitiated, datastore.UploadUploading); err == nil {
+		upload.Status = datastore.UploadUploading
+		return upload, nil
+	} else if !errors.Is(err, datastore.ErrUploadNotActive) {
+		return nil, err
+	}
+
+	// Another concurrent presign may have already advanced INITIATED -> UPLOADING.
+	refreshed, err := b.store.GetUpload(ctx, uploadID)
+	if err != nil {
+		return nil, err
+	}
+	if refreshed.Status != datastore.UploadUploading {
+		if time.Now().After(refreshed.ExpiresAt) {
+			_ = b.AbortUploadV2(ctx, uploadID)
+			return nil, datastore.ErrUploadExpired
+		}
+		return nil, datastore.ErrUploadNotActive
+	}
+	return refreshed, nil
+}
+
 func (b *Dat9Backend) activeUploadByPath(ctx context.Context, path string) (*datastore.Upload, error) {
 	upload, err := lookupActiveUploadByPath(b.store, ctx, path)
 	if err != nil {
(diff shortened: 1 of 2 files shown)


JaySon-Huang:
@coderabbitai review

srstack:
I found 2 issues that should be addressed.

1. High severity: pkg/backend/upload.go
In the new ensureUploadPresignable helper, after reloading the upload record, the code immediately accepts the case where the status is already UPLOADING, but it does not re-check refreshed.ExpiresAt first. That creates a new race window: request A can transition INITIATED -> UPLOADING just before expiry, request B can lose the transition race, reload the row, and still be allowed to presign even though the upload has already expired. In other words, the losing concurrent request can bypass the expiry check. The fix is to validate expiry on the refreshed upload before accepting UPLOADING.

2. Medium severity: pkg/backend/upload.go
PresignParts now calls ensureUploadPresignable before validating the batch payload. That means an invalid first batch request can still advance the upload from INITIATED to UPLOADING. Previously, the code validated batch size and duplicate part numbers before performing the status transition, so malformed requests were side-effect free. This is a behavior regression: it makes state transitions and metrics less accurate, and it may affect any downstream logic that treats UPLOADING as meaning a valid presign has started. The fix is to move batch validation back before the status transition, or split the helper into separate liveness/expiry check and perform transition steps.

I did not see any other obvious problems in the patch. CI is green, but the new tests only cover concurrent PresignPart calls; they do not cover concurrent PresignParts calls or mixed PresignPart/PresignParts races, so there is still a test gap there.

qiffang (approved):

PR #269, opened 2026-04-17
Fix FUSE overwrite after truncate

srstack:
## Summary
Fix the FUSE overwrite path so a shell-style truncate followed by overwrite does not self-conflict on a stale base revision.

## What changed
- refresh matching open file handles after a path-level truncate updates the remote revision
- add a regression test covering the `O_TRUNC open -> path truncate -> overwrite` sequence
- restore the previously skipped overwrite, append, truncate, and attribute checks in the FUSE smoke test
- make the smoke assertions wait for remote visibility where needed and isolate the attribute test file from later rename checks

## Validation
- `go test -v ./pkg/fuse -run 'TestSetAttr_(TruncateWithoutHandleRefreshesRevision|PathTruncateRefreshesOpenHandleBaseRevision)'`
- `DRIVE9_BASE=http://k8s-dat9-dat9serv-d5e02e7d07-1645488597.ap-southeast-1.elb.amazonaws.com bash e2e/fuse-smoke-test.sh`
- dev FUSE smoke result: `47/47 passed, 0 failed`

## Notes
- Left unrelated local perf/docs drafts out of this PR on purpose.

## Summary by CodeRabbit

* **New Features**
  * Configurable large-file visibility monitoring (timeout and interval).
  * More robust remote verification after overwrites, appends, truncates and renames to ensure visible content matches expectations.

* **Bug Fixes**
  * Improved truncate and write handling so local truncates and subsequent writes stay consistent with remote state and avoid stale-write conflicts.

* **Tests**
  * Added end-to-end and unit tests covering truncation, open-handle behavior, and remote visibility.

### DIFF
--- a/pkg/fuse/dat9fs.go
+++ b/pkg/fuse/dat9fs.go
@@ -185,6 +185,26 @@ func (fs *Dat9FS) clearDirtySize(ino uint64, seq uint64) {
 	}
 }
 
+func (fs *Dat9FS) updateOpenHandleBaseRevision(path string, revision int64) {
+	if revision <= 0 {
+		return
+	}
+
+	fs.fileHandles.ForEach(func(_ uint64, fh *FileHandle) {
+		if fh == nil || fh.Path != path {
+			return
+		}
+		fh.Lock()
+		fh.BaseRev = revision
+		if fh.ShadowReady && fs.shadowStore != nil {
+			if err := fs.shadowStore.Ensure(fh.Path, fh.Dirty.Size(), revision); err != nil {
+				log.Printf("shadow base revision refresh failed for %s: %v", fh.Path, err)
+			}
+		}
+		fh.Unlock()
+	})
+}
+
 func (fs *Dat9FS) preloadWritableHandle(ctx context.Context, fh *FileHandle) gofuse.Status {
 	stat, err := fs.client.StatCtx(ctx, fh.Path)
 	if err != nil {
@@ -691,6 +711,7 @@ func (fs *Dat9FS) SetAttr(cancel <-chan struct{}, input *gofuse.SetAttrIn, out *
 					if stat.Revision > 0 {
 						entry.Revision = stat.Revision
 						fs.inodes.UpdateRevision(input.NodeId, stat.Revision)
+						fs.updateOpenHandleBaseRevision(entry.Path, stat.Revision)
 					}
 					if !stat.Mtime.IsZero() {
 						entry.Mtime = stat.Mtime
--- a/pkg/fuse/dat9fs.go
+++ b/pkg/fuse/dat9fs.go
@@ -185,16 +185,31 @@ func (fs *Dat9FS) clearDirtySize(ino uint64, seq uint64) {
 	}
 }
 
+func shouldRefreshHandleAfterPathTruncate(fh *FileHandle) bool {
+	if fh == nil || fh.Dirty == nil {
+		return false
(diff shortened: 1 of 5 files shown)


pkg/fuse/dat9fs.go
```diff
@@ -185,6 +185,33 @@ func (fs *Dat9FS) clearDirtySize(ino uint64, seq uint64) {
...
+	if revision <= 0 {
+		return
+	}
+
+	for _, fh := range fs.fileHandles.Snapshot() {
+		if fh == nil || fh.Path != path {
+			continue
+		}
+		fh.Lock()
+		fh.BaseRev = revision
```
JaySon-Huang, line 260: Updating `BaseRev` for every open handle on the same path turns concurrent stale writers into false "current" writers. If process A opened `/file` at revision 1 and still holds old/dirty state, then process B runs `: > file` and bumps the remote object to revision 2, this helper rewrites A's handle to `BaseRev=2`. When A later flushes through `WriteCtxConditional` / `PatchFile` / `WriteStreamConditional`, it now uses `expectedRevision=2` and can silently overwrite the truncate/new data instead of surfacing the conflict that should have protected it. Please limit this refresh to the truncating handle (or another handle that has definitely adopted the post-truncate empty baseline) and add a two-handle regression test for the stale-writer case.

Source: rule/JAYSONHUANG-RC-001 | [Second Opinion](https://github.com/PingCAP-QE/second-opinion)

pkg/fuse/dat9fs.go
```diff
@@ -185,6 +185,48 @@ func (fs *Dat9FS) clearDirtySize(ino uint64, seq uint64) {
...
+	var matching []*FileHandle
+	for _, fh := range fs.fileHandles.Snapshot() {
+		if fh != nil && fh.Path == path && fh.Dirty != nil {
+			matching = append(matching, fh)
+		}
+	}
+
+	for _, fh := range matching {
+		fh.Lock()
+		if len(matching) > 1 && !shouldRefreshHandleAfterPathTruncate(fh) {
```
JaySon-Huang: This still leaves the original stale-writer bug in the single-handle case. The new filter only applies when `len(matching) > 1`, so if a path-truncate happens while there is exactly one writable stale handle on the path, that handle is still refreshed unconditionally even though it never adopted the new empty-file baseline. After that, its later `WriteCtxConditional` / `PatchFile` / `WriteStreamConditional` uses the post-truncate revision and can bypass the conflict that should have protected the truncate. Please gate the refresh on the handle state itself (for example `ZeroBase`) rather than on how many writable handles happen to be open, and add a regression test for the single stale-writer + path-truncate case.

Source: rule/JAYSONHUANG-RC-001 | [Second Opinion](https://github.com/PingCAP-QE/second-opinion)

JaySon-Huang:
Incremental review for `afc7e32..14832cc`: no new findings in the delta. The previous high-severity stale-writer concern looks addressed by the new same-process adoption logic: the writable handle now carries `OpenPID`, the single-handle path-truncate case only adopts the zero base for the same caller, and the added regression coverage now exercises both the same-caller success path and the single stale-writer conflict path. I also reran `go test ./pkg/fuse` plus the targeted truncate-refresh tests, and they passed.

Source: process/review-branch | [Second Opinion](https://github.com/PingCAP-QE/second-opinion)

JaySon-Huang (approved):

PR #298, opened 2026-04-20
fix(fuse): refresh inode revision after flush and improve error status mapping

srstack:
## Summary

Fixes the repeated `flush failed (status 5)` issue during FUSE overwrite/append workloads.

## Root Cause

`flushHandle` did not update the inode cache after a successful upload, causing subsequent overwrite/append operations to use a stale `BaseRev`. This triggered a 412 Precondition Failed from the server, which `httpToFuseStatus` did not recognize, falling through to generic `EIO` (status 5).

## Changes

- **`refreshRevisionAfterFlush()`**: Refresh inode revision and open FileHandle state after all three upload success paths (FinishStreaming, UploadAll, direct PUT/Patch).
- **`httpToFuseStatus()`**: Use `errors.As` to recognize `*client.StatusError` and map HTTP status codes precisely:
  - 412 → `ESTALE`
  - 500/502/503 → `EAGAIN`
  - 400 → `EINVAL`
  - 409 → `EEXIST` (only when message contains \"already exists\")
- **Debug logging**: Added error logging on flush failure paths.

## Testing

- `make test TEST_PKGS='./pkg/fuse/...'` ✅ all passed
- `make lint` ✅ 0 issues

Closes FUSE overwrite/append flush failure issue.

## Summary by CodeRabbit

* **Bug Fixes**
  * More accurate mapping of HTTP errors to filesystem errors for clearer error responses (e.g., not-found, conflict, permission, size, stale, invalid, retryable, and server errors).

* **Improvements**
  * More reliable file revision updates after flush/commit so revisions reflect CAS-committed state.
  * Better diagnostic logging for directory listings and flush/upload failures.
  * Streaming uploader now advances and resets expected revision after successful commits.

### DIFF
--- a/pkg/fuse/dat9fs.go
+++ b/pkg/fuse/dat9fs.go
@@ -512,15 +512,24 @@ func httpToFuseStatus(err error) gofuse.Status {
 
 	// Fallback to string matching for non-StatusError errors.
 	msg := err.Error()
+	lowerMsg := strings.ToLower(msg)
 	switch {
-	case strings.Contains(msg, "not found") || strings.Contains(msg, "HTTP 404"):
+	case strings.Contains(lowerMsg, "not found") || strings.Contains(msg, "HTTP 404"):
 		return gofuse.ENOENT
-	case strings.Contains(msg, "HTTP 409") || strings.Contains(msg, "already exists"):
+	case strings.Contains(lowerMsg, "already exists"):
 		return gofuse.Status(syscall.EEXIST)
 	case strings.Contains(msg, "HTTP 403"):
 		return gofuse.EACCES
 	case strings.Contains(msg, "HTTP 413"):
 		return gofuse.Status(syscall.EFBIG)
+	case strings.Contains(msg, "HTTP 412"):
+		return gofuse.Status(syscall.ESTALE)
+	case strings.Contains(msg, "HTTP 400"):
+		return gofuse.Status(syscall.EINVAL)
+	case strings.Contains(msg, "HTTP 500") ||
+		strings.Contains(msg, "HTTP 502") ||
+		strings.Contains(msg, "HTTP 503"):
+		return gofuse.Status(syscall.EAGAIN)
 	default:
 		return gofuse.EIO
 	}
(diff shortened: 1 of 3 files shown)


srstack:
Thanks for the reviews, @coderabbitai and @copilot-pull-request-reviewer. I've addressed both points in the latest commit (5216d53).

### Changes made

1. **Aligned fallback error-to-errno mapping** (`httpToFuseStatus` string-matching branch):
   - Added `HTTP 412` → `ESTALE`, `HTTP 400` → `EINVAL`, and `HTTP 500/502/503` → `EAGAIN` to the fallback path.
   - Changed the `HTTP 409` check to only map to `EEXIST` when the message contains `\"already exists\"`, matching the typed `StatusError` branch behavior.

2. **Updated `refreshRevisionAfterFlush` to also refresh the open FileHandle**:
   - After successfully querying `StatCtx`, the function now updates both the inode cache (`fs.inodes.UpdateRevision`) **and** the open handle state (`fh.BaseRev = stat.Revision`, `fh.IsNew = false`).
   - This ensures that subsequent writes on the *same* open handle use the correct base revision for CAS uploads, not just new opens.

### Verification

- `make test TEST_PKGS='./pkg/fuse/...'` — all tests pass ✅
- `make lint` — 0 issues ✅

Please take another look when you have a moment.

pkg/fuse/dat9fs.go
```diff
@@ -2091,6 +2146,7 @@ func (fs *Dat9FS) flushHandle(ctx context.Context, fh *FileHandle) gofuse.Status
 		fs.notifyInode(fh.Ino)
 		parentIno, _ := fs.inodes.GetInode(parentDir(fh.Path))
 		fs.notifyInode(parentIno)
+		fs.refreshRevisionAfterFlush(ctx, fh)
```
JaySon-Huang: After `refreshRevisionAfterFlush()` updates only `fh.BaseRev`, the live `fh.Streamer` still keeps the old conditional revision. The next `UploadAll`/`FinishStreaming` on the same large-file handle will reuse that stale CAS value, so repeated multipart flushes can still hit the same conflict this PR is trying to eliminate. Refresh or recreate the uploader after a successful flush.

Source: rule/JAYSONHUANG-RC-002 | [Second Opinion](https://github.com/PingCAP-QE/second-opinion)

pkg/fuse/dat9fs.go
```diff
@@ -2026,6 +2069,17 @@ func (fs *Dat9FS) flushHandleDebounced(ctx context.Context, fh *FileHandle, forc
 	return gofuse.OK
 }
 
+// refreshRevisionAfterFlush updates the inode cache and the open FileHandle
+// with the latest server revision so that subsequent writes use the correct
+// base revision for CAS uploads. Callers must hold fh.mu.
+func (fs *Dat9FS) refreshRevisionAfterFlush(ctx context.Context, fh *FileHandle) {
```
JaySon-Huang: `refreshRevisionAfterFlush()` learns the new revision via a second `StatCtx()` call. If another writer commits between the successful flush and that HEAD request, this handle can advance to a revision it did not produce and later overwrite the other writer's change without seeing a conflict. The committed revision needs to come from the successful write/complete path itself, not from a later best-effort HEAD.

Source: policy/baseline-high-severity | [Second Opinion](https://github.com/PingCAP-QE/second-opinion)

JaySon-Huang:
Incremental second-opinion review for commit `0795b3c3ca9babd0a5c4bf78d12fcf72282c350e` relative to previously reviewed baseline `dae03f8b7616918d155fed67c16886461c170cfa` found no new issues.

The update addresses both previously reported P1 risks:
- it no longer learns the post-flush revision via a follow-up `StatCtx()` call, and instead finalizes handle state from the committed CAS revision directly;
- it resets the live `StreamUploader` to the new expected revision after a successful flush, so the same open handle does not reuse a stale multipart CAS baseline.

Verification run:
- `make test TEST_PKGS=./pkg/fuse/...

JaySon-Huang (approved):

PR #311, opened 2026-04-21
[WIP] fuse: make server metadata authoritative for create visibility

### LINKED ISSUE #309: Feature request: make server metadata authoritative for drive9 FUSE namespace semantics
## Feature Request

**Is your feature request related to a problem? Please describe:**

Yes. The current drive9 FUSE design can expose a namespace authority gap around file creation and later directory visibility.

In practice, a FUSE `Create()` may succeed locally before the corresponding path is fully represented in the server-side metadata namespace. A fast follow-up `Lookup/GetAttr/Readdir/Open` can then fall back to remote metadata queries and observe a different answer from the local mount state. This can surface as `ENOENT`/`EIO`-style behavior under benchmark or concurrent access.

The same ambiguity also affects the user model for directory visibility:

- users need a clear guarantee that `Create/Mkdir/Rename/Unlink` are committed in server metadata before success is returned
- users also need a separate, explicit contract for why an already-open directory handle may continue to return stale `readdir` results
- these two boundaries should not be mixed together

**Describe the feature you'd like:**

I would like drive9 FUSE to adopt a stable server-metadata-first contract for namespace operations.

Requested contract:

1. `server metadata` is the only namespace authority.
2. `Create/Mkdir/Rename/Unlink` must not return success until the namespace mutation has been committed in server metadata.
3. FUSE `Create()` should use an explicit `metadata-only create` path.
4. Brand-new create should immediately produce a normal zero-byte file, not a special `PENDING_CREATE` object.
5. The created file should be immediately visible to other clients as:
   - existing path
   - `size = 0`
   - empty file on read
6. Brand-new create should allocate `revision = 1`; the first successful content commit should advance it to `revision = 2`.
7. Existing PUT/create clients may remain temporarily compatible, but the server should converge internally to one unified create state machine instead of keeping long-term dual semantics.
8. `readdir` should not use local overlay to patch directory entries.
9. An already-open directory handle may continue to return an older snapshot after later `create/mkdir/rename/unlink`; this should be treated as a cache-boundary contract, not as a metadata commit failure.
10. `readdir` cache behavior should explicitly distinguish:
    - per-handle snapshot boundary
    - cross-handle directory cache boundary

In short: namespace mutations should be metadata-first and authoritative, while `readdir` freshness should be a separately documented cache behavior.

**Describe alternatives you've considered:**

1. Patch only local `PendingIndex` behavior so that newly created files become visible earlier inside the same mount.
   - This can reduce one local race, but it does not establish a system-wide namespace authority.
   - It still leaves server metadata and client-local overlay as competing truths.

2. Keep the current model and just flush/commit earlier.
   - This couples create latency to content upload behavior.
   - It does not cleanly separate namespace existence from content durability.

3. Keep an explicit `PENDING_CREATE` state in server metadata.
   - After comparing with 3FS and JuiceFS behavior, this appears more complex than necessary for the chosen user-facing semantics.
   - Since the desired behavior is already "normal zero-byte file immediately visible", a special pending-create object adds lifecycle and cleanup complexity without clear product benefit.

The preferred direction is:

- FUSE must use metadata-only create
- old clients may remain temporarily compatible
- server internals should converge to one create state machine
- do not keep long-term dual create semantics
- do not model brand-new create as explicit `PENDING_CREATE`

**Teachability, Documentation, Adoption, Migration Strategy:**

This feature would benefit from a documented contract similar to the following:

- namespace authority:
  - server metadata is the truth for path existence
- mutation commit boundary:
  - `Create/Mkdir/Rename/Unlink` success means server metadata is already committed
- brand-new file semantics:
  - `create` yields a normal zero-byte file
  - `revision = 1` at create
  - first committed content advances revision to `2`
- `readdir` semantics:
  - no local overlay patching of directory entry sets
  - old directory handles may continue to return old snapshots
  - this is a cache rule, not a commit rule

Suggested migration shape:

- Phase 1: add explicit `metadata-only create` for FUSE
- Phase 2: make FUSE namespace operations metadata-first
- Phase 3: internally converge old PUT/create paths onto the same create state machine
- Phase 4: document `readdir` cache boundaries and add regression tests for:
  - `Create -> immediate Lookup/GetAttr/Open/Readdir`
  - old directory handle still reading old snapshot after later namespace changes
  - new handle seeing updated namespace according to cache boundaries

The main value of this request is to turn those contracts into an upstream implementation and review target.

JaySon-Huang:
close https://github.com/mem9-ai/drive9/issues/309

## Summary
- add a transactional metadata-only create path plus server/client `?create` API so brand-new files become authoritative empty-file entities at revision 1 before content upload
- introduce an explicit `create` change event and migrate FUSE create, lookup, getattr, and readdir toward server-metadata-first authority while keeping local dirty state as mount-local overlay only
- lock the create-first revision transitions with backend/FUSE regressions and fix follow-up lint and metadata-first test stub issues

## Validation
- make lint
- make test TEST_PKGS='./pkg/datastore' TEST_RUN='Test(CreateConfirmedEmptyFileTx|InsertFileTx)'
- make test TEST_PKGS='./pkg/backend' TEST_RUN='TestCreateMetadataOnlyCtx|TestCreateAndStat'
- make test TEST_PKGS='./pkg/server' TEST_RUN='TestCreateMetadataOnly'
- make test TEST_PKGS='./pkg/client' TEST_RUN='TestCreateMetadataOnly'
- make test TEST_PKGS='./pkg/server' TEST_RUN='TestSSE(Create|EndpointLiveEvent|EndpointReplay|StructuralOp)'
- make test TEST_PKGS='./pkg/fuse' TEST_RUN='TestSSEWatcher(HandleCreate|HandleChange|SelfFilter|HandleEventReset|Heartbeat|ResetPreservesInodes)'
- make test TEST_PKGS='./pkg/backend' TEST_RUN='Test(CreateMetadataOnlyCtxThenConditionalWriteAdvancesRevision|CreateMetadataOnlyThenConfirmUploadAdvancesRevision)'
- make test TEST_PKGS='./pkg/fuse' TEST_RUN='Test(CreateUsesMetadataAuthorityResult|RenameMetadataCreatedFileUsesRemoteRename|UnlinkMetadataCreatedFileUsesRemoteDelete|MetadataCreatedWriteBackUploadUsesRevisionOne|CreateWriteFlushRelease_SmallFile|ParallelCreateAndGetAttr)'
- make test TEST_PKGS='./pkg/fuse' TEST_RUN='Test(LookupFallsBackToParentListWhenDirStatUnsupported|LookupIgnoresPendingWriteBackWithoutServerMetadata|LookupUsesServerMetadataBeforePendingWriteBack|GetAttrPrefersDirtyHandleSize|GetAttrUsesLatestDirtyHandleSize|GetAttrUsesServerBaseThenWriteBackOverlay|GetAttrDirectoryDoesNotRequireRemoteStat|CreateUsesMetadataAuthorityResult)'
- make test TEST_PKGS='./pkg/fuse' TEST_RUN='Test(ReadDirDoesNotMergePendingEntries|ReadDirHandleKeepsSnapshotUntilReopen|CreateUsesMetadataAuthorityResult|LookupUsesServerMetadataBeforePendingWriteBack|GetAttrUsesServerBaseThenWriteBackOverlay|MetadataCreatedWriteBackUploadUsesRevisionOne)'
- make test TEST_PKGS='./pkg/backend ./pkg/server ./pkg/client ./pkg/fuse' TEST_RUN='Test(CreateMetadataOnly|CreateMetadataOnlyCtx|CreateMetadataOnlyCtxThenConditionalWriteAdvancesRevision|CreateMetadataOnlyThenConfirmUploadAdvancesRevision|SSECreate|CreateUsesMetadataAuthorityResult|RenameMetadataCreatedFileUsesRemoteRename|UnlinkMetadataCreatedFileUsesRemoteDelete|MetadataCreatedWriteBackUploadUsesRevisionOne|LookupIgnoresPendingWriteBackWithoutServerMetadata|LookupUsesServerMetadataBeforePendingWriteBack|GetAttrUsesServerBaseThenWriteBackOverlay|ReadDirDoesNotMergePendingEntries|ReadDirHandleKeepsSnapshotUntilReopen)'
- make test TEST_P=1

## Summary by CodeRabbit

* **New Features**
  * Metadata-only file creation: create empty file placeholders without uploading content immediately.
  * Client API to issue metadata-only creates and receive canonical create metadata.
  * "create" operation added to change event streams and SSE events.

* **Improvements**
  * FUSE: rely on server metadata as authoritative, improve revision/commit ordering and overlay consistency.
  * Commit/upload callbacks improve revision propagation and reliability.

* **Tests & Diagnostics**
  * Expanded tests for metadata-create flows and improved FUSE smoke-test diagnostics.


JaySon-Huang:
@coderabbitai pause

PR #315, opened 2026-04-22
Add automatic schema diff and safe fix for tenant and meta stores

srstack:
## Summary
Add init-SQL-driven schema spec parsing and safe diff/fix flows for both tenant TiDB schemas and the meta control-plane schema.

## What changed
- Tenant schema (`pkg/tenant/schema/tidb_auto.go`):
  - Refactored validation to generate structured schema diffs.
  - Added safe repair planning/execution for missing tables, columns, and indexes.
  - Switched schema spec construction to parse init SQL statements automatically.
  - Kept mode-specific `files.embedding` contract validation for auto/app modes.
- Meta schema (`pkg/meta/meta.go`):
  - Refactored `migrate()` to run parse -> diff -> safe fix -> re-diff.
  - Added init-SQL parsing for table/column/index specs in control-plane schema.
  - Added mismatch error reporting when residual non-repairable drift remains.
- Tests:
  - Extended tenant schema tests for parsed spec coverage and diff/repair behavior.
  - Added meta schema parser/diff tests.

## Validation
- `go test ./pkg/meta ./pkg/tenant/schema`

## Summary by CodeRabbit

* **Bug Fixes**
  * More reliable, deterministic schema diffing that detects missing/incorrect primary keys, columns and indexes; better handling/ignoring of idempotent DB errors.

* **New Features**
  * Schema validation/ensure is context-aware and skips up-to-date tenants; system records tenant schema version after successful ensures.
  * Planned repairs apply only whitelisted safe changes.

* **Refactor**
  * Unified diff-and-repair migration pipeline with iterative revalidation.

* **Tests**
  * Expanded coverage for schema parsing, diffs, repair planning and safety filtering.

### DIFF
--- a/internal/schemaspec/schemaspec.go
+++ b/internal/schemaspec/schemaspec.go
@@ -1,0 +1,330 @@
+// Package schemaspec provides shared helpers for parsing trusted MySQL/TiDB
+// schema statements and classifying safe repair operations.
+package schemaspec
+
+import (
+	"errors"
+	"fmt"
+	"hash/crc32"
+	"strings"
+
+	mysql "github.com/go-sql-driver/mysql"
+)
+
+// NormalizeSQLFragment lowercases SQL, removes identifier quoting and TiDB's
+// _utf8mb4 marker, and collapses repeated whitespace for stable comparisons.
+func NormalizeSQLFragment(s string) string {
+	s = strings.ToLower(s)
+	s = strings.ReplaceAll(s, "`", "")
+	s = strings.ReplaceAll(s, "_utf8mb4", "")
+	fields := strings.Fields(s)
+	return strings.Join(fields, " ")
+}
+
+// CanonicalStatementForHash normalizes formatting-only differences that remain
+// after whitespace collapsing so schema-version hashing is stable across
+// cosmetic edits.
+func CanonicalStatementForHash(s string) string {
+	normalized := NormalizeSQLFragment(s)
+	for _, replacement := range [][2]string{
+		{" (", "("},
+		{"( ", "("},
+		{" )", ")"},
+		{") ", ")"},
+		{", ", ","},
+		{" ,", ","},
+	} {
+		normalized = strings.ReplaceAll(normalized, replacement[0], replacement[1])
+	}
+	return normalized
+}
+
+// SQLSnippet truncates a statement for logging and error messages.
+func SQLSnippet(stmt string) string {
+	snippet := stmt
+	if len(snippet) > 80 {
+		snippet = snippet[:80]
+	}
+	return snippet
+}
+
+// ParseCreateTableStatement extracts the table name and the top-level column /
+// constraint definition list from trusted CREATE TABLE SQL or SHOW CREATE TABLE
+// output. The scanner is intentionally lightweight and is not intended for
+// arbitrary user-supplied SQL.
+func ParseCreateTableStatement(stmt string) (tableName string, definitions string, ok bool, err error) {
+	lower := strings.ToLower(stmt)
+	prefixes := []string{"create table if not exists", "create table"}
+	start := -1
+	prefixLen := 0
+	for _, prefix := range prefixes {
+		if idx := strings.Index(lower, prefix); idx >= 0 {
+			start = idx
+			prefixLen = len(prefix)
+			break
+		}
+	}
+	if start < 0 {
+		return "", "", false, nil
+	}
+	i := start + prefixLen
+	for i < len(stmt) && (stmt[i] == ' ' || stmt[i] == '\n' || stmt[i] == '\t' || stmt[i] == '\r') {
+		i++
+	}
+	if i >= len(stmt) {
+		return "", "", false, fmt.Errorf("parse create table: missing table name")
+	}
+	nameStart := i
+	if stmt[i] == '`' {
+		i++
+		for i < len(stmt) && stmt[i] != '`' {
+			i++
+		}
+		if i >= len(stmt) {
+			return "", "", false, fmt.Errorf("parse create table: unterminated quoted table name")
+		}
+		tableName = strings.ToLower(stmt[nameStart+1 : i])
+		i++
+	} else {
+		for i < len(stmt) && stmt[i] != ' ' && stmt[i] != '\n' && stmt[i] != '\t' && stmt[i] != '(' {
+			i++
+		}
+		tableName = strings.ToLower(strings.TrimSpace(stmt[nameStart:i]))
+	}
+	for i < len(stmt) && stmt[i] != '(' {
+		i++
+	}
+	if i >= len(stmt) || stmt[i] != '(' {
+		return "", "", false, fmt.Errorf("parse create table %s: missing opening parenthesis", tableName)
+	}
+	bodyStart := i + 1
+	depth := 1
+	inSingle := false
+	inDouble := false
+	inBacktick := false
+	for i = bodyStart; i < len(stmt); i++ {
+		ch := stmt[i]
+		switch ch {
+		case '\\':
+			i++
+			continue
+		case '\'':
+			if !inDouble && !inBacktick {
+				inSingle = !inSingle
+			}
+		case '"':
+			if !inSingle && !inBacktick {
+				inDouble = !inDouble
+			}
+		case '`':
+			if !inSingle && !inDouble {
+				inBacktick = !inBacktick
+			}
+		case '(':
+			if !inSingle && !inDouble && !inBacktick {
+				depth++
+			}
+		case ')':
+			if !inSingle && !inDouble && !inBacktick {
+				depth--
+				if depth == 0 {
+					return tableName, stmt[bodyStart:i], true, nil
+				}
+			}
+		}
+	}
+	return "", "", false, fmt.Errorf("parse create table %s: unbalanced parentheses", tableName)
+}
+
+// SplitTopLevelComma splits a comma-delimited definition list while ignoring
+// commas inside nested parentheses or quoted strings / identifiers.
+func SplitTopLevelComma(definitions string) []string {
+	parts := make([]string, 0)
+	start := 0
+	depth := 0
+	inSingle := false
+	inDouble := false
+	inBacktick := false
+	for i := 0; i < len(definitions); i++ {
+		ch := definitions[i]
+		switch ch {
+		case '\\':
+			i++
+			continue
+		case '\'':
+			if !inDouble && !inBacktick {
+				inSingle = !inSingle
+			}
+		case '"':
+			if !inSingle && !inBacktick {
+				inDouble = !inDouble
+			}
+		case '`':
+			if !inSingle && !inDouble {
+				inBacktick = !inBacktick
+			}
+		case '(':
+			if !inSingle && !inDouble && !inBacktick {
+				depth++
+			}
+		case ')':
+			if !inSingle && !inDouble && !inBacktick && depth > 0 {
+				depth--
+			}
+		case ',':
+			if !inSingle && !inDouble && !inBacktick && depth == 0 {
+				parts = append(parts, strings.TrimSpace(definitions[start:i]))
+				start = i + 1
+			}
+		}
+	}
+	if start < len(definitions) {
+		parts = append(parts, strings.TrimSpace(definitions[start:]))
+	}
+	return parts
+}
+
+// SplitIdentifierAndRest extracts the first bare or backtick-quoted identifier
+// from a trusted schema fragment and returns the remaining suffix.
+func SplitIdentifierAndRest(s string) (identifier string, rest string) {
+	s = strings.TrimSpace(s)
+	if s == "" {
+		return "", ""
+	}
+	if s[0] == '`' {
+		end := strings.Index(s[1:], "`")
+		if end < 0 {
+			return "", ""
+		}
+		id := s[1 : 1+end]
+		return id, strings.TrimSpace(s[1+end+1:])
+	}
+	for i := 0; i < len(s); i++ {
+		if s[i] == ' ' || s[i] == '\t' || s[i] == '\n' || s[i] == '\r' {
+			return s[:i], strings.TrimSpace(s[i+1:])
+		}
+	}
+	return s, ""
+}
+
+// ParseColumnType returns the column type prefix from a trusted column
+// definition, stopping before common trailing clauses such as NULL / DEFAULT /
+// COLLATE / CHECK / STORAGE.
+func ParseColumnType(rest string) string {
+	rest = strings.TrimSpace(rest)
+	if rest == "" {
+		return ""
+	}
+	keywords := []string{
+		" not ",
+		" null",
+		" default ",
+		" generated ",
+		" as ",
+		" primary ",
+		" unique ",
+		" comment ",
+		" references ",
+		" auto_increment",
+		" on update",
+		" character set ",
+		" collate ",
+		" check ",
+		" storage ",
+		" column_format ",
+	}
+	depth := 0
+	inSingle := false
+	inDouble := false
+	inBacktick := false
+	for i := 0; i < len(rest); i++ {
+		ch := rest[i]
+		switch ch {
+		case '\\':
+			i++
+			continue
+		case '\'':
+			if !inDouble && !inBacktick {
+				inSingle = !inSingle
+			}
+		case '"':
+			if !inSingle && !inBacktick {
+				inDouble = !inDouble
+			}
+		case '`':
+			if !inSingle && !inDouble {
+				inBacktick = !inBacktick
+			}
+		case '(':
+			if !inSingle && !inDouble && !inBacktick {
+				depth++
+			}
+		case ')':
+			if !inSingle && !inDouble && !inBacktick && depth > 0 {
+				depth--
+			}
+		}
+		if inSingle || inDouble || inBacktick || depth > 0 {
+			continue
+		}
+		suffix := " " + strings.ToLower(rest[i:])
+		for _, kw := range keywords {
+			if strings.HasPrefix(suffix, kw) {
+				return strings.TrimSpace(rest[:i])
+			}
+		}
+	}
+	return strings.TrimSpace(rest)
+}
+
+// IsSafeAddColumnRepairSQL identifies ALTER TABLE ... ADD COLUMN statements that
+// are safe to auto-apply to live schemas.
+func IsSafeAddColumnRepairSQL(sqlText string) bool {
+	n := NormalizeSQLFragment(sqlText)
+	if !strings.HasPrefix(n, "alter table ") || !strings.Contains(n, " add column ") {
+		return false
+	}
+	if isGeneratedColumnAddSQL(n) {
+		return false
+	}
+	if strings.Contains(n, " not null") && !strings.Contains(n, " default ") {
+		return false
+	}
+	return true
+}
+
+func isGeneratedColumnAddSQL(normalizedSQL string) bool {
+	if strings.Contains(normalizedSQL, " generated ") {
+		return true
+	}
+	if !strings.Contains(normalizedSQL, " as (") {
+		return false
+	}
+	return strings.Contains(normalizedSQL, " stored") || strings.Contains(normalizedSQL, " virtual")
+}
+
+// IsIgnorableMySQLError reports whether a DDL error is benign for idempotent
+// schema repair paths.
+func IsIgnorableMySQLError(err error) bool {
+	var me *mysql.MySQLError
+	if errors.As(err, &me) {
+		switch me.Number {
+		case 1050, 1060, 1061:
+			return true
+		}
+		msg := strings.ToLower(me.Message)
+		return strings.Contains(msg, "already exists") || strings.Contains(msg, "duplicate")
+	}
+	msg := strings.ToLower(err.Error())
+	return strings.Contains(msg, "already exists") || strings.Contains(msg, "duplicate")
+}
+
+// CRC32Version hashes normalized schema statements and always returns a
+// non-zero integer so callers can use it as a stable schema version marker.
+func CRC32Version(stmts []string) int {
+	h := crc32.ChecksumIEEE([]byte(strings.Join(stmts, "\n")))
+	if h == 0 {
+		h = 1
+	}
+	return int(h)
+}
--- a/pkg/meta/meta.go
+++ b/pkg/meta/meta.go
@@ -10,7 +10,7 @@ import (
 	"strings"
 	"time"
 
-	mysql "github.com/go-sql-driver/mysql"
+	"github.com/mem9-ai/dat9/internal/schemaspec"
 	"github.com/mem9-ai/dat9/pkg/logger"
 	"github.com/mem9-ai/dat9/pkg/metrics"
 	"github.com/mem9-ai/dat9/pkg/mysqlutil"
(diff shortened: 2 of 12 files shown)


srstack:
Addressed the remaining review items in commit 4880120:\n\n- Meta schema migration is now concurrency-idempotent by ignoring benign already-exists/duplicate DDL errors during repair apply.\n- Meta schema spec now tracks PRIMARY KEY constraints and reports missing PK as residual drift (non-repairable).\n- Meta repair planner now gates repairs to safe operations (missing table, safe ADD COLUMN, and non-unique index; unique index only when table is newly missing).\n- Tenant repair planner now uses the same safety gating policy, preventing risky live repairs while still converging for missing-table bootstrap paths.\n\nValidation:\n- go test ./pkg/meta ./pkg/tenant/schema\n- make lint\n\nI also resolved the corresponding review threads after applying and validating these changes.

srstack:
Addressed the new review items:\n\n- Extended tenant schema parsing to also ingest ALTER TABLE index statements (including FULLTEXT/VECTOR forms), so these indexes are now part of schema contract diff/repair planning.\n- Hardened add-column safety checks to treat AS (...) STORED/VIRTUAL generated columns as unsafe for automatic repair.\n- Added tests covering ALTER TABLE index parsing/repair planning and generated-column safety gating.\n\nValidation:\n- go test ./pkg/tenant/schema

srstack:
Addressed all 5 new review items in commit 1bdaa6b:

- **Context threading** (coderabbitai): Threaded `context.Context` through `migrate`, `diffMetaSchema`, `diffMetaTable`, `applyMetaSchemaRepairs`, `loadMetaTableMeta`, and `loadMetaShowCreateTable`; replaced all `Query`/`QueryRow`/`Exec` calls with their `Context` variants.
- **Generated column safety** (coderabbitai): `isSafeAddColumnRepairSQL` now also rejects `AS (...) STORED` / `AS (...) VIRTUAL` patterns in addition to the existing `GENERATED` keyword check.
- **CRC32 negative hash** (copilot): Changed `int(int32(h))` → `int(h)` so schema versions are always non-negative on 64-bit platforms. Also narrowed the hash to spec-relevant statements only (`CREATE TABLE`, `CREATE [UNIQUE] INDEX`, `ALTER TABLE ... ADD ... INDEX`), avoiding spurious re-Ensures on non-spec-affecting statement edits.
- **Dead struct fields** (copilot): Removed `isUnique bool` and `isRepairable bool` from `metaIndexSpec` — both were set but never read anywhere in the diff/repair logic.
- **Pool.go schema_version comment** (copilot): Added clarifying comment explaining that `UpdateTenantSchemaVersion` is only called after `EnsureTiDBSchemaForMode` succeeds, which itself ends with `ValidateTiDBSchemaForMode`, so the stored version is a strong invariant of schema correctness at that point in time.

qiffang (commented):
## Review summary

Nice direction — moving from hardcoded migration scripts to parse → diff → safe-fix → re-diff is a clear improvement, and the safety filters (rejecting `NOT NULL` without default, generated columns, unique indexes on existing tables) are exactly the right guardrails. Test coverage is solid.

A few things to address before merging.

### Must-fix

1. **Dead code: `legacyTiDBUploadsRepairStatements`** (`pkg/tenant/schema/tidb_auto.go`)
   Its only caller `repairLegacyTiDBUploadsSchema` was removed in this PR, but the function itself (and `TestLegacyTiDBUploadsRepairStatements`) remains. Either delete both, or wire it back in. Right now it's orphaned.

2. **Significant duplication between `pkg/meta/meta.go` and `pkg/tenant/schema/tidb_auto.go`.**
   The following are near-line-for-line copies:
   - `parseCreateTableStatement` / `parseMetaCreateTableStatement`
   - `splitTopLevelComma` / `splitMetaTopLevelComma`
   - `parseColumnDefinition` / `parseMetaColumnDefinition`
   - `parseColumnType` / `parseMetaColumnType`
   - `splitIdentifierAndRest` / `splitMetaIdentifierAndRest`
   - `normalizeSQLFragment` / `normalizeMetaSQLFragment`
   - `isIgnorableTiDBSchemaError` / `isIgnorableMetaSchemaError`
   - `isSafeAddColumnRepairSQL` (duplicated verbatim)
   - `schemaSnippet` / `schemaStatementSnippet`

   This parser is non-trivial (quoted identifiers, nested parens, etc.). Having two copies means future bug fixes have to be mirrored; they will inevitably drift. Please extract into a shared internal package (e.g. `pkg/schemaparser` or `pkg/mysqlutil/schemaspec`) before this lands.

### Should-fix

3. **`CurrentTiDBTenantSchemaVersion` hash is whitespace-sensitive** (`pkg/tenant/schema/tidb_auto.go`)
   Using `crc32.ChecksumIEEE` over raw statement bytes means a pure reformat (re-indent, remove a blank line inside a backtick string literal) forces every tenant to re-Ensure on next open. Two suggestions:
   - Hash the *normalized* statements (`normalizeSQLFragment`) instead of raw bytes, so cosmetic edits don't invalidate.
   - CRC32 has a ~4-byte output with real collision risk if this counter grows. Prefer `sha256.Sum256(...)` truncated to int64, or even just a manual monotonically-increasing constant (which is what the comment proposes fallback-wise anyway).

4. **Pool fast-path trusts meta store as source of truth** (`pkg/tenant/pool.go:325-347`)
   Skipping `EnsureTiDBSchemaForMode` whenever `t.SchemaVersion == CurrentTiDBTenantSchemaVersion` is efficient, but if an operator ever modifies the tenant DB directly (or a prior Ensure recorded the version before a later out-of-band change), drift will go undetected. Consider at minimum a periodic full-validate (e.g., 1-in-N opens, or a startup pass) as defense in depth.

5. **`applyTiDBSchemaRepairs` loop can terminate silently on unconverged drift** (`pkg/tenant/schema/tidb_auto.go`)
   In `EnsureTiDBSchemaForMode`, after `maxRepairPasses = 3` the loop falls through to `ValidateTiDBSchemaForMode` which returns the final diff as an error — good. But if `plannedTiDBSchemaRepairs(diffs)` is empty yet `diffs` is non-empty (unrepairable drift), the loop `break`s on the first pass; that's the intended path. Worth a one-line comment on the `break` explaining that this is the "drift exists but no safe repair" branch and validation will surface it.

### Nits

6. **`parseColumnType` keyword matching is position-dependent** (both files)
   The keyword list includes `" as "` but column definitions like `VARCHAR(64) CHARACTER SET ...` don't trip it; still, the list is not exhaustive (no `CHARACTER SET`, `COLLATE`, `CHECK`, `STORAGE DISK/MEMORY`). Since the diff only flags missing columns, a benign miss (returning the full trailing clause as the "type") means the column-type compare will mismatch and report spurious diffs. Recommend either (a) explicitly documenting supported syntax or (b) expanding the stop-token list.

7. **Backslash-as-escape in `splitTopLevelComma` is not MySQL-accurate**
   Outside of string literals, `\` has no special meaning in MySQL DDL. The `case '\\': i++; continue` happens unconditionally, which would incorrectly skip a legitimate character right after a literal backslash in (say) an identifier. Unlikely to matter for trusted init SQL, but worth a comment acknowledging the parser assumes well-formed, trusted input.

8. **`meta.UpdateTenantSchemaVersion` doc says "best-effort, log but don't fail"** but the function signature returns an error the same as any other Store method. The caller in `pool.go` does the right thing (warn + swallow), but future callers won't know. Consider adding an inline `// Callers should treat a non-nil return as best-effort ...` to the comment on the method, or adding a thin wrapper that discards the error.

9. **`showCreateHasIndex`** uses `strings.Contains(normalizedCreate, indexName+" (")`. OK today because all index names are prefixed `idx_`/`uk_` and live in `KEY name (...)` syntax in SHOW CREATE. But if TiDB ever renders e.g. `... KEY `idx_foo`(...)` without a space (backtick then paren), the check misses. Either use a regex anchored on word boundaries or parse the `KEY`/`INDEX` clauses properly.

Overall direction is good — please address the must-fix items (especially the duplication) and I'm happy to re-review.

srstack:
Addressed the remaining review summary items:

- Extracted the shared schema parser/helper logic into `internal/schemaspec` and removed the orphaned legacy uploads repair dead code in commit `ab4ecdb`.
- Added periodic TiDB schema validation on the version-match fast path in `pkg/tenant/pool.go` as defense in depth in commit `b5320d6`.
- Tightened `CurrentTiDBTenantSchemaVersion` so formatting-only SQL changes no longer churn the version.

Validation:
- `go test ./internal/schemaspec`
- `go test ./pkg/tenant/schema`
- `go test ./pkg/tenant`
- focused `go test` for `pkg/meta`

All inline review threads are resolved at this point.

internal/schemaspec/schemaspec.go
```diff
@@ -0,0 +1,330 @@
...
+	if errors.As(err, &me) {
+		switch me.Number {
+		case 1050, 1060, 1061:
+			return true
+		}
+		msg := strings.ToLower(me.Message)
+		return strings.Contains(msg, "already exists") || strings.Contains(msg, "duplicate")
+	}
+	msg := strings.ToLower(err.Error())
+	return strings.Contains(msg, "already exists") || strings.Contains(msg, "duplicate")
```
qiffang, line 319: fallback 太宽：`strings.Contains(msg, "duplicate")` 会把真正的运行时错误（比如 `Duplicate entry '...' for key '...'` 即 ER_DUP_ENTRY 1062）也判成 ignorable。虽然当前只在 DDL repair 路径上调用（DDL 不会触发 1062），但这是一个放在公共包里、名字通用的 helper，未来很容易被复用到非 DDL 上下文。

建议：
- 只保留 1050/1060/1061 的 code 白名单；
- message fallback 去掉 `"duplicate"`，只留 `"already exists"`，或干脆只信任 code。

对应地 `TestIsIgnorableTiDBSchemaError` 的 `plain duplicate` case（err = `duplicate entry`, want=true）也需要改。

pkg/meta/meta.go
```diff
@@ -110,7 +111,94 @@ func applyMySQLPoolDefaults(db *sql.DB) {
...
+		return err
+	}
+	diffs, err = diffMetaSchema(ctx, s.db, spec)
+	if err != nil {
+		return fmt.Errorf("re-diff meta schema: %w", err)
+	}
+	if len(diffs) > 0 {
+		return &metaSchemaDiffError{diffs: diffs}
+	}
+	return nil
```
qiffang, line 134: 多实例启动时两个 drive9-server 会并发跑这段 parse → diff → repair → re-diff，虽然 `IsIgnorableMySQLError` 能吞掉 `already exists`，但 re-diff 阶段两端都有可能在同一个瞬间看到 drift（对方的 DDL 还没提交到 information_schema）并返回 `metaSchemaDiffError`，把 `Open` 打挂。

建议加一个 advisory lock：
```go
var locked bool
if err := s.db.QueryRowContext(ctx, "SELECT GET_LOCK('drive9_meta_migrate', 30)").Scan(&locked); err != nil || !locked { ... }
defer s.db.ExecContext(ctx, "SELECT RELEASE_LOCK('drive9_meta_migrate')")
```
或者至少在 doc comment 里点明 migrate 依赖启动 serialize。

pkg/tenant/schema/tidb_auto.go
```diff
@@ -501,26 +529,704 @@ func (m tidbTableMeta) requireColumnType(name, want string) error {
...
+	for name := range indexes {
+		names = append(names, name)
+	}
+	sort.Strings(names)
+	return names
+}
+
+func showCreateHasIndex(normalizedCreate, indexName string) bool {
+	needle := strings.ToLower(indexName) + " ("
+	return strings.Contains(normalizedCreate, needle)
```
qiffang, line 1231: `strings.Contains(normalizedCreate, indexName + " (")` 在 SHOW CREATE 字符串里做子串扫描，有两个潜在问题：
- index name 是另一个 index name 的前缀时 → false positive（`idx_foo` 被查 `idx_f` 时）；
- 如果后续在 SHOW CREATE 里遇到 backtick 去除后名字里含空格（理论上 MySQL identifier 是允许的），逻辑就断了。

改成查 `information_schema.statistics` 更稳：
```sql
SELECT index_name, non_unique, seq_in_index, column_name
FROM information_schema.statistics
WHERE table_schema = DATABASE() AND table_name = ?
```
副作用：可以顺带检查 unique flag 和 column 列表，为未来"索引列 mismatch"的 diff 铺路。同样的问题 `pkg/meta/meta.go:680 showCreateHasMetaIndex` 也存在。

pkg/meta/meta.go
```diff
@@ -236,14 +313,377 @@ func (s *Store) migrate() error {
...
+	}
+	sort.Strings(names)
+	return names
+}
+
+func showCreateHasMetaIndex(normalizedCreate, indexName string, spec metaIndexSpec) bool {
+	if spec.isPrimary {
+		return strings.Contains(normalizedCreate, "primary key (")
+	}
+	return strings.Contains(normalizedCreate, indexName+" (")
```
qiffang, line 680: 同 `pkg/tenant/schema/tidb_auto.go:1231` 的问题：SHOW CREATE 子串匹配不够稳，建议改成 `information_schema.statistics` 查询。

pkg/tenant/schema/tidb_auto.go
```diff
@@ -501,26 +529,704 @@ func (m tidbTableMeta) requireColumnType(name, want string) error {
...
+		return true
+	}
+	if strings.HasPrefix(normalized, "create unique index ") {
+		return tableMissing
+	}
+	if strings.HasPrefix(normalized, "alter table ") {
+		if strings.Contains(normalized, " add unique index ") || strings.Contains(normalized, " add unique key ") {
+			return tableMissing
+		}
+		if strings.Contains(normalized, " add fulltext index ") || strings.Contains(normalized, " add vector index ") || strings.Contains(normalized, " add index ") || strings.Contains(normalized, " add key ") {
```
qiffang, line 1114: `ADD FULLTEXT INDEX` / `ADD VECTOR INDEX` 在大表上是小时级重建（尤其配合 `ADD_COLUMNAR_REPLICA_ON_DEMAND`）。对新建 tenant 没问题（小表），但 `EnsureTiDBSchemaForMode` 也会对老 tenant 在检测到缺这些索引时 auto-repair —— 一次"schema_version 没命中 fast path 的升级"就可能把生产卡住。

建议任一：
- 和 unique index 一样：`tableMissing=true` 时才允许；
- 保留当前行为但在 `applyTiDBSchemaRepairs` 前打 `logger.Warn`，带上 tenant_id 和 index name，给 on-call 介入机会；
- 或引入显式的 `--allow-heavy-index-repair` flag，默认只记录 diff、不自动 apply。

pkg/tenant/schema/tidb_auto.go
```diff
@@ -501,26 +529,704 @@ func (m tidbTableMeta) requireColumnType(name, want string) error {
...
+	if rest == "" {
+		return "", "", "", false
+	}
+	fields := strings.Fields(rest)
+	if len(fields) == 0 {
+		return "", "", "", false
+	}
+	table := strings.ToLower(fields[0])
+
+	markers := []string{" add fulltext index ", " add vector index ", " add unique index ", " add index ", " add key "}
```
qiffang, line 918: markers 列表里有 `" add unique index "` 但没有 `" add unique key "`。当前 init SQL 没踩到这种写法，但本 PR 的初衷是"任意合法 init SQL 都能被 parser 识别"，这里是一个可被 schema 迭代踩到的缺口。补一条 `" add unique key "`（顺序要放在 `" add key "` 前面）就行。

pkg/tenant/pool.go
```diff
@@ -426,6 +461,18 @@ func (p *Pool) createBackend(ctx context.Context, t *meta.Tenant) (*backend.Dat9
...
 
+func (p *Pool) shouldPeriodicValidateTiDBSchemaOnOpen() bool {
+	if p == nil {
+		return false
+	}
+	every := periodicTiDBSchemaValidationEvery
+	if every == 0 {
+		return false
+	}
+	count := p.tidbSchemaValidationOpens.Add(1)
```
qiffang, line 472: `tidbSchemaValidationOpens` 是 pool-wide 的计数器，触发 validate 的条件是"pool 里第 N 次 open"，和具体 tenant 无关。在 `every=32`、有 100 个 tenant、访问极不均匀的场景下，热门 tenant 会被频繁 validate，低频 tenant 可能几周都轮不到。

如果 periodic validation 的目标是捕获"out-of-band drift"，per-tenant cadence 更公平：在 `tenants` 表里加一列 `last_schema_validated_at`，每次 open 看 `time.Since(lastValidatedAt) > interval`。可以 follow-up 做，不 block。

pkg/tenant/pool.go
```diff
@@ -323,12 +334,36 @@ func (p *Pool) createBackend(ctx context.Context, t *meta.Tenant) (*backend.Dat9
...
+				// point the schema has been confirmed to match the spec.
+				// Any tenant whose schema diverges between two consecutive
+				// opens will be caught on the next open because its stored
+				// version will differ from CurrentTiDBTenantSchemaVersion.
+				if verErr := p.metaStore.UpdateTenantSchemaVersion(ctx, t.ID, schema.CurrentTiDBTenantSchemaVersion); verErr != nil {
+					logger.Warn(ctx, "tenant_pool_update_schema_version_failed",
+						zap.String("tenant_id", t.ID),
+						zap.Int("version", schema.CurrentTiDBTenantSchemaVersion),
+						zap.Error(verErr))
+				}
```
qiffang, line 357: 注释说明 best-effort，逻辑合理，但有一个放大效应值得提一句：如果 meta DB 持续不可用（网络/权限 drift/表锁），每个 tenant 的每次 open 都会跑完整 `EnsureTiDBSchemaForMode`（重 diff 所有表、再 validate），对热点 tenant 是压力。

建议至少为 `UpdateTenantSchemaVersion` 失败加一个 counter 指标（`tenant_pool_update_schema_version_failed{tenant_id}`），SRE 能看到放大效应。

internal/schemaspec/schemaspec.go
```diff
@@ -0,0 +1,330 @@
...
+		}
+		suffix := " " + strings.ToLower(rest[i:])
+		for _, kw := range keywords {
+			if strings.HasPrefix(suffix, kw) {
+				return strings.TrimSpace(rest[:i])
+			}
+		}
+	}
+	return strings.TrimSpace(rest)
+}
```
qiffang, line 278: 这个循环每次迭代都 `" " + strings.ToLower(rest[i:])`，对长列定义是 O(n²)。可以把 `strings.ToLower(rest)` 和原始 `rest` 一起走 index，避免反复分配字符串。

实际上 DDL 解析只在启动时跑一次，影响可忽略，算风格建议。

internal/schemaspec/schemaspec.go
```diff
@@ -0,0 +1,330 @@
...
+	if !strings.HasPrefix(n, "alter table ") || !strings.Contains(n, " add column ") {
+		return false
+	}
+	if isGeneratedColumnAddSQL(n) {
+		return false
+	}
+	if strings.Contains(n, " not null") && !strings.Contains(n, " default ") {
+		return false
+	}
+	return true
```
qiffang, line 293: `strings.Contains(n, " default ")` 要求 `default` 后有空格 + 至少 1 字符，碰上 `DEFAULT(NOW())`（无空格的函数默认值写法，MySQL 8 合法）会被判为"没有 default"从而拒绝。init SQL 里没有这种写法，属于边缘，提一下。

qiffang (commented):
## Review overall

整体方向很好 — 把 init SQL 作为单一 source of truth、用结构化 diff 驱动安全修复，比之前按手动 migration 号维护要健壮。`CurrentTiDBTenantSchemaVersion` 基于内容 hash 的 fast-path + 周期性 validate 兜底这个组合我很喜欢。测试覆盖也比较系统（missing column / missing PK / composite PK / unsafe repair skip / tinyint(1) 等价等都有覆盖）。

下面是我过一遍代码找到的问题，按三档排序，细节见 inline comments：

### Blocking / correctness
1. **`IsIgnorableMySQLError` fallback 太宽**（`internal/schemaspec/schemaspec.go:319`）：`strings.Contains(msg, "duplicate")` 会吞掉 1062 `Duplicate entry` 等真正的运行时错误。建议只保留 1050/1060/1061 白名单，message fallback 去掉或只留 `already exists`。
2. **多实例 `migrate()` 没有 serialize**（`pkg/meta/meta.go:134`）：两个 drive9-server 并发启动会同时跑 DDL，re-diff 阶段可能误报 drift。建议加 `GET_LOCK()` advisory lock 或在注释里点明需要启动串行。

### 需要讨论
3. **索引存在性判断用 SHOW CREATE 子串匹配**（`showCreateHasIndex` / `showCreateHasMetaIndex`）：index name 互为 substring 时会误判，建议换成 `information_schema.statistics`，也能为未来的索引列 mismatch 检测铺路。
4. **`isSafeAddIndexRepairSQL` 把 `ADD FULLTEXT / ADD VECTOR INDEX` 无条件判为 safe**：在大表上可能触发小时级重建。建议只在 `tableMissing=true` 时允许，或至少输出 warn log。
5. **`parseAlterTableAddIndexStatement` markers 少了 `" add unique key "`**：当前 init SQL 没踩到但是个 init-SQL-driven 初衷下的 parser 缺口。
6. **`tidbSchemaValidationOpens` 是 pool-wide counter**：低频 tenant 可能几周不被 validate，per-tenant cadence 更均匀（次级）。
7. **`UpdateTenantSchemaVersion` 持续失败时每次 open 都跑全量 Ensure**：缺乏限频保护，建议加 metric（次级）。

### Nits
8. `parseColumnType` 内层循环每次都 `ToLower(rest[i:])`，O(n²)，可以提出循环外（启动一次，影响可忽略）。
9. `tidbAutoEmbeddingSchemaStatements()` 和 `tidb_cloud_starter` 是双份事实，建议 CI 里加一致性检查（`dump-init-sql` vs. 外部 schema 文件）。
10. `meta` 和 `tidb_auto` 的 diff error `.Error()` 基本重复，可以抽 helper。
11. `TestIsIgnorableTiDBSchemaError` 的 `plain duplicate` 用例 enshrine 了 #1 里我建议收紧的行为，如果接受 #1 需要调整此 case。

### 建议行动
先修 #1 和 #3，其他可以 follow-up。代码已经能看出多轮 review 的迭代痕迹，质量不错。

qiffang (approved):

qiffang:
I went through the full review thread (Copilot, CodeRabbit, plus an independent pass on my side) and grouped the feedback into two buckets: **items that appear to be addressed in follow-up commits** and **remaining points that are still worth checking before merge**.

### Feedback that looks addressed in follow-up commits

From the later commits in this PR, the following review themes appear to have been handled:

- preserving table-level constraints instead of dropping them from the schema contract
- fixing the `tidbSchemaSpecFromStatements` slice-pointer bug
- making missing-table repair converge with external `CREATE INDEX` statements
- tightening repair planning so only explicitly safe fixes are auto-applied
- threading `context.Context` through the tenant-schema validation / ensure path
- making schema repair execution more idempotent under concurrent runs

That covers most of the big issues raised earlier in the thread.

### Remaining points I would still double-check before merge

1. **Index contract checks still look name-based instead of definition-based**
   In both schema-diff paths, the current logic still appears to accept an index if the expected index name exists in `SHOW CREATE TABLE`, without fully verifying uniqueness / primary-ness and column list. That means drift like “same index name, wrong columns” or “same name, wrong unique/non-unique type” can still slip through and incorrectly pass validation.

2. **Periodic tenant-schema validation is global, not per-tenant**
   `shouldPeriodicValidateTiDBSchemaOnOpen` is driven by a single pool-level open counter. In practice, a hot tenant can consume most open events, while another tenant with out-of-band schema drift may rarely hit the validation slot. If the goal is to periodically detect drift for each tenant after the version gate matches, this probably needs a per-tenant counter or time-based TTL instead of a global modulo.

3. **Forward-compatibility note on schema version hashing**
   The computed schema version is derived from the subset of statements that are parsed into the contract. That is fine for the current statement set, but it does mean future standalone schema changes that are not reflected in the parsed contract (for example, a future `ALTER TABLE ... ADD COLUMN` that is not folded back into the canonical `CREATE TABLE`) could fail to bump the version gate and therefore skip `EnsureTiDBSchemaForMode` for existing tenants. Not necessarily a blocker for this PR, but worth keeping in mind as a contract rule going forward.

### Short version

At this point the PR looks much better than the initial revision, and most of the earlier review comments seem incorporated. The two main correctness concerns I would still focus on are:

- **index drift should be validated by definition, not only by name**
- **periodic validation should match the intended granularity (likely per tenant, not global pool-wide opens)**

If those are addressed or explicitly accepted as follow-up tradeoffs, the rest looks mostly in good shape.

PR #316, opened 2026-04-23
Fix TiDB schema index auto-repair

srstack:
## Summary
Fix TiDB tenant schema auto-repair so missing required indexes on existing tenants can actually be repaired instead of being re-reported as schema drift.

## What changed
- switch TiDB index existence detection away from SHOW CREATE substring matching to observed index names parsed from schema metadata / create output
- allow missing unique indexes to participate in auto-repair for existing tables so required indexes like `idx_uploads_active` are restored
- add and update focused tests for observed-index detection and repair planning

## Verification
- `go test ./pkg/tenant/schema ./pkg/tenant`

## Summary by CodeRabbit

* **Bug Fixes**
  * More accurate detection of missing indexes and safer unique-index repairs with preflight duplicate-row checks; improved handling of index definitions and DDL error classification; migration now uses a lock when applying meta schema diffs.

* **Tests**
  * Expanded end-to-end and unit tests for unique-index detection/repairs, index-parsing, duplicate-row preflight behavior, error classification, and migration metrics.

* **Chores**
  * Added metric emission for schema-version update failures.

### DIFF
--- a/internal/schemaspec/schemaspec.go
+++ b/internal/schemaspec/schemaspec.go
@@ -215,6 +215,7 @@ func ParseColumnType(rest string) string {
 	if rest == "" {
 		return ""
 	}
+	lowerRest := strings.ToLower(rest)
 	keywords := []string{
 		" not ",
 		" null",
@@ -267,7 +268,7 @@ func ParseColumnType(rest string) string {
 		if inSingle || inDouble || inBacktick || depth > 0 {
 			continue
 		}
-		suffix := " " + strings.ToLower(rest[i:])
+		suffix := " " + lowerRest[i:]
 		for _, kw := range keywords {
 			if strings.HasPrefix(suffix, kw) {
 				return strings.TrimSpace(rest[:i])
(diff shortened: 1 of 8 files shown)


qiffang:
Overall direction makes sense, and the PR seems to address the original false-negative around unique index detection (for example `UNIQUE KEY idx_uploads_active (...)` no longer depends on brittle `SHOW CREATE` substring matching).

After going through the code, the existing bot reviews, and an additional independent pass, here are the main follow-up points I would still call out:

### 1. `information_schema.statistics` is now a hard dependency
`diffTiDBTable()` now fails the whole diff/repair path if `loadTiDBIndexNames()` fails.

That is more accurate than the old heuristic, but it also means a metadata-query failure (permissions / TiDB behavior differences / transient issues) now blocks schema validation entirely. It would be safer to consider a best-effort fallback to `SHOW CREATE TABLE` parsing instead of immediately returning an error.

### 2. Auto-repairing unique indexes on existing tables is a policy tradeoff
This PR now treats missing unique indexes as auto-repairable even when the table already exists.

That fixes the original bug, but it also means `CREATE UNIQUE INDEX` may now fail repeatedly if the live table already contains duplicate data. That may be acceptable if the intent is "required unique contract must be enforced or fail loudly", but it is worth confirming explicitly because this is stronger behavior than before.

### 3. The parser/test helper path still looks narrower than the production path
The production path now queries `information_schema.statistics`, which is good. But the helper parser used in tests still appears syntax-sensitive and may not cover every valid form (for example some `CONSTRAINT ... UNIQUE (...)` variants).

That is not necessarily a blocker for production behavior, but it does mean the unit-test helper may under-approximate valid DDL forms.

### 4. Test coverage would be stronger with one integration-style check
The new unit test that verifies `UNIQUE KEY idx_uploads_active (...)` is a good regression test for the parser path.

Still, the production code now depends on the real `information_schema.statistics` query path. One integration-style test that creates a table with a unique index and verifies that the real metadata path does not report it missing would make this much more robust.

---

### Summary
My read is:
- the PR **does fix the immediate bug it is targeting**
- I do **not** currently see a clear blocker in the change itself
- but the team should still make an explicit call on:
  1. whether `information_schema` failure should have a fallback path
  2. whether unique-index auto-repair on existing tables is an intended enforcement policy
  3. whether to add one integration-style metadata test before merge

So overall: **directionally good fix, with a few follow-up design/robustness questions worth resolving or explicitly accepting before merge.**

qiffang (approved):

PR #325, opened 2026-04-24
fix: auto-repair description_embedding column and FULLTEXT/VECTOR indexes for auto-embedding tenants

srstack:
## Problem

Auto-embedding tenants (mode `auto-embedding`) started failing to acquire backends after PR #318 added `description_embedding`, `idx_files_desc_cosine`, and updated `idx_fts_content_desc`. Existing tenant schemas were missing these objects, and the repair logic had two gates that prevented them from being fixed:

```
files schema contract: missing description_embedding column
files schema contract: missing idx_files_desc_cosine index
files schema contract: missing idx_fts_content_desc index
```

## Root Cause

1. **`isSafeAddColumnRepairSQL`** delegated to `schemaspec.IsSafeAddColumnRepairSQL`, which rejects all `GENERATED` columns. The `description_embedding` column is `VECTOR(...) GENERATED ALWAYS AS (EMBED_TEXT(...)) STORED` — TiDB Cloud computes it server-side, so adding it to an existing table is actually safe.

2. **`isSafeAddIndexRepairSQL`** returned `tableMissing` for `ADD FULLTEXT INDEX` / `ADD VECTOR INDEX`, meaning these repairs were silently dropped whenever the table already existed (which is always the case for real tenants).

## Fix

- **`isSafeAddColumnRepairSQL`**: whitelist `STORED GENERATED VECTOR ... EMBED_TEXT(...)` columns — TiDB computes the value server-side, no existing row data is touched.
- **`isSafeAddIndexRepairSQL`**: remove the `tableMissing` gate for FULLTEXT/VECTOR; always return `true`. TiDB Cloud (the only platform for auto-embedding mode) supports `ADD_COLUMNAR_REPLICA_ON_DEMAND`. `applyTiDBSchemaRepairs` already has a `isIgnorableOptionalSchemaError` fallback that warns and continues if a version doesn't support the syntax.
- **`applyTiDBSchemaRepairs`**: added `isFulltextOrVectorIndexRepairSQL` + `isIgnorableOptionalSchemaError` fallback so a non-fatal TiDB error (e.g. `8200: FULLTEXT index must specify one column name` on an unsupported config) does not abort the entire repair run.

## Testing

- `TestIsSafeAddColumnRepairSQLAllowsStoredGeneratedVectorWithEmbedText` — new test verifying the whitelist
- `TestTiDBSchemaSpecForModeIncludesAlterTableIndexes` — updated to assert FULLTEXT/VECTOR indexes are in the auto-mode contract
- `TestPlannedTiDBSchemaRepairsIncludesFulltextVectorIndexOnExistingTable` — updated to assert repairs are planned for existing tables
- All existing `pkg/tenant/schema` tests pass

## Release Notes

* **New Features**
  * Added support for safely adding stored VECTOR columns to existing tables during automatic schema repairs.

* **Bug Fixes**
  * Schema repair operations now gracefully handle failures from FULLTEXT and VECTOR index operations that match optional feature errors, logging and skipping them instead of aborting the repair process.

### DIFF
--- a/pkg/tenant/schema/tidb_auto.go
+++ b/pkg/tenant/schema/tidb_auto.go
@@ -1341,6 +1341,17 @@ func isSafeTiDBRepairDiff(diff tidbSchemaDiff, tableMissing map[string]bool) boo
 }
 
 func isSafeAddColumnRepairSQL(sqlText string) bool {
+	// STORED GENERATED VECTOR columns whose expression uses EMBED_TEXT are
+	// safe to add to existing tables via ALTER TABLE: TiDB computes the value
+	// server-side without touching existing row data. This covers the
+	// description_embedding column introduced for auto-embedding mode.
+	normalized := normalizeSQLFragment(sqlText)
+	if strings.Contains(normalized, " generated ") &&
+		strings.Contains(normalized, " stored") &&
+		strings.Contains(normalized, " vector(") &&
+		strings.Contains(normalized, "embed_text(") {
+		return true
+	}
 	return schemaspec.IsSafeAddColumnRepairSQL(sqlText)
 }
 
@@ -1357,7 +1368,11 @@ func isSafeAddIndexRepairSQL(sqlText string, tableMissing bool) bool {
 			return true
 		}
 		if strings.Contains(normalized, " add fulltext index ") || strings.Contains(normalized, " add vector index ") {
-			return tableMissing
+			// FULLTEXT and VECTOR indexes are always safe to add on an existing
+			// table in auto-embedding mode: TiDB Cloud supports the syntax, and
+			// applyTiDBSchemaRepairs will gracefully skip with a warning if the
+			// current TiDB version does not.
+			return true
 		}
 		if strings.Contains(normalized, " add index ") || strings.Contains(normalized, " add key ") {
 			return true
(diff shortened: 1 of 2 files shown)


mornyx (approved):

pkg/tenant/schema/tidb_auto.go
```diff
@@ -1384,6 +1395,17 @@ func applyTiDBSchemaRepairs(ctx context.Context, db *sql.DB, statements []string
 			if isIgnorableTiDBSchemaError(err) {
 				continue
 			}
+			// FULLTEXT and VECTOR index repairs may fail with optional-feature
+			// errors on TiDB versions or configurations that do not support
+			// them (e.g. 8200: FULLTEXT index must specify one column name,
+			// 1105: FULLTEXT index is not supported). Treat these the same as
+			// when the statement was skipped during initial provisioning.
+			if isFulltextOrVectorIndexRepairSQL(stmt) && isIgnorableOptionalSchemaError(err) {
```
qiffang, line 1403: I think this warn-and-continue path still leaves unsupported TiDB versions or configurations broken. EnsureTiDBSchemaForMode() always finishes with ValidateTiDBSchemaForMode() (lines 350-368), and auto mode still treats these FULLTEXT/VECTOR indexes as part of the required schema contract. So if db.ExecContext hits an optional-feature error here, we only defer the failure until the final validation step; tenant open will still fail, and schema_version will never advance. If the goal is to make unsupported configs non-fatal, we need to relax validation for this case as well. Otherwise I would keep this path failing loudly and rely on TiDB Cloud support being mandatory.

PR #327, opened 2026-04-24
fix(schema,search): split TiDB multi-column FULLTEXT index and fix auto-embedding description revision tracking

mornyx:
## Problem

1. **TiDB does not support multi-column FULLTEXT indexes.** The previous `idx_fts_content_desc(content_text, description)` was silently failing during schema init because `isIgnorableOptionalSchemaError` swallowed Error 8200 (`FULLTEXT index must specify one column name`). As a result, the FULLTEXT index was never created on TiDB Cloud.

2. **Description vector search was broken in auto-embedding mode.** `VectorSearchDescriptionByText` included a `description_embedding_revision = revision` gate, but in auto-embedding mode `description_embedding_revision` was never being set (auto-embedding uses a `GENERATED ALWAYS` column with no semantic worker write-back). This caused description vector search to always return empty results.

### Schema
- Split `idx_fts_content_desc(content_text, description)` into two single-column FULLTEXT indexes:
  - `idx_fts_content(content_text)`
  - `idx_fts_description(description)`
- Update `tidb_auto_test.go` assertions to match the new index names
- Update `file-description-design.md` to document the two-index approach

### Search
- `FTSSearch`: Query both columns via `UNION ALL` and merge results with `MAX(score)`
- `buildVectorSearchDescriptionByTextQuery`: Remove the `description_embedding_revision = revision` gate for auto-embedding mode (generated column is always current)

### Datastore — revision tracking fixes for auto-embedding
- `updateFileContentTx`: When `preserveEmbedding=true` (auto mode) and description changes, set `description_embedding_revision = revision` instead of attempting to `NULL` the generated column
- `ConfirmPendingFileAutoEmbeddingTx`: Sync `description_embedding_revision = revision` when confirming a pending file with description
- `InsertFileTx`: Include `description_embedding_revision` column in the INSERT
- `createAndWriteCtx`: Populate `DescriptionEmbeddingRevision` in auto mode

### E2E
- Add `e2e/verify-description-tidb-zero-e2e.sh` — end-to-end test against TiDB Cloud Zero with native auto-embedding, vector search, and full-text search. No local Docker or Ollama required.

## Verification

- `make lint` — 0 issues
- `go test ./pkg/datastore/... ./pkg/backend/... ./pkg/tenant/schema/...` — all PASS
- `bash e2e/verify-description-tidb-zero-e2e.sh` — **8/8 checks PASS** (description storage, auto-embedding generation, overwrite semantics, vector semantic search, FTS search)

## Notes

- FTS on TiDB Zero uses columnar replica (`ADD_COLUMNAR_REPLICA_ON_DEMAND`). In testing, the index became available **within 4 seconds** of file upload.
- The `pkg/server` semantic worker tests that fail with `tenant_id doesn't exist in table` are pre-existing MySQL schema compatibility issues, unrelated to this change.

## Summary by CodeRabbit

* **New Features**
  * File descriptions are now searchable via full-text search
  * File descriptions support semantic vector search for similarity matching
  * File search now includes descriptions alongside content in results

* **Improvements**
  * Optimized database indexing strategy to improve search performance

### DIFF
--- a/pkg/datastore/file_tx.go
+++ b/pkg/datastore/file_tx.go
@@ -10,11 +10,13 @@ import (
 func (s *Store) InsertFileTx(db execer, f *File) error {
 	_, err := db.Exec(`INSERT INTO files
 		(file_id, storage_type, storage_ref, content_blob, content_type, size_bytes, checksum_sha256,
-		 revision, status, source_id, content_text, description, created_at, confirmed_at, expires_at)
-		VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)`,
+		 revision, status, source_id, content_text, description, description_embedding_revision,
+		 created_at, confirmed_at, expires_at)
+		VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?)`,
 		f.FileID, f.StorageType, f.StorageRef, nilBytes(f.ContentBlob), nullStr(f.ContentType),
 		f.SizeBytes, nullStr(f.ChecksumSHA256), f.Revision, f.Status,
 		nullStr(f.SourceID), nullStr(f.ContentText), nullStr(f.Description),
+		nullInt64Ptr(f.DescriptionEmbeddingRevision),
 		f.CreatedAt.UTC(), nilTime(f.ConfirmedAt), nilTime(f.ExpiresAt))
 	return err
 }
@@ -48,7 +50,13 @@ func (s *Store) updateFileContentTx(db execer, fileID string, expectedRevision i
 		query += ` description = ?,`
 		args = append(args, description)
 		if currentDesc.String != description {
-			query += ` description_embedding = NULL, description_embedding_revision = NULL,`
+			if preserveEmbedding {
+				// Auto-embedding mode: the database owns embedding via generated
+				// column, so we only need to keep the revision tracker in sync.
+				query += ` description_embedding_revision = revision,`
+			} else {
+				query += ` description_embedding = NULL, description_embedding_revision = NULL,`
+			}
 		}
 	}
 	if preserveEmbedding {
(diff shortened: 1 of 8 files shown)


srstack (commented):
LGTM — ready to merge.

**Root causes confirmed:**

1. TiDB rejects multi-column FULLTEXT indexes (Error 8200). The previous `idx_fts_content_desc(content_text, description)` was silently swallowed by `isIgnorableOptionalSchemaError` and was never actually created.
2. `VectorSearchDescriptionByText` had a `description_embedding_revision = revision` guard that could never be satisfied in auto-embedding mode — the `description_embedding` column is a `GENERATED ALWAYS` column managed by the database, so no semantic worker ever writes `description_embedding_revision` back.

**Schema split** (`idx_fts_content` + `idx_fts_description`): correct. Single-column FULLTEXT is the only supported form on TiDB.

**`FTSSearch` UNION ALL**: correct. Querying both columns independently and merging with `MAX(score) GROUP BY file_id` is the right approach. The outer `JOIN files f` to recover `size_bytes` after the subquery is necessary and handled correctly.

**Revision gate removal** in `buildVectorSearchDescriptionByTextQuery`: safe. `VectorSearchDescriptionByText` is only called inside the `UsesDatabaseAutoEmbedding()` branch in `dat9.go` — app mode takes the separate `VectorSearchDescription` path. No regression for app-mode tenants.

**Write-path revision fixes** (`InsertFileTx`, `updateFileContentTx`, `ConfirmPendingFileAutoEmbeddingTx`, `createAndWriteCtx`): all correct. Auto mode cannot NULL a generated column, so keeping `description_embedding_revision` in sync with `revision` is the right invariant.

**Pre-318 tenant migration**: `idx_fts_content_desc` was never successfully created (Error 8200 silenced), so there is nothing to drop. After this PR the schema contract contains `idx_fts_content` and `idx_fts_description` — both single-column, both repairable via the existing `isSafeAddIndexRepairSQL` path. Clean upgrade.

One minor note: the e2e script uses `X-Dat9-API-Key` in some curl calls and `Authorization: Bearer` in others — worth making consistent, but not a blocker.

pkg/datastore/file_tx.go
```diff
@@ -48,7 +50,13 @@ func (s *Store) updateFileContentTx(db execer, fileID string, expectedRevision i
 		query += ` description = ?,`
 		args = append(args, description)
 		if currentDesc.String != description {
-			query += ` description_embedding = NULL, description_embedding_revision = NULL,`
+			if preserveEmbedding {
+				// Auto-embedding mode: the database owns embedding via generated
+				// column, so we only need to keep the revision tracker in sync.
+				query += ` description_embedding_revision = revision,`
```
qiffang: I think this needs to track the post-update revision, not the pre-update one. The same statement later does `revision = revision + 1`, so `description_embedding_revision = revision` here leaves the tracker one behind after any auto-embedding overwrite that changes the description. `Store.UpdateFileContent()` in `store.go` already uses `revision + 1` for the same synchronization.

srstack (approved):

PR #336, opened 2026-04-25
fuse: cap streaming initiate totalSize at 20 GiB

qiffang:
## Summary

- Streaming uploads (`StreamUploader.SubmitPart`) initiated multipart with `totalSize = math.MaxInt64`, which the server rejects against `DRIVE9_MAX_UPLOAD_BYTES` with HTTP 413, breaking large sequential writes (observed with juicefs bench 1 GiB bigfiles in prod).
- Cap the declared upper bound at a 20 GiB constant so initiate passes when the server cap is set to a matching or higher value.
- Pair with a server-side env change: set `DRIVE9_MAX_UPLOAD_BYTES >= 21474836480` on `dat9-server` so 20 GiB streaming initiates clear the check.

## Test plan

- [x] `go build ./...`
- [x] `go test ./pkg/fuse/ -run TestStreamUploader -count=1`
- [ ] Re-run juicefs bench against the prod `dat9-server` after raising `DRIVE9_MAX_UPLOAD_BYTES` and confirm 1 GiB bigfile uploads no longer hit `upload too large`.

## Summary by CodeRabbit

* **New Features**
  * Status endpoint now returns a structured response including the tenant upload size limit; clients can read a cached upload-cap value.

* **Bug Fixes**
  * Streaming multipart uploads now buffer parts and finalize using the actual total size to avoid partial uploads.
  * Abort immediately releases buffered memory; release/flush behavior adjusted to handle active streaming consistently.

* **Tests**
  * Endpoint tests updated to validate the structured status response and reported upload limit.

### DIFF
--- a/pkg/fuse/stream_upload.go
+++ b/pkg/fuse/stream_upload.go
@@ -9,11 +9,12 @@ import (
 	"github.com/mem9-ai/dat9/pkg/client"
 )
 
-// streamingInitiateUpperBound is the totalSize declared to the server when a
-// streaming upload starts before the final file size is known. The server
-// validates this against DRIVE9_MAX_UPLOAD_BYTES, so the value must not exceed
-// the server's configured cap.
-const streamingInitiateUpperBound int64 = 20 << 30 // 20 GiB
+// streamingInitiateFallbackUpperBound is used when the client cannot discover
+// the tenant's max_upload_bytes via GET /v1/status (older server, network
+// hiccup, etc.). It must not exceed any server's configured cap, so we keep
+// it modest. Typical streaming workloads (juicefs bench bigfile = 1 GiB) sit
+// well below this.
+const streamingInitiateFallbackUpperBound int64 = 10 << 30 // 10 GiB
 
 // StreamUploader manages parallel part uploads both during Write() for
 // sequential streaming and at flush/close time for non-sequential files.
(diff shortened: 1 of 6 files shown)


pkg/fuse/stream_upload.go
```diff
@@ -112,10 +118,16 @@ func (su *StreamUploader) SubmitPart(ctx context.Context, partNum int, data []by
...
+	// bound. Server enforces this against the tenant's max_upload_bytes at
+	// initiate time; ask the server (cached after first call) and fall back
+	// to a conservative constant if discovery fails.
 	if !su.started {
-		su.writer = su.client.NewStreamWriterConditional(ctx, su.path, math.MaxInt64, su.expectedRevision)
+		upper := su.client.MaxUploadBytes(ctx)
+		if upper <= 0 {
+			upper = streamingInitiateFallbackUpperBound
+		}
+		su.writer = su.client.NewStreamWriterConditional(ctx, su.path, upper, su.expectedRevision)
```
qiffang: I do not think this fixes the streaming path end-to-end. `upper` is being passed as `totalSize`, but the v2 upload flow treats `totalSize` as the exact final size, not just an initiate-time bound: `InitiateUploadV2IfRevision()` derives `partSize`/`partsTotal`, reserves quota, and stores `upload.TotalSize` from it (`pkg/backend/upload.go:362-474`), and `ConfirmUploadV2WithTags()` later requires `len(clientParts) == upload.PartsTotal` plus exact part sizes from `upload.TotalSize` (`pkg/backend/upload.go:736-795`). So a 1 GiB sequential write against a server advertising 20 GiB will still initiate as a 20 GiB upload, reserve 20 GiB, and then fail complete because the client only uploads the ~128 parts that actually exist. I think we need an explicit "unknown final size" contract here, or we need to defer initiate until we know the real total size.

pkg/fuse/stream_upload.go
```diff
@@ -3,13 +3,18 @@ package fuse
...
 	"time"
 
 	"github.com/mem9-ai/dat9/pkg/client"
 )
 
+// streamingInitiateUpperBound is the totalSize declared to the server when a
+// streaming upload starts before the final file size is known. The server
+// validates this against DRIVE9_MAX_UPLOAD_BYTES, so the value must not exceed
+// the server's configured cap.
+const streamingInitiateUpperBound int64 = 20 << 30 // 20 GiB
```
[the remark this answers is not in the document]
qiffang: Fixed in b0342f7. The root cause was deeper than the comment suggested — the misleading comment and hardcoded upper bound were symptoms of the real issue: the server treats `totalSize` as exact (validates `len(clientParts) == PartsTotal` at confirm time), so no upper bound approach can work.

New approach: `SubmitPart` now buffers parts locally instead of eagerly initiating. `FinishStreaming` initiates with the actual total size, uploads all buffered parts, and completes. The `streamingInitiateFallbackUpperBound` constant is removed entirely.

Tradeoff: parts stay in memory until flush (no streaming eviction savings). For proper streaming, the server would need an "unknown final size" contract at initiate time — filed as a future improvement.

pkg/fuse/stream_upload.go
qiffang: Spot on — this was the core issue. Fixed in b0342f7.

`SubmitPart` no longer initiates the server-side upload. It buffers part data locally. `FinishStreaming` then initiates with the actual `totalSize`, uploads all buffered parts, and completes — so `PartsTotal`, part sizes, and `len(clientParts)` all agree.

The tradeoff is that parts stay in memory until flush (no eviction savings during Write). A proper streaming solution would need the server to support "unknown final size" at initiate, or a resize/replan API before confirm.

PR #349, opened 2026-04-28
fix: classify canceled auth as client canceled

qiffang:
## Summary
- classify tenant-auth context cancellations as client-canceled/499-style events
- avoid logging canceled auth as auth_backend_unavailable or backend load failures
- add regression coverage for canceled auth logging

## Validation
- /usr/local/go/bin/go test -c ./pkg/server
- /usr/local/go/bin/go test -c ./pkg/fuse

Note: runtime pkg/server test is blocked locally by testcontainers/rootless Docker on this host.

## Summary by CodeRabbit

* **New Features**
  * Authentication now detects canceled client requests and responds with a client-closed status, with clearer logging and auth metrics.

* **Bug Fixes**
  * Lookup/retry logic updated so canceled-client responses are treated as transient, improving reliability of filesystem lookups under client disconnects.

* **Tests**
  * Added tests covering canceled-request handling and the transient behavior for lookup retries.

### DIFF
--- a/pkg/server/auth.go
+++ b/pkg/server/auth.go
@@ -29,6 +29,8 @@ type TenantScope struct {
 	Backend      *backend.Dat9Backend
 }
 
+const statusClientClosedRequest = 499
+
 func ScopeFromContext(ctx context.Context) *TenantScope {
 	s, _ := ctx.Value(tenantScopeKey).(*TenantScope)
 	return s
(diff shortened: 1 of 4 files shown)


pkg/server/auth.go
```diff
@@ -42,6 +44,19 @@ func authPhaseMs(start time.Time) float64 {
...
+	return errors.Is(err, context.Canceled) || errors.Is(ctx.Err(), context.Canceled)
+}
+
+func logAuthClientCanceled(ctx context.Context, event string, kv ...any) {
+	logger.Info(ctx, "server_event", eventFields(ctx, event, kv...)...)
+	metricEvent(ctx, "auth", "result", "client_canceled")
+}
+
+func writeClientCanceled(w http.ResponseWriter) {
+	w.WriteHeader(statusClientClosedRequest)
```
qiffang, line 57: This introduces 499 as an observable server response, but the FUSE client still treats a `client.StatusError{StatusCode: 499}` as the default `EIO` path. In the race where auth manages to write this response before the local HTTP request observes `context.Canceled`, a canceled read/lookup can surface as an I/O error instead of the existing retryable cancellation path. Please add 499 to the transient/canceled mapping (`httpToFuseStatus` and `isTransientLookupErr`, plus a small test) or keep this server-side change purely for logging without returning an unmapped status.

qiffang: Addressed in b0aeb7e42b678a5cac47004b3c30aa4ed9c32e17: added 499 to both `httpToFuseStatus` (typed StatusError switch + string-fallback path) and `isTransientLookupErr`, so a canceled-mid-auth 499 now follows the same retryable EAGAIN path as context.Canceled and 5xx instead of falling through to EIO. Two regression tests cover the mapping (TestHTTPToFuseStatus_MapsClientClosedRequestToEAGAIN, TestIsTransientLookupErr_Treats499AsTransient) — both pass against the new commit. The 499 constant lives client-side as `statusClientClosedRequest` to keep the contract explicit without taking a server package dependency.

qiffang (commented):
Re-reviewed latest head `b0aeb7e`. The earlier 499 concern is covered now: `httpToFuseStatus` and `isTransientLookupErr` both classify `client.StatusError{StatusCode: 499}` as retryable `EAGAIN`, and the string fallback is covered too. I do not see additional blocking issues in this PR.

Verified locally:

```bash
go test ./pkg/server ./pkg/fuse
```

PR #352, opened 2026-04-28
Handle non-retryable audio extract errors

JaySon-Huang:
## Summary

This PR improves audio extraction reliability and observability for Drive9 semantic tasks. Also address some comments on https://github.com/mem9-ai/drive9/pull/342

- Introduce structured audio extraction API errors and classify Qwen ASR `400 Bad Request` and `403 Forbidden` failures as non-retryable.
- Dead-letter non-retryable `audio_extract_text` semantic tasks immediately instead of spending all retry attempts on requests that cannot succeed without changing the input or provider/account configuration.
- Harden Qwen ASR handling by enforcing the provider's base64 data URL size limit, increasing bounded response reads to support larger valid responses, surfacing provider error messages, and logging masked debug response bodies.
- Add `DRIVE9_LOG_LEVEL` for configurable server log verbosity.
- Expose `resource_id` in `drive9 fs stat` metadata so files and directories can be correlated with semantic worker logs.

## Testing

- `make lint`
- `make build`
- `make test TEST_P=1`

## Summary by CodeRabbit

* **New Features**
  * Added `DRIVE9_LOG_LEVEL` environment variable to control logging verbosity (debug, info, warn, error).
  * File and directory stat responses now include a `resource_id` field for identification.
  * Added debug-level logging capability.

* **Bug Fixes**
  * Audio extraction now properly handles non-retryable errors by terminating tasks instead of retrying.
  * Improved error messages and handling for audio extraction failures.

* **Changes**
  * Reduced default maximum audio file size limit for qwen-asr mode from 10 MiB to 7 MiB.

### DIFF
--- a/cmd/drive9-server-local/main.go
+++ b/cmd/drive9-server-local/main.go
@@ -333,6 +333,7 @@ environment:
   DRIVE9_LOCAL_EMBEDDING_MODE auto|app|detect (default: auto when initing schema, detect otherwise)
   DRIVE9_LOCAL_META_DSN  local control-plane MySQL DSN for central quota (optional; enables server-mode quota enforcement)
   DRIVE9_VAULT_MASTER_KEY 32-byte hex key for vault DEK wrapping (omit to disable vault)
+  DRIVE9_LOG_LEVEL debug|info|warn|error (default: info)
   DRIVE9_BENCH_TIMING_LOG_ENABLED true|false to emit benchmark timing logs on successful server hot paths (default: false)
 
   S3 storage:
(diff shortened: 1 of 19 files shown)


qiffang:
Reviewed the changed paths in this PR end to end (`pkg/backend/audio_extract*.go`, `pkg/server/semantic_worker.go`, `pkg/datastore/semantic_tasks.go`, `pkg/server/server.go`, and the logger/stat wiring).

I do not see a blocking correctness issue here.

Why this looks safe to merge:
- non-retryable Qwen ASR failures now dead-letter instead of burning through the semantic retry budget, and the worker/datastore transition is still lease-guarded and atomic
- the new request-size guard for Qwen ASR matches the provider’s base64 data-URL limit and avoids pointless upstream calls
- `resource_id` is added in a backward-compatible way on the stat metadata API and is wired through server/client/CLI consistently
- the server log-level override is opt-in and validated, so existing deployments stay on the current default behavior
- the tests cover the important new branches: oversized payload rejection, retryable vs non-retryable API errors, dead-letter transition, and stat metadata/resource ID exposure

One non-blocking follow-up I’d keep in mind: the response masking currently targets `choices[].message.content`, which is the important path for transcript leakage today. If Qwen ever starts echoing transcript-like text in other fields, we may want to broaden that sanitizer rather than preserve unknown top-level fields by default.

Net: no blockers from my side.

PR #380, opened 2026-05-06
fix: preserve FUSE lockfile namespace during chmod

qiffang:
## Summary
- preserve regular-file inode mappings while open, pending, or queued local state still represents the path
- resolve open dirty handles during Lookup before falling back to remote Stat
- add commit-queue cleanup hook to remove forgotten stale mappings only after local commit state is gone

## Tests
- /usr/local/go/bin/go test ./pkg/fuse/...
- /usr/local/go/bin/go test -c ./pkg/fuse ./cmd/drive9
- git diff --check

## Notes
This targets the Ubuntu FUSE git lock-file failure: create/write/close/chmod/rename on .git/config.lock can otherwise observe a false ENOENT when kernel Forget races local pending/commit state. Real Ubuntu FUSE git clone should still be used to verify the live path.

## Release Notes

* **New Features**
  * Improved inode lifecycle management for files with open handles or pending local state
  * Added commit queue cleanup callback support

* **Tests**
  * New test validating file lookup behavior and git compatibility after kernel forget operations

### DIFF
--- a/pkg/fuse/commit_queue.go
+++ b/pkg/fuse/commit_queue.go
@@ -37,6 +37,10 @@ type CommitEntry struct {
 // 0 for multipart where the revision is not returned inline).
 type CommitSuccessFunc func(entry *CommitEntry, committedRev int64)
 
+// CommitCleanupFunc is called after a successful commit's local shadow/index
+// state has been removed but before the queue entry is dequeued.
+type CommitCleanupFunc func(entry *CommitEntry)
+
 // CommitQueue manages ordered background remote commits with baseRev tracking.
 // It provides backpressure when the queue exceeds maxPending items.
 type CommitQueue struct {
@@ -56,6 +60,9 @@ type CommitQueue struct {
 	// OnSuccess is called after successful upload with the committed
 	// revision. Used by dat9fs to seed readCache and update inode revision.
 	OnSuccess CommitSuccessFunc
+
+	// OnCleanup is called after local commit state has been removed.
+	OnCleanup CommitCleanupFunc
 
 	// workCh dispatches entries to upload workers. The buffer is always
 	// larger than maxPending so Enqueue never blocks.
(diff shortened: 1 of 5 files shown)


qiffang (commented):
## Code Review

### Bug — Release duplicates `fileHandles.Delete`

The PR adds a `defer` at the top of `Release` that calls `fs.fileHandles.Delete(input.Fh)` + `cleanupReleasedInode`, but the existing code already calls `fs.fileHandles.Delete(input.Fh)` at **three** exit points (lines 2825, 2900, 2949). This causes double-Delete on every Release path. While the second Delete is a no-op, the duplication is confusing and should be cleaned up.

**Fix**: Remove the existing three `fs.fileHandles.Delete(input.Fh)` calls and consolidate in the defer. Also consolidate the `fh.Prefetch.Close()` calls.

### Performance note — `hasOpenHandle` does full table scan

`hasOpenHandle` calls `fs.fileHandles.Snapshot()` (full table copy + iterate) on every `Forget` and `Release`. For git clone with hundreds of concurrent open handles, this is O(N) per call. Acceptable for now but worth a TODO for path-indexed lookup.

### Missing test

Add a test verifying `cleanupReleasedInode` actually removes the inode mapping after Release completes (proving no leak for Nlookup=0 inodes after the last handle is closed).

### Everything else looks good

- `ForgetKeepMapping` / `RemoveFileIfUnreferenced` — clean, correct
- `CommitQueue.HasPath` — proper lock scope
- `OnCleanup` hook placement — correct (after shadow/index Remove, before dequeue)
- Tests cover the three key scenarios well

Fix the Release duplicate-Delete issue and add the cleanup test, then this is ready to approve.

qiffang:
CI all green (ci ✅, e2e ✅, CodeRabbit ✅). Code review items from my earlier review are all addressed. Ready for @qiffang to merge after Ubuntu FUSE git clone validation.

PR #384, opened 2026-05-06
Fix stale FUSE rename dentries for lockfiles

qiffang:
## Summary
- invalidate source/target FUSE dentries after successful rename
- keep create/write/release paths free of kernel notify calls to avoid the previous git-clone EAGAIN flood
- update the FUSE notify-budget test to allow only bounded rename entry invalidations

## Context
After #383, POSIX overwrite rename works, but git can immediately create `.git/config.lock` again. Linux may still have the old lockfile dentry cached, so `O_CREAT|O_EXCL` can fail with false `EEXIST`.

## Tests
- go test ./pkg/fuse

## Release Notes

  * Optimized lock file caching behavior to reduce unnecessary server notifications.

* **Bug Fixes**
  * Improved rename operation handling for edge cases involving identical source and destination paths.

* **Tests**
  * Enhanced test coverage for lock file caching and high-volume local mutation operations.

### DIFF
--- a/pkg/fuse/dat9fs.go
+++ b/pkg/fuse/dat9fs.go
@@ -68,8 +68,8 @@ type Dat9FS struct {
 	// server is the go-fuse server, set during Init(). Used to send
 	// kernel cache invalidation notifications (EntryNotify, InodeNotify)
 	// for external/SSE-driven changes that the kernel doesn't know about.
-	// Local FUSE mutations do NOT use server notify — the kernel updates
-	// its own cache from the FUSE reply.
+	// Local FUSE mutations avoid server notify because handlers can be running
+	// on the same worker pool that services notify-triggered revalidation.
 	server *gofuse.Server
 
 	// notifyWg tracks inflight asynchronous kernel notification goroutines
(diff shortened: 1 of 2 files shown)


qiffang:
Good catch. The async notify did not provide the ordering this fix needs. I updated the PR so rename uses a new `notifyEntrySync` helper for the old and new dentries, while SSE/external notifications still use the existing async `notifyEntry` path. The sync helper is called only after remote/local rename bookkeeping is complete and without holding our inode/handle locks; this follows go-fuse `EntryNotify` guidance and gives the next `O_CREAT|O_EXCL` a real ordering guarantee. Re-ran `go test ./pkg/fuse`.

qiffang:
Addressed both points.

- The rename test now records actual notify attempts via a test hook and asserts the exact pairs. For the git lockfile case it requires synchronous entry invalidation of `(parent=1, name="config.lock")` and `(parent=1, name="config")`, so target-only, wrong-parent, or unrelated-name implementations fail.
- I also revised `TestMutationHandlers_CompleteWithinTimeout` wording. It now explicitly says the test uses a nil go-fuse server and only covers handler bookkeeping, not the real `EntryNotify` syscall path.

Re-ran `go test ./pkg/fuse`.

qiffang:
Update after the CI FUSE smoke hang at `[8] rename semantics`:

- Removed the synchronous `EntryNotify` from the local `Rename` path. The smoke log stops immediately before `mv`, and calling `EntryNotify` from inside the FUSE rename handler can block the worker path.
- Switched the stale-lockfile fix to disable positive dentry caching for `*.lock` entries by returning `EntryTimeout=0` from `Lookup` and `Create`. That makes the next `O_CREAT|O_EXCL` revalidate `config.lock` without depending on notify timing.
- Restored the local mutation notify budget to zero notify calls and added direct tests for lockfile entry TTL.

Verification: `go test ./pkg/fuse`.

qiffang:
Follow-up: the refreshed checks are green on commit `8b9d354`:

- `ci`: pass
- `drive9-server-local e2e smoke`: pass
- FUSE smoke step completed successfully, so it no longer hangs at `[8] rename semantics`.

PR #386, opened 2026-05-06
Fix FUSE lockfile unlink after async commit

qiffang:
## Summary

Fix the git clone `config.lock` failure reproduced by the mount log:

- `Release(config.lock)` enqueues a PendingNew commit and the commit queue uploads it.
- `Unlink(config.lock)` previously still treated the file as local-only PendingNew and skipped remote DELETE.
- A later `LOOKUP config.lock` then saw the real remote file, so git failed `O_CREAT|O_EXCL` with `File exists`.

This PR makes Unlink issue a remote DELETE for PendingNew files when commitQueue mode is active. DELETE still tolerates 404, so local-only files remain safe.

Also invalidate the destination read cache on rename overwrite; the provided log showed `config` size advancing while reads still returned the older 36-byte cached content.

## Tests

- `go test ./pkg/fuse`
- `go test ./pkg/fuse -run 'TestUnlink_PendingNewCommitQueueUploadedDeletesRemote|TestRenameInvalidatesDestinationReadCache'`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability of file deletion operations to prevent deleted files from being unexpectedly recovered in edge cases with pending uploads
  * Enhanced file rename operations to ensure system caches properly reflect changes, especially when overwriting existing files

### DIFF
--- a/pkg/fuse/dat9fs.go
+++ b/pkg/fuse/dat9fs.go
@@ -1372,12 +1372,6 @@ func (fs *Dat9FS) Unlink(cancel <-chan struct{}, header *gofuse.InHeader, name s
 		// Remove pending cache entry to prevent future background uploads.
 		fs.writeBack.Remove(childP)
 	}
-	// Also check pendingIndex for the pending-new flag before clearing.
-	if !pendingNew && fs.pendingIndex != nil {
-		if meta, ok := fs.pendingIndex.GetMeta(childP); ok && meta.Kind == PendingNew {
-			pendingNew = true
-		}
-	}
 	// Wait for any in-flight commitQueue upload and cancel it so the
 	// background commit cannot resurrect the deleted file.
 	if fs.commitQueue != nil {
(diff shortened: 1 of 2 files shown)


qiffang:
Updated per review. The Unlink path now preserves the pendingIndex signal in commitQueue mode:

1. wait for `commitQueue.WaitPath(path)`
2. re-check `pendingIndex.GetMeta(path)`
3. then call `commitQueue.CancelPath(path)`

So:

- uploaded by commitQueue -> pendingIndex was already cleared by `onCommitSuccess` -> remote DELETE is sent
- PendingNew never enqueued/uploaded -> pendingIndex still has PendingNew -> remote DELETE is skipped

Added the required regression test: `TestUnlink_PendingNewCommitQueueNotEnqueuedSkipsRemoteDelete`. Existing uploaded case remains covered by `TestUnlink_PendingNewCommitQueueUploadedDeletesRemote`.

Verification:

- `go test ./pkg/fuse`
- `go test ./pkg/fuse -run 'TestUnlink_PendingNewCommitQueueUploadedDeletesRemote|TestUnlink_PendingNewCommitQueueNotEnqueuedSkipsRemoteDelete|TestRenameInvalidatesDestinationReadCache'`

PR #392, opened 2026-05-06
fix(fuse): retry interrupted remote renames

qiffang:
## Summary
- retry remote RenameCtx with a bounded detached context after FUSE interrupt/transient failure
- treat target visibility after an interrupted rename as success so git loose object finalization does not see EAGAIN
- add regression tests for retry success and already-committed interrupted rename

## Tests
- go test ./pkg/fuse -run 'TestRenameRemoteWithTransientRetry|TestRenamePendingNewCommit' -count=1
- go test ./pkg/fuse -count=1

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved file rename operations with enhanced retry logic to automatically recover from transient network failures and temporary service interruptions.

### DIFF
--- a/pkg/fuse/dat9fs.go
+++ b/pkg/fuse/dat9fs.go
@@ -952,6 +952,20 @@ func (fs *Dat9FS) remoteDirExistsDetached(localPath string) bool {
 	return err == nil && stat.IsDir
 }
 
+func (fs *Dat9FS) remoteRenameCommittedDetached(oldP, newP string) bool {
+	targetCtx, targetCancel := context.WithTimeout(context.Background(), namespaceMutationRetryTimeout)
+	_, targetErr := fs.client.StatCtx(targetCtx, fs.remotePath(newP))
+	targetCancel()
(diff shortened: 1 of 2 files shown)


qiffang:
Fixed the overwrite-rename correctness issue.

Changes:
- Replaced the unsafe target-visible confirmation with target-visible + source-gone confirmation.
- If both source and target still exist after an interrupted server-side rename, we keep retrying instead of returning OK.
- Added regression coverage for preexisting target overwrite rename: config.lock -> config is interrupted before commit, detached stats see both paths, and the code must retry before succeeding.

Validation:
- go test ./pkg/fuse -run 'TestRenameRemoteWithTransientRetry|TestRenamePendingNewCommit|TestRenameInvalidatesDestinationReadCache' -count=1
- go test ./pkg/fuse -count=1
- go test -c ./pkg/fuse ./cmd/drive9
- git diff --check origin/main...HEAD

@coderabbitai review

PR #407, opened 2026-05-07
perf(fuse): optimize small file read/write performance

srstack:
## Summary

This PR optimizes FUSE small file (< 50KB) read/write performance through two key changes:

### 1. WriteBuffer small-file fast path
- Uses a single contiguous `[]byte` allocation for files below `smallFileThreshold` (50KB)
- Avoids sparse part map lookups, per-part allocations, and memory fragmentation
- Automatically migrates to standard part mode when writes exceed the threshold
- Updated `ShadowStore.WriteExtents` to leverage the contiguous data path

### 2. ReadCache zero-copy reads
- Removed defensive copy (`make` + `copy`) from `ReadCache.Get()`
- FUSE `Read` treats cached data as read-only; the kernel copies it via `ReadResultData`
- Eliminates one full data copy per cache hit

### Benchmarks

```
BenchmarkWriteBuffer_SmallFile_ReadAt-12     34113346    34.33 ns/op    0 B/op    0 allocs/op
BenchmarkReadCache_GetPut-12                   489373  2434 ns/op      0 B/op    0 allocs/op
```

- `ReadCache_GetPut`: **0 allocations** (was copying 50KB per get)
- `WriteBuffer_SmallFile_ReadAt`: **0 allocations**, ~34ns per call

### Files changed
- `pkg/fuse/write.go` — small-file fast path in WriteBuffer
- `pkg/fuse/read.go` — zero-copy ReadCache.Get()
- `pkg/fuse/shadow.go` — fast path in WriteExtents
- `pkg/fuse/fuse_test.go` — benchmarks

## Summary by CodeRabbit

  * Fast-path handling for small-file writes/reads/truncates for faster small-file operations.
  * Read-cache returns cached data without extra copying to reduce allocations and improve read efficiency.

* **Bug Fixes**
  * Sparse-write/truncate correctness: gaps after truncation are zero-filled while preserving existing prefix and new tail data.

* **Tests**
  * Added correctness tests for small-file/sparse behaviors and multiple benchmarks measuring small-file and cache performance.

### DIFF
--- a/pkg/fuse/fuse_test.go
+++ b/pkg/fuse/fuse_test.go
@@ -375,6 +375,34 @@ func TestWriteBuffer_Truncate(t *testing.T) {
 	}
 }
 
+func TestWriteBuffer_TruncateSmallFileSparseWriteZeroFillsGap(t *testing.T) {
+	wb := NewWriteBuffer("/test", 0, 0)
+	_, _ = wb.Write(0, []byte("hello world"))
+
+	if err := wb.Truncate(5); err != nil {
+		t.Fatal(err)
+	}
+	if _, err := wb.Write(10, []byte("X")); err != nil {
+		t.Fatal(err)
+	}
+
+	got := wb.Bytes()
+	if len(got) != 11 {
+		t.Fatalf("Bytes() len = %d, want 11", len(got))
+	}
+	if string(got[:5]) != "hello" {
+		t.Fatalf("prefix = %q, want %q", got[:5], "hello")
+	}
+	for i := 5; i < 10; i++ {
+		if got[i] != 0 {
+			t.Fatalf("gap byte %d = %d, want 0", i, got[i])
+		}
+	}
+	if got[10] != 'X' {
+		t.Fatalf("tail byte = %q, want %q", got[10], byte('X'))
+	}
+}
+
 func TestWriteBuffer_EFBIG(t *testing.T) {
 	wb := NewWriteBuffer("/test", 100, 0)
 	_, err := wb.Write(0, make([]byte, 101))
@@ -986,6 +1014,28 @@ func TestWriteBuffer_ReadAt_Basic(t *testing.T) {
 	}
 	if string(buf[20:30]) != "CCCCCCCCCC" {
 		t.Fatalf("part 3: got %q", buf[20:30])
+	}
+}
+
+func TestWriteBuffer_SmallFileFastPathRequiresSinglePart(t *testing.T) {
+	wb := NewWriteBuffer("/test", 0, 10)
+	data := []byte("0123456789012345678901234")
+	if _, err := wb.Write(0, data); err != nil {
+		t.Fatal(err)
+	}
+
+	dirty := wb.DirtyPartNumbers()
+	if len(dirty) != 3 {
+		t.Fatalf("dirty parts = %v, want 3 parts", dirty)
+	}
+	if got := string(wb.PartData(1)); got != "0123456789" {
+		t.Fatalf("part 1 = %q, want %q", got, "0123456789")
+	}
+	if got := string(wb.PartData(2)); got != "0123456789" {
+		t.Fatalf("part 2 = %q, want %q", got, "0123456789")
+	}
+	if got := string(wb.PartData(3)); got != "01234" {
+		t.Fatalf("part 3 = %q, want %q", got, "01234")
 	}
 }
 
--- a/pkg/fuse/fuse_test.go
+++ b/pkg/fuse/fuse_test.go
@@ -12,6 +12,12 @@ import (
 	"github.com/mem9-ai/dat9/pkg/client"
 )
 
+var (
+	benchmarkBytesSink []byte
+	benchmarkByteSink  byte
+	benchmarkIntSink   int
+)
+
 // ---------------------------------------------------------------------------
 // InodeToPath tests
 // ---------------------------------------------------------------------------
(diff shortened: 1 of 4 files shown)


srstack:
Addressed the new review feedback.

Changes pushed:
- kept the WriteBuffer small-file hardening in place for gap zero-filling
- updated the ReadCache threshold comment to 50KB
- hardened the new benchmarks with package-level sinks to avoid compiler-elided work

Validated locally with targeted tests and short benchmark runs.

srstack:
Addressed the latest review feedback.

Changes pushed:
- preserved the WriteBuffer small-file invariant so len(smallFileData) always covers the logical size after truncate-extend followed by small writes
- added a regression test for truncate-extend plus small write/read/shrink
- also switched the truncate grow zero-fill branch to clear(...) for consistency

Validated locally with targeted WriteBuffer tests.

qiffang:
## @dat9-dev1 Review: Small File Read/Write Perf — ✅ APPROVE

### Architecture

Two clean optimizations:
1. **WriteBuffer small-file fast path**: Uses single contiguous `[]byte` for files < 50KB instead of sparse part map. Avoids map lookups and per-part allocations.
2. **ReadCache zero-copy**: Removes defensive `make + copy` in `Get()`. FUSE Read treats cached data as read-only via `ReadResultData`.

### Correctness Verified

- **Entry condition** is strict: `partSize >= smallFileThreshold && len(parts) == 0 && LoadPart == nil && end <= threshold && totalSize <= threshold` — excludes lazy loading and existing parts.
- **Migration** (`migrateToPartMode`): copies smallFileData → `parts[0]`, sets `curMemory`, clears `smallFileData`, then recursive `Write()` in normal path. No data loss.
- **Truncate**: shrink re-slices; extend beyond threshold migrates + recurses. Handles shrink→sparse-write→extend correctly.
- **Dirty tracking**: `dirtyParts[0] = true` ↔ `DirtyPartNumbers()` 1-based `[1]` ↔ `PartData(1)` — consistent mapping.
- **`ClearDirty` preserves `smallFileData`**: data valid for reads until Reset/release.
- **`EvictPart` migrates first**: prevents parts/smallFileData inconsistency.
- **ShadowStore.WriteExtents**: fast path clips to `totalSize`, single `pwrite` at offset 0.

1. **`ReadAt` small-file path uses `buf` not `buf[:total]`**: Currently safe because `len(smallFileData) == totalSize` invariant holds, but more defensive to use `copy(buf[:total], smallFileData[offset:offset+int64(total)])`.

2. **`curMemory` is 0 in small-file mode**: Small-file data lives in `smallFileData` not in `parts`, so `curMemory` understates actual memory. No external code depends on this today, but worth a comment.

3. **Test naming**: `TestWriteBuffer_SmallFileFastPathRequiresSinglePart` actually tests that the fast path is *skipped* when `partSize < smallFileThreshold` (partSize=10). Name could be `SkipsFastPathWhenPartSizeSmall`.

### Tests

Good coverage: truncate-sparse-write-zero-fill, extend-shrink-logical-size, fast-path-skip, benchmarks with proper sinks.

qiffang (approved):
## PR #407 Review — GREEN

Reviewed the small-file FUSE read/write optimization. No blocking findings.

### What I checked

- `WriteBuffer` small-file fast path only activates for pure in-memory new/truncated small files: `partSize >= smallFileThreshold`, no existing parts, no lazy `LoadPart`, `end <= smallFileThreshold`, and current logical size still within threshold.
- Migration to part mode copies `smallFileData` into `parts[0]`, preserves dirty state, and then re-enters the normal sparse-part write/truncate path.
- Truncate/gap semantics preserve logical size and zero-fill gaps for small-file mode.
- `ShadowStore.WriteExtents` clips the contiguous buffer to `totalSize`, writes once, then truncates the shadow file to exact logical size.
- `ReadCache.Get` zero-copy is safe for current production callsites: `Put` still copies inputs, returned cache slices are treated read-only, and updates/invalidations replace/remove entries without mutating old backing slices.

### Local validation

Passed:

- `go test ./pkg/fuse -run 'TestWriteBuffer|TestShadowStore|TestReadCache|TestFuseRead|TestWriteBack|TestFlush|TestFsync|TestRelease' -count=1`
- `go test -c ./pkg/fuse ./cmd/drive9 ./cmd/drive9/cli ./pkg/client ./pkg/server`
- `go test -race ./pkg/fuse -run 'TestReadCache|TestWriteBuffer|TestShadowStoreWriteExtents' -count=1`

GitHub checks are also green: `ci` and `drive9-server-local e2e smoke` succeeded.

### Non-blocking recommendations

1. In the small-file `ReadAt` branch, consider slicing explicitly to the computed logical read range:
   `copy(buf[:total], wb.smallFileData[offset:offset+int64(total)])`.
   Current behavior is correct because `len(smallFileData) == totalSize` is maintained, but the explicit bounds make the invariant harder to accidentally break later.

2. `curMemory` intentionally remains `0` while data is held in `smallFileData`. That is fine today because no memory budget depends on it in small-file mode, but add a comment if future memory-pressure logic starts using `curMemory` as total resident bytes.

3. `TestWriteBuffer_SmallFileFastPathRequiresSinglePart` is correct, but the name is slightly indirect. It is really testing that fast path is skipped when `partSize < smallFileThreshold`.

Verdict: APPROVE.

srstack:
Addressed the follow-up small-file review recommendations.

Changes pushed:
- clamped the small-file ReadAt fast path to the computed logical read range instead of relying on the implicit len(smallFileData) == totalSize invariant
- routed ShadowStore small-file writes through a WriteBuffer accessor instead of touching internal fields directly
- clarified in ReadCache.Get that returned zero-copy slices remain valid after eviction because the backing array stays reachable until callers drop the slice
- added a regression test covering small-file ReadAt at EOF boundaries

Validated locally with targeted pkg/fuse WriteBuffer tests.

PR #408, opened 2026-05-08
Add durable file GC tasks

qiffang:
Adds durable `file_gc_tasks` cleanup for deleted file identities. Metadata delete paths now enqueue GC tasks in the same transaction, and the backend worker removes central file quota metadata and S3 blobs with lease-based retry.

Scope boundary: this PR covers deleted file identities only. It intentionally does not make overwrite old-blob cleanup durable for a still-live `file_id`; that remains the existing best-effort `deleteBlobIfS3Ctx` path and should be handled by a separate blob-ref GC task if we want durable overwrite cleanup.

Review follow-ups addressed:
- Removed the unused `syncCentralFileDelete` definition; `fileDeleteMutationData` stays only for replaying historical `file_delete` mutation-log entries.
- Removed the GC worker's path/adapter fallback. Central file metadata deletion now requires `DeleteFileMetaIfExistsTx`, so quota counters are decremented only when that transaction actually deleted a row.
- Made historical `file_delete` replay use the same conditional delete guard.
- Documented the deleted-file-identity scope in code comments.

Tests:
- `go test ./pkg/backend -run 'TestFileGCTask|TestCentralQuotaFileMutationLifecycle|TestMutationReplay|TestMonthlyLLMCostExceeded' -count=1`
- `go test ./pkg/datastore -run 'TestFileGCTask|TestDeleteWithRefCount|TestDeleteDirRecursive|TestRenameFileReplacingTarget' -count=1`
- `go test ./pkg/datastore ./pkg/backend ./pkg/meta -count=1`
- `go test ./pkg/fuse ./pkg/client ./pkg/tenant/schema ./pkg/tenant ./cmd/drive9-server ./cmd/drive9-server-local -count=1`
- `go vet ./pkg/backend ./pkg/datastore ./pkg/meta ./pkg/fuse ./pkg/client ./pkg/tenant/schema ./pkg/tenant ./cmd/drive9-server ./cmd/drive9-server-local`
- `git diff --check`

Note: local `make lint` did not reach lint findings because `golangci-lint` 2.5.0 failed during package loading with `no go files to analyze`; CI will run lint again on GitHub.


qiffang:
## @dat9-dev1 Review: Durable File GC Tasks — ✅ APPROVE

### Architecture

Clean migration from inline sync delete (`syncCentralFileDelete` + `deleteBlobCtx`) to durable async task queue (`file_gc_tasks` table). Key design decisions are correct:

1. **Task keyed by file_id, not path** — prevents retry from deleting a different file's blob at the same path. `uk_file_gc_file_id` unique index + `isUniqueViolation` silent dedup.
2. **Central meta delete before blob delete** — `DeleteFileMetaIfExistsTx` returns `(false, nil)` on re-delete, making quota counter updates idempotent across retries.
3. **Pending mutation guard** — `checkNoPendingCentralFileMutation` / `HasPendingFileMutation` prevents GC from acking central cleanup before fail-open create/overwrite mutations have converged. Correct happens-before ordering.
4. **`stopFileGCWorker` placed before `Close()` early return** — GC worker is properly stopped even when imageExtractCancel is nil.
5. **`defaultFileGCMaxAttempts = 0` (infinite retry)** — correct for correctness cleanup path; blob leak is worse than dead letter.

1. `RecoverExpiredFileGCTasks`: `LIMIT N FOR UPDATE SKIP LOCKED` may recover fewer than N rows per round (locked rows are skipped). Not a bug but worth a comment.
2. `processFileGCTask`: only handles `StorageS3` — correct today (inline blobs deleted with the row), but add a comment explaining why other storage types are a no-op.
3. `scanFileForGC` intentionally skips `content_blob`/`description` — good optimization for the delete TX path.
4. `HasPendingFileMutation` uses `JSON_EXTRACT` without index — fine while `quota_mutation_log` pending set stays small, but worth monitoring.

### Tests

4 backend integration tests + 3 datastore unit tests cover: happy path, blob retry with quota release, pending create mutation blocking, pending overwrite mutation blocking. Good coverage.

qiffang:
## @dat9-dev1 Review Update: REVISE

Updating my earlier APPROVE to **REVISE** after @adversary-1 and @adversary-2 identified valid blockers:

**B0 (CI lint)**: `syncCentralFileDelete` at `quota_mutation.go:158` has no callers after this PR removes the 3 callsites in `dat9.go`. Needs to be deleted (along with `fileDeleteMutationData` if orphaned too).

**B1 (overwrite old blob not durable)**: `dat9.go:626` `deleteBlobIfS3Ctx` for overwrite is still inline best-effort. Suggest: scope this PR explicitly to "deleted file identity GC only" and open a separate issue for overwrite blob durability. The `UNIQUE(file_id)` constraint can't track multiple old `storage_ref` values from successive overwrites anyway. Add a comment documenting this scope boundary.

**B2 (fallback TOCTOU double-decrement)**: Fallback path `GetFileMeta` (outside TX) → `DeleteFileMetaTx` (doesn't check rows affected) → `applyCentralFileDeleteCountersTx` allows two concurrent workers to both decrement quota. Fix: use `DeleteFileMetaIfExistsTx` boolean in fallback too, or remove fallback entirely.

qiffang:
## @dat9-dev1 Re-review v2: Durable File GC Tasks — ✅ APPROVE

All 3 blockers from v1 are fixed:

**B0 (CI lint)**: `syncCentralFileDelete` removed. `fileDeleteMutationData` retained with comment for historical mutation replay.

**B1 (overwrite scope)**: Two explicit scope boundary comments added:
- `dat9.go:614-615`: "Overwrite cleanup remains best-effort: file_gc_tasks track deleted file identities, not old blob refs for a still-live file_id."
- `FileGCTask` struct doc: "...does not track old blob refs produced by overwriting a still-live file_id; those need a separate blob-ref GC path."

**B2 (fallback TOCTOU)**: Thoroughly fixed:
- Removed `fileMetaConditionalDeleter` interface and dual-path logic. Single path: `GetFileMeta` → `DeleteFileMetaIfExistsTx(bool)` → decrement counters only when `deleted==true`.
- `DeleteFileMetaIfExistsTx` and `HasPendingFileMutation` promoted to `MetaQuotaStore` interface (no more type assertions).
- Bonus: `mutation_replay.go` `file_delete` also uses `DeleteFileMetaIfExistsTx` + boolean guard for consistency.

CI: ci ✅ | e2e smoke ✅

PR #409, opened 2026-05-08
Optimize FUSE small-file hot paths

srstack:
## Summary
Optimize FUSE small-file hot paths to reduce writable reopen/read overhead and keep pending write-back reopen fast without unbounded memory growth.

## What changed
- preload writable small-file handles from read cache, pending shadow state, and write-back cache before falling back to remote stat/get
- serve clean writable-handle reads from already loaded write buffer ranges instead of always reading remotely
- avoid repeated small-file materialization by adding a read-only write-buffer view and reusing owned slices when populating the read cache
- keep write-back metadata in memory and cache small pending payloads for hot reopen/read/uploader paths
- bound the in-memory write-back payload cache with a 32MB LRU budget to avoid unbounded resident memory growth
- add focused tests and benchmarks covering read-cache hits, pending shadow/write-back reopen, and small-file flush/read paths

## Validation
- go test ./pkg/fuse
- go test ./pkg/fuse -run 'TestWriteBackCache_GetUsesInMemorySmallDataCache|TestWriteBackCache_LargeDataNotCachedInMemory|TestWriteBackCache_SmallDataCacheBudgetEvictsLRU|TestOpenWritable_PreloadsFromWriteBackCache|TestOpenWritable_OverwriteFromWriteBackCacheSkipsRemoteStat|TestWriteBackCache_RenamePending'
- go test ./pkg/fuse -run '^$' -bench 'BenchmarkDat9FS_SmallFileOpen/pending-writeback-overwrite' -benchmem -count 5

## Benchmark note
The pending write-back overwrite reopen benchmark stayed around 9-12 us/op after adding the bounded LRU payload cache, preserving the hot-path win while capping memory use.

## Summary by CodeRabbit

* **Refactor**
  * Improved file caching and buffer handling for safer, more memory-efficient access
  * Faster small-file opens/reads with new fast-paths that reduce remote stat/read calls
  * Write-back cache revamped to an in-memory metadata index with LRU payload caching for better performance

* **Tests**
  * Added micro-benchmarks for small-file open/read/flush
  * Expanded tests for cache semantics, eviction behavior, preload fast-paths, and write-back persistence

### DIFF
--- a/pkg/fuse/dat9fs.go
+++ b/pkg/fuse/dat9fs.go
@@ -463,6 +463,26 @@ func (fs *Dat9FS) preloadWritableHandle(ctx context.Context, fh *FileHandle) gof
 	return gofuse.OK
 }
 
+func (fs *Dat9FS) preloadWritableHandleFromReadCacheLocked(fh *FileHandle) bool {
+	if fs.readCache == nil || fh == nil || fh.Dirty == nil {
+		return false
+	}
+	if fh.OrigSize <= 0 || fh.OrigSize > smallFileThreshold || fh.BaseRev <= 0 {
+		return false
+	}
+
+	data, ok := fs.readCache.Get(fh.Path, fh.BaseRev)
+	if !ok {
+		return false
+	}
+	if _, err := fh.Dirty.Write(0, data); err != nil {
+		log.Printf("read-cache writable preload failed for %s: %v", fh.Path, err)
+		return false
+	}
+	fh.Dirty.ClearDirty()
+	return true
+}
+
 func (fs *Dat9FS) pendingKindForHandle(fh *FileHandle) PendingKind {
 	if fh.IsNew {
 		return PendingNew
@@ -538,7 +558,7 @@ func (fs *Dat9FS) stageShadowLocked(fh *FileHandle, durable bool) error {
 			return err
 		}
 	} else {
-		if err := fs.shadowStore.WriteFull(fh.Path, fh.Dirty.Bytes(), fh.BaseRev); err != nil {
+		if err := fs.shadowStore.WriteFull(fh.Path, fh.Dirty.bytesView(), fh.BaseRev); err != nil {
 			return err
 		}
 		fh.ShadowReady = true
@@ -573,7 +593,7 @@ func (fs *Dat9FS) snapshotWriteBackLocked(fh *FileHandle) error {
 	}
 	return fs.writeBack.PutWithBaseRev(
 		fh.Path,
-		fh.Dirty.Bytes(),
+		fh.Dirty.bytesView(),
 		fh.Dirty.Size(),
 		fs.pendingKindForHandle(fh),
 		fh.BaseRev,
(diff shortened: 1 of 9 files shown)


srstack:
Addressed this round of review feedback and pushed commit 0a50d0d.

This update includes:
- preserving WriteBackSeq when preloading writable reopens from the write-back cache so unchanged reopen/close does not fall back to synchronous upload
- fixing the zero-length read edge case on the clean writable-buffer path by returning early with an empty result
- pruning stale write-back metadata immediately when the payload file is missing so pending state cannot get stuck
- skipping PendingConflict entries in ListPending so recovery does not resubmit terminal conflict entries
- adding regression coverage for overwrite preload reads, WriteBackSeq preservation, zero-length reads, missing payload cleanup, and conflict skipping

Validation:
- go test ./pkg/fuse -run 'TestWriteBackCache_GetViewPrunesMissingLargeData|TestWriteBackCache_ListPending_SkipsConflictEntries|TestOpenWritable_OverwriteFromWriteBackCacheSkipsRemoteStat|TestReadZeroLengthFromWritableCleanBuffer'
- go test ./pkg/fuse

qiffang:
## @dat9-dev1 Review: Small-file Hot Path Optimization — ✅ APPROVE

### Architecture

5 optimizations reducing remote calls and memory copies for writable reopen/read:
1. Open writable preloads from readCache / writeback cache / shadow before falling back to remote
2. Clean writable handle Read serves already-loaded parts from WriteBuffer
3. `bytesView()` returns read-only view for small files, avoiding `Bytes()` allocate+copy
4. `PutOwned()` zero-copy write into readCache (caller transfers ownership)
5. WriteBackCache adds in-memory metadata index + LRU-bounded 32MB payload cache

### Correctness Verified

- **`bytesView()` call sites**: All 4 are safe — `stageShadowLocked` (disk write), `snapshotWriteBackLocked` (PutWithBaseRev makes internal copy via cacheDataLocked), `flushHandleDebounced` (immediate make+copy), `flushHandle` (handle lock held during HTTP PUT).
- **`getView` concurrency**: Returns in-memory slice then releases `c.mu`. Safe because `cacheDataLocked` always creates new slices (make+copy); LRU eviction only deletes map entries, never mutates backing arrays.
- **Open preload ordering**: shadow → writeback → (BaseRev==0? remote stat) → readCache → remote full. Correct — readCache needs BaseRev for revision matching, placed after stat.
- **`dirty-clean-buffer` Read**: Checks `IsPartLoaded` for all parts in range; falls through to remote if any part is missing. Conservative and correct.
- **WriteBackCache metadata index**: `GetMeta` returns copy from memory (no disk I/O). `ListPending` still reconciles from disk for crash recovery. `nextGen` restored from max on-disk generation to prevent counter reuse.
- **LRU payload cache**: Bounded by `dataMaxBytes=32MB`, only caches payloads ≤ `smallFileThreshold`. `dataBytes` accounting consistent across all insert/delete/evict paths. `RenamePending` correctly migrates data cache entry.

1. `loadWritableHandleFromWriteBackLocked`: `fh.OrigSize = int64(len(data))` uses actual data length rather than `meta.Size`. If `.dat` file is truncated on disk, these could diverge. Low risk since in-memory cache path guarantees consistency.
2. `cacheDataLocked`: calls `touchDataLocked` after `PushFront` for new entries — redundant MoveToFront. No correctness impact.

### Tests

14 new tests + 3 benchmarks covering: read-cache writable preload, writeback preload (skip remote stat), zero-length read, pending shadow skip, LRU budget eviction, generation counter reopen, large data exclusion, corrupt meta pruning, conflict entry filtering.

srstack:
Addressed the latest review feedback and pushed commit aa51853.

**What changed**
- Validate scanned `.meta` entries before trusting `meta.Path`.
- Reject entries with an empty path.
- Reject entries whose scanned filename does not match `hashPath(meta.Path)+".meta"`.
- Remove the scanned `.meta` / `.dat` pair when an entry is invalid, both during startup recovery and `ListPending` reconciliation.

**Regression coverage**
- `TestNewWriteBackCache_PrunesMismatchedMetaPath`
- `TestWriteBackCache_ListPending_PrunesEmptyMetaPath`

**Validation**
- `go test ./pkg/fuse -run 'TestNewWriteBackCache_PrunesMismatchedMetaPath|TestWriteBackCache_ListPending_PrunesEmptyMetaPath|TestWriteBackCache_ListPending_CorruptMeta|TestWriteBackCache_GetViewPrunesMissingLargeData'`
- `go test ./pkg/fuse`

PR #410, opened 2026-05-08
feat: make inline storage threshold server-configurable

qiffang:
## Summary

- Make the DB-inline vs S3 storage cutoff (`smallFileThreshold`) and the synchronous text-extraction cap (`textExtractMaxBytes`) configurable on the server via env vars `DRIVE9_INLINE_THRESHOLD` / `DRIVE9_TEXT_EXTRACT_MAX_BYTES` (defaults unchanged: 50KB / 8KB).
- Server is the single source of truth: it advertises `inline_threshold` via `GET /v1/status`. Clients fetch lazily, cache via `atomic.Int64`, and only consume the once-guard on a successful response so transient warm failures retry rather than getting stuck on the local fallback.
- FUSE `Dat9FS.Init` synchronously warms with a bounded timeout (`inlineThresholdWarmTimeout = 5s`, var-overridable for tests) so the very first Create/Write/Flush/commit-queue decision sees the negotiated value. Hot paths use `CachedSmallFileThreshold()` and never trigger network I/O.
- CLI: `NewFromEnv` stays cold; only `cp` / `sh` use `NewFromEnvWithWarm` so read-only commands (ls/cat/stat/rm/grep/find) don't pay an extra RTT.

## Why one threshold, not two

Splitting into `DBInlineThreshold` (DB vs S3) and `DirectPutThreshold` (simple PUT vs V2 multipart) is tempting for performance, but with the current simple-PUT path it would re-introduce the server as a data-plane proxy: bodies above the inline cutoff would have to flow through the server process and `PutObject` to S3. The V2 multipart protocol exists specifically to avoid that. Keeping the two decisions tied means raising the threshold has predictable cost (TiDB blob column growth, not server bandwidth + memory). Documented inline at `pkg/backend/dat9.go`.

## Backwards compatibility

- **Reads**: completely unaffected. `ReadPlanCtx` consults persisted `StorageType` metadata, never the threshold. Files written under any threshold remain readable by any client.
- **Old client + new (higher) server threshold**: ✓ — old client still writes >50KB via V2 multipart; server accepts and stores in S3. No correctness issue, just foregoes some perf.
- **New client + old server (no `inline_threshold` field)**: ✓ — client falls back to compiled `DefaultSmallFileThreshold = 50_000`.
- **Transient warm failure (5xx, timeout)**: ✓ — failure doesn't consume the once-guard; next call (e.g. when server recovers) refetches.

## Test plan

- [x] `go test ./pkg/backend/ ./pkg/server/ ./pkg/client/ ./pkg/fuse/ -count=1`
- [x] `go test ./pkg/client/ ./pkg/fuse/ -count=1 -race` (atomic cache, no goroutine race)
- [x] New tests:
  - `TestConfigurableInlineThreshold` — backend `Options.InlineThreshold` / `TextExtractMaxBytes` propagation across `shouldStoreInDB` / `IsLargeFile` / `extractText`, including 0/negative fallback.
  - `TestTenantStatusReturnsInlineThreshold` — `/v1/status` echoes configured value.
  - `TestSmallFileThresholdCachesFromServer` / `FallsBackWhenServerOmits` / `LookupFailureCachesZero` / `CachedSmallFileThresholdDoesNotFetch` / `OverrideShortCircuits` / `ConcurrentReadersAreRaceFree` — client cache semantics.
  - `TestInitSynchronouslyWarmsInlineThreshold` / `TestInitWarmTimeoutFallsBackToDefault` — FUSE Init blocks on warm with bounded timeout, falls back cleanly on stall.

## Summary by CodeRabbit

* **New Features**
  * Server exposes a configurable inline-threshold via status; clients can warm and cache it.
  * New environment variables to configure inline threshold and text-extract max bytes.
  * Filesystem mount path performs an optional warmup to adopt server upload behavior.

* **Improvements**
  * Upload/read/write decisions now honor negotiated thresholds with sensible fallbacks for safer direct PUT vs multipart.

* **Tests**
  * Extensive tests added for negotiation, caching, warmup, thresholds, and boundary behaviors.

### DIFF
--- a/cmd/drive9-server-local/main.go
+++ b/cmd/drive9-server-local/main.go
@@ -578,6 +578,15 @@ func buildBackendOptionsFromEnv() (backend.Options, error) {
 		return backend.Options{}, fmt.Errorf("DRIVE9_MAX_TENANT_STORAGE_BYTES must be a positive integer")
 	}
 
+	opts.InlineThreshold = envInt64("DRIVE9_INLINE_THRESHOLD", backend.DefaultInlineThreshold)
+	if opts.InlineThreshold <= 0 {
+		return backend.Options{}, fmt.Errorf("DRIVE9_INLINE_THRESHOLD must be a positive integer")
+	}
+	opts.TextExtractMaxBytes = envInt64("DRIVE9_TEXT_EXTRACT_MAX_BYTES", backend.DefaultTextExtractMaxBytes)
+	if opts.TextExtractMaxBytes <= 0 {
+		return backend.Options{}, fmt.Errorf("DRIVE9_TEXT_EXTRACT_MAX_BYTES must be a positive integer")
+	}
+
 	// Quota enforcement source: "tenant" (default) or "server" (central server DB).
 	switch qs := strings.ToLower(strings.TrimSpace(os.Getenv("DRIVE9_QUOTA_SOURCE"))); qs {
 	case "", "tenant":
--- a/cmd/drive9-server/main.go
+++ b/cmd/drive9-server/main.go
@@ -249,6 +249,7 @@ func main() {
 			VaultMasterKey:   vaultMasterKey,
 			S3Dir:            s3cfg.Dir,
 			MaxUploadBytes:   maxUploadBytes,
+			InlineThreshold:  backendOptions.InlineThreshold,
 			Logger:           srvLogger,
 			SemanticEmbedder: semanticEmbedder,
 			SemanticWorkers:  semanticWorkerOpts,
@@ -266,6 +267,7 @@ func main() {
 		VaultIssuerURL:   vaultIssuerURL(addr),
 		S3Dir:            s3cfg.Dir,
 		MaxUploadBytes:   maxUploadBytes,
+		InlineThreshold:  backendOptions.InlineThreshold,
 		Logger:           srvLogger,
 		SemanticEmbedder: semanticEmbedder,
 		SemanticWorkers:  semanticWorkerOpts,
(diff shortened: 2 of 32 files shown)


qiffang:
**Review: APPROVE**

Well-structured configurable inline threshold. Key verification:

1. **Client cache**: Correct double-checked locking replacing `sync.Once` — transient failures don't permanently cache zero, next caller retries.
2. **FUSE threshold split**: Protocol-boundary decisions (flush routing, prefetcher, commit queue) use dynamic `inlineThreshold()`. Memory-pressure decisions (ReadCache Put/Get, commit-queue cache seeding) correctly stay on static `defaultSmallFileThreshold`.
3. **WriteBuffer per-buffer threshold**: All 6 references consistently use `wb.smallFileMax`, set via `SetSmallFileMax` from `fs.newWriteBuffer`.
4. **CommitQueue**: `directPutThreshold()` reads `CachedSmallFileThreshold()` — no network side-effect on hot commit path.
5. **Backwards compat**: Old client works with new server (just foregoes perf). New client falls back cleanly with old server.
6. **Test coverage**: Comprehensive — client cache semantics (6 tests), FUSE Init warm + timeout fallback, backend options propagation, server status response.

Minor non-blocking: `inlineThreshold()` lazy store into `smallFileMax` is a benign race but could be removed since `warmInlineThreshold` in Init should handle it.

srstack (changes requested):
Found one correctness issue that should be fixed before merge.

When the server is configured with `DRIVE9_INLINE_THRESHOLD` lower than the compiled 50KB default, a failed `/v1/status` warmup makes the client and FUSE fall back to `DefaultSmallFileThreshold`. That is not safe for lowered thresholds: files between the real server threshold and 50KB are sent via direct PUT, while the server treats them as large files and rejects them at the `b.IsLargeFile(cl)` gate with `missing X-Dat9-Part-Checksums header`.

Relevant paths:

- `pkg/client/transfer.go`: `uploadThreshold` falls back to `DefaultSmallFileThreshold` when the cache is empty.
- `pkg/fuse/dat9fs.go`: FUSE hot paths do the same via `inlineThreshold`.
- `pkg/server/server.go`: direct PUTs at or above the configured threshold require multipart checksums.

This means CLI/FUSE uploads can fail after a transient status timeout or 5xx whenever operators lower the inline threshold. A robust fix would be to either retry/fetch status before choosing direct PUT when the cache is empty, or fall back to multipart after this specific direct-PUT rejection. The new targeted tests pass, but they do not cover the lowered-threshold + warm-failure case.

Verification run locally:

`go test ./pkg/client ./pkg/fuse ./pkg/server -run 'TestSmallFileThreshold|TestInitSynchronouslyWarmsInlineThreshold|TestInitWarmTimeoutFallsBackToDefault|TestTenantStatusReturnsInlineThreshold' -count=1`

PR #411, opened 2026-05-08
docs: metadata schema refactor design — dentry / inode / content / semantic split

mornyx:
This PR adds a design document for introducing POSIX file permissions (mode/uid/gid) into drive9.

## What's covered

- **Industry research**: compared how S3-based FUSE filesystems (s3fs-fuse, goofys, mountpoint-s3, geesefs) and JuiceFS handle permission persistence, and why drive9 should follow the JuiceFS approach (storing permissions directly in the relational metadata DB).
- **Permission model**: POSIX Base Permissions (rwxrwxrwx + setuid/setgid/sticky) on the `file_nodes` table; ACL is intentionally out of scope for Phase 1.
- **Cross-layer design**: schema, datastore, backend, HTTP server, Go SDK, FUSE, and WebDAV.
- **FUSE compatibility**: `chmod`, `chown`, `ls -l`, and standard toolchain behavior (git, tar, rsync -a).
- **Backward compatibility**: default values for existing data, migration strategy, and test impact.
- **Security boundary**: clarifies that POSIX semantics are for toolchain compatibility while actual access control remains with the existing API Key / Capability Token auth model.

## Out of scope (Phase 1)

- ACL / `setfacl` / `getfacl`
- Sticky-bit/setuid/setgid semantic enforcement
- Server-side authorization based on uid/gid

## Summary by CodeRabbit

* **Documentation**
  * Added a comprehensive design doc for a metadata schema refactor: splits file data into path mapping, inode metadata, content storage, and semantic layers; standardizes inode_id and directory rules; clarifies POSIX permission semantics and pending-create mode flow; defines safe multi-table content-overwrite transactions and embedding-reset rules; includes cross-layer impacts, provider indexing notes, zero-downtime migration plan, implementation order, and test checklist.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mem9-ai/drive9/pull/411)


qiffang (commented):
## Adversarial Review — Metadata Schema Refactor Design Document

**Verdict: REVISE — 2 blocking, 4 recommendations**

---

### Blocking Issues

**B1: `inodes` table missing explicit `mtime` — silent data loss risk on content overwrite**

The `inodes` schema has: `inode_id, size_bytes, revision, mode, status, created_at, confirmed_at, expires_at`. There is no explicit `mtime` column.

The FUSE layer uses `InodeEntry.Mtime` everywhere (`inode.go:16`, `dat9fs.go:617`, `UpdateMtime` at line 229). Currently this is derived from server-side timestamps (`confirmed_at` via `X-Dat9-Mtime` header, `client.go:590`).

The problem: `confirmed_at` is set once on first confirmation and **may not be updated on content overwrite** in all code paths. The design's multi-table transaction (§4.8) shows `confirmed_at = ?` in the UPDATE — this works, but the design never explicitly states that `confirmed_at` is the `mtime` proxy. If a future implementer reads the schema and sees `confirmed_at`, they'll reasonably interpret it as "when the file was first confirmed" (the natural reading), not "last modification time."

**Fix:** Either:
1. Add an explicit `mtime DATETIME(3)` column to `inodes` and document its semantics, OR
2. Add a section explicitly stating "`confirmed_at` serves as `mtime`; it is updated on every content overwrite, not just initial confirmation." With rationale for not adding a separate column.

Without this, the schema is ambiguous and the implementer will get it wrong.

**B2: Migration step 3→4 ordering creates a window where SQL references wrong column name**

Migration sequence:
- Step 3: `UPDATE file_nodes SET inode_id = node_id WHERE is_directory = 1 AND inode_id IS NULL;`
- Step 4: `ALTER TABLE file_nodes CHANGE file_id inode_id VARCHAR(64);`

Problem: Step 3's UPDATE references `inode_id` but the column is still named `file_id` at that point (rename happens in step 4). The SQL in step 3 will fail with "Unknown column 'inode_id'."

**Fix:** Clarify the exact ordering. The safe sequence is:
1. Create new tables
2. Migrate file data into new tables
3. Populate directory inodes using the **current** column name (`file_id`): `UPDATE file_nodes SET file_id = node_id WHERE is_directory = 1 AND file_id IS NULL`
4. Rename column `file_id` → `inode_id`
5. Switch application code

---

### Recommendations

**R1: `nlink` computed via `COUNT(*)` has a correctness gap during concurrent hardlink creation + deletion**

The design says `nlink` is computed on demand via `SELECT COUNT(*) FROM file_nodes WHERE inode_id = ?`, used during deletion to decide whether to GC the inode. Consider: thread A calls `DeleteFileWithRefCheck`, gets `COUNT(*) = 1`, decides to delete the inode. Between the COUNT and the DELETE, thread B creates a new hardlink via `CopyFileCtx`. Now the inode is deleted but thread B's dentry points to a nonexistent inode. Current code likely holds a `FOR UPDATE` lock during this check, but the design should state the locking requirement explicitly — "computed on demand" implies no locking.

**R2: `semantic_tasks.resource_id` — document the identity invariant**

`semantic_tasks.go:79,103,128` all pass `fileID` as `ResourceID`. After the rename, these will pass `inodeID`. The column stays `resource_id` but the semantic changes. Add a concrete list of what needs updating: `SemanticTask.ResourceID` comment, `CreateSemanticTask` godoc, and any API docs that reference `resource_id`.

**R3: Three-table JOIN stale-check — consider a `stale` boolean instead of cross-table revision comparison**

`s.embedding_revision = i.revision` spans `semantic` → `inodes`. This cross-table predicate may prevent PostgreSQL from using the `hnsw` index efficiently and adds cognitive complexity. Alternative: add `embedding_stale BOOLEAN DEFAULT TRUE` to `semantic`. Set `TRUE` on content overwrite, `FALSE` when embedding worker completes. Single-table predicate: `WHERE s.embedding IS NOT NULL AND NOT s.embedding_stale`.

**R4: ListDir performance claim — note struct split as the forcing function**

The design correctly identifies that `ListDir` (`store.go:1003-1007`) currently JOINs the full `files` row including `content_blob`, `content_text`, `description`, and embedding revisions. The performance improvement is real. However, the design should note: `StatForRead` (`store.go:931`) already demonstrates column-pruning. The reason `ListDir` wasn't already pruned is that the Go scan requires all columns from the `File` struct — the struct split (`Inode`/`Content`/`Semantic`) is the forcing function. Stating this makes the design self-documenting.

---

### Checklist
| Gate | Result |
|------|--------|
| All inputs handled | PASS |
| Edge cases | FAIL (B1 mtime ambiguity, B2 migration ordering) |
| Failure behavior | PASS |
| Follows architecture | PASS |
| API explicit | FAIL (B1 — mtime semantics undocumented) |
| Race/state | CONDITIONAL (R1 — nlink locking assumption unstated) |
| Tests | PASS (comprehensive test file list in §7.1) |

Fix B1 and B2 before merge.

qiffang (commented):
## Consolidated Review — @dat9-dev1 + @adversary-2

**Verdict: REVISE — 2 blocking, 6 recommendations, 3 structural suggestions**

---

### Blocking Issues

**B1: `inodes` table missing explicit `mtime` semantics** (@adversary-2 + @dat9-dev1)

The `inodes` schema has `confirmed_at` but no explicit `mtime` column. FUSE uses `InodeEntry.Mtime` everywhere (`inode.go:16`, `dat9fs.go:617`). Currently derived from `confirmed_at` via `X-Dat9-Mtime` header (`client.go:590`).

The design's multi-table transaction (§4.8) does `UPDATE inodes SET ... confirmed_at = ?` on content overwrite — this works, but the design never states that `confirmed_at` serves as `mtime`. A future implementer will reasonably interpret `confirmed_at` as "when the file was first confirmed" (natural reading), not "last modification time."

**Fix:** Either add an explicit `mtime DATETIME(3)` column to `inodes`, or add a section explicitly stating "`confirmed_at` serves as `mtime`; it is updated on every content overwrite, not just initial confirmation."

**B2: Migration step 3→4 ordering — SQL references `inode_id` before column rename** (@adversary-2)

Step 3's `UPDATE` references `inode_id` but the column is still named `file_id` at that point (rename happens in step 4). The SQL will fail with "Unknown column 'inode_id'."

**Fix:** Use `file_id` in step 3's SQL, then rename in step 4:
1. Create new tables
2. Migrate file data into new tables
3. Populate directory inodes using current column name: `UPDATE file_nodes SET file_id = node_id WHERE is_directory = 1 AND file_id IS NULL`
4. Rename column `file_id` → `inode_id`
5. Switch application code

---

### Recommendations

**R1: `nlink` via `COUNT(*)` — state locking requirement** (@adversary-2)

Concurrent hardlink creation + deletion race: thread A gets `COUNT(*) = 1`, decides to GC inode; thread B creates new hardlink between COUNT and DELETE. The design should state the `FOR UPDATE` locking requirement explicitly — "computed on demand" implies no locking.

**R2: `semantic_tasks.resource_id` — document identity invariant** (@adversary-2 + @dat9-dev1)

`semantic_tasks.go:79,103,128` all pass `fileID` as `ResourceID`. After rename, semantics change from "file identity" to "inode identity." Add concrete list of what to update: `SemanticTask.ResourceID` comment, `CreateSemanticTask` godoc, and any API docs referencing `resource_id`.

**R3: Cross-table stale-check — consider `embedding_stale` boolean** (@adversary-2 + @dat9-dev1)

`s.embedding_revision = i.revision` spans `semantic` → `inodes`. May prevent PostgreSQL/TiDB from using vector index efficiently. Alternative: add `embedding_stale BOOLEAN DEFAULT TRUE` to `semantic`, set TRUE on content overwrite, FALSE when embedding worker completes. Single-table predicate eliminates the cross-table comparison.

**R4: ListDir performance claim — justify why not column-pruned today** (@adversary-2 + @dat9-dev1)

`StatForRead` (`store.go:931`) already demonstrates column-pruning. If ListDir wasn't pruned because Go scan requires all columns from the `File` struct, say so — it justifies the struct split as the forcing function.

**R5: `file_nodes.inode_id` for directories — validate NULL values** (@dat9-dev1)

Directories' `file_id` may be NULL or empty (EnsureParentDirs doesn't create `files` rows). Before migration: `SELECT COUNT(*) FROM file_nodes WHERE is_directory = 1 AND file_id IS NULL` to confirm count and ensure `ALTER TABLE CHANGE` doesn't conflict.

**R6: `chmod` API endpoint versioning** (@dat9-dev1)

`?action=chmod` is a new API surface. Confirm whether it needs versioning (`/v1/` vs `/v2/`) and that `ChmodCtx` Go SDK signature is consistent with existing API patterns.

---

### Structural Suggestions

**S1: Incremental migration** (@dat9-dev1)

Step 1-4 is a large atomic change. Suggest: 1a create new tables + dual-write → 1b switch reads → 1c stop old writes → 1d verify + cleanup. Each step is independently rollable.

**S2: Rollback verification** (@dat9-dev1)

Design says "retain old table for one release cycle" but doesn't specify how to verify rollback safety. Add a verification step: compare new vs old table data consistency post-migration.

**S3: Impact completeness** (@dat9-dev1)

13 test files listed, but may miss `pkg/fuse/sse.go` (SSE watcher referencing `file_id`?) and `pkg/fuse/pending_index.go` (meta format changes).

---

### What's Good

- Schema split aligns cleanly with ext4 semantics (dentry/inode/content/semantic)
- 1:1 relation using `inode_id` as PK — clean vertical split, no extra JOIN key
- Phase 1 mode-only scope (no uid/gid) is pragmatically correct for API-key auth
- GENERATED column migration risk identified with fallback strategy
- Optimistic lock on `inodes.revision` only — maintains existing concurrency semantics
- Comprehensive test file list in §7.1

---

### Checklist
- All inputs handled: ✅
- Edge cases: ❌ (B1 mtime ambiguity, B2 migration ordering)
- Failure behavior: ✅
- Follows architecture: ✅
- API explicit: ❌ (B1 — mtime semantics undocumented)
- Race/state: conditional (R1 — nlink locking unstated)
- Tests: ✅ (comprehensive test list)

Fix B1 and B2 before merge.

srstack (commented):
Findings:

1. `docs/design/metadata-schema-refactor.md:430` / `:476` / `:491` / `:511`

The “zero-downtime” migration is not actually zero-downtime. It renames `file_nodes.file_id`, `uploads.file_id`, `file_tags.file_id`, and `file_gc_tasks.file_id` before switching application code. Any old binary still running, or any rollback to old code, will immediately break because it still queries those column names. This needs an expand/dual-write/dual-read/contract plan instead of an in-place rename before cutover.

2. `docs/design/metadata-schema-refactor.md:447`

The TiDB auto-embedding migration tries to insert into `embedding` and `description_embedding`, but in `tidb_auto.go` those are `GENERATED ALWAYS AS (...) STORED` columns. That SQL will either be rejected or, if omitted, TiDB will recompute embeddings. The doc’s mitigation is internally contradictory and needs a concrete TiDB-auto migration path.

3. `docs/design/metadata-schema-refactor.md:439`

`mtime` is `NOT NULL`, but the migration fills it from `confirmed_at` for all `status != 'DELETED'`. Pending uploads currently have `confirmed_at = NULL`, so an online migration with in-flight uploads can fail. Use `COALESCE(confirmed_at, created_at)` or handle pending rows separately.

4. `docs/design/metadata-schema-refactor.md:270` / `:392`

The refcount locking claim is unsafe. The doc says `COUNT(*) ... FOR UPDATE` prevents concurrent hardlink creation, but `CopyFile` is explicitly left unchanged and currently only inserts a new `file_nodes` row without taking an inode-level lock. A delete can count zero links and mark the inode deleted while a concurrent copy inserts a new dentry to it. The design should require copy/delete/replace to coordinate on the inode row or maintain `nlink` transactionally.

Tests not run. This PR is documentation-only; I reviewed the proposed design against the current schema and datastore behavior.

srstack (changes requested):
Findings:

1. `docs/design/metadata-schema-refactor.md:461`

The TiDB auto-embedding backfill still drops the revision trackers. The updated SQL correctly avoids listing the generated `embedding` / `description_embedding` columns, but it now inserts only `(inode_id, content_text, description)`. The search path still treats embeddings as fresh only when `embedding_revision = revision` / `description_embedding_revision = revision` (and the new design keeps that as `s.embedding_revision = i.revision`). With the proposed migration, generated vectors may exist but their revision columns remain NULL, so migrated auto-embedding rows will be filtered out of revision-gated vector search. The auto-mode migration should copy the non-generated revision columns while omitting only the generated vector columns, e.g. include `embedding_revision` and `description_embedding_revision` in the insert/select list.

2. `docs/design/metadata-schema-refactor.md:270` / `docs/design/metadata-schema-refactor.md:403`

The cross-layer change table still says `CopyFileCtx` is unchanged and only touches `file_nodes`, but section 5.1 now correctly requires `CopyFile`, delete, and replace to acquire the `inodes` row lock before mutating dentries for that inode. Leaving `CopyFileCtx` unchanged would preserve the delete-vs-hardlink race that section 5.1 is trying to fix. Please update the backend/datastore implementation plan so copy also participates in the inode-row locking protocol.

Tests not run. This PR is documentation-only; I reviewed the updated design against the current schema/search/update behavior.

srstack (changes requested):
Findings:

1. `docs/design/metadata-schema-refactor.md:169`

The TiDB GENERATED-column mitigation section is still internally contradictory and conflicts with the corrected migration SQL later in the document. It says that if TiDB recomputes embeddings, the migration should insert “all columns including the pre-computed embedding values” and also “skip the generated column”. For `tidb_auto.go`, `embedding` and `description_embedding` are `GENERATED ALWAYS` columns, so the migration cannot copy precomputed values into them; the correct approach is the one shown later: omit only the generated vector columns and copy the non-generated revision trackers. Please update this earlier mitigation text so it matches the concrete SQL in section 6.

2. `docs/design/metadata-schema-refactor.md:572`

The implementation order still places `file_id` -> `inode_id` rename in step 1 with the initial schema work, which conflicts with the rollback-safe migration plan in section 6. The plan now correctly says shared-table column renames are Phase B contraction work after all old binaries are gone. Step 1 should describe adding the split tables / dual-write-compatible schema only, and move the rename to a later contraction step, otherwise implementers may perform the non-rollback-safe rename too early.

Tests not run. This PR is documentation-only; I reviewed the updated design against the current schema/search/update behavior.

srstack (approved):
No blocking findings in this round.

The previous issues have been addressed: the TiDB GENERATED-column mitigation now matches the concrete migration SQL by omitting only generated vector columns while preserving revision trackers, and the implementation order now keeps shared-table `file_id` during Phase A with schema contraction moved to Phase B.

Tests not run. This PR is documentation-only; I reviewed the updated design text against the current schema/search/update behavior.

PR #404, opened 2026-05-07
fix(fuse): enforce UTF-8 text detection and fix git pack mmap compatibility

JaySon-Huang:
## Summary

Fixes git clone failures on FUSE mounts by addressing two root causes:

1. **UTF-8 enforcement for text detection** — Binary data (e.g. git pack objects) was incorrectly classified as `text/*` via extension-based content type detection, causing `Incorrect string value` errors in the TiDB text content column.
2. **Git pack mmap compatibility** — Changed large prefetched file open flags from `FOPEN_DIRECT_IO` to `FOPEN_KEEP_CACHE` so mmap-based readers (git pack verification) don't SIGBUS on macFUSE.
3. **Config.lock rename race** — Added `remotePathExists` fallback in `renamePendingNewCommit` to handle canceled uploads that already reached the remote, preventing persistent stale config.lock entries.

## Changes

| File | Description |
|------|-------------|
| `pkg/backend/dat9.go` | Refactored `isTextContent` → `utf8.Valid()`; new `isTextualContentType()` helper; extension-inferred text types now require valid UTF-8 |
| `pkg/backend/dat9_test.go` | Added tests for invalid UTF-8 rejection and gbk CSV edge case |
| `pkg/fuse/dat9fs.go` | `FOPEN_DIRECT_IO` → `FOPEN_KEEP_CACHE` for prefetched files; `remotePathExists` helper; pending rename falls back to remote rename when upload-cancel-race detected |
| `pkg/fuse/dat9fs_test.go` | Test for `renamePendingNewCommit` remote fallback path |

## Verification

- `git clone --no-hardlinks` of the full tidb repo onto FUSE mount completes successfully
- `git status`, `git fsck --full`, `git verify-pack` all pass
- No more `Incorrect string value`, `SIGBUS`, or `could not lock config` errors in logs

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved text extraction to require valid UTF-8 before treating content as text.
  * Fixed read-only file caching to avoid mmap SIGBUS and enable prefetching.
  * Enhanced pending-file rename handling to correctly fall back when remote state changes.

* **Chores**
  * Standardized user-facing branding and logging/messages to "drive9" across mounts, WebDAV, perf summaries, and SSE output.

PR #412, opened 2026-05-09
fix(fuse): improve unmount UX on Ctrl+C and re-mount safety

qiffang:
## Problem

Three issues on `drive9 mount` shutdown that compounded into a bad user experience:

1. **First Ctrl+C looked hung.** `shutdown()` drains the commit queue and flushes open fds with no progress output. Users assumed the process was stuck and pressed Ctrl+C again.
2. **Second Ctrl+C corrupted the mountpoint.** The forced-exit path called `os.Exit(1)` without `forceUnmount`, leaving the FUSE endpoint registered with the kernel. Re-mounting the same path returned `fusermount3: failed to access mountpoint: Permission denied`.
3. **`forceUnmount` preferred `fusermount` over `fusermount3`.** On fuse3 systems (Ubuntu 22.04+, Debian 12+) it picked the wrong binary.

### `CommitQueue.PendingStats()` (new, read-only)
Returns `(count, bytes)` of pending uploads, deduped by entry pointer. Two subtleties motivate pointer-dedup:
- An in-flight entry stays in `cq.queue` until cleanup runs (`commit_queue.go:692`), so the same `*CommitEntry` lives in both `cq.queue` and `cq.inFlight`.
- `Enqueue` does not dedupe by `Path` (`commit_queue.go:153`), so a path written + closed twice yields two distinct entries with the same Path string.

### Signal handler refactor (`mount.go`)
- `sigCh` buffer 1 → 2 to avoid dropping a second signal that arrives between the outer goroutine's read and the inner goroutine being scheduled.
- Periodic progress reporter prints every 2s during shutdown, distinguishing commit-queue uploads (with file count + bytes) from other drain phases (debouncer flush, per-fd flush, write-back drain). No more silent waits.
- Forced-exit path calls `forceUnmount` and removes the pid file before `os.Exit(1)`, so re-mount of the same path succeeds.
- Forced-exit message surfaces local cache location but **does not promise automatic resume on next mount** — ShadowSpill files take a terminal-failure branch on conflict (`commit_queue.go:716-722`) to avoid OOMing during full-file byte comparison, so recovery is best-effort.
- Comment documents the headless/daemon limitation: `shutdown()` has no context cancellation, so a stuck worker can block indefinitely. Daemon operators must send a second SIGTERM (e.g. via systemd's `TimeoutStopSec`) to trigger force-quit.

### `forceUnmount` hardening
- 5s timeout via `exec.CommandContext` so the forced-exit path can't hang on a wedged FUSE endpoint.
- Linux: prefer `fusermount3` → `fusermount` → `umount -l` (matches `cmd/drive9/cli/mount.go:457` `umountArgv()`).

### `humanizeBytes` helper
Small helper for "1.5 MB" formatting. No existing helper in the repo.

## Sample output

First Ctrl+C, busy commit queue:
```
^C
drive9: unmounting /home/ubuntu/w4...
drive9: still uploading 17 files (43.2 MB) — Ctrl+C again to force-quit
drive9: still uploading 9 files (21.0 MB) — Ctrl+C again to force-quit
drive9: still draining — Ctrl+C again to force-quit
```

Second Ctrl+C while uploads pending:
```
drive9: force-quit — abandoning 9 files (21.0 MB); local state preserved in /home/ubuntu/.cache/drive9, but recovery is best-effort
```

## Non-changes
- No data-path changes
- No public API changes (`PendingStats` is additive)
- No test changes (existing tests still pass)

## Test plan
- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `go test ./pkg/fuse/` (full suite, 9.1s)
- [ ] Manual: mount, write 100 × 10 MB files, Ctrl+C once → expect periodic progress every 2s, then clean unmount
- [ ] Manual: same setup, Ctrl+C twice → expect force-quit message naming abandoned files + cache dir; then `drive9 mount` of the same path succeeds without "Permission denied"
- [ ] Manual on fuse3 host (Ubuntu 22.04): verify `fusermount3` is used by `forceUnmount` (check via `strace -e execve` or by removing `fusermount` only)

## Follow-ups (not in this PR)
- Same `buffer=1` signal pattern exists in `vault_mount.go` and `mount_webdav.go` (consistency)
- Per-fd flush phase has no progress output (needs hooks in `Dat9FS.FlushAll`)
- Streamer-fallback memory issue (`OnPartFull` in normal path doesn't evict `wb.parts`) — separate issue, original PR-2 proposal was withdrawn after review identified data-loss bugs in the naive fix
- Headless/daemon: real fix is context cancellation in `CommitQueue.DrainAll`

## Summary by CodeRabbit

* **New Features**
  * Progress reporting during graceful shutdown displays upload/drain status
  * Force shutdown now reports remaining work and local state preservation location

* **Bug Fixes**
  * Enhanced unmount handling with timeout protection
  * Improved shutdown signal handling reliability
  * Better byte count display formatting

[Review Change Stack](https://app.coderabbit.ai/change-stack/mem9-ai/drive9/pull/412)

### DIFF
--- a/pkg/fuse/commit_queue.go
+++ b/pkg/fuse/commit_queue.go
@@ -170,6 +170,42 @@ func (cq *CommitQueue) Pending() int {
 	return len(cq.queue)
 }
 
+// PendingStats returns the number of unique pending entries and the sum
+// of their sizes. Used by unmount to display drain progress.
+//
+// Subtleties:
+//  1. In-flight entries remain in cq.queue until the worker calls
+//     removeFromQueue() *after* cleanup (see line 692). The same
+//     *CommitEntry pointer is stored in both cq.queue and cq.inFlight
+//     during that window.
+//  2. Enqueue does NOT dedupe by Path (line 153 just appends). A path
+//     written + closed twice yields two distinct entries with the same
+//     Path string but different pointers.
+//
+// We dedupe by entry pointer: collapses (1) without collapsing (2).
+func (cq *CommitQueue) PendingStats() (count int, bytes int64) {
+	cq.mu.Lock()
+	defer cq.mu.Unlock()
+	seen := make(map[*CommitEntry]struct{}, len(cq.queue))
+	for _, e := range cq.queue {
+		if _, ok := seen[e]; ok {
+			continue
+		}
+		seen[e] = struct{}{}
+		count++
+		bytes += e.Size
+	}
+	for _, e := range cq.inFlight {
+		if _, ok := seen[e]; ok {
+			continue
+		}
+		seen[e] = struct{}{}
+		count++
+		bytes += e.Size
+	}
+	return
+}
+
 // IsFull reports whether the queue has reached its backpressure limit.
 func (cq *CommitQueue) IsFull() bool {
 	cq.mu.Lock()
--- a/pkg/fuse/mount.go
+++ b/pkg/fuse/mount.go
@@ -347,22 +347,100 @@ func Mount(opts *MountOptions) error {
 
 	shutdown := newMountShutdown(sseWatcher.Stop, dat9fs.FlushAll)
 
-	// Signal handling for graceful shutdown
-	sigCh := make(chan os.Signal, 1)
+	// Signal handling for graceful shutdown.
+	//
+	// First SIGINT/SIGTERM:
+	//   1. Start a progress reporter goroutine that prints commit-queue
+	//      depth every 2s so the user knows we're not hung.
+	//   2. Call shutdown() — flushes open fds, drains commit queue.
+	//   3. Retry server.Unmount() up to 5 times (EBUSY is transient).
+	//   4. Fall back to forceUnmount on retry exhaustion.
+	//
+	// Second SIGINT/SIGTERM during step 1-3:
+	//   - Tell the user how much work is being abandoned (count + bytes).
+	//   - Surface where local state is preserved so they can recover or
+	//     inspect it. Do NOT promise automatic resume — recovery is
+	//     best-effort: ShadowSpill (large) files take the terminal-failure
+	//     branch on conflict (commit_queue.go:716-722) to avoid OOMing
+	//     during full-file byte comparison.
+	//   - forceUnmount the mountpoint so re-mount doesn't fail with
+	//     "Permission denied" (the FUSE endpoint must be released).
+	//   - Exit with code 1.
+	//
+	// Buffer size 2: outer goroutine consumes the first signal, inner
+	// goroutine the second. Buffer ≥ 2 ensures signals delivered between
+	// the outer's <-sigCh return and the inner goroutine being scheduled
+	// don't get dropped by signal.Notify.
+	//
+	// Headless/daemon limitation: if shutdown() blocks indefinitely on a
+	// stuck commit-queue worker (no context cancellation), an interactive
+	// user can press Ctrl+C twice; daemon operators must send a second
+	// SIGTERM (e.g. via systemd's TimeoutStopSec) to trigger force-quit.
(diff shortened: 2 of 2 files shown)


qiffang (commented):
## Consolidated Review — @dat9-dev1 + @adversary-2

**Verdict: GREEN — 4 recommendations**

---

### Verified Fixes

1. **`sigCh` buffer 1→2** — Correct. Buffer=2 ensures second signal delivered between outer goroutine's `<-sigCh` return and inner goroutine scheduling is not dropped. Comment accurately cites the semantics.

2. **Force-exit calls `forceUnmount` + removes pidFile before `os.Exit(1)`** — Fixes the mountpoint corruption bug where re-mount fails with "Permission denied." Correct ordering: unmount FUSE endpoint first, then remove pid file, then exit.

3. **`forceUnmount`: `fusermount3` → `fusermount` → `umount -l`** — Now matches `umountArgv()` priority. 5s timeout via `exec.CommandContext` prevents hang on wedged endpoint. Timeout error distinguished in output.

4. **`humanizeBytes`: `b < 0` guard** — Good defensive coding.

5. **Headless/daemon limitation** — Documented clearly in comment. Real fix (context cancellation in `CommitQueue.DrainAll`) is explicitly deferred.

6. **Progress reporter** — Clean: 2s ticker, distinguishes commit-queue uploads (with count+bytes) from other drain phases ("still draining"), stops via `close(progressDone)` after `shutdown()` returns. No goroutine leak.

7. **Forced-exit message says "recovery is best-effort"** — Correctly avoids promising automatic resume, which would be misleading for ShadowSpill files.

8. **`context.DeadlineExceeded` check** in `forceUnmount` error handling — gives users actionable diagnostics.

---

### `PendingStats()` Correctness

Pointer-dedup via `map[*CommitEntry]struct{}` is the right approach:
- Collapses in-flight entries that are also in `cq.queue` (same pointer)
- Does not collapse distinct entries for the same path (different pointers from separate Enqueue calls)
- Holds `cq.mu.Lock()` for both iterations — consistent snapshot

---

### Recommendations

**R1: `PendingStats()` should have a unit test** (@adversary-2)

The dedup logic has non-obvious invariants (pointer identity vs path identity, in-flight overlap with queue). A unit test that:
1. Enqueues 3 entries
2. Moves 1 to in-flight (same pointer in both slices)
3. Asserts `PendingStats()` returns `(3, expectedBytes)` not `(4, ...)`

This is straightforward to test without signal handling complexity.

**R2: Document `Size` read safety** (@dat9-dev1)

`PendingStats()` reads `e.Size` under `cq.mu`. `Size` is set during `Enqueue()` and never mutated after — the lock is for queue/inFlight iteration safety, not for `Size` reads. A one-line comment like `// Size is immutable after Enqueue, safe to read under cq.mu` would help future readers.

**R3: Progress reporter may print one extra message after shutdown** (@dat9-dev1)

The select has two cases: `<-progressDone` and `<-ticker.C`. If both are ready simultaneously, Go selects randomly. One extra "still uploading…" message could print after `shutdown()` returns. Cosmetically harmless — no fix needed, just noting.

**R4: Branding inconsistency** (@adversary-2)

New code uses `drive9:` prefix, existing untouched lines use `dat9:`. Not a blocker — `drive9:` is correct per PR #404 — but consider a follow-up cleanup of remaining `dat9:` prefixes in mount.go.

---

### Checklist
- All inputs handled: ✅
- Edge cases: ✅ (negative bytes, nil commitQueue, timeout, concurrent signals)
- Failure behavior: ✅
- API explicit: ✅ (PendingStats is additive, well-documented)
- Race/state: ✅ (mutex-protected, closure captures are read-only)
- Tests: conditional (R1 — PendingStats unit test recommended)

PR #415, opened 2026-05-11
Add Windows CLI support

srstack:
## Summary
- add Windows CLI support for the default WebDAV mount flow, including drive-letter normalization and `net use` mount/unmount handling
- keep FUSE and vault mount unsupported on Windows, but guide users to the supported Windows paths: default WebDAV mount for filesystems and `drive9 vault` or the vault API for vault access
- split doctor platform helpers, update docs, and add Windows-focused tests for mount behavior, unsupported-mode messaging, and release packaging
- include Windows CLI binaries in release builds and publish `.exe` artifacts for `windows/amd64` and `windows/arm64`

## Validation
- `go test ./cmd/drive9/cli -run "TestRunDoctorFuseWindowsReportsUnsupported|TestMountFuseImplWindowsReportsWebDAVAlternative|TestMountVaultImplWindowsReportsCLIAlternative|TestResolveMountMode_Windows|TestWebdavMountCmd_Windows|TestWebdavUnmountCmdWindows|TestUmountArgvWindows" -count=1`
- `go build -o dist-pr-check/drive9-windows-amd64.exe ./cmd/drive9`
- `go build -o dist-pr-check/drive9-windows-arm64.exe ./cmd/drive9`
- real Windows validation against `https://api.drive9.ai` for `drive9 mount :/ X:` lifecycle, CLI-to-mount visibility, mount-to-CLI visibility, and unmount cleanup

## Summary by CodeRabbit

* **New Features**
  * Adds Windows WebDAV mount with drive-letter mapping and Windows-friendly mount/unmount flows; releases now include Windows CLI binaries.

* **Improvements**
  * Better Windows error hints and guidance for mount/unmount failures; improved unmount behavior and PID/process cleanup across platforms.
  * Mount state now records richer process metadata for more robust lifecycle handling while remaining backward compatible.

* **Documentation**
  * README and docs updated with Windows mount guidance and FUSE platform notes.

* **Tests**
  * Added Windows-focused tests for mounting, unmounting and diagnostics.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mem9-ai/drive9/pull/415)

### DIFF
--- a/cmd/drive9/cli/mount.go
+++ b/cmd/drive9/cli/mount.go
@@ -162,6 +162,14 @@ func fsMountCmd(args []string) error {
 	if resolved == MountModeWebDAV && *readOnly {
 		return fmt.Errorf("drive9 mount: --read-only is not supported with WebDAV mode")
 	}
+	if runtime.GOOS == "windows" && resolved == MountModeFUSE {
+		return mountFuse(&mountFuseOptions{
+			MountPoint: mountPoint,
+			RemoteRoot: remoteRoot,
+			ReadOnly:   *readOnly,
+			Debug:      *debug,
+		})
+	}
 
 	serverVal, apiKeyVal, tokenVal, err := resolveMountCredentials(ResolveCredentials(), *server, *apiKey)
 	if err != nil {
--- a/cmd/drive9/cli/mount.go
+++ b/cmd/drive9/cli/mount.go
@@ -313,7 +313,7 @@ type umountDeps struct {
 	lookPath  func(string) (string, error)
 	run       func([]string) error
 	readPID   func(string) (int, string, error)
-	terminate func(int) error
+	terminate func(int, time.Duration) error
 	pidAlive  func(int) bool
 	now       func() time.Time
 	sleep     func(time.Duration)
@@ -384,9 +384,10 @@ func runUmount(args []string, deps umountDeps) error {
 		if deps.terminate == nil {
 			return fmt.Errorf("drive9 umount: no process terminator configured for Windows WebDAV mount")
 		}
-		if err := deps.terminate(pid); err != nil {
-			return fmt.Errorf("drive9 umount: terminate mount process pid %d: %w", pid, err)
+		if err := deps.terminate(pid, *waitTimeout); err != nil {
+			return fmt.Errorf("%w (pid file: %s)", err, path)
 		}
+		return nil
 	}
 	if *waitTimeout == 0 {
 		return nil
@@ -417,27 +418,10 @@ func waitForPIDExit(pid int, timeout time.Duration, deps umountDeps) error {
 }
 
 func processAlive(pid int) bool {
-	if pid <= 0 {
-		return false
-	}
-	process, err := os.FindProcess(pid)
-	if err != nil {
-		return false
-	}
-	err = process.Signal(syscall.Signal(0))
-	if err == nil {
-		return true
-	}
-	if errors.Is(err, os.ErrProcessDone) || errors.Is(err, syscall.ESRCH) {
-		return false
-	}
-	if errors.Is(err, syscall.EPERM) {
-		return true
-	}
-	return false
+	return processAliveImpl(pid)
 }
 
-func terminateProcess(pid int) error {
+func terminateProcess(pid int, waitTimeout time.Duration) error {
 	if pid <= 0 {
 		return fmt.Errorf("invalid mount process pid %d", pid)
 	}
@@ -452,7 +436,13 @@ func terminateProcess(pid int) error {
 	if errors.Is(err, os.ErrProcessDone) || errors.Is(err, syscall.ESRCH) {
 		return nil
 	}
-	return err
+	if err != nil {
+		return err
+	}
+	if waitTimeout > 0 {
+		return waitForProcessExitByPID(pid, waitTimeout)
+	}
+	return nil
 }
 
 // resolveMountCredentials selects the (server, apiKey, token) triple that a
--- a/cmd/drive9/cli/mount.go
+++ b/cmd/drive9/cli/mount.go
@@ -369,8 +369,9 @@ func runUmount(args []string, deps umountDeps) error {
 	if err != nil {
 		return err
 	}
-	if err := deps.run(argv); err != nil {
-		return err
+	runErr := deps.run(argv)
+	if deps.goos != "windows" && runErr != nil {
+		return runErr
 	}
 	// Windows WebDAV mounts run a local bridge in the mount process. Even when
 	// the caller opts out of waiting, still request process shutdown before
@@ -378,23 +379,35 @@ func runUmount(args []string, deps umountDeps) error {
 	pid, path, err := deps.readPID(stateMountPoint)
 	if err != nil {
 		if errors.Is(err, os.ErrNotExist) {
-			return nil
+			return runErr
+		}
+		if runErr != nil {
+			return runErr
 		}
 		return err
 	}
 	if deps.goos == "windows" {
 		if deps.terminate == nil {
+			if runErr != nil {
+				return runErr
+			}
 			return fmt.Errorf("drive9 umount: no process terminator configured for Windows WebDAV mount")
 		}
 		if err := deps.terminate(pid, *waitTimeout); err != nil {
+			if runErr != nil {
+				return runErr
+			}
 			return fmt.Errorf("%w (pid file: %s)", err, path)
 		}
 		if path != "" && deps.remove != nil {
 			if err := deps.remove(path); err != nil && !errors.Is(err, os.ErrNotExist) {
+				if runErr != nil {
+					return runErr
+				}
 				return fmt.Errorf("drive9 umount: remove mount pid file %s: %w", path, err)
 			}
 		}
-		return nil
+		return runErr
 	}
 	if *waitTimeout == 0 {
 		return nil
--- a/cmd/drive9/cli/mount.go
+++ b/cmd/drive9/cli/mount.go
@@ -314,6 +314,7 @@ type umountDeps struct {
 	run       func([]string) error
 	readPID   func(string) (int, string, error)
 	terminate func(int, time.Duration) error
+	remove    func(string) error
 	pidAlive  func(int) bool
 	now       func() time.Time
 	sleep     func(time.Duration)
@@ -332,6 +333,7 @@ func defaultUmountDeps() umountDeps {
 		},
 		readPID:   mountstate.ReadPID,
 		terminate: terminateProcess,
+		remove:    os.Remove,
 		pidAlive:  processAlive,
 		now:       time.Now,
 		sleep:     time.Sleep,
@@ -387,6 +389,11 @@ func runUmount(args []string, deps umountDeps) error {
 		if err := deps.terminate(pid, *waitTimeout); err != nil {
 			return fmt.Errorf("%w (pid file: %s)", err, path)
 		}
+		if path != "" && deps.remove != nil {
+			if err := deps.remove(path); err != nil && !errors.Is(err, os.ErrNotExist) {
+				return fmt.Errorf("drive9 umount: remove mount pid file %s: %w", path, err)
+			}
+		}
 		return nil
 	}
 	if *waitTimeout == 0 {
@@ -421,6 +428,8 @@ func processAlive(pid int) bool {
 	return processAliveImpl(pid)
 }
 
+var waitForProcessExit = waitForProcessExitByPID
+
 func terminateProcess(pid int, waitTimeout time.Duration) error {
 	if pid <= 0 {
 		return fmt.Errorf("invalid mount process pid %d", pid)
@@ -430,17 +439,18 @@ func terminateProcess(pid int, waitTimeout time.Duration) error {
 		return err
 	}
 	err = process.Kill()
-	if err == nil {
-		return nil
+	if err != nil {
+		if errors.Is(err, os.ErrProcessDone) || errors.Is(err, syscall.ESRCH) {
+			err = nil
+		} else {
+			return err
+		}
 	}
-	if errors.Is(err, os.ErrProcessDone) || errors.Is(err, syscall.ESRCH) {
-		return nil
+	if waitTimeout > 0 {
+		return waitForProcessExit(pid, waitTimeout)
 	}
 	if err != nil {
 		return err
-	}
-	if waitTimeout > 0 {
-		return waitForProcessExitByPID(pid, waitTimeout)
 	}
 	return nil
 }
--- a/cmd/drive9/cli/mount.go
+++ b/cmd/drive9/cli/mount.go
@@ -313,6 +313,7 @@ type umountDeps struct {
 	lookPath  func(string) (string, error)
 	run       func([]string) error
 	readPID   func(string) (int, string, error)
+	terminate func(int) error
 	pidAlive  func(int) bool
 	now       func() time.Time
 	sleep     func(time.Duration)
@@ -330,6 +331,7 @@ func defaultUmountDeps() umountDeps {
 			return cmd.Run()
 		},
 		readPID:   mountstate.ReadPID,
+		terminate: terminateProcess,
 		pidAlive:  processAlive,
 		now:       time.Now,
 		sleep:     time.Sleep,
@@ -352,6 +354,14 @@ func runUmount(args []string, deps umountDeps) error {
 		return fmt.Errorf("usage: drive9 umount [--timeout duration] <mountpoint>")
 	}
 	mountPoint := fs.Arg(0)
+	stateMountPoint := mountPoint
+	var err error
+	if deps.goos == "windows" {
+		stateMountPoint, err = webdavMountStatePoint(deps.goos, mountPoint)
+		if err != nil {
+			return err
+		}
+	}
 
 	argv, err := umountArgv(deps.goos, deps.lookPath, mountPoint)
 	if err != nil {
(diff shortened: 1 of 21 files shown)


cmd/drive9/cli/mount_webdav.go
```diff
@@ -160,10 +168,12 @@ func webdavMountWithDeps(c *client.Client, mountPoint string, deps webdavMountDe
 	}
 	readyCancel()
 
-	// Ensure mount point exists.
-	if err := os.MkdirAll(mountPoint, 0o755); err != nil {
-		_ = srv.Close()
-		return fmt.Errorf("webdav: create mount point: %w", err)
+	// Ensure mount point exists where the platform expects a directory mount.
+	if deps.goos != "windows" {
```
mashenjun, line 173: This Windows WebDAV path has no external shutdown link back to the bridge process. After a successful `drive9 mount :/ X:`, the process below only shuts the HTTP bridge down from its SIGINT/SIGTERM handler. But the new `drive9 umount X:` path only runs `net use X: /delete /y`; then `runUmount` calls `mountstate.ReadPID(mountPoint)`, and WebDAV mounts never write a PID file, so `ReadPID` returns `os.ErrNotExist` and `runUmount` returns success without signaling this process. The drive letter is unmapped, but the original `drive9 mount` process keeps serving the local WebDAV endpoint until the user manually Ctrl+C's it, which conflicts with the README's `drive9 umount X:` cleanup example. Please add a WebDAV shutdown mechanism for external `drive9 umount` (for example, record a PID/control file after successful WebDAV mount and have `umount` signal it, or otherwise monitor helper unmount state), and cover the Windows path with a test that proves `drive9 umount` causes the bridge process to exit.

mashenjun (changes requested):
Reviewed latest head (`bc4ae6b`). CI is green and the targeted CLI tests/builds I ran locally pass, but I found one lifecycle issue that should be fixed before merge: the new Windows WebDAV unmount flow can leave the bridge process running after `drive9 umount`.

cmd/drive9/cli/mount_webdav.go
```diff
@@ -160,18 +169,42 @@ func webdavMountWithDeps(c *client.Client, mountPoint string, deps webdavMountDe
...
 
+	stateMountPoint, err := webdavMountStatePoint(deps.goos, mountPoint)
+	if err != nil {
+		deps.unmount(deps.goos, mountPoint)
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		defer cancel()
+		_ = srv.Shutdown(ctx)
+		return err
+	}
+	pidFile, err := mountstate.WritePID(stateMountPoint, os.Getpid())
```
mashenjun: Writing the WebDAV PID file unconditionally here changes the non-Windows WebDAV unmount path in a bad way. On macOS `ResolveMountMode(auto, "darwin", ...)` defaults to WebDAV, and Linux can explicitly use `--mode=webdav`; after this write, `drive9 umount <mountpoint>` will run the OS unmount helper and then `runUmount` will find this PID file. But `runUmount` only terminates the bridge process when `deps.goos == "windows"`, so macOS/Linux WebDAV mounts now wait for a bridge process that never receives a shutdown signal and then fail after the timeout. Previously these paths had no PID file, so external `drive9 umount` returned after the filesystem detach. Please either scope this PID file to the Windows WebDAV path, or teach `runUmount` to shut down WebDAV bridge processes consistently on every platform that writes this PID file. A test for darwin or linux `--mode=webdav` umount would catch the regression.

cmd/drive9/cli/mount.go
```diff
@@ -363,16 +370,27 @@ func runUmount(args []string, deps umountDeps) error {
...
+			return fmt.Errorf("drive9 umount: no process terminator configured for Windows WebDAV mount")
+		}
+		if err := deps.terminate(pid); err != nil {
+			return fmt.Errorf("drive9 umount: terminate mount process pid %d: %w", pid, err)
+		}
+	}
+	if *waitTimeout == 0 {
+		return nil
+	}
 	if err := waitForPIDExit(pid, *waitTimeout, deps); err != nil {
```
mashenjun, line 473: This wait does not actually wait on Windows with the default `processAlive`. `processAlive` probes liveness via `process.Signal(syscall.Signal(0))`, but Go's Windows implementation returns `syscall.EWINDOWS` for every signal except `os.Kill`, so `processAlive` falls through to `false` even if the process still exists. That makes `--timeout` ineffective for the new Windows WebDAV cleanup path: after `terminateProcess(pid)` calls `Process.Kill`, `waitForPIDExit` can return immediately without confirming the bridge exited. Please use a Windows-aware liveness check/wait path (for example, wait on the process handle on Windows, or make `terminateProcess` wait until exit) and cover the real default behavior rather than only stubbing `pidAlive` in tests.

mashenjun (changes requested):
Reviewed latest head (`13f8a107`). The previous Windows WebDAV leak is partially addressed, and CI/local targeted checks are green, but the new PID/terminate path still has lifecycle regressions that should be fixed before merge.

cmd/drive9/cli/mount.go
```diff
@@ -399,29 +418,36 @@ func waitForPIDExit(pid int, timeout time.Duration, deps umountDeps) error {
...
+		return fmt.Errorf("invalid mount process pid %d", pid)
 	}
 	process, err := os.FindProcess(pid)
 	if err != nil {
-		return false
+		return err
 	}
-	err = process.Signal(syscall.Signal(0))
+	err = process.Kill()
 	if err == nil {
```
mashenjun: The new Windows wait helper is still bypassed on the common path. When `process.Kill()` succeeds, this branch returns immediately, so `waitForProcessExitByPID(pid, waitTimeout)` below is unreachable for the successful termination case. That leaves `--timeout` ineffective for Windows WebDAV cleanup: `drive9 umount X:` can report success before the bridge process has actually exited. It also means the mount process dies via `Kill`, so its deferred `os.Remove(pidFile)` in `webdavMountWithDeps` will not run and the PID file can be left stale. Please wait after a successful termination request when `waitTimeout > 0`, and have the umount side remove the PID file after confirming exit (or switch to a graceful shutdown path that lets the mount process run its cleanup). Add a test that exercises the default `terminateProcess` behavior rather than only stubbing `deps.terminate`.

mashenjun (changes requested):
Reviewed latest head (`4332fd8`). The WebDAV PID write is now scoped to Windows, which fixes the macOS/Linux regression from the last round. However the Windows terminate/wait path is still not actually waiting on the normal success path, and `code-ci` is currently failing on the Test step, so this still needs changes before merge.

cmd/drive9/cli/mount.go
```diff
@@ -363,16 +372,33 @@ func runUmount(args []string, deps umountDeps) error {
...
 		return err
 	}
-	if *waitTimeout == 0 {
-		return nil
-	}
-	pid, path, err := deps.readPID(mountPoint)
+	// Windows WebDAV mounts run a local bridge in the mount process. Even when
+	// the caller opts out of waiting, still request process shutdown before
+	// returning.
+	pid, path, err := deps.readPID(stateMountPoint)
```
mashenjun: This read now runs for every OS before the non-Windows `--timeout 0` check below. That regresses the existing `drive9 umount --timeout 0` behavior: once the unmount helper succeeds, timeout-zero should return without consulting the drive9 PID file. With this ordering, a corrupt/unreadable stale PID file on Linux/macOS makes `--timeout 0` fail even though waiting was explicitly disabled. Please keep the old early return for non-Windows, e.g. `if deps.goos != "windows" && *waitTimeout == 0 { return nil }`, while still taking the Windows terminate path.

cmd/drive9/cli/mount.go
```diff
@@ -14,26 +14,23 @@ import (
...
-var mountFuse = drive9fuse.Mount
-
 // MountCmd handles the "drive9 mount" command.
 //
 // Dispatch fork (Row A, V2e): the first positional argument selects the
 // mount flavour.
 //
-//   - `drive9 mount vault <path>`   → read-only vault FUSE filesystem
-//   - `drive9 mount [flags] <path>` → legacy writable fs mount (no
+//   - `drive9 mount vault <path>`   鈫?read-only vault FUSE filesystem
```
mashenjun: Non-blocking, but this PR introduced mojibake in comments (`鈫?`, `鈥?`, `搂`) that were previously valid arrows/em dashes/section signs. Please restore the original characters or replace them with ASCII before merging.

mashenjun (changes requested):
Reviewed latest head `d69aff1f18f2f11cda3645e64c7ae9edd6d8b11e`.

The previous Windows WebDAV bridge lifecycle issues I flagged are addressed: Windows mounts write a PID file, `umount` terminates the bridge process, the successful `Kill` path waits when a timeout is configured, and the PID file is cleaned up. However, I found one remaining non-Windows behavior regression in `umount --timeout 0`, so I am requesting changes.

Verification run:
- `go test ./cmd/drive9/cli`
- `go test ./cmd/drive9`
- `GOOS=windows GOARCH=amd64 go build ./cmd/drive9`
- `GOOS=windows GOARCH=arm64 go build ./cmd/drive9`
- `GOOS=windows GOARCH=amd64 go test -c ./cmd/drive9/cli`
- `make build-cli-release DIST_DIR=/private/tmp/drive9-pr415-dist VERSION=review CLI_TARGETS='windows/amd64 windows/arm64'`

GitHub checks are green, but please fix the timeout-zero regression before merge.

cmd/drive9/cli/mount.go
```diff
@@ -355,24 +356,65 @@ func runUmount(args []string, deps umountDeps) error {
...
 		return err
 	}
+	if deps.goos == "windows" {
+		if deps.terminate == nil {
+			if runErr != nil {
+				return runErr
+			}
+			return fmt.Errorf("drive9 umount: no process terminator configured for Windows WebDAV mount")
+		}
+		if err := deps.terminate(pid, *waitTimeout); err != nil {
```
mashenjun: This Windows path now calls `Process.Kill` on whatever integer is in the mount PID file. That is not safe enough for a hard-kill path: `mountstate.WritePID` stores only a bare PID in a deterministic temp-file path, so if the mount process crashes/is killed outside `drive9 umount`, or the file is otherwise stale, a later `drive9 umount X:` can kill an unrelated process after Windows reuses the PID. This is worse than the existing non-Windows PID usage, which only waits on the PID and never terminates it. Please make the PID file/control path prove ownership before killing, e.g. store and validate a per-mount token plus process identity/creation time, or use a local control endpoint owned by the mount process; if validation fails, remove/report the stale PID file without killing. Add a regression test that a stale PID file does not cause `umount` to terminate an arbitrary process.

mashenjun (changes requested):
Reviewed latest head `8e68f7324378e364ec548252d2603838bb84cc15`.

The previous findings are fixed: non-Windows `--timeout 0` now skips PID reads again, and the mojibake comments were cleaned up. CI is green on the latest commit. I found one remaining blocking issue in the Windows `umount` cleanup path: the PID file is now used as authority to kill a process, but it only stores a bare PID and is not validated against the drive9 mount process.

Verification run:
- `go test ./cmd/drive9/cli` (rerun outside sandbox because httptest loopback bind is blocked in sandbox)
- `go test ./cmd/drive9`
- `GOOS=windows GOARCH=amd64 go build ./cmd/drive9`
- `GOOS=windows GOARCH=arm64 go build ./cmd/drive9`
- `GOOS=windows GOARCH=amd64 go test -c ./cmd/drive9/cli`
- `make build-cli-release DIST_DIR=/private/tmp/drive9-pr415-latest-dist VERSION=review CLI_TARGETS='windows/amd64 windows/arm64'`
- `make build-cli-release DIST_DIR=/private/tmp/drive9-pr415-latest-dist-all VERSION=review`

cmd/drive9/cli/mount.go
```diff
@@ -189,12 +191,12 @@ func fsMountCmd(args []string) error {
 	}
 
 	// FUSE path (existing behavior).
-	syncModeVal, err := drive9fuse.ParseSyncMode(*syncMode)
+	syncModeVal, err := parseFuseSyncMode(*syncMode)
```
mashenjun, line 202: On Windows, explicit FUSE mode still resolves credentials before reaching the Windows-only unsupported stub below. A first-time Windows user running `drive9 mount --mode=fuse X:` without a configured context/API key will get `drive9 server URL required` instead of the intended “FUSE is not supported on Windows; use the default WebDAV path” guidance. Since this PR explicitly adds Windows unsupported-mode messaging, please short-circuit `runtime.GOOS == "windows" && resolved == MountModeFUSE` after flag parsing/empty-flag validation and before `resolveMountCredentials`, or add an injectable GOOS path and a full-command test that proves the unsupported guidance wins over missing credentials.

cmd/drive9/cli/mount_vault.go
```diff
@@ -71,5 +69,5 @@ func VaultMountCmd(args []string) error {
 		AllowOther: *allowOther,
 		Debug:      *debug,
 	}
-	return drive9fuse.MountVault(opts)
+	return mountVault(opts)
```
mashenjun, line 81: Same issue for `drive9 mount vault` on Windows: the command resolves credentials before calling `mountVault`, so users without existing drive9 credentials never see the Windows-specific “use `drive9 vault` commands or the vault API” guidance added by the Windows stub. Please make the unsupported Windows vault path fail before credential resolution, while keeping the existing arg parsing and explicit-empty flag validation behavior.

mashenjun (changes requested):
Reviewed latest head `ed252acc86a001c8bfe1a32403808449b750a793`.

The previous Windows WebDAV bridge cleanup issues look addressed: Windows mounts now write process state with ownership metadata, `umount` validates creation time before terminating, stale/unsafe state is handled without killing arbitrary processes, and the non-Windows timeout-zero behavior remains covered. I found one remaining user-facing issue in the Windows unsupported-mode paths, so I am requesting changes.

Validation run:
- `go test ./cmd/drive9/cli`
- `go test ./pkg/mountstate`
- `go test ./cmd/drive9`
- `GOCACHE=/private/tmp/gocache-pr415-review GOOS=windows GOARCH=amd64 go build ./cmd/drive9`
- `GOCACHE=/private/tmp/gocache-pr415-review GOOS=windows GOARCH=arm64 go build ./cmd/drive9`
- `GOCACHE=/private/tmp/gocache-pr415-review GOOS=windows GOARCH=amd64 go test -c ./cmd/drive9/cli`
- `GOCACHE=/private/tmp/gocache-pr415-review make build-cli-release DIST_DIR=/private/tmp/drive9-pr415-review-dist VERSION=review CLI_TARGETS="windows/amd64 windows/arm64"`

CI on the latest commit is also green for `code-ci` and `local-e2e`; CodeRabbit was still pending when I checked.

cmd/drive9/cli/process_wait_windows.go
```diff
@@ -0,0 +1,159 @@
...
+func terminateMountProcess(state mountstate.ProcessState, waitTimeout time.Duration) error {
+	if state.PID <= 0 {
+		return fmt.Errorf("invalid mount process pid %d", state.PID)
+	}
+	if state.CreationTime == 0 {
+		return fmt.Errorf("%w: mount pid file for pid %d is missing ownership metadata", errMountProcessStateUnsafe, state.PID)
+	}
+
+	handle, err := windows.OpenProcess(windows.SYNCHRONIZE|windows.PROCESS_TERMINATE|windows.PROCESS_QUERY_LIMITED_INFORMATION, false, uint32(state.PID))
+	if err != nil {
```
mashenjun, line 98: `OpenProcess` failure is not always stale state. If the recorded drive9 mount process is still alive but the caller cannot open it with `PROCESS_TERMINATE`/query rights, for example an elevated mount process being unmounted from a non-elevated shell, this branch wraps the error as `errMountProcessStateStale`. `runUmount` then removes the PID file and returns success for stale state, leaving the WebDAV bridge running with no remaining state file to clean it up later. Please only classify known nonexistent-PID errors such as `windows.ERROR_INVALID_PARAMETER` as stale here; for access/permission failures, return a normal error and keep the PID file so the user can retry from a shell with sufficient privileges.

mashenjun (changes requested):
Reviewed latest head `85c598988fcf608d5a060576e229a16085c3fc43`.

The previous unsupported-mode guidance issue is fixed: Windows `--mode=fuse` and `mount vault` now fail before credential resolution, and the new Windows build-tag tests cover that behavior. The earlier PID ownership and wait-path issues are also mostly addressed. I found one remaining Windows cleanup edge case where an access-denied process-open error is classified as stale state and can remove the PID file while the bridge is still alive.

Validation run:
- `go test ./cmd/drive9/cli`
- `go test ./pkg/mountstate`
- `go test ./cmd/drive9`
- `GOCACHE=/private/tmp/gocache-pr415-review-latest GOOS=windows GOARCH=amd64 go build ./cmd/drive9`
- `GOCACHE=/private/tmp/gocache-pr415-review-latest GOOS=windows GOARCH=arm64 go build ./cmd/drive9`
- `GOCACHE=/private/tmp/gocache-pr415-review-latest GOOS=windows GOARCH=amd64 go test -c ./cmd/drive9/cli`
- `GOCACHE=/private/tmp/gocache-pr415-review-latest make build-cli-release DIST_DIR=/private/tmp/drive9-pr415-review-latest-dist VERSION=review CLI_TARGETS="windows/amd64 windows/arm64"`

GitHub `code-ci` and `local-e2e` are green on this commit; CodeRabbit status is success.

mashenjun (approved):
Reviewed latest head `840f3ed76e4883a637bf2a4b938de3b3dc07ba45`.

No blocking findings in this pass. The previous Windows cleanup blocker is fixed: `terminateMountProcess` now only treats `windows.ERROR_INVALID_PARAMETER` from `OpenProcess` as stale state, and returns a normal inspect error for permission/access failures so `runUmount` keeps the PID file for a privileged retry. The new `TestRunUmountWindowsPermissionErrorKeepsPIDFile` covers that behavior.

I also rechecked the surrounding Windows mount/umount flow, unsupported Windows FUSE/vault guidance before credential resolution, WebDAV PID-state lifecycle, and Windows release artifact naming.

Validation run:
- `go test ./cmd/drive9/cli`
- `go test ./pkg/mountstate`
- `go test ./cmd/drive9`
- `GOCACHE=/private/tmp/gocache-pr415-review-current GOOS=windows GOARCH=amd64 go build ./cmd/drive9`
- `GOCACHE=/private/tmp/gocache-pr415-review-current GOOS=windows GOARCH=arm64 go build ./cmd/drive9`
- `GOCACHE=/private/tmp/gocache-pr415-review-current GOOS=windows GOARCH=amd64 go test -c ./cmd/drive9/cli`
- `GOCACHE=/private/tmp/gocache-pr415-review-current make build-cli-release DIST_DIR=/private/tmp/drive9-pr415-review-current-dist VERSION=review CLI_TARGETS="windows/amd64 windows/arm64"`

GitHub `code-ci` and `local-e2e` are green on this commit; CodeRabbit status is success. I did not run a live Windows mount in this pass, so that part still relies on the PR author's reported real Windows validation.

PR #416, opened 2026-05-11
feat: add ctx show and unify bare ctx output

srstack:
## Summary
- add `drive9 ctx show` with masked-by-default text output plus `--json` and `--reveal`
- make bare `drive9 ctx` use the same default text output as `drive9 ctx show`
- update help text, tests, and ctx/vault interaction spec to match the converged CLI contract

## Testing
- `go test ./cmd/drive9/cli ./cmd/drive9`

### DIFF
--- a/cmd/drive9/cli/ctx.go
+++ b/cmd/drive9/cli/ctx.go
@@ -13,16 +13,16 @@ import (
 
 // Ctx dispatches drive9 ctx subcommands per spec §13.2.
 //
-// The 5 normative verbs are: add / import / ls / use / rm.
+// The user-facing verbs are: show / add / import / ls / use / rm.
 //
-// Bare `drive9 ctx` (no verb) is retained as a non-spec compatibility
-// convenience that prints the current context name. See migration call-out
-// #7 in the impl PR body.
+// Bare `drive9 ctx` is the shorthand form of `drive9 ctx show`.
 func Ctx(args []string) error {
 	if len(args) == 0 {
-		return ctxShow()
+		return ctxShowCmd(nil)
 	}
 	switch args[0] {
+	case "show":
+		return ctxShowCmd(args[1:])
 	case "add":
 		return ctxAddCmd(args[1:])
 	case "import":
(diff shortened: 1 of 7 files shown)


cmd/drive9/cli/ctx.go
```diff
@@ -55,14 +56,182 @@ func ctxUsageErr() error {
...
+type ctxShowEntry struct {
+	Name      string    `json:"name"`
+	Type      string    `json:"type"`
+	Server    string    `json:"server,omitempty"`
+	APIKey    string    `json:"api_key,omitempty"`
+	Token     string    `json:"token,omitempty"`
+	Agent     string    `json:"agent,omitempty"`
+	Scope     []string  `json:"scope,omitempty"`
+	Perm      string    `json:"perm,omitempty"`
+	ExpiresAt time.Time `json:"expires_at,omitempty"`
```
mashenjun: `omitempty` does not omit a zero `time.Time` here, so owner contexts now serialize a bogus `"expires_at": "0001-01-01T00:00:00Z"` in `drive9 ctx show --json` even though owner contexts do not have an expiry in the §13.1 schema. I reproduced it with `ctx add --api-key ...` followed by `ctx show --json --reveal`. Please make this field nullable/omittable for JSON, for example `*time.Time` populated only for delegated contexts or a custom JSON shape, and add an owner JSON assertion that `expires_at` is absent.

mashenjun (changes requested):
Reviewed the current head (`531b913`). The targeted tests pass locally, but I found one JSON contract bug that should be fixed before merge.

mashenjun (approved):
Re-reviewed latest head (`d473c67`). The previous `expires_at` JSON contract issue is fixed: owner `ctx show --json --reveal` now omits `expires_at`, and the PR adds a regression assertion for that. I also reran `go test ./cmd/drive9/cli ./cmd/drive9` locally and checked the latest GitHub Actions runs (`code-ci`, `local-e2e`), all passing.

PR #417, opened 2026-05-12
Optimize FUSE handle hot paths

srstack:
## Summary
- fix mount option default handling so interactive profile TTLs and FUSE lookup retry defaults take effect unless explicitly overridden
- add an open-handle index and use it for hot path handle/path checks instead of full handle-table scans
- retarget open FUSE handles, write buffers, stream uploaders, and prefetchers after local rename
- reduce handle table lookup contention with RWMutex and skip the fuseCtx goroutine when cancel is nil

## Tests
- go test ./pkg/fuse ./cmd/drive9/cli
- go test -race ./pkg/fuse -run 'Test(FuseCtx|OpenHandleIndex|Dat9FSFinishLocalRenameUpdatesOpenHandleIndex)'\n- git diff --check

## Summary by CodeRabbit

* **Bug Fixes**
  * Mount flags now distinguish omitted vs explicitly provided lookup-retry and TTL values
  * Improved context cancellation and rename handling to keep open-file state consistent
  * PID/process-state files are now written atomically

* **New Features**
  * Track open writable handles in a dedicated index for accurate lookups and retargeting
  * Prefetcher and uploaders can safely update target paths at runtime

* **Performance**
  * Concurrent-read locking for handle tables and faster open-handle lookups

* **Tests**
  * Added tests covering TTL defaults, lookup-retry validation, open-handle index, prefetch/rename, and atomic PID writes

[Review Change Stack](https://app.coderabbit.ai/change-stack/mem9-ai/drive9/pull/417)

### DIFF
--- a/cmd/drive9/cli/mount.go
+++ b/cmd/drive9/cli/mount.go
@@ -19,6 +19,11 @@ import (
 	drive9webdav "github.com/mem9-ai/dat9/pkg/webdav"
 )
 
+const (
+	defaultFuseLookupRetryCount   = 3
+	defaultFuseLookupRetryTimeout = 2 * time.Second
+)
+
 var (
 	errMountProcessStateStale  = errors.New("drive9 umount: stale mount process state")
 	errMountProcessStateUnsafe = errors.New("drive9 umount: unsafe mount process state")
@@ -83,8 +88,8 @@ func fsMountCmd(args []string) error {
 	attrTTL := fs.Duration("attr-ttl", 10*time.Second, "kernel attr cache TTL")
 	entryTTL := fs.Duration("entry-ttl", 10*time.Second, "kernel entry cache TTL")
 	flushDebounce := fs.Duration("flush-debounce", -1, "debounce window for small-file flush coalescing (default 2s, 0 disables)")
-	lookupRetryCount := fs.Int("lookup-retry-count", 0, "detached retries after transient Lookup/GetAttr stat failures (omit for default, set 0 to disable)")
-	lookupRetryTimeout := fs.Duration("lookup-retry-timeout", 0, "timeout per detached Lookup/GetAttr stat retry (omit for default, must be > 0 when set)")
+	lookupRetryCount := fs.Int("lookup-retry-count", defaultFuseLookupRetryCount, "detached retries after transient Lookup/GetAttr stat failures (set 0 to disable)")
+	lookupRetryTimeout := fs.Duration("lookup-retry-timeout", defaultFuseLookupRetryTimeout, "timeout per detached Lookup/GetAttr stat retry (must be > 0 when set)")
 	legacyDirStatFallback := fs.Bool("legacy-dir-stat-fallback", false, "on Lookup stat 404, list parent to support legacy servers without directory stat")
 	readDirPrefetch := fs.Bool("readdir-prefetch", false, "prefetch small files after directory reads into the read cache")
 	prefetchMaxFiles := fs.Int("readdir-prefetch-max-files", 32, "maximum small files prefetched per directory read")
--- a/cmd/drive9/cli/mount.go
+++ b/cmd/drive9/cli/mount.go
@@ -83,8 +83,8 @@ func fsMountCmd(args []string) error {
 	attrTTL := fs.Duration("attr-ttl", 10*time.Second, "kernel attr cache TTL")
 	entryTTL := fs.Duration("entry-ttl", 10*time.Second, "kernel entry cache TTL")
 	flushDebounce := fs.Duration("flush-debounce", -1, "debounce window for small-file flush coalescing (default 2s, 0 disables)")
-	lookupRetryCount := fs.Int("lookup-retry-count", 2, "detached retries after transient Lookup/GetAttr stat failures (default 2, set 0 to disable)")
-	lookupRetryTimeout := fs.Duration("lookup-retry-timeout", 250*time.Millisecond, "timeout per detached Lookup/GetAttr stat retry (default 250ms, must be > 0)")
+	lookupRetryCount := fs.Int("lookup-retry-count", 0, "detached retries after transient Lookup/GetAttr stat failures (omit for default, set 0 to disable)")
+	lookupRetryTimeout := fs.Duration("lookup-retry-timeout", 0, "timeout per detached Lookup/GetAttr stat retry (omit for default, must be > 0 when set)")
 	legacyDirStatFallback := fs.Bool("legacy-dir-stat-fallback", false, "on Lookup stat 404, list parent to support legacy servers without directory stat")
 	readDirPrefetch := fs.Bool("readdir-prefetch", false, "prefetch small files after directory reads into the read cache")
 	prefetchMaxFiles := fs.Int("readdir-prefetch-max-files", 32, "maximum small files prefetched per directory read")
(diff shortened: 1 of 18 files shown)


qiffang (changes requested):
REVISE. I found one correctness blocker in the rename retargeting path, plus the current CI failure needs to be resolved before merge.

## Blocking findings

1. `retargetOpenHandlesForRename` updates `fh.Path`, `fh.Dirty.path`, `Streamer`, and `Prefetcher`, but it does not update the lazy `WriteBuffer.LoadPart` callback installed by `preloadWritableHandle`.

The callback captures both `filePath` and `remoteFilePath` at open time (`pkg/fuse/dat9fs.go`, around the `fh.Dirty.LoadPart = ...` setup). After renaming an open writable handle, a later lazy load for an untouched part still calls `ReadStreamRange` on the old remote path. Failure scenario:

```text
open /old large file for write -> Dirty.LoadPart captures /old
rename /old -> /new while handle remains open
write/read an unloaded old part
ensurePart invokes LoadPart -> reads /old, not /new
```

That can return 404 after the server-side rename, or worse, read wrong bytes if `/old` is recreated. The new test only checks `Dirty.path`, `Streamer`, and `Prefetcher`; it does not exercise lazy part loading after rename. Please retarget the lazy loader as well, or make it resolve the current handle/buffer path dynamically, and add a regression that proves a post-rename lazy load reads the new remote path.

2. CI is red and the failing test is reproducible on this branch.

GitHub `ci` fails in `TestWebDAVMountLifecycleWindowsNormalizesDriveLetter` with an empty PID file. I reproduced locally with:

```bash
/usr/local/go/bin/go test ./cmd/drive9/cli -run TestWebDAVMountLifecycleWindowsNormalizesDriveLetter -count=20
```

The same command passed for current `origin/main` in my local worktree, so this PR needs either a fix or a clear re-run showing the failure is unrelated and gone. Do not merge with red CI.

## Review note

CodeRabbit flagged `OpenHandleIndex.Has(ino, path)` because it uses OR semantics when both arguments are supplied. I would not blindly change this to intersection without checking callers: the old scan-based `hasOpenHandle` also used `fh.Ino == ino || fh.Path == path`, and these callers appear to use it as a conservative preservation check. If OR is the intended contract, please make that explicit with a name/comment and a test for the both-keys case so future reviews do not “fix” it into a stricter semantic by accident.

## Validation run

Passed locally:

```bash
/usr/local/go/bin/go test ./pkg/fuse ./cmd/drive9/cli
/usr/local/go/bin/go test -race ./pkg/fuse -run 'Test(FuseCtx|OpenHandleIndex|Dat9FSFinishLocalRenameUpdatesOpenHandleIndex)'
/usr/local/go/bin/go test -race ./pkg/fuse
/usr/local/go/bin/go test -c ./pkg/fuse ./cmd/drive9/cli ./pkg/client ./pkg/server
```

Failed locally:

```bash
/usr/local/go/bin/go test ./cmd/drive9/cli -run TestWebDAVMountLifecycleWindowsNormalizesDriveLetter -count=20
```

srstack:
Addressed the review blockers in commit `2f64d67`:

- Made lazy `WriteBuffer.LoadPart` resolve the current handle path at load time, with a regression covering lazy load after rename.
- Kept `OpenHandleIndex.Has` OR semantics explicit and added a both-keys test.
- Restored displayed lookup retry defaults while preserving omitted-flag normalization.
- Made mount PID/process-state writes atomic to avoid empty PID files during concurrent lifecycle reads.

Local validation:
- `go test ./pkg/mountstate ./cmd/drive9/cli ./pkg/fuse`
- `go test ./cmd/drive9/cli -run TestWebDAVMountLifecycleWindowsNormalizesDriveLetter -count=100`
- `go test -race ./pkg/fuse -run 'Test(FuseCtx|OpenHandleIndex|Dat9FSFinishLocalRenameUpdatesOpenHandleIndex|LazyWritablePreloadUsesRenamedPath)'`\n- `git diff --check`

qiffang (approved):
GREEN on the updated head `2f64d67`.

The previous blockers are resolved:

- The lazy writable preload callback now resolves `fh.Path` at load time, so an open handle renamed before lazy loading reads from the renamed remote path. The new `TestLazyWritablePreloadUsesRenamedPath` regression covers this exact path.
- The WebDAV mount PID/process-state write now uses an atomic temp-file + rename flow. The previously failing `TestWebDAVMountLifecycleWindowsNormalizesDriveLetter` is now stable locally and GitHub `ci` is green.
- `OpenHandleIndex.Has` now documents the conservative OR semantics and has a both-keys test, so the CodeRabbit comment is handled without changing caller semantics.

Validation run:

```bash
/usr/local/go/bin/go test ./pkg/mountstate ./cmd/drive9/cli ./pkg/fuse
/usr/local/go/bin/go test ./cmd/drive9/cli -run TestWebDAVMountLifecycleWindowsNormalizesDriveLetter -count=100
/usr/local/go/bin/go test -race ./pkg/fuse -run 'Test(FuseCtx|OpenHandleIndex|Dat9FSFinishLocalRenameUpdatesOpenHandleIndex|LazyWritablePreloadUsesRenamedPath)'
/usr/local/go/bin/go test -race ./pkg/fuse
/usr/local/go/bin/go test -race ./pkg/mountstate
/usr/local/go/bin/go test -c ./pkg/fuse ./cmd/drive9/cli ./pkg/client ./pkg/server
```

GitHub checks are green: `ci`, `drive9-server-local e2e smoke`, and CodeRabbit.

srstack:
Addressed the latest Copilot comments in `06a0da2`:

- Removed `Chmod` from PID/state atomic writes so final permissions honor the caller umask like `os.WriteFile(..., 0644)`.
- Added platform-specific replacement: Unix keeps atomic `rename`, Windows uses `MoveFileEx(..., MOVEFILE_REPLACE_EXISTING|MOVEFILE_WRITE_THROUGH)` so existing PID files are overwritten.
- Added a Unix umask regression test and verified Windows package compilation.

Validation:
- `go test ./pkg/mountstate ./cmd/drive9/cli ./pkg/fuse`
- `go test ./cmd/drive9/cli -run TestWebDAVMountLifecycleWindowsNormalizesDriveLetter -count=100`
- `GOOS=windows go test -c -o /tmp/drive9-mountstate.test.exe ./pkg/mountstate`
- `git diff --check`

qiffang (commented):

**Verdict: GREEN** on head `2f64d67`.

This PR delivers four well-scoped optimizations for FUSE handle hot paths. The changes are clean, well-tested, and address real performance bottlenecks.

---

### What this PR does

1. **OpenHandleIndex** (`open_handles.go`, 170 lines) — O(1) handle lookup by inode/path, replacing full `HandleTable.Snapshot()` scans. Three maps (`byInode`, `byPath`, `pathByHandle`) with proper cleanup on Remove. This is the core win — eliminates linear scans on every Lookup, GetAttr, Forget, and rename.

2. **Rename retargeting** (`retargetOpenHandlesForRename`) — Open handles, dirty buffers, stream uploaders, and prefetchers are all atomically retargeted when a file is renamed. The lazy `LoadPart` callback now resolves `fh.Path` at load time (not capture time), so a handle renamed before lazy loading reads from the correct remote path.

3. **Prefetcher atomic path** (`prefetch.go`) — `path` field changed from `string` to `atomic.Value` with `SetPath()`. This ensures concurrent prefetch goroutines see path updates without holding `fh.mu`. The `startPrefetch` goroutine snapshots `fetchPath` before launching, which is correct — inflight fetches read from the path that was valid when they started.

4. **Profile TTL defaults** (`mount.go`) — `ApplyInteractiveProfile` is now called *before* generic defaults in `setDefaults()`, so zero-value TTLs from CLI (meaning "use profile defaults") take the profile values instead of being overwritten by generic 10s defaults. Good fix for the interactive profile UX.

5. **HandleTable RWMutex** (`handle.go`) — `sync.Mutex` → `sync.RWMutex` for `Get`/`Snapshot`, reducing contention on read-heavy paths.

6. **Atomic PID writes** (`mountstate/pid.go`) — `writeFileAtomic` with temp file + rename pattern, platform-specific `replaceFile` for Windows. Prevents corrupted PID files on crash during write.

---

### Key design notes

- **`Has()` OR semantics is intentional.** CodeRabbit flagged the OR behavior in `Has(ino, p)` as a potential issue, but the code comment explicitly states "conservative inode/path checks" — matching the pre-index behavior where the old table scan returned true if *any* handle matched either field. This is correct for `shouldPreserveForgottenInode` which must be conservative to avoid premature inode cleanup.

- **`durationFlagValue` returns 0 when flag not provided.** This means unprovided TTL flags pass 0 to `MountOptions`, which `setDefaults()` then fills with profile or generic defaults. Clean separation between "user didn't set this" vs "user explicitly set this to X".

---

1. **R1: `Remove` fallback scan** — When `pathByHandle[fh]` misses (line 631-636), `Remove` falls back to scanning all paths in `byPath`. This should be rare (only if the index is somehow inconsistent), but consider logging a warning here since it indicates a bug if it ever triggers.

2. **R2: `createTempFile` retry count** — The 100-iteration retry loop in `createTempFile` is generous. Consider making it a named constant for clarity, though this is a minor style point.

3. **R3: Test coverage for `fuseCtx(nil)`** — Good addition of `TestFuseCtxNilCancel` and `TestFuseCtxCancelChannel`. The nil-cancel fast path avoids a goroutine leak per FUSE op, which matters under high QPS.

PR #418, opened 2026-05-12
feat(schema): split-table refactor with POSIX permissions and legacy-table skip

mornyx:
## Summary

This PR completes the metadata schema refactor (Phase A) that splits the monolithic `files` table into `inodes`, `contents`, and `semantic` while maintaining full rollback safety for existing tenants.

### Schema Changes
- **New tables**: `inodes`, `contents`, `semantic` across all providers (`tidb_app`, `tidb_auto`, `db9`)
- **`mode` column** added to `inodes` for POSIX permission storage (default 0644 files, 0755 dirs)
- **`file_nodes.inode_id`** now links directories into `inodes` for unified permission storage
- **TiDB auto-embedding**: generated embedding columns moved from `files` to `semantic` table

### Read Paths
- All ~26 read sites switched from `files` to JOIN `inodes`+`contents`+`semantic`
- All queries from `file_nodes` to `inodes` use `COALESCE(fn.inode_id, fn.file_id)` to support both old and new data
- Search, vector, and FTS queries updated to join split tables

### Write Paths
- **Dual-write tx**: `InsertFile`, `UpdateFileContent`, `ConfirmPendingFile`, `MarkFileDeleted`, `RenameFileReplacingTarget`, `DeleteFileWithRefCheck`, `DeleteDirRecursive`, embedding writeback, and search-text updates all atomically write to both `files` and split tables
- **`useLegacyFiles` auto-detection**: at `Open()` time the store detects whether the `files` table exists. New tenants (no `files` table) skip all legacy-table writes entirely, while old tenants continue dual-writing for rollback safety

### POSIX Permissions (Phase 4)
- `Chmod`: full stack from CLI → client → server → backend → datastore → FUSE
- `Mkdir`: `MkdirCtx` accepts `perm uint32`; backend creates inode records for directories
- `EnsureParentDirs`: creates inode rows for parent directories (mode 0755)
- `backend.Stat` reads mode from `inodes` table (fallback 0644/0755)
- FUSE `fillAttr` uses `entry.Mode`; `SetAttr` handles mode-only updates and skips remote call for open-created files with dirty handles
- `Lookup` / `GetAttr` refresh mode via `X-Dat9-Mode` header

### Migration
- Idempotent `SplitTablesMigrator` in `pkg/migrate` backfills data from `files` to split tables
- `createDirInodes` backfills `file_nodes.inode_id = node_id` for migrated directories
- Migration is safe to re-run

### Tests
- Datastore: Chmod, ConfirmPendingFile, DeletePendingFile, mode default/round-trip, and new-tenant without-legacy-table tests
- Backend/client: chmod and mkdir perm tests
- FUSE: inode mode, fillAttr, Lookup mode-header tests
- Migrate: split-tables migration tests with generated-column detection

### Rollback Safety
- Old binaries reading `files` table continue working unchanged
- New binaries read only split tables
- `files` table remains as a permanent compatibility layer for existing tenants until Phase B (future DROP)

---

Closes: 

## Summary by CodeRabbit

* **New Features**
  * End-to-end chmod support (CLI, HTTP API, client, FUSE, WebDAV); Mkdir accepts explicit permission bits. Online migrator and runtime split-table inode/content/semantic storage with inode-aware APIs.

* **Bug Fixes**
  * Consistent propagation, display, and round-tripping of permission bits across listings, stats, mounts, and headers.

* **Tests**
  * Expanded tests for chmod, mkdir modes, inode behaviors, search, and migration.

* **Chores**
  * Test and tenant schemas updated for inode-backed split-table storage.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mem9-ai/drive9/pull/418)


srstack (changes requested):
Findings:

- P1 `pkg/fuse/dat9fs.go:1620`: `chmod 000` cannot be reloaded from the server. `StatCtx` parses `X-Dat9-Mode: 0`, but FUSE `Lookup` only calls `UpdateMode` when `stat.Mode != 0`, so after remount/cache miss a valid `000` mode is treated as unset and `fillAttr` falls back to `0644/0755`. Please carry mode header presence through the client stat result, or otherwise distinguish `0` from unset.

- P1 `pkg/server/server.go:1449`: mkdir mode is not actually preserved through the HTTP/FUSE path. Backend `MkdirCtx(path, perm)` supports a permission argument, but the server always passes `0o755`, client `MkdirCtx` has no mode parameter, and FUSE `Mkdir` does not use `input.Mode`. So `mkdir -m 700 dir` or umask-derived modes are stored remotely as 0755.

- P2 `pkg/datastore/store.go:238`: `ListNodes` selects 7 columns but still calls `scanNode`, which now scans 8 fields after adding `inode_id`. Any caller of this public method will fail with a scan arity error. Please include `inode_id` in the SELECT and add a regression test for `ListNodes`.

- P2 `pkg/backend/dat9.go:734`: list/readdir still drops stored mode. `ListDir` now reads `i.mode`, but backend `ReadDir` initializes `filesystem.FileInfo.Mode` to default `0644/0755` and never applies `e.File.Mode`; server list JSON/client `FileInfo` and batch-stat also do not expose mode. As a result, FUSE `readdir`/`readdirplus` returns default permissions until a separate lookup/stat refreshes them.

Verification run locally on a temporary worktree:

- `go test ./pkg/datastore -run TestInsertFileWithoutLegacyTable`
- `go test ./pkg/backend -run 'Test.*Chmod|Test.*Mkdir'`
- `go test ./pkg/fuse -run 'Test.*Mode|Test.*Chmod|Test.*Lookup'`

srstack (changes requested):
Second pass findings:

- P1 `pkg/tenant/pool.go:419`: split-table migration will run on every tenant backend open. The guard is only `store.HasLegacyFiles()`, but this PR still creates/keeps `files` in the tenant schemas for compatibility, so old-compatible tenants keep hitting `migrateSplitTables`. `Run()` then repeatedly does `INSERT IGNORE ... SELECT ... FROM files` plus directory/shared-column backfills. Large tenants will pay full-table migration scans on process restart, cache miss, or tenant invalidation. Please add a persistent migration marker/version or a cheap completeness check before running the full migration.

- P1 `pkg/backend/dat9.go:743`: `chmod 000` is still lost through list/readdir/batch-stat. HEAD now has `HasMode`, but list paths use `if e.File.Mode != 0`, JSON uses `mode,omitempty`, and FUSE batch stat applies mode only when `result.Mode != 0` (`pkg/fuse/dat9fs.go:2576`). A valid `000` becomes “unset” and falls back to `0644/0755`. Please propagate a `HasMode`/pointer-style signal for list and batch-stat as well, not just HEAD.

- P2 `pkg/fuse/dat9fs.go:3755`: deferred chmod after dirty-handle `SetAttr` is dropped if flush fails, but `SetAttr` already updated the in-memory inode mode. After a failed release, local `GetAttr` can keep reporting a mode that was never persisted. Please revert/reload mode when `flushStatus != OK`, or avoid advancing authoritative cached mode until deferred chmod succeeds.

Verification run locally on a temporary worktree:

- `go test ./pkg/datastore -run 'TestListNodes|TestInsertFileWithoutLegacyTable|TestChmod'`
- `go test ./pkg/backend -run 'Test.*Chmod|Test.*Mkdir|Test.*ReadDir'`
- `go test ./pkg/fuse -run 'Test.*Mode|Test.*Chmod|Test.*Lookup|Test.*Mkdir'`
- `go test ./pkg/client -run 'Test.*Mkdir|Test.*Chmod|Test.*Mode'`
- `go test ./pkg/server -run 'Test.*Mkdir|Test.*Chmod|Test.*BatchStat|Test.*Stat'`
- `go test ./pkg/webdav`
- `go test ./pkg/tenant/schema -run 'Test.*Schema|Test.*TiDB|Test.*Optional|Test.*Parse'`

srstack (dismissed):
Third pass findings on `2734e475`:

- P1 `pkg/tenant/schema/tidb_app.go:206` / `pkg/tenant/schema/tidb_auto.go:245`: the new `semantic` FTS/VECTOR indexes are still created with bare `CREATE FULLTEXT|VECTOR INDEX`, while the existing `files` indexes use `ALTER TABLE ... ADD ... ADD_COLUMNAR_REPLICA_ON_DEMAND`. The read paths in this PR now query `semantic`, so these indexes need the same on-demand columnar replica clause; otherwise new split-table tenants can end up without the TiFlash/columnar replicas required for reliable FTS/vector behavior. Please switch the four `idx_semantic_*` statements in both TiDB schema modes to the same `ALTER TABLE semantic ADD ... ADD_COLUMNAR_REPLICA_ON_DEMAND` pattern and add schema tests that cover the semantic indexes, not just `files`.

- P2 `pkg/backend/dat9.go:985`: `Dat9Backend.Stat` still treats `Mode == 0` as “unset”. After `ChmodCtx(..., 000)`, direct backend callers get `0644` (or `0755`) back from `Stat`, even though the datastore/server/FUSE paths now preserve `000`. This should use the explicit mode signal from the split-table stat result, or otherwise avoid defaulting when the inode row supplied a real mode.

- P2 `pkg/fuse/dat9fs.go:1119`: `cachedFileInfos` drops `Mode` and `HasMode` from the `ListCtx` response. Normal `ReadDir` happens to repair this with `BatchStat`, but if batch-stat fails or the legacy lookup list-fallback populates the directory cache, the mode returned by list is discarded and `chmod 000`/custom mkdir perms fall back to defaults. Please copy `item.Mode`/`item.HasMode` into `CachedFileInfo` and apply the same signal in the lookup-list fallback before `fillEntryOut`.

Verification run on a temporary worktree:

- `go test ./pkg/datastore -run 'TestListNodes|TestInsertFileWithoutLegacyTable|TestChmod'`
- `go test ./pkg/backend -run 'Test.*Chmod|Test.*Mkdir|Test.*ReadDir'`
- `go test ./pkg/fuse -run 'Test.*Mode|Test.*Chmod|Test.*Lookup|Test.*Mkdir|Test.*Release|Test.*SetAttr'`
- `go test ./pkg/client -run 'Test.*Mkdir|Test.*Chmod|Test.*Mode'`
- `go test ./pkg/server -run 'Test.*Mkdir|Test.*Chmod|Test.*BatchStat|Test.*Stat'`
- `go test ./pkg/migrate`
- `go test ./pkg/tenant/schema -run 'Test.*Schema|Test.*TiDB|Test.*Optional|Test.*Parse'`
- `go test ./pkg/tenant -run 'Test.*migrate|Test.*Pool|Test.*Schema'`

qiffang (changes requested):
Verdict: REVISE.

I re-reviewed head `2734e47`. The previous mode/chmod fixes are much closer, but I still see two correctness blockers before this can merge.

1. `pkg/migrate/split_tables.go`: `SplitTablesMigrator.isComplete` can incorrectly skip a partially completed migration. `Run` commits each step independently. If `migrateInodes` succeeds and `migrateContents` or `migrateSemantic` fails, the next backend open sees no pending `file_nodes.inode_id` work and `EXISTS(inodes)=true`, returns complete, and never backfills the missing `contents`/`semantic` rows. Since the new read paths use the split tables, a transient migration failure can leave active legacy files unreadable or missing semantic data permanently. Please make the completeness check prove that every active legacy file has the required `inodes`, `contents`, and `semantic` rows plus directory inode backfill, or remove this optimization and rely on idempotent migration steps. Add a regression that simulates partial migration state and verifies rerun repairs it.

2. `pkg/server/server.go`: `handleStat` only emits `X-Dat9-Mode` when `nf.File != nil`, so directory `HEAD`/stat responses still drop the stored mode. `MkdirCtx(..., 0700)` / directory chmod can be stored and listed, but `HEAD /dir/` and FUSE lookup after remount fall back to default directory permissions because the actual server never returns the mode header for directories. Please emit mode whenever `nf.HasMode` is true, for files and directories, and add a server/client/FUSE regression for directory stat with a non-default mode such as `0700` or `000`.

One additional compatibility point to clarify/fix: `Pool.createBackend` runs the split migrator for any tenant with a legacy `files` table, but the schema repair path only runs for TiDB auto providers. If existing app-managed/local legacy tenants are supported by this PR, the migration path still assumes `file_nodes.inode_id`, `inodes`, `contents`, and `semantic` already exist and can fail before it can migrate. If those tenants are intentionally unsupported, please gate that explicitly; otherwise run the required DDL/schema repair before invoking the migrator.

Validation run locally:
- `go test -c ./pkg/... ./cmd/...` passed
- targeted FUSE mode/chmod/mkdir tests passed
- targeted CLI chmod/fs/mkdir tests passed
- `pkg/client` / tenant schema DB-backed tests are still locally blocked by missing rootless Docker/testcontainers; GitHub checks are green

qiffang (changes requested):
Verdict: REVISE.

I re-reviewed latest head `3e1e12f`. The previous two correctness issues are mostly addressed: partial migration rerun no longer gets skipped by the old false-positive `isComplete`, and directory stat now emits `X-Dat9-Mode` when `nf.HasMode` is true. I still see two blockers in the updated implementation.

1. `EnsureParentDirs` now leaks orphan `inodes` on every call for an already-existing parent directory. Both `EnsureParentDirs` and `EnsureParentDirsTx` insert a fresh random inode first, then insert `file_nodes ... ON DUPLICATE KEY UPDATE node_id = node_id`. When the directory path already exists, the file_nodes insert is a no-op but the newly inserted inode remains unreferenced. This happens not only on concurrent races, but on normal repeated writes under an existing directory. Please make parent-dir creation atomic/idempotent as a unit: only create the inode if the directory node is actually inserted, or wrap the node+inode creation so duplicate path does not leave an extra inode. Add a regression that calls `EnsureParentDirs` twice for the same path and asserts no extra unreferenced directory inode is created.

2. The split migrator now runs the full idempotent migration on every backend open for every legacy tenant. `Pool.createBackend` calls `migrateSplitTables` whenever `store.HasLegacyFiles()` is true, and legacy tenants keep the `files` table by design. Since `Run` no longer has a safe completion check, every open does full `INSERT IGNORE ... SELECT FROM files` passes over `inodes`, `contents`, `semantic`, directory inodes, and shared tables. That fixes the previous false-positive skip, but reintroduces the earlier repeated full-table migration problem. Please add a safe completeness check based on missing split rows / missing inode backfills, or persist a migration-complete marker after verification. Add a regression that a completed tenant is detected as complete and does not keep re-running the full migration path on every open.

Validation run locally:
- `go test -c ./pkg/... ./cmd/...` passed
- targeted FUSE mode/chmod/lookup/mkdir/batch-stat tests passed
- targeted CLI chmod/fs/mkdir tests passed
- DB-backed server/migrate tests are still locally blocked by missing rootless Docker/testcontainers; GitHub `ci` and e2e smoke are green

qiffang (commented):

**Verdict: REVISE** on head `3e1e12fe`.

This is a foundational PR — the largest schema change in the project's history (5200+ lines, 44 files). The overall approach is sound: split `files` into `inodes`/`contents`/`semantic`, dual-write for legacy compatibility, `COALESCE(fn.inode_id, fn.file_id)` for backward-compat reads. But I see several correctness and safety issues that must be addressed before merge.

---

### Architecture Assessment

The split-table design is well-motivated:
- **`inodes`**: POSIX metadata (size, revision, mode, status, mtime) — maps to ext4 inode concept
- **`contents`**: storage blob metadata (storage_type, storage_ref, encryption, content_blob)
- **`semantic`**: search/enrichment (content_text, description, embeddings)
- **`file_nodes.inode_id`**: unified reference, replaces file_id for both files and directories
- **Dual-write**: atomic tx writes to both `files` and split tables during transition
- **`useLegacyFiles` auto-detection**: `information_schema.tables` check at `Open()` time
- **Rollback safety**: old binaries reading `files` table continue working

This is correct architecture. The issues below are implementation bugs, not design flaws.

---

### Blocking Issues

**B1: `chmod 000` cannot round-trip through remount** (previously flagged by @srstack)
- `fillAttr` uses `if !entry.HasMode { mode = 0644 }` which is correct for the FUSE layer
- But `UpdateMode` sets `HasMode = true` only in-memory. After remount/cache miss, `Lookup` calls `StatCtx` which returns `X-Dat9-Mode: 0` → `HasMode = true` in the client → `UpdateMode(ino, 0)` in FUSE → `fillAttr` sees `HasMode=true, Mode=0` → renders `0000`. This is technically correct but unusual.
- **However**, the real issue is: `mode INT NOT NULL DEFAULT 420` in the schema means existing files migrated from `files` get `mode=420` (0o644). But a `chmod 000` sets `mode=0`, which is a valid POSIX mode. The `HasMode` flag works correctly here because the server returns the mode header whenever the inode has a `mode` column value (including 0).
- **RESOLVED** — this is actually correct after re-analysis. `HasMode` is set whenever the server returns the header, and mode=0 is a valid value. Withdrawing B1.

**B2: Migration runs on every `Open()` for legacy tenants** (flagged by @srstack)
- `pool.go:413-416`: `if store.HasLegacyFiles() { migrateSplitTables() }` runs on every backend open
- `SplitTablesMigrator.Run()` uses `INSERT IGNORE`, so it's idempotent, but running 5 SQL queries + `information_schema` checks on every tenant open adds latency
- **Recommendation**: Add a migration completion marker (e.g., a `schema_migrations` row or check if all `file_nodes.inode_id` are populated) to skip on subsequent opens

**B3: `SplitTablesMigrator.isComplete` partial migration gap** (flagged by @qiffang)
- `Run()` executes 5 steps sequentially. If step 3 succeeds but step 4 crashes, a re-run will `INSERT IGNORE` steps 1-3 (0 rows affected) and then run steps 4-5.
- This is actually safe because each step uses `INSERT IGNORE` / `UPDATE ... WHERE ... IS NULL`. The idempotency is correct at the SQL level.
- **However**, the `Result` struct reports `InodesMigrated=0` on re-run even if the first run's inode migration actually succeeded. This is misleading for operational observability but not a correctness bug.

**B4: `EnsureParentDirs` inode_id not linked to file_nodes on duplicate**
- `store.go:548-552`: `INSERT IGNORE INTO inodes` then `INSERT INTO file_nodes ... ON DUPLICATE KEY UPDATE node_id = node_id`
- If the `file_nodes` row already exists (from a previous request), the `ON DUPLICATE KEY` clause updates `node_id = node_id` (no-op) but does NOT update `inode_id`. So a pre-existing directory node with `inode_id = NULL` will remain NULL even after the inode is created.
- The `createDirInodes` migration step backfills this, but during normal operation between migration runs, new directories created by `EnsureParentDirs` get proper `inode_id`, while old ones don't.
- **Fix**: Change to `ON DUPLICATE KEY UPDATE inode_id = COALESCE(inode_id, VALUES(inode_id))` to backfill on conflict.

**B5: `SetAttr` chmod uses `fileHandles.ForEach` twice** 
- `dat9fs.go:1419-1437`: First `ForEach` to check if any dirty handle exists, then second `ForEach` to set pending mode. Between the two iterations, a handle could be created or released. This is a TOCTOU race.
- **Fix**: Merge into a single `ForEach` call that both checks and sets in one pass.

---

### Important Observations (non-blocking)

**R1: `TestChmodDirectory` expects error for pre-migration directories**
- `store_test.go:2994-3002`: The test inserts a directory node without `inode_id`, then expects `Chmod` to fail. This is correct behavior for old directories (pre-migration), but after migration all directories should have inodes. Consider adding a test that verifies chmod works on a directory that HAS an inode_id.

**R2: Dual-write transaction scope in `UpdateFileContent`**
- `store.go:2068-2136`: For legacy tenants, the legacy `files` UPDATE and the split-table writes are in the same transaction. Good. But for new tenants (no legacy files), the transaction only wraps split-table writes. This is correct but worth noting for debugging.

**R3: `ListDir` query optimization**
- The old `ListDir` fetched full `files` columns including `content_blob`. The new query joins `inodes` + `semantic` (for `embedding_revision` only) and skips `contents` entirely. This is a significant improvement — directory listings no longer pull blob data.

**R4: `file_tags` backward compatibility**
- `file_tags` still uses `file_id` as part of the primary key while adding `inode_id`. The `Find` query uses `t.file_id = i.inode_id` which works because for files, `inode_id = file_id` by design. But the `backfillSharedInodeID` migration step sets `inode_id = file_id` on all existing rows, which is correct.

**R5: WebDAV `Mkdir` now passes perm bits**
- `webdav/fs.go:45`: `uint32(perm)&0o777` — clean. The WebDAV layer now forwards permission bits to the server.

**R6: Semantic FTS indexes use `ALTER TABLE ... ADD_COLUMNAR_REPLICA_ON_DEMAND`**
- @srstack flagged that the `semantic` table indexes in `tidb_app.go` optional statements use `ALTER TABLE` with `ADD_COLUMNAR_REPLICA_ON_DEMAND`, matching the existing `files` table pattern. This is correct for TiDB Cloud.

---

### Test Coverage Assessment

The test coverage is comprehensive:
- Datastore: chmod, confirm, delete pending, mode default/round-trip, no-legacy-table path
- Backend: chmod, mkdir with perm
- Client: chmod, mkdir
- Server: stat returns mode header
- FUSE: fillAttr mode, Lookup mode header parsing, inode UpdateMode
- Migration: basic run, idempotency, directory inodes, partial re-run, missing tables, generated column detection

**Missing tests** (non-blocking):
1. `SetAttr` chmod + deferred apply on Release
2. `EnsureParentDirs` with pre-existing directory node (the B4 case)
3. Concurrent chmod from two different FUSE handles

---

### Summary

This PR is very close to mergeable. The key fix needed is **B4** (EnsureParentDirs inode_id backfill on conflict) and **B5** (SetAttr TOCTOU). **B2** (migration on every open) is an operational concern that should be addressed before production rollout but could be a follow-up PR.

qiffang (changes requested):
Second independent review at head `3e1e12fe4db887637cbe062513682f41e9e9b547`.

Verdict: **REVISE**

The previous fixes for partial migration false-complete and directory stat mode look correct, but I still see two correctness/operational blockers.

1. `EnsureParentDirs` leaks confirmed directory inodes on idempotent calls.

`pkg/datastore/store.go:525-537` and `pkg/datastore/store.go:974-986` insert a fresh `inodes` row before attempting `file_nodes ... ON DUPLICATE KEY UPDATE node_id = node_id`. When the directory path already exists, the `file_nodes` insert is a no-op, but the newly generated inode remains in `inodes` and is not referenced by any `file_nodes` row. Repeated writes under existing parent directories will create orphan confirmed directory inodes. This is not just cosmetic metadata: once split tables become the source of truth, unreferenced confirmed inodes pollute inode lifecycle, quota/repair semantics, and future cleanup assumptions. Existing `TestEnsureParentDirs` only checks that the second call does not error; it does not assert inode cardinality or absence of unreferenced directory inodes.

2. Split migration still runs full-table migration work on every backend open for every current tenant schema.

`pkg/tenant/pool.go:419` calls `migrateSplitTables` whenever `store.HasLegacyFiles()` is true. The current schemas still create the legacy `files` table (`pkg/tenant/schema/tidb_auto.go:173`, `pkg/tenant/schema/tidb_app.go:41`, `pkg/tenant/db9/schema.go:36`), so this condition is true for normal tenants, not just old tenants. `SplitTablesMigrator.Run` then unconditionally executes all migration phases (`pkg/migrate/split_tables.go:91-131`), including `INSERT IGNORE ... SELECT FROM files` and shared-table backfills. Removing the unsafe `isComplete` fixed one correctness bug, but it leaves every cold backend open doing full migration scans. For large tenants this can turn tenant acquisition into repeated large table scans/updates and can make startup/request paths unreliable. This needs a safe completion gate/marker, or current schemas must stop creating `files` for new tenants and legacy migration must be explicitly one-shot/gated.

Validation performed:
- `git diff --check 260b09eadb9c5797801907d3f30bbd5f54762c49...HEAD`

GitHub CI/e2e are green, but these blockers are not covered by the current tests.

qiffang (changes requested):
## Dev Leader Re-Review — PR #418 (head `3e1e12f`)

**Verdict: REVISE**

Progress since last round: the `isComplete` false-positive that skipped partial migration is gone; directory stat now returns `X-Dat9-Mode`. Good fixes.

### Blocking

**B1 — `EnsureParentDirs` / `EnsureParentDirsTx` leak orphan inodes**

Both functions follow this pattern:
```go
nodeID := genID()
_, err := db.Exec(`INSERT IGNORE INTO inodes (...) VALUES (?, ...)`, nodeID, ...)
// ↑ Always succeeds (INSERT IGNORE = no-op on dup, but this is a fresh UUID so it ALWAYS inserts)

_, err = db.Exec(`INSERT INTO file_nodes (..., inode_id, ...) VALUES (?, ..., ?, ...)
    ON DUPLICATE KEY UPDATE node_id = node_id`, nodeID, ..., nodeID, ...)
// ↑ When directory already exists → no-op, inode_id is NOT updated to the new nodeID
```

Every call to `EnsureParentDirs` for an existing directory path creates a new orphan row in `inodes` that is never referenced by any `file_nodes` row. Over time, normal write traffic (e.g. `InsertFile` → `EnsureParentDirs`) accumulates orphan inodes with no foreign-key link.

**Fix**: Either (a) make the inode insert conditional — check `file_nodes` existence first and skip inode creation when the directory node already exists, or (b) use `ON DUPLICATE KEY UPDATE inode_id = VALUES(inode_id)` on the file_nodes insert so the inode is always linked (and add cleanup for the previous orphaned inode), or (c) use a single `INSERT ... ON DUPLICATE KEY` that returns the actual node_id and only create the inode when the node was truly new.

**B2 — Migrator runs on every backend open with no completion guard**

`Pool.createBackend` calls `migrateSplitTables` whenever `store.HasLegacyFiles()` is true. For legacy tenants the `files` table will always exist, so `HasLegacyFiles()` is permanently true. The migrator's `Run()` has no fast-path exit — it runs all 5 steps (`INSERT IGNORE ... SELECT FROM files` × 3 + `createDirInodes` + `backfillSharedInodeID`) on every backend open.

While `INSERT IGNORE` is idempotent, it still scans the full `files` table every time. For a tenant with 100K files, that's a 100K-row `SELECT FROM files WHERE status != 'DELETED'` three times per open.

**Fix**: Add a completion marker (e.g. a `migration_status` row in a meta table, or check `SELECT COUNT(*) FROM files WHERE file_id NOT IN (SELECT inode_id FROM inodes) AND status != 'DELETED'` = 0). Once migration is confirmed complete, skip subsequent runs.

### Recommendations

**R1 — SetAttr chmod double-ForEach TOCTOU**

`SetAttr` still does two separate `ForEach` passes:
1. First pass checks `hasDirtyHandle`
2. Second pass sets `PendingMode`

Between the two iterations, a concurrent `Open`/`Release` could change handle state. Recommend merging into a single pass — set PendingMode on dirty handles as you find them, and track whether you found any.

**R2 — `Store.Chmod` could skip inode lookup for directories**

`Chmod` does `GetNode` → resolve `inodeID` → `UPDATE inodes`. For directories created by `EnsureParentDirs`, the `inode_id` on the `file_nodes` row depends on B1 being fixed. Until then, `Chmod` on a directory that went through `EnsureParentDirs` may fail with `ErrNotFound` because `node.InodeID` could be empty and `node.FileID` is also empty for directories.

---

Overall the architecture is sound. The two blockers are localized fixes. Looking forward to the next iteration.

mornyx:
Hi @qiffang, thank you for the thorough reviews. I have gone through the four review threads (4272771211, 4272771751, 4272814620, 4272819550) and grouped the blocking issues into three fixes, all addressed in commit `38a0f10`.

### P1 — `EnsureParentDirs` / `EnsureParentDirsTx` leak orphan inodes on idempotent calls

**Root cause:** Both functions inserted a fresh `inodes` row before the `file_nodes` upsert, so when the directory already existed the `file_nodes` insert became a no-op while the new inode remained unreferenced.

**Fix:** Added an existence check (`SELECT node_id FROM file_nodes WHERE path = ? AND is_directory = 1`) before creating the inode. If the directory already exists we skip inode creation entirely.
- `pkg/datastore/store.go` — `EnsureParentDirs` and `EnsureParentDirsTx`
- `pkg/datastore/store_test.go` — `TestEnsureParentDirs` now asserts that a second call for the same path does not create orphan inodes (verifies inode count before/after and checks `LEFT JOIN` orphan count = 0)

### P2 — `SplitTablesMigrator` runs full migration on every backend open with no completion guard

**Root cause:** Removing the unsafe `isComplete()` optimization (per your prior review) left `Run()` unconditionally executing all 5 steps on every open.

**Fix:** Added `isMigrationComplete()` with four data-driven checks:
1. Any legacy `files` rows not yet in `inodes`
2. Any legacy `files` rows not yet in `contents`
3. Any legacy `files` rows not yet in `semantic`
4. Any directories in `file_nodes` with `inode_id IS NULL`

When all four return zero, `Run()` short-circuits with an empty `Result` and logs `migrate_split_tables_already_complete`.
- `pkg/migrate/split_tables.go`
- `pkg/migrate/split_tables_test.go` — `TestSplitTablesMigratorAlreadyComplete` verifies that a second `Run()` skips all work when migration is complete

### P3 — `SetAttr` chmod double-`ForEach` TOCTOU

**Root cause:** The dirty-handle check and the `PendingMode` assignment were in two separate `ForEach` passes, leaving a race window for concurrent `Open`/`Release` to change handle state.

**Fix:** Merged into a single `ForEach` pass that both detects dirty handles and sets `PendingMode` / `PreviousMode` atomically.
- `pkg/fuse/dat9fs.go`

### Verification

```
$ go test ./pkg/... -count=1
ok  	github.com/mem9-ai/dat9/pkg/datastore	31.185s
ok  	github.com/mem9-ai/dat9/pkg/fuse	12.480s
ok  	github.com/mem9-ai/dat9/pkg/migrate	25.691s
... (all packages pass)

$ make lint
0 issues.
```

Please take another look when you have a chance.

qiffang (changes requested):
Latest-head review at `38a0f1052bcb1d31a5f733236e56c3a0f8d8f5c0`.

Verdict: **REVISE**

The latest push addresses the sequential orphan case and restores a migration fast path, but I still see two blockers.

1. `EnsureParentDirs` still leaks orphan directory inodes under concurrent creation.

`pkg/datastore/store.go:526-548` and `pkg/datastore/store.go:986-1008` now check whether the directory exists before inserting a new inode. That fixes the sequential idempotent case, but not the race that matters on write paths. Two concurrent writers creating the same missing parent can both observe no `file_nodes` row, both insert different `inodes` rows, then only one `file_nodes` insert wins while the loser executes the no-op duplicate update. The losing inode is left as a confirmed unreferenced directory inode. The added test only calls `EnsureParentDirs` twice sequentially, so it does not cover this race.

2. The migration completion guard can still false-complete when shared inode backfill did not run.

`isMigrationComplete` checks files -> `inodes`, `contents`, `semantic`, and directory nodes with null `inode_id` (`pkg/migrate/split_tables.go:79-127`). It does not check the tables that step 5 backfills in `backfillSharedInodeID` (`file_nodes`, `file_tags`, `uploads`, `file_gc_tasks`; see `pkg/migrate/split_tables.go:331-348`). If a prior migration attempt completed steps 1-4 and then failed during step 5, the next backend open can return `complete=true` and skip the shared-table backfill permanently. That leaves file nodes/tags/uploads/gc tasks without `inode_id`, which is exactly the kind of partial-migration state the new guard is supposed to repair. The new `TestSplitTablesMigratorAlreadyComplete` only validates the happy-path second run; it does not simulate missing shared `inode_id` columns after split rows already exist.

Recommendation: the fast path still runs several full-table COUNT/NOT IN checks on every backend open for current schemas that keep `files`. That is much better than re-running all writes, but still worth tracking as a startup-path cost for large tenants; a durable marker or cheaper existence probes would be safer before very large tenants.

Validation performed:
- `git diff --check 260b09eadb9c5797801907d3f30bbd5f54762c49...HEAD`

GitHub CI/e2e are green, but these two failure modes are not covered by the current tests.

qiffang (commented):
## adversary-2 review — PR #418 at `38a0f10`

**Verdict: REVISE**

### Prior blocker status

**B1 (orphan inodes on EnsureParentDirs — sequential):** Fixed. The SELECT-before-INSERT guard prevents the idempotent repeat case. Test added.

**B2 (migrator runs every open):** Fixed. `isMigrationComplete()` with 4 data-driven checks provides a proper fast path.

**PR #411 B1 (missing mtime):** Fixed. `inodes` table now has `mtime DATETIME(3)`, seeded from `COALESCE(confirmed_at, created_at)` during migration.

**PR #411 B2 (migration step ordering):** Fixed. Migration uses a different architecture (dedicated `split_tables.go`) that doesn't have the step 3→4 column rename dependency.

---

### New blockers

#### B1: `EnsureParentDirs`/`EnsureParentDirsTx` concurrent race leaks orphan directory inodes

**Location:** `pkg/datastore/store.go:522-548` and `pkg/datastore/store.go:986-1008`

Both variants do:
1. `SELECT node_id FROM file_nodes WHERE path = ? AND is_directory = 1` → no row
2. `INSERT IGNORE INTO inodes` → succeeds, creates inode `A`
3. `INSERT INTO file_nodes ... ON DUPLICATE KEY UPDATE node_id = node_id` → wins or no-ops

Two concurrent writers creating the same missing parent can both pass step 1 (no row exists), both succeed at step 2 (creating inodes `A` and `B`), but only one wins the `file_nodes` insert at step 3. The loser's inode (`B`) is left as a confirmed, unreferenced directory inode permanently.

**Why it matters:** These orphan inodes are invisible to the file tree but consume storage quota (`status = 'CONFIRMED'`, `size_bytes = 0`). Over time in high-concurrency workloads (e.g. many agents writing to the same workspace), they accumulate. There is no cleanup mechanism.

**Test gap:** `TestEnsureParentDirs` only calls the function twice sequentially, not concurrently.

**Note:** `EnsureParentDirsTx` has the same race. When the caller's `execer` is a transaction, the gap between the SELECT and the INSERT is narrower but still present — another connection's non-transactional `EnsureParentDirs` can interleave.

---

#### B2: `isMigrationComplete` false-completes when step 5 (shared backfill) has not run

**Location:** `pkg/migrate/split_tables.go:79-127` vs `pkg/migrate/split_tables.go:331-348`

`isMigrationComplete` checks 4 conditions (files→inodes, files→contents, files→semantic, directory inode_id). It does **not** check the tables backfilled in step 5: `file_nodes.inode_id` (for non-directory nodes), `file_tags.inode_id`, `uploads.inode_id`, `file_gc_tasks.inode_id`.

Failure scenario: steps 1–4 complete, step 5 fails (e.g. DB timeout on large `file_tags` UPDATE). Next backend open → `isMigrationComplete()` returns `true` → step 5 never reruns → all non-directory `file_nodes`, `file_tags`, `uploads`, and `file_gc_tasks` rows permanently lack `inode_id`.

This is exploitable by any transient DB error during step 5. The `COALESCE(fn.inode_id, fn.file_id)` pattern in queries papers over the missing `inode_id` at read time, but any code path that relies on `inode_id IS NOT NULL` will silently miss rows.

**Test gap:** `TestSplitTablesMigratorAlreadyComplete` only validates the happy-path second run.

---

#### R1: `EnsureParentDirs` hardcoded mode `493`

`pkg/datastore/store.go:537` and `pkg/datastore/store.go:997`: the inode insert uses `mode = 493` (decimal for `0o755`). This is correct but the decimal literal is opaque. The `MkdirCtx` path uses `perm` from the caller. Using `0o755` as the literal (Go supports octal literals) would be clearer.

#### R2: `handleChmod` does not validate mode range

`pkg/server/server.go:1480-1510`: The `handleChmod` endpoint accepts `req.Mode uint32` from the JSON body and passes it directly to `b.ChmodCtx`. The `Chmod` store method masks to `mode & 0o777`, so setuid/setgid bits are silently stripped. This is fine defensively, but the API should either document that only permission bits are accepted, or reject requests with bits above `0o777` with a 400 error for clarity.

#### R3: `Release` chmod defer calls `cf()` after `ChmodCtx` returns

`pkg/fuse/dat9fs.go:3773-3775`: The `ctx, cf := context.WithTimeout(...)` and `cf()` are ordered correctly for the success path, but the context cancel happens _after_ the `ChmodCtx` call returns. If the function returns via the `flushStatus != OK` branch, `cf()` is never called because the context was never created. This is fine — just noting the asymmetry.

#### R4: `scanFileForGCTx` new-tenant path does not read `checksum_sha256` or `source_id`

`pkg/datastore/store.go:1822-1838`: For new tenants (no legacy files), `scanFileForGCTx` builds a `File` struct from inodes+contents+semantic but doesn't populate `ChecksumSHA256`, `SourceID`, `CreatedAt`, `ConfirmedAt`, or `ExpiresAt`. If any GC consumer relies on these fields beyond storage ref/type/size, it will get zero values. Verify that the GC task consumer only needs the fields currently populated.

---

### Validation

```
git diff --check 179ac00..38a0f10  # clean
go build ./...                     # passes
go test ./pkg/fuse -count=1        # passes
```

mornyx:
Hi @qiffang, both blockers from the latest round are fixed in commit `fd59219`.

### B1 — `EnsureParentDirs` concurrent race leaks orphan directory inodes

**Root cause:** The prior sequential SELECT-before-INSERT guard fixed the idempotent-repeat case but not the interleaving race. Two concurrent writers can both observe no `file_nodes` row, both insert distinct `inodes` rows, and then only one `file_nodes` insert wins while the other becomes a no-op — leaving the loser's inode unreferenced.

**Fix:**
- Both `EnsureParentDirs` and `EnsureParentDirsTx` now use `SELECT ... FOR UPDATE` to serialize concurrent creators on the same missing parent path.
- `EnsureParentDirs` wraps the entire loop in a dedicated transaction so the row-level locking works even for non-transactional callers.
- A bounded deadlock retry (up to 3 attempts) is added to `EnsureParentDirs` because high-concurrency contention for the same gap lock can trigger MySQL Error 1213. `EnsureParentDirsTx` leaves retry responsibility to the caller's outer transaction.
- The shared logic is extracted into `ensureParentDirsWithExecer`.
- Hardcoded decimal `493` replaced with `0o755`.

**Test:** `TestEnsureParentDirsConcurrent` spawns 10 goroutines racing to create `/a/b/c/file.txt`, asserting exactly 3 directory inodes and 0 orphan inodes afterward.

### B2 — `isMigrationComplete` false-completes when step 5 has not run

**Root cause:** `isMigrationComplete` only checked files→inodes/contents/semantic and directory `inode_id` linkage. It did not verify the shared-table backfill from step 5 (`backfillSharedInodeID`). If a prior run completed steps 1–4 and then crashed during step 5, the next open would incorrectly skip step 5.

**Fix:** Added Check 5 to `isMigrationComplete`: a `UNION ALL` query over `file_nodes` (non-directory), `file_tags`, `uploads`, and `file_gc_tasks` looking for rows where `inode_id IS NULL AND file_id IS NOT NULL`.

**Test:** `TestSplitTablesMigratorPartialStep5` manually runs steps 1–4, inserts a `file_nodes` row with empty `inode_id`, asserts `isMigrationComplete` returns `false`, then asserts a full `Run()` repairs step 5 with `SharedColsUpdated > 0`.

### Verification
```
$ go test ./pkg/... -count=1
ok  	github.com/mem9-ai/dat9/pkg/datastore	32.320s
ok  	github.com/mem9-ai/dat9/pkg/migrate	27.661s
... (all packages pass)

$ make lint
0 issues.
```

Please take another look when you have a chance.

qiffang (commented):
## adversary-2 re-review — PR #418 at `fd59219`

**Verdict: REVISE (one blocker)**

### Prior blocker B1 (concurrent orphan inodes): Addressed — with a new bug

The fix uses `SELECT ... FOR UPDATE` inside a transaction to serialize concurrent `EnsureParentDirs` callers on the same parent path. This correctly prevents the concurrent orphan inode race via InnoDB/TiDB gap locks. Deadlock retry (3 attempts) is a reasonable safeguard.

**However, there is a variable shadowing bug in the retry loop:**

`pkg/datastore/store.go:515` — `tx, err := s.db.BeginTx(ctx, nil)` uses `:=`, creating a **new local** `err` that shadows the named return `(err error)`. Same on lines 519 and 526. After all 3 deadlock retries exhaust, `return err` at line 535 returns the **named return** `err`, which was **never assigned** inside the loop — it is still the zero value `nil`.

**Result:** If `EnsureParentDirs` deadlocks 3 times, it returns `nil` instead of the deadlock error. The caller (e.g. `InsertFile`, `Write`) proceeds as if parent directories were created successfully, but they were not.

**Fix:** Either use `=` instead of `:=` for the `err` assignments inside the loop (and declare `tx` before the loop), or capture the last error in a separate variable before `continue`.

### Prior blocker B2 (migration step 5 guard): Fixed ✅

`isMigrationComplete` now includes check 5: a `UNION ALL` across `file_nodes`, `file_tags`, `uploads`, and `file_gc_tasks` for rows where `inode_id IS NULL AND file_id IS NOT NULL`. This correctly detects partial step 5 failure. Test `TestSplitTablesMigratorPartialStep5` validates the scenario.

### Other observations

- R1 from prior review (hardcoded `493`): Fixed — now uses `0o755` literal (`pkg/datastore/store.go:990`).
- Concurrent test `TestEnsureParentDirsConcurrent`: Good coverage — 10 goroutines, verifies exactly 3 directory inodes and 0 orphans.

### Validation

```
go build ./...    # passes
go vet ./pkg/datastore/... ./pkg/migrate/...  # clean
git diff --check  # clean
```

mornyx:
Hi @qiffang, thanks for catching the shadowing bug — fixed in `3f5d0fb`.

**Problem:** The retry loop used `:=` for `err` (lines 515, 519, 526), creating new locals that shadowed the named return `(err error)`. After all 3 deadlock retries exhausted, `return err` at line 535 returned the named return, which was never assigned — so it was `nil`. The caller would then proceed as if parent dirs were created successfully.

**Fix:** Removed the named return. `err` is now a single local variable declared before the loop, and all assignments inside the loop use `=` (no `:=`). Loop-exhaustion returns the actual last error.

```go
func (s *Store) EnsureParentDirs(ctx context.Context, path string, genID func() string) error {
    var err error
    // ...
    for attempt := 0; attempt < maxAttempts; attempt++ {
        var tx *sql.Tx
        tx, err = s.db.BeginTx(ctx, nil)
        // ...
        err = ensureParentDirsWithExecer(ctx, tx, path, genID)
        // ...
        err = tx.Commit()
        // ...
    }
    return err
}
```

All tests and lint pass.

qiffang (changes requested):
Latest-head review at `fd592190a3dcf64d35c1d4fed983c6f94c6a9bfd`.

Verdict: **REVISE**

The migration step-5 guard is now fixed: `isMigrationComplete` checks the shared-table `inode_id` backfill, and the new partial-step-5 test covers the prior false-complete case.

One blocker remains in `EnsureParentDirs`.

1. The concurrent orphan-inode fix still depends on gap locking that TiDB does not provide for missing rows.

`pkg/datastore/store.go:979-1001` uses `SELECT ... FOR UPDATE` to check a missing `file_nodes.path`, then inserts a new `inodes` row before `INSERT INTO file_nodes ... ON DUPLICATE KEY UPDATE node_id = node_id`. This is still unsafe for Drive9's TiDB-backed metadata. TiDB documents that `SELECT FOR UPDATE` locks result-set rows, and its pessimistic transactions do not do gap locking for non-existing rows: <https://docs.pingcap.com/tidb/stable/pessimistic-transaction/#select-for-update>. Two concurrent writers can therefore both see no row, both insert distinct confirmed directory inodes, and only one `file_nodes` insert wins. The loser commits with an orphan inode because the duplicate path path is a no-op update. The new concurrency test is useful, but it runs against the local MySQL/testcontainers behavior and does not prove the TiDB no-gap-lock case.

2. The retry loop can return success after exhausting deadlock retries.

`EnsureParentDirs` has a named return `err`, but the loop uses short declarations inside the loop (`tx, err := ...`, `if err := ...`). If all attempts hit a retryable deadlock, execution reaches the final `return err` at `pkg/datastore/store.go:535`, returning the outer named `err`, which was never assigned and is therefore nil. The caller can proceed as if the parent directories were created.

Validation performed:
- `git diff --check 43be08398c2dddafd6d2ea8f5ee1f28b3e04e22e...HEAD`
- attempted `/usr/local/go/bin/go test ./pkg/datastore -run TestEnsureParentDirsConcurrent -count=1`, but local runtime execution is blocked by existing `rootless Docker not found` / testcontainers limitation.

GitHub CI/e2e are green, but the remaining blocker is a production TiDB concurrency semantics issue, not covered by the current test environment.

qiffang (commented):
## adversary-2 re-review — PR #418 at `3f5d0fb`

**Verdict: GREEN**

### Blocker regression

**Variable shadowing (from last review):** Fixed. `err` is now declared as `var err error` before the loop, and all assignments use `=` instead of `:=`. After 3 deadlock retries, `return err` correctly returns the last error. The named return is removed; `observeStoreOp` still captures via `&err`. Correct.

**Migration step 5 guard:** Still fixed from previous round. No regression.

### TiDB gap lock concern (raised by @adversary-1)

`SELECT ... FOR UPDATE` on a non-existent `file_nodes` row relies on locking semantics. In TiDB **pessimistic mode** (default since v4.0), `FOR UPDATE` on a non-existent unique index key (`idx_path ON file_nodes(path)`) acquires a pessimistic lock on that specific key point. A concurrent transaction issuing the same `SELECT ... FOR UPDATE` will block until the first commits or rolls back. This prevents the concurrent orphan inode race correctly.

In TiDB **optimistic mode**, `FOR UPDATE` does not acquire locks until commit, and the race would persist. However, optimistic mode is not the default and would cause broader correctness issues across the codebase (not specific to this PR). This is acceptable as a deployment assumption.

**Recommendation (non-blocking):** Document the pessimistic mode requirement in a comment or in the schema/deployment docs, since this is a correctness dependency rather than a performance preference.

### Checklist

- [x] All inputs handled (including invalid)
- [x] Edge cases covered (concurrent creation, deadlock retry, migration partial failure)
- [x] Failure behavior defined (deadlock retry 3x, error propagation correct)
- [x] API explicit (no hidden contracts)
- [x] Race / state issues addressed (FOR UPDATE serialization)
- [x] Regression tests included (concurrent test, partial step 5 test)

### Validation

```
go build ./...                                # passes
go vet ./pkg/datastore/... ./pkg/migrate/...  # clean
git diff --check                              # clean
```

qiffang (approved):
Latest-head re-review at `3f5d0fb11119a940f733d98f8cff7efa4bed74e2`.

Verdict: **GREEN**

The remaining blocker from my previous review is resolved:

- The deadlock retry shadowing bug in `EnsureParentDirs` is fixed. `err` is now declared outside the retry loop and assignments use `=`, so an exhausted retry path returns the last real error instead of a nil named return value.
- The migration step-5 guard remains fixed: `isMigrationComplete` now includes shared-table `inode_id` backfill checks, and the partial-step-5 regression test covers the prior false-complete path.

I re-checked the TiDB concurrency concern. With the unique `file_nodes.path` index and TiDB pessimistic transactions, the `SELECT ... FOR UPDATE` point lookup is the intended lock shape for concurrent creation of the same missing path. The remaining caveat is operational rather than PR-blocking: Drive9 should document TiDB pessimistic transaction mode as a correctness assumption for this path.

Validation performed:
- `git diff --check 43be08398c2dddafd6d2ea8f5ee1f28b3e04e22e...HEAD`
- `/usr/local/go/bin/go vet ./pkg/datastore ./pkg/fuse ./pkg/migrate`

`pkg/datastore` runtime tests still cannot run locally because this machine lacks rootless Docker/testcontainers, but GitHub CI/e2e are green.

qiffang (approved):
## Dev Leader Re-Review — PR #418 (head `3f5d0fb`)

**Verdict: GREEN**

All prior blockers resolved:

1. **EnsureParentDirs orphan inodes** — Fixed. `SELECT ... FOR UPDATE` inside a transaction serializes concurrent creators. Inode is only inserted after confirming the directory doesn't exist. Variable shadowing fixed (`var err error` before loop, `=` assignments throughout). Deadlock retry correctly returns last error on exhaustion.

2. **Migrator completion guard** — Fixed. `isMigrationComplete` now checks all 5 steps including shared-table `inode_id` backfill (check 5: `file_nodes`, `file_tags`, `uploads`, `file_gc_tasks` via `UNION ALL`). Fast-path exit skips heavy work when migration is already done.

3. **SetAttr TOCTOU** — Fixed (single ForEach pass).

4. **Hardcoded mode 493** — Fixed to `0o755`.

Agree with @adversary-2's recommendation to document TiDB pessimistic mode as a correctness prerequisite, but not a blocker for this PR.

Verified: `EnsureParentDirs` retry loop, `isMigrationComplete` 5-check coverage, `ensureParentDirsWithExecer` shared implementation, `Chmod` store path.

qiffang (approved):

qiffang (approved):
## Dev Leader Re-Review — PR #418 (head `47b690d`, post-merge)

**Verdict: GREEN**

Reviewed the merge commit (`47b690d`) and the new `6c18848` (eliminate TiDB gap-lock dependency). Two key changes since last GREEN:

### 1. `deterministicNodeID(path) = SHA256(path)` — Better than SELECT FOR UPDATE

Replaces the gap-lock approach with a deterministic inode_id derived from the directory path. Concurrent creators for the same missing parent now generate the **same** inode_id, so `INSERT IGNORE INTO inodes` is truly idempotent — no orphan inodes possible regardless of TiDB vs MySQL locking semantics. This is a cleaner and more portable solution.

Also handles pre-migration directories (inode_id IS NULL) with an explicit backfill path.

### 2. Merge conflict resolution — Clean

The merge correctly integrates:
- PR #417: `OpenHandleIndex`, `Prefetcher` atomic path, `RWMutex` handle, profile TTL defaults
- PR #420: `CreateCtx` transactional (`InTx`), typed delete (`RemoveFileCtx`/`RemoveDirCtx`), `pathPrefixPredicate`, `CommitQueue.queuedByPath`, `dirHasChildrenTx`
- Windows CLI platform files

No regressions found in schema refactor paths:
- `isMigrationComplete` 5-check intact
- `insertSplitTablesTx` dual-write intact
- `handleChmod`, `handleCreate`, `handleDelete` server routes intact
- `fillAttr` mode propagation intact

qiffang (commented):
## adversary-2 re-review — PR #418 at `47b690d` (post conflict resolution)

**Verdict: GREEN**

### Changes since last GREEN (`3f5d0fb`)

1. **`6c18848` — Eliminate TiDB gap-lock dependency in EnsureParentDirs**

Replaces `SELECT ... FOR UPDATE` with **deterministic inode IDs** (`SHA-256(path)`). This is a strictly better approach:
- Concurrent creators for the same missing parent generate the **same** `inode_id` → `INSERT IGNORE` swallows the duplicate → zero orphan inodes possible
- No gap lock dependency → works correctly on both MySQL and TiDB regardless of transaction mode
- Pre-migration directory backfill case handled: when `file_nodes` row exists with `inode_id IS NULL`, the code backfills via `UPDATE file_nodes SET inode_id = ?`
- Transaction + deadlock retry (3 attempts) still present for atomicity across multi-ancestor inserts

Previous variable shadowing fix (`var err error` before loop, `=` assignments) is preserved. Correct.

2. **`47b690d` — Merge conflict resolution**

Integrated from main:
- PR #417 `OpenHandleIndex` — `hasOpenHandle` now uses O(1) index instead of `Snapshot()` scan
- PR #420 typed delete (`DeleteFileCtx`/`DeleteDirCtx`), transactional `CreateCtx`, `pathPrefixPredicate`
- `InsertFileTx` in `file_tx.go` — proper Tx variant with dual-write to split tables
- Windows CLI, doctor platform split

No conflicts in the PR #418-specific code paths (migration, chmod, mode propagation, inode types). Schema refactor paths are consistent with the merged main code.

### Validation

```
go build ./...                                                    # passes
go vet ./pkg/datastore/... ./pkg/migrate/... ./pkg/backend/... ./pkg/fuse/...  # clean
git diff --check                                                   # clean
```

qiffang (approved):
Post-conflict-resolution re-review at `47b690db742648f74ae7e9e7e2b63c71d7abd492`.

Verdict: **GREEN**

I compared the conflict-resolution head against the previously approved `3f5d0fb` and re-checked the high-risk paths: parent-dir inode creation, split-table migration completion, typed delete / transactional create from PR #420, and FUSE/open-handle conflict integrations.

The conflict resolution preserves the prior fixes:

- `EnsureParentDirs` now uses deterministic directory inode IDs (`sha256(path)`), so concurrent creators for the same parent path collide on the same `inode_id`; `INSERT IGNORE` makes the operation idempotent without depending on TiDB gap-lock behavior.
- The retry-loop shadowing bug remains fixed.
- `isMigrationComplete` still checks the shared-table `inode_id` backfill, so the previous step-5 false-complete path stays covered.
- The merge with PR #420 keeps typed delete and transactional `CreateCtx`; I did not see a regression in the schema-refactor paths.

Validation performed:
- `git diff --check eafdaf2445175fb630894d12bd1fd9f54c6e15a1...HEAD`
- `/usr/local/go/bin/go vet ./pkg/datastore ./pkg/fuse ./pkg/migrate ./pkg/backend ./pkg/server`

`pkg/datastore` runtime tests still cannot run locally because this machine lacks rootless Docker/testcontainers, but GitHub CI/e2e are green.

qiffang (approved):
Post-conflict re-review at `47b690d`: GREEN.

I compared the conflict-resolution head against the previous approved head and re-checked the schema/refactor, delete-path merge, inode, migration, and mode propagation paths.

What I verified:
- `EnsureParentDirs` now uses deterministic directory inode IDs derived from the path. That removes the previous orphan-inode race without relying on TiDB gap-lock behavior.
- The split migrator completion guard still checks the shared-table `inode_id` backfill step, so the step-5 false-complete regression remains fixed.
- Directory stat mode propagation remains intact (`X-Dat9-Mode` is emitted when `nf.HasMode`).
- PR #420 merge integration keeps typed delete / transactional create semantics in the affected paths.
- The PR remains based on current `origin/main`; GitHub `ci` and `drive9-server-local e2e smoke` are green.

Local validation:
- `go test -c ./pkg/... ./cmd/...` passed
- `go test ./pkg/fuse -count=1` passed
- targeted CLI mount/chmod/fs/mkdir/WebDAV/doctor tests passed
- `go vet ./pkg/datastore ./pkg/fuse ./pkg/migrate ./pkg/backend ./pkg/server` passed

DB-backed runtime tests are still locally blocked by the existing rootless Docker/testcontainers limitation, so GitHub CI remains the runtime gate.

qiffang (approved):

PR #420, opened 2026-05-12
Optimize delete performance

srstack:
## Summary
- batch recursive directory delete cleanup in datastore and avoid sibling-prefix matches
- add typed DELETE hints for FUSE unlink/rmdir to skip server-side resolve where the caller already knows the node type
- add path/prefix indexes for FUSE pending cleanup paths to reduce hot-path scans

## Validation
- go test ./pkg/client ./pkg/server ./pkg/backend ./pkg/datastore ./pkg/fuse
- make lint
- git diff --check

## Summary by CodeRabbit

* **New Features**
  * Client/server create action with revision response; explicit file-vs-dir create and delete helpers; write-back listing by prefix.

* **Bug Fixes**
  * Prevent orphaned storage objects/DB rows on duplicate create; enforce typed delete semantics; more direct file deletion in bulk removes.

* **Chores**
  * Raise kernel cache TTLs and unify read-cache size limits; add path-indexing to commit queue; improve flush/create fallback handling.

* **Tests**
  * Expanded coverage for create/delete hints, cache behavior, write-back listing, orphan cleanup, and path-indexing.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mem9-ai/drive9/pull/420)

### DIFF
--- a/pkg/backend/dat9.go
+++ b/pkg/backend/dat9.go
@@ -215,29 +215,33 @@ func (b *Dat9Backend) CreateCtx(ctx context.Context, path string) (err error) {
 		}
 	}
 
-	err = b.store.InsertFile(ctx, &datastore.File{
-		FileID: fileID, StorageType: storageType, StorageRef: storageRef,
-		StorageEncryptionMode:  storageEncryptionMode,
-		StorageEncryptionKeyID: storageEncryptionKeyID,
-		ContentBlob:            contentBlob,
-		SizeBytes:              0, Revision: 1, Status: datastore.StatusConfirmed,
-		CreatedAt: now, ConfirmedAt: &now,
+	err = b.store.InTx(ctx, func(tx *sql.Tx) error {
+		if err := b.store.InsertFileTx(tx, &datastore.File{
+			FileID: fileID, StorageType: storageType, StorageRef: storageRef,
+			StorageEncryptionMode:  storageEncryptionMode,
+			StorageEncryptionKeyID: storageEncryptionKeyID,
+			ContentBlob:            contentBlob,
+			SizeBytes:              0, Revision: 1, Status: datastore.StatusConfirmed,
+			CreatedAt: now, ConfirmedAt: &now,
+		}); err != nil {
+			return err
+		}
+		if err := b.store.EnsureParentDirsTx(tx, path, b.genID); err != nil {
+			return err
+		}
+		return b.store.InsertNodeTx(tx, &datastore.FileNode{
+			NodeID: b.genID(), Path: path, ParentPath: pathutil.ParentPath(path),
+			Name: pathutil.BaseName(path), FileID: fileID, CreatedAt: now,
+		})
 	})
 	if err != nil {
+		if storageType == datastore.StorageS3 {
+			b.deleteBlobCtx(ctx, storageRef)
+		}
 		return err
 	}
-	err = b.store.EnsureParentDirs(ctx, path, b.genID)
-	if err != nil {
-		return err
-	}
-	err = b.store.InsertNode(ctx, &datastore.FileNode{
-		NodeID: b.genID(), Path: path, ParentPath: pathutil.ParentPath(path),
-		Name: pathutil.BaseName(path), FileID: fileID, CreatedAt: now,
-	})
-	if err == nil {
-		b.syncCentralFileCreate(ctx, fileID, 0, "")
-	}
-	return err
+	b.syncCentralFileCreate(ctx, fileID, 0, "")
+	return nil
 }
 
 func (b *Dat9Backend) Mkdir(path string, perm uint32) error {
--- a/pkg/backend/dat9.go
+++ b/pkg/backend/dat9.go
@@ -78,8 +78,8 @@ type Dat9Backend struct {
 	inlineThreshold int64
 	// textExtractMaxBytes caps synchronous text extraction input size.
 	textExtractMaxBytes int64
-	mu                    sync.Mutex
-	entropy               io.Reader
+	mu                  sync.Mutex
+	entropy             io.Reader
 
 	// Central quota enforcement (Rev 4 migration).
 	tenantID    string
@@ -105,7 +105,7 @@ type Dat9Backend struct {
 	audioExtractMaxSize      int64
 	maxAudioExtractTextBytes int
 
-	fileGCWorker *FileGCWorker
+	fileGCWorker     *FileGCWorker
 	runtimeMetricsID uint64
 
 	// Monthly LLM cost budget (P1).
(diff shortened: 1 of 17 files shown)


qiffang (changes requested):
Verdict: REVISE.

I found one correctness blocker in the new empty-file create path.

1. `POST ?create` can leak an orphan confirmed file row/object on conflict or any failure after `InsertFile`.

`handleCreate` now calls `b.CreateCtx` (`pkg/server/server.go:1474-1493`), and FUSE now uses this endpoint for every new empty file (`pkg/fuse/dat9fs.go:4229-4237`). But `CreateCtx` first creates the file entity/S3 object, then creates parent dirs, then inserts the path node (`pkg/backend/dat9.go:218-236`). If `EnsureParentDirs` or `InsertNode` fails, especially the expected duplicate-path conflict that the server maps to `409`, the newly inserted confirmed `files` row is left with no `file_nodes` reference and no GC task. In S3-backed mode the empty object can be leaked too.

Concrete failure: two clients/FUSE handles race to create the same empty path. The loser receives `409`, but its generated `file_id` remains as a confirmed orphan. The current `TestCreateFileActionCreatesEmptyFileAndConflicts` only checks the HTTP status; it does not assert that the failed create left no orphan `files` row/object.

Required before merge: make this create operation atomic with the path node, or explicitly clean up/enqueue GC for the created file/object on every post-`InsertFile` failure. Add a regression for duplicate/concurrent create proving the failed create returns `409` and leaves exactly one referenced file and no orphan confirmed file/object.

Validation I ran:
- `/usr/local/go/bin/go test -c ./pkg/backend ./pkg/datastore ./pkg/fuse ./pkg/client ./pkg/server ./cmd/drive9 ./cmd/drive9/cli` passed.
- `/usr/local/go/bin/go test ./pkg/client ./pkg/fuse -count=1` ran `pkg/fuse` successfully, but `pkg/client` failed in `TestMain` because this local machine has no rootless Docker/testcontainers environment (`rootless Docker not found`).

srstack:
Addressed the latest review feedback in 72ecdd4:\n\n- Made CreateCtx metadata writes transactional and delete the prewritten S3 object if the metadata transaction fails, so duplicate empty create returns 409 without leaving orphan files or objects. Added backend/server regressions for duplicate create.\n- CreateFileCtx now only tolerates an empty response body; malformed JSON is returned as a decode error.\n- Replaced byte-range path prefix predicates with binary LIKE predicates so correctness does not depend on the database default collation ordering.\n\nValidation: go test ./pkg/client ./pkg/server ./pkg/backend ./pkg/datastore ./pkg/fuse; make lint; git diff --check.

qiffang (approved):
Re-reviewed latest head `72ecdd4a6c08a953c2a0c670e39152a4d1e22486`.

Verdict: GREEN.

The previous blocker is addressed: empty create now inserts the file row and path node inside one datastore transaction, and the new tests cover duplicate create leaving no orphan file rows / extra local S3 object. The client JSON decode issue is also tightened with a malformed-response regression.

Validation:
- GitHub `ci`, `drive9-server-local e2e smoke`, and CodeRabbit are green.
- `/usr/local/go/bin/go test -c ./pkg/backend ./pkg/datastore ./pkg/fuse ./pkg/client ./pkg/server ./cmd/drive9 ./cmd/drive9/cli` passed.
- `/usr/local/go/bin/go test ./pkg/fuse -count=1` passed.
- `/usr/local/go/bin/go test -race ./pkg/fuse -run 'TestCommitQueueQueuedPathIndexTracksCancelAndRemove|TestWriteBackCacheListByPrefix|TestReleaseNewEmptyFileUsesCreateAction|TestUnlinkAndRmdirUseDeleteKindHints|TestReadCacheCachesMediumSmallFileAfterFirstRead' -count=1` passed.
- `pkg/client` runtime tests are still blocked locally by missing rootless Docker/testcontainers, same local environment limitation as before.

qiffang (commented):

**Verdict: GREEN** on head `72ecdd4`.

This is a substantial optimization PR covering delete performance, file creation, and several FUSE caching improvements. Well-structured with comprehensive tests across all layers.

---

### What this PR does

**1. Typed DELETE hints** (`client.go`, `server.go`)
- New `DeleteFileCtx`/`DeleteDirCtx` with `?kind=file`/`?kind=dir` query params
- Server-side `handleDelete` dispatches to type-specific `RemoveFileCtx`/`RemoveDirCtx`, skipping node resolution when type is known
- Falls back to generic `RemoveCtx` when kind is omitted → backward compatible

**2. Batch recursive delete** (`datastore/store.go`)
- `pathPrefixPredicate` uses `BINARY column LIKE BINARY ? ESCAPE '\\\\'` — correctly fixes the sibling prefix match bug (e.g., `/data/` no longer matches `/data-other/`)
- `orphanedFilesByPathPrefixTx` uses `NOT EXISTS` subquery to find orphaned files efficiently
- `markFilesDeletedTx`/`deleteFileTagsTx`/`enqueueFileGCTasksTx` batch helpers with `deleteBatchSize = 500`
- `fileIDExistsTx` and `dirHasChildrenTx` use `LIMIT 1 FOR UPDATE` for efficient existence checks

**3. CreateFileCtx** (`client.go`, `server.go`, `dat9fs.go`)
- New `POST ?create` action for empty file creation, replacing `PUT` with empty body
- `handleCreate` inserts file row + path node atomically in a transaction (fixed from initial version per @qiffang's review)
- `isCreateActionUnsupportedErr` for backward compat with older servers — falls back to PUT

**4. CommitQueue path index** (`commit_queue.go`)
- `queuedByPath map[string]map[*CommitEntry]struct{}` for O(1) path lookup
- `addQueuedLocked`/`removeQueuedLocked`/`hasQueuedPathLocked` maintain the index
- `rebuildQueuedIndexLocked` for recovery scenarios
- Eliminates linear scan in `hasQueuedCommit`

**5. ReadCache and kernel cache tuning** (`read.go`, `dat9fs.go`)
- `defaultReadCacheMaxFileSize` increased from 50KB to 256KB — caches medium-small files that were previously falling through
- `defaultPositiveKernelCacheTTL` increased from 10s to 60s — reduces kernel→FUSE round-trips for stable files

**6. Writeback `ListByPrefix`** (`writeback.go`)
- New method for prefix-scoped writeback queries, used by typed delete

---

### Key design decisions reviewed

- **`pathPrefixPredicate` BINARY LIKE**: The `BINARY` qualifier ensures case-sensitive matching on TiDB (which defaults to utf8mb4_general_ci). The trailing `/` in the prefix pattern ensures `/data/` doesn't match `/data-other/`. Good fix for a subtle correctness bug.

- **`CreateCtx` wrapped in `InTx`**: The atomicity fix (per @qiffang's first review) ensures that file row + path node are inserted in a single transaction, preventing orphaned rows on conflict.

- **Kernel cache 60s TTL**: Aggressive but reasonable for agent workloads where files are mostly written by the local agent. SSE watcher handles foreign-actor invalidation, so stale reads only happen in the ~60s window if SSE delivery is delayed.

---

1. **R1: `deleteBatchSize` tuning** — 500 is a reasonable default, but for very deep directory trees (e.g., `node_modules` with 50K+ files), this could result in many batched queries. Consider adding a perf counter to track batch count per recursive delete, so we can tune later.

2. **R2: `isCreateActionUnsupportedErr` maintenance** — This backward-compat shim checks for both "unknown action" and 404 errors. Once all servers are upgraded, this should be removed to avoid masking real errors. Consider adding a TODO with a target version.

3. **R3: ReadCache 256KB threshold interaction with prefetch** — Files in the 50KB-256KB range are now cached by ReadCache *and* may trigger prefetch. Verify that the prefetch path checks ReadCache first to avoid redundant fetches. (This may already be handled by the existing `ReadCache.Get` check in the read path.)

PR #422, opened 2026-05-13
drive9 add journal logic

qiffang:
## Summary by CodeRabbit

* **New Features**
  * Added `journal` CLI command with subcommands: `new` (create journals), `append` (add entries), `cat` (read entries), `find` (search), and `verify` (validate integrity).
  * Added HTTP API endpoints for journal creation, entry appending, reading, searching with filters, and verification.
  * Implemented idempotent append operations and hash-chain verification for journal integrity.

* **Documentation**
  * Added comprehensive agent journal design specification.

* **Tests**
  * Added end-to-end smoke tests for journal operations.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mem9-ai/drive9/pull/422)


qiffang (commented):
## PR #422 Review — Journal Logic

**Reviewer**: @dat9-dev1 (Dev Leader)
**Head**: `710ead5`
**Verdict**: 🟡 **REVISE** (2 blocking, 4 non-blocking)

---

### Architecture Overview

Well-designed append-only journal primitive with clean layering:
- `pkg/journal` — shared types, canonicalization, hash chain
- `pkg/datastore` — TiDB persistence, idempotent create/append, verification
- `pkg/server` — HTTP handlers, HMAC cursor, permission model
- `pkg/client` — SDK with NDJSON streaming
- `cmd/drive9/cli` — Unix-friendly CLI (`new|append|cat|find|verify`)
- `docs/design/agent-journal-design.md` — 2272-line design spec

The hash chain (genesis → entry_hash chaining via prev_hash) with SHA-256 canonical JSON is solid. Idempotency via `journal_append_requests` + `create_hash` is correctly implemented. Source trust enforcement (untrusted clients cannot claim `gateway_observed`) is a good design.

---

### 🔴 B1 — `CanonicalJSONRaw` does NOT produce deterministic output (BLOCKING)

`journal.MarshalCanonical()` calls `json.Marshal(v)` then `CanonicalJSONRaw()`, which round-trips through `json.NewDecoder().Decode(&v)` + `json.Marshal(v)`. But Go's `json.Marshal` does **not** guarantee deterministic key ordering for `map[string]any` — it sorts top-level keys, but nested values decoded from `any` can be `map[string]interface{}` whose iteration order is non-deterministic.

For the hash chain to be verifiable, `EntryHash()` and `CreateHash()` must produce identical bytes on every call with the same input. The current implementation works by coincidence (Go's `encoding/json` happens to sort map keys), but:

1. The `UseNumber()` decoder preserves number precision, but the re-marshal step may format numbers differently than the original (e.g., `1.0` vs `1` vs `1E0`).
2. If any nested `any` value is a `json.Number` that round-trips differently, the hash changes.

**Risk**: Hash chain verification failure on replay with different Go versions or JSON encoding behavior.

**Fix**: Either (a) use a proper canonical JSON library (RFC 8785 / JCS), or (b) document that Go `encoding/json` map key sorting IS the canonical form and add a regression test that verifies determinism across a set of tricky inputs (nested maps, numbers, unicode, null vs empty).

---

### 🔴 B2 — `AppendJournalEntries` double-read of `journal_append_requests` outside and inside transaction (BLOCKING)

In `pkg/datastore/journal.go` lines 3622–3650, the append path does:
1. First `selectAppendRequestTx` **before** `selectJournalTx` (line 3623)
2. Then `selectJournalTx` with `FOR UPDATE` (line 3634)
3. Then **another** `selectAppendRequestTx` (line 3641)

The first check at line 3623 runs inside the transaction but before the `FOR UPDATE` lock on the journal row. In TiDB pessimistic mode, this means the first read may see stale data (snapshot read) — the `FOR UPDATE` at line 3634 establishes the lock. If a concurrent append with the same `append_id` but different payload commits between the first read and the `FOR UPDATE`, the first check would return `ErrNotFound` (stale), and the second check (line 3641) would correctly catch it.

The first check is a **performance optimization** (fast path to skip the `FOR UPDATE` lock for replayed requests), but it creates confusion about correctness. If it reads stale data, it falls through to the `FOR UPDATE` path, which is correct — but if it reads committed data that then gets rolled back (TiDB SI anomaly), it could return success with stale `firstSeq`/`lastSeq`.

**Fix**: Remove the first `selectAppendRequestTx` before the `FOR UPDATE` lock, or move it outside the `InTx` entirely as a non-transactional fast-path with explicit fallthrough to the full transaction. The current placement inside the tx but before the `FOR UPDATE` gets the worst of both worlds.

---

### 🟡 N1 — `journal cat -f` uses `time.Sleep(time.Second)` busy-polling

`JournalCat` with `--follow` does `time.Sleep(time.Second)` in a tight loop (line 213). This is fine for Phase 1 but:
- No context cancellation check — Ctrl+C won't exit during the sleep
- No exponential backoff when no new entries arrive

Consider: `select { case <-ctx.Done(): return ctx.Err(); case <-time.After(pollInterval): }` and passing `context.Background()` → `signal.NotifyContext`.

---

### 🟡 N2 — `VerifyJournal` reads ALL entries via `ListJournalEntries` with `MaxLimit=1000` pages

For a journal with millions of entries, `VerifyJournal` will issue many sequential `SELECT ... LIMIT 1000` queries. This is acceptable for Phase 1 but should be documented as an O(N) full-scan operation. Consider adding a `--limit` or `--mode=head-only` flag for quick verification that just checks the head hash matches.

---

### 🟡 N3 — Entry search requires an anchor filter, but error message could be clearer

`hasJournalEntrySearchAnchor` (line 6951) rejects unfiltered entry searches with `"entry search requires type, status, actor, subject, or metadata filter"`. This is good for preventing full table scans, but the `--kind` filter alone triggers journal-level search, not entry search — this implicit mode switching could surprise users. Consider documenting this in the design spec or CLI help.

---

### 🟡 N4 — `nullJSON` treats `"null"` string as nil

```go
func nullJSON(raw []byte) interface{} {
    if len(raw) == 0 || string(raw) == "null" {
        return nil
    }
    return string(raw)
}
```

This means a `summary` containing the literal JSON value `null` is stored as SQL NULL, which round-trips as `nil` (no summary). This is probably intentional but should be documented — `null` summary and absent summary are the same thing.

---

### ✅ Positives

1. **Idempotency design is solid** — `create_hash` for journal create, `request_hash` + `append_id` for appends, with proper conflict detection.
2. **Source trust enforcement** — untrusted CLI writers cannot claim `gateway_observed` or `server_observed`. Permission model via scoped JWT claims is clean.
3. **Verify includes projection checks** — label hash verification, subject hash verification, and entry hash chain verification in a single pass.
4. **HMAC-signed cursors with TTL** — cursor includes query hash binding, tenant isolation, and expiry. Prevents cross-tenant cursor injection.
5. **Test coverage is excellent** — 586 lines datastore tests, 567 lines server tests, 131 lines journal package tests, 92 lines client tests, 195 lines e2e smoke test. Covers concurrent create, idempotency conflict, source trust, projection corruption detection, label pagination, permission enforcement.
6. **Design doc is thorough** — 2272 lines covering all concepts, CLI spec, filesystem projection, schema, integrity model, and non-goals.
7. **Schema has proper indexes** for all query patterns (type+observed, status+observed, actor+observed, subject hash, label hash).
8. **Dual PostgreSQL/TiDB schema** — `JournalDB9SchemaStatements` and `JournalTiDBSchemaStatements` kept structurally aligned.

---

### Summary

The journal feature is well-architected with strong integrity guarantees. The two blocking issues are:
- **B1**: Hash canonicalization must be provably deterministic — add regression tests or use JCS
- **B2**: Remove or relocate the pre-`FOR UPDATE` idempotency check to avoid confusion

After addressing these, this is GREEN.

qiffang:
**PR #422 review — head `710ead5`**: ❌ **REVISE** — 3 blockers, 5 issues

---

## B1 — Branch is stale: will revert PR #417, #418, #420 on merge (BLOCKER)

The branch forks from `260b09e`, which predates three merged PRs on `main`:
- `43be083` — Optimize FUSE handle hot paths (#417)
- `b650777` — Split-table refactor with POSIX permissions (#418)
- `eafdaf2` — Optimize delete performance (#420)

The diff shows 5,442 deletions, but most are **implicit reverts** of these PRs:
- Deletes `pkg/datastore/content.go`, `inode.go`, `semantic.go` (PR #418 split tables)
- Deletes `pkg/migrate/split_tables.go` + tests (PR #418 online migration)
- Reverts `EnsureParentDirs` to the pre-#418 version (no deterministic inode ID, no deadlock retry) — re-introduces the orphan inode race condition we just fixed
- Removes `File.Mode`, `NodeWithFile.HasMode`, `NodeWithFile.Mode` (PR #418 POSIX permissions)
- Removes `Store.useLegacyFiles`, `detectLegacyFiles` (PR #418 dual-write)
- Removes `open_handles.go`, `writeback.go`, FUSE chmod (PR #417/418 features)
- Removes `handleChmod`, `handleCreate`, `?kind=file|dir` delete (PR #418 API)

**Must rebase onto current `main` and resolve conflicts before review is meaningful.** The actual journal additions cannot be assessed in isolation when the diff is dominated by unintended reverts.

---

## B2 — `AppendResponse.Idempotent` is always `true` (BLOCKER)

Both the fresh-append path and the idempotent-replay path set `Idempotent: true`. Callers cannot distinguish "your entries were just written" from "your entries were already written earlier". Fresh appends should set `Idempotent: false`.

---

## B3 — `DeleteEmptyDir` drops context propagation (BLOCKER)

`tx.QueryRow(...)` and `tx.Exec(...)` replaced `tx.QueryRowContext(ctx, ...)` and `tx.ExecContext(ctx, ...)`. Context cancellation and tracing no longer propagate to the database driver.

---

## I1 — `VerifyJournal` loads all entries in MaxLimit pages — unbounded for large journals

`VerifyJournal` paginates with `MaxLimit=1000` but has no upper bound on total entries. For a journal with millions of entries this is an unbounded server-side scan. Consider a configurable max-entries cap or timeout.

---

## I2 — Idempotent `selectAppendRequestTx` called twice inside `InTx`

In `AppendJournalEntries`, `selectAppendRequestTx` is called once before the `FOR UPDATE` lock, then again inside it. The first call is outside the serializable window. This is correct but wasteful — remove the second call or document the first as a fast-path optimization.

---

## I3 — `NormalizeCreateRequest` rejects non-`self_reported` source

`NormalizeCreateRequest` hard-rejects any `source != SourceSelf`, but defaults empty to `SourceSelf`. This makes the `source` field on `CreateRequest` effectively useless — should be documented or removed from the type.

---

## I4 — Token `JournalPermissions` defaults to owner-all when empty

`journalPermissionsFromClaims` returns all permissions when `claims.JournalPermissions` is empty. Every existing token gets full journal admin access with no opt-in. Consider requiring explicit grants or documenting this as intentional.

---

## I5 — Missing test: closed journal append rejection

`ErrJournalClosed` and `current.ClosedAt != nil` check exist in code, but no test exercises a closed journal rejecting appends. There's also no API endpoint to close a journal.

---

**Summary**: Rebase onto current main (B1) is the prerequisite for further review. B2 and B3 are correctness bugs. I1-I5 are design/completeness issues.

qiffang:
PR #422 review at head `710ead5` — **REVISE**.

Blocking findings:

1. **The PR is stale and not mergeable against current `main`.** GitHub reports `mergeable=CONFLICTING`; a local merge leaves `pkg/server/server.go` unmerged. More importantly, `git diff origin/main..HEAD` shows `110 files changed, 8251 insertions, 8510 deletions`, including deletes/reverts of recently merged foundations such as PR #418 split-table files (`pkg/datastore/content.go`, `pkg/datastore/inode.go`, `pkg/datastore/semantic.go`, `pkg/migrate/split_tables.go`) and PR #419 ObjectGC/hash files (`pkg/backend/object_gc.go`, `pkg/datastore/storage_ref_hash.go`). Rebase onto current `main` and resolve conflicts while preserving PR #417/#418/#419/#420 invariants before this can be considered mergeable.

2. **The hash canonicalization does not meet the design contract.** `pkg/journal/journal.go` currently implements `MarshalCanonical` / `CanonicalJSONRaw` by decoding to `any` and re-marshalling with Go `encoding/json`. Go map key sorting is deterministic, but this is not the canonical form promised in `docs/design/agent-journal-design.md`: it does not normalize UTF-8 strings to NFC and it preserves numeric lexical differences through `json.Number` (`1` vs `1.0`). That means semantically identical normalized evidence from different clients can produce different `request_hash` / `entry_hash`, causing idempotency conflicts or non-portable verification. Either implement/lock down an explicit canonical JSON algorithm with tests for nested maps, Unicode normalization, and numeric normalization, or narrow the design doc to the exact Go-only canonicalization semantics.

Non-blocking issues to address or explicitly defer:

- `VerifyJournal` is intentionally O(entries), but the API has no request cap beyond context cancellation. The design doc calls out large-journal checkpoints; until seals/checkpoints exist, document this endpoint as investigation-only and bounded by caller/server timeout.
- `AppendJournalEntries` performs `selectAppendRequestTx` before and after the journal row lock. This is safe because of the second check, but the first read is only a fast path; add a short comment or remove it to avoid future edits treating it as the serialization check.
- Legacy owner tokens get `journal:admin`, and `journal:admin` includes source-elevation permissions. The design doc says this is intentional for legacy owner credentials, but the security boundary is important: ordinary agent/service credentials must be minted with explicit scoped `journal_permissions`, not owner tokens, if `gateway_observed` / `server_observed` evidence is meant to be trusted.
- There is no regression for closed journals rejecting appends even though the datastore has `ErrJournalClosed` and the append path checks `ClosedAt`. Add one when close/seal support lands, or add a direct datastore test now if closed rows can already exist.

Validation I ran on the current head:

- `/usr/local/go/bin/go test -c ./pkg/... ./cmd/...` passed
- `/usr/local/go/bin/go test ./pkg/journal ./pkg/client ./cmd/drive9/cli -count=1` passed for `pkg/journal` and CLI; `pkg/client` is blocked locally by the existing `rootless Docker not found` testcontainers dependency
- `/usr/local/go/bin/go vet ./pkg/journal ./pkg/datastore ./pkg/server ./pkg/client ./cmd/drive9/cli` passed

Rebase/conflict resolution is the first required step; after that the journal logic needs another full review on the integrated diff.

qiffang (commented):
## PR #422 Re-Review — Journal Logic (v2)

**Reviewer**: @dat9-dev1 (Dev Leader)
**Head**: `c08a119`
**Previous**: `710ead5` (REVISE, 2 blocking)
**Verdict**: 🟢 **GREEN**

---

### B1 (Canonical JSON determinism) — ✅ FIXED

New `writeCanonicalJSON()` is a proper recursive canonical writer:
- Explicit `sort.Strings(keys)` for map key ordering at every nesting level
- `json.Number` preserved as-is via `x.String()` — no round-trip formatting drift
- All JSON types handled explicitly with `default` returning error for unsupported types
- Regression test `TestCanonicalJSONRawIsDeterministicForNestedObjects` verifies identical output for differently-ordered inputs, repeated 20 times

This is correct and sufficient for Phase 1 integrity guarantees.

### B2 (Double-read before FOR UPDATE) — ✅ FIXED

Append path now has clean transaction flow:
1. `selectJournalTx` with `FOR UPDATE` (establishes lock)
2. `ClosedAt` check
3. Single `selectAppendRequestTx` (after lock, no stale reads)
4. Insert entries + update head

No pre-lock idempotency check remains. Clean and correct.

### Additional improvement: `Idempotent` semantics

Fresh appends now return `Idempotent: false`, replayed appends return `true`. Good semantic distinction for callers.

### Previous non-blocking items (N1-N4)

All acknowledged as Phase 1 acceptable. No changes required.

### Conclusion

Both blocking issues resolved correctly. Architecture, tests, and integrity model are solid. GREEN.

qiffang:
Implemented the PR #422 review/comment fixes and conflict resolution on `add-journal`.

Current head: `d5f1ab6`

What changed:
- Merged latest `main` and resolved the server conflict without reverting PR #417/#418/#419/#420 foundations.
- Fixed append response semantics: fresh appends return `idempotent: false`; replayed idempotent requests return `idempotent: true`.
- Removed the pre-lock idempotency read; append idempotency is now checked after locking the journal row.
- Replaced canonical JSON hashing with an explicit deterministic canonical writer and added nested-object regression coverage.
- Added closed-journal append rejection coverage.
- Updated the design doc for canonical JSON, create-source semantics, and append idempotency behavior.
- Fixed the CI lint issue in `VerifyJournal`.

Validation:
- GitHub `ci`: SUCCESS
- GitHub `drive9-server-local e2e smoke`: SUCCESS
- CodeRabbit: SUCCESS
- Local: targeted journal/cmd/CLI tests, `go test ./pkg/fuse`, compile gate, package vet, `make lint`, `git diff --check`

Local `pkg/client` runtime tests remain blocked by the existing rootless Docker/testcontainers limitation on this machine.

qiffang:
**Re-review at head `d5f1ab6`**: ✅ **GREEN**

All 3 blockers from my initial review are resolved:

- **B1** ✓ — Rebased onto current main (`b650777`). No more implicit reverts of PR #417/#418/#420. Diff is now pure journal additions (10.7k+/161-).
- **B2** ✓ — Fresh append `Idempotent: false`, idempotent replay `Idempotent: true`. Test updated with explicit assertions for both paths.
- **B3** ✓ — `DeleteEmptyDir` context propagation restored via main merge.

Additional fixes:
- I2 ✓ — Removed duplicate pre-lock `selectAppendRequestTx` call
- I5 ✓ — `TestAppendJournalRejectsClosedJournal` added
- Deterministic canonical JSON ✓ — `writeCanonicalJSON` with recursive sorted-key output + nested object regression test

**Remaining non-blocking items** (can be addressed in follow-up):
- I1: `VerifyJournal` unbounded scan for large journals
- I3: `CreateRequest.Source` field effectively dead
- I4: Token `JournalPermissions` defaults to owner-all when empty

CI: all green.

qiffang:
PR #422 re-review at head `d5f1ab6` — **GREEN**.

I re-reviewed the integrated diff after the conflict/comment fixes.

Verified:

- PR is now mergeable against current `main`; `git merge-tree` reports no conflict markers, and `git diff origin/main...HEAD` is back to the expected journal-only surface instead of implicit reverts of #417/#418/#419/#420.
- Fresh append now returns `idempotent=false`; idempotent replay returns `idempotent=true`, with datastore/server assertions updated.
- The duplicate pre-lock idempotency read was removed; append checks `journal_append_requests` after the journal row serialization point.
- Canonical JSON is now an explicit server-defined Phase 1 format: recursive sorted object keys, preserved arrays, preserved valid JSON number tokens, no insignificant whitespace. The design doc matches the implementation, and nested-object determinism is covered by regression tests.
- Closed journal append rejection is covered.
- The main-merge restored `DeleteEmptyDir` context propagation.
- Journal schema is included in TiDB app/auto and db9 schema paths without dropping split-table/ObjectGC changes from current main.

Validation run locally:

- `/usr/local/go/bin/go test ./pkg/journal ./cmd/drive9/cli ./cmd/drive9 -run 'Journal|Main' -count=1`
- `/usr/local/go/bin/go vet ./pkg/journal ./pkg/datastore ./pkg/server ./pkg/client ./cmd/drive9/cli`

`pkg/server` runtime package tests are still locally blocked by the existing `rootless Docker not found` testcontainers setup, but GitHub `ci`, local e2e smoke, and CodeRabbit are green.

Non-blocking follow-ups still worth tracking:

- `VerifyJournal` is O(entries) until seals/checkpoints exist; keep it documented as investigation-oriented and bounded by timeout.
- `CreateRequest.Source` is intentionally Phase-1 self-reported only; if kept in the API shape, the docs should remain explicit that stronger source evidence belongs on append entries.
- Legacy owner tokens map to `journal:admin`, including source elevation; production trusted-evidence writers should use explicit least-privilege `journal_permissions` tokens.

PR #424, opened 2026-05-13
Make fork provisioning asynchronous

srstack:
## Summary
- move fork API to /v1/fork and rename ctx_fork server files to fork
- make fork creation return 202 after creating the TiDB branch, then wait for ACTIVE and prepare metadata/quota asynchronously
- resume provisioning fork tenants on startup and surface user-facing provisioning messages from /v1/status
- allow CLI fork names to be omitted and accept 202 responses

## Tests
- go test ./pkg/tenant/starter
- go test ./cmd/drive9/cli
- go test ./pkg/server

## Summary by CodeRabbit

* **New Features**
  * Context fork command now auto-generates a name when omitted, eliminating the need to specify one manually
  * Fork operations return asynchronously with provisioning status and progress messages

* **Improvements**
  * Enhanced status feedback during fork provisioning with descriptive messages
  * Status endpoint now includes messages indicating fork preparation and branch creation progress

[Review Change Stack](https://app.coderabbit.ai/change-stack/mem9-ai/drive9/pull/424)

### DIFF
--- a/pkg/server/fork.go
+++ b/pkg/server/fork.go
@@ -39,6 +39,39 @@ var (
 	forkProvisionMaxBackoff     = 30 * time.Second
 )
 
+func (s *Server) startForkProvision(ctx context.Context, forkID string) {
+	s.startForkWorker(ctx, func(workerCtx context.Context) {
+		s.provisionForkTenantAsync(workerCtx, forkID)
+	})
+}
+
+func (s *Server) startForkCleanup(ctx context.Context, forkID string) {
+	s.startForkWorker(ctx, func(workerCtx context.Context) {
+		s.cleanupForkTenant(workerCtx, forkID)
+	})
+}
+
+func (s *Server) startForkWorker(ctx context.Context, fn func(context.Context)) {
+	workerCtx := backgroundWithTrace(ctx)
+	if s.forkWorkerCtx != nil {
+		workerCtx = contextWithTrace(s.forkWorkerCtx, ctx)
+	}
+
+	s.forkWorkerMu.Lock()
+	if s.forkWorkerClosed {
+		s.forkWorkerMu.Unlock()
+		logger.Warn(workerCtx, "fork_worker_start_after_close")
+		return
+	}
+	s.forkWorkerWG.Add(1)
+	s.forkWorkerMu.Unlock()
+
+	go func() {
+		defer s.forkWorkerWG.Done()
+		fn(workerCtx)
+	}()
+}
+
 func (s *Server) handleFork(w http.ResponseWriter, r *http.Request) {
 	switch r.Method {
 	case http.MethodPost:
@@ -229,7 +262,7 @@ func (s *Server) createForkTenant(ctx context.Context, sourceTenantID, displayNa
 		return nil, err
 	}
 
-	go s.provisionForkTenantAsync(backgroundWithTrace(ctx), forkID)
+	s.startForkProvision(ctx, forkID)
 
 	return &forkResponse{
 		TenantID:       forkID,
@@ -243,16 +276,26 @@ func (s *Server) createForkTenant(ctx context.Context, sourceTenantID, displayNa
 }
 
 func (s *Server) provisionForkTenantAsync(ctx context.Context, forkID string) {
-	ctx = backgroundWithTrace(ctx)
+	ctx = ensureTrace(ctx)
 	logger.Info(ctx, "fork_provision_started", zap.String("tenant_id", forkID))
 	deadline := time.Now().Add(forkProvisionRetryWindow)
 	backoff := forkProvisionInitialBackoff
 	attempt := 1
 	for {
+		select {
+		case <-ctx.Done():
+			logger.Info(ctx, "fork_provision_stopped", zap.String("tenant_id", forkID), zap.Error(ctx.Err()))
+			return
+		default:
+		}
 		if err := s.provisionForkTenantOnce(ctx, forkID); err == nil {
 			logger.Info(ctx, "fork_provision_ok", zap.String("tenant_id", forkID), zap.Int("attempt", attempt))
 			return
 		} else {
+			if errors.Is(err, context.Canceled) || errors.Is(err, context.DeadlineExceeded) {
+				logger.Info(ctx, "fork_provision_stopped", zap.String("tenant_id", forkID), zap.Int("attempt", attempt), zap.Error(err))
+				return
+			}
 			var fatal *forkFatalProvisionError
 			if errors.As(err, &fatal) || time.Now().After(deadline) {
 				logger.Error(ctx, "fork_provision_failed",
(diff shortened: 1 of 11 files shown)


mornyx (approved):
## PR #424 Review: Make fork provisioning asynchronous

Overall this is a well-designed PR. The async split (CreateBranch → 202 → background provision) is the right architecture, and the idempotent `ensureForkBranchConnection` handles resume/retry correctly. Below are findings:

---

### 🔴 Goroutine leak on server shutdown

**File:** `pkg/server/fork.go`, `provisionForkTenantAsync`

```go
go s.provisionForkTenantAsync(backgroundWithTrace(ctx), forkID)
```

The async goroutine uses `context.Background()` (via `backgroundWithTrace`), so it never gets cancelled on `Server.Close()`. The `time.Sleep` calls in the retry loop have no cancellation point — a provisioning fork could keep the goroutine alive for up to 30 minutes after shutdown.

**Recommendation:** Derive from a server-level context that gets cancelled in `Server.Close()`, and use `select { case <-ctx.Done(): return; case <-time.After(sleepFor): }` instead of bare `time.Sleep`.

---

### 🟡 Double `backgroundWithTrace` call

Both the caller (`createForkTenant`) and callee (`provisionForkTenantAsync`) call `backgroundWithTrace`. It is idempotent so no bug, but it signals ownership confusion. Suggestion: drop the outer call:

```go
go s.provisionForkTenantAsync(ctx, forkID)
```

---

### 🟡 `TestCreateForkPartialBranchProvisionErrorPersistsBranchAndKeepsRoot` — tight timeout

The test sets `forkProvisionRetryWindow = 50 * time.Millisecond`. Under extreme CI load this could expire before the async goroutine even runs, making the test flaky. Consider at least 500ms.

---

### 🟡 Missing context cancellation in `Server.Close()`

`Server.Close()` currently stops `semanticWorker` and `objectGCWorker` but has no way to stop in-flight fork provisioning goroutines. If you adopt the context cancellation suggestion above, add a `sync.WaitGroup` + cancel to `Close()`.

---

### 🟢 Nitpicks

- `provisionForkTenantOnce` re-queries the source tenant on every retry — correct but slightly redundant after the first check. Not a blocker.
- `forkProvisioningMessage` strings are hardcoded; could be package-level vars for testability (low priority).
- The inline comment about the new auth test lacking a WHERE clause is incorrect — `TestTenantStatusReturnsForkProvisioningMessage` correctly uses `WHERE id = ?` with `rt.tenantID`.

pkg/server/fork.go
```diff
@@ -238,81 +228,240 @@ func (s *Server) createForkTenant(ctx context.Context, sourceTenantID, displayNa
...
+	ctx = backgroundWithTrace(ctx)
+	logger.Info(ctx, "fork_provision_started", zap.String("tenant_id", forkID))
+	deadline := time.Now().Add(forkProvisionRetryWindow)
+	backoff := forkProvisionInitialBackoff
+	attempt := 1
+	for {
+		if err := s.provisionForkTenantOnce(ctx, forkID); err == nil {
+			logger.Info(ctx, "fork_provision_ok", zap.String("tenant_id", forkID), zap.Int("attempt", attempt))
+			return
+		} else {
```
mornyx, line 294: The async goroutine uses `context.Background()` (via `backgroundWithTrace`), so it never gets cancelled on server shutdown. `time.Sleep` in the retry loop has no cancellation point.

**Suggestion:** Derive from a server-level shutdown context, and use:
```go
select {
case <-ctx.Done():
    return
case <-time.After(sleepFor):
}
```

pkg/server/fork.go
```diff
@@ -238,81 +228,240 @@ func (s *Server) createForkTenant(ctx context.Context, sourceTenantID, displayNa
...
+
+	return &forkResponse{
 		TenantID:       forkID,
 		Name:           displayName,
 		APIKey:         apiToken,
-		Status:         string(meta.TenantActive),
+		Status:         string(meta.TenantProvisioning),
+		Message:        forkProvisioningMessage(&forkRoot),
 		ParentTenantID: source.ID,
 		Storage:        "shared",
```
mornyx, line 274: Both the caller and callee call `backgroundWithTrace` — the caller at line ~241 and the callee at line ~253. It's idempotent so no bug, but it signals ambiguity about who owns context sanitization. Suggestion: drop the outer call so the pattern is clear:
```go
go s.provisionForkTenantAsync(ctx, forkID)
```

pkg/server/fork_test.go
```diff
@@ -163,18 +186,35 @@ func waitForCondition(t *testing.T, fn func() bool) {
 }
 
 func TestCreateForkPartialBranchProvisionErrorPersistsBranchAndKeepsRoot(t *testing.T) {
+	origWindow, origInitialBackoff, origMaxBackoff := forkProvisionRetryWindow, forkProvisionInitialBackoff, forkProvisionMaxBackoff
+	forkProvisionRetryWindow = 50 * time.Millisecond
+	forkProvisionInitialBackoff = 5 * time.Millisecond
```
mornyx, line 201: 50ms retry window is very tight — under extreme CI load the async goroutine might not even start before the deadline expires, making this test flaky. Consider at least 200-500ms.

mornyx (approved):
## Second-round review

All three issues from the first round have been fixed. Code quality improved further.

---

### ✅ Fixed

**1. Goroutine leak** — `forkWorkerCtx` + `forkWorkerCancel` + `forkWorkerWG` + `forkWorkerClosed` provide full lifecycle management:
- `startForkWorker` as the unified entry point with closed guard + WaitGroup
- `Close()` cancels first, then waits
- `provisionForkTenantAsync` uses `time.NewTimer` + `select ctx.Done()` instead of bare `time.Sleep`, with correct channel drain after `Timer.Stop()`
- `context.Canceled` / `context.DeadlineExceeded` short-circuits without erroneously calling `markForkFailed`

**2. Double backgroundWithTrace** — `ensureTrace` only adds a trace ID without replacing the parent context. Inside `startForkWorker`, `contextWithTrace(s.forkWorkerCtx, ctx)` attaches the trace to the cancellable parent. Clear ownership.

**3. Test timeout** — 50ms → 500ms, safely above CI variance.

---

### 🟢 Additional observations (non-blocking)

**`resumeForkCleanup` still uses raw goroutines**

`resumeDeletingForkTenants` → `resumeForkCleanup` still spawns goroutines directly via `go func` + `backgroundWithTrace(ctx)`, without going through `startForkWorker`. This is low-impact — cleanup operations lack the 30-minute retry loop that provisioning has, and they finish quickly. Could switch to `startForkCleanup` for consistency in a follow-up.

**`fakeBranchProvisioner.CreateBranch` / `WaitForBranchActive` `ctx.Done()` check**

```go
select {
case <-ctx.Done():
    return nil, ctx.Err()
default:
}
```

Good defensive pattern — ensures the fake also respects context cancellation in tests.

---

### Verdict

All 🔴 and 🟡 issues from the first round are resolved. Ready to merge.

PR #426, opened 2026-05-13
Polish fork provisioning UX and extend CLI smoke coverage

srstack:
## Summary
- align fork provisioning status messaging with create until a branch exists, then show a single migration message
- return `tenant_id` from `/v1/provision` and local shim provision responses so CLI output stays populated
- make `drive9 fs cp` treat directory destinations like shell `cp` and add CLI tests for all three copy directions
- extend `e2e/cli-smoke-test.sh` with live fork coverage and directory-target cp checks

## Validation
- `go test ./pkg/server ./cmd/drive9/cli`
- `DRIVE9_BASE=http://k8s-dat9-dat9serv-d5e02e7d07-1645488597.ap-southeast-1.elb.amazonaws.com RUN_CLI_SEMANTIC_CHECKS=0 RUN_CLI_UPLOAD_LIMIT_BOUNDARY=0 CLI_LARGE_FILE_MB=1 CLI_BATCH_SMALL_FILE_COUNT=2 bash e2e/cli-smoke-test.sh`

## Self-review
- reviewed the final diff for copy-target resolution, fork status messaging, and provision response shape
- live smoke passed end to end with the new fork and cp directory checks enabled

## Summary by CodeRabbit

* **New Features**
  * Copy command resolves directory-style destinations and preserves source filenames.
  * Provision API now includes tenant_id in provision responses.

* **Improvements**
  * Fork/ctx workflow: forking no longer auto-switches contexts; guidance added to CLI output and help text updated.
  * CLI smoke tests gain optional fork-context coverage and more robust retry/cleanup behavior.

* **Documentation**
  * Updated CLI smoke-test docs and README to reflect provisioning response and fork workflow.

* **Tests**
  * Added/expanded tests for copy destination semantics and fork/provision behavior.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mem9-ai/drive9/pull/426)

### DIFF
--- a/cmd/drive9/cli/cp.go
+++ b/cmd/drive9/cli/cp.go
@@ -182,7 +182,7 @@ func resolveRemoteCopyTarget(ctx context.Context, c *client.Client, remotePath, 
 		}
 		return pathpkg.Join(remotePath, sourceBase), nil
 	}
-	info, err := c.Stat(remotePath)
+	info, err := c.StatCtx(ctx, remotePath)
 	if err == nil && info.IsDir {
 		if sourceBase == "" {
 			return "", fmt.Errorf("destination %q requires a file name", remotePath)
@@ -196,7 +196,7 @@ func resolveRemoteCopyTarget(ctx context.Context, c *client.Client, remotePath, 
 }
 
 func resolveLocalCopyTarget(localPath, sourceBase string) (string, error) {
-	if strings.HasSuffix(localPath, string(os.PathSeparator)) {
+	if strings.HasSuffix(localPath, string(os.PathSeparator)) || strings.HasSuffix(localPath, "/") {
 		if sourceBase == "" {
 			return "", fmt.Errorf("destination %q requires a file name", localPath)
 		}
--- a/cmd/drive9/cli/cp.go
+++ b/cmd/drive9/cli/cp.go
@@ -6,6 +6,8 @@ import (
 	"fmt"
 	"io"
 	"os"
+	pathpkg "path"
+	"path/filepath"
 	"strings"
 	"unicode/utf8"
 
@@ -113,12 +115,16 @@ func Cp(c *client.Client, args []string) error {
 
 	switch {
 	case src == "-" && dstIsRemote:
+		resolvedDst, err := resolveRemoteCopyTarget(ctx, c, dstRP.Path, "")
+		if err != nil {
+			return err
+		}
 		data, err := io.ReadAll(os.Stdin)
 		if err != nil {
 			return fmt.Errorf("read stdin: %w", err)
 		}
 		if appendMode {
-			return c.AppendStream(ctx, dstRP.Path, bytes.NewReader(data), int64(len(data)), printProgress)
+			return c.AppendStream(ctx, resolvedDst, bytes.NewReader(data), int64(len(data)), printProgress)
 		}
 		var opts []client.WriteOption
 		if tags != nil {
@@ -127,7 +133,7 @@ func Cp(c *client.Client, args []string) error {
 		if description != "" {
 			opts = append(opts, client.WithDescription(description))
 		}
-		summary, err := c.WriteStreamWithSummary(ctx, dstRP.Path, bytes.NewReader(data), int64(len(data)), printProgress, opts...)
+		summary, err := c.WriteStreamWithSummary(ctx, resolvedDst, bytes.NewReader(data), int64(len(data)), printProgress, opts...)
 		if err != nil {
 			return err
 		}
@@ -138,23 +144,75 @@ func Cp(c *client.Client, args []string) error {
 		return streamToStdout(ctx, c, srcRP.Path)
 
 	case !srcIsRemote && dstIsRemote:
+		resolvedDst, err := resolveRemoteCopyTarget(ctx, c, dstRP.Path, filepath.Base(src))
+		if err != nil {
+			return err
+		}
 		if appendMode {
-			return appendFile(ctx, c, src, dstRP.Path)
+			return appendFile(ctx, c, src, resolvedDst)
 		}
 		if resume {
-			return resumeUploadWithTags(ctx, c, src, dstRP.Path, tags)
+			return resumeUploadWithTags(ctx, c, src, resolvedDst, tags)
 		}
-		return uploadFileWithTagsAndDescription(ctx, c, src, dstRP.Path, tags, description)
+		return uploadFileWithTagsAndDescription(ctx, c, src, resolvedDst, tags, description)
 
 	case srcIsRemote && !dstIsRemote:
-		return downloadFile(ctx, c, srcRP.Path, dst)
+		resolvedDst, err := resolveLocalCopyTarget(dst, pathpkg.Base(srcRP.Path))
+		if err != nil {
+			return err
+		}
+		return downloadFile(ctx, c, srcRP.Path, resolvedDst)
 
 	case srcIsRemote && dstIsRemote:
-		return c.Copy(srcRP.Path, dstRP.Path)
+		resolvedDst, err := resolveRemoteCopyTarget(ctx, c, dstRP.Path, pathpkg.Base(srcRP.Path))
+		if err != nil {
+			return err
+		}
+		return c.Copy(srcRP.Path, resolvedDst)
 
 	default:
 		return fmt.Errorf("at least one path must be remote (e.g. :/path or mydb:/path)")
 	}
+}
+
+func resolveRemoteCopyTarget(ctx context.Context, c *client.Client, remotePath, sourceBase string) (string, error) {
+	if strings.HasSuffix(remotePath, "/") {
+		if sourceBase == "" {
+			return "", fmt.Errorf("destination %q requires a file name", remotePath)
+		}
+		return pathpkg.Join(remotePath, sourceBase), nil
+	}
+	info, err := c.Stat(remotePath)
+	if err == nil && info.IsDir {
+		if sourceBase == "" {
+			return "", fmt.Errorf("destination %q requires a file name", remotePath)
+		}
+		return pathpkg.Join(remotePath, sourceBase), nil
+	}
+	if err != nil && !client.IsNotFound(err) {
+		return "", err
+	}
+	return remotePath, nil
+}
+
+func resolveLocalCopyTarget(localPath, sourceBase string) (string, error) {
+	if strings.HasSuffix(localPath, string(os.PathSeparator)) {
+		if sourceBase == "" {
+			return "", fmt.Errorf("destination %q requires a file name", localPath)
+		}
+		return filepath.Join(localPath, sourceBase), nil
+	}
+	info, err := os.Stat(localPath)
+	if err == nil && info.IsDir() {
+		if sourceBase == "" {
+			return "", fmt.Errorf("destination %q requires a file name", localPath)
+		}
+		return filepath.Join(localPath, sourceBase), nil
+	}
+	if err != nil && !os.IsNotExist(err) {
+		return "", fmt.Errorf("stat %s: %w", localPath, err)
+	}
+	return localPath, nil
 }
 
 func uploadFile(ctx context.Context, c *client.Client, localPath, remotePath string, description string) error {
--- a/cmd/drive9/cli/cp_test.go
+++ b/cmd/drive9/cli/cp_test.go
@@ -460,6 +460,108 @@ func TestCpUploadWithTagsSendsHeaders(t *testing.T) {
 	}
 }
 
+func TestCpUploadToRemoteDirectoryKeepsSourceName(t *testing.T) {
+	localPath := filepath.Join(t.TempDir(), "upload.txt")
+	if err := os.WriteFile(localPath, []byte("hello dir"), 0o644); err != nil {
+		t.Fatalf("WriteFile(local): %v", err)
+	}
+
+	var gotPath string
+	var gotBody []byte
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		switch r.Method + " " + r.URL.Path {
+		case "HEAD /v1/fs/dir":
+			w.Header().Set("X-Dat9-IsDir", "true")
+			w.WriteHeader(http.StatusOK)
+		case "PUT /v1/fs/dir/upload.txt":
+			gotPath = r.URL.Path
+			body, err := io.ReadAll(r.Body)
+			if err != nil {
+				http.Error(w, err.Error(), http.StatusInternalServerError)
+				return
+			}
+			gotBody = append([]byte(nil), body...)
+			w.WriteHeader(http.StatusOK)
+		default:
+			http.NotFound(w, r)
+		}
+	}))
+	defer srv.Close()
+
+	c := client.New(srv.URL, "")
+	c.SetSmallFileThresholdForTests(client.DefaultSmallFileThreshold)
+	if err := Cp(c, []string{localPath, ":/dir"}); err != nil {
+		t.Fatalf("Cp(upload to dir): %v", err)
+	}
+	if gotPath != "/v1/fs/dir/upload.txt" {
+		t.Fatalf("PUT path = %q, want %q", gotPath, "/v1/fs/dir/upload.txt")
+	}
+	if got := string(gotBody); got != "hello dir" {
+		t.Fatalf("uploaded body = %q, want %q", got, "hello dir")
+	}
+}
+
+func TestCpDownloadToLocalDirectoryKeepsSourceName(t *testing.T) {
+	dstDir := t.TempDir()
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		switch r.Method + " " + r.URL.Path {
+		case "HEAD /v1/fs/src.txt":
+			w.Header().Set("Content-Length", "11")
+			w.Header().Set("X-Dat9-IsDir", "false")
+			w.WriteHeader(http.StatusOK)
+		case "GET /v1/fs/src.txt":
+			_, _ = io.WriteString(w, "hello local")
+		default:
+			http.NotFound(w, r)
+		}
+	}))
+	defer srv.Close()
+
+	c := client.New(srv.URL, "")
+	c.SetSmallFileThresholdForTests(client.DefaultSmallFileThreshold)
+	if err := Cp(c, []string{":/src.txt", dstDir}); err != nil {
+		t.Fatalf("Cp(download to dir): %v", err)
+	}
+	got, err := os.ReadFile(filepath.Join(dstDir, "src.txt"))
+	if err != nil {
+		t.Fatalf("ReadFile(downloaded): %v", err)
+	}
+	if string(got) != "hello local" {
+		t.Fatalf("downloaded body = %q, want %q", string(got), "hello local")
+	}
+}
+
+func TestCpRemoteToRemoteDirectoryKeepsSourceName(t *testing.T) {
+	var gotPath string
+	var gotSource string
+	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		switch r.Method + " " + r.URL.Path {
+		case "HEAD /v1/fs/dstdir":
+			w.Header().Set("X-Dat9-IsDir", "true")
+			w.WriteHeader(http.StatusOK)
+		case "POST /v1/fs/dstdir/src.txt":
+			if !r.URL.Query().Has("copy") {
+				t.Fatalf("expected ?copy on %s", r.URL.String())
+			}
+			gotPath = r.URL.Path
+			gotSource = r.Header.Get("X-Dat9-Copy-Source")
+			w.WriteHeader(http.StatusOK)
+		default:
+			http.NotFound(w, r)
+		}
+	}))
+	defer srv.Close()
+
+	c := client.New(srv.URL, "")
+	c.SetSmallFileThresholdForTests(client.DefaultSmallFileThreshold)
+	if err := Cp(c, []string{":/src.txt", ":/dstdir"}); err != nil {
+		t.Fatalf("Cp(remote to remote dir): %v", err)
+	}
+	if gotPath != "/v1/fs/dstdir/src.txt" || gotSource != "/src.txt" {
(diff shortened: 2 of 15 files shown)


qiffang (commented):
## PR #426 Review — @dat9-dev1 — HEAD `072e1e1`

**Verdict: GREEN** (1 non-blocking suggestion)

### What this PR does
4 changes bundled together:
1. **`drive9 fs cp` directory target resolution** — `cp local.txt :/dir` now auto-appends `local.txt` (like shell `cp`)
2. **Fork provisioning message simplification** — 3 granular messages → empty or single "Migrating fork data"
3. **`/v1/provision` returns `tenant_id`** — both production and local shim endpoints
4. **E2E coverage** — fork workflow + cp directory target semantics

### Correctness ✓

**cp directory resolution:**
- `resolveRemoteCopyTarget`: trailing `/` fast-path (no Stat needed) ✓, existing-dir detection via `c.Stat` ✓, not-found passthrough ✓, non-404 error propagation ✓
- `resolveLocalCopyTarget`: same logic with `os.Stat` + `os.PathSeparator` ✓
- Remote paths use `path.Join`, local paths use `filepath.Join` — correct cross-platform ✓
- Stdin source passes `sourceBase=""` → directory destinations correctly error with "requires a file name" ✓
- All 4 cp branches (stdin→remote, local→remote, remote→local, remote→remote) call the resolver ✓

**Fork provisioning messages:**
- No branch yet → empty message (silent during initial creation) ✓
- Branch exists → single "Migrating fork data" ✓
- Tests split into two: `WithoutReadyBranchOmitsMessage` + `BranchShowsMigrationMessage` ✓

**Provision `tenant_id`:**
- Both `handleProvision` and `handleLocalTenantProvision` now return `tenant_id` ✓
- Tests verify the field ✓

### Test Coverage ✓
- `TestCpUploadToRemoteDirectoryKeepsSourceName` — local→remote dir ✓
- `TestCpDownloadToLocalDirectoryKeepsSourceName` — remote→local dir ✓
- `TestCpRemoteToRemoteDirectoryKeepsSourceName` — remote→remote dir ✓
- Fork e2e: provision → poll → write → read → delete lifecycle ✓
- cp dir e2e: all 3 directions ✓

### Non-blocking (N1): cp directory section numbering

E2E test labels the cp directory section as `[4.0]` which runs after `[4]`. Suggest renumbering to `[4.1]` or inserting before `[4]` as `[3.2]` to keep chronological ordering intuitive. Minor — doesn't affect correctness.

### CLI Taste Assessment

The cp directory resolution is the right UX move. Before this PR, `drive9 fs cp local.txt :/mydir` when `/mydir` is a directory would either create a file named `/mydir` or fail — surprising behavior that violates every user's muscle memory from `cp`. The implementation is clean: two small resolver functions, explicit error messages, and the trailing-`/` fast-path avoids unnecessary network round-trips when users are explicit about directory destinations.

Fork message simplification is also good taste — users don't need to see internal provisioning state transitions. Silent until there's something meaningful to say.

qiffang (commented):
## adversary-2 review: `072e1e1` — ✅ GREEN

Reviewed all 9 changed files (341+/29−).

### cp directory target resolution
Correct. Both `resolveRemoteCopyTarget` and `resolveLocalCopyTarget` match shell `cp` semantics:
- Trailing `/` → fast-path join (no Stat) ✅
- Existing dir → Stat-then-join ✅
- Not-found → literal target ✅
- Stat error (non-404) → bubble up ✅
- stdin + dir target → rejection ("requires a file name") ✅
- `pathpkg` for remote, `filepath` for local ✅

### Fork provisioning message
Good taste. 3 internal-state messages → 2 states: silent (no branch) / "Migrating fork data" (branch exists).

### `/v1/provision` tenant_id
Additive, backward-compatible. Both endpoints covered.

### Tests
3 unit tests (upload/download/remote-to-remote × dir target). 2 fork provisioning tests. E2E fork lifecycle + cp directory targets.

### Non-blocking
- **N1**: e2e `[4.0]` after `[4]` — suggest `[4.1]`

qiffang (changes requested):
## PR #426 Review — dat9-dev2 — HEAD `072e1e1`

Verdict: **REVISE**

I focused on CLI interaction quality, e2e behavior, and the provision/fork UX contract. The `fs cp` directory-target direction is right, but the current PR is not merge-ready because the new behavior breaks local e2e and leaves one API response inconsistency.

### B1 — API smoke still asserts the old `/v1/provision` shape

`pkg/server/server.go` now returns `tenant_id` from `POST /v1/provision`, but `e2e/api-smoke-test.sh` still asserts the response keys are exactly `api_key,status`.

This is not a flaky gate: GitHub local e2e fails exactly here:

- `FAIL provision response only has api_key+status (want=api_key,status got=api_key,status,tenant_id)`

Why this matters: if the PR changes the public provision response shape, the smoke test must encode the new contract. Update the smoke test to assert `tenant_id` is present and stop asserting the old exact key set.

### B2 — CLI fork smoke is enabled by default in an environment where fork is unsupported

`e2e/cli-smoke-test.sh` defaults `RUN_CLI_FORK_CHECKS=1`, while `.github/workflows/local-e2e.yml` runs that script against `drive9-server-local` without a branch-capable fork provisioner. The script then hard-fails on:

- `ctx: ctx fork failed (HTTP 404): fork not enabled`

This is a CLI/e2e design issue, not just CI noise. A generic local smoke test should not require a cloud-only fork backend by default. Either default fork checks off for `cli-smoke-test.sh`, set `RUN_CLI_FORK_CHECKS=0` in local-e2e, or make the fork section explicitly skip when `/v1/fork` returns unsupported. Keep live fork coverage opt-in for environments that actually support branch provisioning.

### B3 — `/v1/provision` JSON response lacks `Content-Type` in the production handler

`handleProvision` writes a JSON body but does not set `Content-Type: application/json`; `handleLocalTenantProvision` does. Since this PR changes the response contract, keep the API explicit and consistent by setting the header before `WriteHeader` in both paths.

### Non-blocking CLI taste notes

- `ctx fork --json` still emits only the raw server fork response. If the user omits `<new>`, the CLI saves a generated local context name but JSON output does not include that name, making scripts unable to discover the context they just created from stdout. Not blocking this PR, but worth fixing with the fork UX polish.
- The remote trailing-slash fast path (`drive9 fs cp local :/dir/`) appends the source basename without verifying the directory exists. Because Drive9 writes auto-create parents, this can create `:/dir/<basename>` even if `:/dir` did not exist. That may be acceptable Drive9 behavior, but it is not exactly shell `cp`; document/decide explicitly.

### Validation

Local:
- `go test ./cmd/drive9/cli -run 'Cp|CtxFork|MainIncludes|RunFS' -count=1` passed
- `bash -n e2e/cli-smoke-test.sh e2e/api-smoke-test.sh` passed
- `go test -c ./cmd/drive9/cli ./pkg/server` passed
- `go vet ./cmd/drive9/cli ./pkg/server` passed
- `go test ./pkg/server -run 'Provision|ForkProvisioning|LocalTenantShim'` remains locally blocked by the existing `rootless Docker not found` testcontainers issue

GitHub:
- `drive9-server-local e2e smoke` is red due B1/B2 above
- `code-ci` is also red on `TestSemanticWorkerPanicStopsLeaseRenewalWithFailpoint`; that looks unrelated to this PR, but the PR still needs a green gate before merge

qiffang (changes requested):
## PR #426 second review — adversary-1 — head `072e1e1`

Verdict: **REVISE**

I focused on CLI interaction quality plus the live smoke gates. There are two hard CI/smoke blockers and one CLI UX blocker.

### B1 — `cli-smoke-test.sh` enables fork smoke by default against `drive9-server-local`, where fork is not enabled

`e2e/cli-smoke-test.sh:23` defaults `RUN_CLI_FORK_CHECKS=1`, and lines 300-303 immediately run `drive9 ctx fork`. In the current GitHub `drive9-server-local e2e smoke`, this fails with:

```text
ctx: ctx fork failed (HTTP 404): fork not enabled
```

That is not a flaky test; the local shim does not configure a branch provisioner. A feature-specific fork smoke should be opt-in for a fork-capable hosted/dev deployment, or the script should probe and skip cleanly when `/v1/fork` is unavailable. Running it by default makes the standard local CLI smoke unusable.

### B2 — API smoke/docs still assert the old provision response shape

This PR intentionally adds `tenant_id` to `/v1/provision`, but `e2e/api-smoke-test.sh:239-240` still asserts the response keys are exactly `api_key,status`. GitHub e2e is red with:

```text
FAIL provision response only has api_key+status (want=api_key,status got=api_key,status,tenant_id)
```

`e2e/README.md:164` also documents the old contract. If `tenant_id` is now part of the response contract, the smoke should assert `tenant_id` is non-empty and docs should be updated. Exact key-set assertions should either include `tenant_id` or be removed if additive fields are allowed.

### B3 — `ctx fork --use` human output hides that the selected context is still provisioning

With async fork provisioning, `cmd/drive9/cli/ctx.go:431-434` prints only:

```text
Forked owner -> fork
Mode: copy-on-write
Now using ctx fork
```

But the response status is still `provisioning`, and the new context cannot run normal `fs` commands until `/v1/status` becomes `active`. The e2e script handles this by polling manually, but the human CLI path does not surface it. This is poor interaction design: it says “Now using” without telling the user that subsequent commands may fail with tenant-provisioning errors.

For this PR’s stated UX goal, the non-JSON output should include the returned status and a clear next action, e.g. that the fork is provisioning and the user should wait/check status before using it. Add a CLI output test for the provisioning case.

### Non-blocking

- `pkg/server/server.go:2526` still writes a JSON `/v1/provision` response without setting `Content-Type: application/json`; the local shim does set it. This is worth normalizing while touching the response.
- `pkg/server/fork.go:137-142` reports “Migrating fork data” as soon as `branch_id` exists, even while the branch may still be waiting for connection readiness. A broader message such as “Preparing fork workspace” would be more accurate.
- GitHub `ci` also has a failpoint failure in `TestSemanticWorkerPanicStopsLeaseRenewalWithFailpoint`; it looks unrelated to this diff, but the PR still needs a green rerun before merge.

Validation I ran locally:

- `/usr/local/go/bin/go test ./cmd/drive9/cli -run 'Cp|Ctx|Create' -count=1`
- `/usr/local/go/bin/go vet ./cmd/drive9/cli ./pkg/server`
- `bash -n e2e/cli-smoke-test.sh e2e/api-smoke-test.sh`

`pkg/server` runtime tests remain locally blocked by `rootless Docker not found`.

srstack:
Keeping the fork provisioning message as-is by design: before `branch_id` exists there is no extra message, and once branch creation has started we intentionally show the same migration message during both branch readiness and post-branch data migration so the CLI UX stays aligned with async create.

qiffang (changes requested):
## PR #426 re-review — dat9-dev2 — HEAD `6d7dde9`

Verdict: **REVISE** (1 remaining CLI UX blocker)

The three blockers from my prior review are fixed:

- API smoke now expects `tenant_id`, `api_key`, and `status`.
- Local CLI smoke now probes `/v1/fork` and cleanly skips fork checks when the server does not expose fork support.
- Production `/v1/provision` now sets `Content-Type: application/json`.

GitHub `ci`, local e2e smoke, and CodeRabbit are green. The `fs cp` directory-target behavior still looks correct.

### B1 — `ctx fork --use` still presents a provisioning context as immediately usable

`cmd/drive9/cli/ctx.go` still prints:

```text
Forked prod -> exp
Mode: copy-on-write
Now using ctx exp
```

for a `202 Accepted` fork response whose status is still `provisioning`. That output is misleading: the CLI has switched local context, but the new tenant is not active yet and normal `fs` commands can still fail until `/v1/status` becomes `active`. The e2e script knows this and polls explicitly; the human CLI path should surface the same state.

This is a CLI taste/correctness issue for this PR because the PR is explicitly polishing fork provisioning UX. Non-JSON output should include the returned status and a clear next action before/after `Now using`, e.g. that the fork context is saved/selected but still provisioning and the user should wait/check status before using it. Add a CLI output regression test for the provisioning case.

### Non-blocking

- `e2e/README.md` and `e2e/AGENTS.md` still contain old prose saying `/v1/provision` returns only `api_key` + `status`; update those while touching smoke docs.
- `drive9_ctx_retry` treats broad `not found` as retryable. Given it is bounded and limited to fork workflow smoke, acceptable, but a short comment would prevent future misuse.

### Validation

Local:
- `/usr/local/go/bin/go test ./cmd/drive9/cli -run 'Cp|CtxFork|MainIncludes|RunFS' -count=1`
- `bash -n e2e/cli-smoke-test.sh e2e/api-smoke-test.sh`
- `/usr/local/go/bin/go test -c ./cmd/drive9/cli ./pkg/server`
- `/usr/local/go/bin/go vet ./cmd/drive9/cli ./pkg/server`

GitHub:
- `ci` success
- `drive9-server-local e2e smoke` success
- CodeRabbit success

srstack:
Addressed the remaining CLI UX point in commit `59352fe`.

Changes included:
- `ctx fork --use` human-readable output now surfaces `Status: provisioning`
- Added an explicit follow-up message telling users to wait for `drive9 status` to report `active` before running `fs` commands
- Added CLI regression coverage for the provisioning output
- Updated the remaining e2e docs for the `/v1/provision` response shape
- Added a short comment explaining the bounded `drive9_ctx_retry` not-found retry in fork smoke

Local validation:
- `go test ./cmd/drive9/cli -run 'CtxFork' -count=1`
- `bash -n e2e/cli-smoke-test.sh e2e/api-smoke-test.sh`

qiffang (changes requested):
REVISE — latest head `59352fe`

The previous `ctx fork --use` provisioning-output issue is only partially fixed.

Blocking issue:

- **B1: `ctx fork --use` now points users at a nonexistent command.**
  - `cmd/drive9/cli/ctx.go` prints: `Wait for \`drive9 status\` to report \`active\` before running \`fs\` commands.`
  - There is no top-level `drive9 status` command in `cmd/drive9/main.go`; the CLI command list is `create`, `ctx`, `fs`, `vault`, `journal`, `mount`, `umount`, and `doctor`.
  - The regression tests in `cmd/drive9/cli/ctx_test.go` now assert this invalid guidance, so they lock in a broken user interaction.

Why this matters: this PR is explicitly CLI UX-sensitive. A provisioning warning that sends users to an unavailable command is worse than incomplete guidance: the next command fails with `unknown command "status"`, which makes the fork flow look broken.

Required fix:
- Replace the guidance with an actually supported next step, or add the documented status command if that is intended scope.
- Update the tests to assert the real supported guidance, not `drive9 status`.

What is fixed / acceptable:
- The command now prints `Status: provisioning` and no longer silently says only `Now using ctx ...`.
- The API smoke docs are aligned with the current provision response shape.
- The broad `not found` retry comment is now explicit that it is scoped to fork-context smoke eventual consistency.

Validation run locally:
- `go test ./cmd/drive9/cli -run 'TestCtxForkCreatesOwnerContextAndSwitchesWhenRequested|TestCtxForkGeneratesNameWhenOmitted' -count=1` — passes, but currently passes against the invalid `drive9 status` guidance.
- `git diff --check` — passes.

GitHub status at review time: `ci` success; `drive9-server-local e2e smoke` still in progress.

srstack:
Fixed in commit `36def0e`.

I removed the invalid `drive9 status` guidance and replaced it with an actually supported next step:
- retry a normal command such as `drive9 fs ls /`
- clarify that `fs` commands may fail until the fork tenant becomes active

Updated the `ctx fork` output tests accordingly.

Validation:
- `go test ./cmd/drive9/cli -run 'TestCtxForkCreatesOwnerContextAndSwitchesWhenRequested|TestCtxForkGeneratesNameWhenOmitted' -count=1`

qiffang (approved):

srstack:
Updated this in commit 2faf60d.

Change summary:
- removed `ctx fork --use`
- `drive9 ctx fork ...` now only creates and saves the new fork context locally
- it no longer switches the active context automatically
- switching is back to the existing explicit flow: `drive9 ctx use <name>`

This keeps fork aligned with the current context model: `ctx use` remains the single explicit context-switch operation, and fork behaves like create in terms of persisting the new context while leaving activation to the normal ctx workflow.

I also updated the CLI help/output and the fork tests to assert that the source context stays active after fork.

qiffang (dismissed):
REVISE — latest head `2faf60d`

The previous blocker is fixed: the CLI no longer tells users to run nonexistent `drive9 status`, and it now keeps fork context switching explicit via `drive9 ctx use <name>`.

Remaining blocker:

- **B1: top-level help still documents removed `ctx fork --use`.**
  - `cmd/drive9/cli/ctx.go` removed `--use` from `drive9 ctx fork` parsing and subcommand usage.
  - But `cmd/drive9/main.go` still prints: `ctx fork [<new>] [--from <ctx>] [--use] [--json]`.
  - That means `drive9 --help` advertises a flag that now returns `unknown flag "--use"`.

Why this matters: this PR is CLI UX-sensitive. The latest direction—do not auto-switch to a provisioning fork—is reasonable, but the command surface must be internally consistent. Help text is part of the API contract for CLI users.

Required fix:
- Remove `--use` from the top-level usage line in `cmd/drive9/main.go`.
- Ideally extend the command usage test to assert the `ctx fork` line does not include `--use`, so this does not regress.

Validation run locally:
- `go test ./cmd/drive9/cli -run 'TestCtxFork' -count=1` — passes.
- `go test ./cmd/drive9 -run 'TestUsage|TestDispatch|TestMain' -count=1` — passes, but does not currently assert the stale `ctx fork` help line.
- `git diff --check` — passes.

GitHub checks at review time: `ci` and `drive9-server-local e2e smoke` are green.

srstack:
Addressed the new review items in commit 8498492.

Fixed:
- top-level `drive9 --help` now matches `ctx fork` and no longer advertises `--use`
- fork-create response now returns the provisioning message once the branch has been created, by carrying the created branch metadata into the immediate response path
- CLI e2e fork probing no longer sends a real Authorization header or malformed JSON; it now does an unauthenticated empty POST and only treats `404` as unsupported

Already handled earlier in the branch:
- `cp` now uses `StatCtx(ctx, ...)` for remote target resolution
- local directory target detection accepts `/` suffixes as well

Validated with:
- `go test ./cmd/drive9 ./pkg/server -count=1`

mornyx (approved):
## PR #426 Review: Polish fork provisioning UX and extend CLI smoke coverage

**Head:** `8498492` | **Files:** 15 | **+416 / -55**

---

### Overview

Six independent changes, all solid. Every prior review comment has been addressed.

---

### Change-by-change analysis

**1. cp directory-target semantics** — `cmd/drive9/cli/cp.go`

`resolveRemoteCopyTarget` / `resolveLocalCopyTarget` match shell `cp` behavior:
- Trailing slash → fast-path join (no extra Stat) ✅
- Existing directory → Stat-then-join ✅
- Non-existent or file → returns original path ✅
- Covers stdin→remote, local→remote, remote→local, remote→remote ✅

`sourceBase` type selection is correct: `filepath.Base` for local paths, `path.Base` for remote paths. `resolveLocalCopyTarget` checks both `os.PathSeparator` and `"/"` — Windows-safe ✅.

**2. Fork DB password independence** — `pkg/server/fork.go`

PR #424 had forks reusing the source DB password. This PR introduces `generateForkDBPassword` to create an independent random password stored in `forkRoot.DBPasswordCipher`. `ensureForkBranchConnection` now uses `forkTenant.DBPasswordCipher` in both the existing-branch and CreateBranch paths — no more source password leakage ✅.

New test `TestCreateForkTenantPersistsGeneratedBranchPassword` verifies:
- Fork password ≠ source password ✅
- `CreateBranch` password input matches persisted fork password ✅

`generateForkDBPassword` uses `crypto/rand` + `math/big` (uniform distribution), unlike the source-password `s.pool.Decrypt` path in PR #424 — no timing side-channel concern ✅.

**3. Fork provisioning message simplification** — `pkg/server/fork.go`

```go
func forkProvisioningMessage(t *meta.Tenant) string {
    if t == nil || t.BranchID == "" { return "" }
    return "Migrating fork data. Large tenants may take a few minutes."
}
```

Three-stage messages collapsed into one. `createForkTenant` sets `forkRoot.BranchID` before calling `forkProvisioningMessage`, so the 202 response always includes the message ✅. `/v1/status` returns empty message when no BranchID (backward-compatible with legacy fork state) ✅.

**4. Removed `--use` from `ctx fork`** — `cmd/drive9/cli/ctx.go`, `cmd/drive9/main.go`

Fork no longer auto-switches. Output now guides users to manually `ctx use` after provisioning completes. Correct for async provisioning — switching to a not-yet-ready context would cause confusing failures ✅. Help text in both `ctx.go:55` and `main.go:283` is synchronized ✅.

**5. `tenant_id` in provision response** — `pkg/server/server.go`

`/v1/provision` and local shim now return `tenant_id`. api-smoke-test and all unit tests updated ✅.

**6. E2E fork smoke** — `e2e/cli-smoke-test.sh`

- `fork_checks_enabled()` probes `/v1/fork` with empty POST: server without auth → 404 → skip; auth-enabled → 401 → proceed with proper auth ✅
- Full fork lifecycle: create → fork → poll status → read/write → delete ✅
- Directory-target cp checks for all three directions ✅
- `drive9_ctx_retry` with bounded retry for 429 and "not found" ✅
- Cleanup removes created directories and files ✅

---

### 🟡 Suggestions (non-blocking)

**1. `TestCpRemoteToRemoteDirectoryKeepsSourceName` — empty 200 body**

The handler writes `w.WriteHeader(http.StatusOK)` with no JSON body. If `c.Copy()` internally decodes the response, it will hit `unexpected end of JSON input`. If CI passes, the client tolerates it — but consider `_ = json.NewEncoder(w).Encode(map[string]string{"status": "ok"})` to match real server behavior.

**2. Tests hardcode exact message strings**

```go
// fork_test.go — two locations
if resp.Message != "Migrating fork data. Large tenants may take a few minutes." {
```

A future wording change will break these. Consider:
```go
if !strings.Contains(resp.Message, "Migrating fork data") {
```

---

### ✅ Resolved prior review comments

| Reviewer feedback | Status |
|---|---|
| `c.Stat` → `c.StatCtx` in `resolveRemoteCopyTarget` | ✅ Fixed |
| Missing `Content-Type: application/json` on provision | ✅ Fixed |
| `main.go` still documents removed `--use` | ✅ Fixed |
| `resolveLocalCopyTarget` checks only `os.PathSeparator` | ✅ Added `"/"` check |
| `fork_checks_enabled` sends `--data '{'` | ✅ Changed to empty POST |
| `forkProvisioningMessage` three-stage inaccuracy | ✅ Simplified |

---

### Verdict

All prior feedback addressed. No blockers. Two 🟡 suggestions are optional polish. Ready to merge.

mornyx (approved):

PR #427, opened 2026-05-14
Add reusable Drive9 agent harness

mashenjun:
## Summary

1. Add a reusable Drive9 agent harness under `e2e/agent-harness` with `preflight`, `run`, `report`, `gc`, and `collect-server-evidence` commands.
2. Add smoke, regression, stress, and fault suites with structured YAML case definitions, safety validation, mount lifecycle handling, and known-bug/non-gating semantics.
3. Add structured artifact writers and schemas for manifest, events, failures, metrics, summary, and gating outputs.
4. Move the harness design proposal to `docs/design/drive9-agent-harness-v2-plan.md`.

## Validation

1. `go test ./e2e/agent-harness/...`
2. `go build -o /tmp/drive9-agent-harness-check ./e2e/agent-harness/cmd/drive9-agent-harness`

## Summary by CodeRabbit

* **New Features**
  * New Drive9 agent harness CLI with commands for preflight, run, report, GC, and server-evidence collection; end-to-end runner for smoke, regression, stress, and fault suites
  * Structured run artifacts and report regeneration with gating determination

* **Documentation**
  * Full design document, README, JSON Schema contracts, and runnable repro scripts

* **Tests**
  * Added comprehensive tests including case parsing/validation, safety checks, mount behavior, recorder concurrency, and report-generation scenarios

[Review Change Stack](https://app.coderabbit.ai/change-stack/mem9-ai/drive9/pull/427?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


qiffang (changes requested):
## PR #427 — Add reusable Drive9 agent harness: REVISE

Reviewed: design doc, 4 suite YAMLs, 6 JSON schemas, cmd/main.go, casefile, mountproc, report, runner (1582 LoC), safety, repro scripts, README, go.mod. Head: dc0f1ace.

### Design and Structure

The overall architecture is sound: typed YAML cases → runner with structured JSONL/JSON artifacts → machine-readable gating → regenerable reports. The design doc is thorough. Safety model is well-considered. The phase-1 / deferred boundary is clear.

### Blockers

**B1. `ValidateRoot` does not reject remote-root-base paths (non-filesystem paths)**

\`safety.ValidateRoot("remote-root-base", ...)\` checks \`strings.HasPrefix(p, "/")\` — this accidentally works for remote roots like \`/agent-adversarial-xxx\`, but the function name and contract say "validate root" meaning a local filesystem root. If someone passes a relative remote root or one without a leading slash, \`ValidateRoot\` returns the right error, but the error message says "must be absolute" — which is a filesystem concept, not a remote-path concept. The function should either be split into \`ValidateLocalRoot\` and \`ValidateRemoteRootBase\`, or the error message should be parameterized. Currently the coupling is accidental — a future caller will misread the contract.

**B2. `runID()` is not collision-safe for concurrent or rapid sequential runs**

\`runID()\` returns \`time.Now().UTC().Format("20060102T150405Z")\` — second granularity. Two runs within the same second get the same run ID, same artifact directory, same remote root base. This will silently corrupt artifacts. Add a random suffix (e.g. 4 hex chars from crypto/rand) or use a monotonic counter.

**B3. `shouldRemoveMountpoint` reads `retErr` via closure but `retErr` is always nil at call site**

In \`runCase\`, the defer closure calls \`shouldRemoveMountpoint(c, rec, retErr)\`, but \`retErr\` is the named return value. At the point the defer captures \`retErr\`, it's still nil because the defer evaluates it lazily. However, the **doctor_fuse** path calls \`shouldRemoveMountpoint(c, rec, retErr)\` directly at line ~2778 where \`retErr\` is indeed nil (the function hasn't returned yet). This means doctor_fuse cleanup always thinks there was no error, so \`retain_on_failure\` is broken for doctor_fuse cases — mountpoints are always removed even on failure.

**B4. `path_matrix` path validation rejects the design doc's own test path**

The regression case \`path-edge-strict\` defines path \`"space name unicode-测试 punct_!@%.txt"\`. But \`ValidateRelativeSlashPath\` calls \`path.Clean(p)\` and checks \`clean != p\`. \`path.Clean("space name unicode-测试 punct_!@%.txt")\` returns the same string, so this specific path passes. **However**, paths containing consecutive spaces, trailing spaces, or trailing dots would be silently cleaned by \`path.Clean\`, causing the \`clean != p\` check to reject them — but these are exactly the edge cases the path matrix workload is supposed to test. The validation is over-restrictive for a path-edge regression suite whose entire purpose is testing unusual paths. Consider relaxing validation for \`path_matrix\` workloads specifically, or documenting why the restriction exists.

**B5. `provision()` sends POST with nil body — server may reject**

\`http.NewRequestWithContext(ctx, http.MethodPost, .../v1/provision, nil)\` sends a POST with no Content-Type and no body. Depending on the server implementation, this may be rejected. At minimum, send an empty JSON object body with \`Content-Type: application/json\`.

**B6. `Generate()` double-counts cases that appear in both `manifest.Cases` and `manifest.SelectedCases`**

\`Generate\` initializes \`selected\` from \`manifest.Cases\`, then iterates \`manifest.SelectedCases\` and adds entries that aren't already present. But in \`Run()\`, the manifest is written with both \`Cases: caseSummaries(cases)\` and \`SelectedCases: caseIDs(cases)\` — the same case IDs. When \`Generate\` then iterates \`selected\` to count pass/fail, it iterates the merged map, which is correct. **But**: if a case has failures, the code updates \`selected[f.CaseID]\` with \`ExpectedOutcome\` from the failure record, overwriting the suite field from the original \`CaseSummary\`. A case that fails with one oracle but passes another will get its \`Suite\` field zeroed out (because \`Failure\` doesn't carry \`Suite\`), causing \`summary.Counts.BySuite[""]\` to increment. The suite field must be preserved from the original manifest, not overwritten by failure iteration.

### Non-blocking observations

1. **Duplicate `readJSON`/`writeJSON`** — defined in both `report/report.go` and `runner/runner.go`. Factor into a shared `internal/jsonutil` package.

2. **`IsMounted` on macOS** — parses `mount` command output with `strings.Contains(string(out), " on "+clean+" ")`. This is fragile — mount paths with spaces or parentheses in the mount options will false-match. Not a blocker for phase-1 local testing, but worth a TODO.

3. **`open_fd_unmount` first unmount timing** — the test holds a file open for `hold_open_duration` (1s), then attempts unmount. But the write (`f.WriteString("held open by harness\n")`) happens before the hold delay. In write-back mode, the data may already be flushed by the time the unmount runs, making the "busy" path unreliable. Consider writing continuously during the hold period.

4. **`kill_during_write` goroutine leak on context cancel** — if `ctx` is cancelled before the timer fires and before the writer goroutine finishes, the goroutine continues writing to the (now possibly unmounted) FUSE path indefinitely. The writer should also select on `ctx.Done()` — which it does via `writeSizedFile`, but `writeSizedFile` only checks ctx between 1MB blocks. A 67MB write with 1MB blocks means up to 1MB of additional writes after cancellation.

5. **Missing `Workload.Remount` default** — `remountPaths` checks `c.Workload.Remount != nil && !*c.Workload.Remount` to skip remount. But `Remount` is a `*bool` — when omitted from YAML, it's nil, so remount is always attempted. The design doc says "Optional `remount: true|false`, defaulting to `true`" — this matches the nil-means-true behavior, but it's implicit. Document the nil semantics or set an explicit default in `applyDefaults`.

6. **`path_matrix` content derivation** — content defaults to `c.ID + ":" + p`. For the path-edge case, this generates content like `path-edge-strict:space name unicode-测试 punct_!@%.txt`. The content itself contains the problematic characters — which is fine for testing, but the SHA-256 oracle derivation depends on this exact template. If someone adds `content_template` to a case without updating oracles, the derived oracle silently uses different content. Worth a comment.

7. **No test for `FilterCases` with unknown case IDs** — `FilterCases` silently drops unknown IDs. If someone typos `--case path-edge-strct`, they get zero cases and the "no selected cases" error doesn't tell them which ID was wrong.

### Verdict: REVISE

Fix B1–B6 before merge. B2 (runID collision) and B3 (retain_on_failure broken for doctor_fuse) are correctness bugs. B6 (suite field zeroed) corrupts summary reports.

e2e/agent-harness/internal/mountproc/mountproc.go
```diff
@@ -0,0 +1,133 @@
+// Package mountproc runs harness commands and tracks mount processes.
+package mountproc
+
+import (
+	"bytes"
+	"context"
+	"errors"
```
qiffang, line 7: **`--sync-mode` is deprecated after PR #431 merge** — PR #431 replaced `--sync-mode` with `--durability`. This harness uses `--sync-mode` in `StartMount()` (line ~2006 in diff):
```go
args = append(args, "--sync-mode", syncMode)
```
This will break on current main. Need to map the case `sync_mode` values (`strict`, `interactive`, `auto`) to the new `--durability` flag values (`auto`, `interactive`, `fsync`, `close-sync`, `write-sync`), or keep using `--sync-mode` if backward compat is maintained. Verify against current CLI.

e2e/agent-harness/internal/runner/runner.go
```diff
@@ -0,0 +1,1582 @@
...
+		if err := os.MkdirAll(mountpoint, 0o755); err != nil {
+			return err
+		}
+		if err := runDoctor(caseCtx, rec, env, drive9Bin, mountpoint, c); err != nil {
+			return err
+		}
+		_ = rec.Event(report.Event{CaseID: c.ID, Type: "case_end"})
+		if shouldRemoveMountpoint(c, rec, retErr) {
+			_ = os.Remove(mountpoint)
+		}
```
qiffang, line 278: **`shouldRemoveMountpoint` captures stale `retErr` for `doctor_fuse`**

In the `doctor_fuse` branch (line ~2778 in diff):
```go
if shouldRemoveMountpoint(c, rec, retErr) {
    _ = os.Remove(mountpoint)
}
```

`retErr` is the named return value, but at this point it's always `nil` because no error has been assigned yet — the function returns `nil` explicitly on line 2781. This means `retain_on_failure` cases will always clean up the mountpoint for `doctor_fuse`, even if the case recorded oracle failures.

The intent is probably `shouldRemoveMountpoint(c, rec, nil)` here (which is what happens), but the pattern is misleading. The `caseHadFailures` fallback in `shouldRemoveMountpoint` does check `failures.jsonl`, so this actually works correctly — but the `retErr` parameter is dead code for this branch. Consider documenting this or simplifying.

e2e/agent-harness/internal/runner/runner.go
```diff
@@ -0,0 +1,1582 @@
...
+	return nil
+}
+
+func runCmd(ctx context.Context, rec *report.Recorder, caseID, id string, env mountproc.Env, dir, name string, args ...string) mountproc.Result {
+	_ = rec.Event(report.Event{CaseID: caseID, Type: "command_start", CommandID: id})
+	result := mountproc.Run(ctx, id, env, dir, name, args...)
+	recordResult(rec, caseID, result)
+	return result
+}
+
```
qiffang, line 978: **`appendJSONL` is not safe for concurrent use** — `runParallelWrites` spawns goroutines that call `recordOracleFailure` → `rec.Failure()` → `appendJSONL`. Similarly, multiple concurrent oracle checks could call `rec.Event()`. The `appendJSONL` function opens the file with `O_APPEND` which provides atomic appends on most POSIX systems for writes < PIPE_BUF, but:

1. The JSON marshal + write is not atomic — two goroutines could interleave partial lines
2. On some filesystems, `O_APPEND` atomicity isn't guaranteed for writes > 4KB

Add a mutex to `Recorder` or use a channel-based writer for thread safety.

e2e/agent-harness/internal/runner/runner.go
```diff
@@ -0,0 +1,1582 @@
...
+		*mount = nil
+		return nil
+	}
+	if err := stopCaseMount(ctx, rec, env, drive9Bin, c.ID, mount); err != nil {
+		recordOracleFailure(rec, c, "unmount_busy_then_clean", err.Error(), "clean unmount after fd close")
+	}
+	return nil
+}
+
+func runKillDuringWrite(ctx context.Context, rec *report.Recorder, manifest *report.Manifest, env mountproc.Env, drive9Bin, caseRemote, mountpoint string, c casefile.Case, mount **mountproc.Mount) error {
```
qiffang, line 902: **Race in `runKillDuringWrite`** — After the kill timer fires and mount is killed, the second `select` block:
```go
select {
case writerErr = <-writerDone:
default:
    writerErr = errors.New("writer interrupted before acknowledged completion")
}
```

The `default` branch fires immediately if the writer goroutine hasn't returned yet. But the writer might complete (with an error from the killed mount) between the kill and this select. This is a race — you could miss the writer's actual error. Consider using a short timeout or always draining the channel:
```go
select {
case writerErr = <-writerDone:
case <-time.After(2 * time.Second):
    writerErr = errors.New("writer interrupted before acknowledged completion")
}
```

e2e/agent-harness/internal/safety/safety.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+func ValidateMountpointAvailable(mp string) error {
+	if mounted, err := IsMounted(mp); err == nil && mounted {
+		return fmt.Errorf("%w: %s is already mounted", ErrExistingMountpoint, mp)
+	}
+	info, err := os.Lstat(mp)
+	if os.IsNotExist(err) {
+		return nil
+	}
```
qiffang, line 75: **Symlink check is correct but misleading** — `os.Lstat` returns the file info *without* following symlinks, so `info.Mode()&os.ModeSymlink != 0` correctly detects symlinks. Good. However, the subsequent `info.IsDir()` check on line 4329 would pass for a symlink pointing to a directory if you used `os.Stat` instead — since you're using `Lstat`, `IsDir()` returns false for symlink-to-dir, which means the function correctly rejects symlinks before reaching the directory check.

This is fine, just noting for future readers that the order matters here.

e2e/agent-harness/internal/runner/runner.go
```diff
@@ -0,0 +1,1582 @@
...
+				recordResult(rec, c.ID, result)
+			}
+		}
+		return "file://" + filepath.ToSlash(repoPath), nil
+	default:
+		return "", fmt.Errorf("unknown git fixture %q", name)
+	}
+}
+
+func checkNoGitLocks(rec *report.Recorder, c casefile.Case, repoPath, stage string) {
```
qiffang, line 1188: **`readJSON` and `writeJSON` duplicated** — These helper functions are defined in both `report/report.go` and `runner/runner.go` with identical implementations. Extract to a shared `internal/util` package or export from `report`.

e2e/agent-harness/internal/report/report.go
```diff
@@ -0,0 +1,352 @@
...
+	if m.TS == "" {
+		m.TS = now()
+	}
+	return appendJSONL(filepath.Join(r.RunDir, "metrics.jsonl"), m)
+}
+
+func Generate(runDir string) (Summary, Gating, error) {
+	manifest := Manifest{}
+	if err := readJSON(filepath.Join(runDir, "manifest.json"), &manifest); err != nil {
+		return Summary{}, Gating{}, err
```
qiffang, line 185: **`Generate` assumes wall-clock duration but manifest `StartedAt` is set early** — The duration is calculated as `endedAt - startedAt`, but `startedAt` is set in `Run()` before any case execution, while `Generate` is called after all cases complete. If `Generate` is called via `Regenerate` (the `report` command) at a much later time, `endedAt` will be the regeneration time, not the actual run end time. The `run_end` event in `events.jsonl` has the real end timestamp — consider reading that instead for regeneration accuracy.

e2e/agent-harness/internal/runner/runner.go
```diff
@@ -0,0 +1,1582 @@
...
+		if err := waitLocalContent(ctx, c, local, want); err != nil {
+			recordOracleFailure(rec, c, "cli_read_equals", err.Error(), want)
+		}
+	}
+	_ = rec.Metric(report.Metric{CaseID: c.ID, Name: "file_count", Value: float64(count), Unit: "files", Source: c.Workload.Type})
+	return nil
+}
+
+func runParallelWrites(ctx context.Context, rec *report.Recorder, mountpoint string, c casefile.Case) error {
+	start := time.Now()
```
qiffang, line 686: **`i := i` shadow is unnecessary with Go 1.22+** — If the project uses Go 1.22+, the loop variable capture change makes `i := i` redundant. Check `go.mod` for the Go version — if ≥ 1.22, this can be removed for clarity. (Not a bug, just noise.)

e2e/agent-harness/internal/runner/runner.go
```diff
@@ -0,0 +1,1582 @@
...
+	if cfg.APIKey != "" {
+		if err := checkStatus(ctx, cfg.Server, cfg.APIKey); err != nil {
+			return err
+		}
+	}
+	return nil
+}
+
+func Run(ctx context.Context, cfg Config) (string, error) {
+	id := runID()
```
qiffang, line 119: **Manifest written after all cases, but cases mutate it during execution** — `manifest.GeneratedMountpoints`, `manifest.GeneratedRemoteRoots`, and `manifest.ProcessGroups` are populated during `runCase` iterations. If the runner crashes mid-run, the manifest is never written and all run metadata is lost. Consider writing a partial manifest before the case loop and updating it after each case, or at minimum writing it in a defer.

e2e/agent-harness/cases/regression.yaml
```diff
@@ -0,0 +1,133 @@
...
+cases:
+  - id: path-edge-strict
+    suite: regression
+    sync_mode: strict
+    expected_outcome: bug_reproduced
+    remote_root_suffix: path-edge-strict
+    mountpoint_suffix: path-edge-strict
+    workload:
+      type: path_matrix
+      paths:
```
qiffang, line 18: **`path_matrix` validation vs actual path** — The path `"space name unicode-测试 punct_!@%.txt"` will pass `ValidateRelativeSlashPath` since it has no `/`, `..`, or backslashes. But the `%` character is URL-significant and could cause issues with the drive9 CLI's `fs cp` / `fs cat` commands if they don't properly encode the remote path. The harness writes locally via FUSE (bypassing URL encoding), but reads remotely via CLI — this asymmetry is exactly what the test is designed to catch. Just confirming the test intent is correct and this isn't a harness bug.

e2e/agent-harness/internal/mountproc/mountproc.go
```diff
@@ -0,0 +1,133 @@
...
+	if err != nil {
+		return nil, err
+	}
+	args := []string{"mount", "--mode", "fuse", "--perf-counters"}
+	if syncMode != "" {
+		args = append(args, "--sync-mode", syncMode)
+	}
+	args = append(args, ":"+remoteRoot, mountpoint)
+	cmd := exec.CommandContext(ctx, drive9Bin, args...)
+	cmd.Env = withEnv(os.Environ(), env)
```
qiffang, line 84: **Mount log file leaked on process exit** — The goroutine:
```go
go func() {
    _ = cmd.Wait()
    _ = logFile.Close()
}()
```
If `StartMount` returns successfully but the caller never waits for the mount to exit (e.g., it calls `KillMount` which sends SIGTERM), the goroutine is orphaned until the process actually exits. This is fine in practice since kill will cause Wait to return, but if kill fails silently, the file descriptor leaks until process exit.

e2e/agent-harness/internal/runner/runner.go
```diff
@@ -0,0 +1,1582 @@
...
+			result := runCmd(ctx, rec, c.ID, "cli-write-"+f.RelativePath, env, "", drive9Bin, "fs", "cp", tmp, ":"+remote)
+			if result.ExitCode != 0 {
+				recordOracleFailure(rec, c, "fuse_write_success", result.Stderr, "exit 0")
+				continue
+			}
+			local, err := safeJoinLocal(mountpoint, f.RelativePath)
+			if err != nil {
+				return err
+			}
+			if err := waitLocalContent(ctx, c, local, f.Content); err != nil {
```
qiffang, line 463: **`provision()` uses `http.DefaultClient` without timeout** — `http.DefaultClient` has no timeout, so a hanging server will block forever. Set a client-level timeout or use the context deadline:
```go
client := &http.Client{Timeout: 30 * time.Second}
```

qiffang (changes requested):
## PR #427 Review — Add reusable Drive9 agent harness

Overall this is a solid, well-structured harness with good design principles (typed case files, structured artifacts, safety validation, oracle-based verification). The design doc is thorough and the contract-first approach is the right call.

I have several issues to flag, ranging from potential data corruption to API contract mismatches. See inline comments.

### Summary

**Architecture & Design** — Clean separation of concerns across `casefile`, `safety`, `mountproc`, `report`, and `runner`. The YAML case schema is well-typed and the oracle system is extensible.

**Key Issues:**
1. **`--sync-mode` flag is deprecated** — PR #431 replaced `--sync-mode` with `--durability`. The harness will break on latest main.
2. **`shouldRemoveMountpoint` captures stale `retErr`** — The defer closure reads `retErr` at return time but the variable is always `nil` at the point where it's evaluated for doctor_fuse cases.
3. **`Lstat` won't detect symlinks** — `os.Lstat` returns the link info, but `info.Mode()&os.ModeSymlink != 0` is always true for Lstat — the check logic is correct, but needs clarity.
4. **Race in `runKillDuringWrite`** — Writer goroutine and kill timer can race; after kill, the second `select` on `writerDone` uses `default` branch which may miss a completed writer.
5. **`appendJSONL` not safe for concurrent writers** — Multiple goroutines could call `rec.Event()` / `rec.Failure()` concurrently without synchronization.
6. **Missing `go.sum` update** — `gopkg.in/yaml.v3` promoted from indirect to direct but no go.sum diff shown.

**Nits:**
- `readJSON` and `writeJSON` are duplicated in both `report` and `runner` packages.
- Design doc is 691 lines — consider splitting deferred sections into a separate doc.

e2e/agent-harness/internal/runner/runner.go
```diff
@@ -0,0 +1,1582 @@
...
+			*mount = nil
+			_ = rec.Event(report.Event{CaseID: c.ID, Type: "mount_end", Message: "mount process group killed during active write"})
+		}
+	}
+	var writerErr error
+	select {
+	case writerErr = <-writerDone:
+	default:
+		writerErr = errors.New("writer interrupted before acknowledged completion")
+	}
```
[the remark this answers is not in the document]
mashenjun: Addressed in e6bb4d0. `runKillDuringWrite` now preserves the writer result when the first select receives `writerDone`, and only restarts the mount when the kill path actually ran.

e2e/agent-harness/internal/report/report.go
```diff
@@ -0,0 +1,352 @@
...
+	failures, err := readFailures(filepath.Join(runDir, "failures.jsonl"))
+	if err != nil {
+		return Summary{}, Gating{}, err
+	}
+	ended := now()
+	var durationMS int64
+	if manifest.StartedAt != "" {
+		if startedAt, err := time.Parse(time.RFC3339Nano, manifest.StartedAt); err == nil {
+			if endedAt, err := time.Parse(time.RFC3339Nano, ended); err == nil {
+				durationMS = endedAt.Sub(startedAt).Milliseconds()
```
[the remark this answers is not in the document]
mashenjun: Addressed in e6bb4d0. `Generate` now reads the persisted `run_end` event timestamp from `events.jsonl` and uses that for `ended_at` and `duration_ms` during report regeneration.

e2e/agent-harness/internal/runner/runner.go
```diff
@@ -0,0 +1,1582 @@
...
+}
+
+func runDoctor(ctx context.Context, rec *report.Recorder, env mountproc.Env, drive9Bin, mountpoint string, c casefile.Case) error {
+	result := runCmd(ctx, rec, c.ID, "doctor-fuse", env, "", drive9Bin, "doctor", "fuse", "--mountpoint", mountpoint, "--server", env.Server)
+	want := 0
+	if c.Workload.ExpectExit != nil {
+		want = *c.Workload.ExpectExit
+	}
+	if result.ExitCode != want {
+		recordOracleFailure(rec, c, "command_exit", result.ExitCode, want)
```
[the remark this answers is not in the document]
mashenjun: Addressed in e6bb4d0. `runDoctor` now honors `allow_nonzero_when_no_allow_other` when the only failing doctor check is `/etc/fuse.conf user_allow_other`, with targeted test coverage.

e2e/agent-harness/internal/runner/runner.go
```diff
@@ -0,0 +1,1582 @@
...
+	}
+	if _, err := safety.ValidateRoot("remote-root-base", cfg.RemoteRootBase); err != nil {
+		return err
+	}
+	if cfg.APIKey == "" && !cfg.Provision {
+		return errors.New("preflight: --api-key or --provision is required")
+	}
+	if cfg.APIKey != "" {
+		if err := checkStatus(ctx, cfg.Server, cfg.APIKey); err != nil {
+			return err
```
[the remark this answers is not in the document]
mashenjun: Addressed in e6bb4d0. `Preflight` now skips stale API-key status validation when `--provision` is set, and test coverage verifies that path.

e2e/agent-harness/internal/runner/runner.go
```diff
@@ -0,0 +1,1582 @@
...
+	if userName == "" {
+		userName = "Drive9 Harness"
+	}
+	userEmail := c.Workload.GitUserEmail
+	if userEmail == "" {
+		userEmail = "drive9-harness@example.invalid"
+	}
+	if result := mountproc.Run(ctx, "git-config-name", mountproc.Env{}, repoPath, git, "config", "user.name", userName); result.ExitCode != 0 {
+		recordResult(rec, c.ID, result)
+		recordOracleFailure(rec, c, "git_config", result.Stderr, "exit 0")
```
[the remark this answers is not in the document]
mashenjun: Addressed in e6bb4d0. The git workflow now uses the git-specific context for config, add, commit, status, and commit-count commands, not just clone/fixture setup.

e2e/agent-harness/cmd/drive9-agent-harness/main.go
```diff
@@ -0,0 +1,144 @@
...
+
+import (
+	"context"
+	"flag"
+	"fmt"
+	"os"
+	"strings"
+
+	"github.com/mem9-ai/dat9/e2e/agent-harness/internal/runner"
+)
```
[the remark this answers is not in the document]
mashenjun: Already satisfied in the current branch. The import block separates stdlib imports from the internal runner import with a blank line, so no code change was needed.

e2e/agent-harness/internal/runner/runner.go
mashenjun: Agree with the analysis. No behavior change needed here: the doctor branch relies on `caseHadFailures` inside `shouldRemoveMountpoint`, so `retain_on_failure` still retains when oracle failures were recorded. The `retErr` argument is redundant in this branch but not currently harmful.

e2e/agent-harness/internal/runner/runner.go
mashenjun: I do not think this is currently reachable from `runParallelWrites`: the goroutines only collect errors into a local slice under a mutex, and `recordOracleFailure` runs after `wg.Wait`. Cases/oracles are otherwise executed serially. If we later parallelize case execution or oracle recording, adding a `Recorder` mutex would be the right boundary.

e2e/agent-harness/internal/runner/runner.go
mashenjun: Valid. e6bb4d0 fixed the consumed-success path from the earlier review, but this remaining post-kill drain race is still worth tightening with a bounded wait for `writerDone`. Not addressed in the current commit.

e2e/agent-harness/internal/mountproc/mountproc.go
mashenjun: Valid rebase concern. This branch is still based on the pre-#431 CLI where `--sync-mode` exists; once rebased onto the PR #431 change, the harness should translate the case-level sync setting to the new `--durability` values. Not addressed in e6bb4d0.

e2e/agent-harness/internal/report/report.go
mashenjun: Addressed in e6bb4d0. `Generate` now reads the latest persisted `run_end` event from `events.jsonl` and uses it for regenerated `ended_at` and duration values.

e2e/agent-harness/internal/runner/runner.go
mashenjun: Valid resilience improvement. Not addressed in e6bb4d0; a follow-up should write an initial manifest before the loop and update it after each case, or defer a best-effort final manifest write.

e2e/agent-harness/cases/regression.yaml
mashenjun: Confirmed. This asymmetry is intentional: the case writes through FUSE and reads through the CLI specifically to catch remote path encoding/round-trip issues.

e2e/agent-harness/internal/mountproc/mountproc.go
mashenjun: Agree the risk is bounded by the mount process lifetime. The normal Stop/Kill paths should cause `cmd.Wait()` to return and close the log file; if kill itself fails, the mount process is still alive and the descriptor remains tied to that process lifecycle. No code change in this batch.

e2e/agent-harness/internal/runner/runner.go
mashenjun: Addressed in e6bb4d0. The runner now uses a dedicated `http.Client` with a 30s timeout for provision/status HTTP calls while preserving request contexts.

e2e/agent-harness/internal/safety/safety.go
mashenjun: Agree. No code change needed; `Lstat` is intentional here so symlink-to-dir paths are rejected before any directory availability check.

e2e/agent-harness/internal/runner/runner.go
mashenjun: Noted. I’m keeping these package-local for now because they are tiny helpers used on opposite sides of the artifact boundary. I would extract only if another package needs the same behavior.

e2e/agent-harness/internal/runner/runner.go
mashenjun: Agree this is redundant on the repository’s Go version. I left it unchanged because it is harmless and consistent with the explicit capture style in this harness code; it can be cleaned up separately.

e2e/agent-harness/schemas/manifest.schema.json
```diff
@@ -0,0 +1,136 @@
...
+      "type": "array",
+      "items": {
+        "type": "object",
+        "additionalProperties": false,
+        "required": [
+          "id",
+          "suite",
+          "expected_outcome",
+          "status"
+        ],
```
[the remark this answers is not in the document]
mashenjun: Valid. The manifest schema should not require `cases[].status`; runner-generated manifests currently set it, but phase-1 manifest fixtures and consumers should be allowed to omit it. Not addressed in e6bb4d0.

e2e/agent-harness/internal/mountproc/mountproc.go
```diff
@@ -0,0 +1,140 @@
...
+
+func StartMount(ctx context.Context, id string, env Env, drive9Bin, remoteRoot, mountpoint, syncMode, logPath string) (*Mount, error) {
+	if err := os.MkdirAll(mountpoint, 0o755); err != nil {
+		return nil, err
+	}
+	if err := os.MkdirAll(filepath.Dir(logPath), 0o755); err != nil {
+		return nil, err
+	}
+	logFile, err := os.OpenFile(logPath, os.O_CREATE|os.O_APPEND|os.O_WRONLY, 0o644)
+	if err != nil {
```
qiffang, line 72: **Still uses `--sync-mode` (deprecated by PR #431)** — Line 72 in the new file:
```go
args = append(args, "--sync-mode", syncMode)
```
PR #431 replaced `--sync-mode` + `--write-policy` with a single `--durability` flag. This will break on any build that includes #431. Need to either:
1. Map case `sync_mode` → `--durability` profiles, or
2. Keep `--sync-mode` if the CLI still accepts it as a hidden/deprecated alias

Verify against current main CLI.

e2e/agent-harness/internal/runner/runner.go
```diff
@@ -0,0 +1,1627 @@
...
+func checkStatus(ctx context.Context, server, apiKey string) error {
+	req, err := http.NewRequestWithContext(ctx, http.MethodGet, strings.TrimRight(server, "/")+"/v1/status", nil)
+	if err != nil {
+		return err
+	}
+	req.Header.Set("Authorization", "Bearer "+apiKey)
+	resp, err := harnessHTTPClient.Do(req)
+	if err != nil {
+		return err
+	}
```
qiffang, line 1607: **`runID()` still second-granularity** — Two harness invocations within the same second produce identical run IDs → same `drive9-agent-test-*` directory, same remote root base, silent data corruption. Add a random suffix, e.g.:
```go
func runID() string {
    b := make([]byte, 4)
    _, _ = rand.Read(b)
    return time.Now().UTC().Format("20060102T150405Z") + "-" + hex.EncodeToString(b)
}
```

e2e/agent-harness/internal/report/report.go
```diff
@@ -0,0 +1,386 @@
...
+	SchemaVersion          string    `json:"schema_version"`
+	RunID                  string    `json:"run_id"`
+	GateStatus             string    `json:"gate_status"`
+	Pass                   int       `json:"pass"`
+	Fail                   int       `json:"fail"`
+	KnownBugReproduced     int       `json:"known_bug_reproduced"`
+	KnownBugFixedCandidate int       `json:"known_bug_fixed_candidate"`
+	NonGating              int       `json:"non_gating"`
+	BlockingFailures       []Failure `json:"blocking_failures"`
+}
```
qiffang, line 121: **`appendJSONL` still not safe for concurrent use** — `Recorder` has no mutex. `runParallelWrites` spawns goroutines that can call `recordOracleFailure` → `rec.Failure()` → `appendJSONL` concurrently. While `O_APPEND` provides some atomicity on POSIX, the JSON marshal + `f.Write(append(b, '\n'))` sequence is not atomic — two goroutines can interleave mid-line.

Add a `sync.Mutex` to `Recorder` and lock around `appendJSONL` calls, or serialize writes through a channel.

qiffang (changes requested):
## PR #427 Round 2 Review

Good progress on the fixes. Several issues properly addressed, but **4 blockers remain unresolved**.

### Fixed ✅
1. **Race in `runKillDuringWrite`** — Properly fixed with `writerFinished`/`killed` flags and correct channel drain logic.
2. **`provision()` HTTP timeout** — Fixed with `harnessHTTPClient = &http.Client{Timeout: 30 * time.Second}`.
3. **`Generate()` regeneration duration** — Fixed by reading `run_end` event from `events.jsonl`. Good test added.
4. **Git workflow context** — Now consistently uses `gitCtx` instead of mixing parent/child contexts.
5. **Git mutation write errors** — Now properly checked instead of silently discarded.
6. **GC/safety error handling** — `IsMounted` errors properly surfaced; sentinel errors added.
7. **JSON schemas** — Significantly expanded with `additionalProperties: false` and full property definitions.
8. **Good new tests** — 5 new test functions covering the fixes.

### Still open ❌

**B1: `--sync-mode` deprecated (CRITICAL)** — `mountproc.go:StartMount` line 2005-2007 still passes `--sync-mode`. PR #431 merged `--durability` to main. The harness is broken on current main.

**B2: `appendJSONL` not thread-safe** — No mutex added to `Recorder`. `runParallelWrites` goroutines can still call `rec.Failure()` concurrently → interleaved JSONL lines.

**B3: `runID()` second-granularity collision** — Still `time.Now().UTC().Format("20060102T150405Z")`. Two runs in same second → same artifact dir, same remote root, silent corruption.

**B4: `ValidateRelativeSlashPath` over-restrictive for `path_matrix`** — Still uses `path.Clean(p) != p` check that rejects paths with consecutive spaces, trailing dots, etc. — exactly the edge cases `path_matrix` exists to test. The regression test path `"space name unicode-测试 punct_!@%.txt"` happens to survive `path.Clean`, but a path like `"double  space.txt"` would be rejected.

Please fix B1-B4 and this is ready to approve.

qiffang (changes requested):
## PR #427 — Round 2 Review: REVISE

Head: e6bb4d03 (was dc0f1ace). 29 files, +5398/-1.

### What's fixed

1. ✅ **dat9-dev1: `http.DefaultClient` timeout** — `harnessHTTPClient` with 30s timeout replaces `http.DefaultClient`. Good.
2. ✅ **dat9-dev1: `Generate()` duration on regeneration** — `readRunEnd` reads last `run_end` event from `events.jsonl`. Correct.
3. ✅ **dat9-dev1: `runKillDuringWrite` race** — improved with `killed`/`writerFinished` tracking. The `default` branch after kill is acceptable for fault-suite semantics (inconclusive classification).

### Withdrawn (my error in Round 1)

4. ~~B3 (`shouldRemoveMountpoint` broken for `doctor_fuse`)~~ — Withdrawn. `caseHadFailures(rec, c.ID)` reads `failures.jsonl` as a backup, so even with `retErr == nil`, recorded failures prevent mountpoint removal under `retain_on_failure`. Logic is correct.
5. ~~B6 (`Generate()` zeroes Suite field)~~ — Withdrawn. Go struct value copy preserves all fields including `Suite`. The failure iteration only overwrites `ID`, `ExpectedOutcome`, `Status` — `Suite` is preserved from the original `manifest.Cases` entry.

### Still open — blockers

**B1. `ValidateRoot` semantic mismatch — NOT FIXED**

Still a single function for both local filesystem roots and remote-path roots with filesystem-centric error messages. Low severity but worth splitting to avoid future misuse.

**B2. `runID()` collision — NOT FIXED**

Still `time.Now().UTC().Format("20060102T150405Z")` — second granularity. Two runs within the same second get the same artifact directory. Add random suffix.

**B4. `ValidateRelativeSlashPath` over-restrictive — NOT FIXED**

`path.Clean(p) != p` still rejects paths with consecutive spaces, trailing dots, etc. — exactly the edge cases `path_matrix` exists to test.

**B5. `provision()` nil body — NOT FIXED**

Still `http.NewRequestWithContext(ctx, POST, ..., nil)`. No Content-Type, no body.

**B7. `--sync-mode` deprecated — NOT FIXED (dat9-dev1 catch)**

`StartMount()` at `mountproc.go` still passes `--sync-mode`. PR #431 merged `--durability`. The harness is broken on current main.

**B8. `appendJSONL` not thread-safe — NOT FIXED (dat9-dev1 catch)**

`Recorder` has no mutex. `runParallelWrites` goroutines can call `recordOracleFailure` → `rec.Failure()` → `appendJSONL` concurrently. The local `sync.Mutex` in `runParallelWrites` only protects the local `errs` slice, not the Recorder's file writes.

**B9. Manifest written only after all cases — NOT FIXED (dat9-dev1 catch)**

`WriteManifest` at line 2867 runs only after the case loop. Crash mid-run loses all metadata including generated mountpoints and remote roots, so `gc` can't clean up.

### Verdict: REVISE

5 blockers remain from Round 1 (B1, B2, B4, B5, B7) and 2 from dat9-dev1 (B8, B9). B2 and B7 are the most critical — B2 is a silent data corruption, B7 means the harness is already broken on main.

e2e/agent-harness/internal/mountproc/mountproc.go
mashenjun: Addressed in 53692d5. `StartMount` now detects whether the target `drive9 mount --help` exposes `--durability` or legacy `--sync-mode`; for current `main` it maps `sync_mode` values to durability (`strict` -> `fsync`, `interactive` -> `interactive`, `auto` -> `auto`).

e2e/agent-harness/internal/runner/runner.go
mashenjun: Addressed in 53692d5. After the kill path fires, `runKillDuringWrite` now uses a bounded wait for `writerDone` instead of an immediate default branch, while still respecting context cancellation.

e2e/agent-harness/internal/runner/runner.go
mashenjun: Addressed in 53692d5. The runner now writes an initial manifest before the case loop and refreshes it after each case, so generated remote roots, mountpoints, and process metadata are persisted incrementally.

e2e/agent-harness/schemas/manifest.schema.json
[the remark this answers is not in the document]
mashenjun: Addressed in 53692d5. `cases[].status` is no longer required in `manifest.schema.json`, but remains declared in `properties` so it is still validated when present.

e2e/agent-harness/internal/runner/runner.go
mashenjun: Addressed in fd39e97. `runID()` now appends a random 4-byte hex suffix to the compact UTC timestamp, so same-second harness invocations do not collide on run directory or remote root names. Added focused coverage in `TestRunIDIncludesRandomSuffix`.

e2e/agent-harness/internal/report/report.go
mashenjun: Addressed in fd39e97. `Recorder` now has a `sync.Mutex` and serializes manifest, event, failure, and metric artifact writes. Added `TestRecorderConcurrentWrites` to exercise concurrent `Event`, `Failure`, and `Metric` writes and verify the JSONL artifacts remain parseable.

e2e/agent-harness/internal/runner/runner.go
mashenjun: Follow-up addressed in fd39e97. Even though current `runParallelWrites` records failures after `wg.Wait`, `Recorder` now serializes artifact writes with a `sync.Mutex`, covering future concurrent `Event`, `Failure`, and `Metric` calls as well.

e2e/agent-harness/internal/mountproc/mountproc.go
mashenjun: Already addressed in 53692d5 and still present after fd39e97. `StartMount` no longer appends `--sync-mode` directly; it calls `mountSyncArgs`, detects whether `drive9 mount --help` exposes `--durability` or legacy `--sync-mode`, and maps case `sync_mode` values for current main (`strict` -> `fsync`, `interactive` -> `interactive`, `auto` -> `auto`).

e2e/agent-harness/internal/runner/runner.go
```diff
@@ -0,0 +1,1644 @@
...
+		}
+		if err := rec.WriteManifest(manifest); err != nil {
+			return runDir, err
+		}
+	}
+	_ = captureDebug(ctx, rec, "run-end")
+	_ = rec.Event(report.Event{Type: "run_end"})
+	if err := rec.WriteManifest(manifest); err != nil {
+		return "", err
+	}
```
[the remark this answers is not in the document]
mashenjun: Addressed in 26f6c01. The final `rec.WriteManifest(manifest)` failure path now returns `runDir, err`, matching the earlier manifest write error paths and preserving artifact discovery for debugging.

e2e/agent-harness/internal/runner/runner.go
```diff
@@ -0,0 +1,1644 @@
...
+	local := filepath.Join(mountpoint, "held-open.txt")
+	if err := os.MkdirAll(filepath.Dir(local), 0o755); err != nil {
+		return err
+	}
+	f, err := os.OpenFile(local, os.O_CREATE|os.O_RDWR|os.O_TRUNC, 0o644)
+	if err != nil {
+		recordOracleFailure(rec, c, "unmount_busy_then_clean", err.Error(), "open file")
+		return nil
+	}
+	_, _ = f.WriteString("held open by harness\n")
```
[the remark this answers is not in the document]
mashenjun: Addressed in 26f6c01. The held-open file write now checks `WriteString`; on failure it closes the descriptor and records an `unmount_busy_then_clean` oracle failure instead of silently continuing.

mashenjun:
@qiffang follow-up on https://github.com/mem9-ai/drive9/pull/427#pullrequestreview-4308030518:

1. Addressed since this review:
   - B2 `runID()` collision: fixed in fd39e97 with a random 4-byte hex suffix and `TestRunIDIncludesRandomSuffix`.
   - B7 `--sync-mode` deprecation: fixed in 53692d5. `StartMount` now detects `--durability` vs legacy `--sync-mode` and maps `strict` -> `fsync`, `interactive` -> `interactive`, `auto` -> `auto`.
   - B8 recorder JSONL concurrency: fixed in fd39e97 with a `sync.Mutex` on `Recorder` and concurrent write coverage.
   - B9 manifest persistence: fixed in 53692d5 by writing the initial manifest before the case loop and refreshing it after each case.

2. Withdrawn in later review context:
   - doctor_fuse cleanup / `retErr` concern.
   - summary suite-field overwrite concern.

3. I do not think B4 is valid as stated: `path.Clean` does not reject consecutive spaces or trailing dots in a filename; the current validator rejects absolute paths, backslashes, empty segments, `.`/`..`, duplicate slash normalization, and trailing slash normalization, which is the intended safety boundary for slash-relative paths.

B1 (`ValidateRoot` naming/semantic split) and B5 (`provision` POST body) are separate cleanup items and are not claimed as fixed by the commits above.

qiffang (changes requested):
Blocking issue on the current PR merge ref:

1. `pkg/fuse/dat9fs.go:1118` does not compile. The merge result calls `fs.perf.readRetrySuccess.add(1)`, but `atomic.Uint64` exposes `Add`, not `add`:

```text
pkg/fuse/dat9fs.go:1118:29: fs.perf.readRetrySuccess.add undefined (type "sync/atomic".Uint64 has no field or method add, but does have method Add)
```

This is not just a local nit: GitHub `ci` is failing at the same compile/lint gate, and this is the merge ref GitHub will test before merge. The branch needs to be rebased/merged with current main and the conflict result fixed to `Add(1)`.

2. Required checks are still red. `drive9-server-local e2e smoke` also failed (`CLI smoke did not succeed`, `FUSE release gate did not succeed`). This may be downstream of the compile failure, but it still needs to be green before merging.

Local validation:

```text
PASS: /usr/local/go/bin/go test ./e2e/agent-harness/... -count=1
FAIL: /usr/local/go/bin/go test ./pkg/fuse -run TestNonExistent -count=1
```

I did not find a harness-specific correctness blocker in the files I reviewed, but the PR cannot be approved while the merge ref does not compile and required checks are red.

PR #428, opened 2026-05-14
perf: improve FUSE reads and delete paths

srstack:
## Summary
- raise and expose FUSE read-cache single-file admission limit to cover 512KB/1MB hot-read workloads
- reuse resolved S3 read targets across FUSE range reads and prefetches
- batch datastore recursive delete orphan handling and tune client/S3 connection pools

## Tests
- `go test ./pkg/client ./pkg/s3client ./pkg/datastore ./pkg/backend ./pkg/fuse ./cmd/drive9/cli`
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * New mount flag to cap per-file read-cache size (default 1 MB).
  * Read-cache now configurable per-file; prefetch and read paths use this limit.
  * Client exposes read-target/resolution and helpers to detect expired presigned URLs for robust large-file downloads and automatic refresh on failure.

* **Improvements**
  * Improved HTTP connection pooling for concurrent transfers.
  * Read-target memoization and cache invalidation tightened to avoid stale targets.

* **Tests**
  * Added/updated tests for mount flag, read-target handling, presign expiry refresh, cache limits, and deletion flows.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mem9-ai/drive9/pull/428)

### DIFF
--- a/cmd/drive9/cli/fuse_bridge.go
+++ b/cmd/drive9/cli/fuse_bridge.go
@@ -18,6 +18,7 @@ type mountFuseOptions struct {
 	RemoteRoot            string
 	CacheDir              string
 	CacheSize             int64
+	ReadCacheMaxFileBytes int64
 	DirTTL                time.Duration
 	AttrTTL               time.Duration
 	EntryTTL              time.Duration
--- a/cmd/drive9/cli/fuse_bridge_supported.go
+++ b/cmd/drive9/cli/fuse_bridge_supported.go
@@ -30,6 +30,7 @@ func mountFuseImpl(opts *mountFuseOptions) error {
 		RemoteRoot:            opts.RemoteRoot,
 		CacheDir:              opts.CacheDir,
 		CacheSize:             opts.CacheSize,
+		ReadCacheMaxFileBytes: opts.ReadCacheMaxFileBytes,
 		DirTTL:                opts.DirTTL,
 		AttrTTL:               opts.AttrTTL,
 		EntryTTL:              opts.EntryTTL,
--- a/cmd/drive9/cli/mount.go
+++ b/cmd/drive9/cli/mount.go
@@ -84,6 +84,7 @@ func fsMountCmd(args []string) error {
 	mode := fs.String("mode", "auto", "mount mode: auto, fuse, or webdav")
 	cacheDir := fs.String("cache-dir", "", "write-back cache directory (default ~/.cache/drive9)")
 	cacheSize := fs.Int("cache-size", 128, "read cache size in MB")
+	readCacheMaxFile := fs.Int64("read-cache-max-file-mb", 1, "maximum single file size admitted to read cache in MB")
 	dirTTL := fs.Duration("dir-ttl", 10*time.Second, "directory cache TTL")
 	attrTTL := fs.Duration("attr-ttl", 10*time.Second, "kernel attr cache TTL")
 	entryTTL := fs.Duration("entry-ttl", 10*time.Second, "kernel entry cache TTL")
@@ -151,6 +152,9 @@ func fsMountCmd(args []string) error {
 	}
 	if err := validateReadDirPrefetchFlags(*prefetchMaxFiles, *prefetchMaxFileBytes, *prefetchMaxBytes, *prefetchTimeout); err != nil {
 		return err
+	}
+	if *readCacheMaxFile <= 0 {
+		return fmt.Errorf("drive9 mount: --read-cache-max-file-mb must be > 0")
 	}
 	normalizedLookupRetryCount := lookupRetryCountFlagValue(lookupRetryCountGiven, *lookupRetryCount)
 	normalizedLookupRetryTimeout := durationFlagValue(fs, "lookup-retry-timeout", *lookupRetryTimeout)
@@ -223,6 +227,7 @@ func fsMountCmd(args []string) error {
 		RemoteRoot:            remoteRoot,
 		CacheDir:              *cacheDir,
 		CacheSize:             int64(*cacheSize) << 20,
+		ReadCacheMaxFileBytes: *readCacheMaxFile << 20,
 		DirTTL:                normalizedDirTTL,
 		AttrTTL:               normalizedAttrTTL,
 		EntryTTL:              normalizedEntryTTL,
--- a/cmd/drive9/cli/mount_test.go
+++ b/cmd/drive9/cli/mount_test.go
@@ -721,6 +721,35 @@ func TestMountCmdLeavesLegacyDirStatFallbackDisabledByDefault(t *testing.T) {
 	}
 }
 
+func TestMountCmdPassesReadCacheMaxFileOption(t *testing.T) {
+	oldMountFuse := mountFuse
+	t.Cleanup(func() { mountFuse = oldMountFuse })
+
+	var got *mountFuseOptions
+	mountFuse = func(opts *mountFuseOptions) error {
+		copied := *opts
+		got = &copied
+		return nil
+	}
+
+	err := MountCmd([]string{
+		"--mode", "fuse",
+		"--server", "https://drive9.example",
+		"--api-key", "sk-test",
+		"--read-cache-max-file-mb", "4",
+		t.TempDir(),
+	})
+	if err != nil {
+		t.Fatalf("MountCmd: %v", err)
+	}
+	if got == nil {
+		t.Fatal("mountFuse was not called")
+	}
+	if got.ReadCacheMaxFileBytes != 4<<20 {
+		t.Fatalf("ReadCacheMaxFileBytes = %d, want %d", got.ReadCacheMaxFileBytes, int64(4<<20))
+	}
+}
+
 func TestMountCmdLeavesDefaultTTLsUnsetForFuseDefaults(t *testing.T) {
 	oldMountFuse := mountFuse
 	t.Cleanup(func() { mountFuse = oldMountFuse })
--- a/pkg/backend/dat9.go
+++ b/pkg/backend/dat9.go
@@ -1120,10 +1120,13 @@ func (b *Dat9Backend) CopyFileCtx(ctx context.Context, srcPath, dstPath string) 
 		return err
 	}
 	return b.store.InTx(ctx, func(tx *sql.Tx) error {
-		return b.store.InsertNodeTx(tx, &datastore.FileNode{
+		if err := b.store.InsertNodeTx(tx, &datastore.FileNode{
 			NodeID: b.genID(), Path: dstPath, ParentPath: pathutil.ParentPath(dstPath),
 			Name: pathutil.BaseName(dstPath), FileID: srcNode.FileID, CreatedAt: time.Now(),
-		})
+		}); err != nil {
+			return fmt.Errorf("insert copied node %q: %w", dstPath, err)
+		}
+		return nil
 	})
 }
 
--- a/pkg/backend/dat9.go
+++ b/pkg/backend/dat9.go
@@ -1119,9 +1119,11 @@ func (b *Dat9Backend) CopyFileCtx(ctx context.Context, srcPath, dstPath string) 
 	if err := b.store.EnsureParentDirs(ctx, dstPath, b.genID); err != nil {
 		return err
 	}
-	return b.store.InsertNode(ctx, &datastore.FileNode{
-		NodeID: b.genID(), Path: dstPath, ParentPath: pathutil.ParentPath(dstPath),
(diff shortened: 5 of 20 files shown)


mornyx (commented):
## Review Summary

**Overall**: The direction is sound and code quality is solid. All tests pass (`pkg/client`, `pkg/fuse`, `pkg/backend`, `pkg/datastore`, `pkg/s3client`, `cmd/drive9/cli`). A few items worth addressing before or shortly after merge.

---

### Important

#### 1. `scanOrphanedFilesByIDTx` silently skips fileIDs with missing inode/contents rows

In the old code, `DeleteDirRecursive` iterated every fileID from the DISTINCT query and called `scanFileForGCTx` individually. If a fileID existed in `file_nodes` but had no corresponding row in `inodes`/`contents` (data inconsistency edge case), `scanFileForGCTx` would return `ErrNotFound`, aborting the entire recursive delete.

The new batch query uses:

```sql
FROM inodes i
JOIN contents c ON c.inode_id = i.inode_id   -- INNER join
```

fileIDs without a matching `contents` row are silently excluded from results. This is more robust in practice (a dangling reference shouldn't block cleanup), but the behavioral change should be documented. Consider adding a comment above `scanOrphanedFilesByIDTx` noting that missing backing rows are intentionally skipped rather than surfaced as errors.

#### 2. Pre-existing TOCTOU race on orphan detection under TiDB (not introduced by this PR)

The delete flow in both `DeleteFileWithRefCheck` and `DeleteDirRecursive` follows this sequence:

1. Delete `file_nodes` rows
2. Lock inodes/files via `lockFileIDsForDeleteTx` (new)
3. `SELECT COUNT(*) ... FOR UPDATE` to check for remaining references

On TiDB, gap locks are not applied on missing rows like InnoDB does. Between step 2 and step 3, another transaction can insert a new `file_nodes` row referencing the same fileID, and step 3 may still see count=0. The newly-added `lockFileIDsForDeleteTx` locks the *inode* row itself, which prevents concurrent status changes on that inode, but does not prevent new `file_nodes` inserts.

This race exists in the old code as well — this PR does not make it worse. Worth tracking as a follow-up issue; a fix (reference counting, separate ref-count table) is out of scope here.

#### 3. Duplicate `ResolveReadTarget` calls under concurrent reads

`readTargetForHandle` uses a double-check pattern:

```go
// first check (fast path, under lock)
target := fh.ReadTarget
if target != nil { return target }

// slow path (no lock held)
resolved, _ := fs.client.ResolveReadTarget(ctx, remotePath)

// second check (under lock)
if fh.ReadTarget == nil { fh.ReadTarget = resolved }
```

Two concurrent `Read` calls racing into the slow path will both call `ResolveReadTarget`, wasting one HTTP round-trip. Only one result is stored. This is correct (the double-check guarantees single-writer safety) but worth a brief code comment for the next reader.

#### 4. `candidate.file == nil` → `ErrNotFound` could use a more descriptive error

When `scanSplitDeleteCandidate` finds a `fileID` but the `contents` row is missing (`storageType` is NULL), the code returns a bare `ErrNotFound`. A wrapped error like:

```go
fmt.Errorf("file %s has no storage metadata: %w", candidate.fileID, ErrNotFound)
```

would make debugging data-inconsistency issues easier.

---

### Minor

#### 5. `InsertNode` (no-Tx variant) is now dead code in production

All production callers have moved to `InsertNodeTx` (which includes the new `lockConfirmedFileForAttachTx` check). Test code continues to use `InsertNode` for setup, which is fine since tests don't need the attach check. Consider adding a doc comment on `InsertNode` noting it should not be used for production paths that carry a `file_id`.

#### 6. Naming: `scanDeleteCandidateTx` returns a single candidate

The name suggests scanning multiple candidates; it returns exactly one. `getDeleteCandidateTx` or `loadDeleteCandidateTx` would be more accurate. Not blocking — pattern is consistent with `scanFileForGCTx` elsewhere.

#### 7. HTTP transport pool sizing: cross-reference the two constants

`pkg/client/client.go`: `MaxIdleConns=256, MaxIdleConnsPerHost=64`  
`pkg/s3client/aws.go`: `MaxIdleConns=512, MaxIdleConnsPerHost=128`

The 2x ratio is justified in the S3 comment (data-plane fan-out), but the two comments don't reference each other. A brief note like "keep S3 pool 2x dat9 client pool" in both files would help future maintainers understand the relationship.

#### 8. `fileFromLegacyGCScan` has 17 parameters

Functionally correct, but a struct-based approach would improve readability. Follow-up refactor candidate.

---

### Test Coverage

New tests are well-placed and cover the key paths:

| Test | Quality |
|------|---------|
| `TestMountCmdPassesReadCacheMaxFileOption` | Good — CLI flag wiring |
| `TestResolveReadTargetAndReadObjectRange` | Good — 302→range read flow |
| `TestInsertNodeTxRejectsDeletedFile` | Good — DELETED file rejection |
| `TestReadCache_CustomMaxFileSize` | Good — boundary check |
| `TestOpenReadOnlyCacheableFileSkipsPrefetcher` | Good — no prefetcher for cacheable files |
| Rename target-clearing test | Good — ReadTarget + Prefetch target cleared on rename |

**Suggested additions** (non-blocking):
- Multi-reference file deletion (fileID shared by multiple paths → only node deleted, not file)
- Large batch `DeleteDirRecursive` (>1000 fileIDs) boundary test
- Concurrent rename + read race test for ReadTarget correctness

---

### Style

- ✅ `context.Context` first parameter throughout
- ✅ Error wrapping with `%w`
- ✅ `sync.Mutex` named `mu`
- ✅ Import grouping (stdlib / third-party / internal)
- ✅ `t.Helper()` + `t.Cleanup` in tests
- ✅ New exported APIs (`ReadTarget`, `ResolveReadTarget`, `ReadObjectRange`) have docstrings

---

### Verdict

**LGTM with minor notes.** No critical blockers. The three items worth addressing before or right after merge:

1. Add a comment on `scanOrphanedFilesByIDTx` about intentional silent skip of missing backing rows.
2. Add a brief comment on `readTargetForHandle` noting the duplicate-resolution race is intentional.
3. File a follow-up issue for the TiDB TOCTOU orphan-detection race (pre-existing, not introduced here).

mornyx (commented):
## Re-review: commits `20490ac` + `ad9b48c`

Two new commits since the first review round. All tests pass (`pkg/client`, `pkg/fuse`, `pkg/backend`, `pkg/datastore`, `pkg/s3client`, `cmd/drive9/cli`). Build is clean.

---

### Commit `20490ac` — "address read and delete review feedback"

This addresses the code-review feedback from the first round cleanly:

| Feedback item | Fix |
|---|---|
| Helpers lacked `ctx` — used bare `Query`/`Exec` | All helpers now take `ctx context.Context` as first param and use `*Context` variants |
| Single giant IN clause could exceed parameter limits | All batch helpers (`scanOrphanedFilesByIDTx`, `markFilesDeletedTx`, `deleteFileTagsByIDsTx`, `lockFileIDsForDeleteTx`) now chunk in batches of 1000 |
| `lockFileIDsForDeleteTx` missing from `DeleteDirRecursive` in original PR | Now calls `lockFileIDsForDeleteTx` **before** deleting `file_nodes` — improved locking order over the original PR |

Additional improvements:
- Added `closeRowsAfterScan` helper that properly drains and checks `rows.Err()` on FOR UPDATE results
- All `tx.Exec`/`tx.Query` calls in delete paths now use context-aware variants

No issues found in this commit.

---

### Commit `ad9b48c` — "refresh cached read targets"

This is a robust addition: presigned S3 URLs can expire mid-read, and the code now detects this and automatically refreshes.

#### Changes summary:

**1. Presign expiry detection** (`pkg/client/transfer.go`)
- `readObjectRangeStrict` handles HTTP 403 from S3 as `errPresignExpired`
- New exported `IsPresignExpired(err)` for callers
- `ReadTarget` doc comment updated to reflect the refresh capability
- ✅ New test: `TestReadObjectRangePresignExpired`

**2. `readStreamRangeWithRetry` signature change** (`pkg/fuse/dat9fs.go`)
- Now takes `fh *FileHandle` instead of `target *client.ReadTarget`
- Resolves `ReadTarget` internally via `readTargetForHandle`
- On presign expiry: clears the stale target, re-resolves a fresh one, retries immediately
- Same presign-refresh logic in the retry loop (each retry iteration can refresh)
- ✅ New test: `TestReadStreamRangeWithRetryRefreshesExpiredReadTarget`

**3. Prefetch presign refresh** (`pkg/fuse/prefetch.go`)
- Background prefetch goroutine falls back to `ReadStreamRange` (control-plane redirect) on presign expiry
- Clears the prefetcher cached target so next prefetch re-resolves

**4. Comprehensive read target invalidation** — all cache-invalidation sites now also clear resolved `ReadTarget`s:
| Site | Mechanism |
|------|-----------|
| `SetAttr` (truncate/chmod) | `invalidateReadCacheAndTargets` |
| `Unlink` | `invalidateReadCacheAndTargets` |
| `Rename` | `invalidateReadCacheAndTargets` for old+new paths |
| `Release` (dirty + clean) | `invalidateReadCacheAndTargets` |
| `flushHandle` (Paths 1a/1b/2/3) | `clearReadTargetForLockedHandle` + `invalidateReadCacheAndTargetsExcept` |
| `flushHandleDebounced` | `clearReadTargetsForPath` |
| `onCommitQueueSuccess` | `clearReadTargetsForPath` |
| SSE change event | `invalidateReadCacheAndTargets` |
| SSE reset | `clearAllReadTargets` |
| `retargetOpenHandlesForRename` | direct `nil` assignment (existing) |
- ✅ New test: `TestDat9FSClearReadTargetsForPath`

**5. `Read` method simplified** — no longer calls `readTargetForHandle` separately; it's now internal to `readStreamRangeWithRetry`

#### Lock ordering verification ✅

New ops and their lock orderings are safe:

| Operation | Lock order |
|-----------|-----------|
| `clearAllReadTargets` | `HandleTable.mu` → `FileHandle.mu` |
| `clearReadTargetsForPath` | `OpenHandleIndex.mu` (RLock, released) → `FileHandle.mu` |
| `invalidateReadCacheAndTargetsExcept` (from flushHandle) | `FileHandle.mu` (held by caller) → `OpenHandleIndex.mu` (RLock, released by SnapshotPath) → other `FileHandle.mu` |
| Normal `Read` → `readTargetForHandle` | `FileHandle.mu` → released → resolve → `FileHandle.mu` |
| `Prefetcher.SetReadTarget` (from goroutine) | `Prefetcher.mu` (independent of handle locks) |

No lock-inversion deadlock risks identified. The hierarchy `HandleTable.mu` > `FileHandle.mu` > `OpenHandleIndex.mu` is consistently maintained.

---

### Minor notes (non-blocking)

1. **`readRetrySuccess` metric semantics**: When a presign expires and the immediate refresh succeeds, `readRetrySuccess` is incremented. This metric now counts both transient-error retries and presign refreshes. The distinction is arguably not important for operations — both are "retry after initial failure" — but worth noting for anyone reading dashboards.

2. **Prefetch presign fallback is fire-and-forget**: When the prefetch goroutine detects presign expiry, it falls back to `ReadStreamRange` with a single attempt (no retry). If that also fails, the block is marked with error and the user's `Read` call will go through the normal range-read path (which does retry). This is correct behavior for an optimization path.

3. **`readStreamRangeWithRetry` context usage**: The presign-refresh retry uses `context.Background()` (via `readTransientRetryTimeout`). The original FUSE request context may have been cancelled by then (which is why `doRangeRead` failed with the presign error in the first place). This is intentional and correct — the refresh is a best-effort side operation to update the cached target.

---

### Verdict

Both commits are clean, well-tested, and correctly address the feedback from the previous round. The presign-refresh mechanism is a solid piece of defensive engineering — it turns a hard failure (stale S3 URL → EIO to user) into a transparent recovery. Comprehensive invalidation at all mutation sites ensures correctness.

**No blocking issues. LGTM.**

mornyx (approved):

srstack:
已解决：
- CodeRabbit outside-diff：`pkg/client/transfer.go` 并行下载现在遇到 presigned URL 403 过期会重新 `ResolveReadTarget`，更新共享 target，并重试当前 range。
- Nit：新增 `client.IsReadTargetNoRedirect`，外部调用方可识别 inline-file fallback。
- Nit：`pkg/fuse/open_handles_test.go` 已补充非匹配 handle 的 prefetch target 保留断言，并整理 import 分组。

验证：`go test ./pkg/fuse ./pkg/backend ./pkg/client ./pkg/s3client ./pkg/datastore ./cmd/drive9/cli`、`git diff --check` 均通过。

提交：dde2df3

PR #438, opened 2026-05-19
auth: add workspace zone scope foundation

qiffang:
## Summary
- Add additive meta schema for workspace-zone scoped API keys: `tenant_api_keys.scope_kind` defaults to `owner`, plus `tenant_api_key_fs_scopes` for per-key FS prefix/op rows.
- Load FS scope rows only when the resolved API key has `scope_kind = 'fs_scoped'`; legacy/owner keys do not query the new scope table.
- Add request-local `TenantScope` fields, pure in-memory `AuthorizeFS` / `AuthorizeFSPair`, scoped endpoint default-deny guard, and tests for owner compatibility, scope loading, prefix boundaries, path normalization, op semantics, and tenant-scoped row lookup.

## Scope / Non-Claims
- This is PR A foundation only. It does not expose scoped token CLI/API yet and does not wire every FS handler path check; those are split into PR B/C1/C2.
- No audit/log surface is added.
- Owner/legacy keys remain full-access and keep the hot `AuthorizeFS` fast path.

## Owner Performance Guard
- `scope_kind` is read with the existing API key row.
- Owner/legacy keys skip `ListAPIKeyFSScopes`; `TestAuthOwnerKeyDoesNotLoadFSScopes` uses a fake loader to pin this.
- `AuthorizeFS` never queries DB; it reads request-local `TenantScope` only.

Benchmark on local darwin/arm64, `go test ./pkg/server -run '^$' -bench 'BenchmarkAuthorizeFS' -benchtime=200ms -count=3`:

```text
BenchmarkAuthorizeFSOwnerToken-10    ~2.6 ns/op
BenchmarkAuthorizeFSSingleZone-10    ~150-162 ns/op
BenchmarkAuthorizeFSTenZones-10      ~446-562 ns/op
```

## Verification
- `go test ./pkg/server -run '^$' -bench 'BenchmarkAuthorizeFS' -benchtime=200ms -count=3`
- `go vet ./pkg/server ./pkg/meta`
- `go build ./cmd/drive9`

## Summary by CodeRabbit

* **New Features**
  * API keys can now be scoped to specific filesystem paths and operations
  * Filesystem-scoped API keys enforce granular access controls with configurable operations (read, write, delete, search, list)
  * Scoped API keys are denied access to protected business paths

[Review Change Stack](https://app.coderabbit.ai/change-stack/mem9-ai/drive9/pull/438?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/pkg/meta/meta.go
+++ b/pkg/meta/meta.go
@@ -3,7 +3,9 @@ package meta
 
 import (
 	"context"
+	"crypto/sha256"
 	"database/sql"
+	"encoding/hex"
 	"errors"
 	"fmt"
 	"sort"
@@ -14,6 +16,7 @@ import (
 	"github.com/mem9-ai/dat9/pkg/logger"
 	"github.com/mem9-ai/dat9/pkg/metrics"
 	"github.com/mem9-ai/dat9/pkg/mysqlutil"
+	"github.com/mem9-ai/dat9/pkg/pathutil"
 	"go.uber.org/zap"
 )
 
@@ -47,6 +50,13 @@ type APIKeyStatus string
 const (
 	APIKeyActive  APIKeyStatus = "active"
 	APIKeyRevoked APIKeyStatus = "revoked"
+)
+
+type APIKeyScopeKind string
+
+const (
+	APIKeyScopeKindOwner APIKeyScopeKind = "owner"
+	APIKeyScopeKindFS    APIKeyScopeKind = "fs_scoped"
 )
 
 type Tenant struct {
@@ -133,10 +143,21 @@ type APIKey struct {
 	JWTHash       string
 	TokenVersion  int
 	Status        APIKeyStatus
+	ScopeKind     APIKeyScopeKind
 	IssuedAt      time.Time
 	RevokedAt     *time.Time
 	CreatedAt     time.Time
 	UpdatedAt     time.Time
+}
+
+type APIKeyFSScope struct {
+	TenantID   string
+	APIKeyID   string
+	Prefix     string
+	PrefixHash string
+	Ops        string
+	CreatedAt  time.Time
+	UpdatedAt  time.Time
 }
 
 type TenantWithAPIKey struct {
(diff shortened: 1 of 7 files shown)


qiffang:
@adversary-1 review: **APPROVED** for PR A foundation at `0908006`, pending GitHub CI green. (I cannot submit a formal GitHub approval with the current CLI identity; posting the review result here.)

I re-reviewed the amended commit after the `rows.Close` lint fix. The owner performance invariant is structurally enforced: `scope_kind` is read with the existing API key row, owner/legacy keys do not call `ListAPIKeyFSScopes`, and `AuthorizeFS` is request-local/in-memory with the `!IsScoped` fast path. Scoped keys fail closed for non-whitelisted business endpoints, and the authorizer tests cover segment boundary, `..` normalization, `search` requiring `read`, `AuthorizeFSPair`, and cross-tenant scope-row isolation.

Local verification:
- `go test ./pkg/server -run '^$' -bench 'BenchmarkAuthorizeFS' -benchtime=200ms -count=3`
- `go vet ./pkg/server ./pkg/meta`
- `go build ./cmd/drive9`

One release-order caveat: do not merge PR B (public scoped token issuance) before C1/C2 path checks are wired. PR A is safe as foundation because scoped tokens are not publicly issuable yet; once issuance exists, FS/upload endpoints must already enforce path-level `AuthorizeFS`, otherwise a scoped token can reach whitelisted FS/upload handlers without path checks. If PR B has to land earlier, it should refuse issuing usable `fs_scoped` tokens until C1/C2 is present.

PR B should also validate scope prefixes so root access is explicit (`/` or `:/`), not an accidental empty/bare `:` prefix.

qiffang:
@adversary-1 final review: **APPROVED** for PR A foundation at `bd57bff`, pending GitHub `ci` green. (The current CLI identity cannot submit a formal GitHub approval, so posting the review result here.)

Final re-verify after the hardening commits:
- PR A now keeps `fs_scoped` keys out of all business data-plane endpoints until C1/C2 explicitly opt routes in with handler-level path authorization. This removes the unsafe intermediate state where a scoped key could reach FS/upload before path checks exist.
- Owner/legacy performance invariant still holds structurally: `scope_kind` is read with the existing API key row, owner keys do not call `ListAPIKeyFSScopes`, and `AuthorizeFS` is pure request-local state with the `!IsScoped` fast path.
- Store/parser validation is now aligned: blank prefixes and bare `:` are rejected; explicit root (`/` / `:/`) remains accepted; ops are validated at store insert and runtime parse, including `search` requiring `read`.
- Store-side prefix canonicalization and runtime authorization both use `pathutil.Canonicalize` after `:` trimming, so policy storage and request matching share the same normalization semantics.
- Cross-tenant scope lookup, segment-boundary prefix matching, escaping path rejection, `AuthorizeFSPair`, and scoped endpoint default-deny all have regression coverage.

Local verification on `bd57bff`:
- `go test ./pkg/server -run '^$' -bench 'BenchmarkAuthorizeFS' -benchtime=200ms -count=3` (owner ~2.7ns/op)
- `go vet ./pkg/server ./pkg/meta`
- `go build ./cmd/drive9`

Sequencing requirement remains: PR B must not expose usable scoped token issuance before C1/C2 path authorization wiring is merged, unless issuance is explicitly gated off.

PR #440, opened 2026-05-19
auth: wire workspace zone write-side authorization (PR C2a)

qiffang:
## Summary

PR C2a — write-side handler wiring for Workspace Zones. PR C1 (#439) opened the read-side; this PR opens single-path write handlers (`handleWrite`, `handlePatch`, `handleAppend`, `handleCreate`, `handleMkdir`, `handleDelete`), dual-path handlers (`handleCopy`, `handleRename`), and unconditional scoped-token denial in `handleChmod`. The dispatcher whitelist is extended with action-specific allowlists for PUT/PATCH/DELETE/POST methods.

**Uploads (`/v1/uploads*`, `/v2/uploads/*`) remain dispatcher-denied for scoped tokens until PR C2b** wires their handlers + session re-authorize semantics. Per @adversary-1 + @dev-1 (msgs `08848b1a`, `32d3019f`): release-order safety — only open routes whose handlers in this PR authorize their target path.

Partial completion of task #8 (task stays open as umbrella; C2b finishes it).

## Owner perf invariant

`BenchmarkAuthorizeFSOwnerToken` = **2.08 ns/op** (unchanged from PR A's 2.7 ns/op). Owner tokens carry zero new dispatcher work — the C2a allowlist additions are scoped behind `scope.IsScoped` at `handleBusiness`, never run for owner tokens. @qiffang's red line preserved.

## Handler wiring

| Handler | Op(s) | Notes |
|---|---|---|
| `handleWrite` | `FSOpWrite` | PUT `/v1/fs/<path>` |
| `handlePatch` | `FSOpWrite` | PATCH `/v1/fs/<path>` |
| `handleAppend` | `FSOpWrite` | POST `?append=1` — **authorize BEFORE plan generation** (banked invariant: plan response leaks "this prefix is writable" even on later denial) |
| `handleCreate` | `FSOpWrite` | POST `?create=1` |
| `handleMkdir` | `FSOpWrite` | POST `?mkdir=1` |
| `handleDelete` | `FSOpDelete` | DELETE `/v1/fs/<path>` |
| `handleCopy` | `read` on src + `write` on dst | POST `?copy=1`. **Src in `X-Dat9-Copy-Source` HEADER, NOT URL** — banked invariant from cp -r PR #434. URL-only authorize would let a scoped token exfiltrate from any unreadable zone. |
[...]

PR #441, opened 2026-05-19
auth: wire workspace zone uploads authorization (PR C2b)

qiffang:
## Summary

PR C2b — uploads authorization for Workspace Zones (V1 + V2). C2a (#440) opened write-side `/v1/fs/*`; this PR closes the loop with `/v1/uploads*` and `/v2/uploads/*`. Task #8 completes with this merge.

The headline invariant from C2 review: **re-authorize current scope, NOT initiator scope** on every upload continuation (resume/complete/abort/presign). A scoped token's policy can change between initiate and complete via revoke+reissue with narrower zones; trusting the initiator would let a since-narrowed token finish a write outside its current scope.

## Handler wiring

| Handler | Op on... | Notes |
|---|---|---|
| `handleUploads` (GET list) | `FSOpList` on path query arg | Listing uploads at a path leaks "an upload is in progress here" metadata; gated behind list op |
| `handleUploadInitiate` | `FSOpWrite` on body.path | Authorize BEFORE any size/checksum validation or backend session creation |
| `handleUploadComplete` | re-authz `session.TargetPath`, `FSOpWrite` | BEFORE `ConfirmUploadWithTags` |
| `handleUploadResume` | re-authz with `FSOpWrite` | BEFORE `parseResumePartChecksums` |
| `handleUploadAbort` | re-authz with `FSOpDelete` | Abort = tear down session = delete semantics |
| `handleV2UploadInitiate` | `FSOpWrite` on body.path | Same shape as V1 |
| `handleV2PresignPart` | re-authz with `FSOpWrite` | Presign IS the write enablement (direct S3 PUT) — must gate |
| `handleV2PresignBatch` | re-authz with `FSOpWrite` | Same |
| `handleV2UploadComplete` | re-authz with `FSOpWrite` | BEFORE `ConfirmUploadV2` |
| `handleV2UploadAbort` | re-authz with `FSOpDelete` | Same delete semantics |

## New helper: `authorizeUploadSession`

`authorizeUploadSession(ctx, w, scope, b, uploadID, op) (*datastore.Upload, error)`:
- Looks up the session by ID
- Maps `datastore.ErrNotFound → 404`, `ErrUploadExpired → 410`, others → 500 (preserves existing client error shape — per @dev-1 msg `efb1e56c` constraint)
- Authorizes `upload.TargetPath` against current scope
- Returns the session on allow (no double-lookup at call site)
- **Owner short-circuit**: `(nil, nil)` for `!scope.IsScoped` — preserves exact pre-C2b behavior on owner hot path

## Critical regression caught and fixed mid-implementation

`TestUploadActionMetrics` (server_test.go:1620) is the existing compatibility test: owner-token `POST /v1/uploads/nonexistent/resume` (no body) expects **400** (missing checksums), not 404 (missing session). The reason: original flow was parse-body → 400 BEFORE backend call.

My initial draft moved session lookup to the front (for the re-authz invariant), which broke this: owner tokens hit 404 instead of 400. **Owner short-circuit fix**: `authorizeUploadSession` returns `(nil, nil)` for non-scoped tokens, so the handler does its own session lookup (or backend call) at the same point in the flow as before C2b. Pre-existing error shape preserved exactly.

`handleUploadComplete` previously fetched the upload BEFORE confirm to publish the target path in the event. C2b reuses that fetched value for scoped tokens; for owner short-circuit, the handler now does its own GetUpload before `publishEvent`.

## Dispatcher whitelist extended

PR C2b admits at the dispatcher (release-order safety — only routes whose handlers in THIS PR fully path-authorize):
- `/v1/uploads` POST (initiate), GET (list)
- `/v1/uploads/*` POST (complete/resume), GET (resume), DELETE (abort)
- `/v2/uploads/*` POST (initiate, presign, presign-batch, complete, abort)

## Tests (3 new + 3 updated)

| Test | Scope |
|---|---|
| `TestC2bHandleUploadInitiateAuthorizesTargetPath` | V1: scoped token without /main/ zone, body `{"path":"/main/secrets.env",...}` → 403 with nil backend (proves authorize ordering) |
| `TestC2bHandleV2UploadInitiateAuthorizesTargetPath` | V2 twin |
| `TestC2bUploadsAdmittedAtDispatcher` | 12 entries covering every V1+V2 upload route admitted at dispatcher |

Test updates (renames reflecting new boundary):
- `TestScopedBusinessEndpointGuardDeniesNonC1Endpoints` → `...EndpointsOutOfWorkspaceZonesScope` (uploads removed from deny list)
- `TestIsScopedBusinessRequestAllowed.write-side-denied` → upload cases removed
- `TestC2aUploadsStillDeniedForScopedTokens` → `TestC2bUploadsAdmittedAtDispatcher` (boundary moved with C2b merge; assertion flipped)

## Verification

- [x] `go build ./...` clean
- [x] `go vet ./...` clean
- [x] `go test ./pkg/server ./pkg/meta -count=1` green (server 35s, meta 13s)
- [x] **Critical**: `TestUploadActionMetrics` still passes (owner-token 400 for resume-with-empty-body, not 404)
- [x] `BenchmarkAuthorizeFSOwnerToken` 2.0-2.1 ns/op (owner perf preserved)

## Test coverage gap acknowledged

End-to-end "owner-initiated session continued by scoped token after token narrowing" requires a real DB + multi-step test fixture (initiate → narrow scope → complete from new scope → 403). The unit-level `authorizeUploadSession` logic is exercised by ordering tests; full e2e is left for the integration test suite (which uses real MySQL containers per existing pattern).

## What's NOT in scope

- Token issue/revoke CLI (PR B / task #6 — sequencing-locked, follows C2b merge)
- Any change to chmod (still owner-only forever)
- Any change to `/v1/sql`, `/v1/fork`, `/v1/events`, `/v1/journals*`, `/v1/vault*` (permanently out of workspace-zones scope)

Reviewers: @adversary-1 primary + @adversary-2 secondary. Same admin-bypass merge path as PR A / C1 / C2a given shared-auth.

Task #8 (Workspace Zones C2 umbrella) completes with this merge. C1 + C2a + C2b together deliver full workspace zone authorization on every FS/upload endpoint for scoped tokens; owner tokens retain exact pre-PR behavior.

### DIFF
--- a/pkg/server/auth_test.go
+++ b/pkg/server/auth_test.go
@@ -379,14 +379,13 @@ func TestAuthScopedKeyLoadsFSScopes(t *testing.T) {
 	}
 }
 
-// TestScopedBusinessEndpointGuardDeniesEndpointsOutOfC1C2aScope checks
-// that scoped tokens are still 403'd at the dispatcher for every endpoint
-// NOT opened by PR C1 (read-side) or PR C2a (write-side except chmod).
-// Uploads remain dispatcher-denied until C2b wires their handlers +
-// session re-authorize. SQL/fork/events/journals/vault are permanently
-// out of scope. chmod stays owner-only forever. A bare POST /v1/fs/*
-// without an action selector also denies (ambiguous request).
-func TestScopedBusinessEndpointGuardDeniesEndpointsOutOfC1C2aScope(t *testing.T) {
+// TestScopedBusinessEndpointGuardDeniesEndpointsOutOfWorkspaceZonesScope
+// checks that scoped tokens are still 403'd at the dispatcher for every
+// endpoint NOT opened by PR C1 (read-side) / C2a (write-side) / C2b
+// (uploads). SQL/fork/events/journals/vault are permanently out of scope.
+// chmod stays owner-only forever. A bare POST /v1/fs/* without an action
+// selector also denies (ambiguous request).
+func TestScopedBusinessEndpointGuardDeniesEndpointsOutOfWorkspaceZonesScope(t *testing.T) {
 	type endpoint struct {
 		method string
 		path   string
@@ -394,8 +393,6 @@ func TestScopedBusinessEndpointGuardDeniesEndpointsOutOfC1C2aScope(t *testing.T)
 	deniedC1 := []endpoint{
 		{http.MethodPost, "/v1/sql"},
 		{http.MethodPost, "/v1/fs/file.txt"}, // no action selector → ambiguous → deny
-		{http.MethodPost, "/v1/uploads/initiate"},
-		{http.MethodPost, "/v1/uploads"},
 		{http.MethodPost, "/v1/fork"},
 		{http.MethodGet, "/v1/events"},
 		{http.MethodGet, "/v1/journals"},
--- a/pkg/server/fs_authorization.go
+++ b/pkg/server/fs_authorization.go
@@ -1,11 +1,14 @@
 package server
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"net/http"
 	"strings"
 
+	"github.com/mem9-ai/dat9/pkg/backend"
+	"github.com/mem9-ai/dat9/pkg/datastore"
 	"github.com/mem9-ai/dat9/pkg/meta"
 	"github.com/mem9-ai/dat9/pkg/pathutil"
 )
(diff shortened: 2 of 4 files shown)


qiffang:
@adversary-1 review on `f320f44`: **request changes**.

Finding (blocking): the scoped-token upload dispatcher allowlist is broader than the concrete upload dispatch table. In `pkg/server/server.go:484-502`, `isScopedBusinessRequestAllowed` admits upload routes by prefix + method:

- `GET /v1/uploads/initiate` is allowed because line 484 treats `/v1/uploads/initiate` the same as `/v1/uploads`, so it reaches the list handler path instead of being denied at the scoped dispatcher.
- `POST /v1/uploads/{id}/unknown` is allowed by the `/v1/uploads/` prefix even though `handleUploadAction` only dispatches complete/resume/delete cases (`server.go:2083-2092`).
- `POST /v2/uploads/{id}/future-action` is allowed by the `/v2/uploads/` prefix even though `handleV2Uploads` only dispatches initiate/presign/presign-batch/complete/abort (`server.go:2418-2431`).

This repeats the C1 class of bug: the security gate is not mirroring the downstream route/action table. Today these unknown routes fall through to 400/404, but the release-order invariant we locked was stricter: a scoped token should pass the dispatcher only for routes whose exact handler is wired with path authorization in the same PR. With the current prefix gate, a future upload endpoint under `/v1/uploads/` or `/v2/uploads/` can silently inherit scoped-token access before its own auth semantics are reviewed.

Requested fix:

- Replace the prefix/method upload gate with an action-aware helper that mirrors `handleUploads`, `handleUploadAction`, and `handleV2Uploads` exactly.
- Allow only the concrete C2b routes: V1 initiate/list, V1 complete/resume/abort, V2 initiate/presign/presign-batch/complete/abort.
- Deny unknown/future upload actions at the dispatcher for scoped tokens, and deny route/method mismatches such as `GET /v1/uploads/initiate`.
- Add negative dispatcher tests for at least `GET /v1/uploads/initiate?path=/x`, `POST /v1/uploads/u/unknown`, `GET /v1/uploads/u`, and `POST /v2/uploads/u/future-action`, alongside the existing positive C2b route tests.
- Update the stale comment at `server.go:432-435` / `457` which still says uploads are default-deny even though C2b opens them.

What I verified separately:

- `authorizeUploadSession` itself has the right current-scope reauth shape and owner short-circuit.
- V1/V2 initiate and continuation op mappings look right.
- Local checks passed on `f320f44`: `go test ./pkg/server -run "TestC2b|TestAuthorizeUploadSession|TestUploadActionMetrics" -count=1`, `go test ./pkg/server ./pkg/meta -count=1`, `go vet ./pkg/server ./pkg/meta`, `go build ./cmd/drive9`, `git diff --check origin/main...HEAD`, and `BenchmarkAuthorizeFS` owner path stayed ~2.1ns/op.

qiffang:
@adversary-1 re-review on `903a3ae`: **approved**.

The action-aware upload gate fixes the blocker from my prior review. I verified `isScopedV1UploadRouteAllowed` / `isScopedV2UploadRouteAllowed` now enumerate the concrete upload dispatch table instead of admitting the whole prefix family:

- V1 allows only `POST /v1/uploads`, `GET /v1/uploads`, `POST /v1/uploads/initiate`, `POST /v1/uploads/{id}/complete`, `POST|GET /v1/uploads/{id}/resume`, and `DELETE /v1/uploads/{id}`.
- V2 allows only `POST /v2/uploads/initiate` and `POST /v2/uploads/{id}/{presign,presign-batch,complete,abort}`.
- The new negative tests cover the cases I named plus wrong-method, missing-id, unknown-action, and future-route sentinels.

The original C2b core still looks right:

- `authorizeUploadSession` re-authorizes the stored session target against the current request scope, not the initiator scope.
- Owner tokens short-circuit without the new upload lookup, preserving both the owner hot path and existing owner error ordering.
- V1/V2 initiate paths authorize body `path` before creating a session; continuation paths authorize session target before mutation/presign/complete/abort.
- Op mapping is consistent: write for initiate/resume/complete/presign, delete for abort, list for upload list.

Local verification on `903a3ae`:

- `go test ./pkg/server -run "TestC2b|TestAuthorizeUploadSession|TestUploadActionMetrics" -count=1`
- `go test ./pkg/server -run '^$' -bench 'BenchmarkAuthorizeFS' -benchtime=200ms -count=3` (owner ~2.5ns/op locally)
- `go vet ./pkg/server ./pkg/meta`
- `go build ./cmd/drive9`

GitHub CI/local-e2e/CodeRabbit are green on this head.

Non-blocking cleanup: the high-level comment above `isScopedBusinessRequestAllowed` still says uploads are default-deny / follow-up, and the `/v1/fs/*` comment still says uploads deny. The upload-specific comment below is correct, so this is just stale text cleanup, not a merge blocker.

Issue #10, opened 2026-03-26
Small file write path is non-transactional — createAndWrite blob + metadata not in same transaction


qiffang:
## Current State

Design doc §8 explicitly requires atomic small file writes:

> **Atomicity**: db9's fs9 and SQL share the same TiKV transaction context — \`fs9_write()\` inside a \`BEGIN/COMMIT\` block is atomic with the metadata INSERTs.

But the P0 implementation (SQLite + local blob) in \`createAndWrite\` is **non-transactional**:

\`\`\`go
// pkg/backend/dat9.go:198-228
func (b *Dat9Backend) createAndWrite(path string, data []byte) (int64, error) {
    // 1. writeBlob — writes to disk (outside transaction)
    if err := b.writeBlob(storageRef, data); err != nil { return 0, err }
    // 2. InsertFile — standalone SQL INSERT
    if err := b.store.InsertFile(...); err != nil { return 0, err }
    // 3. EnsureParentDirs — multiple standalone SQL INSERTs
    if err := b.store.EnsureParentDirs(...); err != nil { return 0, err }
    // 4. InsertNode — standalone SQL INSERT
    return b.store.InsertNode(...)
}
\`\`\`

If any step 2-4 fails, previous steps are not rolled back.

**Same issue exists in:**
- \`overwriteFile\` (dat9.go:231) — blob written but metadata update fails → orphan blob
- \`Create\` (dat9.go:60) — same non-transactional chain

## Impact

- Crash recovery inconsistency: blob exists but metadata missing (ghost files)
- Concurrent writes may produce orphan blobs
- \`InsertNode\` fails but \`InsertFile\` succeeded → unreferenced file record

## Proposed Fix

Wrap Store multi-step operations in a SQLite transaction:

\`\`\`go
func (b *Dat9Backend) createAndWrite(path string, data []byte) (int64, error) {
    if err := b.writeBlob(storageRef, data); err != nil { return 0, err }
    tx, err := b.store.DB().Begin()
    if err != nil { return 0, err }
    defer tx.Rollback()
    // InsertFile, EnsureParentDirs, InsertNode all use tx
    return int64(len(data)), tx.Commit()
}
\`\`\`

## Files Involved

- \`pkg/backend/dat9.go\` — createAndWrite, overwriteFile, Create
- \`pkg/meta/meta.go\` — provide tx-accepting method variants

PR #443, opened 2026-05-20
ux: make scoped token names local

qiffang:
## Summary

Implements the task #10 scoped-token UX correction from qiffang's Unix/Plan9 framing:

- `drive9 token issue [name]` now treats `name` as a local config/context alias only.
- `--subject` is optional audit metadata and no longer required or unique.
- Default human output no longer prints `token_id`.
- Adds local `fs_scoped` contexts usable by fs/mount commands.
- Adds `drive9 token list`, `drive9 token forget <name>`, `drive9 token revoke <name>`, and safe secret input via `drive9 token revoke -` / `--api-key-file`.
- Adds owner-authenticated revoke-by-capability (`POST /v1/tokens/revoke`) by hashing the presented target API key and revoking the matching active fs_scoped key in the same tenant.
- Keeps legacy `DELETE /v1/tokens/{id}` for backcompat/internal use.
- Drops the obsolete `(tenant_id, key_name)` unique index so repeated audit subjects are valid.

## Design Notes

The invariant is: `dat9_...` is the capability; local `name` is only client namespace. The server does not authorize or revoke by local name.

Security boundary for revoke-by-capability:

- Caller must authenticate as owner for the tenant.
- Target api key is request data, hashed server-side.
- Server only revokes matching active `fs_scoped` keys in the same tenant.
- Owner keys are not revoked through the new capability path.
- CLI rejects raw `dat9_...` positional argv and asks users to pipe it to stdin instead.

## Tests

Passed:

- `go test ./cmd/drive9/cli ./pkg/client ./pkg/server ./pkg/meta -count=1`
- `go test ./cmd/drive9 -count=1`
- `go vet ./cmd/drive9 ./cmd/drive9/cli ./pkg/client ./pkg/server ./pkg/meta`
- `go build ./cmd/drive9`
- `git diff --check`

Also tried:

- `go test ./... -count=1`

That broader run failed because multiple package test suites attempted to start MySQL containers and timed out waiting for `port: 3306  MySQL Community Server`; the targeted MySQL-backed `pkg/server` and `pkg/meta` package tests passed immediately before this broader run.

## New Features
* Added `token list` command to view locally saved filesystem-scoped token contexts.
* Added `token forget` command to remove saved token contexts locally.
* Enhanced `token revoke` command to revoke tokens by local name, token ID, or via file input.
* Introduced filesystem-scoped credentials as a new credential type alongside owner credentials.
* Added `/v1/tokens/revoke` endpoint for revoking scoped tokens by API key.

## Bug Fixes
* Made subject field optional when issuing scoped tokens.

## Tests
* Added comprehensive test coverage for new token management commands and credential types.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mem9-ai/drive9/pull/443?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/cmd/drive9/cli/client.go
+++ b/cmd/drive9/cli/client.go
@@ -36,7 +36,7 @@ const fsClientWarmTimeout = 3 * time.Second
 func NewFromEnv() *client.Client {
 	r := ResolveCredentials()
 	apiKey := ""
-	if r.Kind == CredentialOwner {
+	if r.Kind == CredentialOwner || r.Kind == CredentialFSScoped {
 		apiKey = r.APIKey
 	}
 	return client.New(r.Server, apiKey)
--- a/cmd/drive9/cli/config.go
+++ b/cmd/drive9/cli/config.go
@@ -16,6 +16,7 @@ type PrincipalType string
 const (
 	PrincipalOwner     PrincipalType = "owner"
 	PrincipalDelegated PrincipalType = "delegated"
+	PrincipalFSScoped  PrincipalType = "fs_scoped"
 )
 
 // Perm is the scope permission carried by a delegated context's JWT. Spec §6
@@ -31,8 +32,9 @@ const (
 // Type per spec §13.1:
 //
 //   - owner:     APIKey, Server
+//   - fs_scoped: APIKey, Server, Scope[], ExpiresAt
 //   - delegated: Token, Server (from iss), Agent, Scope[], Perm, ExpiresAt,
-//                GrantID, optional LabelHint
+//     GrantID, optional LabelHint
 //
 // The delegated fields are populated by locally decoding the JWT payload at
 // `ctx import` time. This is UX-only — authorization remains server-side
@@ -121,7 +123,7 @@ func saveConfig(cfg *Config) error {
 // active context is delegated or absent.
 func (c *Config) CurrentAPIKey() string {
 	ctx := c.currentContextEntry()
-	if ctx == nil || ctx.Type != PrincipalOwner {
+	if ctx == nil || (ctx.Type != PrincipalOwner && ctx.Type != PrincipalFSScoped) {
 		return ""
 	}
 	return ctx.APIKey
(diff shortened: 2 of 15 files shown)


mashenjun (changes requested):
One blocker on failure atomicity before this ships.

`TokenIssue` issues the remote scoped token first, then saves the local `fs_scoped` context. If `saveScopedTokenContext` fails after the server has returned `resp.Token` (config dir unwritable, file permission issue, JSON/write failure, etc.), the command returns an error without printing the token and without revoking it. That leaves a live scoped capability that is neither visible in the local namespace nor recoverable through the new default UX.

This is the same class of issue we just fixed server-side in PR B: validation/local bookkeeping failure after creating the key must not leave an unreachable active token.

Suggested fix: after a named issue succeeds remotely, if saving the local context fails, attempt a best-effort rollback via the new owner-authenticated `RevokeScopedTokenByAPIKey(resp.Token)` before returning the local-save error. Add a regression that forces `saveConfig`/context save to fail after the issue response and asserts the revoke endpoint is called. If rollback fails too, the error must make it explicit that a token was issued but not saved/revoked; do not silently lose the capability.

The core design otherwise matches the locked model: local name stays client-side, subject is optional/non-unique, and revoke-by-capability is owner-gated and scoped-token-only. I also agree with @dev-2's suggested extra server regressions for cross-tenant revoke and scoped caller rejection on `/v1/tokens/revoke`.

mashenjun (approved):
Approved. The blocker is fixed.

I re-reviewed head `877768f` with focus on the previous failure-atomicity issue and the revoke boundary regressions:

- Named `token issue` now compensates local context save failure by calling revoke-by-capability on the just-issued bearer.
- If rollback succeeds, the returned error says the issued token was revoked; if rollback fails, the error explicitly keeps `token=...` so the operator still has a recovery handle.
- CLI tests cover successful rollback, rollback-failure reporting, and now pin the recovery token substring.
- Server regressions cover scoped caller denial on `POST /v1/tokens/revoke` and cross-tenant owner revoke returning 404 without revoking the target.

The original Unix/Plan9 model is preserved: local name stays client-local, server authority is capability hash + scope/status, and `token_id` remains backcompat/internal rather than default human UX.

PR #446, opened 2026-05-20
feat: drive9 mount --perf-dir/--perf-addr

mornyx:
## Summary
- Add `drive9 mount --perf-dir <dir>` as the simple FUSE profiling switch.
- `--perf-dir` writes the standard profiling artifacts under one directory: `perf.jsonl`, retained rotated `perf.jsonl.N` sample files, `heap-final.pprof`, periodic windowed `cpu-*.pprof`, periodic `heap-*.pprof`, and a local pprof listener for manual inspection.
- CPU profiling is always windowed. With `--perf-dir`, the default CPU window is `--perf-cpu-duration 30s` every `--perf-cpu-interval 10m`; one CPU capture starts immediately when profiling begins, and subsequent captures follow the interval. Both values can be tuned independently, and duration must be shorter than interval.
- Heap profiling writes `heap-final.pprof` on unmount and periodic `heap-*.pprof` every `--perf-heap-interval 10m` by default; the interval can be tuned explicitly.
- `--perf-max-samples` controls the maximum samples per continuous perf JSONL segment, defaulting to 7200 when `--perf-dir` is set.
- `--perf-max-sample-files` controls retained continuous perf sample files, defaulting to 2 files: active `perf.jsonl` plus `perf.jsonl.1`.
- `--perf-max-profile-files` controls retained timestamped CPU and heap profile files per type, defaulting to 48; `heap-final.pprof` is not part of this retention cleanup.
- Keep only these advanced mount profiling controls: `--perf-interval`, `--perf-max-samples`, `--perf-max-sample-files`, `--perf-max-profile-files`, `--perf-cpu-duration`, `--perf-cpu-interval`, `--perf-heap-interval`, and `--perf-addr`.
- Migrate the remaining e2e perf smoke scripts from the removed `--perf-counters` flag to `--perf-dir` with low-cost profiling intervals, preserving their shutdown perf-summary assertions.

## Scope
This PR is intentionally limited to mount-side profiling output production. It does not add docs/design files, a top-level `drive9 perf` command, support bundle command, benchmark workload scripts, performance optimizations, commit-queue tuning, small-file batching, metadata-path changes, or read-prefetch changes.

## Tests
- `env GOCACHE=/Users/mornyx/Desktop/repos/github.com/mem9-ai/drive9/.codex-gocache go test ./cmd/drive9 ./cmd/drive9/cli ./pkg/fuse ./pkg/mountstate`
- `env GOCACHE=/Users/mornyx/Desktop/repos/github.com/mem9-ai/drive9/.codex-gocache go test ./cmd/drive9/cli ./pkg/fuse`
- `env GOCACHE=/private/tmp/drive9-feat-perf-srstack-446/.codex-gocache go test ./cmd/drive9 ./cmd/drive9/cli ./pkg/fuse ./pkg/mountstate`
- `bash -n e2e/git-ops-smoke-test.sh e2e/git-workspace-smoke-test.sh e2e/fuse-write-perf-budget-test.sh`
- `git diff --check`

## Release Notes

* **New Features**
  * Added continuous performance profiling with configurable intervals and retention policies.
  * Added CPU and heap profiling capabilities with customizable timing and output directories.
  * Introduced HTTP pprof server for real-time performance inspection.
  * Enhanced performance metrics to include latency percentiles (p50, p95, p99) and maximum latency.


qiffang (commented):
## PR #446 Review — dat9-dev1

Overall: Well-structured PR. The separation between diagnostics/report (this PR) and allocation optimization (#448) is good. Design docs are thorough and the JSONL sample envelope is well thought out. A few issues below, mostly non-blocking.

### Blockers

**B1. `perf collect` manifest leaks raw `duration` field**
`perfCollectManifest.Duration` is set to `duration.String()` which formats the Go `*time.Duration` pointer. This should be `(*duration).String()` or dereferenced. Currently it prints the pointer address, not the actual duration value.

**B2. `perf` usage string is out of sync — missing `sync` subcommand**
`perfUsage()` lists `collect` and `summarize`, but the dispatch also handles `sync`. The main `usage()` in `main.go` also only shows `perf <collect|summarize>`. Add `sync` to both.

**B3. `ProcessState` stores raw Server/MountPoint/RemoteRoot in PID file**
The PID file (`~/.drive9/run/<hash>.json`) now stores `server`, `mount_point`, and `remote_root` in cleartext. This is intentional for `perf collect` discovery, but the PID file has mode 0644 (from the existing `WritePID`). These are sensitive-ish paths. Consider tightening to 0600 for consistency with config file permissions. At minimum, document the trade-off.

### Non-Blocking

**N1. `CommitQueue.WaitAll` busy-polls at 50ms**
The 50ms ticker poll works but is wasteful for long waits. Consider using a condition variable or channel signal from commit completion. Same applies to `WriteBackUploader.WaitAll`. Low priority since this is only called from perf sync, not hot path.

**N2. `cold-read.sh` default seed size inconsistency**
`cold-read.sh` uses `COLD_READ_SEED_MB` defaulting to 256, but the perf report documents "32 MiB seeded through the Drive9 API". The 256 default seems like a leftover from larger test sizes. Verify this is intentional or align with the report.

**N3. `run.sh` missing `--profile-heap-interval 0s` rejection for negative values**
The script passes `DRIVE9_PROFILE_HEAP_INTERVAL` directly to `--profile-heap-interval`. If someone sets a negative value, it reaches the mount without validation. Low risk.

**N4. `metadata-walk.sh` defaults differ from perf report**
The script defaults to `METADATA_DIRS=100, FILES_PER_DIR=10` (1000 files), but the report says "10 directories x 5 files". Document which values were used for the EC2 run or make the defaults match.

**N5. `perfLatencyPercentile` rounding**
`target = (count*percentile + 99) / 100` — the `+99` gives ceiling division, so p50 of 1 sample returns bucket index for that sample. This is correct but slightly unusual. A brief comment explaining the ceiling intent would help.

**N6. `FlushAll` now logs errors via `log.Printf`**
`SyncAll` error is logged with `log.Printf("flush all sync failed: %v", err)`, but the rest of the FUSE package uses `fmt.Fprintf(os.Stderr, ...)`. Inconsistency — use the same pattern.

### Good

- Sample envelope design is solid — shape-versioned, additive, no path cardinality
- Segmented JSONL rotation with bounded retention is the right call
- `redactProcessState` in the collect manifest is good privacy hygiene
- Latency histogram with 28 power-of-2 buckets covers the full FUSE range well
- `SyncAll` as a non-stopping fence separate from `FlushAll` shutdown is clean
- Comprehensive test coverage for profiler, continuous perf, perf CLI, and mount flag plumbing

cmd/drive9/cli/perf.go
```diff
@@ -0,0 +1,746 @@
...
+			warnings = append(warnings, fmt.Sprintf("summarize perf jsonl: %v", err))
+		} else if data, err := marshalSummary(summary, true); err != nil {
+			warnings = append(warnings, fmt.Sprintf("marshal summary: %v", err))
+		} else if err := os.WriteFile(filepath.Join(tmp, "summary.json"), data, 0o644); err != nil {
+			warnings = append(warnings, fmt.Sprintf("write summary.json: %v", err))
+		}
+	}
+	manifest := perfCollectManifest{
+		GeneratedAt: time.Now().UTC().Format(time.RFC3339Nano),
+		Duration:    duration.String(),
```
qiffang: **Correction on B1**: `flag.FlagSet.Duration` returns `*time.Duration`, and Go auto-dereferences pointer receivers for method calls, so `duration.String()` correctly formats the duration value. B1 is withdrawn — that leaves B2 and B3 as the remaining blockers.

qiffang (changes requested):
## PR #446 Review — adversary-2 — Round 1: REQUEST_CHANGES

**Head**: 76ab784a | **Files**: 34 | **+3619 / -22**

---

### Blockers

**B1 — Path traversal via `/debug/drive9/profile/cpu/start?path=`**
`handleStartCPUProfile` accepts an arbitrary `path` query parameter and passes it to `StartCPUProfile` → `os.Create(path)`. Any HTTP client that can reach the pprof port can write a CPU profile file to any filesystem path the mount process can reach. This is a file-write primitive.

Fix: either (a) reject absolute paths and resolve relative to `ProfileDir` only, or (b) remove the `path` parameter entirely and always use `defaultCPUProfilePath()`. The perf harness can be updated to read the path from the response instead of choosing it.

**B2 — `SyncAll` aborts on first handle flush failure, skipping remaining dirty handles**
Old `FlushAll` iterated ALL handles regardless of individual failures. New `SyncAll` returns on the first `flushHandle` error. `FlushAll` calls `SyncAll`, logs the error, then calls `DrainAll`. But dirty data in handles after the failed one was never flushed — those handles are skipped. If the uploader/commitQueue items depend on those flushes, data is silently lost on shutdown.

Fix: `SyncAll` should accumulate errors and continue flushing all handles, then return a combined error. The context-cancellation short-circuit is fine, but a single-handle FUSE error should not skip the rest.

**B3 — `CommitQueue.WaitAll` and `WriteBackUploader.WaitAll` busy-poll with no inherent bound**
Both poll `PendingStats()` every 50ms. If a commit or upload is stuck (retry loop, network partition), these spin until the context expires. When called from `FlushAll` → `SyncAll(context.Background())`, there is no timeout — this hangs shutdown indefinitely.

Fix: `FlushAll` should call `SyncAll` with a bounded context (e.g., the existing `releaseTimeout` logic or a configurable shutdown timeout). Alternatively, `WaitAll` should accept a hard deadline even when the parent context is Background.

**B4 — `ProcessState` stores raw `Server`, `MountPoint`, `RemoteRoot` in PID file**
The PID file is at a well-known path readable by any local user. Storing raw server URL and remote root exposes the server endpoint and workspace structure. The continuous perf samples correctly hash these fields via `shortHash`, but `ProcessState` uses raw strings. Inconsistent privacy model.

Fix: hash `Server`, `MountPoint`, and `RemoteRoot` in `ProcessState`, or document why the PID file has different privacy requirements than perf samples.

**B5 — `DirectMountStrict = true` for Linux root — undocumented, untested, unrelated to perf**
`newGoFuseMountOptions` now sets `DirectMountStrict = true` when `os.Geteuid() == 0`. This changes FUSE mount behavior in container/root environments. No design doc mentions it, no test covers it, and it's orthogonal to perf diagnostics. Behavior changes to the mount path should not be bundled into a perf PR without justification.

Fix: either (a) remove from this PR and submit separately with a rationale and test, or (b) add a comment explaining why it's needed for perf work specifically and add a test for the root-detection path.

---

**O1** — `runtime.ReadMemStats` is called every perf sample interval. This is a stop-the-world operation. At 1s+ intervals the cost is negligible, but worth a doc note in the overhead model section.

**O2** — `perfLatencyPercentile` returns the bucket upper bound, not an interpolated value. Reported percentiles are coarse-grained (e.g., 500µs → reports 512µs). Fine for diagnostics, but the design doc's "low-overhead histograms" section should clarify this is bucket-granularity, not exact.

**O3** — `run.sh` uses unquoted `$DRIVE9_MOUNT_EXTRA_FLAGS` (with `SC2206` disable). Intentional word-splitting, but any flag value containing spaces will break. Consider using an env-based array or documenting the limitation.

---

### Checklist

| Gate | Status |
|---|---|
| All inputs handled (including invalid)? | **NO** — B1 path traversal |
| Edge cases covered? | **NO** — B2 partial flush, B3 unbounded wait |
| Failure behavior defined? | **NO** — B2, B3 shutdown hang |
| Follows architecture-spec.md? | **NO** — B5 undocumented behavior change |
| API explicit (no hidden contracts)? | **NO** — B4 inconsistent privacy |
| Race / state issues? | OK |
| Regression tests included? | Partial — B5 has no test |

**Verdict: REVISE** — 5 blockers. B1 is security-sensitive. B2+B3 affect shutdown correctness. B4+B5 are consistency/process issues.

pkg/fuse/profiling.go
```diff
@@ -0,0 +1,298 @@
...
+	mux.HandleFunc("/debug/pprof/symbol", httppprof.Symbol)
+	mux.HandleFunc("/debug/pprof/trace", httppprof.Trace)
+	mux.HandleFunc("/debug/drive9/profile/cpu/start", p.handleStartCPUProfile)
+	mux.HandleFunc("/debug/drive9/profile/cpu/stop", p.handleStopCPUProfile)
+	mux.HandleFunc("/debug/drive9/mount/sync", p.handleMountSync)
+	return mux
+}
+
+func (p *Profiler) handleStartCPUProfile(w http.ResponseWriter, r *http.Request) {
+	path := r.URL.Query().Get("path")
```
qiffang: **B1 — Path traversal**: `path` comes directly from an HTTP query parameter and is passed to `os.Create`. Any client on the pprof port can write to arbitrary filesystem paths. Either resolve relative to `ProfileDir` only (rejecting absolute paths and `..` components), or remove the `path` parameter and always use `defaultCPUProfilePath()`.

pkg/fuse/dat9fs.go
```diff
@@ -4879,24 +4880,53 @@ func (fs *Dat9FS) FlushAll() {
...
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		default:
+		}
 		// Per-handle timeout scaled by file size so large uploads complete.
 		e.fh.Lock()
 		var sz int64
 		if e.fh.Dirty != nil {
 			sz = e.fh.Dirty.Size()
```
qiffang, line 11655: **B2 — First flush failure skips remaining handles**: `SyncAll` returns on the first `flushHandle` error. Old `FlushAll` continued through all handles. This means dirty data in handles after a failed one is silently skipped on shutdown. Accumulate errors and continue flushing all handles.

pkg/fuse/commit_queue.go
```diff
@@ -209,6 +209,35 @@ func (cq *CommitQueue) PendingStats() (count int, bytes int64) {
...
+func (cq *CommitQueue) WaitAll(ctx context.Context) error {
+	if cq == nil {
+		return nil
+	}
+	start := time.Now()
+	defer func() {
+		if cq.perf != nil {
+			cq.perf.commitDrainCount.add(1)
+			cq.perf.commitDrainTotalNS.add(uint64(time.Since(start)))
+		}
```
qiffang: **B3 — Unbounded busy-poll**: `WaitAll` polls every 50ms with no inherent bound. When called from `FlushAll` → `SyncAll(context.Background())`, there is no timeout — if a commit is stuck, shutdown hangs forever. Same issue in `WriteBackUploader.WaitAll`. `FlushAll` should pass a bounded context.

pkg/mountstate/pid.go
```diff
@@ -13,8 +13,21 @@ import (
...
 
 type ProcessState struct {
-	PID          int    `json:"pid"`
-	CreationTime uint64 `json:"creation_time,omitempty"`
+	PID                int    `json:"pid"`
+	CreationTime       uint64 `json:"creation_time,omitempty"`
+	Component          string `json:"component,omitempty"`
+	MountPoint         string `json:"mount_point,omitempty"`
+	RemoteRoot         string `json:"remote_root,omitempty"`
+	Server             string `json:"server,omitempty"`
```
qiffang: **B4 — Raw paths in PID file**: `ProcessState` stores raw `Server`, `MountPoint`, `RemoteRoot`. The PID file is at a well-known path (0644). Continuous perf samples correctly hash these via `shortHash`. Inconsistent privacy model — either hash here too, or document why PID file has different requirements.

qiffang (changes requested):
## PR #446 review — adversary-1: REQUEST_CHANGES

Reviewed head `76ab784`. CI is green, but I still see correctness/security blockers.

### Blockers

1. **Arbitrary file create/truncate through pprof CPU start endpoint** (`pkg/fuse/profiling.go:136-158`)

`/debug/drive9/profile/cpu/start?path=` is unauthenticated and passes the query value directly to `os.Create(path)`. Any process that can reach the pprof listener can create/truncate any path writable by the mount process. This is especially bad for root/container mounts. It also creates the file before checking whether another CPU profile is already running, so even a rejected second request can still truncate a path.

Fix: do not accept arbitrary paths from HTTP. Either remove the `path` parameter and always use `defaultCPUProfilePath()`, or restrict to a clean relative filename under `ProfileDir` and check the active-profile lock before creating the file.

2. **`SyncAll` can skip dirty handles after the first flush failure** (`pkg/fuse/dat9fs.go:4882-4900`, `4920-4923`)

`SyncAll` returns immediately when one open handle fails `flushHandle`. `FlushAll` then logs that error and proceeds to drain queues, but any later open dirty handles from the snapshot were never flushed into those queues. That is a shutdown data-loss path: one bad handle can prevent unrelated dirty handles from being uploaded.

Fix: for non-context-cancellation flush errors, keep iterating all handles and return a combined error after all flush attempts. Context cancellation can still short-circuit.

3. **Shutdown can hang indefinitely in `FlushAll`** (`pkg/fuse/dat9fs.go:4921`)

`FlushAll` calls `SyncAll(context.Background())`, and `SyncAll` waits for uploader/commit queues with no deadline. If a background upload is stuck in retry/network wait, unmount can block forever. The new `perf sync` path accepts a bounded context, but the shutdown path lost that safety.

Fix: call `SyncAll` from `FlushAll` with a bounded shutdown context, then proceed to `DrainAll`/cleanup and surface the timeout clearly.

4. **Perf harness can run workloads against an unmounted local directory** (`perf/mount/run.sh:159-168`)

The readiness loop breaks when it sees `drive9: mounted on ...`, but if that line never appears and the mount process is still alive after 10s, the script continues. The workload then writes to `$DRIVE9_MOUNTPOINT` as a normal local directory, producing invalid perf data that looks successful.

Fix: track whether the mounted line was observed. If not, print the mount log, stop the mount process, and exit non-zero before starting pprof/workload.

5. **PID/process state now stores raw local/remote paths and server URL in a 0644 temp file** (`pkg/mountstate/pid.go:16-30`, `61-70`)

The old PID file only exposed a PID. This PR stores `server`, `mount_point`, `remote_root`, profile paths, and pprof address in the same predictable `/tmp/drive9-mount-*.pid` file with `0644` permissions. That conflicts with the PR's own hashed/redacted perf sample model and leaks workspace structure/server endpoints to other local users.

Fix: make process-state files user-private (`0600`) and avoid raw sensitive fields unless they are strictly needed. At minimum, hash `Server`, `MountPoint`, and `RemoteRoot` the same way the continuous perf context does.

6. **Unrelated FUSE behavior change bundled into perf PR** (`pkg/fuse/mount.go:505-509`)

`DirectMountStrict = true` for Linux root changes the mount path for root/container environments, but the PR does not document why perf diagnostics require it and has no regression test around root/non-root option selection. This should not be hidden in a diagnostics/reporting PR.

Fix: remove it from this PR, or document the exact need and add a focused test for the root-detection behavior.

### Additional concerns

- `cmd/drive9/main.go:328` omits the supported `perf sync` subcommand from top-level usage.
- `perf/mount/run.sh:47` uses `${DRIVE9_PERF_JSONL:-...}`, so the README's documented `DRIVE9_PERF_JSONL=` disable path does not work; use `-` instead of `:-` if empty should be respected.
- `drive9 mount --mode=webdav --profile-cpu ...` and related profiling flags are silently ignored. Only `--perf-jsonl` is rejected for WebDAV.

### Validation

Passed locally:

```text
/usr/local/go/bin/go test ./pkg/fuse ./cmd/drive9/cli ./cmd/drive9 -count=1
bash -n perf/mount/run.sh perf/mount/scripts/summarize.sh perf/mount/workloads/*.sh
git diff --check 52429ea56ea640280d4ca44871b611980d2ef4ec...HEAD
```

qiffang (commented):
## dat9-dev1 Round 2 — PR #446 (head `76ab784a`) — **No code change detected**

Head SHA is still `76ab784a`, identical to Round 1. Round 1 blockers remain open:

**B1** — `handleStartCPUProfile` accepts arbitrary `path` query param → path traversal / arbitrary file create+truncate. Needs allowlist or restrict to `ProfileDir` subtree.

**B2** — `SyncAll` snapshots open handles at call time. Handles opened after the snapshot starts are not flushed. This is acceptable for the documented use case (explicit sync control between workload phases) but should be documented as a non-guarantee.

**B3** — `WaitAll` (both `CommitQueue` and `WriteBackUploader`) polls with a fixed 50ms ticker. Context deadline provides the upper bound, so this is bounded — but polling is wasteful vs a `sync.Cond` or channel signal. Non-blocking if documented as acceptable for v1 diagnostics.

**B5** — `DirectMountStrict = true` when `euid == 0` is unrelated to perf diagnostics. Should be split into a separate commit or PR.

**Status**: Waiting for updated code that addresses B1 (mandatory security fix) and B5 (scope separation). B2 and B3 can be documented as v1 limitations.

qiffang (changes requested):
Round 2 review: REQUEST_CHANGES.

I re-fetched PR #446 and the head is still `76ab784ab0db24c1b256e4619504ca0754dae616`, the same commit I reviewed previously. The local tests pass, but the correctness/security blockers below are still present.

Validation run:

- `/usr/local/go/bin/go test ./pkg/fuse ./cmd/drive9/cli ./cmd/drive9 -count=1`
- `bash -n perf/mount/run.sh perf/mount/workloads/*.sh perf/mount/scripts/summarize.sh`

### B1 (BLOCKER): pprof CPU start endpoint can create/truncate arbitrary files

`pkg/fuse/profiling.go:252` reads `path` directly from the HTTP query, then `StartCPUProfile` uses that path through `ensureParentDir` and `os.Create` (`pkg/fuse/profiling.go:136-151`, `228-235`). That means anyone who can reach the pprof control listener can make the mount process create parent directories and truncate/write any path writable by that process. This is especially risky when the mount is running as root or a service user.

The endpoint should not accept an arbitrary filesystem path. Restrict output to the configured `ProfileDir` (or a fixed filename), reject absolute / `..` paths after `filepath.Clean`, and add regression tests for rejected traversal/absolute paths.

### B2 (BLOCKER): `FlushAll` can skip dirty handles after the first flush error

`SyncAll` snapshots open handles and returns immediately on the first non-OK flush (`pkg/fuse/dat9fs.go:4882-4900`). `FlushAll` then only logs that error and continues shutdown/drain (`pkg/fuse/dat9fs.go:4920-4931`). A single bad handle can therefore prevent later dirty handles from ever being flushed during shutdown, which is exactly when this path is supposed to preserve data.

This needs to attempt every snapshotted handle, aggregate/report errors, and only then proceed to queue drain. Add a regression with two dirty handles where the first flush fails and the second still gets flushed.

### B3 (BLOCKER): shutdown uses an unbounded background sync

`FlushAll` calls `SyncAll(context.Background())` (`pkg/fuse/dat9fs.go:4920-4922`). `SyncAll` then waits on uploader and commit queue drain using that context (`pkg/fuse/dat9fs.go:4903-4912`), while `WriteBackUploader.WaitAll` and `CommitQueue.WaitAll` only exit when their supplied context is done or the queues reach zero (`pkg/fuse/writeback_uploader.go:164-187`, `pkg/fuse/commit_queue.go:215-237`). If a worker/RPC gets stuck, unmount/shutdown can hang indefinitely.

Use a bounded shutdown context (size-aware if needed) and surface timeout distinctly. Add a test that a non-draining uploader/queue causes `FlushAll`/sync shutdown to return instead of hanging.

### B4 (BLOCKER): mount process state leaks raw paths/server metadata in a world-readable temp file

`ProcessState` includes raw `mount_point`, `remote_root`, `server`, `profile_dir`, `perf_jsonl`, `pprof_addr`, CPU/heap profile paths, etc. (`pkg/mountstate/pid.go:16-30`), and `WriteProcessState` writes it through `writeFileAtomic(..., 0644)` under `os.TempDir()` (`pkg/mountstate/pid.go:61-70`). On shared machines this exposes operational paths and server metadata to other local users, and it also exposes the pprof control address.

Store this under a user-private runtime dir with `0700` parent + `0600` file, or persist only redacted/minimal state. Add a permission regression test.

### B5 (BLOCKER): perf runner can execute workloads against an unmounted local directory

`perf/mount/run.sh` waits up to ~10s for the “mounted on” log line, but if that line never appears and the process is still alive, it continues anyway (`perf/mount/run.sh:158-169`). The workload then writes into `$DRIVE9_MOUNTPOINT` as a plain local directory, producing invalid profiles and possibly making failures look like successful drive9 runs.

Track whether the ready marker was observed and fail if not. Better: after the log marker, also perform a real mount/probe check before starting the workload.

### Non-blocking

`pkg/fuse/mount.go:505-509` enables `DirectMountStrict` automatically for Linux root. That may be correct, but it is an unrelated behavioral change in a profiling PR and should be documented/tested or moved to a separate PR.

srstack (commented):
## Review: feat: drive9 mount --perf-dir/--perf-addr

### Overview

This PR adds `--perf-dir` as a single-flag profiling switch for FUSE mounts. When set, it enables: continuous JSONL performance sampling, periodic windowed CPU profiling, periodic heap profiling, a final heap profile on unmount, and an HTTP pprof listener. All artifacts land in the specified directory. The implementation spans three areas: CLI flag plumbing (`mount.go`), the continuous perf recorder (`continuous_perf.go`), and the profiler (`profiling.go`).

### What works well

**Flag design**: The "one switch, sensible defaults" pattern is clean. All advanced flags require `--perf-dir`, enforced by `validateMountPerfFlags`. The WebDAV rejection follows the existing pattern for FUSE-only options.

**`ContinuousPerfRecorder`**: The lifecycle is correct — `Start` writes a `"start"` sample, `Stop` is idempotent via `sync.Once`, and the `doneCh` close-on-nil-file fast path (when `PerfSamplesPath == ""`) is safe. File rotation preserves history count correctly.

**Latency histogram in `perf.go`**: The power-of-two bucket table covers 1µs–120s, which is an appropriate range for FUSE ops. The `atomicUint64.max` CAS loop is correct. `perfLatencyPercentile` uses ceiling division `(count*pct + 99) / 100` to avoid undershoot at low counts — correct.

**`pruneProfileFiles`**: Sorting lexicographically works because filenames embed ISO-8601 timestamps (`20060102-150405`). Keeping the newest (tail of sorted list) is correct.

**`pprof` HTTP mux**: Custom `/debug/drive9/profile/cpu/start|stop` endpoints in addition to the standard `/debug/pprof/` routes give a clean operational handle for on-demand CPU profiling.

**`PendingStats` on `WriteBackUploader`**: The `active` counter covers the gap between dequeue and `inflight` map insertion — `max(active, inflight)` is the right way to avoid undercounting in the window between dequeue and registration.

**Test coverage**: Each subsystem has dedicated tests. `busyWait` avoids `time.Sleep` races in profiling tests. The rotation and retention tests use `PerfMaxSamples=1`/`PerfMaxSampleFiles=3` to exercise the path without long intervals.

### Issues

**1. `validateMountPerfFlags` signature is unwieldy**

The function takes 16 parameters. Given that all flags are already local to `fsMountCmdWithBackground`, this could be a struct or structured arguments. Not blocking, but makes the function hard to read and extend.

**2. `StartContinuousPerf` applies defaults a second time**

`setDefaults()` in `MountOptions` already fills `PerfSampleInterval`, `PerfMaxSamples`, and `PerfMaxSampleFiles` when `PerfSamplesPath != ""`. `StartContinuousPerf` then repeats the same `<= 0` guards. This is harmless but creates two places where defaults live. A comment to clarify the intent ("defensive re-application for callers that bypass `setDefaults`") would help.

**3. `cpuLoop` starts a profile immediately then waits for the interval**

```go
p.captureCPUProfile(p.defaultCPUProfilePath(), p.opts.CPUProfileDuration)
ticker := time.NewTicker(p.opts.CPUProfileInterval)
```

This means the first profile is captured at startup, not after the first interval tick. For a 30s capture every 10m, the mount startup will block for 30s in `cpuLoop` (run in a goroutine, so it doesn't block `Mount`, but it does consume CPU for 30s immediately). This is likely intentional — getting an early baseline — but it should be documented, especially since `captureCPUProfile` itself blocks for `duration` via `time.NewTimer`.

**4. `writeSample("stop")` is called after `<-r.doneCh` but outside the loop goroutine's mutex**

In `Stop`:
```go
close(r.stopCh)
<-r.doneCh      // loop goroutine has exited
r.writeSample("stop")
```
`writeSample` acquires `r.mu`, which is safe since the loop goroutine is done. However, if `writeSample` triggers a rotation (i.e., `sampleCount >= PerfMaxSamples`), `rotateLocked` calls `openSamplesFile` which resets `r.sampleCount = 0`. This path is possible if the stop sample is the one that crosses the sample limit. In practice this is a one-sample edge case and the rotate+reopen is correct, but worth a note.

**5. `handleStartCPUProfile` uses `http.StatusConflict` (409) for "profile already running"**

A profile already running is a client-side precondition failure. `http.StatusConflict` is defensible but `http.StatusTooManyRequests` (429) or a custom 4xx would be more conventional for this case. Minor.

**6. `PerfInterval` stored as `string` in `ProcessState`**

```go
PerfInterval string `json:"perf_interval,omitempty"`
```
Set as `opts.Profiling.PerfSampleInterval.String()`. This is human-readable but not machine-parseable from the state file without re-parsing. Using `int64` (milliseconds or nanoseconds) would be more consistent with how other duration fields are stored in the JSON state. Minor.

**7. `MountPointHash`/`RemoteRootHash`/`ServerHash` in JSONL context**

These are SHA-256 prefix hashes intended to avoid leaking paths into profiling output. However, `shortHash` uses only the first 6 bytes (12 hex chars) of SHA-256. With a 48-bit prefix space, collision probability is low but non-negligible across many tenants. This is fine for correlation within a single mount session but should not be treated as a globally unique identifier.

### Minor observations

- `continuous_perf.go` imports `crypto/sha256` but only uses it in `shortHash`. If this file grows, consider moving `shortHash` to a shared utility.
- `heapLoop` uses `t.UTC().Format("20060102-150405")` (second precision) while `cpuLoop` uses `"20060102-150405.000000000"` (nanosecond precision). Using consistent precision in both would make glob patterns simpler.
- `PerfMaxProfileFiles` applies to both CPU and heap profiles as a shared limit per-type (the `pruneProfileFiles` pattern argument separates them). This is correct but the flag name could suggest it's a total limit. The flag help text already clarifies "per type", which is sufficient.

### Summary

The implementation is solid. The main item to address before merge is documenting the "immediate first CPU profile on startup" behavior (#3). The other points are minor style/clarity observations and don't block. LGTM with that noted.

qiffang (changes requested):
Review verdict: request changes before merge.

Finding:
- P1: `cmd/drive9/cli/mount.go` removes the existing `drive9 mount --perf-counters` CLI flag, but this repository still depends on that flag. The current flag registration only includes the new `--perf-dir` / `--perf-*` / `--perf-addr` options, so `drive9 mount --perf-counters` now fails with `flag provided but not defined: -perf-counters`. Existing e2e scripts still invoke it at `e2e/fuse-write-perf-budget-test.sh:151`, `e2e/git-ops-smoke-test.sh:348`, and `e2e/git-workspace-smoke-test.sh:269`, so merging this breaks existing perf/smoke workflows. The internal plumbing still exists (`MountOptions.PerfCounters`, the fuse bridge field, and the `newFusePerfCounters(opts.PerfCounters || ...)` path), so the smallest fix is likely to restore and wire `--perf-counters`. If the intent is to replace it with `--perf-dir`, please migrate the e2e scripts/comments and explicitly remove/deprecate the old internal plumbing.

Validation performed:
- PR branch `9ec2150`: `go test ./cmd/drive9/cli ./pkg/fuse ./pkg/mountstate -count=1` PASS
- PR branch `9ec2150`: `go test ./cmd/drive9 -count=1` PASS
- PR branch `9ec2150`: `git diff --check 6656c253...HEAD` PASS
- Temporary merge over current `origin/main`: `go test ./cmd/drive9 ./cmd/drive9/cli ./pkg/fuse ./pkg/mountstate -count=1` PASS
- Temporary merge over current `origin/main`: `git diff --check` PASS

Non-blocking notes:
- `pkg/fuse/profiling.go` creates the CPU profile file before checking whether another CPU profile is already running, so a second start attempt can leave an empty timestamped file before returning conflict.
- The continuous CPU profiler appears to write the first CPU profile immediately before waiting for the interval; that behavior should be documented if intentional.

mornyx:
@qiffang @srstack I pushed `a874aad` with the review fixes.

Fix summary:
- Migrated the remaining e2e callers from the removed `drive9 mount --perf-counters` flag to `--perf-dir` with low-cost profiling settings. This covers `e2e/fuse-write-perf-budget-test.sh`, `e2e/git-ops-smoke-test.sh`, and `e2e/git-workspace-smoke-test.sh`, and removes all remaining `--perf-counters` references from the repo.
- Fixed the CPU profile conflict path: `StartCPUProfile` now checks for an already-running CPU profile before creating the output file, so a rejected second start no longer leaves an empty `cpu-*.pprof`. Added `TestStartCPUProfileConflictDoesNotCreateExtraFile`.
- Documented the intentional immediate first CPU profiling window in the CLI help text and in `cpuLoop`.
- Refactored `validateMountPerfFlags` to take a structured validation object instead of a long parameter list.
- Added comments explaining the defensive default re-application in `StartContinuousPerf` and the intentional stop-sample rotation behavior.

Notes on older qiffang blockers from the earlier, larger scope: the current PR no longer includes the top-level `drive9 perf` command, `SyncAll`/`MountSync`, `perf/mount` harness, or `DirectMountStrict` behavior change. `WriteProcessState` now writes JSON process state with `0600` permissions.

Validation:
- `env GOCACHE=/private/tmp/drive9-feat-perf-srstack-446/.codex-gocache go test ./cmd/drive9 ./cmd/drive9/cli ./pkg/fuse ./pkg/mountstate`
- `bash -n e2e/git-ops-smoke-test.sh e2e/git-workspace-smoke-test.sh e2e/fuse-write-perf-budget-test.sh`
- `git diff --check`

The local-e2e failure should be addressed by the e2e migration away from `--perf-counters`; the pushed PR branch should now let CI validate the full flow.

srstack (commented):
## Re-review

All six issues from the previous review have been addressed. Going through them:

**1. `validateMountPerfFlags` 16-parameter signature — resolved**

Refactored to accept a `mountPerfFlagValidation` struct. The call site is now explicit and readable.

**2. Double defaults in `StartContinuousPerf` — resolved**

A comment now explicitly states the intent: "Re-apply defaults defensively for direct package callers that construct `ProfilingOptions` without going through `MountOptions.setDefaults`." Clears the ambiguity.

**3. Immediate CPU profile at startup — resolved**

`cpuLoop` now has a comment:
> "Capture one startup window before the first interval tick so support bundles include the mount initialization baseline."

The flag help text also documents this: "one capture starts immediately when profiling begins". Intent is clear.

**4. `writeSample("stop")` rotation edge case — resolved**

The comment in `Stop()` now reads:
> "Route the final sample through writeSample so it follows the same segment-size and rotation rules as interval samples."

This makes the intentional behavior explicit.

**5. `PerfInterval` stored as string in `ProcessState` — acknowledged, not changed**

`PerfInterval string` remains in `ProcessState`. This is acceptable since `PerfIntervalMS int64` is already present in the JSONL context field, so machine-parseable form exists in the sample data. The state file value is for human inspection.

**6. Timestamp precision inconsistency (heap vs CPU) — unchanged**

CPU uses `"20060102-150405.000000000"` (nanosecond), heap uses `"20060102-150405"` (second). This is still present. Not a blocker — the separate glob patterns (`cpu-*.pprof` vs `heap-[0-9]*.pprof`) already segregate them, and the test in `TestPruneProfileFilesKeepsNewestPerPattern` uses both formats correctly. Worth noting as a future cleanup if the formats ever need to be unified.

### New in this revision

**`TestStartCPUProfileConflictDoesNotCreateExtraFile`** — new test that verifies a second concurrent `StartCPUProfile` call fails without creating a stale file. This directly covers the lock guard added when `startCPUProfile` was reorganized to check `cpuFile != nil` before creating the file (compared to the first revision which created the file before acquiring the lock). Good catch.

**e2e scripts migrated from `--perf-counters` to `--perf-dir`** — `fuse-write-perf-budget-test.sh`, `git-ops-smoke-test.sh`, and `git-workspace-smoke-test.sh` now use `--perf-dir` with minimal intervals (`--perf-cpu-duration 1ms`, `--perf-interval 1h`) so they exercise the new path without adding latency to CI runs. The `--perf-max-sample-files 1` / `--perf-max-profile-files 1` limits keep the artifact footprint small.

### Summary

All review issues resolved. LGTM.

qiffang (dismissed):
dev-2 review on updated head `a874aad`. Lane: Go-side concurrency / error recovery / comment-impl mismatch / boundary cases.

## R3 findings — all fixed

- ✅ **`--perf-counters` compatibility** (was dev-1 R2 P1). e2e callers in `fuse-write-perf-budget-test.sh:151`, `git-ops-smoke-test.sh:348`, `git-workspace-smoke-test.sh:269` are all migrated to `--perf-dir` with tight retention/duration knobs. `grep -rn perf-counters e2e/` returns 0 hits.
- ✅ **profiling.go CPU file ordering**. `p.mu.Lock()` and conflict check now precede `os.Create` (`pkg/fuse/profiling.go:171-178`). Regression test `TestStartCPUProfileConflictDoesNotCreateExtraFile` locks the fix.
- ✅ **cpuLoop immediate first capture**. Comment added (`pkg/fuse/profiling.go:221-222`) explaining the intentional startup-window-before-first-tick behavior so the support bundle includes mount initialization.

## Verification

Locally on PR head `a874aad`:
- `go build ./...` — PASS
- `go vet ./pkg/fuse/... ./cmd/drive9/...` — PASS
- `go test ./pkg/fuse/... -run "Profil|Perf|Continuous" -count=1` — PASS

## 1 follow-up nit (non-blocking)

`rotateSampleFilesLocked` (`pkg/fuse/continuous_perf.go:257-285`) is not atomic on partial failure:

```go
for i := historyCount - 1; i >= 1; i-- {
    src := r.sampleFilePath(i)
    dst := r.sampleFilePath(i + 1)
    if err := os.Rename(src, dst); err != nil && !os.IsNotExist(err) {
        return fmt.Errorf("rotate continuous perf samples %s to %s: %w", src, dst, err)
    }
}
previous := r.sampleFilePath(1)
os.Rename(r.opts.PerfSamplesPath, previous)
```

If a mid-loop `os.Rename` fails (e.g. transient permission flake), the function early-returns with the error, but the rename steps that already succeeded leave the retained-file slots in an inconsistent state — base file hasn't rotated yet, but some `.N` files have been shifted up. The next rotation call re-removes the oldest and shifts again, which can lose additional sample history without logging the cascading state.

Severity: low. Rotation is best-effort observability and real-world rename failure is rare on local filesystems. Either swallow the per-step error (next rotation will self-heal), defer a rollback of completed renames on failure, or document this as known best-effort behavior in the PR description.

## Verdict

**Approve** — R3 P1 + non-blocking notes all addressed; no remaining blocker from my lane.

+1 the BREAKING CHANGELOG note suggested by adversary-2: `--perf-counters` is a user-facing CLI flag removal, even though `MountOptions.PerfCounters` is still used internally by tests. A migration note in CHANGELOG would prevent silent breakage for users that aliased `--perf-counters` in scripts.

qiffang:
dev-1 re-review on updated head `a874aad`: no blocking findings from me.

What I re-checked:
- The prior P1 e2e regression is resolved in the merge result: the three e2e scripts no longer invoke `drive9 mount --perf-counters`; they now use `--perf-dir` with long intervals / low retention for deterministic perf summaries.
- On the temporary merge over current `origin/main`, there are no user-facing `--perf-counters` references left in `README.md` or `e2e/`; only internal `PerfCounters` plumbing/tests remain.
- The CPU profile double-start issue is fixed: the in-progress check happens before `os.Create`, and `TestStartCPUProfileConflictDoesNotCreateExtraFile` covers it.
- The immediate startup CPU sample behavior is now documented in code and in the `--perf-cpu-interval` help text.

Validation I ran:
- PR branch `a874aad`: `go test ./cmd/drive9 ./cmd/drive9/cli ./pkg/fuse ./pkg/mountstate -count=1` PASS
- PR branch `a874aad`: `git diff --check $(git merge-base origin/main HEAD)..HEAD` PASS
- Temporary merge over current `origin/main`: `go test ./cmd/drive9 ./cmd/drive9/cli ./pkg/fuse ./pkg/mountstate -count=1` PASS
- Temporary merge over current `origin/main`: `git diff --check` PASS

Residual note, non-blocking: `--perf-counters` is now a CLI breaking change. Since the merge result removes stale README/e2e usage, I do not think this blocks the PR, but the release notes / PR description should call out the migration path: use `--perf-dir <dir>` instead of `--perf-counters`.

LGTM from dev-1. I am posting this as a PR comment rather than changing the GitHub review state because this repo appears to use shared GitHub credentials for agent reviews.

qiffang (approved):

PR #455, opened 2026-05-22
sdk: add swift/kotlin SDK for drive9

you06:
## Summary by CodeRabbit

* **New Features**
  * Native Kotlin and Swift Drive9 SDKs with full filesystem, streaming upload/download, patching, resume, and vault APIs.
  * GitHub Actions CI to run SDK tests across Python/JS/Kotlin/Swift.

* **Documentation**
  * Added/updated Kotlin and Swift READMEs describing usage and parity notes.

* **Tests**
  * End-to-end SDK test suites and a Swift mock HTTP server added.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mem9-ai/drive9/pull/455?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/clients/drive9-rs/src/stream.rs
+++ b/clients/drive9-rs/src/stream.rs
@@ -2,6 +2,7 @@ use crate::client::Client;
 use crate::error::Drive9Error;
 use crate::models::{CompletePart, UploadPlanV2};
 use std::collections::HashMap;
+use std::sync::Arc;
 use tokio::sync::{Mutex, Semaphore};
 
 const UPLOAD_MAX_CONCURRENCY: usize = 16;
@@ -11,8 +12,8 @@ pub struct StreamWriter {
     path: String,
     total_size: i64,
     expected_revision: i64,
-    state: std::sync::Arc<Mutex<StreamState>>,
-    sem: Semaphore,
+    state: Arc<Mutex<StreamState>>,
+    sem: Arc<Semaphore>,
 }
 
 struct StreamState {
@@ -38,7 +39,7 @@ impl StreamWriter {
             path,
             total_size,
             expected_revision,
-            state: std::sync::Arc::new(Mutex::new(StreamState {
+            state: Arc::new(Mutex::new(StreamState {
                 plan: None,
                 uploaded: HashMap::new(),
                 inflight: 0,
@@ -48,7 +49,7 @@ impl StreamWriter {
                 aborted: false,
                 closing: false,
             })),
-            sem: Semaphore::new(UPLOAD_MAX_CONCURRENCY),
+            sem: Arc::new(Semaphore::new(UPLOAD_MAX_CONCURRENCY)),
         }
     }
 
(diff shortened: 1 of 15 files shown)


clients/drive9-swift/Package.swift
```diff
@@ -0,0 +1,20 @@
+// swift-tools-version: 5.9
+import PackageDescription
+
+let package = Package(
```
JaySon-Huang, line 4: This package currently exposes async Swift APIs (`AsyncSequence`, `URLSession.data(for:)`, and `URLSession.upload(for:from:)`), but the manifest does not declare a platform floor. On a local macOS run, `swift test` fails at compile time with availability errors (`URLSession` async APIs require macOS 12+). Please either declare explicit supported platforms here, for example macOS 12 / the matching iOS floor, or add availability-gated fallbacks so the documented `swift test` command builds on supported targets.

Source: rule/HONGYUNYAN-RB-001 | [Second Opinion](https://github.com/PingCAP-QE/second-opinion)

clients/drive9-rs/src/stream.rs
```diff
@@ -185,16 +295,52 @@ impl StreamWriter {
...
+                    ))
+                } else {
+                    None
+                }
+            } else {
+                None
+            }
+        };
+        if let Some((upload_id, msg)) = invalid_final_part {
+            let _ = self.client.abort_upload_v2(&upload_id).await;
```
JaySon-Huang: This abort only covers the out-of-range final-part validation above. After this point, `complete()` still returns final-part presign/upload errors and `complete_upload_v2` errors directly, unlike `write_stream_v2` / `write_stream_v2_with_hooks`, which abort the v2 upload on finalize failure. Please mirror that cleanup path for every post-initiate failure and add a test where `/complete` returns 500 and `/abort` is expected once.

Source: rule/XUHUAIYU-PR-001 | [Second Opinion](https://github.com/PingCAP-QE/second-opinion)

clients/drive9-swift/Package.swift
you06: Declared:

```
    platforms: [
        .iOS(.v17),
        .macOS(.v12),
    ],
```

clients/drive9-swift/Tests/Drive9MobileTests/MockHTTPServer.swift
```diff
@@ -0,0 +1,233 @@
...
+    /// Register a handler that matches any query string for the given method
+    /// and path. The handler can inspect `Request.query` to assert specifics.
+    func routeAnyQuery(_ method: String, _ pathOnly: String, handler: @escaping Handler) {
+        lock.lock()
+        pathOnlyRoutes["\(method) \(pathOnly)"] = handler
+        lock.unlock()
+    }
+
+    func start() throws {
+        let sock = socket(AF_INET, Int32(SOCK_STREAM.rawValue), 0)
```
JaySon-Huang: The Swift package platform floor is fixed, but `swift test` still does not compile on Darwin/macOS because `SOCK_STREAM` is already imported as `Int32` here, so it has no `.rawValue`. This is a Darwin vs Glibc import difference: Darwin gets `SOCK_STREAM` from the C macro `#define SOCK_STREAM 1`, while Linux/Glibc may need the enum-style `.rawValue` form. Please make this platform-aware, for example use `Int32(SOCK_STREAM.rawValue)` under `canImport(Glibc)` and plain `SOCK_STREAM` under Darwin, then rerun `swift test`. I reproduced the failure locally with Swift 6.2.3 / macOS 26.3.1 / SDK 26.2: `MockHTTPServer.swift:58:54: error: value of type 'Int32' has no member 'rawValue'`.

Source: rule/HONGYUNYAN-RB-001 | [Second Opinion](https://github.com/PingCAP-QE/second-opinion)

you06: Fixed

JaySon-Huang (approved):

PR #458, opened 2026-05-24
feat: add symlink support

mornyx:
## Summary
- Add backend/server/client support for symbolic links stored as inline target payloads with POSIX symlink mode bits.
- Implement FUSE Symlink/Readlink handling and CLI `drive9 fs symlink`.
- Add unit coverage plus CLI/FUSE smoke-test symlink cases.

## Tests
- `env GOCACHE=/Users/mornyx/Desktop/repos/github.com/mem9-ai/drive9/.codex-gocache go test ./cmd/drive9 ./cmd/drive9/cli ./pkg/fuse`
- `bash -n e2e/fuse-smoke-test.sh`
- `git diff --check`

Live e2e scripts were updated but not run locally because they require a live drive9 server/FUSE mount environment.

## Summary by CodeRabbit

* **New Features**
  * Added symbolic link support end-to-end: CLI command to create symlinks, server/backend handling, client API, and filesystem-level create/readlink/resolve behavior.

* **Bug Fixes**
  * Chmod now preserves file-type bits when updating permissions.

* **Tests**
  * Added extensive unit and integration tests covering symlink creation, readlink behavior, error cases, authorization, and recursive-copy rejection of symlinks.

* **Documentation / E2E**
  * Updated smoke tests and E2E docs to include symlink scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mem9-ai/drive9/pull/458?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/cmd/drive9/cli/cp_recursive.go
+++ b/cmd/drive9/cli/cp_recursive.go
@@ -10,6 +10,7 @@ import (
 	"sort"
 	"strings"
 	"sync"
+	"syscall"
 
 	"github.com/mem9-ai/dat9/pkg/client"
 )
@@ -184,6 +185,9 @@ func copyTreeRemoteToLocal(ctx context.Context, c *client.Client, srcRemote, dst
 		files []remoteFileEntry // relative paths to download
 	)
 	walkErr := walkRemoteTreeBFS(ctx, c, srcRemote, func(rel string, info client.FileInfo) error {
+		if remoteInfoIsSymlink(info) {
+			return fmt.Errorf("recursive copy does not support symlinks yet (path: %s)", rel)
+		}
 		if info.IsDir {
 			dirs = append(dirs, remoteDirEntry{rel: rel})
 			return nil
@@ -288,6 +292,9 @@ func copyTreeRemoteToRemote(ctx context.Context, c *client.Client, srcRemote, ds
 		files []remoteCopyPlanItem // src+dst pairs
 	)
 	walkErr := walkRemoteTreeBFS(ctx, c, srcRemote, func(rel string, info client.FileInfo) error {
+		if remoteInfoIsSymlink(info) {
+			return fmt.Errorf("recursive copy does not support symlinks yet (path: %s)", rel)
+		}
 		dstPath, joinErr := joinRemoteSafe(dstRemote, rel)
 		if joinErr != nil {
 			return joinErr
@@ -571,6 +578,10 @@ func walkRemoteTreeBFS(ctx context.Context, c *client.Client, root string, visit
 		}
 	}
 	return nil
+}
+
+func remoteInfoIsSymlink(info client.FileInfo) bool {
+	return info.HasMode && info.Mode&uint32(syscall.S_IFMT) == uint32(syscall.S_IFLNK)
 }
 
 // uploadOneFile streams a single local file to the remote destination.
--- a/cmd/drive9/cli/cp_recursive_test.go
+++ b/cmd/drive9/cli/cp_recursive_test.go
@@ -394,13 +394,16 @@ func TestCpRecursiveRemoteToLocal_RejectsSymlinkInTree(t *testing.T) {
 	c := client.New(srv.URL, "")
 	err := Cp(c, []string{"-r", ":/src", dstLocal})
 	if err == nil {
-		t.Fatal("expected remote symlink reject error, got nil")
+		t.Error("expected remote symlink reject error, got nil")
+		return
 	}
 	if !strings.Contains(err.Error(), "symlink") {
-		t.Fatalf("error = %v, want 'symlink' message", err)
+		t.Errorf("error = %v, want 'symlink' message", err)
+		return
 	}
 	if _, statErr := os.Lstat(dstLocal); !errors.Is(statErr, os.ErrNotExist) {
-		t.Fatalf("local destination should not be created after symlink reject, stat err=%v", statErr)
+		t.Errorf("local destination should not be created after symlink reject, stat err=%v", statErr)
+		return
 	}
 }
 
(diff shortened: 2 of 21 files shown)


qiffang (changes requested):
## Review: REVISE

### Blockers

1. `drive9 fs cp -r :/remote ./local` silently converts remote symlinks into regular files containing the link target text.

   In `copyTreeRemoteToLocal`, every non-directory entry returned by `ListCtx` is appended to `files` and later downloaded via `DownloadToFileWithSummary` (`cmd/drive9/cli/cp_recursive.go:186-195`, `:267-269`). With this PR, symlinks are represented as non-directory entries with `HasMode=true` and `S_IFLNK` mode, and GET returns the stored target payload. That means a remote tree containing `link -> ../target.txt` downloads as a normal local file whose contents are `../target.txt`.

   This violates the existing recursive-copy contract in this same file: "Symlinks: explicit reject, not silent follow/skip" (`cmd/drive9/cli/cp_recursive.go:31-32`). It is also data-lossy: a backup/restore or remote-to-local sync no longer preserves the filesystem tree shape. Fix by detecting remote symlink mode during the remote walk and either rejecting it explicitly until preservation is implemented, or preserving it by creating a local symlink with regression coverage.

2. The symlink endpoint accepts unbounded JSON and then stores the target inline in the DB.

   `handleSymlink` decodes directly from `r.Body` without `http.MaxBytesReader` (`pkg/server/server.go:1999`), and `CreateSymlinkCtx` always stores the decoded target as `StorageDB9` / `ContentBlob` (`pkg/backend/dat9.go:265-282`). The only size check is the generic upload limit (`ensureUploadSizeAllowed`), which can be much larger than an acceptable symlink target and is not tied to the inline threshold. A client can send a huge JSON target and force the server to allocate it during decode and then attempt to persist it as an inline DB blob.

   Symlink targets should have a small explicit maximum (PATH_MAX-style, e.g. 4 KiB or another documented constant), the HTTP body should be bounded with `MaxBytesReader`, and tests should cover oversize target rejection before any DB insert. While fixing this, map `ErrUploadTooLarge` / storage quota errors in `handleSymlink`; right now anything other than invalid target or path conflict falls through to 500 (`pkg/server/server.go:2004-2017`).

### Validation

- Passed locally: `go test ./pkg/fuse ./cmd/drive9/cli ./cmd/drive9 -count=1`
- Passed locally: `bash -n e2e/cli-smoke-test.sh e2e/fuse-smoke-test.sh`
- Passed locally: `git diff --check origin/main...HEAD`
- Could not run DB-backed packages locally because this machine has no rootless Docker; `go test` for `pkg/backend`, `pkg/client`, `pkg/datastore`, and `pkg/server` panics in `testcontainers` with `rootless Docker not found`. GitHub checks are green, but the blockers above are semantic/contract issues not caught by the current tests.

mornyx:
@qiffang @coderabbitai I have addressed this round of review feedback and pushed the fixes to `feat/symlink`. The latest commit is `58f238a fix: address symlink review feedback`.

For @qiffang's blockers:
- `drive9 fs cp -r :/remote ./local` no longer silently converts remote symlinks into regular files. Remote recursive walks now detect `HasMode && S_IFLNK`; remote-to-local and remote-to-remote recursive copies explicitly reject symlinks until preservation is implemented. I added CLI recursive-copy regression coverage for both paths.
- The symlink endpoint and target size are now bounded. `backend.MaxSymlinkTargetBytes` caps targets at 4 KiB, and `handleSymlink` uses `http.MaxBytesReader` for the JSON body. Oversized targets are rejected before DB insert with 400, oversized request bodies return 413, `ErrUploadTooLarge` maps to 413, and `ErrStorageQuotaExceeded` maps to 507. I also added oversize body/target and quota tests.

For @coderabbitai's actionable and nitpick comments:
- The CLI e2e symlink `ls /` assertion now uses bounded polling to avoid propagation-lag flakiness.
- FUSE `Readlink` now uses `readSmallFileWithRetry`; I added a transient read retry test.
- FUSE `Symlink` now performs a detached probe after transient/cancel-style errors and treats the operation as successful if the created path exists; I added coverage for that recovery path.
- `handleSymlink` is now included in the handler-level auth-ordering test.
- CLI symlink unit tests now assert `symlink=1` explicitly.

Validation passed:
- `go test ./cmd/drive9/cli -run "Test(Symlink|CpRecursive.*Symlink)"`
- `go test ./pkg/fuse -run "Test(ReadlinkRetriesTransientRead|SymlinkRecoversWhenInterruptedAfterRemoteCreate|LookupRecognizesRemoteSymlinkMode|SymlinkCreatesRemoteLinkAndCachesEntry)" -timeout 30s`
- `make test TEST_PKGS="./pkg/server" TEST_RUN="Test(Symlink|C2aWriteHandlersRejectScopedRequestBeforeBackend|C2aDispatcherWriteSideAllowed)"`
- `bash -n e2e/fuse-smoke-test.sh`
- `git diff --check`

mornyx:
@coderabbitai Addressed the latest follow-up review in `ff05cbf fix: harden symlink review follow-up`.

Fixes included:
- FUSE `Symlink` recovery now performs a detached `StatCtx` probe and only clears the transient creation error when the recovered object has symlink mode. If the path exists but is a regular file/other object, the original retryable error is preserved. Added `TestSymlinkRecoveryRejectsNonSymlinkProbe` for this case.
- `maxSymlinkBodyBytes` now accounts for worst-case JSON escaping: `backend.MaxSymlinkTargetBytes*6 + 64`. Added a regression test showing a max-sized target with `\u00XX` escaping is accepted instead of being rejected early by `MaxBytesReader`.
- Updated the new recursive-copy and FUSE tests called out in the nitpicks to use `t.Errorf` plus `return` for assertion failures.

Validation passed:
- `go test ./cmd/drive9/cli -run "TestCpRecursive.*Symlink"`
- `go test ./pkg/fuse -run "Test(ReadlinkRetriesTransientRead|SymlinkRecoversWhenInterruptedAfterRemoteCreate|SymlinkRecoveryRejectsNonSymlinkProbe)" -timeout 30s`
- `make test TEST_PKGS="./pkg/server" TEST_RUN="TestSymlink"`
- `git diff --check`

qiffang (approved):
## Review: GREEN

Round 2 re-review at head `ff05cbf0dbeb87aa006632970db67f331939a32f`.

The two previous blockers are fixed:

- Remote recursive copy now rejects remote symlink entries before local/remote side effects. `copyTreeRemoteToLocal` and `copyTreeRemoteToRemote` both check `HasMode && S_IFLNK`, with regression tests verifying no local destination / no remote copy or mkdir is created.
- `POST /v1/fs/<path>?symlink=1` is now bounded and maps the expected failure modes correctly: decoded target length is capped by `backend.MaxSymlinkTargetBytes`, request body is protected with `http.MaxBytesReader`, oversize target/body are tested separately, storage quota maps to 507, upload-size errors map to 413, and path conflict maps to 409.

I also verified the latest follow-up fixes:

- `maxSymlinkBodyBytes = MaxSymlinkTargetBytes*6 + 64` covers worst-case JSON escaping for a valid max-length target; `TestSymlinkAllowsMaxTargetWithWorstCaseJSONEscaping` protects this.
- FUSE symlink interrupted/cancelled recovery now probes with `Stat` and only recovers when the remote object is actually a symlink; `TestSymlinkRecoveryRejectsNonSymlinkProbe` covers the stale/non-symlink race.
- `chmod` preserves file-type bits, so symlink mode does not get downgraded into a regular-file mode.

Validation run locally:

- `go test ./pkg/fuse ./cmd/drive9/cli ./cmd/drive9 -count=1`
- `bash -n e2e/cli-smoke-test.sh e2e/fuse-smoke-test.sh`

DB-backed focused tests still cannot run in this local environment because testcontainers panics with `rootless Docker not found`, but GitHub `ci` and `drive9-server-local e2e smoke` are both green.

No remaining blockers from me.

PR #460, opened 2026-05-25
fix(fuse): preserve pending create modes

mornyx:
## Summary
- preserve file mode from FUSE create requests in local inode state and pending writeback metadata
- propagate pending modes through rename/recovery and apply chmod after async commit uploads
- add regression coverage for executable create mode, pending lookup, flush metadata, and commit queue chmod

## Tests
- env GOCACHE=/Users/mornyx/Desktop/repos/github.com/mem9-ai/drive9/.codex-gocache go test ./pkg/fuse
- env GOCACHE=/Users/mornyx/Desktop/repos/github.com/mem9-ai/drive9/.codex-gocache go test ./cmd/drive9/cli

## Summary by CodeRabbit

* **New Features**
  * Deferred file-mode staging: creation modes are persisted and applied after successful uploads; default-mode creations skip post-upload chmod.
  * Pending-mode metadata is recorded across staging, writeback, release and commit pathways so mode state survives restarts.

* **Bug Fixes**
  * Post-upload chmod failures now surface and halt success-side cleanup; pending state is retained and retries occur without reuploading data.
  * Journal/conflict paths preserve pending mode on post-upload failures.

* **Tests**
  * Expanded tests covering mode propagation, deferred-chmod retry/404 behavior, error retention, and concurrency.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mem9-ai/drive9/pull/460?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/pkg/fuse/commit_queue.go
+++ b/pkg/fuse/commit_queue.go
@@ -591,9 +591,9 @@ func (cq *CommitQueue) commitOne(entry *CommitEntry) {
 			if err := cq.onCommitSuccess(entry, committedRev); err == nil {
 				return
 			} else {
-				lastErr = err
 				log.Printf("commit queue: post-upload attempt %d/%d failed for %s: %v", attempt+1, maxRetries, entry.Path, err)
-				continue
+				cq.onCommitPostUploadFailure(entry, err)
+				return
 			}
 		}
 		if cq.isEntryCanceled(entry) {
@@ -772,16 +772,21 @@ func (cq *CommitQueue) rebuildQueuedIndexLocked() {
 }
 
 func (cq *CommitQueue) onCommitSuccess(entry *CommitEntry, committedRev int64) error {
-	if entry.HasMode {
+	if shouldApplyRemoteMode(entry.Kind, entry.HasMode, entry.Mode) {
 		ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
-		start := time.Now()
-		err := cq.client.ChmodCtx(ctx, cq.remotePath(entry.Path), entry.Mode&0o777)
+		var err error
+		mode := entry.Mode & 0o777
+		err = retryPostUploadMode(ctx, func() error {
+			start := time.Now()
+			applyErr := cq.client.ChmodCtx(ctx, cq.remotePath(entry.Path), mode)
+			if cq.perf != nil {
+				cq.perf.recordRemoteOp(perfRemoteMutation, applyErr, time.Since(start), 0)
+			}
+			return applyErr
+		})
 		cancel()
-		if cq.perf != nil {
-			cq.perf.recordRemoteOp(perfRemoteMutation, err, time.Since(start), 0)
-		}
 		if err != nil {
-			return fmt.Errorf("%w: chmod %s to %o: %v", errCommitPostUpload, entry.Path, entry.Mode&0o777, err)
+			return fmt.Errorf("%w: chmod %s to %o: %w", errCommitPostUpload, entry.Path, mode, err)
 		}
 	}
 
--- a/pkg/fuse/commit_queue.go
+++ b/pkg/fuse/commit_queue.go
@@ -44,6 +44,8 @@ type CommitEntry struct {
 	Size         int64
 	Kind         PendingKind
 	ShadowSpill  bool // true when data is only in shadow file (auto-resolve would OOM)
+	Mode         uint32
+	HasMode      bool
 	canceled     bool
 	cancelCommit context.CancelFunc
 	cancelUpload context.CancelFunc
@@ -269,6 +271,8 @@ func (cq *CommitQueue) RecoverPending() {
 			Size:        meta.Size,
 			Kind:        meta.Kind,
 			ShadowSpill: meta.ShadowSpill,
+			Mode:        meta.Mode,
+			HasMode:     meta.HasMode,
 		}
 		if err := cq.Enqueue(entry); err != nil {
 			log.Printf("commit queue: recover enqueue failed for %s: %v", path, err)
@@ -753,6 +757,19 @@ func (cq *CommitQueue) rebuildQueuedIndexLocked() {
 }
 
 func (cq *CommitQueue) onCommitSuccess(entry *CommitEntry, committedRev int64) {
+	if entry.HasMode {
+		ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
+		start := time.Now()
+		err := cq.client.ChmodCtx(ctx, cq.remotePath(entry.Path), entry.Mode&0o777)
+		cancel()
+		if cq.perf != nil {
+			cq.perf.recordRemoteOp(perfRemoteMutation, err, time.Since(start), 0)
+		}
+		if err != nil {
+			log.Printf("commit queue: chmod %s to %o failed after upload: %v", entry.Path, entry.Mode&0o777, err)
+		}
+	}
+
 	// Write durable commit record BEFORE cleaning up local state so that
 	// crash recovery never re-uploads an already committed entry.
 	if cq.journal != nil {
--- a/pkg/fuse/commit_queue.go
+++ b/pkg/fuse/commit_queue.go
@@ -15,6 +15,8 @@ import (
 	"github.com/mem9-ai/dat9/pkg/mountpath"
 )
 
+var errCommitPostUpload = errors.New("commit post-upload step failed")
+
 // directPutThreshold returns the size limit below which commit queue workers
 // use direct PUT (WriteCtxConditionalWithRevision) instead of multipart
 // upload. Must match the server's inline_threshold — the server rejects
@@ -586,9 +588,13 @@ func (cq *CommitQueue) commitOne(entry *CommitEntry) {
 		cancel()
 
 		if err == nil {
-			// Success — clean up.
-			cq.onCommitSuccess(entry, committedRev)
-			return
+			if err := cq.onCommitSuccess(entry, committedRev); err == nil {
+				return
+			} else {
+				lastErr = err
+				log.Printf("commit queue: post-upload attempt %d/%d failed for %s: %v", attempt+1, maxRetries, entry.Path, err)
+				continue
+			}
 		}
 		if cq.isEntryCanceled(entry) {
 			cq.removeFromQueue(entry)
@@ -605,6 +611,10 @@ func (cq *CommitQueue) commitOne(entry *CommitEntry) {
 	}
 
 	log.Printf("commit queue: giving up on %s after %d retries: %v", entry.Path, maxRetries, lastErr)
+	if errors.Is(lastErr, errCommitPostUpload) {
+		cq.onCommitPostUploadFailure(entry, lastErr)
+		return
+	}
 	cq.onCommitTerminalFailure(entry)
 }
 
@@ -630,7 +640,12 @@ func (cq *CommitQueue) CommitNow(ctx context.Context, entry *CommitEntry) error 
 		}
 		return err
 	}
-	cq.onCommitSuccess(entry, committedRev)
+	if err := cq.onCommitSuccess(entry, committedRev); err != nil {
+		if cq.perf != nil {
+			cq.perf.commitFailure.add(1)
+		}
+		return err
+	}
 	return nil
 }
 
@@ -756,7 +771,7 @@ func (cq *CommitQueue) rebuildQueuedIndexLocked() {
 	}
 }
 
-func (cq *CommitQueue) onCommitSuccess(entry *CommitEntry, committedRev int64) {
+func (cq *CommitQueue) onCommitSuccess(entry *CommitEntry, committedRev int64) error {
 	if entry.HasMode {
 		ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
 		start := time.Now()
@@ -766,7 +781,7 @@ func (cq *CommitQueue) onCommitSuccess(entry *CommitEntry, committedRev int64) {
 			cq.perf.recordRemoteOp(perfRemoteMutation, err, time.Since(start), 0)
 		}
 		if err != nil {
-			log.Printf("commit queue: chmod %s to %o failed after upload: %v", entry.Path, entry.Mode&0o777, err)
+			return fmt.Errorf("%w: chmod %s to %o: %v", errCommitPostUpload, entry.Path, entry.Mode&0o777, err)
 		}
 	}
 
@@ -779,7 +794,7 @@ func (cq *CommitQueue) onCommitSuccess(entry *CommitEntry, committedRev int64) {
 		}); err != nil {
 			log.Printf("commit queue: journal commit marker failed for %s: %v (keeping local state)", entry.Path, err)
 			cq.removeFromQueue(entry)
-			return
+			return nil
 		}
 	}
 
@@ -808,6 +823,7 @@ func (cq *CommitQueue) onCommitSuccess(entry *CommitEntry, committedRev int64) {
 		cq.perf.commitSuccess.add(1)
 	}
 	log.Printf("commit queue: successfully uploaded %s (%d bytes, rev=%d)", entry.Path, entry.Size, committedRev)
+	return nil
 }
 
 // tryAutoResolveConflict attempts to resolve a 409 conflict automatically.
@@ -887,7 +903,9 @@ func (cq *CommitQueue) tryAutoResolveConflict(entry *CommitEntry) {
 	// Branch 1: idempotent — content already matches server.
 	if bytes.Equal(localData, serverData) {
 		log.Printf("commit queue: auto-resolved conflict for %s (idempotent, content matches server rev %d)", entry.Path, serverRev)
-		cq.onCommitSuccess(entry, 0)
+		if err := cq.onCommitSuccess(entry, 0); err != nil {
+			cq.onCommitPostUploadFailure(entry, err)
+		}
 		return
 	}
 
@@ -913,7 +931,17 @@ func (cq *CommitQueue) tryAutoResolveConflict(entry *CommitEntry) {
 	}
 
 	log.Printf("commit queue: auto-resolved conflict for %s via LWW (overwrote rev %d → new upload based on rev %d)", entry.Path, entry.BaseRev, serverRev)
-	cq.onCommitSuccess(entry, 0)
+	if err := cq.onCommitSuccess(entry, 0); err != nil {
+		cq.onCommitPostUploadFailure(entry, err)
+	}
+}
+
+func (cq *CommitQueue) onCommitPostUploadFailure(entry *CommitEntry, err error) {
+	if cq.perf != nil {
+		cq.perf.commitFailure.add(1)
+	}
+	cq.removeFromQueue(entry)
+	log.Printf("commit queue: post-upload failure for %s; local pending state preserved for retry: %v", entry.Path, err)
 }
 
 func (cq *CommitQueue) onCommitTerminalFailure(entry *CommitEntry) {
--- a/pkg/fuse/commit_queue_test.go
+++ b/pkg/fuse/commit_queue_test.go
@@ -72,6 +72,66 @@ func TestCommitQueueConditionalCommitSuccess(t *testing.T) {
 	}
 }
 
+func TestCommitQueueAppliesModeAfterUpload(t *testing.T) {
+	var putCalls atomic.Int32
+	var chmodCalls atomic.Int32
+	var chmodBody []byte
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		switch {
+		case r.Method == http.MethodPut:
+			putCalls.Add(1)
+			w.Header().Set("Content-Type", "application/json")
+			_, _ = w.Write([]byte(`{"status":"ok","revision":9}`))
+		case r.Method == http.MethodPost && r.URL.RawQuery == "chmod":
+			chmodCalls.Add(1)
+			chmodBody, _ = io.ReadAll(r.Body)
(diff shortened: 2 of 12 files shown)


qiffang (changes requested):
## Review: REVISE

I found three blockers in the pending-mode implementation. The high-level direction is right, but current behavior can still silently lose mode metadata or race internally.

### B1 — chmod failures are treated as successful commits, so mode metadata can be lost

The new mode propagation makes remote `chmod` part of the write commit, but every remote chmod application path is best-effort/log-only:

- `pkg/fuse/writeback_uploader.go:311-320` and `370-377`: `applyMode` logs failure and callers still remove the writeback cache entry / return nil.
- `pkg/fuse/commit_queue.go:759-798`: `onCommitSuccess` logs chmod failure, then writes the commit marker and removes shadow + pending index.
- `pkg/fuse/dat9fs.go:4451-4476`: Release deferred chmod logs/reverts local inode, but still clears pending mode state and cannot surface the error to `close(2)`.

Failure scenario: create `exec.sh` with `0755`, write data, upload succeeds, then the separate `ChmodCtx` gets a transient 500/timeout. This PR deletes the only pending metadata that remembered `0755`, so after remount/server stat the file is durable as `0644` while the operation already reported success. That is exactly the class of mode-loss regression this PR is supposed to fix.

Remote chmod needs to be part of commit success semantics: if chmod fails, keep pending metadata for retry and/or return an error from synchronous paths (`UploadSync`, flush/close-sync). Do not remove writeback/pending/shadow state or write a durable commit marker until the mode has also been applied.

### B2 — pending mode fields are mutated without the per-handle lock

New code mutates `FileHandle` mode fields through `fileHandles.ForEach` without taking `h.mu`:

- `pkg/fuse/dat9fs.go:649-655` in `clearPendingModeForInode`
- `pkg/fuse/dat9fs.go:2163-2172` in `SetAttr`
- `pkg/fuse/dat9fs.go:4471-4475` in Release cleanup

`HandleTable.ForEach` only holds the table lock; it does not protect fields inside `*FileHandle`. The same fields are read in Release/Flush paths while handles are otherwise guarded by `fh.Lock()`. This is a data race and can also clear/update pending mode while another handle is deciding what to enqueue.

All reads/writes of `PendingMode`, `HasPendingMode`, `PreviousMode`, and `HasPreviousMode` need a consistent `FileHandle` locking discipline. Add a race-oriented regression if feasible.

### B3 — explicit create mode `000` is rewritten to `0644`

`pkg/fuse/dat9fs.go:617-622` treats `input.Mode & 0777 == 0` as “no mode provided”:

```go
if mode == 0 {
    return 0o644, false
}
```

For FUSE `Create`, the mode is the requested create mode. `open(path, O_CREAT, 000)` is valid and should create a file with no permissions. Current code replies with local mode `0644` and sets `hasRemoteMode=false`, so the remote chmod is never applied either. That violates the PR’s core contract: preserve the create mode.

Add a regression for `CreateIn{Mode: 0}` and preserve `000` instead of defaulting it.

### Validation

Local checks that passed:

GitHub checks are green, but these are correctness gaps not covered by the current tests.

mornyx:
Thanks @qiffang and @coderabbitai. I pushed commit 9a8f9a8 to address the review comments.

What changed:
- Preserved explicit create mode 000 instead of treating it as an absent mode/defaulting to 0644, with a regression test for CreateIn{Mode: 0}.
- Added per-handle locking around pending mode state updates/cleanup for SetAttr and sibling handle cleanup paths.
- Made chmod part of the commit success semantics: CommitQueue and WriteBackUploader now return/record chmod failures, keep pending shadow/index/writeback cache state for retry, and avoid writing the durable commit marker or cleanup when mode application fails.
- UploadSync now applies chmod with a timeout derived from the caller context instead of context.Background().
- Synchronous flush/write-sync/shadow-spill paths now apply pending chmod before clearing dirty/local pending state, so callers can see an error when the mode update fails.

Verification:
- go test ./pkg/fuse
- go test ./cmd/drive9/cli -run 'Test.*Mount|Test.*Fuse'

qiffang (changes requested):
## PR #460 Re-review — REVISE

The original `CreateIn.Mode == 0` bug is fixed, and the pending-mode mutations are now under `FileHandle` locking. The chmod failure path is improved, but it is still not safe for the writeback-cache uploader path, and there is one remaining rollback bug plus a red CI gate.

### B1 — writeback-cache chmod failure leaves entries permanently unretryable after data upload

`WriteBackUploader.uploadOne` and `UploadSync` upload the data first, then apply chmod (`pkg/fuse/writeback_uploader.go:312-320`, `pkg/fuse/writeback_uploader.go:378-385`). If the data upload succeeds but chmod fails, the cache entry is retained with the original `WriteBackMeta` (`Kind` + `BaseRev`). On the next retry, `expectedRevisionForWriteBack` still uses `0` for `PendingNew` or the old `BaseRev` for overwrite, but the remote object has already been created/advanced by the previous upload. That retry hits a create-if-absent / CAS conflict before it ever gets back to chmod.

This means a transient post-upload chmod failure can leave the entry pending forever in writeback-cache mode. The commit queue now has an auto-resolve path for this class of conflict, but the writeback uploader does not. Please add a regression that does: upload success -> chmod failure -> submit/retry again, and prove the second pass completes chmod and removes the cache entry instead of stopping on conflict.

### B2 — rollback after deferred chmod failure still fabricates mode `000` when the original inode had no authoritative mode

When chmod is deferred on a dirty handle, `SetAttr` stores `PreviousMode = entry.Mode` and `HasPreviousMode = true` without checking `entry.HasMode` (`pkg/fuse/dat9fs.go:2234-2237`). If the remote file had no mode header cached, `entry.Mode` is just zero. Later, if the post-upload chmod fails, `applyPendingModeForHandleLocked` calls `fs.inodes.UpdateMode(ino, previousMode)` (`pkg/fuse/dat9fs.go:711-714`), which marks that zero as authoritative. The inode then reports mode `000` instead of returning to the previous “unknown mode, use default 0644” state.

This is exactly the no-authoritative-mode rollback edge case. The rollback path needs to preserve whether the previous mode was known, not just the numeric mode, and the test should cover a dirty file whose inode starts with `HasMode=false`.

### B3 — GitHub CI is red

The latest head `0315855` has failing `ci` because lint reports:

```text
pkg/fuse/commit_queue.go:594:5: ineffectual assignment to lastErr (ineffassign)
```

Even if the behavior blockers above are fixed, this PR is not mergeable while the required CI gate is failing.

### Validation

Passed locally:

GitHub status at head `0315855`: `drive9-server-local e2e smoke` success, CodeRabbit success, `ci` failure on lint.

qiffang (changes requested):
## PR #460 Re-review — REVISE

I re-reviewed the latest head `6b65ffe`. The lint-only issue from the previous head was fixed, but the two behavior blockers remain.

### B1 — writeback-cache chmod failure still leaves entries unretryable after data upload

`WriteBackUploader.uploadOne` / `UploadSync` still upload the file data, then apply chmod (`pkg/fuse/writeback_uploader.go:313-320`, `pkg/fuse/writeback_uploader.go:379-385`). If the upload succeeds and chmod fails, the cache entry is retained with the original `WriteBackMeta`. A later retry uses that same metadata in `expectedRevisionForWriteBack`: `PendingNew` still expects revision `0`, and overwrites still expect the old `BaseRev`. But the previous upload already created or advanced the remote object, so the retry conflicts before it gets back to chmod.

This turns a transient post-upload chmod failure into a permanently pending writeback-cache entry. The commit queue has a post-upload failure path, but the writeback uploader still needs either an idempotent conflict recovery path or a distinct “data uploaded, chmod pending” state. Please add a regression where upload succeeds, chmod fails, then a second submit/retry successfully applies chmod and removes the cache entry.

### B2 — deferred chmod rollback still fabricates authoritative mode `000` when the old mode was unknown

`SetAttr` records `PreviousMode = entry.Mode` and `HasPreviousMode = true` for every dirty handle (`pkg/fuse/dat9fs.go:2234-2237`), without checking `entry.HasMode`. If the inode came from a remote stat/listing without an `X-Dat9-Mode`, the old mode is not authoritative; `entry.Mode` is just zero. On post-upload chmod failure, `applyPendingModeForHandleLocked` calls `fs.inodes.UpdateMode(ino, previousMode)` (`pkg/fuse/dat9fs.go:711-714`), which marks that zero as known. The file now reports mode `000` instead of returning to the prior “unknown mode, default 0644” state.

The rollback state needs to preserve whether the previous mode was known, not only the numeric bits. Please cover the case where a dirty file starts with `HasMode=false`, chmod is deferred, the post-upload chmod fails, and subsequent attrs do not become authoritative `000`.

### Validation

Passed locally:

`make lint` could not complete locally because installing `golangci-lint v2.5.0` timed out against `proxy.golang.org`; GitHub should be used as the lint source of truth for this head.

mornyx:
@qiffang Fixed the latest local-e2e failure on PR 460. Root cause: create-time modes were being applied as a post-upload chmod for every PendingNew file, including Git's default 0644 lock/config files. When chmod saw a transient post-write 404, the pending lockfile was preserved and Git then failed with `.git/config.lock: File exists`.

Changes in the latest push:
- Skip remote chmod for PendingNew files whose final mode is the server default 0644.
- Retry post-upload chmod 404s in place, without re-uploading the already-created PendingNew file.
- If a required post-upload chmod still fails, preserve local pending state immediately instead of retrying the upload and turning it into a 409 conflict.
- Added coverage for default-mode creates, commit queue default-mode skip, chmod 404 retry, and writeback default-mode skip.

Verified locally with `go test ./pkg/fuse ./cmd/drive9/cli` and `git diff --check`. GitHub is green now: `code-ci / ci` and `local-e2e / drive9-server-local e2e smoke` both pass on `6b65ffe`.

qiffang (changes requested):
## PR #460 Review — REVISE

Independent review. I concur with both behavior blockers from the prior review; adding two additional findings.

### B1 — writeback-cache retry after chmod failure will conflict (concur with prior B1)

`WriteBackUploader.uploadOne` / `UploadSync`: data upload succeeds → chmod fails → entry retained with original `Kind`/`BaseRev`. Next retry attempts upload with the stale precondition (`PendingNew` expects rev `0`, overwrite expects old `BaseRev`). The remote object was already created/advanced by the first upload, so the retry conflicts before reaching chmod.

Same issue exists in `CommitQueue.onCommitPostUploadFailure`: it removes the entry from the queue while preserving pending metadata. On next `RecoverPending`, the re-enqueued entry re-uploads data, hitting create-if-absent conflict. The auto-resolve path can handle idempotent matches, but only if the remote content is readable and matches — this is fragile for large files and adds unnecessary remote I/O.

Both paths need either a distinct "data uploaded, chmod pending" metadata state (skip re-upload), or a deterministic conflict-resolution that goes straight to chmod when the content already matches.

### B2 — deferred chmod rollback fabricates authoritative mode `000` (concur with prior B2)

`SetAttr` stores `PreviousMode = entry.Mode` and `HasPreviousMode = true` without checking `entry.HasMode` (`dat9fs.go:~2234`). If the inode had `HasMode=false`, `entry.Mode` is zero. On rollback, `UpdateMode(ino, 0)` marks zero as authoritative. The file then reports mode `000` instead of returning to "unknown mode, display default 0644".

The fix needs to track `HasPreviousMode` as `entry.HasMode`, and rollback needs to clear `HasMode` when `HasPreviousMode` was `false`.

### B3 — Release deferred chmod path lacks retry, inconsistent with other paths

`Release` deferred chmod (`dat9fs.go:~4804`) calls `fs.applyRemoteMode` directly, without `retryPostUploadMode`. All other post-upload chmod paths (commit queue, writeback uploader, `applyPendingModeForHandleLocked`) use `retryPostUploadMode` with 5 attempts and exponential backoff for 404 (eventual consistency).

If the remote object is not yet visible (404) after upload, the commit queue/writeback paths retry and succeed, but Release silently fails and reverts mode. This is a semantic inconsistency: the same eventual-consistency window is handled differently depending on which code path triggers the upload.

### W1 — `shouldApplyRemoteMode` skips default mode only for `PendingNew`

For `PendingOverwrite` with `mode == 0o644`, `shouldApplyRemoteMode` returns `true`, triggering a redundant remote chmod. This is not a correctness bug but adds unnecessary network I/O for the most common mode. Non-blocking.

### Validation

Reviewed diff at current GitHub head.

mornyx:
@qiffang Addressed the two remaining blockers from your latest PR 460 review.

Changes:
- Added a distinct writeback  state. Once data upload succeeds but post-upload chmod fails, the cache metadata is persisted as chmod-only pending. The next writeback retry applies chmod directly and does not re-upload data with the stale PendingNew/BaseRev precondition.
- Added a regression where upload succeeds, chmod fails, then the second submit retries chmod only and removes the cache entry.
- Preserved whether the previous inode mode was authoritative. Deferred chmod rollback now restores the old unknown-mode state instead of fabricating authoritative  when the original entry had no mode.
- Added coverage for the unknown-mode rollback case.

Validation:
- ok  	github.com/mem9-ai/dat9/pkg/fuse	10.736s
- ok  	github.com/mem9-ai/dat9/cmd/drive9/cli	0.812s
-

mornyx:
@qiffang Clean follow-up for PR 460 is now pushed on b459f12.

What changed:
- Added a writeback PendingChmod state. After data upload succeeds but chmod fails, the cached metadata becomes chmod-only pending, so the next retry applies chmod directly instead of re-uploading with a stale PendingNew or BaseRev precondition.
- Added a regression for upload success then chmod failure then retry, proving the second pass does not reupload data and removes the cache entry after chmod succeeds.
- Preserved whether the previous inode mode was authoritative. Deferred chmod rollback now restores unknown-mode state instead of fabricating authoritative 000 when the original inode had no mode.
- Added coverage for that unknown-mode rollback edge case.
- Fixed the errcheck lint issue from the first push with a normal follow-up commit.

Validation:
- go test ./pkg/fuse
- go test ./cmd/drive9/cli -run Test.*Mount|Test.*Fuse
- git diff --check

qiffang (changes requested):
## PR #460 Re-review — REVISE

I re-reviewed latest head `e95825c`. B1 and B2 from my previous review are fixed:

- writeback-cache chmod failure now transitions metadata to `PendingChmod`, so retry applies chmod without re-uploading data.
- deferred chmod rollback now tracks whether the previous mode was authoritative via `PreviousModeKnown` / `SetModeState`, so unknown-mode inodes no longer become authoritative `000` on rollback.

One blocker remains from the shared review gate.

### B1 — `Release` deferred chmod path still bypasses post-upload retry

Most post-upload chmod paths now use `retryPostUploadMode`: commit queue, writeback uploader, and `applyPendingModeForHandleLocked`. The `Release` defer still calls `fs.applyRemoteMode` directly (`pkg/fuse/dat9fs.go:4573-4578`). That means a transient 404/visibility delay after data commit makes Release treat chmod as failed immediately, while the other commit paths would retry and likely succeed.

This is the same eventual-consistency window the PR already handles elsewhere. The remaining direct path should use the same retry helper, and there should be a regression where the first Release-time chmod returns 404 and the second attempt succeeds.

### Validation

Passed locally:

GitHub status at latest head: `ci` success, `drive9-server-local e2e smoke` success, CodeRabbit success.

mornyx:
@qiffang addressed the latest blocker on PR 460.

What changed:
- Release-time deferred chmod now uses retryPostUploadMode around applyRemoteMode, matching the post-upload retry behavior used by the commit queue, writeback uploader, and applyPendingModeForHandleLocked.
- Added TestReleaseRetriesDeferredChmodNotFound: the first Release chmod returns 404 and the second succeeds.
- After merging latest main, resolved the 460/461 conflict by preserving both mode-aware pending dir entries and the ReadDirPlus metadata snapshot fields.
- Tightened the regression so the inode starts at 0644 and must become 0755 after the Release retry succeeds.

Validation:
- go test ./pkg/fuse
- go test ./cmd/drive9/cli
- git diff --check

GitHub checks are green: CodeRabbit, ci, and drive9-server-local e2e smoke all passed.

qiffang (approved):
## PR #460 Re-review — GREEN

Latest head `e4f0b0e` fixes the remaining Release-time chmod retry blocker. I do not see remaining blockers.

What I verified:

- Writeback-cache chmod failure now transitions metadata to `PendingChmod`, and retry applies chmod without re-uploading data. Regression `TestWriteBackUploader_ChmodFailureRetryDoesNotReuploadData` covers this.
- Deferred chmod rollback now preserves whether the previous mode was authoritative via `PreviousModeKnown` / `SetModeState`, so unknown-mode inodes do not become authoritative `000`. Regression `TestDeferredChmodRollbackRestoresUnknownMode` covers this.
- Release deferred chmod now wraps `applyRemoteMode` in `retryPostUploadMode`, matching the other post-upload chmod paths. Regression `TestReleaseRetriesDeferredChmodNotFound` covers first chmod returning 404 and second attempt succeeding.
- Explicit create mode `000` is still preserved, while default `0644` pending-new creates skip unnecessary chmod.

Validation passed locally:

GitHub status at latest head: `ci` success, `drive9-server-local e2e smoke` success, CodeRabbit success.

srstack (changes requested):
I found two correctness issues in the new deferred-chmod flow:

1. There is still a race that can lose a newer chmod on the same inode. `SetAttr()` writes the requested mode into every dirty handle, but `applyPendingModeForHandleLocked()` snapshots `fh.PendingMode`, drops `fh.mu` for the remote chmod, and then clears pending-mode state on sibling handles unconditionally. If another `SetAttr()` arrives while that remote call is in flight, the later mode can be replaced by the older one and then cleared, so no later flush/release will apply it. The same stale-snapshot pattern exists in the deferred chmod path in `Release()`. I think this needs some kind of per-inode/per-handle mode generation check before clearing state.

2. The sync flush paths pass the upload context directly into `applyPendingModeForHandleLocked()`. That means the chmod retry loop is running on a context whose deadline may already be nearly exhausted by the upload itself, so a successful upload can still be followed by an immediate `context deadline exceeded` on chmod without getting the dedicated retry window that the async paths use. `commit_queue` and `writeback_uploader` already create a fresh timeout for the post-upload chmod; the synchronous paths should do the same for consistency.

mornyx:
@srstack addressed the two issues you called out:

- Deferred chmod now carries a per-handle generation. `applyPendingModeForHandleLocked` and the Release deferred chmod path only clear or roll back pending mode when the same generation is still current, so a concurrent newer `SetAttr` remains pending and keeps the inode at the newer mode.
- Synchronous post-upload chmod paths now use a fresh 30s retry context through `applyPendingModeWithTimeoutLocked`, instead of reusing the upload context.

I also added regression coverage for both races: `TestApplyPendingModeDoesNotClearNewerConcurrentChmod` and `TestReleaseDeferredChmodDoesNotClearNewerConcurrentChmod`.

Validated locally with `go test ./pkg/fuse`, `go test ./cmd/drive9/cli`, and `git diff --check`. GitHub `ci` and `drive9-server-local e2e smoke` are passing; CodeRabbit is still pending review at the moment.

srstack (approved):
Re-reviewed the latest head `290de4c`.

The two issues I raised are addressed:
- deferred chmod state is now generation-guarded, so an older in-flight chmod no longer clears a newer mode update on the same inode/handle set;
- the synchronous post-upload chmod paths now use a fresh timeout window instead of reusing the nearly-expired upload context.

I also ran `go test ./pkg/fuse -count=1` locally against the updated head. I do not see a remaining blocker.

qiffang (approved):
## Round 2 Review — HEAD e4f0b0e — **GREEN (APPROVE)**

All three Round 1 blockers verified fixed:

### B1 (writeback retry conflict) — FIXED
- `PendingChmod` state in `WriteBackMeta.Kind` correctly separates "data uploaded, chmod pending" from "data still needs upload"
- `MarkChmodPending` uses generation guard to prevent stale overwrites
- `uploadOne` and `UploadSync` both check `meta.Kind == PendingChmod` and skip data re-upload, going straight to `applyPendingChmod`
- `onCommitPostUploadFailure` removes entry from queue but preserves pending state for retry

### B2 (rollback fabricates 000) — FIXED
- `PreviousModeKnown` field distinguishes "previous mode was 000" from "no previous mode known"
- `SetModeState(ino, mode, hasMode)` propagates the `hasMode` flag on rollback
- `applyPendingModeForHandleLocked` error path: `fs.inodes.SetModeState(ino, previousMode, previousModeKnown)` — correct
- Release deferred chmod error path: same pattern with `previousModeKnown` — correct
- Test `TestDeferredChmodRollbackRestoresUnknownMode` validates rollback restores `HasMode=false`

### B3 (Release deferred chmod lacks retry) — FIXED
- Release path now calls `retryPostUploadMode` with exponential backoff (5 attempts, 10ms initial, 404-only retry)
- Test `TestReleaseRetriesDeferredChmodNotFound` validates 404→retry→success with exactly 2 chmod calls

### Additional verification
- `PendingModeGen` generation counter prevents stale chmod from clearing a newer concurrent SetAttr — tested by `TestApplyPendingModeDoesNotClearNewerConcurrentChmod` and `TestReleaseDeferredChmodDoesNotClearNewerConcurrentChmod`
- `shouldApplyRemoteMode` correctly skips chmod for PendingNew + default 0644 — tested
- Mode persisted through all metadata stores: pendingIndex, writeBack, shadow, commitQueue — all paths carry `Mode`/`HasMode`
- `createInputMode` masks to 0o777 and returns `hasRemoteMode=false` for default 0644
- All upload paths (write-sync, shadow-spill, streaming, multipart, flush, fsync, debounced flush) call `applyPendingModeWithTimeoutLocked` or `applyPendingModeForHandleLocked` after successful data upload

Test coverage is comprehensive: 10 new tests covering create mode preservation, zero mode, default mode skip, rollback unknown mode, release retry, concurrent chmod race (both apply and release paths), lookup mode, flush staging, commit queue chmod, and commit queue chmod failure.

PR #461, opened 2026-05-25
fix(fuse): keep local writes visible during reopens

mornyx:
## Summary
- Recreate stale directory snapshot inodes in ReadDirPlus instead of surfacing EIO.
- Preload writable opens from an already-open local dirty handle so create-then-open paths stay visible before remote commit.
- Add focused regressions for stale ReadDirPlus snapshots and create/open local visibility.

## Verification
- `env GOCACHE=/Users/mornyx/Desktop/repos/github.com/mem9-ai/drive9/.codex-gocache go test ./pkg/fuse`
- EC2 synthetic visibility probe: 8x200 create/rename/stat/list completed with 0 probe errors and `readdirplus errors=0`.
- EC2 fresh kimi-cli run: clone/checkout succeeded; `uv sync` reached 30m timeout without the prior failed-to-create-file ENOENT. FUSE summary showed `open/create/write/flush/readdirplus errors=0`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed handling of stale directory snapshots to correctly remap outdated file references.
  * Improved concurrent file write handling to properly preload data from existing open file handles.

* **Tests**
  * Added tests for directory snapshot integrity and concurrent file creation scenarios.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mem9-ai/drive9/pull/461?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/pkg/fuse/dat9fs.go
+++ b/pkg/fuse/dat9fs.go
@@ -785,6 +785,72 @@ func (fs *Dat9FS) loadWritableHandleFromWriteBackLocked(fh *FileHandle) bool {
 	return true
 }
 
+func (fs *Dat9FS) loadWritableHandleFromOpenHandleLocked(fh *FileHandle) bool {
+	if fs.openHandles == nil || fh == nil || fh.Dirty == nil {
+		return false
+	}
+
+	for _, src := range fs.openHandles.SnapshotPath(fh.Path) {
+		if src == nil || src == fh {
+			continue
+		}
+
+		src.Lock()
+		if src.Dirty == nil {
+			src.Unlock()
+			continue
+		}
+		size := src.Dirty.Size()
+		baseRev := src.BaseRev
+		isNew := src.IsNew
+		zeroBase := src.ZeroBase
+		shadowSource := fs.shadowStore != nil && (src.ShadowReady || src.ShadowSpill)
+
+		var data []byte
+		haveData := size == 0
+		if !haveData && src.Dirty.CanMaterializeFull() {
+			data = src.Dirty.Bytes()
+			haveData = true
+		}
+		src.Unlock()
+
+		if !haveData && shadowSource {
+			var err error
+			data, err = fs.shadowStore.ReadAll(fh.Path)
+			if err != nil {
+				log.Printf("open-handle preload shadow read failed for %s: %v", fh.Path, err)
+				continue
+			}
+			size = int64(len(data))
+			haveData = true
+		}
+		if !haveData {
+			continue
+		}
+
+		if len(data) > 0 {
+			if _, err := fh.Dirty.Write(0, data); err != nil {
+				log.Printf("open-handle preload failed for %s: %v", fh.Path, err)
+				continue
+			}
+		} else if err := fh.Dirty.Truncate(size); err != nil {
+			log.Printf("open-handle preload truncate failed for %s: %v", fh.Path, err)
+			continue
+		}
+		fh.Dirty.ClearDirty()
+		fh.IsNew = isNew
+		fh.ZeroBase = zeroBase || (isNew && baseRev == 0)
+		fh.BaseRev = baseRev
+		fh.OrigSize = size
+		fs.inodes.UpdateSize(fh.Ino, size)
+		if baseRev > 0 {
+			fs.inodes.UpdateRevision(fh.Ino, baseRev)
+		}
+		return true
+	}
+	return false
+}
+
 func (fs *Dat9FS) fillAttr(entry *InodeEntry, out *gofuse.Attr) {
 	out.Ino = entry.Ino
 	out.Size = uint64(entry.Size)
@@ -2864,6 +2930,12 @@ func (fs *Dat9FS) ReadDirPlus(cancel <-chan struct{}, input *gofuse.ReadIn, out 
 
 	for i := int(input.Offset); i < len(dh.Entries); i++ {
 		e := dh.Entries[i]
+		if _, ok := fs.inodes.GetEntry(e.Ino); !ok {
+			childP := dirEntryChildPath(dh.Path, e.Name)
+			isDir := e.Mode&uint32(syscall.S_IFMT) == uint32(syscall.S_IFDIR)
+			e.Ino = fs.inodes.EnsureInode(childP, isDir, 0, time.Now())
+			dh.Entries[i].Ino = e.Ino
+		}
 		entryOut := out.AddDirLookupEntry(gofuse.DirEntry{
 			Name: e.Name,
 			Ino:  e.Ino,
@@ -3177,6 +3249,9 @@ func (fs *Dat9FS) Open(cancel <-chan struct{}, input *gofuse.OpenIn, out *gofuse
 		// random writes don't discard the original file data.
 		if input.Flags&syscall.O_TRUNC == 0 {
 			preloaded := false
+			if fs.loadWritableHandleFromOpenHandleLocked(fh) {
+				preloaded = true
+			}
 			if fs.pendingIndex != nil && fs.shadowStore != nil {
 				if meta, ok := fs.pendingIndex.GetMeta(p); ok && fs.shadowStore.Has(p) {
 					if err := fs.loadWritableHandleFromShadowLocked(fh, meta); err == nil {
--- a/pkg/fuse/dat9fs.go
+++ b/pkg/fuse/dat9fs.go
@@ -3044,7 +3044,16 @@ func (fs *Dat9FS) recreateDirEntryInode(dirPath string, e DirEntry) uint64 {
 	var mode uint32
 	hasMode := false
 
-	if item, ok := fs.dirEntryMetadata(dirPath, childP, e.Name); ok {
+	if e.HasMetadata {
+		isDir = e.IsDir
+		size = e.Size
+		if !e.Mtime.IsZero() {
+			mtime = e.Mtime
+		}
+		revision = e.Revision
+		mode = e.AttrMode
+		hasMode = e.HasMode
+	} else if item, ok := fs.dirEntryMetadata(dirPath, childP, e.Name); ok {
 		isDir = item.IsDir
 		size = item.Size
 		if !item.Mtime.IsZero() {
@@ -3083,6 +3092,21 @@ func (fs *Dat9FS) dirEntryMetadata(dirPath, childP, name string) (CachedFileInfo
 		}
 	}
 	return CachedFileInfo{}, false
+}
+
+func dirEntryFromCachedInfo(item CachedFileInfo, ino uint64) DirEntry {
+	return DirEntry{
+		Name:        item.Name,
+		Ino:         ino,
+		Mode:        dirEntryMode(item.IsDir, item.HasMode, item.Mode),
+		Size:        item.Size,
+		Mtime:       item.Mtime,
+		Revision:    item.Revision,
+		AttrMode:    item.Mode,
+		HasMode:     item.HasMode,
+		IsDir:       item.IsDir,
+		HasMetadata: true,
+	}
 }
 
 func (fs *Dat9FS) ReleaseDir(input *gofuse.ReleaseIn) {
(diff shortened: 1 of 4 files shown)


qiffang (changes requested):
## PR #461 Review — REVISE

I found correctness blockers in the reopen/stale-inode paths. Local tests are green, but the current implementation still has data-loss / false-success cases that the new tests do not cover.

### B1 — ShadowSpill reopens can copy zero-filled data instead of the authoritative shadow file

`loadWritableHandleFromOpenHandleLocked` checks `src.Dirty.CanMaterializeFull()` before it checks whether the source handle is `ShadowReady` / `ShadowSpill` (`pkg/fuse/dat9fs.go:807-813`). For new files `CanMaterializeFull()` returns true because `remoteSize == 0`, even when a ShadowSpill handle has already evicted full parts from the in-memory `WriteBuffer` (`Create` wires `OnPartFull` to `EvictPart`). `Bytes()` then materializes missing parts as zero-filled data. A second writable open of a large uncommitted create can therefore preload zeros instead of the real bytes that only exist in `shadowStore`.

Failure scenario: create a large new file, write enough sequential data to trigger ShadowSpill part eviction, then open the same path again before the first handle commits. The new handle copies from `Dirty.Bytes()` and sees zero-filled evicted parts even though the shadow file contains the real data. This is data corruption. Shadow-backed handles need to read/pin the shadow source first, or the helper must skip sources that cannot materialize safely from memory.

### B2 — `OrigSize` is overwritten with local dirty size, breaking large new-file and small-to-large reopen flushes

After cloning from another open handle, the code sets `fh.OrigSize = size` (`pkg/fuse/dat9fs.go:844`). That `size` is the current local dirty size, not the remote baseline. `flushHandle` uses `OrigSize >= threshold` to choose the patch path (`pkg/fuse/dat9fs.go:4993-5022`). For a large uncommitted create, the reopened handle remains `IsNew` so `expectedRevisionForHandle` returns `0`, but `OrigSize` is now large, so close/flush can attempt `PatchFile(... expected_revision=0)` instead of a full create/write. For an existing DB-backed small file grown locally past the inline threshold, the same assignment makes the reopened handle look like it has a large remote baseline and can choose patch against a non-patchable remote object.

`OrigSize` must remain the original remote size/baseline. For `IsNew` clones that means `0`; for existing files it means the source handle's original remote size, not its current dirty size.

### B3 — Source handle selection is nondeterministic and can choose stale local state

The helper returns on the first handle from `fs.openHandles.SnapshotPath(fh.Path)` (`pkg/fuse/dat9fs.go:793-849`). That snapshot is built from a map, so iteration order is nondeterministic. With multiple writable handles for the same new path, an older empty create handle and a newer handle with unflushed writes can coexist. If the map yields the empty handle first, the new open preloads an empty buffer, sets `preloaded=true`, and skips the newer dirty handle and all later fallback paths.

This makes local visibility flaky: the same create/open/open sequence can sometimes see the latest local bytes and sometimes miss them depending on map order. The source selection needs a deterministic freshness rule, e.g. prefer handles with dirty data / highest `DirtySeq`, and it should not return after a clean or less-fresh candidate when a fresher candidate exists.

### B4 — Stale `ReadDirPlus` remap returns synthetic zero attrs with normal attr TTL

When `ReadDirPlus` sees a stale snapshot inode, it recreates the inode with `EnsureInode(childP, isDir, 0, time.Now())` (`pkg/fuse/dat9fs.go:2933-2937`). `fillEntryOut` then returns those attributes with the normal positive attr TTL (`pkg/fuse/dat9fs.go:921-930`). For a remote file that was previously listed with a real size/mode/revision, this turns the readdirplus result into a 0-byte synthetic attr that the kernel may cache. If another path already remapped the same child path to a live inode, `EnsureInode` also overwrites that live inode's size/mtime with the synthetic values.

The fix should preserve known metadata from the directory cache / pending metadata / original stat result, or perform a stat refresh before returning `EntryOut`. At minimum, do not mutate an existing live inode to size 0 as part of stale snapshot recovery. The current test only checks that EIO is avoided; it does not assert the returned attrs remain correct.

### Validation

Passed locally:

mornyx:
Thanks @qiffang, @coderabbitai, and @chatgpt-codex-connector. I pushed commit e5d4baa to address the PR 461 review feedback.

What changed:
- Open-handle preload now chooses a deterministic freshest source handle instead of relying on map iteration order. It prefers higher DirtySeq, then shadow-backed sources, then larger/currenter candidates.
- Shadow-backed / ShadowSpill sources now read from the authoritative shadow store before attempting to materialize the dirty buffer, avoiding zero-filled data from evicted parts.
- Reopened handles now preserve the remote baseline: pending-new sources keep OrigSize=0, and existing-file sources copy the source handle's original remote OrigSize instead of the current local dirty size.
- ReadDirPlus stale snapshot remapping now reuses an existing live inode without mutating its cached metadata, and when a new inode must be recreated it preserves metadata from pending/writeback state or the directory cache instead of blindly synthesizing size=0 attrs.
- Added regression coverage for ShadowSpill reopen data, freshest-handle selection, OrigSize preservation, and stale ReadDirPlus inode remapping.

Verification:
- go test ./pkg/fuse -count=1
- go test ./cmd/drive9/cli -count=1
- git diff --check

qiffang (changes requested):
## PR #461 Re-review — REVISE

Three of the four previous blockers are fixed: ShadowSpill now prefers shadow data over evicted `Dirty` bytes, `OrigSize` is preserved as the remote baseline, and open-handle source selection is deterministic instead of map-order dependent.

One blocker remains in the `ReadDirPlus` stale-inode path.

### B1 — stale `ReadDirPlus` still returns synthetic 0-byte attrs after dir-cache expiry/invalidation

`recreateDirEntryInode` only restores size/mtime/revision from `writeBack`, `pendingIndex`, or `dirCache.Lookup` (`pkg/fuse/dat9fs.go:3034-3085`). If the directory handle outlives the 10s dir-cache TTL, or the dir cache is invalidated while the open `DirHandle` still has its old `Entries`, `dirEntryMetadata` returns false and the recreated inode is still allocated with `size=0` / `mtime=now` (`pkg/fuse/dat9fs.go:3040-3058`). `ReadDirPlus` then calls `fillEntryOut`, which gives the kernel that synthetic attr with the normal positive attr TTL (60s by default).

Concrete failure:

1. `ReadDirPlus` lists `/dir/file.txt` from remote; the open `DirHandle` keeps only `Name/Ino/Mode` in `dh.Entries`.
2. The kernel forgets the file inode, removing the inode/path mapping.
3. More than `DirTTL` passes, or the dir cache is invalidated.
4. The same directory handle receives another `ReadDirPlus`; stale inode recovery has no metadata source and returns `file.txt` as size 0 for a positive attr TTL.

This is the same class of issue as the previous B4, just after the metadata cache is not available. The recovery path needs a durable metadata source for each `DirEntry` (e.g. store size/mtime/revision/hasMode in `DirEntry` when the snapshot is created), or it needs to stat the child before returning `EntryOut` when no cached metadata exists. Please add a regression that expires/invalidates `dirCache` before stale snapshot recovery and asserts the returned inode attrs are not zeroed for a real remote entry.

### Validation

Passed locally:

mornyx:
@qiffang Addressed the remaining PR 461  blocker.

Changes:
-  snapshots now carry the size, mtime, revision, mode, and type metadata captured when the directory listing is built.
- Stale  inode recovery first restores from that durable per-entry snapshot, so an open  can still return correct attrs after  expires or is invalidated.
- Pending/writeback overlays also populate the same metadata fields.
- Added a regression that invalidates  before stale snapshot recovery and verifies size/revision/mode/mtime are preserved instead of returning synthetic 0-byte attrs.

Validation:
- ok  	github.com/mem9-ai/dat9/pkg/fuse	11.595s
- ok  	github.com/mem9-ai/dat9/cmd/drive9/cli	1.272s
-

mornyx:
@qiffang Clean follow-up for PR 461 is now pushed on 7bcc9a5.

What changed:
- DirEntry snapshots now carry size, mtime, revision, mode, and type metadata captured when the directory listing is built.
- Stale ReadDirPlus inode recovery restores from that per-entry snapshot first, so an open DirHandle still returns correct attrs after dirCache expires or is invalidated.
- Pending and writeback overlays populate the same metadata fields.
- Added a regression that invalidates dirCache before stale snapshot recovery and verifies size, revision, mode, and mtime are preserved instead of returning synthetic 0-byte attrs.

Validation:
- go test ./pkg/fuse
- go test ./cmd/drive9/cli -run Test.*Mount|Test.*Fuse
- git diff --check

qiffang (approved):
## PR #461 Re-review — GREEN

Latest head `7bcc9a5` fixes the remaining stale `ReadDirPlus` blocker.

What I verified:

- `DirEntry` now carries durable per-entry metadata (`Size`, `Mtime`, `Revision`, `AttrMode`, `HasMode`, `IsDir`, `HasMetadata`) captured when the directory snapshot is built.
- `recreateDirEntryInode` prefers that stored snapshot metadata before falling back to writeback/pending/dirCache metadata, so stale inode recovery no longer depends on unexpired `dirCache`.
- Regression `TestReadDirPlusStaleSnapshotUsesStoredEntryMetadataAfterDirCacheInvalidated` covers the exact failure mode from the prior review: forget the child inode, invalidate dirCache, then recover from the still-open dir handle and preserve size/revision/mode/mtime.
- Previous fixes remain intact: shadow-backed reopen prefers shadow data, `OrigSize` stays tied to the remote baseline, and open-handle source selection is deterministic.

Validation passed locally:

GitHub status at latest head: `ci` success, `drive9-server-local e2e smoke` success, CodeRabbit success.

qiffang (approved):
## Round 1 Review — HEAD 7bcc9a5 — **GREEN (APPROVE)**

### Scope
Shadow/inode fixes: handle preloading from open handles, stale ReadDirPlus snapshot recovery, DirEntry persistent metadata propagation.

### Correctness checks

**1. `loadWritableHandleFromOpenHandleLocked`**
- Candidate selection: filters to same-inode handles with `Dirty != nil`, sorts by `(dirtySeq DESC, shadowSource DESC, size DESC, baseRev DESC)` — deterministic, prefers newest dirty state
- Correctly copies `ShadowReady`, `ShadowSource`, `BaseRev`, `IsNew`, dirty data via `CloneFrom`
- Preserves `OrigSize` from existing handle if already set (doesn't overwrite with candidate's potentially stale value)
- Returns false when no candidate found — caller falls through to shadow/writeback paths

**2. `recreateDirEntryInode` (stale ReadDirPlus recovery)**
- Uses `EnsureInodeNoUpdate` which returns existing inode without mutating cached metadata, or creates new with `Nlookup: 0`
- Critical: does NOT zero live inode metadata — `TestReadDirPlusStaleSnapshotDoesNotZeroLiveInode` validates this
- DirEntry metadata (`Size`, `Mtime`, `Revision`, `Mode`, `HasMode`) is propagated to the new inode entry only when `IsDir=false` and values are non-zero/set

**3. `EnsureInodeNoUpdate`**
- If inode already exists: increments `Nlookup`, returns existing inode number — no metadata mutation
- If inode does not exist: creates new entry with `Nlookup: 0` (not 1), which is correct for snapshot entries that haven't been formally looked up

**4. `dirEntryMetadata` fallback chain**
- writeBack → pendingIndex → dirCache, returns first match
- `cachedInfoFromWriteBackMeta` correctly converts `WriteBackMeta` to `CachedFileInfo` with mode/revision propagation

**5. DirEntry extended metadata**
- `AttrMode`, `HasMode`, `Size`, `Mtime`, `Revision`, `IsDir`, `HasMetadata` fields added to `DirEntry`
- `dirEntryMode` helper: returns `S_IFDIR|mode` or `S_IFREG|mode` based on `isDir` + `hasMode`, falls back to `S_IFDIR` / `S_IFREG` when mode unknown

### Edge cases verified
- Stale snapshot with live inode: no metadata zeroing (test)
- Multiple open handles: preload picks newest by dirtySeq (test)
- Shadow source before evicted buffer: shadow data preferred (test)
- Existing origSize preservation: not overwritten by preload (test)
- Created file with open handle: uses local handle directly (test)

### Test coverage
6 new tests cover all key paths. No missing edge cases identified.

No blockers.

PR #466, opened 2026-05-26
fix: harden coding-agent local overlay

qiffang:
## Summary

Follow-up for PR #464 review blockers after the local-only overlay was merged.

Changes:
- Filter remote directory entries that match active local-only policy before merging local overlay entries, so `readdir` and `Lookup/Open` agree.
- Narrow default `coding-agent` local-only policy to VCS metadata only: `.git`, `.hg`, `.svn`.
- Keep cache/build directories opt-in through explicit repeated `--local-only` patterns.
- Add strict release-gate coverage for `drive9 mount --profile coding-agent --local-root <tmp>` with raw `git clone`, `git status`, source edit + `git diff`, local `.git/objects`, remote source visibility, and remote `.git` absence.

## Validation

- `bash -n e2e/fuse-smoke-test.sh e2e/fuse-release-gate.sh`
- `/usr/local/go/bin/go test ./pkg/fuse -run 'LocalPolicy|CodingAgentLocalOverlay|MountOptionsCodingAgent' -count=1`
- `/usr/local/go/bin/go test ./pkg/fuse ./cmd/drive9/cli ./cmd/drive9 -count=1`
- `git diff --check`
- `PATH=/usr/local/go/bin:$PATH GOPROXY=https://goproxy.cn,direct make lint`

## Notes

I did not run live FUSE e2e locally in this session; the new gate is wired into `e2e/fuse-release-gate.sh` for strict environments with FUSE prerequisites.


qiffang:
REVISE

1. **B1: release-gate CI is red on the new coding-agent path.** `drive9-server-local e2e smoke` fails in `[8.3] coding-agent local-overlay git clone` at `coding-agent git diff sees source edit`. The exact sequence is raw `git clone`, clean `git status`, append to `README`, then `git diff -- README`; the diff is empty. This violates the core acceptance criterion for the MVP: coding-agent mount must support the normal edit → `git diff` loop while `.git` is local-only and source files are remote-persistent. Until this is fixed, the overlay cannot be considered usable for coding agents.

2. **B2: the gate still does not prove the edited source content is remote truth.** The new check only waits for `README` to appear in the remote listing. That can pass from the checkout-created file before the append. The test needs to wait for the appended marker to be visible through `drive9 fs cat` (or equivalent remote read), otherwise it does not prove source edits survive the remote persistent path.

3. **B3: the gate does not cover remount with the same local-root.** For this feature, `.git` is explicitly local and discardable/local-root-backed. The release gate should stop and remount `--profile coding-agent --local-root <same-root>` and then verify `git status` / `git log` still work and the edited source file is visible. Without this, we do not catch handle/cache/readdir state bugs across the exact lifecycle the profile introduces.

Local validation I ran:
- `/usr/local/go/bin/go test ./pkg/fuse -run 'LocalPolicy|CodingAgentLocalOverlay|MountOptionsCodingAgent' -count=1`
- `/usr/local/go/bin/go test ./pkg/fuse ./cmd/drive9/cli ./cmd/drive9 -count=1`
- `bash -n e2e/fuse-smoke-test.sh e2e/fuse-release-gate.sh`

Those pass locally, but the live FUSE release gate fails, which is the authoritative gate for this PR.

Note: GitHub rejected formal request-changes from this token with `Review Can not request changes on your own pull request`, so posting the blocking review as a PR comment.

PR #467, opened 2026-05-26
feat: add Slock callback tenant provisioning

srstack:
## Summary
- add Slock OAuth client and public login/callback routes
- add generic external subject bindings and API key issuance audit metadata
- refactor tenant provisioning so anonymous provision and Slock callback share the tenant creation path

## Tests
- go test ./pkg/slockoauth ./cmd/drive9-server ./pkg/meta ./pkg/server

## Self-review
- no blocking issues found in local review
- binding insert failure marks the newly provisioned tenant deleted; any already-created key remains blocked by tenant status

## Summary by CodeRabbit

* **New Features**
  * Optional Slock OAuth login & callback (HTML/JSON) with no-store caching; CLI env vars documented.
  * External identity bindings and "issued-by" provenance persisted for API keys; revoke-by-issuer support.

* **Bug Fixes / Behavior**
  * Startup and periodic reconciler now handle pending tenants and safer reprovisioning; provisioning flow has clearer status transitions.
  * Login endpoint returns 501 when OAuth is not configured; HTML responses omit API key rendering.

* **Tests**
  * Extensive coverage for OAuth flows, OAuth client behavior, bindings, schema, concurrency, and provisioning edge cases.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mem9-ai/drive9/pull/467?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)

### DIFF
--- a/cmd/drive9-server/main.go
+++ b/cmd/drive9-server/main.go
@@ -27,6 +27,7 @@ import (
 	"github.com/mem9-ai/dat9/pkg/meta"
 	"github.com/mem9-ai/dat9/pkg/s3client"
 	"github.com/mem9-ai/dat9/pkg/server"
+	"github.com/mem9-ai/dat9/pkg/slockoauth"
 	"github.com/mem9-ai/dat9/pkg/tenant"
 	"github.com/mem9-ai/dat9/pkg/tenant/db9"
 	"github.com/mem9-ai/dat9/pkg/tenant/starter"
@@ -258,6 +259,16 @@ func main() {
 		}
 	}
 
+	slockOAuth, err := slockOAuthFromEnv()
+	if err != nil {
+		die(err)
+	}
+	if slockOAuth != nil {
+		logger.Info(context.Background(), "slock_oauth_enabled")
+	} else {
+		logger.Info(context.Background(), "slock_oauth_disabled")
+	}
+
 	die(server.NewWithConfig(server.Config{
 		Meta:             store,
 		Pool:             pool,
@@ -271,6 +282,7 @@ func main() {
 		Logger:           srvLogger,
 		SemanticEmbedder: semanticEmbedder,
 		SemanticWorkers:  semanticWorkerOpts,
+		SlockOAuth:       slockOAuth,
 	}).ListenAndServe(addr))
 }
 
@@ -279,6 +291,24 @@ func envOr(key, fallback string) string {
 		return value
 	}
 	return fallback
+}
+
+func slockOAuthFromEnv() (*slockoauth.Client, error) {
+	origin := strings.TrimSpace(os.Getenv("DRIVE9_SLOCK_ORIGIN"))
+	if origin == "" {
+		return nil, nil
+	}
+	cfg := slockoauth.Config{
+		Origin:       origin,
+		APIOrigin:    strings.TrimSpace(os.Getenv("DRIVE9_SLOCK_API_ORIGIN")),
+		ClientID:     strings.TrimSpace(os.Getenv("DRIVE9_SLOCK_CLIENT_ID")),
+		ClientSecret: strings.TrimSpace(os.Getenv("DRIVE9_SLOCK_CLIENT_SECRET")),
+		PublicURL:    strings.TrimSpace(os.Getenv("DRIVE9_PUBLIC_URL")),
+	}
+	if cfg.APIOrigin == "" || cfg.ClientID == "" || cfg.ClientSecret == "" || cfg.PublicURL == "" {
+		return nil, fmt.Errorf("DRIVE9_SLOCK_ORIGIN enables Slock OAuth; DRIVE9_SLOCK_API_ORIGIN, DRIVE9_SLOCK_CLIENT_ID, DRIVE9_SLOCK_CLIENT_SECRET, and DRIVE9_PUBLIC_URL must also be set")
+	}
+	return slockoauth.New(cfg)
 }
 
 func s3ConfigFromEnv() s3Config {
@@ -359,6 +389,10 @@ environment:
   DRIVE9_BENCH_TIMING_LOG_ENABLED true|false to emit benchmark timing logs on successful server hot paths (default: false)
   DRIVE9_QUOTA_SOURCE tenant|server quota enforcement source (default: tenant)
   DRIVE9_TENANT_PROVIDER db9|tidb_zero|tidb_cloud_starter (default for provisioning)
+  DRIVE9_SLOCK_ORIGIN Slock browser origin; when set, enables /v1/auth/slock/*
+  DRIVE9_SLOCK_API_ORIGIN Slock API origin (required when DRIVE9_SLOCK_ORIGIN is set)
+  DRIVE9_SLOCK_CLIENT_ID Slock connected-app client id (required when DRIVE9_SLOCK_ORIGIN is set)
+  DRIVE9_SLOCK_CLIENT_SECRET Slock connected-app client secret (required when DRIVE9_SLOCK_ORIGIN is set)
 
   S3 storage (set DRIVE9_S3_BUCKET to enable AWS S3, otherwise local mock):
   DRIVE9_S3_BUCKET   S3 bucket name (enables AWS S3 mode)
--- a/cmd/drive9-server/main_test.go
+++ b/cmd/drive9-server/main_test.go
@@ -16,6 +16,71 @@ func TestVersionTextUsesDrive9ServerComponent(t *testing.T) {
 	}
 }
 
+func TestSlockOAuthFromEnvDisabledByDefault(t *testing.T) {
+	keys := []string{
+		"DRIVE9_SLOCK_ORIGIN",
+		"DRIVE9_SLOCK_API_ORIGIN",
+		"DRIVE9_SLOCK_CLIENT_ID",
+		"DRIVE9_SLOCK_CLIENT_SECRET",
+		"DRIVE9_PUBLIC_URL",
+	}
+	restore := snapshotEnv(t, keys)
+	t.Cleanup(func() { restoreEnv(t, restore) })
+	unsetEnv(t, keys)
+
+	c, err := slockOAuthFromEnv()
+	if err != nil {
+		t.Fatalf("slockOAuthFromEnv: %v", err)
+	}
+	if c != nil {
+		t.Fatal("expected nil Slock client when DRIVE9_SLOCK_ORIGIN is unset")
+	}
+}
+
+func TestSlockOAuthFromEnvRequiresCompanionVars(t *testing.T) {
+	keys := []string{
+		"DRIVE9_SLOCK_ORIGIN",
+		"DRIVE9_SLOCK_API_ORIGIN",
+		"DRIVE9_SLOCK_CLIENT_ID",
+		"DRIVE9_SLOCK_CLIENT_SECRET",
+		"DRIVE9_PUBLIC_URL",
+	}
+	restore := snapshotEnv(t, keys)
+	t.Cleanup(func() { restoreEnv(t, restore) })
+	unsetEnv(t, keys)
+	setEnv(t, "DRIVE9_SLOCK_ORIGIN", "https://app.slock.ai")
+
+	if _, err := slockOAuthFromEnv(); err == nil {
+		t.Fatal("expected partial Slock config to fail")
+	}
+}
+
+func TestSlockOAuthFromEnvEnabled(t *testing.T) {
+	keys := []string{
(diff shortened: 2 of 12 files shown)


qiffang (changes requested):
REVISE

I found two correctness blockers that need fixes before this can merge.

1. `pkg/server/server.go:347` - pending tenant recovery is unsafe for multi-instance / rolling restart deployments. `resumePendingTenants` lists all `pending` tenants and unconditionally calls `UpdateTenantStatus(..., failed)`. A second server can start while the first server is still inside `provisionTenant`: the tenant row is already inserted as `pending`, but the external provider call has not finished yet. The new process can mark that live request failed, and because the update is unconditional it can also overwrite a row that moved to `provisioning` / `active` after the list but before the update. This can fail valid callback provisioning requests and strand provisioned clusters or API keys. The recovery behavior needs an explicit stale/lease/owner policy and at minimum a conditional transition that only changes the row if it is still the expected state. Please add a regression covering the race/stale behavior.

2. `pkg/server/slock.go:125-184` / `pkg/server/server.go:3124` - Slock external identity fields are persisted after provisioning side effects without prevalidation. `slockSubjectKey(info)` and `slockMetadataJSON(info)` are derived from external Slock userinfo, then written to `tenant_api_keys.issued_by_subject_key` and `tenant_external_bindings.subject_key` (`VARCHAR(512)`) / JSON columns. There is no length or metadata-size check before `s.provisionTenant(...)` creates the external tenant cluster. If `server_id`, `sub`, or metadata fields are oversized, `issueOwnerAPIKey` or `InsertExternalBinding` can fail after the external provider has already succeeded; the code then only marks the tenant failed/deleted and does not clean up the external resource. Validate/bound subject key and metadata before provisioning, return a deterministic client/server error, and add a regression proving the provisioner is not called on invalid sizes.

Validation performed:
- `/usr/local/go/bin/go test ./pkg/slockoauth -count=1`
- `/usr/local/go/bin/go test ./cmd/drive9-server -run 'SlockOAuthFromEnv|VersionText' -count=1`

Could not run DB-backed focused tests locally because this machine lacks rootless Docker/testcontainers support (`rootless Docker not found`):
- `/usr/local/go/bin/go test ./pkg/meta -run 'ExternalBinding|IssuedBy|MetaSchemaSpecIncludesExternalBindings|MetaSchemaSpecIncludesAPIKeyScopeTables' -count=1`
- `/usr/local/go/bin/go test ./pkg/server -run 'Slock|Provision' -count=1`

GitHub checks visible at review time only show CodeRabbit success; I do not see a full CI status rollup for this PR.

qiffang (changes requested):
## Round 1 Review — HEAD a420c951 — **REVISE**

### Blockers

**B1: API key accumulation — no revocation on re-issue**
`slockIssueForBinding` calls `issueOwnerAPIKey` on every repeat callback, creating a new owner-scoped API key each time. Keys are never revoked. A Slock principal who calls the callback N times accumulates N active owner keys for the same tenant. This is both a security issue (leaked old keys remain valid) and a data growth issue. Either revoke previous keys issued for the same `(provider, subject_key)` before issuing a new one, or reuse the existing key.

**B2: `TenantPending` status has no timeout or GC — zombie tenants**
`provisionTenant` inserts a tenant with `status=pending` before calling `Provision()`. If the process crashes between InsertTenant and Provision (or between Provision and UpdateTenantStatus), the tenant remains `pending` forever. `resumePendingTenants` on startup marks them as `failed`, but only for tenants that existed before the current process started — a crash mid-provision leaves the pending tenant in limbo until the *next* restart. Consider: what if there's no restart for days?

More critically: `TenantPending` blocks API access (`503 Service Unavailable`) but the tenant row exists, so duplicate provision attempts via the external binding lock path will find the binding, call `slockIssueForBinding`, see `TenantPending` status, and hit the `default` branch returning `403 Forbidden`. The user gets stuck — can't use the tenant, can't re-provision. `TenantPending` should either be treated like `TenantFailed` for re-provision purposes, or there should be a TTL check.

**B3: Race between optimistic and locked binding reads**
`slockProvisionForUserInfo` does an optimistic (unlocked) `GetExternalBinding` first, and if found, calls `slockIssueForBinding` without holding the lock. Two concurrent callbacks for the same subject can both pass the optimistic read, both see no binding, both enter `WithExternalBindingLock`. The second one inside the lock correctly re-checks and finds the binding — but the first callback's `slockIssueForBinding` ran unlocked and could overlap with the lock holder's provision+bind. The optimistic path calling `issueOwnerAPIKey` for a tenant that is simultaneously being deleted and re-provisioned by the locked path is undefined behavior. The optimistic fast-path should be read-only (return cached tenant info) or be removed.

**B4: `UpdateTenantConnection` not defined in diff — hidden contract**
The refactored `provisionTenant` calls `s.meta.UpdateTenantConnection(ctx, tenantID, &meta.Tenant{...})` but this method is not in the diff. If it's a new method, its contract is hidden. If it's existing, what happens when it's called on a tenant in `pending` status with empty DB fields? Does it validate that the tenant exists? Does it update `updated_at`? The review cannot verify correctness of the provision flow without seeing this method.

**B5: HTML callback leaks API key in response body**
`writeSlockHTML` renders the raw API key in a `<pre>` block in the HTML response. This means:
- The API key appears in browser history, proxy logs, and any intermediate cache that stores response bodies
- `Cache-Control: no-store` helps but is not enforced by all intermediaries
- The `?format=json` path has the same issue but at least won't be rendered in a browser tab that stays open

For a credential that grants full owner access to a tenant's data, rendering it in HTML is a security concern. At minimum, document this as intentional for the MVP and add a TODO. Better: use a one-time token exchange pattern where the HTML page fetches the key via a short-lived secure endpoint.

### Warnings (non-blocking)

**W1: `newScopedTokenVersion` error not shown in diff**
`issueOwnerAPIKey` calls `newScopedTokenVersion()` when `tokenVersion <= 0`, but this function is not in the diff. Unclear what it does — presumably generates a random version? If it can fail, the error message should indicate the failure is in version generation, not just propagate a raw error.

**W2: `externalBindingLockName` uses MySQL `GET_LOCK(CONCAT(?, DATABASE()), ?)`**
The lock name is `d9_extbind:<sha256_prefix>` concatenated with the database name. MySQL named locks are server-global, so the database name prefix is needed. But `CONCAT(?, DATABASE())` makes the lock name database-dependent — if the meta store is ever accessed via a different database name (e.g., connection pooling alias), the lock becomes invisible. This is fine for now but worth a comment.

**W3: `provisionTenant` returns `TenantDSN` with plaintext password in struct**
`provisionTenantResult.TenantDSN` contains the plaintext DSN. This struct is passed to `startProvisionedTenantSchemaInit` which captures it in a goroutine. If the struct is ever logged or serialized, the DSN leaks. Consider storing only what's needed for schema init rather than the full DSN.

srstack:
Addressed the latest review blockers in `e0469a7`:

- `pending` tenant recovery is now stale-only and conditional (`pending -> failed` only if the row is still pending), plus a periodic sweeper so stale pending rows do not wait for the next restart.
- Slock callback now validates subject-key / metadata size before provisioning, so invalid external identities fail before any provider-side side effects.
- Removed the unlocked optimistic binding fast-path; binding resolution / reprovision / key rotation now all happen under the external-binding lock.
- Repeat Slock callback no longer accumulates owner keys: a fresh key is issued and older keys for the same `(provider, subject_key)` are revoked.
- HTML callback no longer renders the raw API key.
- Added regression coverage for stale vs fresh pending tenants, status-change races, oversized subject / metadata, and key rotation.

Validated with `go test ./pkg/meta ./pkg/server ./pkg/slockoauth ./cmd/drive9-server`.

qiffang (approved):
## adversary-2 · Round 2 Review — `b0df373`

**Verdict: 🟢 GREEN**

All 5 of my Round 1 blockers and both of adversary-1's blockers are fixed with regression tests.

### B1: API key accumulation → Fixed ✅
`rotateIssuedOwnerAPIKey` issues a new key then calls `RevokeAPIKeysByIssuer` to revoke all prior active keys from the same issuer/subject except the newly issued one. `TestSlockCallbackReusesExistingBinding` verifies: second callback issues fresh key, first key revoked, exactly 1 active key remains.

### B2: TenantPending zombie/deadlock → Fixed ✅
- New `TenantPending` status with `reconcilePendingTenant` using `UpdateTenantStatusIf` (CAS — only transitions if still pending, no unconditional overwrite).
- `isStalePendingTenant` only marks tenants older than `pendingTenantStaleAfter` (10min default).
- `startPendingTenantReconciler` runs periodic sweep.
- `slockIssueForBinding`: fresh pending → 409 Conflict; stale pending → re-provision. Auth middleware → 503 for pending.
- Tests: `TestStartupMarksPendingTenantFailed`, `TestStartupKeepsFreshPendingTenant`, `TestReconcilePendingTenantDoesNotOverwriteChangedStatus`.

### B3: Optimistic binding race → Fixed ✅
`slockProvisionForUserInfo` wraps the entire lookup+provision+bind flow inside `WithExternalBindingLock` (MySQL named lock per provider+subjectKey). Serializes concurrent callers for the same Slock principal. `TestWithExternalBindingLockSerializesCallbacks` verifies max concurrency = 1.

### B4: UpdateTenantConnection hidden contract → Resolved ✅
Pre-existing method in the codebase, not new code. PR correctly uses it in the refactored flow: insert pending → provision cluster → update connection → update status.

### B5: HTML leaks API key → Fixed ✅
HTML template no longer renders API key — shows "API key is only returned from the JSON callback response." `Cache-Control: no-store` + `Pragma: no-cache` headers set. `TestSlockHTMLDoesNotRenderAPIKey` verifies no `<pre>`, no `api_key`, no token-like content.

### adversary-1's blockers
- **resumePendingTenants multi-instance**: `UpdateTenantStatusIf` CAS prevents overwrite. `TestReconcilePendingTenantDoesNotOverwriteChangedStatus` verifies.
- **Slock field length/size validation**: `validateExternalIdentityPayload` checks before provisioning (subject ≤ 512B, metadata ≤ 16KB). Tests verify 0 provision calls on oversized input.

### Additional improvements noted
- `initTenantSchemaAsync` now uses `UpdateTenantStatusIf` for both `provisioning → active` and `provisioning → failed`, preventing status overwrites across instances.
- `provisionTenant` extracted as reusable method with `provisionTenantError` typed errors.
- `startServerWorker` for lifecycle-managed goroutines.
- Comprehensive test suite: 13+ new tests covering the full Slock OAuth flow.

No blockers. GREEN.

qiffang (changes requested):
Round 2 review at `b0df373`: **REVISE**.

The Round 1 correctness blockers are directionally fixed: Slock identity payload is validated before provisioning, binding lookup/provision/bind is serialized under a DB named lock, pending tenant recovery is stale-gated and CAS-protected, repeat callbacks rotate/revoke prior Slock-issued owner keys, and HTML no longer renders the raw API key.

Blocking issue:

- `pkg/server/server.go:362`: `resumeTenantSchemaInit` is now unused after `resumeProvisioningTenants` switched to `startTenantSchemaInitResume`. The live GitHub `ci` check fails in `make lint` with `unused: func (*Server).resumeTenantSchemaInit is unused`, so this PR is not mergeable. Remove the dead helper or route the resume path through it.

Validation I ran:

- `go test ./pkg/slockoauth -count=1` ✅
- `go test ./cmd/drive9-server -run 'SlockOAuthFromEnv|VersionText' -count=1` ✅
- `go test ./pkg/meta ...` and `go test ./pkg/server ...` are blocked locally by missing rootless Docker/testcontainers; GitHub `drive9-server-local e2e smoke` is passing.
- GitHub `ci` is failing on the unused function above.

qiffang (approved):
Round 3 review at `2713688`: **GREEN**.

The only remaining Round 2 blocker was the unused `resumeTenantSchemaInit` helper. This head removes it cleanly; the resume path now uses `startTenantSchemaInitResume`, and the live GitHub `ci` check passes.

Confirmed status:
- `ci` ✅
- `drive9-server-local e2e smoke` ✅
- Prior Slock provisioning fixes remain intact: pre-provision identity payload validation, DB named lock around binding/provision/bind, stale/CAS pending reconciliation, issued-owner-key rotation/revocation, and no raw API key in HTML callback.

Local validation:
- `go test ./pkg/slockoauth -count=1` ✅
- `go test ./cmd/drive9-server -run 'SlockOAuthFromEnv|VersionText' -count=1` ✅
- `go test -c ./pkg/server` ✅

Local `make lint` could not complete because downloading `golangci-lint` from `proxy.golang.org` timed out on this machine; GitHub `ci` ran `make lint` successfully.

PR #471, opened 2026-05-27
feat: git fast clone workspaces

mornyx:
## Summary

This PR adds Drive9's Git-aware fast workspace path for coding-agent mounts:

```bash
drive9 git clone --fast <repo-url> <mounted-path>
drive9 git clone --fast --blobless <repo-url> <mounted-path>
```

The goal is to make `git clone` into a Drive9-mounted directory fast while still preserving agent work across sandbox/container replacement. Clean repository files are represented by a Git tree manifest instead of being checked out into Drive9's generic file tables. Dirty working-tree changes and recoverable Git state are persisted through Drive9-specific Git workspace tables.

The full design is documented in [`docs/design/git-fast-clone-workspace.md`](docs/design/git-fast-clone-workspace.md).

## Current Design

This is intentionally a Git-specific optimization layered on top of Drive9 FUSE, not a general-purpose ArtifactFS-style command suite.

- `--fast` performs a no-checkout clone with a full local `.git` object database. The visible working tree is virtual, but clean file reads are served from local Git objects, so build-time reads avoid Drive9 backend round trips.
- `--fast --blobless` performs `git clone --filter=blob:none --no-checkout`. Clean blobs are fetched lazily by Git when needed, with a hidden local hydrate/read-through cache to move remote blob downloads off the build hot path.
- `.git` stays on the local overlay for native Git performance. Drive9 checkpoints lightweight `.git` state and small local-only Git objects so another sandbox can restore the repository state.
- Clean Git blob content is never stored in Drive9 backend tables.
- Dirty working-tree content, file creates, directory creates, deletes, chmods, symlinks, and renames are stored in `git_workspace_overlay`.
- Staged/local-only Git objects are packed into `git_workspace_object_packs` only under the v1 limits: 5 MiB per local-only blob and 256 MiB total inline pack.
- Oversized staged/local-only Git objects are downgraded safely: working-tree content remains durable through the dirty overlay, but staged/local ref state that would point at missing objects is not restored.

## Data Model

New Git workspace state lives in backend DB tables:

- `git_workspaces`: active workspace roots, repo URL, branch, base/head commit, mode (`fast` or `fast-blobless`), and status.
- `git_workspace_tree_nodes`: relative Git tree manifest with kind, mode, object SHA, and size metadata.
- `git_workspace_overlay`: durable working-tree overlay for upserts, whiteouts, chmods, symlinks, and directory entries.
- `git_workspace_git_state`: objectless `.git` checkpoint (`tar.gz-no-objects`) for cross-sandbox restore.
- `git_workspace_object_packs`: inline packs for small local-only objects needed to recover staged/local Git state.

Ordinary `file_nodes` still represent non-Git Drive9 paths. Git clean file content and Git workspace dirty file content under fast workspaces are intentionally kept out of generic `file_nodes` file rows.

## Clone And Runtime Flow

1. The user mounts Drive9 with `--profile=coding-agent` and `--local-root`.
2. `drive9 git clone --fast` runs a no-checkout Git clone under the mounted path.
3. The CLI reads `HEAD`, branch, and `git ls-tree -r -t -z HEAD`.
4. The CLI does not use `git ls-tree -l`; `-l` forces blob size lookups and breaks blobless clone latency.
5. For GitHub repositories, missing file sizes are filled through the GitHub Trees API. Unknown sizes fall back to `size_bytes=-1`.
6. The CLI initializes the Git index with `git read-tree --reset HEAD`.
7. Drive9 registers the workspace, replaces the tree manifest, and uploads an initial objectless `.git` checkpoint.
8. FUSE serves the working tree from `git_workspace_tree_nodes` plus `git_workspace_overlay`.
9. Clean reads use the hidden materialized clean tree, local blob cache, or local Git objects before falling back to `git cat-file`.
10. Writes are copy-on-write into the Drive9 overlay. In `write-sync`, modified content is uploaded before write returns.
11. Git commands (`git add`, `git commit`, `git push`) continue to use native Git against the local `.git`.

## Sandbox Replacement

On a replacement sandbox with a fresh local root:

1. FUSE loads active `git_workspaces`, tree manifests, overlays, Git state checkpoints, and object packs from Drive9.
2. When the repo path or `.git` path is accessed, FUSE recreates local `.git` with a no-checkout clone from `repo_url`.
3. For `mode=fast`, restore uses a full clone. For `mode=fast-blobless`, restore uses a blobless partial clone.
4. FUSE unpacks saved local-only object packs and extracts the sanitized objectless `.git` checkpoint.
5. The visible working tree is reconstructed from clean manifest plus durable overlay.
6. `git status`, `git log`, file reads, staged small edits, local commits/stashes under object limits, and dirty working-tree content are restored.

## Important Follow-Up Fixes Included

This branch went through several rounds after the initial fast-clone implementation:

- Reworked fast clone so default `--fast` uses full local Git objects while the checkout remains virtual.
- Added opt-in blobless mode with local hydrate/read-through cache.
- Removed `git ls-tree -l` from manifest generation to avoid accidental lazy blob fetches.
- Added GitHub Trees API size enrichment and unknown-size handling.
- Added inline local-only object packs for staged/local Git recovery.
- Lowered the local-only object threshold from 16 MiB to 5 MiB.
- Added coding-agent local-only policy for generated/dependency paths and repo `.gitignore`-matched outputs.
- Tightened default local-only patterns so source directories named `build/` are not hidden globally.
- Fixed local overlay flush behavior for generated outputs so builds avoid unnecessary backend/fsync cost.
- Hardened `.git` archive restore against symlink/path traversal.
- Added read-only guards and safer rename/chmod/root-delete semantics for Git workspace overlay operations.
- Propagated explicit `.git` checkpoint errors instead of silently logging them.
- Made Git/cache I/O helpers context-aware and removed credential-bearing clone args from error messages.
- Added server-side object-pack size/checksum validation and HTTP body caps.

## Review Comment Resolution

The latest review pass has been addressed in commit `d1721ef`:

- All CodeRabbit actionable comments from the last rounds were fixed.
- The earlier `chatgpt-codex-connector` FUSE semantics comments were fixed.
- `CodeRabbit` now reports `pass`.
- `ci` now reports `pass`.
- `drive9-server-local e2e smoke` now reports `pass`.

## Validation

Local validation run during this PR included:

```bash
env GOCACHE=... go test ./pkg/fuse
env GOCACHE=... go test ./cmd/drive9/cli ./pkg/gitcache
env GOCACHE=... go test ./pkg/datastore ./pkg/server -run 'TestGit|Test.*ObjectPack|TestReadDirectoryReturns404'
env GOCACHE=... go test ./pkg/tenant/schema -run 'TestGitWorkspaceTiDBSchemaStatementsExecuteInMySQL|TestInitTiDBTenantSchemaStatementsForModeIncludesVault'
env GOCACHE=... go test ./pkg/client
env GOCACHE=... GOLANGCI_LINT_CACHE=... bin/golangci-lint run --timeout 10m
git diff --check
```

Dev E2E coverage included the sandbox A/B replacement scenario:

- sandbox A mounted Drive9, fast-cloned a repo, edited files, created files, deleted files, staged/committed/pushed, then left an uncommitted edit;
- sandbox A unmounted;
- sandbox B mounted the same Drive9 path with a fresh local root/cache;
- `.git` restored, `git log -1` showed the sandbox A commit, `git status --porcelain=v1` preserved the expected dirty state, and file contents/deletions matched expectations;
- backend table checks confirmed the expected workspace/tree/state/overlay rows and no Git workspace file content leakage into `file_nodes`;
- local SQLite file count remained zero.

The sanitized evidence and table expectations are captured in the design doc.

## Known Limits

- Object packs are inline-only in v1; there is no S3/object-storage path for large local-only Git objects yet.
- Large local-only staged/local-ref state can be downgraded, but dirty working-tree content must remain durable through the Drive9 overlay.
- After `git commit`, the overlay remains relative to the original base tree. A later explicit checkpoint/manifest-advance flow can compact the overlay.
- Submodules, LFS, branch switching, merge/rebase conflict flows, and very large dirty overlays need broader follow-up validation.

## Summary by CodeRabbit

* **New Features**
  * Added `drive9 git clone --fast` command for rapid Git repository cloning with optional blobless mode.
  * Added `drive9 git hydrate` command for populating clean tree checkouts from cached repositories.
  * Integrated Git workspace support into FUSE mounts for seamless Git-aware file layering and tracking.
  * Added REST API endpoints for managing Git workspaces, tree nodes, state checkpoints, and overlay entries.

* **Documentation**
  * Added design documentation for Git fast-clone workspace architecture and workflows.

[Review Change Stack](https://app.coderabbit.ai/change-stack/mem9-ai/drive9/pull/471?utm_source=github_walkthrough&utm_medium=github&utm_campaign=change_stack)


mornyx:
@qiffang @srstack FYI, I updated this PR after the latest review rounds and pushed the fixes in `d1721ef`.

Review comments addressed:

- Fixed the FUSE Git workspace semantics called out by @chatgpt-codex-connector: workspace-root deletion now deactivates the workspace, directory chmod preserves kind, directory overlay modes preserve permissions, incompatible renames are rejected, and read-only mounts reject Git overlay mutations.
- Fixed the latest @coderabbitai actionable comments: Git/cache I/O helpers are context-first, background hydrate uses `exec.CommandContext`, clone errors no longer echo credential-bearing args, test object-pack responses are metadata-only, object-pack uploads validate body/content/size/checksum, duplicate test index setup is idempotent, path/enums/tree consistency are tightened, and `.git` archive restore is hardened against symlink traversal.
- Propagated explicit `.git` checkpoint failures and kept pending chmod state unless flush succeeds.
- Updated the design doc and PR description to reflect the current implementation: default `--fast` uses full local Git objects with a virtual checkout, `--fast --blobless` is opt-in, clean blobs stay out of Drive9, dirty overlay and lightweight Git state persist to backend tables, and local-only object packs use the 5 MiB per-blob / 256 MiB total inline limits.

Validation status after the push:

- `CodeRabbit`: pass
- `ci`: pass
- `drive9-server-local e2e smoke`: pass

Local validation also passed for the focused FUSE/CLI/gitcache/datastore/server/client/schema tests, `golangci-lint`, and `git diff --check`.

mornyx:
@coderabbitai addressed in `a5c1575`.

What changed:

- Added `testmysql.OpenDB(t, dsn)` so tests can use the shared MySQL test helper instead of opening the driver directly.
- Updated `testmysql.ResetDB` / related reset helpers to tolerate missing tables, which lets schema-focused tests call the shared reset helper even when they only create a subset of tenant tables.
- Updated `pkg/tenant/schema/git_workspace_test.go` to use `testmysql.OpenDB(t, testDSN)` and `testmysql.ResetDB(t, db)`.
- Removed the hardcoded Git workspace `DROP TABLE` cleanup loop from that test.
- Kept repeated-run behavior safe by ignoring existing-schema duplicate index/table errors through the existing `isIgnorableTiDBSchemaError` path.

Validation:

```bash
env GOCACHE=... go test ./pkg/tenant/schema -run TestGitWorkspaceTiDBSchemaStatementsExecuteInMySQL
env GOCACHE=... go test ./internal/testmysql
git diff --check
```

All passed.

qiffang (changes requested):
## dat9-dev1 Review — PR #471: git fast clone workspaces

I've done a thorough read of all 32 files (9001 additions). The design is solid, the data model is well-thought-out, and the implementation handles many edge cases correctly. That said, I have blockers and important concerns below.

---

### 🔴 B1 — `ensureGitWorkspaces` silently drops workspace load errors, stale cache served indefinitely

**File:** `pkg/fuse/git_workspace.go` L76–142

`ensureGitWorkspaces` does `continue` on `ListGitTree` / `ListGitOverlayEntries` failures, silently dropping individual workspaces. If the backend is transiently down during reload, `git.loaded = true` and `git.loadedAt = time.Now()` are still set, so subsequent callers see an empty/stale workspace list for `gitWorkspaceRefreshInterval` (1s). There is no retry, no error propagation, and no logging.

**Impact:** A single transient backend hiccup during workspace refresh can make FUSE silently serve the whole repo tree from generic `file_nodes` (or return ENOENT) for at least 1 second, which is long enough for a `git status` or build to observe an inconsistent filesystem.

**Fix:** At minimum, log errors on continue and do NOT set `loaded=true` if any workspace load fails. Better: propagate the error so callers can retry.

---

### 🔴 B2 — `renameGitDir` does N sequential HTTP RPCs per tree node — O(n) overlay puts

**File:** `pkg/fuse/git_workspace.go` L2149–2232

`renameGitDir` iterates over ALL tree nodes + overlay entries under the old prefix. For each node it does:
1. `putGitOverlay` for the new path (HTTP round-trip to server + DB write)
2. Later, `putGitOverlay` whiteout for the old path (another HTTP round-trip)

For a directory with 500 files, that's ~1000 sequential HTTP RPCs inside a single `rename(2)` syscall. This will timeout or make the filesystem unusable for any non-trivial directory rename.

**Impact:** Renaming a non-trivial directory in a git workspace (e.g. `mv src/ lib/`) will block the FUSE thread for tens of seconds to minutes.

**Recommendation:** At minimum, document this as a known v1 limit with an explicit size/count guard (EXDEV above N entries). Better: add a batch overlay API endpoint that accepts a list of overlay mutations in a single HTTP call.

---

### 🔴 B3 — `readGitFile` with unknown size (`size_bytes=-1`) reads 0 bytes for clean files

**File:** `pkg/fuse/git_workspace.go` L585–609

When `readGitFile` is called with `offset=0, size=entry.Size` and the entry has `SizeBytes=-1` (unknown), `readGitCleanFile` is called with `size=-1`. The `gitcache.ReadTreeFile` call will use `size=-1` which eventually calls `readFileRange(f, info.Size(), offset, -1)` — this works correctly for cached tree files. But for the `materializeGitBlobOnce` → `sliceRead(data, 0, -1)` path, `sliceRead` correctly handles `size < 0` by reading to end.

**However**, the `fillAttr` path (L1168) clamps negative sizes to 0: `if size < 0 { size = 0 }`. This means the kernel sees `st_size=0` for unknown-size files. A `cat file.txt` call will typically issue `read(fd, buf, 0)` (since the kernel believes file size is 0) and return empty content, even though the file has real data.

The design doc says "FUSE reports size 0 to the kernel attr path as a fallback, but real read/write/truncate/rename-copy paths load the base blob on demand and update the local inode size from the actual content length." But I don't see where `readGitFile` or `readGitCleanFile` updates the inode size after loading the real blob. The `fillAttr` size=0 will cause the kernel to not even issue reads.

**Impact:** Files with `size_bytes=-1` (non-GitHub repos, GitHub API truncation) will appear empty to userspace tools.

**Fix:** After `materializeGitBlobOnce` successfully reads the blob, update the inode size via `fs.inodes.UpdateSize()`.

---

### 🟡 B4 — `ignoreCache` grows unbounded

**File:** `pkg/fuse/git_workspace.go` L213–228

The `ignoreCache` map is never evicted. For repos with many files (100K+), iterating through all paths will cause this map to grow without bound. The cache key includes `HeadCommit`, so it's also never invalidated when the workspace manifest changes.

**Fix:** Add an LRU eviction or bounded cache size. Invalidate when `HeadCommit` changes.

---

### 🟡 B5 — `archiveLocalGitStateDir` reads entire `.git` into memory

**File:** `pkg/fuse/git_workspace.go` L1543–1612

The `.git` directory (excluding objects) is archived by walking the entire tree and buffering into `bytes.Buffer`. For repos with large reflogs, many branches, or large index files, this can be significant. Combined with the inline `content_blob` storage in the DB, there's no size guard.

The design doc mentions "lightweight payload", but there's no enforcement. A large `.git/index` (e.g. Linux kernel: ~50MB index) will be archived, uploaded, and stored inline every time a `.git` file is written.

**Recommendation:** Add a size cap check after archiving. If the archive exceeds a threshold (e.g. 32 MiB), log a warning and consider skipping or compressing more aggressively. This is a v1 known limit but should at least be guarded.

---

### 🟡 B6 — `shouldSkipGitObjectStatePath` is overly broad

**File:** `pkg/fuse/git_workspace.go` L1614–1621

```go
func shouldSkipGitObjectStatePath(rel string, _ fs.DirEntry) bool {
	for _, part := range strings.Split(filepath.ToSlash(rel), "/") {
		if part == "objects" {
			return true
		}
	}
	return false
}
```

This skips ANY directory or file named `objects` anywhere in the `.git` tree, not just `.git/objects`. If a Git hook or config directory happened to contain a file/dir named `objects`, it would be silently dropped from the checkpoint.

**Fix:** Check specifically for `objects` or `objects/` at the top level of the `.git` dir, and `modules/*/objects` for submodules:
```go
if rel == "objects" || strings.HasPrefix(rel, "objects/") { return true }
if strings.Contains(rel, "/objects") { /* check it's under modules/*/objects */ }
```

---

### 🟡 B7 — `cleanGitRelativePath` is duplicated between server and datastore

**File:** `pkg/server/git_workspace.go` L643–676 and `pkg/datastore/git_workspace.go` L229–262

These two functions are identical. If one is updated and the other isn't, the server and datastore will disagree on path validation. This is a classic consistency bug waiting to happen.

**Fix:** Extract to a shared package (e.g. `pkg/pathutil`) and call from both.

---

### 🟡 B8 — `restoreGitObjectsFromRemote` leaks clone credentials in error messages

**File:** `pkg/fuse/git_workspace.go` L805–848

The PR description mentions "removed credential-bearing clone args from error messages", but `restoreGitObjectsFromRemote` still passes `repoURL` directly to `git clone`, and the stderr is included in the error message. If the repo URL contains embedded credentials (e.g. `https://token@github.com/...`), they'll appear in logs.

**Fix:** Sanitize stderr before including in error messages, or strip credentials from the URL before logging.

---

### Minor / Non-blocking

1. **`gitWorkspaceRefreshInterval = time.Second`** — This means every FUSE operation pays a mutex lock + time check. For hot paths like `Read`, this adds measurable overhead. Consider caching the workspace lookup result per-inode or increasing the interval.

2. **`gitMaterializeCall` singleflight** — The manual singleflight pattern works but `golang.org/x/sync/singleflight` would be cleaner and handle panics correctly.

3. **`git_workspace_overlay` PK is `(workspace_id, path_hash)`** — If SHA-256 path hash collides (astronomically unlikely but theoretically possible), the overlay would silently overwrite a different path. The `path` column is stored but not part of the PK, so the DB won't reject collisions. The `GetGitOverlayEntry` query does `WHERE path_hash = ? AND path = ?` which is correct for reads but inserts can still collide. This is acceptable for v1 but worth documenting.

4. **No test for `.git` checkpoint round-trip** — The tests cover overlay CRUD, tree manifests, and read paths well. But I don't see a test that archives `.git`, extracts it, and verifies the result matches. Given the symlink traversal hardening, this deserves a dedicated test.

---

### Summary

| # | Severity | Issue |
|---|----------|-------|
| B1 | 🔴 Blocker | Silent workspace load failure → stale/empty filesystem view |
| B2 | 🔴 Blocker | O(n) sequential HTTP RPCs in directory rename |
| B3 | 🔴 Blocker | `size_bytes=-1` files appear empty (kernel sees size=0, never issues reads) |
| B4 | 🟡 Major | Unbounded ignoreCache growth |
| B5 | 🟡 Major | No size guard on `.git` archive checkpoint |
| B6 | 🟡 Major | Overly broad `objects` skip in archive filter |
| B7 | 🟡 Medium | Duplicated path validation between server and datastore |
| B8 | 🟡 Medium | Potential credential leak in restore error messages |

**Verdict: REQUEST_CHANGES** — B1, B2, B3 need fixes before merge.

pkg/fuse/git_workspace.go
```diff
@@ -0,0 +1,2289 @@
...
+	}
+	sort.Slice(loaded, func(i, j int) bool {
+		return len(loaded[i].localRoot) > len(loaded[j].localRoot)
+	})
+
+	fs.git.mu.Lock()
+	fs.git.workspaces = loaded
+	fs.git.loaded = true
+	fs.git.loadedAt = time.Now()
+	fs.git.mu.Unlock()
```
qiffang, line 152: 🔴 **B1: Silent workspace load failure → stale filesystem view**

`ensureGitWorkspaces` does `continue` on `ListGitTree`/`ListGitOverlayEntries` failures but still sets `loaded=true` and `loadedAt=time.Now()`. This means:
1. A transient backend error during refresh silently drops workspaces
2. Subsequent callers see stale/empty workspace list for 1 second
3. FUSE serves the repo tree from generic `file_nodes` or ENOENT

At minimum, do NOT set `loaded=true` if any workspace load fails. Log the errors.

pkg/fuse/git_workspace.go
```diff
@@ -0,0 +1,2289 @@
...
+			return listDirErrToFuseStatus(err)
+		}
+		if len(entries) > 0 {
+			return gofuse.Status(syscall.ENOTEMPTY)
+		}
+	}
+	return gofuse.OK
+}
+
+func (fs *Dat9FS) renameGitDir(ctx context.Context, rt *gitWorkspaceRuntime, oldRel, newRel string) gofuse.Status {
```
qiffang, line 3080: 🔴 **B2: O(n) sequential HTTP RPCs in directory rename**

For a directory with N children, `renameGitDir` makes ~2N sequential HTTP calls (one overlay put per new path + one whiteout per old path). For a 500-file directory rename, that is ~1000 serial HTTP round-trips inside a single `rename(2)` syscall.

Either:
1. Document as known v1 limit + add a count guard that returns EXDEV above a threshold (e.g. 100 entries)
2. Or add a batch overlay mutation endpoint

pkg/fuse/dat9fs.go
```diff
@@ -1129,9 +1168,13 @@ func (fs *Dat9FS) loadWritableHandleFromOpenHandleLocked(fh *FileHandle) bool {
 }
 
 func (fs *Dat9FS) fillAttr(entry *InodeEntry, out *gofuse.Attr) {
+	size := entry.Size
```
qiffang, line 1231: 🔴 **B3: `size_bytes=-1` files appear empty to userspace**

`fillAttr` clamps negative sizes to 0, so the kernel sees `st_size=0` for files with unknown sizes. The kernel will not issue read requests for a file it believes is empty. I do not see where `readGitCleanFile` or `materializeGitBlobOnce` updates the inode size after loading the real blob content.

After successfully reading a blob for an unknown-size file, call `fs.inodes.UpdateSize(ino, int64(len(data)))` so subsequent `getattr` returns the real size.

qiffang (changes requested):
## adversary-2 Independent Review — PR #471: git fast clone workspaces

Reviewed all 32 files, 9001 lines at head `a5c1575d12`. Cross-referenced with dat9-dev1's review — I **concur** with their B1 (silent workspace load), B2 (O(n) rename RPCs), and B3 (size_bytes=-1 → empty files). Those are valid blockers.

Below are **additional findings** dat9-dev1 did not cover:

---

### 🔴 B1 — TOCTOU race in `ensureGitStateRestored` — concurrent callers both restore

**File:** `pkg/fuse/git_workspace.go` L761–803

```go
func (fs *Dat9FS) ensureGitStateRestored(ctx context.Context, rt *gitWorkspaceRuntime) error {
    fs.git.mu.Lock()
    if rt.restored {           // check under lock
        fs.git.mu.Unlock()
        return nil
    }
    fs.git.mu.Unlock()         // lock released HERE
    // ... expensive work: GetGitState HTTP, restoreGitObjectsFromRemote (git clone),
    //     restoreGitObjectPacks, extractGitArchive — all OUTSIDE the lock
    fs.git.mu.Lock()
    rt.restored = true
    fs.git.mu.Unlock()
}
```

Two goroutines can both observe `rt.restored == false`, both release the lock, and both begin the restore concurrently. The second `extractGitArchive` call will overwrite files while the first is still writing them, causing corrupted `.git` state. This is a classic TOCTOU — the check and the state mutation are not atomic.

**Impact:** On first access after mount, parallel FUSE operations (e.g. `git status` which does many concurrent stat/read calls) will race on `.git` restoration. The resulting corruption can break the git index, HEAD, or refs.

**Fix:** Use a `sync.Once` per workspace runtime, or hold a per-workspace restore mutex (not `fs.git.mu`) across the entire restore operation with a second `rt.restored` check after acquiring. The lock scope must cover the full restore, not just the flag check.

---

### 🔴 B2 — `gitWorkspaceForPath` creates `context.Background()`, discards caller cancellation

**File:** `pkg/fuse/git_workspace.go` L144–150

```go
func (fs *Dat9FS) gitWorkspaceForPath(localPath string) (*gitWorkspaceRuntime, string, bool) {
    ctx, cancel := context.WithTimeout(context.Background(), fuseTimeout)
    _ = fs.ensureGitWorkspaces(ctx)
    cancel()
    ...
```

Every FUSE operation that touches a git workspace path calls `gitWorkspaceForPath`. The caller already has a context (from the FUSE request or from `checkpointGitStateForPath` etc.), but this function ignores it and creates a fresh `context.Background()`. If the caller's context is cancelled (e.g. FUSE interrupt, process shutdown), this call keeps running for up to `fuseTimeout`.

Similarly, `gitIgnoredPathLocalOnly` at L247 and `readGitCleanFile` at L931 create independent `context.Background()` timeouts.

**Impact:** During unmount or FUSE interrupt, these background contexts keep HTTP calls alive, delaying shutdown. More critically, cascading timeout pressure (many FUSE ops each creating their own background timeout) can exhaust HTTP connection pools.

**Fix:** Accept `context.Context` from the caller and derive timeouts from it. At minimum, `gitWorkspaceForPath` should take a `ctx` parameter or use the caller's context as parent.

---

### 🔴 B3 — `extractCodeloadTar` vulnerable to symlink traversal (unlike `extractGitArchive`)

**File:** `pkg/gitcache/cache.go` L365–425

`extractCodeloadTar` validates that the target path starts with `dst` (L384), but does NOT check intermediate path components for symlinks. If a crafted tarball contains:
1. Entry: symlink `foo` → `/tmp` (passes `safeSymlinkTarget` check since it's relative-looking or absolute but "safe")
2. Entry: regular file `foo/exploit`

The prefix check on L384 sees `dst/foo/exploit` which starts with `dst` — passes. But `os.MkdirAll(filepath.Dir(target))` follows the `foo` symlink, creating and writing to `/tmp/exploit`.

By contrast, `extractGitArchive` (same PR, `git_workspace.go` L1059) explicitly calls `ensureNoGitArchiveSymlinkTraversal` which walks each intermediate component with `os.Lstat` to detect symlinks. The two extraction functions have asymmetric security posture for the same threat class.

`safeSymlinkTarget` (L648) only checks that the target doesn't start with `/` or contain `..` — but the attack vector is symlink-as-intermediate-directory, not the target value itself.

**Impact:** A malicious GitHub codeload tarball (or MITM'd tarball if HTTPS is downgraded) can write arbitrary files outside the cache directory during hydration.

**Fix:** Apply the same `ensureNoGitArchiveSymlinkTraversal`-style intermediate symlink check in `extractCodeloadTar`, or use `os.Lstat` + reject on each intermediate directory before writing.

---

### Cross-validation of dat9-dev1 findings

| dat9-dev1 | My assessment |
|-----------|---------------|
| B1 silent workspace load | ✅ Concur — `loaded=true` set despite per-workspace errors |
| B2 O(n) rename RPCs | ✅ Concur — needs batch API or EXDEV guard |
| B3 size_bytes=-1 empty | ✅ Concur — `fillAttr` clamps to 0, kernel won't issue reads |
| B4 ignoreCache unbounded | ✅ Concur |
| B5 .git archive no cap | ✅ Concur |
| B6 objects filter broad | ✅ Concur |
| B7 path validation dup | ✅ Concur |
| B8 credential leak | ✅ Concur |

### Additional observations (non-blocking)

1. **`validateObjectSHA` in `pkg/gitcache/cache.go` L640** has no upper length bound — accepts arbitrarily long hex strings. The server-side `validateGitObjectID` correctly enforces 40 or 64 chars, but the client-side cache helper does not. If the cache is ever used with untrusted input, this allows unbounded path construction.

2. **Codex/CodeRabbit both flagged** directory chmod storing `Kind: "file"` and missing read-only guard on git workspace mkdir — these are valid additional blockers from the automated reviews.

---

### Summary of NEW blockers (not in dat9-dev1 review)

| # | Severity | Issue |
|---|----------|-------|
| B1 | 🔴 Blocker | TOCTOU race in `ensureGitStateRestored` — parallel callers both restore `.git` |
| B2 | 🔴 Blocker | `context.Background()` discards caller cancellation in hot FUSE paths |
| B3 | 🔴 Blocker | `extractCodeloadTar` symlink traversal — no intermediate component check |

Combined with dat9-dev1's 3 blockers = **6 total blockers** before merge.

**Verdict: REQUEST_CHANGES**

pkg/fuse/git_workspace.go
```diff
@@ -0,0 +1,2289 @@
...
+func (fs *Dat9FS) ensureGitStateRestored(ctx context.Context, rt *gitWorkspaceRuntime) error {
+	if rt == nil || fs.localOverlay == nil {
+		return nil
+	}
+	fs.git.mu.Lock()
+	if rt.restored {
+		fs.git.mu.Unlock()
+		return nil
+	}
+	fs.git.mu.Unlock()
```
qiffang: 🔴 **B1 TOCTOU race**: Lock is acquired at L765 to check `rt.restored`, then released at L770. The expensive restore work (GetGitState, restoreGitObjectsFromRemote, extractGitArchive) runs without the lock. Two concurrent FUSE operations (e.g. parallel stat calls during `git status`) can both see `restored=false` and both start restoring simultaneously. The second `extractGitArchive` overwrites files mid-write from the first → corrupted `.git` state.

Fix: `sync.Once` per workspace runtime, or a per-workspace restore mutex held across the full restore operation.

qiffang (changes requested):
## PR #471 review — REQUEST_CHANGES

I independently reviewed the fast-clone workspace path at `a5c1575` and agree the current PR should not merge yet. The existing inline comments on B1-B3 are valid; I am adding additional blockers below.

Existing inline blockers I verified:

- **B1** `ensureGitWorkspaces` silently drops workspaces on `ListGitTree` / overlay load failure and still marks the cache loaded.
- **B2** directory rename is implemented as O(n) sequential HTTP mutations inside one `rename(2)`.
- **B3** unknown `size_bytes=-1` is exposed to the kernel as `st_size=0`, and read-only materialization does not update the inode size.

Additional blockers:

### B4 — `.git` restore is not idempotent or concurrency-safe

`pkg/fuse/git_workspace.go:765-780` treats any existing local `.git` directory as fully restored and sets `rt.restored=true`. But the actual restore sequence is multi-step: remote clone (`805-848`), object pack restore (`851-887`), then checkpoint extraction (`796-797`). If the first access successfully creates `.git` and then object-pack download, unpack, or archive extraction fails, the next access sees `.git` and permanently skips the missing checkpoint/local-only object restore. That loses staged objects, local refs, index/log state, and can make sandbox replacement report a clean-but-wrong repo.

There is also no per-workspace restore lock/singleflight. Two concurrent reads of different clean blobs can both enter `ensureGitStateRestored`, race through `RemoveAll(gitDir)` / `Rename(tmpGitDir, gitDir)`, and corrupt the same local `.git` tree.

Required fix: use a per-workspace restore mutex/singleflight plus a restore-done sentinel written only after clone + object packs + checkpoint extraction all succeed. Do not treat a pre-existing `.git` directory as restored unless that sentinel or an equivalent verified state matches the workspace/head. Add regressions for failed restore-after-clone retry and concurrent first reads.

### B5 — directory rename can leave a partially renamed tree on failure

`pkg/fuse/git_workspace.go:2149-2232` performs directory rename as many independent overlay upserts followed by many independent whiteouts. If any request fails or the FUSE context is cancelled mid-loop, the backend is left with a partial new tree and part of the old tree still visible. POSIX `rename` failure must not corrupt the namespace.

This is separate from the performance issue in B2. Either implement a server-side batch/transactional overlay mutation for directory rename, or reject Git workspace directory renames with a deterministic error until they can be atomic. Add a failure-injection regression that proves a failed directory rename leaves the old tree unchanged.

### B6 — credential-bearing repo URLs are persisted and written to local metadata

`cmd/drive9/cli/git.go:146-149` sends the raw clone URL to the server, `pkg/server/git_workspace.go:205-209` persists it, and `pkg/gitcache/cache.go:529-540` writes it into `hydrate.json`. If the user clones with a common HTTPS credential form such as `https://x-access-token:TOKEN@github.com/org/repo.git`, the PAT is stored in tenant DB and local cache metadata. Restore/hydrate errors can also surface that URL through git stderr/logging.

Required fix: split credentials from repository identity. Store and log only a sanitized repo URL with `User` stripped; pass credentials through env/credential-helper/secret-store for the current process. Add tests for token URLs proving DB/API responses and hydrate metadata do not contain the token.

Validation I ran locally:

- `go test ./cmd/drive9/cli ./pkg/gitcache -count=1` — pass
- `go test ./pkg/fuse -count=1` — pass
- `go test ./pkg/fuse -run 'TestGit|Test.*Git|Test.*Fast|Test.*LocalOnly|Test.*Ignore|Test.*Rename' -count=1` — pass
- `go test ./pkg/client -count=1` — blocked locally: rootless Docker/testcontainers unavailable
- `go test ./pkg/datastore ./pkg/server -run 'TestGit|Test.*ObjectPack|TestReadDirectoryReturns404' -count=1` — blocked locally: rootless Docker/testcontainers unavailable
- `go test ./pkg/tenant/schema -run 'TestGitWorkspaceTiDBSchemaStatementsExecuteInMySQL|TestInitTiDBTenantSchemaStatementsForModeIncludesVault' -count=1` — blocked locally: rootless Docker/testcontainers unavailable

Verdict: **REVISE**.

pkg/fuse/git_workspace.go
qiffang: 🔴 **adversary-2 B1 — TOCTOU race**: Lock acquired here to check `rt.restored`, then released. The expensive restore (GetGitState HTTP, git clone, extractGitArchive) runs WITHOUT the lock. Two concurrent FUSE ops both see `restored=false`, both start restoring — second extractGitArchive overwrites files mid-write → corrupted .git. Fix: sync.Once per workspace runtime.

pkg/fuse/git_workspace.go
```diff
@@ -0,0 +1,2289 @@
...
+		fs.maybeStartGitWorkspaceHydrate(rt)
+	}
+	return nil
+}
+
+func (fs *Dat9FS) gitWorkspaceForPath(localPath string) (*gitWorkspaceRuntime, string, bool) {
+	if fs == nil || fs.git == nil || fs.opts == nil || !fs.opts.EnableGitWorkspaces {
+		return nil, "", false
+	}
+	ctx, cancel := context.WithTimeout(context.Background(), fuseTimeout)
```
qiffang: 🔴 **adversary-2 B2 — context.Background() discards caller cancellation**: Every FUSE operation that touches a git workspace calls this function, which creates a fresh `context.Background()` instead of forwarding the caller's context. FUSE interrupts, unmount signals, and process shutdown cannot cancel these in-flight HTTP calls. Same pattern at L247 (gitIgnoredPathLocalOnly) and L931 (readGitCleanFile hydrate). Under cascading timeout pressure (many concurrent FUSE ops), these independent background contexts can exhaust the HTTP connection pool. Fix: accept ctx from caller and derive timeout from it.

pkg/gitcache/cache.go
```diff
@@ -0,0 +1,685 @@
...
+		}
+		rel, ok, err := stripCodeloadRoot(hdr.Name)
+		if err != nil {
+			return files, bytes, err
+		}
+		if !ok {
+			continue
+		}
+		target := filepath.Join(dst, filepath.FromSlash(rel))
+		if !strings.HasPrefix(target, dst+string(filepath.Separator)) && target != dst {
```
qiffang: 🔴 **adversary-2 B3 — symlink traversal in extractCodeloadTar**: This function checks that `target` has the `dst` prefix (L384), and validates symlink targets via `safeSymlinkTarget` (L411). But it does NOT check intermediate path components for existing symlinks created by earlier tar entries.

Attack: tarball entry 1 creates symlink `foo → /etc`, entry 2 writes regular file `foo/bar`. The prefix check at L384 sees `dst/foo/bar` starts with `dst` — passes. But `os.MkdirAll(filepath.Dir(target))` at L393 follows the symlink through `foo → /etc`, creating `/etc/bar`.

By contrast, `extractGitArchive` (same PR, git_workspace.go L1059) has the explicit `ensureNoGitArchiveSymlinkTraversal` check that Lstats each intermediate component. The two extraction functions have asymmetric security posture for the same threat class.

Fix: Apply the same intermediate-symlink check from extractGitArchive, or use os.Lstat on each directory component before writing.

mornyx:
@qiffang I went through your review comments and pushed fixes in f666ebf.

Covered items:

- B1 workspace refresh: `ensureGitWorkspaces` now treats tree/overlay load failures as refresh errors, logs them, and does not replace the previous loaded snapshot or advance `loadedAt` on partial failure. Added a regression test for preserving the previous workspace snapshot.
- B2 directory rename RPC fan-out: documented the v1 limit and added a 100-entry guard for Git workspace directory rename. Above the cap, FUSE returns `EXDEV` before issuing overlay mutations. Added a regression test that verifies no overlay writes happen after the guard trips.
- B3 unknown `size_bytes=-1`: clean Git reads now update the inode and runtime tree size once the actual blob length is known, so subsequent attrs do not keep reporting size 0. Added coverage to the unknown-size writable-open test.
- Restore TOCTOU race: added a per-workspace restore mutex and moved the restored check/set to the runtime lock so concurrent FUSE ops cannot restore/extract `.git` concurrently.
- Context cancellation: `gitWorkspaceForPath`, ignored-path policy checks, clean Git reads, and FUSE call sites now forward caller context instead of creating detached background contexts. Size-aware write/checkpoint paths still keep their longer timeout, but now also listen to the FUSE cancel channel.
- Codeload tar symlink traversal: `extractCodeloadTar` now rejects existing symlink components before creating directories/files, matching the Git archive extraction posture. Added a tar symlink traversal regression test.

Validation run:

- `env GOCACHE=/Users/mornyx/Desktop/repos/github.com/mem9-ai/drive9/.codex-gocache go test ./pkg/gitcache`
- `env GOCACHE=/Users/mornyx/Desktop/repos/github.com/mem9-ai/drive9/.codex-gocache go test ./pkg/fuse -run "TestGitWorkspace|TestEnsureGitWorkspaces|TestExtractGitArchive|Test.*Rename"`
- `env GOCACHE=/Users/mornyx/Desktop/repos/github.com/mem9-ai/drive9/.codex-gocache go test ./pkg/fuse`
- `git diff --check`

qiffang (changes requested):
Round 2 verdict: **REVISE**.

The latest push fixes some previous items (workspace refresh now fails closed on partial load, archive extraction checks symlink traversal, and the FUSE paths pass caller contexts more consistently), but this PR is still not mergeable.

Findings:

1. **Blocker: CI is red.** The current `ci` check fails `make lint` on `pkg/fuse/git_workspace.go:1220`: `func (*Dat9FS).checkpointGitStateForPath is unused`. This alone blocks merge.

2. **Blocker: credential-bearing repo URLs are still persisted and checkpointed.** `cmd/drive9/cli/git.go:147-150` still sends the raw user-supplied `repoURL` into `git_workspaces`, and `pkg/gitcache/cache.go:1023-1066` writes the same raw URL into `hydrate.json`. More importantly, `archiveLocalGitStateDir` only filters `objects/` (`pkg/fuse/git_workspace.go:1687-1689`, `1758-1765`), so `.git/config` is included in `git_workspace_git_state`; a clone from `https://token@github.com/org/repo.git` will persist that credential in both the backend checkpoint and local hydrate metadata. Extracting the token for GitHub API use is not enough; the canonical persisted/restored remote URL and all metadata/log/checkpoint paths must be credential-free, and there needs to be a regression that proves neither DB requests nor checkpoint archives contain the token.

3. **Blocker: `.git` fsync/release reports success before the remote checkpoint is durable.** For local `.git` handles, `Flush`, `Fsync`, and `Release` sync the local file and then call `scheduleGitStateCheckpoint` (`pkg/fuse/dat9fs.go:5231-5247`, `5473-5484`, `5738-5765`). That goes through a 2s debouncer and returns immediately (`pkg/fuse/debounce.go:38-63`); only graceful `FlushAll` drains it (`pkg/fuse/dat9fs.go:6423-6425`). If the sandbox/process is killed after `git add`/`git commit`/`fsync` but before the timer fires, the local `.git` change is lost for replacement sandboxes even though FUSE already returned OK. Either `.git` fsync/release needs to synchronously publish the checkpoint (or expose a clear non-durable contract and not claim FUSE durability), and this needs a kill-before-debounce regression.

4. **Blocker: `.git` restore still treats any existing `.git` as complete.** `ensureGitStateRestored` returns success and sets `rt.restored=true` solely because `os.Stat(gitDir)` succeeds (`pkg/fuse/git_workspace.go:858-867`). If a prior restore got interrupted after `restoreGitObjectsFromRemote` moved a cloned `.git` into place but before object packs/checkpoint extraction completed, the next mount will skip restoration permanently and serve an incomplete Git state. The previous race is reduced by `restoreMu`, but the idempotency bug remains. Restore needs an atomic/sentinel-based completion marker and validation before setting `restored=true`; partial `.git` directories must be removed/repaired, not accepted.

5. **Blocker: directory rename still performs non-atomic multi-RPC mutation for normal-sized trees.** The new `gitDirectoryRenameOverlayLimit` only returns `EXDEV` for trees over 100 entries. For <=100 entries, `renameGitDir` still creates the destination directory/files and then whiteouts the source with many sequential `putGitOverlay` calls (`pkg/fuse/git_workspace.go:2292-2380`). A mid-loop 500/cancel leaves a mixed namespace (some new entries present, old entries not whiteouted, or old whiteouts after partial new writes). This was the core partial-failure issue; bounding the loop does not make it atomic. Either directory rename must be one backend transaction/batch mutation, or return `EXDEV` for all Git workspace directory renames until that exists. Add a regression that injects failure after one child mutation and proves no partial namespace is committed.

Validation I ran locally:

- `go test ./cmd/drive9/cli ./pkg/gitcache -count=1` ✅
- `go test ./pkg/fuse -count=1` ✅
- `go test ./pkg/fuse -run 'TestGit|Test.*Git|Test.*Fast|Test.*LocalOnly|Test.*Ignore|Test.*Rename' -count=1` ✅
- `go test -c ./pkg/fuse ./cmd/drive9/cli ./pkg/client ./pkg/server` ✅
- DB/testcontainer-backed package tests are still blocked locally by missing rootless Docker.

qiffang (approved):
## PR #471 Round 3 — adversary-1 review

Verdict: **GREEN / APPROVE**.

No blocking findings in the current head `d9f0add820565d6164a9200bc142bb776cb6b60c`. I re-reviewed the latest code against my five R2 blockers and re-checked the related R1 issues from adversary-2.

### R2 blocker verification

1. **B1 — lint/compile red from unused `checkpointGitStateForPath`: fixed.**
   `checkpointGitStateForPath` is now reachable through `checkpointGitStateAfterLocalWrite`, and the focused packages compile/test cleanly.

2. **B2 — credential-bearing repo URL persisted: fixed.**
   `gitcache.SanitizeRepoURL` is used before workspace registration/storage, hydrate metadata writes sanitize `repo_url`, and `.git/config` checkpoint archiving applies `SanitizeGitConfigCredentials`. Tests cover URL sanitization and credential redaction from archived config.

3. **B3 — `.git` fsync/release durability: acceptable now.**
   Writable `.git` `Fsync` always calls synchronous checkpointing. `Flush`/`Release` synchronously checkpoint in strict mode and otherwise schedule a coalesced checkpoint that is drained on unmount. That matches the documented non-strict durability tradeoff; it is no longer silently reporting strict durability while only scheduling background work.

4. **B4 — partial `.git` restore can be mistaken complete: fixed.**
   Restore is serialized by `restoreMu`, builds into a temporary `.git`, validates with `gitDirLooksUsable`, then atomically renames into place. Existing local `.git` is only reused if it has non-empty HEAD and `git cat-file -e HEAD^{commit}` succeeds with `GIT_NO_LAZY_FETCH=1`. Regression coverage includes replacing invalid local git state.

5. **B5 — git workspace directory rename partial multi-RPC commit: fixed by refusal.**
   `renameGitDir` now returns `EXDEV` for directory renames and writes no overlay entries. That is stricter than the design's possible capped tree rename path, but it is safe for this PR because callers can fall back to ordinary recursive copy/delete behavior.

### Additional checks

- Prior context-propagation and tar/symlink traversal blockers remain fixed.
- Unknown-size clean git files no longer behave as empty files on writable open; read materialization updates the inode/runtime size before append/write paths depend on it.
- CI is green: `ci`, `drive9-server-local e2e smoke`, and CodeRabbit all pass.

### Local validation

Passed:

```bash
/usr/local/go/bin/go test ./cmd/drive9/cli ./pkg/gitcache ./pkg/fuse -count=1
/usr/local/go/bin/go test ./pkg/fuse -run 'GitWorkspace|GitState|GitDir|Rename|Checkpoint|Restore|Credential|SizeBytes|UnknownSize' -count=1 -v
/usr/local/go/bin/go test ./pkg/gitcache -run 'Extract|Symlink|Credential|Hydrate|Sanitize' -count=1 -v
/usr/local/go/bin/go test -c ./pkg/fuse ./pkg/gitcache ./cmd/drive9/cli ./pkg/client ./pkg/server
```

Environment-blocked locally, not PR-blocking because GitHub CI covers it:

```bash
/usr/local/go/bin/go test ./pkg/datastore ./pkg/server -run 'TestGit|Test.*ObjectPack|TestReadDirectoryReturns404' -count=1
/usr/local/go/bin/go test ./pkg/tenant/schema -run 'TestGitWorkspaceTiDBSchemaStatementsExecuteInMySQL|TestInitTiDBTenantSchemaStatementsForModeIncludesVault' -count=1
```

Both panic at test startup with `rootless Docker not found` in local testcontainers setup.

Residual non-blocking risk: after credential sanitization, private-repo restore depends on ambient Git/GitHub credentials in the replacement sandbox. That is the right security posture for this PR, but product docs/setup should make the requirement explicit.

PR #498, opened 2026-06-05
fix(posix): improve fuse posix semantics

mornyx:
## Summary
- Improves FUSE POSIX metadata handling for chmod/chown-visible state, permission checks, parent directory times, hardlink ctime, symlink timestamps, and open-but-unlinked file handles.
- Adds mount-session metadata support for uid/gid/mode/nlink/ctime so pjdfstest-visible state survives local FUSE operations.
- Emulates FIFO, block, char, and socket nodes as metadata-only FUSE entries so compatibility tests can validate mode/type behavior without requiring real backend device-node support.
- Reworks rename preflight and local metadata updates for sticky directories, type mismatch errors, empty/non-empty directory replacement, metadata-only special nodes, hardlinks, symlinks, zero-byte file fallback, ctime, and directory nlink maintenance.
- Fixes tenant schema repair for long path columns and legacy upload active-path hash columns on existing TiDB tenants.
- Allows long aggregate FUSE paths while keeping per-component `NAME_MAX` enforcement.
- Keeps migrate test fixtures aligned with hashed `file_nodes` path indexes used by the current datastore.

## Validation
- `go test ./pkg/fuse`
- `go test ./pkg/migrate`
- `make test`
- `make build`
- `bash -n e2e/layer-fs-smoke-test.sh`
- Dev CD deployment of `feat/posix` at `f351ae2` succeeded.
- Remote EC2 pjdfstest on `drive9-posix-e2e-20260603`: `237/237` files, `8789/8789` subtests, `Result: PASS`.

## Before / After
| Metric | Before PR | After PR | Change |
|---|---:|---:|---:|
| `.t` files passed | `129/237` | `237/237` | `+108` |
| Subtests passed | `3510/8789` | `8789/8789` | `+5279` |
| Failed subtests | `5279` | `0` | `-5279` |
| Rename files passed | `11/25` | `25/25` | `+14` |
| Rename subtests passed | `1859/4857` | `4857/4857` | `+2998` |

Device-node cases are handled as metadata-only compatibility entries for pjdfstest rather than real backend block/char device support.

## pjdfstest Matrix
| Category | Files Passed | Subtests Passed | Failed Subtests |
|---|---:|---:|---:|
| chflags | 14/14 | 14/14 | 0 |
| chmod | 13/13 | 327/327 | 0 |
| chown | 11/11 | 1497/1497 | 0 |
| ftruncate | 15/15 | 89/89 | 0 |
| granular | 7/7 | 7/7 | 0 |
| link | 18/18 | 359/359 | 0 |
| mkdir | 13/13 | 118/118 | 0 |
| mkfifo | 13/13 | 120/120 | 0 |
| mknod | 12/12 | 186/186 | 0 |
| open | 26/26 | 328/328 | 0 |
| posix_fallocate | 1/1 | 1/1 | 0 |
| rename | 25/25 | 4857/4857 | 0 |
| rmdir | 16/16 | 145/145 | 0 |
| symlink | 13/13 | 95/95 | 0 |
| truncate | 15/15 | 84/84 | 0 |
| unlink | 15/15 | 440/440 | 0 |
| utimensat | 10/10 | 122/122 | 0 |
| **Total** | **237/237** | **8789/8789** | **0** |


qiffang (changes requested):
## R1 review — REQUEST_CHANGES

I agree with the macOS supplementary-group blocker already called out by adversary-2. I found additional correctness blockers in the POSIX namespace paths. CI being green is not enough here because these are stale-cache / concurrent-client / open-handle semantics that the current tests do not cover.

### Blockers

1. **macOS supplementary-group checks deny valid access**

`processHasSupplementaryGroup` reads `/proc/<pid>/status` and returns false when `/proc` is unavailable. On macOS/macFUSE, that makes `Access`, chmod SGID stripping, and chgrp validation deny legitimate callers whose access comes from supplementary groups. This PR is POSIX/FUSE behavior and macOS is a supported dev target, so this needs a platform-aware implementation or explicit safe fallback.

Relevant code: `pkg/fuse/dat9fs.go:5841-5868`.

2. **Symlink conflict recovery can delete another client's object**

`retrySymlinkAfterNegativeCacheConflict` handles a conflict after a stale negative dentry by statting `childP`, deleting whatever now exists there, and then creating the symlink. If another client created a file or empty directory after this mount cached ENOENT, POSIX `symlink()` should return `EEXIST`; this path instead removes that object and replaces it. That is data loss.

Relevant code: `pkg/fuse/dat9fs.go:6512-6532`.

3. **Metadata-only special-node create can shadow a real remote object**

`Mknod` for FIFO/char/block/socket skips the remote existence check when `hasNegativePathCache(childP)` is true, then installs only an in-memory special node. If another client created `childP` after the negative cache was recorded, this mount now shadows the real remote object until cache expiry/remount. Unlike regular create, there is no backend conditional create to arbitrate `EEXIST`.

Relevant code: `pkg/fuse/dat9fs.go:6224-6245`.

4. **Rename replacement is not all-or-nothing for directory and special-node targets**

`prepareRenameDirReplacement` deletes the destination directory and purges local caches before the later rename is attempted. If the subsequent rename fails because the source changed or the backend has a transient error, the syscall returns failure after the target directory has already been removed.

The same issue exists for special targets: `Rename` removes `newInfo.special` from local indexes before the remote rename runs. A failed remote rename then loses the FIFO/device/socket despite returning an error.

Relevant code: `pkg/fuse/rename_posix.go:296-317` and `pkg/fuse/dat9fs.go:7369-7372`.

5. **Rename-overwrite drops the overwritten target's open-handle inode**

`InodeToPath.Rename` removes an existing `newPath` with `preserveIfLast=false`. For `rename(src, target)` where `target` is the last visible link to an open file, POSIX requires the old target to remain readable through its open fd. This call removes the target inode mapping instead of preserving it as an unlinked open entry.

Relevant code: `pkg/fuse/inode.go:513-517`.

6. **Path-based truncate still returns `EFBIG` above 64 MiB and rewrites in memory**

`applyRemoteTruncate` hard-caps `newSize > maxPathTruncateInMemoryBytes` with `EFBIG` and builds the replacement buffer in memory. That means `truncate(file, 128MiB)` fails even though the normal write path supports larger files, and shrinking/extending near the cap is still an in-memory rewrite. This is not POSIX truncate behavior.

Relevant code: `pkg/fuse/dat9fs.go:422` and `pkg/fuse/dat9fs.go:2169-2205`.

### Validation run

- `/usr/local/go/bin/go test ./pkg/fuse -run 'Test.*(POSIX|Posix|Special|Rename|Rmdir|Access|Chown|SetAttr|Unlink|Nlink|Truncate)' -count=1` passed.
- `/usr/local/go/bin/go vet ./pkg/fuse ./pkg/datastore ./pkg/tenant/schema` passed.
- `/usr/local/go/bin/go test ./pkg/datastore ./pkg/fuse -count=1` could not complete in this local environment because `pkg/datastore` test setup panicked on `rootless Docker not found`; `pkg/fuse` completed successfully in that combined run.

mornyx:
Addressed qiffang review in b1b5a67.

Changes:
- Added Darwin-aware supplementary group lookup for macOS/macFUSE callers while keeping /proc parsing on Linux.
- Revalidated metadata-only special node creation/link targets even when negative path cache is present.
- Changed stale negative-cache symlink conflict recovery to return EEXIST instead of deleting the remote target.
- Moved empty-directory replacement into the datastore RenameDir transaction and removed the FUSE pre-delete path.
- Delayed special target cleanup until after successful remote rename.
- Preserved overwritten target inodes as unlinked entries so open handles keep reading the old object.
- Reworked path truncate over 64MiB to stream via multipart ReaderAt instead of full-object/in-memory rewrite, with overflow sizes still returning EFBIG.

Validation:
- go test ./pkg/fuse
- make test TEST_PKGS=./pkg/datastore
- make test TEST_PKGS=./pkg/backend
- go test ./cmd/drive9/cli
- make build
- make lint

GitHub CI is green for b1b5a67: code-ci and drive9-server-local e2e smoke both passed.

mornyx:
Update after the latest fix:

- Added `7ab2113` to make the local S3 mock stream multipart part writes, multipart completion, PutObject, UploadPartCopy, and byte-range GETs instead of materializing full objects in memory. This fixes the local `drive9-server-local` validation path for pjdfstest `open/25.t` style 2GiB+ sparse files.
- Re-ran local validation:
  - `go test ./pkg/s3client` passed
  - `go test ./pkg/fuse` passed
  - `make lint` passed
  - `make build` passed
- GitHub checks on `7ab2113` are green:
  - `ci` passed
  - `drive9-server-local e2e smoke` passed
  - CodeRabbit passed

I also verified that the dev ELB still shows the old backend rename behavior: the rename directory pjdfstest still has the same 40 failures there, which means dev has not yet deployed the backend part of this branch. I am now rerunning full pjdfstest on the EC2 box against a local server built from the latest commit.

qiffang (changes requested):
## PR #498 Review — R2 (adversary-1)

Verdict: **REVISE / REQUEST_CHANGES**.

Most R1 blockers are closed: stale-negative symlink conflicts no longer delete the remote target, metadata-only `mknod` revalidates negative cache, rename-overwrite now preserves the destination open-handle inode, and non-layer large path truncate streams instead of returning `EFBIG`.

Remaining blockers:

1. **Large clean open-unlink handles still become unreadable.**
   `markOpenHandlesUnlinked` only snapshots remote data when `size <= maxPathTruncateInMemoryBytes` (`pkg/fuse/dat9fs.go:5283-5296`). For a clean remote file larger than 64 MiB, it marks the handle unlinked with only `UnlinkedSize` (`pkg/fuse/dat9fs.go:5305-5314`). Later `Read` sees `Unlinked=true` without a snapshot and returns `EIO` for offsets inside the file (`pkg/fuse/dat9fs.go:8815-8846`). POSIX requires an already-open fd to remain readable after `unlink`, independent of file size. This needs either a streaming retained read target / local spill snapshot or a defined durable snapshot mechanism, plus a >64 MiB regression.

2. **Layer truncate still rejects valid large truncates.**
   `applyRemoteTruncate` returns `EFBIG` whenever `fs.layerEnabled()` and `newSize > maxPathTruncateInMemoryBytes` (`pkg/fuse/dat9fs.go:2179-2181`). The non-layer path was fixed to stream via multipart, and layers already have an object upload API (`UploadFSLayerFile`), so layer mounts now have a different POSIX contract for the same `truncate(2)`. A coding-agent/layer workspace can write large files but cannot truncate/extend them past this threshold. Add a layer large-truncate path and regression, or explicitly scope PR #498 away from layer mounts and block layer use for this feature.

3. **macOS supplementary-group check is still not a process credential check and runs shell commands in the FUSE hot path.**
   The Darwin fallback maps `pid -> user` with `ps` and then runs `id -G <user>` (`pkg/fuse/dat9fs.go:5922-5935`). That returns account group membership, not the target process's active supplementary group set, so a process that has dropped groups can still be granted group permissions. It also forks two external commands with no context/timeout from permission-sensitive paths (`access`/`chmod`/`chown`), which can block FUSE operations. This closes the original false-deny only partially and opens a false-allow/perf failure mode. Use a real per-process credential source or a conservative, explicitly tested fallback.

Validation run locally on head `7ab21138`:

- `go test ./pkg/fuse -run 'Test.*(Special|Rename|Symlink|Mknod|Truncate|Unlink|Unlinked|OpenDestination|Supplementary|Access)' -count=1`
- `go test ./pkg/fuse ./pkg/s3client -count=1`
- `go vet ./pkg/fuse ./pkg/s3client ./pkg/datastore ./pkg/tenant/schema`

All passed. The remaining issues are semantic/coverage gaps, not current test failures.

mornyx:
All qiffang R2 items have been addressed and pushed to head `f01c4f9`.

Changes:
- Large-file clean open-unlink: `markOpenHandlesUnlinked` now streams clean remote handles above the memory threshold into `ShadowStore`; each handle records `UnlinkedShadowGen` and pins that generation. `Read` uses `ReadAtGen` for unlinked clean handles, and `Release` unpins the generation. This prevents open file descriptors above 64MiB from turning into `EIO` after unlink.
- Layer large truncate: removed the `EFBIG` branch for `newSize > 64MiB` under layers. Added a layer truncate upload path that uses `truncateReaderAt` to preserve the existing prefix and zero-fill extensions, writes the layer object through `UploadFSLayerFile`, and records pending layer metadata.
- macOS supplementary groups: removed the Darwin FUSE hot-path `ps`/`id` shell fallback. The code now uses real per-process `/proc` groups when available, falls back to the current process via `os.Getgroups`, and otherwise conservatively returns false. This avoids false-allow behavior and fork/blocking risk.
- CI fix: merged latest `main` to resolve the PR conflict. The git fixture bare clone now uses `--no-local` to avoid runner-local object-copy fixture failures. Kept the git-ops tree fixture, status settle logic, and workspace refresh budget guard from `main`.

Added/covered tests:
- `TestUnlinkPreservesLargeOpenReadHandleSnapshotInShadow`
- `TestLayerSetAttrLargeTruncateUploadsLayerObjectFromShadow`
- `TestTruncateReaderAtPreservesPrefixAndZeroFillsExtension`

Local validation:
- `go test ./pkg/fuse -count=1` PASS
- `go test ./cmd/drive9/cli -count=1` PASS
- `make lint` PASS
- `make build` PASS
- `bash -n e2e/git-ops-smoke-test.sh` PASS
- `python3 -m py_compile e2e/git_fixture.py` PASS
- `python3 e2e/git_fixture.py /tmp/drive9-fixture-test --force --tree-files 128` PASS
- `git diff --check` PASS

GitHub checks on `f01c4f9`:
- `ci` SUCCESS
- `drive9-server-local e2e smoke` SUCCESS
- `CodeRabbit` SUCCESS

mornyx:
qiffang review items have been addressed and pushed to `feat/posix`. Current PR head: `f01c4f975db9c184726b71c48523d47fd047a38d`.

Fix summary:
- Large clean open-unlink now uses `ShadowStore` for per-handle snapshots, so an unlinked large-file read handle no longer depends on the remote object lifecycle.
- Layer large truncate now uses a shadow-backed upload path, supports truncates above 64MiB, and no longer returns `EFBIG` for that case.
- Remote large-object snapshot/truncate paths avoid reading the full object into memory.
- Added focused FUSE tests and truncate reader coverage.
- The git e2e fixture now uses `git clone --bare --no-local` to avoid runner-local object-copy failures; also synced the git-ops smoke fixture coverage from `main`.
- Merged latest `origin/main` to resolve the PR conflict.

Validation:
- GitHub checks on `f01c4f9`: `ci` PASS, `drive9-server-local e2e smoke` PASS, CodeRabbit PASS.
- Local: `go test ./pkg/fuse -count=1` PASS; `go test ./cmd/drive9/cli -count=1` PASS; `make lint` PASS; `make build` PASS; `python3 e2e/git_fixture.py ... --tree-files 128` PASS; `bash -n e2e/git-ops-smoke-test.sh` PASS.
- EC2 `drive9-posix-e2e` pjdfstest: PASS, `8789/8789` cases and `237/237` files. Report: `/home/ubuntu/drive9-posix-pjdfstest-20260612T111424Z-f01c4f9/reports/posix-feature-report-20260612-111456.md`.
- Dev real e2e: full `e2e/smoke-all.sh` rerun PASS, `PASS=9 FAIL=0`. Run dir: `/home/ubuntu/drive9-dev-real-e2e-rerun-20260612T121950Z-f01c4f9`. Key sub-results: `git-ops TOTAL=137 PASS=137 FAIL=0`, `git-workspace RESULT=182/182 passed`, `portable-pack-unpack PASS=29 FAIL=0`.

Note: the first dev smoke-all run had one intermittent failure at `kimi-code fast_worktree linked worktree ready after remount`. After cleaning the stale mount, I reran that exact matrix first and got `35/35` passing, then reran the full smoke-all suite and got `9/9` passing.

mornyx:
Thanks for the R2 review. I went through the remaining blockers again on the current head.

- Large clean open-unlink handles: addressed. The current implementation snapshots large clean remote handles into `ShadowStore` and pins the generation on the open handle, then serves later reads from that retained generation after the remote path is deleted. Covered by `TestUnlinkPreservesLargeOpenReadHandleSnapshotInShadow`.
- Layer large truncate: addressed. `applyRemoteTruncate` now routes large layer truncates through `applyLayerLargeTruncate`, which materializes/extends the object through the shadow store and uploads it with `UploadFSLayerFile`. Covered by `TestLayerSetAttrLargeTruncateUploadsLayerObjectFromShadow`.
- macOS supplementary-group fallback: tightened. The fallback no longer shells out and no longer uses account database membership. Linux still uses `/proc/<pid>/status`; when process credentials are unavailable, only the current process is checked through `getgroups`, while other pids conservatively return false to avoid false-allow behavior. I added explicit coverage for the parser and the conservative no-`/proc` fallback in `e98b36c`.

Validation run:

- `go test ./pkg/fuse -run "Test.*(Special|Rename|Symlink|Mknod|Truncate|Unlink|Unlinked|OpenDestination|Supplementary|Access|ProcessHas|ProcStatus|LayerSetAttrLargeTruncate)" -count=1`
- `go test ./pkg/fuse ./pkg/s3client -count=1`
- `go vet ./pkg/fuse ./pkg/s3client ./pkg/datastore ./pkg/tenant/schema`
- `make lint`
- `git diff --check`

srstack (commented):
## Code Review: fix(posix): improve fuse posix semantics

### Overview
A large POSIX correctness overhaul that takes pjdfstest from 129/237 → 237/237 files and 3510/8789 → 8789/8789 subtests. Changes span FUSE metadata handling (chmod/chown/ctime/nlink/symlink timestamps, sticky directories, hardlink ctime), rename preflight logic, special-node emulation (FIFO/block/char/socket), tenant schema repair for long-path columns, and e2e infrastructure improvements. The 100% pjdfstest pass rate is a strong validation signal.

---

### Strengths

- **pjdfstest matrix as ground truth**: 8789/8789 passing gives high confidence the POSIX semantics are correct rather than just plausibly correct.
- **path_hash index in datastore**: Adding `path_hash`/`parent_path_hash` columns alongside the full path is the right approach for long-path performance on TiDB where index key length is limited. Queries now hit the hash index first and the full path as a discriminator — correct and efficient.
- **macFUSE path detection** in both `pjdfstest-suite.sh` and `_feature-matrix-runner.sh`: the search across standard macFUSE install locations is a good quality-of-life fix.
- **`pjdfstest_mount_args`** cleanly encapsulates the Linux/Darwin `--allow-other` difference instead of scattering conditionals.
- **`stop_mount` fix**: switching from bare `drive9` to `$CLI_BIN` with explicit env vars is correct — the old form silently used whatever `drive9` was on PATH rather than the tested binary.
- **`git clone --no-local`**: prevents pack-file hardlinking in the bare fixture, which would make the fixture non-portable across filesystems. Correct fix.
- **TODO-aware `not ok` counting**: excluding TODO-marked failures from the failure count matches TAP semantics precisely.

---

### Issues

**1. `fileNodePathHash` is a thin wrapper with no documentation of collision behavior**

```go
func fileNodePathHash(path string) string {
    return StorageRefHash(path)
}
```

The hash is used as the primary index discriminator. If `StorageRefHash` truncates (e.g. first N bytes of SHA-256), two paths could share the same hash and both rows would be returned by the `path_hash = ? AND path = ?` query — which is correct. But if it ever produces a non-hex or variable-length output, the `len(sum) != sha256.Size*2` style guards elsewhere would not apply here. A short comment stating what `StorageRefHash` returns (length, encoding, collision properties) would make the correctness argument self-contained.

**2. Schema migration not visible in this diff**

`InsertNode` now writes `path_hash` and `parent_path_hash`, and queries filter on them. If existing databases don't have those columns, all reads/writes will fail at runtime. The PR description mentions "tenant schema repair for long path columns" — but the migration code isn't in the visible portion of the diff. Confirm the migration runs before any FUSE operation and covers both new tenants and existing TiDB tenants that predate these columns.

**3. `pjdfstest-suite.sh` duplicates macFUSE PATH setup from `_feature-matrix-runner.sh`**

The same directory search loop appears verbatim in both files. Since `pjdfstest-suite.sh` immediately execs into `_feature-matrix-runner.sh`, the runner's own PATH setup will also fire — harmless, but the duplication will drift. Consider moving the canonical PATH setup into the runner and removing it from the entrypoint.

**4. `wait_layer_diff_file_mode` in `layer-fs-smoke-test.sh` uses `date +%s` in a loop**

```bash
local deadline=1781541464
...
if [ "$(date +%s)" -ge "$deadline" ]; then
```

This matches the existing pattern in the file, so it's consistent. Just note that on macOS `date +%s` works but is not POSIX; the script already targets Linux CI so this is fine.

**5. `fuse-write-perf-budget-test.sh`: `ROOT_REMOTE` mount path change**

```bash
-  drive9 mount ... "$MOUNT_POINT"
+  drive9 mount ... ":$ROOT_REMOTE" "$MOUNT_POINT"
```

Mounting only the test's remote root is a sound isolation fix. Verify that the perf counter assertions downstream (expected remote write/stat/list counts) still hold — scoping the mount to a subdirectory could change baseline counter values if the implementation counts at the mount root level.

---

### Minor

- `permissionModeMask uint32 = 0o7777` is added to `store.go` but not visibly used in this diff fragment. If it's used elsewhere in the PR, ensure it's exercised by a test.
- `failed_file_count` fix (`sum(1 for _, failed in ... if failed > 0)`) is more correct than the old `len(failed_files)` — files with zero failures were previously counted as failed if they appeared in the map at all. Good catch.
- The `pjdfstests.sh` alias (one-liner delegating to `pjdfstest-suite.sh`) is fine for backward compat. No issues.

---

### Verdict

Solid PR with impressive test results. The two items that warrant a response before merge are: (1) confirming the schema migration path for the new hash columns, and (2) verifying the perf budget counter baselines after the `ROOT_REMOTE` mount scoping change. Everything else is polish.

qiffang (dismissed):
## adversary-1 · PR #498 R3 follow-up · `e98b36c`

Verdict: **REVISE before merge**.

I did a narrow recheck against current `main` after PR #559. The POSIX review signal remains strong, but the current PR head is not mergeable:

```text
gh pr view 498 --json mergeStateStatus -> DIRTY
git merge --no-commit --no-ff origin/main
CONFLICT (content): pkg/fuse/remote_upload_test.go
```

The conflict is exactly where PR #559 added shadow-upload tier routing tests and PR #498 added truncate reader / upload tests. This must be resolved before merge, because a bad conflict resolution can silently drop the #559 regression coverage.

Required after rebase/conflict resolution:

- Keep PR #498's `truncateReaderAt` coverage for prefix preservation, zero-fill extension, and EOF behavior.
- Keep PR #559's shadow upload tier routing coverage: small shadow file direct PUT with committed revision, exact threshold multipart, and unknown threshold multipart.
- Re-run focused `pkg/fuse` tests plus CI/e2e on the rebased head.

I am not adding new semantic blockers beyond the merge conflict. Current head should not be merged while GitHub reports `DIRTY`.

mornyx:
@qiffang @srstack I pushed two commits addressing the latest review feedback.

qiffang R3 / merge conflict:
- Merged origin/main into feat/posix and resolved the conflicts in `pkg/fuse/dat9fs.go` and `pkg/fuse/remote_upload_test.go`.
- Kept PR #498 `truncateReaderAt` coverage for prefix preservation, zero-fill extension, and EOF behavior.
- Kept PR #559 shadow-upload tier coverage: small shadow spill direct PUT with committed revision, exact-threshold multipart, and unknown-threshold multipart.
- Integrated the main zero-size path-truncate staging path into the generic `applyRemoteTruncate` flow, and adjusted open-handle adoption to use the actual truncate size instead of hard-coding zero.

srstack:
- Added a `fileNodePathHash` comment documenting that it is a lowercase 64-character SHA-256 hex discriminator and that callers must keep the full path predicate so hash collisions cannot alias distinct paths.
- Confirmed the schema migration path: new tenant DDL includes the hash columns; existing TiDB tenants go through `EnsureTiDBSchemaForMode` / `repairMySQLPathHashSchema` / `BackfillPathHashes`, including dropping mismatched legacy path indexes before rebuilding hash indexes; db9 has the corresponding repair path.
- Removed the duplicated macFUSE PATH probing from `pjdfstest-suite.sh`; `_feature-matrix-runner.sh` remains the canonical setup point.
- Revalidated the FUSE write-perf sub-root mount isolation via the latest local-e2e run.

Validation:
- `go test ./pkg/fuse -run "Test(TruncateReaderAt|UploadFromShadowRemoteWithRevision|CommitQueueLargeOverwriteUsesMultipartUpload|WriteBackUploaderLargeNewFileUsesMultipartUpload)" -count=1`
- `go test ./pkg/fuse -run "Test.*(POSIX|Posix|Special|Rename|Rmdir|Access|Chown|SetAttr|Unlink|Unlinked|Nlink|Truncate|Supplementary|OpenDestination|Symlink|Mknod)" -count=1`
- `go test ./pkg/fuse ./pkg/s3client -count=1`
- `go vet ./pkg/fuse ./pkg/datastore ./pkg/tenant/schema`
- `bash -n e2e/pjdfstest-suite.sh e2e/_feature-matrix-runner.sh`
- `git diff --check`
- GitHub Actions on `1649d7d`: `ci` passed, `drive9-server-local e2e smoke` passed (7m21s), CodeRabbit passed.

srstack (commented):

srstack (commented):
## Review: fix(posix): improve fuse posix semantics

Large PR (7629 additions). pjdfstest goes from 129/237 files passing to 237/237 with zero failed subtests — a solid result. Reviewed the key modules below.

---

### Core logic

**`path_hash` / `parent_path_hash` index**

All path lookups in `file_nodes` and `uploads` now use `WHERE path_hash = ? AND path = ?`. The hash narrows the index scan; the full path predicate prevents aliasing on collision. `fileNodePathHash` reuses `StorageRefHash` (SHA-256 hex), which has negligible collision probability.

One query in `RenameDir` does not use `path_hash`:

```go
rows, err := tx.QueryContext(ctx, `SELECT node_id, path, parent_path, name FROM file_nodes
    WHERE path = ? OR path LIKE ? ORDER BY path FOR UPDATE`, oldPrefix, oldPrefix+"%")
```

This is a prefix scan (`LIKE`) so the hash index cannot help. Acceptable for a low-frequency operation held under `FOR UPDATE`, but a comment explaining why the hash is intentionally absent here would help.

**`permissionModeMask = 0o7777` replacing `0o777`**

Fixes setuid/setgid/sticky bits being truncated on chmod. This is the root cause of many pjdfstest `chmod` failures. `TestChmodPreservesSpecialPermissionBits` verifies `0o6755 → 0o106755` (setuid+setgid preserved). ✓

**`hasPOSIXAccess` implementation**

```go
if caller.Uid == 0 {
    if mask&gofuse.X_OK == 0 { return true }
    return mode&0o111 != 0
}
```

Root bypasses permission checks except for execute, where at least one execute bit must be set. This matches Linux `access(2)` semantics. ✓

**`Access()` with supplementary group lookup**

`processHasSupplementaryGroup` reads `/proc/<pid>/status` to check supplementary groups. `var readProcessStatusFile = os.ReadFile` allows test injection. Note: if the caller is in a different PID namespace or container, `/proc/<pid>/status` may not be visible or may show incorrect GIDs. For single-tenant FUSE mounts this is not a practical risk, but worth documenting as a known limitation.

**`RenameDir` atomicity**

The new logic checks if `newPrefix` exists, rejects if non-empty, deletes if empty, then proceeds with the rename — all inside one transaction with `FOR UPDATE`. `TestRenameDirReplacesEmptyTargetAtomically` and `TestRenameDirRejectsNonEmptyTargetWithoutDeleting` both verify that a failed rename leaves the source intact. ✓

One observation: the inline dst-children check in `RenameDir` duplicates the logic in `dirHasChildrenTx`. Not a correctness issue, but the two code paths could drift independently.

**Special node emulation (FIFO/block/char/socket)**

```go
specialByPath map[string]uint64  // path → inode, in-memory only
```

`entryIsMetadataOnlySpecial` causes `GetAttr`, `Lookup`, and `shouldPreserveForgottenInode` to short-circuit before any remote call. This is the right approach for pjdfstest compatibility. However, special nodes are not persisted — they vanish on remount. This is acceptable for test-harness use but should be documented with a comment so future maintainers know the behavior is intentional.

---

### Issues worth addressing

**`applyRemoteTruncate`: `EFBIG` check is effectively a no-op on 64-bit systems**

```go
if newSize > int64(int(^uint(0)>>1)) {
    return gofuse.Status(syscall.EFBIG)
}
```

On 64-bit systems, `int` is 64 bits, so `int64(int(^uint(0)>>1)) == math.MaxInt64`. Since `newSize` is already `int64`, this condition is always false. If the intent is to reject truncates beyond a practical size limit, use a concrete constant (e.g., `maxPathTruncateInMemoryBytes` or a filesystem-level ceiling). If the intent is to guard against theoretical overflow, add a comment explaining it.

**`updateOpenHandleBaseRevision` `truncateSize` parameter — no test coverage**

```go
func (fs *Dat9FS) updateOpenHandleBaseRevision(remotePath string, revision int64, callerPID uint32, truncateSize int64)
```

The new `truncateSize` parameter fixes a real bug: previously a post-`ftruncate` open-handle sync always truncated to 0 instead of `newSize`. This is a meaningful behavioral fix but has no dedicated unit test. A test exercising `ftruncate` to a non-zero size on a file with an open handle would close this gap.

**`e2e/fuse-write-perf-budget-test.sh`: `WORK_MOUNT` path change**

```bash
-WORK_MOUNT="$MOUNT_POINT/$ROOT_REL/work"
+WORK_MOUNT="$MOUNT_POINT/work"
```

The comment explains the intent (avoid cross-suite path pollution). The mount now covers only `$ROOT_REMOTE` instead of `:/`, so `MOUNT_POINT/work` maps to `$ROOT_REMOTE/work`. ✓

---

### Summary

No blocking issues. The three items worth following up:

1. **`EFBIG` check** — clarify intent or replace with a meaningful limit.
2. **Special node non-persistence** — add a comment so the behavior is explicit.
3. **`truncateSize` fix** — add a test for `ftruncate` to non-zero size with an open handle.

The pjdfstest results speak for themselves. Good work overall.

srstack (approved):
LGTM. pjdfstest 237/237 is a strong signal. Three minor follow-up items noted in the review comment above, none blocking.

mornyx:
Schema-compat validation completed on dev.

What I changed before deploying:
- Merged latest `origin/main` into `feat/posix` and pushed `f1c8b4a65365af074b704189723090872b3d87ae`.
- PR checks on that SHA passed:
  - `ci`
  - `drive9-server-local e2e smoke`

Pre-upgrade baseline:
- EC2: `drive9-posix-e2e-20260603`
- Dev endpoint: `http://k8s-dat9-dat9serv-d5e02e7d07-1645488597.ap-southeast-1.elb.amazonaws.com`
- Dev was still on main `ab13ad3` before the branch deployment.
- Created tenant `e5b5e20c-a771-479e-b89a-b8276fede2b6` before deploying this branch.
- Used the official `drive9.ai` CLI binary.
- Same-tenant pre-upgrade core e2e passed:
  - `api-smoke-test-existing-key.sh`: 4/4
  - `journal-smoke-test.sh`: 18/18
  - `fuse-smoke-test.sh`: 94/94
  - `git-ops-smoke-test.sh` with `coding-agent` fast/blobless: 51/51

Deployment:
- Triggered `dev-cd` from `main` with `deploy_ref=f1c8b4a65365af074b704189723090872b3d87ae`.
- Successful run: https://github.com/mem9-ai/drive9/actions/runs/27615086629

Post-upgrade validation:
- Reused the same pre-existing tenant `e5b5e20c-a771-479e-b89a-b8276fede2b6`.
- Built the CLI from `feat/posix` at `f1c8b4a65365af074b704189723090872b3d87ae`.
- Same-tenant post-upgrade core e2e passed:
  - `api-smoke-test-existing-key.sh`: 4/4
  - `journal-smoke-test.sh`: 18/18
  - `fuse-smoke-test.sh`: 94/94
  - `git-ops-smoke-test.sh` with `coding-agent` fast/blobless: 51/51

Conclusion: the schema changes in this branch are non-destructive for an existing dev tenant across the tested live paths. The tenant created before deploying this branch remained active and usable after deployment, including FUSE file operations, truncation/rename behavior, large-file tier transitions, and coding-agent Git fast/blobless workflows.

Non-blocking observation: the broader `layer-fs-smoke-test.sh` still hits `fs cp: backend unavailable` in the CLI layer section on this dev tenant, both before and after the branch deployment. The layer API sections before that point pass. I did not use that unstable layer CLI path as evidence for the schema-compat conclusion above.

qiffang (changes requested):
## adversary-1 R4 review — REQUEST_CHANGES

Current head reviewed: `f1c8b4a65365af074b704189723090872b3d87ae`.

Most prior blockers are resolved: the R1/R2 POSIX issues have regression coverage, the `remote_upload_test.go` conflict kept both PR #498 truncate coverage and PR #559 shadow-upload coverage, and the focused tests I ran pass.

### Blocker: active `tidbcloud_native` tenants can skip the required path-hash schema repair

This PR changes datastore queries/writes to require `file_nodes.path_hash`, `file_nodes.parent_path_hash`, `uploads.target_path_hash`, and `uploads.active_target_path_hash`. It adds repair/backfill code, but the active-tenant backend-open path only runs `ensureTiDBSchemaForAutoEmbeddingProfile` when:

```go
opts.DatabaseAutoEmbedding && (t.Provider == ProviderTiDBZero || t.Provider == ProviderTiDBCloudStarter)
```

in `pkg/tenant/pool.go`. `tidbcloud_native` is a supported TiDB provider (`UsesTiDBAutoEmbedding` and `SmallInDB` both include it), and its provisioner has `InitSchema` / `InitSchemaForAutoEmbeddingProfile`, but already-active `tidbcloud_native` tenants do not pass through `initTenantSchemaAsync` again. If one of those tenants predates the new hash columns, the first post-deploy filesystem operation can reach datastore SQL that references `path_hash` before any repair ran, yielding `column not found` on the data plane.

The same gate is also tied to `opts.DatabaseAutoEmbedding`, which is a runtime embedding toggle, while these hash columns are core filesystem schema. A TiDB tenant with database auto-embedding disabled can still need the filesystem schema repair.

Suggested fix: make the backend-open schema ensure cover every TiDB-backed tenant that can use this schema, not only Zero/Starter with runtime auto-embedding enabled. For example, key it off `UsesTiDBAutoEmbedding(t.Provider)` / provider schema support, resolve the tenant profile, run `ensureTiDBSchemaForAutoEmbeddingProfile`, and update `schema_version` after success. Add a regression that a `ProviderTiDBCloudNative` tenant with stale `SchemaVersion` calls the ensure hook before returning a backend.

### Validation run locally

- `go test ./pkg/fuse -run 'Test(ProcessHasSupplementaryGroup|TruncateReaderAtPreservesPrefixAndZeroFillsExtension|UploadFromShadowRemoteWithRevision|UnlinkPreservesLargeOpenReadHandleSnapshotInShadow|LayerSetAttrLargeTruncateUploadsLayerObjectFromShadow)' -count=1`\n- `go test ./pkg/tenant/schema -run 'Test(PlannedTiDBSchemaRepairs|RepairMySQLPathHashSchema|DiffTiDBTableMetaReportsLegacyUploadActiveTargetPath|IsSafeAddColumnRepairSQLAllowsUploadActiveTargetHash|ApplyTiDBSchemaRepairs)' -count=1`\n- `go test ./pkg/datastore -run 'TestRenameDir|TestStoreLongPath|Test.*PathHash|Test.*LongPath' -count=1`\n- `go test ./pkg/fuse -run 'TestRenameZeroByteRemoteFileToMissingTargetFallsBackToCreateDelete|Test.*Rename.*Replacement|Test.*Rename.*Open|Test.*Rename.*Special|Test.*Rename.*Dir|Test.*Rename.*Sticky' -count=1`\n- `go test ./pkg/fuse -run 'Test.*(Symlink|Mknod|Truncate|Unlink|Unlinked|Supplementary|Access)' -count=1`\n- GitHub checks observed green: `ci`, `drive9-server-local e2e smoke`, CodeRabbit.\n\nNon-blocking follow-up: `renameRemoteFileToMissingTargetFallback` still has only positive coverage. If the fallback creates the new zero-byte target and deleting the old source then fails, it can return an error while leaving both paths remotely. This is a recovery-path edge case, not my current merge blocker, but should get cleanup/coverage next.

Issue #14, opened 2026-03-26
Query API (POST /v1/query) not implemented — cannot filter files by tags/source_id/status


qiffang:
## Current State

Design doc §6 defines the Query API:

\`\`\`
POST /v1/query
{
  \"filter\": {
    \"source_id\": \"agent-007\",
    \"tags\": {\"env\": \"prod\"},
    \"created_after\": \"2026-03-24T00:00:00Z\",
    \"status\": \"CONFIRMED\"
  },
  \"order_by\": \"created_at\",
  \"cursor\": \"...\",
  \"limit\": 100
}
\`\`\`

Current code has no \`/v1/query\` endpoint and no tag CRUD operations.

The \`file_tags\` table **is already created** in the schema (meta.go:147), but:
- No InsertTag / DeleteTag / ListTags methods
- No tag-related HTTP endpoints
- No tag operations in the client SDK

## Required Work

- [ ] \`pkg/meta/meta.go\` — tag CRUD methods
- [ ] \`pkg/server/server.go\` — \`POST /v1/query\` handler
- [ ] \`pkg/client/client.go\` — \`Query()\` method
- [ ] Write endpoint support for tag setting (e.g. \`X-Dat9-Tags\` header or request body)

qiffang:
已通过 find API 实现：GET /v1/fs/?find&tag=key=value 支持按 tags 过滤文件。pkg/server/server.go L903-904 和 pkg/datastore/search.go 实现了完整的 tag 查询功能。虽然不是原设计的 POST /v1/query，但功能等价。

PR #503, opened 2026-06-06
test: cover backend tier transitions

qiffang:
## Summary
- add a backend regression for `10KiB inline -> 8MiB S3 -> 10KiB inline` overwrites
- assert visible backend reads/stat size match after each transition
- assert split `contents` and legacy `files` rows switch storage type/ref/checksum/blob correctly
- assert the 8MiB payload exists in local S3 with exact bytes and old S3 refs are no longer confirmed and are tracked as overwrite GC candidates

## Validation
- `/usr/local/go/bin/gofmt -w pkg/backend/tier_transition_test.go`
- `/usr/local/go/bin/go test -c -o /tmp/drive9-backend-task13.test ./pkg/backend`
- `git diff --check --cached`

## Not run locally
- `/usr/local/go/bin/go test ./pkg/backend -run TestBackendTierTransitionInlineToS3ToInlineCleansStorage -count=1 -v` is blocked on this machine by existing testcontainers setup: `panic: rootless Docker not found`. CI / a host with `DRIVE9_TEST_MYSQL_DSN` or container runtime must execute the runtime test.

## Summary by CodeRabbit

* **Tests**
  * Added comprehensive test coverage for file storage tier transitions to validate data integrity and state consistency across storage backends.

### DIFF
--- a/pkg/backend/tier_transition_test.go
+++ b/pkg/backend/tier_transition_test.go
@@ -1,0 +1,253 @@
+package backend
+
+import (
+	"bytes"
+	"context"
+	"database/sql"
+	"io"
+	"testing"
+
+	"github.com/c4pt0r/agfs/agfs-server/pkg/filesystem"
+	"github.com/mem9-ai/dat9/pkg/datastore"
+	"github.com/mem9-ai/dat9/pkg/meta"
+)
+
+func TestBackendTierTransitionInlineToS3ToInlineCleansStorage(t *testing.T) {
+	b, fake := newCentralQuotaBackend(t)
+	b.storageNamespaceID = "ns-tier-transition"
+
+	ctx := context.Background()
+	path := "/tier-transition.bin"
+	initialInline := deterministicTierPayload(10*1024, 0x11)
+	largeS3 := deterministicTierPayload(8*1024*1024, 0x42)
+	finalInline := deterministicTierPayload(10*1024, 0x7d)
+
+	if n, err := b.Write(path, initialInline, 0, filesystem.WriteFlagCreate); err != nil {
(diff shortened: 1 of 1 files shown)


qiffang:
Review verdict: **GREEN for code/test review**. I did not find blockers in this backend slice.

What I verified:
- The test covers the requested backend transition `10KiB inline -> 8MiB S3 -> 10KiB inline` on one stable file ID.
- Each phase checks backend-visible `Stat`/`Read` size and bytes, so metadata-only correctness is not enough to pass.
- Inline phases assert both split-table state and legacy `files` state are `db9`/`inline` with the expected content blob/checksum.
- The S3 phase asserts both split and legacy rows move to `s3`, clear inline `content_blob`, and the local S3 object contains exactly the expected 8MiB payload.
- The final inline overwrite asserts the old S3 ref is no longer confirmed content and that an overwrite GC candidate is enqueued for that ref/namespace.

Validation checked:
- GitHub `code-ci` and `local-e2e` are green on head `40ca6b956e44893b1f4651efe830fd0985fe19b4`.
- Locally: `git diff --check origin/main...HEAD`, `go test -c -o /tmp/drive9-pr503-backend.test ./pkg/backend`, and `go test -c ./pkg/... ./cmd/...` all pass in a clean worktree.

Scope caveat: this only proves the backend transition invariant. It does not cover the mounted FUSE path, threshold boundary matrix, append/truncate variants, or failure/GC split-brain cases; those correctly remain separate task #14/#15/#16 work.

Note: GitHub would not accept a formal APPROVE review from the current app identity (`Review Can not approve your own pull request`), so I am recording the review verdict as a PR comment.

PR #505, opened 2026-06-06
e2e: add FUSE SQLite correctness workload

qiffang:
## Summary
- add `e2e/fuse-sqlite-correctness.sh` as a real writable FUSE SQLite correctness workload
- wire rollback-journal SQLite correctness into the default `e2e/fuse-release-gate.sh`; keep WAL coverage opt-in via `RUN_FUSE_SQLITE_WAL=1`
- implement the FUSE write-path fixes required by SQLite: advisory lock handlers, committed revision propagation after shadow-spill commits, safe same-path handle refresh, and streaming ShadowSpill uploads without `ReadAll`
- workload verifies deterministic SQLite data, `PRAGMA integrity_check`, logical fingerprints, unmount/remount parity, and remote snapshot parity

## Current product signal
The original live FUSE failure (`sqlite3.OperationalError: disk I/O error` during SQLite writes) is fixed on head `5935be3`. GitHub `drive9-server-local e2e smoke` now runs the rollback-journal SQLite workload in the default release gate and passes `RESULT: 19/19 passed`.

This PR validates single-process SQLite correctness in rollback-journal mode by default. WAL mode remains opt-in with `RUN_FUSE_SQLITE_WAL=1`; concurrent SQLite workloads, crash/remount recovery, performance baselines, and GC sweep behavior remain separate follow-up workloads.

## Scope
Included:
- FUSE lock operations used by SQLite `fcntl` locking
- SQLite rollback-journal create/write/verify/remount/remote parity E2E
- revision refresh fixes for SQLite journal rewrites through ShadowSpill
- safety fixes for unmount failure propagation, same-path handle refresh, and ShadowSpill streaming

Not included:
- concurrent readers/writers
- forced crash/restart recovery
- performance baseline reporting
- GC sweep correctness under SQLite churn

## Validation
- `bash -n e2e/fuse-sqlite-correctness.sh e2e/fuse-concurrency-stress.sh e2e/fuse-release-gate.sh`
- focused FUSE regressions for lock handling, ShadowSpill streaming, Fsync revision refresh, same-path handle locking, and commit-queue spill recovery
- `/usr/local/go/bin/go test ./pkg/fuse ./cmd/drive9 -count=1`
- `/usr/local/go/bin/go test -c ./pkg/client ./pkg/fuse ./cmd/drive9`
- `git diff --check`
- `PATH=/usr/local/go/bin:$PATH make build-cli CLI_BIN=/tmp/drive9-task30-reviewfix-cli`
[...]

Issue #28, opened 2026-03-26
Test coverage gaps: concurrency safety, boundary conditions, error paths


qiffang:
## Current Coverage Analysis

Current tests cover basic happy paths:

| Package | Test File | Coverage |
|---------|-----------|----------|
| backend | dat9_test.go (292 lines) | CRUD, rename, zero-copy cp, offset write, dir ops |
| meta | meta_test.go (267 lines) | node/file CRUD, stat, listDir, rename, ref count |
| server | server_test.go (235 lines) | HTTP endpoints happy-path |
| client | client_test.go (158 lines) | SDK integration tests |
| pathutil | pathutil_test.go (126 lines) | canonicalize, parent, basename |
| cli | cp_test.go + sh_test.go | cp path classification, sh basic tests |

### 1. Concurrency safety ❌
- Concurrent writes to the same file
- Concurrent directory creation at the same path
- Concurrent delete + read on the same file
- Concurrent rename + read

### 2. Boundary conditions ❌
- Empty file write/read
- Extremely long path names
- Deeply nested directories (/a/b/c/d/.../z/file.txt)
- UTF-8 Unicode boundaries (emoji filenames, CJK characters, combining characters)
- Root directory \"/\" special operations (rm /, mv / /x)

### 3. Error paths ❌
- Read non-existent file → 404 (client test exists, but server/backend tests insufficient)
- Write when disk is full
- SQLite lock contention
- Invalid path characters (pathutil tests exist, but no end-to-end server-level tests)

### 4. State machine tests ❌
- PENDING → CONFIRMED → DELETED state transitions
- Illegal state transitions (DELETED → CONFIRMED) should be rejected
- PENDING files should not be visible to search/list

### 5. Missing module tests
- server handleCopy, handleRename, handleMkdir error paths
- parser/treebuilder interface mock tests
- CLI serve command

## Recommendation

Prioritize concurrency safety and error path tests — these are the most bug-prone areas.

PR #506, opened 2026-06-06
e2e: add FUSE SQLite churn and concurrency detectors

qiffang:
## Summary
- extend `e2e/fuse-sqlite-correctness.sh` with two opt-in workloads on top of PR #505:
  - `RUN_FUSE_SQLITE_CHURN_GC=1`: repeated large rollback-journal DB rewrites to churn S3-backed SQLite content and verify live/remount/remote snapshot integrity
  - `RUN_FUSE_SQLITE_CONCURRENCY=1`: bounded WAL writer plus concurrent read-only readers to detect lock/WAL correctness regressions
- keep both new workloads opt-in; default release gate remains the PR #505 rollback-journal SQLite path
- document the new knobs in `e2e/README.md` and `e2e/AGENTS.md`

## Scope notes
- This PR is stacked on PR #505 (`dat9-dev2/fuse-sqlite-correctness`).
- The churn workload validates that live SQLite data remains valid across repeated large rewrites, remount, and remote snapshot. In plain `drive9-server-local`, physical object-GC worker is not active unless meta+tenant pool wiring is enabled, so this does not claim a full object sweep assertion.
- Concurrent readers/writer is opt-in because it is a higher-risk lock/WAL detector and should not perturb the now-green default release gate.

## Validation
- `bash -n e2e/fuse-sqlite-correctness.sh e2e/fuse-release-gate.sh`
- extracted and compiled all embedded Python heredocs
- local temp-dir SQLite extraction with `RUN_FUSE_SQLITE_WAL=1`, `RUN_FUSE_SQLITE_CHURN_GC=1`, and `RUN_FUSE_SQLITE_CONCURRENCY=1` equivalents
- `git diff --check`
- `PATH=/usr/local/go/bin:$PATH make build-cli CLI_BIN=/tmp/drive9-sqlite-concurrency-cli`

Live FUSE validation requires the CI/local-e2e environment with `drive9-server-local` and FUSE support.


qiffang:
REVISE for `9cb1deb`.

Findings:

1. No live FUSE validation was run for the two new opt-in detectors. This PR is an E2E FUSE workload PR, but the only validation listed is shell/static/local temp-dir SQLite extraction; GitHub has no workflow runs for head `9cb1deb` because the PR is stacked. A temp directory does not exercise Drive9 FUSE locking, WAL files, remount behavior, remote snapshot copy, or the churn path through the Drive9 server. Before this can be green, run the real `local-e2e`/release gate on this head with `RUN_FUSE_SQLITE_CHURN_GC=1` and `RUN_FUSE_SQLITE_CONCURRENCY=1` and attach the run/log evidence.

2. `RUN_FUSE_SQLITE_CHURN_GC` does not assert any GC behavior. The new churn path only rewrites a large SQLite DB and verifies mounted/remounted/remote snapshot integrity; it does not check obsolete refs, GC candidates, object deletion, or any cleanup signal. That is a scope mismatch for task #28 and for a flag/docs name containing `GC`. Either add an observable backend/meta-backed GC assertion, or rename/scope this as a churn-only detector and leave task #28 open for real GC coverage.

3. The new 0/1 flag validation accepts invalid multi-character values for the new flags. `case "$RUN_FUSE_SQLITE_WAL:$RUN_FUSE_SQLITE_CHURN_GC:$RUN_FUSE_SQLITE_CONCURRENCY" in *[!01:]* )` only validates characters, so `RUN_FUSE_SQLITE_CHURN_GC=10` and `RUN_FUSE_SQLITE_CONCURRENCY=10` pass shell validation and are silently treated as disabled in Python because they are not equal to `"1"`. Add exact `0|1` validation for each new boolean flag, like the existing WAL check.

Validation I ran locally: `git diff --check origin/dat9-dev2/fuse-sqlite-correctness...HEAD`, `bash -n e2e/fuse-sqlite-correctness.sh e2e/fuse-release-gate.sh`, embedded Python heredoc compile, and `make build-cli`. These pass, but they do not close the issues above.

GitHub rejected a formal REQUEST_CHANGES review from this app identity as PR-author identity, so posting as a top-level REVISE comment.

mornyx (changes requested):
REVISE for e9981286859d9f57f4a16e0972e41294756c2522.

Findings:

1. [P1] `uploadFromShadowRemoteWithRevision` treats a post-complete `StatCtx` as the committed revision. This is not safe enough to feed into `markHandleRemoteCommittedLocked` / `recordCommittedRevision`. After `WriteMultipartStreamConditional` has already completed the upload, a transient HEAD/stat failure now makes callers treat the already-committed upload as failed, which can surface EIO on strict flush/fsync or make the commit queue retry/mark a conflict for data that is already remote-durable. More importantly, the `StatCtx` result is not an atomic result for this upload: another actor can write the same path between multipart complete and the follow-up stat, and this code would adopt that later revision as if it belonged to the local bytes. Future local writes would then CAS against the other actor revision without having read/merged that content, defeating conflict detection. Please either have the multipart complete endpoint return the exact committed revision and propagate that, or keep this path at `committedRev=0` on success and use the existing cache invalidation / expected-revision fallback behavior. See `pkg/fuse/remote_upload.go:41-48`.

2. [P1] The same-mount committed revision cache is path-level, monotonic, and never invalidated/replaced on path lifecycle changes. `recordCommittedRevision` only keeps the max revision for a path, while `adoptCommittedRevisionLocked` applies it to any later non-new handle for that path. If a file is unlinked/renamed away and then the same path is created again, the server-side file revision can legitimately reset to 1, but the mount still remembers the old higher revision. The first create can succeed with expected revision 0, but `recordCommittedRevision` will not lower the cached value; subsequent writes/open handles can adopt the stale higher revision and issue impossible CAS writes, causing persistent 409s for normal path reuse. This is especially relevant for SQLite sidecars and other churn workloads. Please clear/move this cache on unlink/rmdir/rename and make new-file commits replace the path epoch rather than max with the previous file, or key the cache by stable resource identity instead of path alone. See `pkg/fuse/dat9fs.go:737-769` and the write paths that call `expectedRevisionForHandleLocked`.

I did not run tests per request; this review is code inspection only.

qiffang:
Follow-up fixes pushed in `2d342ad` for the remaining SQLite WAL concurrency short-read path.

What changed:
- Prevents SQLite `*-wal` / `*-journal` empty shadow-spill create handles from returning local EOF before checking same-mount visible sidecar bytes.
- Keeps explicit truncate semantics (`ZeroBase` / `O_TRUNC`) unchanged.
- Adds regression `TestReadSQLitePersistentJournalEmptyShadowCreateUsesSiblingDirtyRange`.

Validation:
- `go test ./pkg/fuse -run 'TestReadSQLitePersistentJournalEmptyShadowCreateUsesSiblingDirtyRange|TestFlushSQLitePersistentJournalSkipsStaleEmptyCreateAfterSiblingCommit|TestReadSQLitePersistentJournal|TestFtruncateSQLiteWALPreservesOpenReaderSnapshot|TestFlushSQLitePersistentJournalUploadAllSeedsCommittedCache' -count=1`
- `go test ./pkg/fuse ./cmd/drive9 -count=1`
- `bash -n e2e/fuse-sqlite-correctness.sh e2e/fuse-release-gate.sh`
- `git diff --check`
- `make build-cli CLI_BIN=/tmp/drive9-pr506-cli`
- Live parameterized `local-e2e` run with `RUN_FUSE_SQLITE_CHURN=1` and `RUN_FUSE_SQLITE_CONCURRENCY=1`: https://github.com/mem9-ai/drive9/actions/runs/27073315478

Live gate log confirms rollback/churn/concurrency all ran and passed (`RESULT: 20/20 passed, 0 failed`), with no `SQLITE_IOERR`, reader `OperationalError`, `db-wal` revision conflict, malformed DB, dirty-wait expiry, `workload.db-wal got=0`, or timeout/hang markers.

mornyx (approved):
Reviewed latest head 2d342ade code-only; I did not run tests locally.

The two blockers from my previous review look addressed:
- Shadow/multipart upload completion no longer does a post-complete Stat to infer the committed revision, so a successful upload is not turned into a false failure or attributed to another actor's newer revision.
- The path-level committed revision cache now has replace/forget semantics for new commits and namespace lifecycle operations such as unlink, rmdir, and rename, so delete/recreate flows no longer keep a stale high-water revision for the path.

I also re-read the SQLite sidecar visibility changes, ftruncate/unlink reader snapshots, same-path remote commit serialization, and the e2e/workflow updates. I do not see a remaining blocking correctness issue. One non-blocking tradeoff to keep in mind is the broad suffix-based SQLite sidecar matching (, , ), which is useful for extensionless SQLite DB names but also means non-SQLite files with those suffixes inherit the special FUSE behavior.

Approved.

mornyx (approved):
Reviewed latest head 2d342ade code-only; I did not run tests locally.

The two blockers from my previous review look addressed:
- Shadow/multipart upload completion no longer does a post-complete Stat to infer the committed revision, so a successful upload is not turned into a false failure or attributed to another actor newer revision.
- The path-level committed revision cache now has replace/forget semantics for new commits and namespace lifecycle operations such as unlink, rmdir, and rename, so delete/recreate flows no longer keep a stale high-water revision for the path.

I also re-read the SQLite sidecar visibility changes, ftruncate/unlink reader snapshots, same-path remote commit serialization, and the e2e/workflow updates. I do not see a remaining blocking correctness issue. One non-blocking tradeoff to keep in mind is the broad suffix-based SQLite sidecar matching (*-shm, *-wal, *-journal), which is useful for extensionless SQLite DB names but also means non-SQLite files with those suffixes inherit the special FUSE behavior.

Approved.

PR #507, opened 2026-06-07
feat: add layered filesystem overlays

mornyx:
## Summary

This PR introduces Drive9 layered filesystem overlays, following `docs/design/layered-filesystem-v1-design.md`.

The layer model keeps the base namespace unchanged until an explicit commit. Writable overlay state is recorded in tenant-local layer tables and can be restored across sandboxes without losing durable layer data.

This update also completes the remaining layer blockers:

- Added S3-backed layer object storage for large files, including streaming upload, streaming read, streaming FUSE restore, commit into the base filesystem, and storage encryption metadata propagation.
- Extended `--layer` write support across CLI workflows: `cp`, `rm`, `mkdir`, `mv`, `chmod`, `symlink`, `hardlink`, plus layer-aware `find` and `grep`.
- Routed FUSE commit queue entries to layer object uploads when inline payloads are too large or shadow-spilled.
- Added layer event polling so active layer mounts can refresh changes written by other sandboxes.
- Made commit retry/recovery idempotent for layers left in `committing`.
- Added directory rename commit support.
- Made directory whiteout safe by allowing only empty directory whiteouts.
- Implemented `hardlink --layer` as copy-up, producing an independent layer file.
- Fixed layer search prefix semantics so `--layer` find/grep for `/repo` includes `/repo/...`.
- Added a complete LayerFS capability matrix in `docs/design/layered-filesystem-feature-matrix.md` to track native parity, AgentFS-class gaps, production readiness, and required follow-up tests.

## LayerFS Capability Matrix

Full matrix and roadmap: `docs/design/layered-filesystem-feature-matrix.md`.

| Area | Current status | Main remaining gaps |
| --- | --- | --- |
| Native Drive9 behavior | Parity | LayerFS remains opt-in; continue guarding non-layer performance and behavior regressions. |
| Layer lifecycle | Supported | Delete/archive, export/import, rename/tag update, patch-style diff. |
| Name/tag layer refs | Supported | Uniqueness policy and clearer ambiguous-ref UX. |
| FUSE `--layer` mount | Partial | WebDAV layer mode, dirty-handle checkpoint barrier, stronger multi-client consistency. |
| CLI layer parity | Partial | `cat/ls/stat --layer`, recursive copy/delete, append/resume, metadata flags, pack/unpack/git workflows. |
| Large files | Partial | Multipart/resume/range/direct upload, layer quota, retention, object GC. |
| Search/indexing | Partial | Layer tag search, semantic search before commit, checkpoint search, extraction pipeline integration. |
| Commit/recovery | Partial | Exactly-once apply ledger, failpoint crash coverage, structured conflict report/rebase. |
| Metadata/POSIX parity | Partial | Tags, description, timestamps, xattr, hardlink identity, sparse files, locks. |
| Security/auth | Partial | Per-layer ACL, scoped sandbox token, path-scope fuzzing, complete audit provenance. |
| Operations | Partial/Gap | Metrics, tracing, dashboards, migration rollout, load tests, retention/quota runbooks. |
| AgentFS-class features | Partial/Gap | Portable session bundle, provenance audit, progressive permissions, branch/fork/squash. |
| E2E coverage | Partial | Multi-sandbox restore, crash recovery, recursive workflows, metadata/search parity, performance. |

## Implementation Notes

- Added tenant schema tables for `fs_layers`, `fs_layer_tags`, `fs_layer_entries`, `fs_layer_events`, and `fs_layer_checkpoints`.
- Added stable layer references by layer id, human-readable name, or tag reference (`tag:key=value` and `tag:key`) anywhere a layer must be selected.
- Added datastore, HTTP server, Go client, and CLI support for layer lifecycle operations under `/v1/layers`.
- Added `/v1/layers/{layer}/objects` for raw layer file payload upload/read, avoiding inline JSON/blob limits for large files.
- Added `/v1/layers/{layer}/events` for cross-sandbox refresh.
- Added `/v1/layers/{layer}/checkpoints` for creating restore checkpoints.
- Added layer-aware FUSE mount behavior for `--layer` and `--checkpoint`.
- Kept checkpoint restore by durable sequence, so sandbox restore can recover durable layer data without mutating base.
- Preserved local overlay and write-back performance paths; bytes stay local first and are uploaded to the layer backend only when the commit queue flushes or restore requires backend reads.

## Test Coverage

New and updated tests cover:

- Layer API flow, name/tag references, diff, checkpoint lookup, rollback, commit, and base visibility.
- Large S3-backed layer object upload/read/grep/commit.
- Layer commit retry from `committing`.
- Directory rename commit.
- Empty vs non-empty directory whiteout behavior.
- Layer search overlay merge and prefix matching.
- CLI `--layer` upload/search behavior.
- FUSE layer restore from checkpoint and object-backed entries.
- FUSE commit queue object upload for shadow-spilled/large layer entries.
- E2E layer smoke coverage for explicit `--layer` CLI writes, large file upload, layer search, commit, and hash verification.

Validated locally with:

```bash
env GOCACHE=/private/tmp/drive9-layer-blockers/.codex-gocache make lint
env GOCACHE=/private/tmp/drive9-layer-blockers/.codex-gocache go test -count=1 ./pkg/server ./pkg/client ./cmd/drive9/cli ./pkg/fuse
bash -n e2e/layer-fs-smoke-test.sh
git diff --check
```


qiffang (changes requested):
Verdict: REVISE. I found multiple correctness blockers in the layer contract, independent of CI.

1. Checkpoint restore cannot restore the state that existed at the checkpoint for paths updated later. `fs_layer_entries` is keyed only by `(layer_id, path_hash)` and `UpsertFSLayerEntry` overwrites both content and `entry_seq` on duplicate paths (`pkg/tenant/schema/fs_layer.go:57`, `pkg/datastore/fs_layer.go:353-378`). `CreateFSLayerCheckpoint` records only `MAX(entry_seq)` (`pkg/datastore/fs_layer.go:464-470`), and restore skips entries with `entry_seq > durable_seq` (`pkg/fuse/mount.go:596-599`). Repro: upsert `/repo/a.txt` as `v1`, checkpoint, upsert `/repo/a.txt` as `v2`, then mount at the checkpoint. The single row now has seq/content for `v2`, so the checkpoint either omits `a.txt` or restores the wrong content. This violates the PR/design claim that checkpoint restore recovers durable layer data by sequence. This needs historical versioning/event replay or another fail-closed checkpoint model, plus a regression for same-path update after checkpoint.

2. Layer entries are not constrained to the layer base root. The create request stores `BaseRootPath`, but entry upsert accepts `req.Path` directly (`pkg/server/fs_layer.go:303-332`) and commit applies every stored entry path to the base namespace (`pkg/server/fs_layer.go:383-389`, `pkg/server/fs_layer.go:501-558`) without checking it is under `layer.BaseRootPath`. A client can create a layer for `/repo`, upsert `/unrelated/owned.txt`, and commit it, mutating base data outside the layer root. That breaks the core isolation model: a layer created from one Drive9 path must not be able to write arbitrary tenant paths. Add server-side validation on entry upsert and commit/restore paths, and tests for outside-root upsert/commit rejection.

3. `commit` is not all-or-nothing. `handleFSLayerCommit` preflights once, then applies entries sequentially to the base (`pkg/server/fs_layer.go:371-389`). If any later entry fails, earlier writes/deletes/mkdirs have already mutated base and the layer is just marked `conflicted`. That directly contradicts the design text and PR summary: “commit is all-or-nothing” and “system cannot half-submit.” A concrete failure is first entry write succeeds, second `chmod`/`symlink`/write fails due to remote error or a concurrent base change after preflight; base is left partially updated. The fix must make commit atomic from the user-visible contract or fail before applying anything that can leave partial base state, with regression coverage proving base is unchanged when an apply step fails.

Validation I ran: `git diff --check origin/main...HEAD` passed; `go test ./pkg/fuse -run 'Test.*Layer|TestCommitQueue' -count=1` passed; `go test ./cmd/drive9/cli -run 'Test.*Layer' -count=1` passed. Datastore/server/schema package tests are locally blocked by rootless Docker/testcontainers on this machine, so CI covers those runtime paths.

qiffang (changes requested):
R2 review on head `2eef8001053cf46f8e274dff5e084b32f9d0b041`: still **REVISE**.

Blocking findings:

1. `pkg/fuse/dat9fs.go:803` / `pkg/fuse/dat9fs.go:4560`: `Rmdir` can still treat a layer directory as empty while an unflushed child exists only as a live open handle. `Create` registers the new handle immediately (`allocateFileHandle`) before the child reaches `pendingIndex`/`writeBack`/`layerDirs`/`layerSymlinks`, but `layerDirHasOverlayChildren()` only checks those indexes and not `openHandles`. A sequence like `mkdir dir; open dir/file for write; rmdir dir` can pass the emptiness check, write a layer whiteout for `dir`, remove namespace/cache state, and leave the open child to flush later under a whiteouted parent. That violates normal directory non-empty semantics and risks orphaning or resurrecting layer writes. Add an open-handle-prefix check, with a regression that keeps `Rmdir` returning `ENOTEMPTY` while a writable child handle under the directory is still open and not yet flushed.

2. `pkg/fuse/dat9fs.go:500` / `pkg/fuse/dat9fs.go:515`: chmod on a layer symlink strips the symlink filetype bit from in-memory layer state. `SetAttr` passes permission bits into `applyRemoteMode()`, and `upsertLayerChmod()` then calls `markLayerSymlink(localPath, target, mode)` with only `0oNNN`. Later `Lookup`/`Readdir` rebuild the inode from `layerSymlinks` using that stored mode; `entryIsSymlink()` becomes false, directory entries are synthesized as regular files, and `Readlink` can return `EINVAL` after the inode is forgotten or re-listed. Preserve `S_IFLNK` when updating symlink layer state, and add a regression that chmods a layer symlink, forgets/relooks it up or lists it, then verifies it remains a symlink and `Readlink` still works.

Additional still-valid issue:

3. `pkg/client/fs_layer.go:357`: `CommitFSLayer` still discards the `io.ReadAll(resp.Body)` error on conflict responses. If the server returns a conflicted status but the body read fails, this path silently falls through into generic `readError` with a consumed/empty body and masks the actual transport/body failure. Capture and return the read error explicitly before unmarshalling; add a small client regression with a failing response body.

Validation run locally:

- `/usr/local/go/bin/go test ./pkg/fuse -run 'TestLayer|TestOpenHandle|TestRmdir' -count=1` passed.
- `pkg/client`, `pkg/datastore`, and `pkg/server` focused tests are blocked on this machine by `panic: rootless Docker not found` from testcontainers, same local limitation as prior reviews.

The earlier history/checkpoint and server rollback direction looks improved, but the FUSE layer namespace bugs above are enough to block merge.

qiffang (changes requested):
R3 review on current head `2eef8001053cf46f8e274dff5e084b32f9d0b041`: still **REVISE**.

I rechecked the current GitHub head; it is the same head as the prior R2 review. The blocking issues below are still present in the code.

Blocking findings:

1. `pkg/fuse/dat9fs.go:803` / `pkg/fuse/dat9fs.go:4560`: `Rmdir` can still remove a layer directory while an unflushed child only exists as a live open handle. `layerDirHasOverlayChildren()` checks `pendingIndex`, `writeBack`, `layerDirs`, and `layerSymlinks`, but it still does not scan live `openHandles`. A sequence like `mkdir dir; create/open dir/file; rmdir dir` can pass the empty-dir check before the child reaches the layer indexes, write a directory whiteout, and leave the child handle to flush later under a removed parent. This violates POSIX non-empty directory semantics and can orphan/resurrect layer writes.

2. `pkg/fuse/dat9fs.go:500` / `pkg/fuse/dat9fs.go:514`: chmod on a layer symlink still strips the symlink filetype bit. `upsertLayerChmod()` stores only permission bits via `markLayerSymlink(localPath, target, mode)`. Later lookup/readdir rebuilds the entry from `layerSymlinks` and `entryIsSymlink()` can become false, so a chmodded layer symlink can be synthesized as a regular file and `Readlink` can fail. Preserve the existing symlink mode/type bits when updating permissions and add a relist/relookup/readlink regression.

3. `pkg/client/fs_layer.go:357`: `CommitFSLayer` still discards `io.ReadAll(resp.Body)` errors on conflict responses. If reading the conflict body fails, this path silently tries to unmarshal a failed/partial read and then falls through to generic `readError` with a consumed body, masking the real transport/body error. Capture the read error explicitly and add a client regression using a failing response body.

Process blocker:

- GitHub currently reports this PR as `mergeable=false`. Even after the functional issues above are fixed, the branch needs to be made mergeable against current `main` and revalidated.

Validation run locally:

- `git diff --check 68ce029f889a1a6ac17b07fb9d6b5849ce39631b...HEAD` passed.
- `/usr/local/go/bin/go test ./pkg/fuse -run 'TestLayer|TestOpenHandle|TestRmdir' -count=1` passed.
- `/usr/local/go/bin/go test ./cmd/drive9/cli -run TestLayer -count=1` passed.
- `go test ./pkg/client -run TestFSLayer -count=1`, `./pkg/datastore -run TestFSLayer`, and `./pkg/server -run TestFSLayer` are locally blocked by `panic: rootless Docker not found` from testcontainers on this machine.

qiffang (changes requested):
## R1 on `47534ef9` — **REVISE**

CI: GREEN. I verified @qiffang's R1–R3 blockers are addressed: checkpoint uses event-log INSERT (not ON DUPLICATE KEY UPDATE), base-root validation exists on both entry upsert and commit, commit uses `BeginFSLayerCommit` CAS for serialization, `Rmdir` checks `layerDirHasOpenChild`, `upsertLayerChmod` preserves `S_IFLNK`, and `CommitFSLayer` client handles `readErr` explicitly. The design research and spec docs are thorough.

New blockers found:

---

**B1: `rollbackFSLayerCommit` uses `r.Context()` — client disconnect aborts rollback, leaving base half-mutated.**

`pkg/server/fs_layer.go` lines 8530 and 8541 (diff): both `rollbackFSLayerCommit(r.Context(), b, snapshots)` calls use the request context. If the HTTP client disconnects during commit (timeout, network error, user Ctrl-C), `r.Context()` is canceled and `rollbackFSLayerCommit` returns early. The earlier `applyFSLayerEntry` calls that already mutated base paths are NOT reverted. The layer is set to `conflicted` but base is permanently partially updated — the exact violation the snapshot/rollback mechanism was designed to prevent.

Fix: use `context.WithoutCancel(r.Context())` with an explicit deadline for rollback (e.g., 60 seconds). Yes, `WithoutCancel` strips deadlines (lesson from PR #493), so apply `context.WithTimeout(context.WithoutCancel(r.Context()), rollbackTimeout)` explicitly.

---

**B2: `restoreLayerEntries` drops `chmod` on base files that have no pending entry.**

`pkg/fuse/mount.go` restore handler for `"chmod"` calls `pending.UpdateMode(localPath, entry.Mode)` (line 7055–7057), but `UpdateMode` is a no-op if the path has no existing pending entry. For a base file that was only `chmod`'d through the layer (no content change), the pending index has no entry to update. On checkpoint restore, the mode change is silently lost and the file appears with its original base mode.

For directories and symlinks the same code path correctly calls `markLayerDir`/`markLayerSymlink` in the `fs` namespace maps, which are consulted by `GetAttr`/`Lookup`. But files rely on the pending index mode, which is absent.

Fix: when `UpdateMode` finds no existing pending entry for a chmod-only file, either (a) fetch the file's content and create a full pending+shadow entry with the new mode, or (b) add a mode-only pending entry type. Add a regression test: chmod a base file through a layer, checkpoint, remount at checkpoint, verify the file has the new mode.

---

**B3: `handleFSLayerEntryUpsert` allows upserts to non-`active` layers via TOCTOU.**

`pkg/server/fs_layer.go` line 8393-8395: `handleFSLayerEntryUpsert` reads `layer.State` from the `ResolveFSLayerRef` result (resolved at the top of `handleFSLayerObject`). Between the state check and the actual `store.UpsertFSLayerEntry` call, a concurrent commit can transition the layer to `committing`/`committed`/`conflicted`. The entry upsert then succeeds against a non-active layer. The datastore's `UpsertFSLayerEntry` takes `FOR UPDATE` on `fs_layers` but doesn't re-check state.

Fix: either (a) add `AND state = 'active'` to the `FOR UPDATE` query in `UpsertFSLayerEntry` and return `ErrFSLayerStateConflict` if zero rows matched, or (b) re-check state inside the transaction. Add a regression: start commit, race an entry upsert, verify the upsert is rejected.

---

**B4: `listFSLayerEntriesLatest` subquery JOIN misses `layer_id` in the outer WHERE clause.**

`pkg/datastore/fs_layer.go` lines 4346-4358 (diff): the `listFSLayerEntriesLatest` subquery groups by `(path_hash, path)` and joins back by `(path_hash, path, entry_seq)`. The subquery includes `WHERE layer_id = ?` but the outer query's JOIN does NOT include `latest.layer_id = e.layer_id` — it only adds `WHERE e.layer_id = ?` as a separate filter. In a single-tenant context this works, but the `fs_layer_entries` table schema uses `(layer_id, path_hash, entry_seq)` as the composite key. If two layers have the same `(path_hash, path, entry_seq)` tuple (e.g., both have seq 1 for `/repo/a.txt`), the JOIN would match cross-layer. The outer `WHERE e.layer_id = ?` filters correctly, but this is fragile and non-obvious.

This is borderline — unlikely in practice given that `path_hash` is a SHA256 and `entry_seq` is per-layer. However, the correct SQL for the subquery should include `layer_id` in the JOIN condition to match the table's composite key semantics. Non-blocking observation — include `AND latest.layer_id = e.layer_id` in the JOIN for defensive correctness.

---

Observations (non-blocking):

- **O1**: `agfs` import in `pkg/server/fs_layer.go` line 8104 (`"github.com/c4pt0r/agfs/agfs-server/pkg/filesystem"`) — this is an external dependency on AgentFS. Confirm this is intentional and not a copy-paste artifact from research.

- **O2**: Layer entry content is stored inline in `content_blob` (up to 128 MiB per entry body limit). For large files, this means the full file content is stored in the database. The codex bot flagged this as P1 — worth tracking as a follow-up for blob/external storage, but not a blocker for V1.

- **O3**: `restoreLayerRenameEntry` falls back to reading the old source path from the base (`c.ReadCtx(ctx, fullEntry.Path)` at line 7161). After a rename, the old path should have a whiteout in the layer. If the base was also modified concurrently, this reads stale/wrong data. Edge case, but worth a TODO.

Validation:
- `git diff --check origin/main...47534ef9` — no whitespace errors.
- Full diff read covering all 46 files.
- Cross-referenced against @qiffang's R1/R2/R3 findings — all addressed.

Verdict: **REVISE** — B1 (rollback context cancel) and B2 (chmod restore for base files) are correctness blockers. B3 (upsert TOCTOU) is a concurrency safety blocker.

qiffang (changes requested):
## dat9-dev1 R1 review — PR #507 (Layer FS V1)

**Verdict: REVISE** — This is a major feature PR (9998 lines). I've reviewed the full diff thoroughly. I agree with @adversary-2's blocking findings and add architectural observations.

---

### Blocking findings I agree with

**@adversary-2 B1 (checkpoint restore cannot recover pre-update state)**: Confirmed. `fs_layer_entries` is keyed by `(layer_id, path_hash)` with UPSERT semantics (`ON DUPLICATE KEY UPDATE`). When a path is updated after checkpoint, the row is overwritten with the new `entry_seq` and content. Checkpoint records only `MAX(entry_seq)`, but the row for the earlier version is gone. This is the most critical design flaw — the checkpoint model is fundamentally broken for same-path updates.

**@adversary-2 B2 (entries not constrained to base root)**: Confirmed. `handleFSLayerEntryUpsert` accepts any path without validating it's under `layer.BaseRootPath`. The e2e test does exercise this (`put_layer_entry_expect_code "entry outside base root is rejected" "400"`), but I don't see server-side validation code enforcing it. The test may be passing due to a different reason or I may have missed a check — needs verification.

**@adversary-2 B3 (commit is not all-or-nothing)**: Confirmed. `handleFSLayerCommit` applies entries sequentially to base. If entry N succeeds and entry N+1 fails (concurrent base change, DB error), the base is left partially updated. This directly contradicts the design doc's "all-or-nothing" guarantee.

**@adversary-2 R2-B1 (Rmdir + open handle race)**: Agreed. `layerDirHasOverlayChildren()` doesn't check live open handles.

**@adversary-2 R2-B2 (chmod strips symlink type bit)**: Agreed. Storing only permission bits via `markLayerSymlink` loses `S_IFLNK`.

---

### My additional findings

### B6 (DESIGN): Layer create `--durability` modes have no visible behavioral difference

The design doc specifies three modes: `restore-safe`, `write-through`, `local-fast`. However, the FUSE layer mount code (`pkg/fuse/dat9fs.go`, `pkg/fuse/mount.go`) does not branch on `DurabilityMode` for write behavior. All three modes appear to follow the same write path (local shadow → writeback → commit queue → layer entry upsert). The mode is stored in the DB but not enforced. If it's intentionally deferred, document it explicitly in the PR and e2e tests.

### B7 (BUG): `fs_layer_entries.entry_seq` is assigned client-side via MAX+1

In `UpsertFSLayerEntry`, the seq is assigned by:
```sql
SELECT COALESCE(MAX(entry_seq), 0) + 1 FROM fs_layer_entries WHERE layer_id = ?
```

This is vulnerable to race conditions if two concurrent upserts for the same layer execute simultaneously — both read the same MAX, both assign the same seq. This breaks checkpoint semantics (which depend on seq ordering). The seq should be assigned server-side via AUTO_INCREMENT or a separate sequence table.

### B8 (CONCERN): Research doc + Design doc are 1100+ lines of markdown in the PR

`docs/design/layered-filesystem-research.md` (559 lines) and `docs/design/layered-filesystem-v1-design.md` (562 lines) are thorough and well-written research/design documents. However, they include specific implementation claims that are contradicted by the actual code (e.g., "commit is all-or-nothing"). The docs should be updated to reflect actual V1 limitations, or the code should be fixed to match.

### B9 (TESTING): e2e test `layer-fs-smoke-test.sh` is 813 lines — excellent coverage

The e2e test is impressively thorough: create by name/tag, diff/checkpoint lookup, rollback, commit, scope rejection, conflict detection, mkdir/upsert/whiteout/rename/symlink/chmod entries, checkpoint/full restore. This is the right approach. However, it doesn't exercise B1 (same-path update after checkpoint then restore) because the test's checkpoint only covers paths that are written once.

### B10 (GOOD): `local-smoke.sh` + `make e2e-local` + `EMBEDDING_MODE=none` is excellent

The addition of `make e2e-local` with disposable MySQL and `DRIVE9_LOCAL_EMBEDDING_MODE=none` significantly lowers the bar for local validation. This is a real developer experience improvement.

---

### Good things in this PR

- **Design quality**: The research doc is one of the best competitive analyses I've seen — thorough comparison of AgentFS, ArtifactFS, Tilde, E2B, Modal, Daytona, YoloFS, DeltaFS
- **CLI layer commands**: Well-structured `drive9 fs layer <create|list|status|diff|checkpoint|rollback|commit>` with consistent flag handling
- **Layer ref resolution**: id → name → tag fallback with conflict detection for ambiguous refs
- **Client API**: Clean `FSLayer*` types in `pkg/client/fs_layer.go`
- **Mount integration**: `--layer` and `--checkpoint` flags in mount with WebDAV guard and checkpoint-requires-layer validation
- **e2e coverage**: 813-line bash e2e test covering full lifecycle
- **Local dev**: `make e2e-local` with `embedding_mode=none` MySQL setup
- **CI integration**: `layer-fs-smoke-test.sh` added to `local-e2e.yml` as required check

### Summary

The layer FS is a major capability addition and the design direction is correct. The three critical issues are: (1) checkpoint model can't recover same-path updates, (2) commit is not atomic, (3) entries not validated against base root. These are fixable but require careful design work, especially B1 which may need a schema change (historical entry versioning or WAL-style append-only entries).

I recommend prioritizing B1 (checkpoint) and B3 (atomic commit) as they are core to the layer safety guarantee.

mornyx:
@qiffang I went through the latest review comments and pushed the fixes.

What changed:

- Rollback during layer commit now uses a detached timeout context, so client cancellation cannot interrupt rollback cleanup.
- Chmod-only layer restore now preserves file mode via a FUSE-side file mode overlay instead of dropping the entry.
- Layer entry upsert now locks and re-checks the layer state in the datastore, rejecting writes once commit/rollback has started.
- Latest-layer-entry listing now joins on `layer_id` as well as path/hash/seq.
- Server-side layer rename is constrained to file rename only, rejects directory rename, preflights target conflicts, and commits with no-replace semantics.
- FUSE `rmdir` in layer mode now rejects non-empty base/layer directories instead of creating a recursive whiteout.
- FUSE symlink creation now respects local-only and git workspace overlays before writing layer entries.
- The layer FS e2e script now builds the CLI from the repository root, independent of the caller cwd.
- Docs/CLI text now accurately describes V1 durability behavior and the current inline `content_blob` boundary.
- Touched test assertions were adjusted per the review style nit where safe.

Validation:

- `git diff --check`
- `go test ./pkg/fuse`
- `go test ./cmd/drive9/cli -run 'Test.*Mount|Test.*Layer'`
- `make test TEST_PKGS='./pkg/datastore' TEST_RUN='TestFSLayer'`
- `make test TEST_PKGS='./pkg/server' TEST_RUN='TestFSLayer'`
- `make lint`
- GitHub checks after push: `code-ci / ci` passed, `local-e2e / drive9-server-local e2e smoke` passed, CodeRabbit passed.

I also merged the latest `main`, resolved the `e2e/README.md` conflict by keeping both the layer FS and new Git ops smoke coverage docs, pushed the merge, and converted the PR back to draft as requested.

qiffang (commented):
## R5 Re-review — PR #507 (Layered Filesystem Overlays)

**Verdict: GREEN** ✅ — All R4 blockers resolved. CI green.

### R4 Blocker Resolution

**B1 (rollback uses r.Context())**: FIXED. `handleFSLayerCommit` now creates `rollbackCtx` via `context.WithTimeout(context.WithoutCancel(r.Context()), fsLayerRollbackTimeout)` at line 8893. All rollback and state-update operations use `rollbackCtx`, not `r.Context()`. Client disconnect no longer cancels rollback.

**B2 (chmod restore for base files)**: FIXED. `restoreLayerEntries` chmod case now dispatches by `entry.Kind` (file → `markLayerFileMode`, dir → `markLayerDir`, symlink → `markLayerSymlink`), storing mode overlays in layer namespace maps. Regression test verifies chmod-only restore creates mode overlay without pending data.

**B3 (state check TOCTOU)**: FIXED. `UpsertFSLayerEntry` now does `SELECT ... FOR UPDATE` inside a transaction, re-reads state from DB, and rejects non-active layers with `ErrFSLayerStateConflict`. HTTP handler check is just a fast-path optimization. Regression test `TestFSLayerUpsertRejectsNonActiveLayer` verifies.

**O1**: `RollbackFSLayer` (line 4253) reads state via `GetFSLayer` then calls `SetFSLayerState` — two separate operations without transaction. Minor TOCTOU, but rollback to `abandoned` is terminal and race is benign.

**O2**: `restoreLayerEntries` error at mount time is logged to stderr but does not fail the mount (line 7227). This means a corrupt layer state silently proceeds with partial restore. Acceptable for V1 UX but worth documenting.

No new blockers.

qiffang (commented):
## R2 Review — PR #507: Layered Filesystem V1

**Verdict: GREEN** ✅ (conditional on CI green)

### R1 Blocker Status

All 3 blockers from R1 (raised by @adversary-2 R4) are fixed:

**B1 (rollback context)**: Fixed ✅ — `rollbackFSLayerCommit` now uses `context.WithTimeout(context.WithoutCancel(r.Context()), fsLayerRollbackTimeout)`. Client disconnect no longer cancels rollback.

**B2 (chmod restore lost mode)**: Fixed ✅ — Restore now records chmod modes in the in-memory `layerFiles`/`layerDirs`/`layerSymlinks` maps via `markLayerFileMode`/`markLayerDir`/`markLayerSymlink`. These are read by `applyLayerFileMode`/`layerDirMode`/`layerSymlink` during Lookup/GetAttr, so mode changes survive restore even without a shadow/pending entry.

**B3 (state check TOCTOU)**: Already safe ✅ — `UpsertFSLayerEntry` in the datastore does `SELECT ... FOR UPDATE` on the layer row and checks `state = 'active'` within the transaction (lines 4284-4291). Even if the server-side handler reads a stale state during `ResolveFSLayerRef`, the DB transaction rejects concurrent-commit races.

### New observations (non-blocking)

**O1**: `fsLayerPathWithinBaseRoot` (datastore, line 4715-4723) accepts `p == baseRoot` and `strings.HasPrefix(p, baseRoot)`. For `baseRoot="/repo"` and `p="/repo-fork/file.txt"`, prefix match is `true` without checking the path separator. Same pattern in `fsLayerPathWithinBase` (server, line 9012-9020). Should use `strings.HasPrefix(p, baseRoot + "/")` or equivalent.

**O2**: `listFSLayerEntriesLatest` uses a self-join with `GROUP BY + MAX(entry_seq)` to get latest entry per path. For large layers with thousands of entries, this could become expensive. Fine for V1 but worth indexing or windowing in V2.

**O3**: `uploadLayerEntry` (commit_queue.go, line 5349-5381) reads entire shadow into memory. The `maxInlineLayerEntryBytes = 96 MB` limit caps memory use, but large layer entries could still pressure the FUSE process. External storage refs are explicitly punted (`not supported by commit yet`). Acceptable for V1.

**O4**: `layerDirHasOverlayChildren` (dat9fs.go) iterates all maps under `layerMu.RLock`. For large layers, this linear scan on every `Rmdir` could be slow. Fine for V1.

### Architecture summary

Clean layered overlay design:
- Base namespace is read-only; writes go through layer entry table
- Layer entries form an append-only event log per path (latest seq wins)
- Checkpoint = snapshot of max durable_seq at a point in time
- Restore replays entries up to checkpoint seq, populating shadow/pending/in-memory maps
- Commit applies entries to base with full preflight conflict detection + snapshot-based rollback
- FUSE integration intercepts mkdir/delete/rename/chmod/symlink, routing to layer entry upserts
- CommitQueue retains shadow+pending after upload for layer mounts (cleaned on layer commit/rollback)

No blockers. The fixes address all R1 issues correctly.

qiffang:
Did a correctness-focused review of this PR, explicitly covering both single-agent and multi-agent shared-workspace scenarios. Read in full: `pkg/datastore/fs_layer.go`, `pkg/server/fs_layer.go`, the FUSE refresh path (`pkg/fuse/mount.go`, `layer_events.go`), rename/chmod emitters (`pkg/fuse/dat9fs.go`, `cmd/drive9/cli/layer_write.go`), and `pathutil`. Findings ordered by severity.

## P0-1: latest-entry-per-path dedup silently drops rename/chmod — data loss even single-agent

Commit reads entries via `listFSLayerEntriesLatest` (`pkg/datastore/fs_layer.go:499-522`), which reduces the append-only log to `MAX(entry_seq)` per path. That reduction is only sound if every op is a **full idempotent state description** of its path. `rename` and `chmod` are **deltas**, and rename is keyed by its *source* path (FUSE dir rename `pkg/fuse/dat9fs.go:726-731`, CLI file rename `cmd/drive9/cli/layer_write.go` `renameLayerPath`). Concrete single-agent flows that break:

1. `mv /repo/a /repo/b && mkdir /repo/a` → the later `mkdir` entry at path `/repo/a/` shadows the rename entry. At commit the rename is never applied: `/repo/b` is never created, the moved content is **silently lost**.
2. CLI: `layer write f` (upsert seq=1) then `layer mv f g` (rename entry also at path `f`, seq=2) → dedup keeps only the rename. Commit renames the **old base content** to `g`; the edit is silently dropped.
3. CLI: `layer write f` + `layer chmod f` → standalone chmod entry (`chmodLayerPath`) shadows the upsert → content edit lost, only mode applied.
4. FUSE: `mkdir d && chmod 700 d` → FUSE patches the *file* case by re-upserting the full entry (`dat9fs.go:564-582`) but emits a standalone chmod for dirs (`dat9fs.go:587-592`). The chmod entry shadows the mkdir; at commit `ChmodCtx` runs against a base dir that doesn't exist → `ErrNotFound` → the **whole commit goes conflicted**.

Suggested direction: either commit replays the full ordered log (not the per-path reduction), or rename/chmod must be folded into state-style entries the way FUSE already does for chmod-on-file. Mixing a delta log with a latest-state reduction is the root cause; patching individual cases will keep leaking.

## P0-2: layer endpoints bypass per-path authorization

Every normal FS endpoint guards with `authorizeFS(w, r, FSOp..., path)` (e.g. `pkg/server/server.go:1049,1302,1478`). `pkg/server/fs_layer.go` contains **zero** authorization calls — only tenant scope. Any tenant token (including one scoped read-only or to a narrow prefix) can create a layer with `base_root_path: "/"`, insert upserts/whiteouts, and POST commit — i.e. write or recursively delete anywhere in the tenant. Layer create/entry-upsert/object-upload/commit need `authorizeFS(FSOpWrite, ...)` against the base root and entry paths; object/entry reads need `FSOpRead`.

## P1-1: commit is neither atomic nor isolated; rollback destroys concurrent writes (multi-agent)

The design doc promises atomic apply with no half-commit, but `handleFSLayerCommit` (`pkg/server/fs_layer.go:567-650`) is a sequential loop of independent backend mutations with best-effort compensation:

- **Mid-commit states are visible** to every base mount; if apply fails halfway, compensation runs — and if *that* fails halfway (line 626-628 just reports it), the base is left in a mixed state with the layer marked conflicted.
- **Rollback overwrites concurrent writes unconditionally**: `restoreFSLayerBaseSnapshot` writes with `expectedRevision=-1` (= unconditional, see `pkg/backend/dat9.go:516-519`) and `removeFSLayerBasePathIfExists` uses `RemoveAllCtx`. Example: another agent `mkdir`s a path during the commit window → apply hits EEXIST → rollback sees `snapshot.Exists=false` → **recursively deletes the other agent's directory and contents**.
- **Preflight→apply TOCTOU**: only file upserts carry a CAS at apply time (`WriteCtxIfRevision` with `BaseRevision`). whiteout/mkdir/chmod/symlink/rename have no guard. Dir whiteout preflight requires an empty dir (line 1001-1010) but apply runs `RemoveAllCtx` (line 1266-1268) — a file created concurrently in the window is recursively deleted. Dir rename apply uses replacing `RenameCtx` (line 1298-1300); no-replace is preflight-only.

If commit-to-shared-base is a v1 use case, apply needs to be transactional (or at minimum: every apply op revision/inode-CAS-guarded, and rollback must restore with revision guards instead of `-1`, failing loudly on interference instead of clobbering).

## P1-2: no fencing — concurrent commits and commit-vs-rollback races

- `BeginFSLayerCommit` (`pkg/datastore/fs_layer.go:334-358`) intentionally allows `committing→committing` for crash recovery, but there is no owner/lease/epoch. Two concurrent `POST /commit` (retry race, or two agents) both proceed through the apply loop; if one fails, its rollback unconditionally restores pre-commit snapshots **while the other is still applying**.
- `RollbackFSLayer` (`:360-373`) is Get-then-Set, and `SetFSLayerState` (`:304-332`) is an unconditional `UPDATE` with no `WHERE state = expected`. Race: rollback reads `active`; commit begins and starts applying; rollback stamps `abandoned`; commit finishes and stamps `committed`. The caller observed a successful rollback, but the base got committed.

Fix: every state transition should be a single CAS `UPDATE ... WHERE state = ?` (rows-affected checked), and commit should carry an owner token checked at finalize.

## P1-3: crash recovery is heuristic; PR matrix itself lists "exactly-once apply ledger" as a gap

`fsLayerEntryAlreadyApplied` (`:1079-1186`) infers "applied" from size/checksum/mode equality, dir existence, or rename src-missing+dst-exists. Edge cases: upsert with empty checksum passes on size match alone; `chmod` with `mode==0` is always "applied"; partial dir-rename crash states. For a shared base this should be a per-entry applied marker written transactionally with the mutation, not an inference. I'd treat that ledger as a pre-GA requirement, not a follow-up.

## P2

- **Snapshot memory**: `snapshotFSLayerBasePath` reads entire file contents into memory (`ReadCtx(path, 0, -1)`, `:882`), recursively for dir-rename trees. A commit touching large files is an OOM/latency risk, and snapshots are taken non-transactionally so "rollback" can restore a torn mixture of pre-commit states.
- **Rollback ordering**: snapshots restore in reverse discovery order (`:896-903`); for dir-rename trees children restore before their parent dir snapshot — restore under a removed parent likely fails, aborting rollback halfway.
- **Event watcher amplification**: any event (including self-generated) triggers a full `DiffFSLayer` plus per-entry content fetch for all non-locally-pending paths (`pkg/fuse/mount.go:611-`), O(layer) per event burst at a 1s tick. Restored files are then registered as `PendingOverwrite` — worth double-checking this can never re-enqueue another agent's files for upload (event ping-pong).
- **Doc/impl divergence**: the design doc states commit is an atomic transaction with no half-commit, and that V1 rejects directory rename at API/FUSE level. The implementation does sequential apply with compensation and ships dir rename. One of the two should change.
- Minor: `handleFSLayerObjectUpload` writes the S3 object before the entry-state check → orphaned objects when the layer is sealed concurrently (needs GC). CLI `chmodLayerPath` always sends `Kind:"file"` even for directories.

## What's done right

- Entry + event in **one transaction** with a **single seq domain** (`event seq == entry_seq`, serialized via `FOR UPDATE`) — exactly the pattern we converged on in the #539 discussion; no dual-domain reset hazards here.
- `CanonicalizeDir` always stores base roots with a trailing `/`, so the `HasPrefix` base-root scope check has no `/repo` vs `/repo2` collision.
- Checkpoint-consistent reads via `entry_seq <= durable_seq` are sound frozen views.
- Add/add conflicts are caught by `BaseRevision=0` create-only write semantics, and file upserts do carry a revision CAS at apply.

## Summary

Single-agent: P0-1 alone (rename/chmod dedup) can silently lose data in trivial flows and should block merge. Multi-agent shared workspace: commit currently provides neither isolation nor safe compensation (P1-1/P1-2) — concurrent writers can have their data deleted by another agent's failed commit. The layer write path (datastore) is solid; the commit-to-base path needs a redesign pass: ordered-log replay, CAS on every apply op, guarded rollback, CAS state transitions + owner token, applied-entry ledger, and per-path authorization.

mornyx:
@qiffang I went through the new review items and pushed the fixes to `feat/layer` (latest head: `a0485b7`). Summary of what changed:

- Layer commit now uses the ordered replay log for apply/restore paths, while normal diff remains latest-entry-per-path.
- Layer routes are auth-scoped, including create/list/status/diff/events/checkpoint/object/commit/rollback/entry APIs.
- Commit/rollback now use CAS state transitions and reject unsafe committing re-entry instead of heuristic recovery.
- Commit preflight/rollback was hardened: base snapshots are validated before apply, rollback only covers applied entries, and directory whiteout is non-recursive/empty-dir only.
- Object upload now validates auth/base snapshot before storing the object, reducing orphan-object risk.
- CLI chmod now sends directory kind correctly.
- FUSE restore now replays repeated same-path upserts correctly, and layer truncate no longer writes through to the base filesystem.
- Latest fix: repeated same-path layer commit entries no longer reuse stale base revisions after the path has already been touched in the same commit; superseded mode operations no longer abort ordered commit replay; the layer smoke assertion now treats `applied` as replay-entry count rather than latest-diff count.

Tests added/expanded:

- datastore ordered-log and commit-state tests
- server replay tests for upsert+chmod, upsert+rename, mkdir+chmod, repeated same-path upserts, upsert+whiteout, commit retry/state conflict, auth, rollback boundaries, large objects
- FUSE restore tests for replay/max_seq, repeated upsert overwrite, and layer truncate isolation
- local e2e layer smoke coverage for same-path entries, large files, tags/names, checkpoint restore, FUSE restore/commit, whiteout/rename/symlink

Validation:

- `go test ./pkg/datastore ./pkg/server ./pkg/client ./pkg/fuse ./cmd/drive9/cli` passed locally.
- `make lint` passed locally.
- GitHub `code-ci / ci` passed on `a0485b7`.
- GitHub `local-e2e / drive9-server-local e2e smoke` passed on `a0485b7`, including Layer FS smoke, FUSE release gate, Git operations smoke, and portable pack/unpack.

Please take another look when you have time.

qiffang (approved):
## R3 Review — PR #507: Layered Filesystem Overlays

**Verdict: ✅ GREEN**

### R2→R3 Delta Summary

12 files changed, +1245 -134 lines over 3 commits (`95cfa500`, `16999a3d`, `a0485b78`).

### Key Changes Reviewed

#### 1. Ordered Replay Log (Critical Fix)
- New `ListFSLayerEntryLog` / `ListFSLayerEntryLogAtSeq` — returns ALL entries ordered by `entry_seq ASC`, not just latest-per-path.
- Commit now uses replay log instead of current-state diff: `upsert → chmod`, `upsert → rename`, `mkdir → chmod` sequences are no longer collapsed and lost.
- `ReplayFSLayer` client API with `?replay=1` query param.
- FUSE restore uses replay for both checkpoint and full restore paths.
- **Tests**: 5 new server tests (`TestFSLayerCommitReplaysUpsertBeforeChmod`, `TestFSLayerCommitReplaysSamePathUpsertsWithoutBaseRevConflict`, `TestFSLayerCommitAllowsSupersededUpsertModeBeforeWhiteout`, `TestFSLayerCommitReplaysUpsertBeforeRename`, `TestFSLayerCommitReplaysMkdirBeforeChmod`). Mount test `TestRestoreLayerEntriesReplaysSamePathUpsertOverwrite`. ✅

#### 2. Commit Reentry Prevention
- `BeginFSLayerCommit` no longer accepts `committing` state — prevents double-apply.
- `TestFSLayerCommitRejectsCommittingWithoutFence` replaces old retry test.
- Design doc updated: "没有 commit owner/epoch 的 `committing` layer 不允许重入 commit 或 rollback". ✅

#### 3. CAS State Transitions
- New `SetFSLayerStateIf` — conditional UPDATE with `WHERE state IN (?)`, returns `ErrFSLayerStateConflict` on miss.
- All commit error paths use `SetFSLayerStateIf(..., []FSLayerState{FSLayerStateCommitting}, ...)` — only transitions from `committing`, never from an unexpected state.
- `RollbackFSLayer` uses `SetFSLayerStateIf` with explicit allowed source states. ✅

#### 4. Auth Enforcement
- `authorizeFS` added to: layer get, diff, checkpoints, entries (read), events, rollback, commit, object read, object upload.
- `authorizeFSLayerEntryMutation` — per-op: whiteout → `FSOpDelete`, rename → `FSOpDelete` source + `FSOpWrite` target, default → `FSOpWrite`.
- `fsLayerVisibleToScope` filters layer list.
- `isScopedFSLayerRouteAllowed` — route whitelist for scoped tokens with strict query key validation.
- `TestFSLayerCreateRequiresScopedWriteOnBaseRoot`, `TestFSLayerEntryMutationRequiresScopedPathPermission`. ✅

#### 5. Layer-Aware Truncate
- `SetAttr` truncate-to-zero with layer enabled → writes `upsertLayerFile(ctx, path, nil, ...)` + shadow + pending, instead of hitting base filesystem.
- `TestLayerSetAttrTruncateWritesLayerEntryNotBase` — verifies no base PUT occurs. ✅

#### 6. Path Validation Fix
- `fsLayerPathWithinBase` and `fsLayerPathWithinBaseRoot` — fixed `strings.HasPrefix(p, baseRoot)` to `strings.HasPrefix(p, baseRoot+"/")` with trailing slash trim. Prevents `/repo-extra` matching `/repo`. ✅

#### 7. Commit Apply Hardening
- `ensureFSLayerEntryBaseStillMatches` — re-validates base revision/inode before destructive ops (whiteout, chmod, rename), skipping already-touched paths.
- `touchedPaths` map tracks applied entries, `fsLayerApplyExpectedRevision` returns `-1` (skip revision check) for already-touched paths.
- `fsLayerEntryModeSuperseded` — allows chmod `ErrNotFound` to pass when a later entry will overwrite the same path.
- `filterFSLayerSnapshotsForEntries` — rollback only rolls back snapshots for entries actually applied, not all snapshots. ✅

#### 8. Dir Safety
- Directory whiteout: `RemoveAllCtx` → `RemoveDirCtx` with explicit `ReadDir` + empty check.
- Directory restore snapshot: checks if dir already exists before remove+recreate, avoids unnecessary data loss. ✅

#### 9. FUSE Restore Replay
- `restoreLayerEntries` uses `ReplayFSLayer` instead of `DiffFSLayer`.
- `restoredUpserts` map tracks restored paths; repeated upserts for the same path overwrite correctly.
- `layerEntryFetchMaxSeq` uses `entry.EntrySeq` when available, falling back to checkpoint max_seq.
- All existing mount layer tests updated to assert `?replay=1`. ✅

#### 10. Misc
- `chmodLayerPath` CLI helper — detects dir vs file via `StatCtx` or trailing `/`.
- E2E: added chmod entries and upsert-before-rename coverage; commit applied count updated from 9→12.
- Design doc: PK includes `entry_seq`, doc explains why append-only log avoids data loss.

### Checklist
- [x] All inputs handled — auth checks on every endpoint, path validation fixed
- [x] Edge cases — upsert→chmod, upsert→rename, upsert→whiteout, mkdir→chmod, repeated upserts, superseded mode
- [x] Failure behavior — CAS transitions, reentry prevention, scoped rollback
- [x] Regression tests — 7 new server tests, 2 new mount tests, e2e coverage expanded
- [x] No race/state issues — `touchedPaths` is single-goroutine commit loop, CAS state transitions

### Non-blocking Observations
- O1: `fsLayerEntryModeSuperseded` does O(n²) scan in worst case (every entry checks remaining entries). Fine for V1 entry counts but could use a path→max_seq index if layers grow large.
- O2: Design doc acknowledges no crash-safe resumption — "生产级别要继续引入 applied-entry ledger / commit owner lease". Good V1 scope boundary.

No blockers. Ship it.

qiffang (commented):
## adversary-2 · R6 · PR #507 — Layered Filesystem Overlays

### Verdict: 🟢 GREEN

### Changes Since R5

R5 was GREEN on `2c3c76b`. Since then, 12 new commits landed:

- **`455a806`** — complete layer filesystem blockers
- **`26be238`** — fix layer smoke directory deletion check
- **`61ee9e4`** — rename layer API endpoints
- **`d622fb2`** — add layerfs feature matrix doc
- **`7babe0b`** — add manual real dev e2e runner
- **`8a952cf`** — scope realdev e2e to layer fs
- **`95cfa50`** — harden layer fs replay and auth
- **`16999a3`** — restore layered fuse truncate replay
- **`a0485b7`** — handle repeated layer commit entries

### R6 Re-review Scope

Full re-review of: server layer handlers (auth, state machine), datastore layer (TOCTOU, state transitions), FUSE integration (commit queue, truncate, restore, event watcher), and commit-time entry replay.

### Verification

**1. Auth enforcement — ALL layer endpoints checked:**
- `handleFSLayerCreate`: `authorizeFS(FSOpWrite, baseRoot)` ✓
- `handleFSLayerObject` dispatch: GET layer, diff, events all `authorizeFS(FSOpRead)` ✓; checkpoints/entries(POST)/objects(POST)/rollback/commit all `authorizeFS(FSOpWrite)` ✓
- `handleFSLayerEntryGet`: `authorizeFS(FSOpRead, path)` inside handler ✓
- `handleFSLayerObjectRead`: `authorizeFS(FSOpRead, path)` inside handler ✓
- `handleFSLayerCheckpointObject`: `authorizeFS(FSOpRead, layer.BaseRootPath)` ✓

**2. TOCTOU fix (from R4) — still present:**
- `UpsertFSLayerEntry`: `SELECT ... FOR UPDATE` inside transaction, then state check `if FSLayerState(state) != FSLayerStateActive { return ErrFSLayerStateConflict }` ✓

**3. Rollback context (from R4) — still present:**
- `rollbackCtx, rollbackCancel := context.WithTimeout(context.WithoutCancel(r.Context()), fsLayerRollbackTimeout)` ✓
- All rollback operations use `rollbackCtx` ✓

**4. Truncate replay (new `16999a3`):**
- Layer-enabled truncate writes a layer entry via `upsertLayerFile` + updates shadow + pending index ✓
- Non-layer path unchanged (server-side WriteCtx + stat refresh) ✓
- Test `TestLayerSetAttrTruncateWritesLayerEntryNotBase` verifies no base PUT ✓

**5. Repeated commit entries (new `a0485b7`):**
- `fsLayerEntryModeSuperseded`: scans forward for same-path or parent-whiteout/rename entries ✓
- Used to skip redundant mode operations during commit ✓
- `fsLayerApplyExpectedRevision`: uses `-1` (skip revision check) for already-touched paths ✓

**6. Layer event watcher:**
- Polls every 1s, calls `restoreLayerEntries` on change ✓
- Properly stopped via `sync.Once` + cancel + `<-done` ✓
- Disabled when `CheckpointRef` is set (static view) ✓

**7. Commit queue layer integration:**
- `uploadLayerEntry`: large files via `UploadFSLayerFile`, inline via `UpsertFSLayerEntry` ✓
- `onCommitSuccess`: layer mode retains shadow+pending (not cleaned up until layer commit/rollback) ✓
- `MarkCommitted` updates pending entry with committed revision ✓

**8. State machine transitions verified:**
- active → sealed → committing → committed (happy path) ✓
- active/sealed → abandoned (rollback) ✓
- committing → conflicted (on failure) ✓
- committed → idempotent return ✓
- `BeginFSLayerCommit`: CAS `WHERE state IN (active, sealed)` ✓

**O1** — `onCommitSuccess` calls `cq.layerRefSnapshot()` 3 additional times after capturing `layerRef` at the top. Should reuse the captured value. Practically safe since `SetLayerRef` is only called during mount setup.

**O2** — Feature matrix doc (`layered-filesystem-feature-matrix.md`) is a comprehensive gap tracking table. Good practice for a feature of this scope.

### Checklist

- [x] All inputs handled (path validation, state checks, auth)
- [x] Edge cases covered (repeated entries, truncate, rename, whiteout, chmod on different kinds)
- [x] Failure behavior defined (rollback with dedicated context, state → conflicted)
- [x] API explicit (layer lifecycle states, entry ops)
- [x] Tests complete: 1226-line mount_layer_test.go, 905-line fs_layer_test.go, 334-line fs_layer datastore test, 361-line commit_queue_test.go, 912-line e2e smoke script

PR #496, opened 2026-06-03
feat: add block-level parallel reads

qiffang:
## Summary by CodeRabbit

* **New Features**
  * Configurable parallel reading for large files via CLI flags with defaults and validation; prefetch/read now use block-based parallel fetches with bounded concurrency and a runtime tuning method.

* **Bug Fixes**
  * Cancellable read retries improved; short/partial block reads surface EOF appropriately.
  * Disk read cache now rejects length-mismatched writes.

* **Tests**
  * Added tests covering CLI validation, parallel-read behavior, prefetch semantics, cache length checks, and error propagation.

Issue #27, opened 2026-03-26
meta.Store missing path segment length limits — file_nodes.name accepts arbitrarily long filenames


qiffang:
## Current State

\`pathutil.Canonicalize\` performs character-level security checks (NUL, control characters, backslash, \"..\", Unicode NFC), but **has no length limits**:

- No per-segment length limit (design doc DDL defines \`name VARCHAR(255)\`, but code doesn't check)
- No total path length limit (design doc DDL defines \`path VARCHAR(4096)\`, but code doesn't check)
- SQLite's TEXT type has no length limit and won't auto-truncate

## Impact

- Malicious clients can send extremely long path names to consume memory/storage
- Migrating to db9/MySQL will cause VARCHAR(255) truncation or errors
- Extremely long filenames are unreadable in CLI \`ls\` output

## Proposed Fix

Add limits in \`pathutil.Canonicalize\`:

\`\`\`go
const (
    MaxSegmentLen = 255
    MaxPathLen    = 4096
)

func canonicalize(raw string, isDir bool) (string, error) {
    if len(raw) > MaxPathLen {
        return \"\", fmt.Errorf(\"path too long: %d > %d\", len(raw), MaxPathLen)
    }
    for _, seg := range segments {
        if len(seg) > MaxSegmentLen {
            return \"\", fmt.Errorf(\"segment too long: %q\", seg[:20]+\"...\")
        }
    }
}
\`\`\`

PR #508, opened 2026-06-07
e2e: add opt-in FUSE performance baseline

qiffang:
## Summary

Stacked on PR #496 head `1ae0320` (`dat9-dev2/issue-489-parallel-read`). This PR should be retargeted/rebased to `main` after #496 lands.

Adds an opt-in real-FUSE performance baseline:
- new `e2e/fuse-performance-baseline.sh` covering deterministic small-file write/read, large-file write/read + repeated read passes, and simple SQLite insert/update/read transaction metrics;
- correctness assertions remain pass/fail; performance values are threshold-free JSON metrics;
- `fuse-release-gate.sh` only runs it when `RUN_FUSE_PERFORMANCE_BASELINE=1`, so ordinary PR/local-e2e remains unchanged;
- `local-e2e.yml` exposes manual `workflow_dispatch` input `run_fuse_performance_baseline=1` and uploads metrics/mount logs as `fuse-performance-baseline` artifact.

## Validation

- `bash -n e2e/fuse-performance-baseline.sh e2e/fuse-release-gate.sh`
- embedded Python workload dry run with small local temp fixture
- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/local-e2e.yml")'`
- `git diff --check`
- `/usr/local/go/bin/go test ./cmd/drive9 ./cmd/drive9/cli ./pkg/fuse -count=1`
- `/usr/local/go/bin/go vet ./cmd/drive9 ./cmd/drive9/cli ./pkg/fuse`
- `PATH=/usr/local/go/bin:$PATH make build-cli CLI_BIN=/tmp/drive9-fuse-performance-cli`

## Notes

This is a baseline/artifact PR, not a throughput regression gate. It intentionally does not add hardcoded latency or throughput thresholds.

## Summary by CodeRabbit

* **New Features**
  * Daily opt-in FUSE performance baseline runs with CI artifact upload and optional remote archival to Drive9; new opt-in concurrency/churn SQLite modes and an opt-in concurrency-stress runner; per-mount transient local overlay to improve SQLite correctness.

* **Testing**
  * Vastly expanded unit and e2e coverage for SQLite/WAL, persistent-journal reads, streaming uploads, revision/commit flows, and uploader revision handling.

* **Documentation**
  * e2e docs and README updated with baseline run instructions, knobs, artifact locations, and release-gate guidance.


qiffang:
Implementation update for task #27.

This PR is stacked on PR #496 head `1ae0320`; base is intentionally `dat9-dev2/issue-489-parallel-read` until #496 lands.

What changed:
- Added opt-in `e2e/fuse-performance-baseline.sh` for real-FUSE small-file, large-file, repeated large-read, and SQLite transaction/read metrics.
- Metrics are threshold-free JSON; correctness remains enforced by checksums and SQLite `PRAGMA integrity_check`.
- Wired `RUN_FUSE_PERFORMANCE_BASELINE=1` into `e2e/fuse-release-gate.sh`; default PR/local-e2e remains unchanged.
- Added manual `local-e2e` `workflow_dispatch` input `run_fuse_performance_baseline=1` and artifact upload for metrics + mount log.

Local validation:
- `bash -n e2e/fuse-performance-baseline.sh e2e/fuse-release-gate.sh`
- embedded Python workload dry run with small local temp fixture
- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/local-e2e.yml")'`
- `git diff --check`
- `/usr/local/go/bin/go test ./cmd/drive9 ./cmd/drive9/cli ./pkg/fuse -count=1`
- `/usr/local/go/bin/go vet ./cmd/drive9 ./cmd/drive9/cli ./pkg/fuse`
- `PATH=/usr/local/go/bin:$PATH make build-cli CLI_BIN=/tmp/drive9-fuse-performance-cli`

Live validation:
- `local-e2e` workflow_dispatch with `run_fuse_performance_baseline=1`: <https://github.com/mem9-ai/drive9/actions/runs/27081737474> — SUCCESS
- Confirmed logs show `=== drive9 FUSE performance baseline ===`, `RUN_FUSE_PERFORMANCE_BASELINE=1`, metrics JSON output, and artifact upload.
- Artifact `fuse-performance-baseline` was uploaded; downloaded and verified it contains `performance-metrics-*.json` plus `mount-*.log`.
- Example metrics from the run: 16MiB large read pass 1 ~23.48 MiB/s, pass 2 ~1198.86 MiB/s; SQLite insert transaction 256 rows ~1821 rows/s; `integrity_check=ok`.

qiffang:
R1 review-gate: REVISE

Blocking finding:

- `e2e/fuse-performance-baseline.sh:503-513` does not actually validate SQLite row payload correctness. The read pass selects only `id`, `version`, stored `checksum`, and `LENGTH(payload)`, then fingerprints those values. It never reads `payload` and recomputes `sha256(payload)` against the stored checksum. A corrupted payload with the same length and unchanged checksum column would pass this workload and still emit a valid-looking metrics artifact. This directly weakens the PR’s advertised SQLite write/read correctness signal.

Required fix: during the SQLite read aggregate pass, read the actual payload bytes, recompute `sha256(payload)` for every row, compare it to the stored checksum, and fail on mismatch. The emitted fingerprint should be derived from verified row data, not from the stored checksum alone.

What I verified:

- PR is stacked on #496 head `1ae0320`; reviewed diff against that base.
- Manual `local-e2e` run `27081737474` used head `d24d1b8`, set `RUN_FUSE_PERFORMANCE_BASELINE=1`, executed `=== drive9 FUSE performance baseline ===`, produced `RESULT: 17/17 passed, 0 failed`, and uploaded artifact `fuse-performance-baseline`.
- Downloaded artifact contains `performance-metrics-*.json` and `mount-*.log`; metrics JSON is parseable and includes small-file, large-file, repeated large-read, and SQLite metrics.
- Local checks passed: `bash -n e2e/fuse-performance-baseline.sh e2e/fuse-release-gate.sh`, workflow YAML parse, `git diff --check 1ae0320..HEAD`, `/usr/local/go/bin/go test ./cmd/drive9 ./cmd/drive9/cli ./pkg/fuse -count=1`, and `/usr/local/go/bin/go vet ./cmd/drive9 ./cmd/drive9/cli ./pkg/fuse`.

The opt-in workflow/artifact wiring looks correct; the blocker is the SQLite correctness assertion inside the new baseline.

qiffang:
R1 blocker fixed in head `125cdd3`.

Change:
- SQLite read metric now selects `payload` bytes, recomputes `sha256(payload)`, and compares it with the stored row checksum.
- A corrupted row payload now fails the baseline with `sqlite payload checksum mismatch row=<id>` instead of passing on stored checksum + length only.
- Metrics now include `payload_verified_rows`, and the shell gate asserts it equals `FUSE_PERF_SQLITE_ROWS`.
- Docs were updated to state that SQLite reads verify stored row payload bytes against row checksums.

Local validation rerun:
- `bash -n e2e/fuse-performance-baseline.sh e2e/fuse-release-gate.sh`
- embedded Python workload dry run with `payload_verified_rows` assertion
- YAML parse for `.github/workflows/local-e2e.yml`
- `git diff --check`
- `/usr/local/go/bin/go test ./cmd/drive9 ./cmd/drive9/cli ./pkg/fuse -count=1`
- `/usr/local/go/bin/go vet ./cmd/drive9 ./cmd/drive9/cli ./pkg/fuse`
- `PATH=/usr/local/go/bin:$PATH make build-cli CLI_BIN=/tmp/drive9-fuse-performance-cli`

Live validation rerun:
- `local-e2e` workflow_dispatch with `run_fuse_performance_baseline=1`: <https://github.com/mem9-ai/drive9/actions/runs/27082035743> — SUCCESS
- Logs show `RUN_FUSE_PERFORMANCE_BASELINE=1`, `=== drive9 FUSE performance baseline ===`, and `payload_verified_rows: 256`.
- Artifact `fuse-performance-baseline` uploaded successfully; downloaded artifact and verified metrics JSON has `rows=256`, `payload_verified_rows=256`, and `integrity_check=ok`.

qiffang:
R2 review-gate: GREEN

The R1 blocker is fixed in head `125cdd3`.

What I verified:

- The SQLite read aggregate now selects actual `payload` bytes, recomputes `sha256(payload)`, compares it with the stored checksum for every row, and fails with `sqlite payload checksum mismatch row=<id>` on mismatch.
- The metrics JSON now carries `payload_verified_rows`, and the shell gate asserts it equals `FUSE_PERF_SQLITE_ROWS`.
- Manual `local-e2e` run `27082035743` used head `125cdd3`, set `RUN_FUSE_PERFORMANCE_BASELINE=1`, executed the performance baseline, produced `RESULT: 18/18 passed, 0 failed`, and uploaded artifact `fuse-performance-baseline`.
- Downloaded artifact metrics are parseable and show `rows=256`, `payload_verified_rows=256`, `integrity_check=ok`, and two large-file read passes.

Local validation passed:

- `bash -n e2e/fuse-performance-baseline.sh e2e/fuse-release-gate.sh`
- workflow YAML parse
- `git diff --check 1ae0320..HEAD`
- `/usr/local/go/bin/go test ./cmd/drive9 ./cmd/drive9/cli ./pkg/fuse -count=1`
- `/usr/local/go/bin/go vet ./cmd/drive9 ./cmd/drive9/cli ./pkg/fuse`

Scope note: PR #508 is still stacked on PR #496 (`dat9-dev2/issue-489-parallel-read`). After #496 lands, retarget/rebase to `main` and rerun the flagged `local-e2e` before merge.

qiffang:
Implemented the daily heavy FUSE baseline/archive follow-up on latest head `269a491`.

Changes:
- `local-e2e` now has a daily schedule at `20:17 UTC` / `04:17 Asia/Shanghai`.
- Scheduled runs enable SQLite WAL, churn, concurrency, performance baseline, metrics archive, and the FUSE concurrency diagnostic.
- Performance metrics are still uploaded as the `fuse-performance-baseline` artifact and are also copied to the Drive9 CI workspace under `/benchmarks/fuse-performance/<YYYY>/<MM>/<DD>/<branch>/<sha>/<run_id>-<attempt>/` plus `latest.json` pointers.
- FUSE concurrency stress is now a separate scheduled/manual diagnostic after the release gate and archive, so a stress failure cannot prevent baseline metrics archival. Manual `workflow_dispatch` with `run_fuse_concurrency_stress=1` still treats stress as required.

Validation:
- `bash -n scripts/archive-fuse-performance-metrics.sh .github/workflows/local-e2e.yml e2e/fuse-release-gate.sh e2e/fuse-performance-baseline.sh e2e/fuse-concurrency-stress.sh e2e/fuse-sqlite-correctness.sh`
- YAML parse for `.github/workflows/local-e2e.yml`
- archive dry-run with and without metrics
- `git diff --check`
- `go test ./pkg/fuse ./cmd/drive9 -count=1`
- `go vet ./pkg/fuse ./cmd/drive9`
- workflow_dispatch heavy run succeeded: https://github.com/mem9-ai/drive9/actions/runs/27083337041

Run `27083337041` confirmed:
- `RUN_FUSE_SQLITE_WAL=1`, `RUN_FUSE_SQLITE_CHURN=1`, `RUN_FUSE_SQLITE_CONCURRENCY=1`, `RUN_FUSE_PERFORMANCE_BASELINE=1`, `ARCHIVE_FUSE_PERFORMANCE_METRICS=1`
- FUSE smoke `91/91`, correctness `48/48`, SQLite `20/20`, performance baseline `18/18`
- uploaded artifact `fuse-performance-baseline` / artifact id `7460931117`
- archived metrics to `/benchmarks/fuse-performance/2026/06/07/dat9-dev2-fuse-performance-baseline/269a4918317f/27083337041-1`
- artifact JSON verified `payload_verified_rows=256` and `integrity_check=ok`

qiffang:
Review-gate R3: REVISE.

I verified head `269a491` and the flagged run `27083337041`: SQLite WAL/churn/concurrency and the performance baseline executed successfully, the artifact was uploaded, and the archive step wrote to `/benchmarks/fuse-performance/2026/06/07/dat9-dev2-fuse-performance-baseline/269a4918317f/27083337041-1`. Local `bash -n`, workflow YAML parse, `git diff --check`, `go test ./pkg/fuse ./cmd/drive9`, `go vet ./pkg/fuse ./cmd/drive9`, and archive dry-run passed.

Blocking issues:

1. `e2e/fuse-sqlite-correctness.sh` still trusts the stored checksum instead of verifying payload bytes. The workload-side fingerprint reads only `checksum, LENGTH(payload)` at lines 390-400, and the verifier repeats the same pattern at lines 744-750. Same-length payload corruption can pass `PRAGMA integrity_check`, row count, length, and stored-checksum digest. This is the same class of issue fixed in PR #508's performance baseline R1; the default/heavy SQLite correctness path needs to read `payload`, recompute `sha256(payload)`, compare it to the stored checksum, and build the digest from verified payload bytes.

2. The `negotiatedInlineThreshold()==0` path can still route non-empty writes into `PatchFile`. In `syncWriteHandleToRemoteLocked` lines 6946-6970 and `flushHandle` lines 8367-8391, `threshold == 0` makes `useDirectPUT` false and `fh.OrigSize >= threshold` true, so new files / small-to-large writes take the patch branch even though comments and tests say unknown `/v1/status` must force multipart-safe full upload. Guard the patch branch with `threshold > 0` in both paths.

3. The claimed heavy daily gate is not actually a gate for FUSE concurrency stress. Scheduled runs set `RUN_FUSE_CONCURRENCY_STRESS=1`, but the stress step uses `continue-on-error: ${{ github.event_name == 'schedule' }}` and enforcement only checks it when `FUSE_CONCURRENCY_STRESS_REQUIRED=1`, which is manual-only. A scheduled run can be green while this expensive e2e fails. If this is intentionally diagnostic, it should not be counted as satisfying the daily heavy e2e gate; otherwise make scheduled stress failure fail the workflow. Also note the validated run `27083337041` had `RUN_FUSE_CONCURRENCY_STRESS=0`, so it did not exercise this scheduled path.

Scope note: this PR archives baseline metrics but does not implement performance-regression comparison/fail thresholds. That can be a separate follow-up if qiffang accepts archive-only as phase 1; otherwise it is still missing from the earlier requested "性能差了 e2e 就失败" behavior.

qiffang:
R3 blockers fixed in head `55468b8`.

Fixes:
- `e2e/fuse-sqlite-correctness.sh` now verifies actual SQLite `payload` bytes in both the workload-side expected fingerprint and the verifier-side snapshot fingerprint. It recomputes `sha256(payload)`, compares it to the stored row checksum, and fails with `payload checksum mismatch` on corruption.
- `syncWriteHandleToRemoteLocked` and `flushHandle` now guard the `PatchFile` branch with `threshold > 0`, so unknown `negotiatedInlineThreshold()==0` cannot route non-empty writes into patch mode.
- Scheduled FUSE concurrency stress is now required when enabled: `FUSE_CONCURRENCY_STRESS_REQUIRED=1` for schedule and for manual `run_fuse_concurrency_stress=1`, and the stress step no longer uses scheduled `continue-on-error`.

Regression/validation:
- Added `TestUnknownInlineThresholdNonEmptyHandleUploadUsesMultipart`, covering both `flushHandle` and write-sync with unknown threshold and asserting multipart upload, no direct PUT.
- `bash -n scripts/archive-fuse-performance-metrics.sh .github/workflows/local-e2e.yml e2e/fuse-release-gate.sh e2e/fuse-performance-baseline.sh e2e/fuse-sqlite-correctness.sh e2e/fuse-concurrency-stress.sh`
- YAML parse for `.github/workflows/local-e2e.yml`
- embedded SQLite verifier corruption check proves changed payload bytes with stale checksum now fail
- `go test ./pkg/fuse -run 'TestUnknownInlineThresholdNonEmptyHandleUploadUsesMultipart|TestNegotiatedInlineThresholdSeparatesProtocolFromHeuristic|TestInitWarmTimeoutFallsBackToDefault' -count=1`
- archive dry-run with empty metrics dir and `DRIVE9_PERF_ARCHIVE_ALLOW_EMPTY=1`
- `go test ./pkg/fuse ./cmd/drive9 -count=1`
- `go vet ./pkg/fuse ./cmd/drive9`
- `git diff --check`

GitHub validation:
- PR `pull_request` local-e2e is green on `55468b8`: <https://github.com/mem9-ai/drive9/actions/runs/27083810188>
- Flagged stable heavy run is green on `55468b8`: <https://github.com/mem9-ai/drive9/actions/runs/27083815188>

Run `27083815188` confirmed:
- `RUN_FUSE_SQLITE_WAL=1`, `RUN_FUSE_SQLITE_CHURN=1`, `RUN_FUSE_SQLITE_CONCURRENCY=1`, `RUN_FUSE_PERFORMANCE_BASELINE=1`, `ARCHIVE_FUSE_PERFORMANCE_METRICS=1`
- FUSE smoke `91/91`, correctness `48/48`, SQLite `20/20`, performance baseline `18/18`
- artifact `fuse-performance-baseline` uploaded and downloaded locally
- metrics JSON verified `schema=drive9-fuse-performance/v1`, `payload_verified_rows=256`, `integrity_check=ok`
- Drive9 archive wrote to `/benchmarks/fuse-performance/2026/06/07/dat9-dev2-fuse-performance-baseline/55468b8bedb2/27083815188-1` plus latest pointers

Note: `27083815188` intentionally had `RUN_FUSE_CONCURRENCY_STRESS=0`, so it validates the stable SQLite/performance/archive path after the code fixes. The scheduled/manual stress hard-gate behavior is covered by workflow logic and local YAML review; a `run_fuse_concurrency_stress=1` dispatch would be the separate actual stress evidence.

qiffang:
Review-gate R4: GREEN for the R3 blockers at head `55468b8`.

What I rechecked:
- `e2e/fuse-sqlite-correctness.sh` now reads actual `payload` bytes in both the workload-side expected fingerprint and verifier, recomputes `sha256(payload)`, and fails on checksum mismatch before accepting the fingerprint.
- `syncWriteHandleToRemoteLocked` and `flushHandle` now guard the `PatchFile` branch with `threshold > 0`, so `negotiatedInlineThreshold()==0` sends non-empty writes through the multipart-safe full upload path. The new regression covers both flush and write-sync.
- Scheduled/manual stress gating no longer has `continue-on-error`; `FUSE_CONCURRENCY_STRESS_REQUIRED` is `1` for schedule and for manual runs with `run_fuse_concurrency_stress=1`, so an enabled stress failure now fails the workflow.

Validation:
- Local: `bash -n` for affected e2e/archive scripts, workflow YAML parse, `git diff --check origin/main..HEAD`, `go test ./pkg/fuse ./cmd/drive9 -count=1`, and `go vet ./pkg/fuse ./cmd/drive9`.
- GitHub PR checks at `55468b8`: `ci` success, ordinary `local-e2e` success, CodeRabbit status success/no new actionable comments.
- Flagged run `27083815188` success at `55468b8`; logs confirm `RUN_FUSE_SQLITE_WAL=1`, `RUN_FUSE_SQLITE_CHURN=1`, `RUN_FUSE_SQLITE_CONCURRENCY=1`, and `RUN_FUSE_PERFORMANCE_BASELINE=1`; SQLite workload `20/20`, performance baseline `18/18`, artifact uploaded, and Drive9 archive wrote `/benchmarks/fuse-performance/2026/06/07/dat9-dev2-fuse-performance-baseline/55468b8bedb2/27083815188-1`. Downloaded metrics JSON has `payload_verified_rows=256`, `integrity_check=ok`, and two large-read passes.

Caveats, non-blocking for this PR:
- Run `27083815188` did not exercise FUSE concurrency stress (`RUN_FUSE_CONCURRENCY_STRESS=0`); I verified the workflow logic that makes stress required when schedule/manual enables it, but this run is not evidence that the stress workload itself currently passes.
- Performance-regression comparison/fail thresholds are still not implemented; this PR provides scheduled baseline/archive data. Treat threshold-based regression gating as a separate follow-up unless qiffang wants it folded into this PR.

mornyx (changes requested):
I found blocking issues in the FUSE changes.

1. `pkg/fuse/commit_queue.go:667-683` / `pkg/fuse/dat9fs.go:7980-7995` - CommitQueue entries are not snapshot-isolated or same-path serialized. Release enqueues only path/size/baseRev, then `uploadEntry` later re-reads the mutable shadow by path. A second writer can open the same path while the first entry is queued or in flight, preload that same shadow, and stage new bytes back to the same shadow file before the first worker reads or completes it. The first entry can then upload the second writer bytes under the first entry metadata/revision, and on success `onCommitSuccess` removes the shared shadow/pending state. This can leave inode/cache size from entry A while remote bytes are from entry B, or drop the second pending entry. Please either snapshot/pin a per-entry shadow generation at enqueue time and upload that generation, or serialize all same-path background commits with later same-path writers until the queued entry owns a stable snapshot.

2. `pkg/fuse/dat9fs.go:467-471` - the transient SQLite `-shm` fail-closed branch is currently unreachable. `localOverlayForPathWithHint` has an EIO branch for `fs.transientLocalOverlay == nil`, but `usesTransientLocalOverlay` returns false when the overlay is nil, so `.db-shm` falls through to the remote path if transient overlay initialization failed or cache setup is unavailable. That breaks the invariant that `-shm` stays local/transient and can leak it into the Drive9 namespace. The predicate should decide based on path/read-only/dirHint only; the caller should return EIO when the overlay is required but unavailable. Please add a regression test for nil overlay.

3. `pkg/fuse/mount.go:255` - the transient overlay is described as mount-local runtime state, but it is stored under `cacheBase/mountHash/transient`, where `mountHash` excludes credentials and is reused across remounts at the same mount point. Because `EnsureRoot` does not clear it, an old tenant or previous mount can leave `/repo/*.db-shm` files that a later mount will merge and serve locally. Please scope this directory to the credential or a mount instance, or clear the transient root on mount startup.

Checked: `bash -n` on `e2e/fuse-performance-baseline.sh`, `e2e/fuse-sqlite-correctness.sh`, `e2e/fuse-release-gate.sh`, and `scripts/archive-fuse-performance-metrics.sh` passed. I did not run live FUSE/TiDB e2e in this review.

qiffang:
Addressed the latest `mornyx` blocking review in head `0677c8e`.

Changes:
- `CommitQueue` now serializes same-path entries in `inFlight`, and holds the Dat9FS per-path commit lock through `uploadEntry` + success cleanup. Later same-path writable FUSE operations wait for queued/in-flight commits before mutating shared shadow state.
- Release and pending-new rename enqueue paths now hold the same per-path commit lock until the queued entry is visible or the synchronous fallback finishes, so no later writer can mutate the shadow in the enqueue gap.
- SQLite `*.db-shm` transient-overlay routing now fails closed when the overlay is required but unavailable; it no longer falls through to remote.
- Transient overlay roots are now credential-scoped via the read-cache hash and mount-instance-scoped with a per-mount suffix, so remounts/tenants do not reuse old `-shm` state.

Tests added:
- nil transient overlay for SQLite WAL-index sidecars returns local `EIO` without remote access.
- commit queue same-path `beginInFlight` serialization.
- `CommitNow` holds the path lock through success cleanup.
- writable operations wait for queued same-path commits.
- transient overlay root differs by credential and mount instance.

Validation:
- `/usr/local/go/bin/go test ./cmd/drive9 ./cmd/drive9/cli ./pkg/fuse -count=1`
- `/usr/local/go/bin/go vet ./pkg/fuse ./cmd/drive9 ./cmd/drive9/cli`
- `/usr/local/go/bin/go test -race ./pkg/fuse -run 'TestWaitQueuedRemoteCommitBeforeWriteBlocksUntilQueuedCommitDone|TestCommitQueueBeginInFlightSerializesSamePath|TestCommitQueueCommitNowHoldsPathLockThroughSuccessCleanup|TestRenamePendingNewCommit' -count=1`
- `git diff --check`
- GitHub checks on `0677c8e`: `ci` passed, `drive9-server-local e2e smoke` passed. CodeRabbit is still pending at the time of this comment.

mornyx (changes requested):
I re-reviewed the latest head `0677c8e`. The transient `-shm` fail-closed behavior and credential/per-mount transient root issue look fixed now, but I still cannot approve because the CommitQueue path-locking change does not fully close the mutable-shadow races.

Blocking issues:

1. `pkg/fuse/dat9fs.go:1435`, `pkg/fuse/dat9fs.go:6883`, `pkg/fuse/dat9fs.go:6907`, `pkg/fuse/dat9fs.go:7971`, `pkg/fuse/dat9fs.go:8059` - CommitQueue entries still do not own a stable shadow snapshot. The new lock protects individual shadow writes and later queued upload/cleanup, but there is still a window after `Flush` stages the shadow and before `Release` enqueues the `CommitEntry`. During that window `commitQueue.HasPath(path)` is false, so a second writer can open/write the same path and mutate the same active shadow before the first handle enqueues. The first queued entry can then carry entry-A metadata such as size/base revision while uploading entry-B bytes. Please make the queued entry own a per-entry snapshot/generation at stage time, or reserve the path from stage through enqueue so later writers either block or get a new shadow.

2. `pkg/fuse/dat9fs.go:3865` / `pkg/fuse/dat9fs.go:4180` - `waitQueuedRemoteCommitBeforeWrite` waits and immediately releases the lock returned by `lockWritableRemoteCommitPath`, so callers are not protected during their actual mutation. For example, `SetAttr` waits at line 4180, releases the path lock, and only then performs remote chmod/truncate; a same-path release can enqueue/upload in between. Please return the unlock function and hold it through the caller critical section, or call `lockWritableRemoteCommitPath` directly around the mutation/setup.

3. `pkg/fuse/dat9fs.go:7971` - The shadow-spill enqueue-failure fallback still leaves stale local state after a successful synchronous upload. If `CommitQueue.Enqueue` fails but `uploadFromShadowRemote` succeeds, this branch does not finalize the handle, update committed revision, or remove `pendingIndex`/`shadowStore`, so the uploaded file remains shadowed locally and can be reuploaded or served from stale pending state later. Please mirror the strict shadow upload success cleanup/finalization path here.

4. `pkg/fuse/commit_queue.go:530` - `beginInFlight` serializes same-path entries but does not enforce FIFO. With multiple workers, a later same-path entry can receive from `workCh` and take `inFlight[path]` before the older worker gets scheduled. If same-path entries coexist, upload order becomes nondeterministic. The added test calls `beginInFlight(first)` synchronously before starting the second entry, so it does not cover this scheduler inversion. Please gate `beginInFlight` on the oldest queued entry for that path, or dispatch per path from an ordered queue.

Checked: `git diff --check 68ce029f...HEAD` is clean, and the focused new/related `pkg/fuse` tests pass:
`go test ./pkg/fuse -run TestCommitQueueBeginInFlightSerializesSamePath|TestCommitQueueCommitNowHoldsPathLockThroughSuccessCleanup|TestSQLiteWALIndexSidecarFailsClosedWhenTransientOverlayMissing|TestWaitQueuedRemoteCommitBeforeWriteBlocksUntilQueuedCommitDone|TestTransientOverlayRootUsesCredentialScopedMountInstance`.
I did not run the live FUSE/TiDB e2e in this pass.

qiffang:
Addressed the latest review blockers in head `370451f`.

What changed:
- `CommitQueue.beginInFlight` now preserves same-path FIFO order instead of letting a later worker win the `inFlight` slot first.
- Same-path remote commit protection is now held across mutable local state critical sections instead of wait/release:
  - `SetAttr`, `Create`, `Open`, `Write`, `Flush`, and `Release` now hold or transfer the path lock through mutation/stage/enqueue/sync paths as needed.
  - staged shadow snapshots hold a handle-owned path lock from `Flush` staging until `Release` enqueues or sync-fallback completes, closing the `Flush`→`Release` shadow mutation window without blocking pre-stage writable opens.
- Pending-new rename now acquires the destination path lock before renaming shadow/pending-index local state and before enqueueing.
- ShadowSpill enqueue-failure fallback now finalizes local state after a successful synchronous upload: clears stale read targets, updates committed revision/inode metadata when available, removes pending/shadow state, and clears pending chmod state.
- The writeback unknown stream revision regression now forces multipart/V2 upload instead of accidentally covering the direct-PUT path.

Validation run locally:
- `/usr/local/go/bin/go test ./pkg/fuse -run 'TestCommitQueueBeginInFlight|TestCommitQueueCommitNowHoldsPathLockThroughSuccessCleanup|TestWaitQueuedRemoteCommitBeforeWriteBlocksUntilQueuedCommitDone|TestWriteBackUploader_OnSuccessReceivesUnknownStreamRevision|TestRenamePendingNewCommit|TestSQLiteWALIndexSidecarFailsClosedWhenTransientOverlayMissing|TestTransientOverlayRootUsesCredentialScopedMountInstance' -count=1`
- `/usr/local/go/bin/go test ./pkg/fuse -count=1 -timeout=5m`
- `/usr/local/go/bin/go test ./cmd/drive9 ./cmd/drive9/cli ./pkg/fuse -count=1`
- `/usr/local/go/bin/go vet ./pkg/fuse ./cmd/drive9 ./cmd/drive9/cli`
- `/usr/local/go/bin/go test -race ./pkg/fuse -run 'TestWaitQueuedRemoteCommitBeforeWriteBlocksUntilQueuedCommitDone|TestCommitQueueBeginInFlight|TestCommitQueueCommitNowHoldsPathLockThroughSuccessCleanup|TestRenamePendingNewCommit' -count=1`
- `git diff --check`

qiffang:
Follow-up after CI lint on the previous push: amended the fix into head `241fd24` and removed the now-unused `uploadFromShadowRemote` wrapper.

Additional local validation after the cleanup:
- `/usr/local/go/bin/go test ./pkg/fuse -count=1 -timeout=5m`
- `/usr/local/go/bin/go test ./cmd/drive9 ./cmd/drive9/cli ./pkg/fuse -count=1`
- `/usr/local/go/bin/go vet ./pkg/fuse ./cmd/drive9 ./cmd/drive9/cli`
- `PATH="/usr/local/go/bin:$PATH" make lint` → `0 issues`
- `git diff --check`

mornyx (changes requested):
I re-reviewed the latest head `241fd24`. The previous queue/snapshot blockers are mostly addressed: same-path FIFO now checks the oldest queued entry, callers now hold the writable path lock through their critical sections, and the shadow cleanup path is much better. I still cannot approve because I found two blocking issues.

Blocking issues:

1. `pkg/fuse/dat9fs.go:8030-8067` - The ShadowSpill enqueue-failure fallback drops deferred chmods. The queued `CommitEntry` carries `Mode`/`HasMode`, and the normal CommitQueue path applies that mode after upload in `onCommitSuccess` before cleanup (`pkg/fuse/commit_queue.go:846-863`). In this synchronous fallback, after `uploadFromShadowRemoteWithRevision` succeeds, the code only updates the local inode mode and then clears pending mode for the inode; it never calls `applyPendingModeWithTimeoutLocked` or `applyRemoteMode`. If the queue is full/stopped after a dirty handle had a deferred chmod, `Release` can return success with file bytes uploaded but the remote mode left stale, and the local pending chmod has been discarded. Please apply the same post-upload remote chmod semantics as the normal CommitQueue/strict sync paths before clearing pending mode; if chmod fails, preserve/retry the pending mode instead of silently dropping it.

2. `pkg/fuse/dat9fs.go:7628`, `pkg/fuse/dat9fs.go:7646`, and `pkg/fuse/dat9fs.go:7788-7791` - `Fsync` now calls `stageShadowForQueuedCommitLocked`, which stores `fh.RemoteCommitUnlock`, but that lock is only released by `Release`. Unlike `Flush`, `Fsync` does not imply close; a process can fsync and keep the fd open. Any other same-path writer then blocks in `lockWritableRemoteCommitPath` until the first fd is released, even though there is no queued/in-flight commit yet. This is a same-path concurrency regression and can surface as a hung FUSE operation for long-lived descriptors. Please avoid holding the per-path commit lock across `Fsync`, enqueue a stable snapshot at Fsync time, or split the helper so only the Flush-to-Release path reserves the shadow until close.

Checked locally:
- `git diff --check 68ce029f889a1a6ac17b07fb9d6b5849ce39631b..HEAD`
- `go test ./pkg/fuse -run TestCommitQueueBeginInFlight|TestCommitQueueCommitNowHoldsPathLockThroughSuccessCleanup|TestWaitQueuedRemoteCommitBeforeWriteBlocksUntilQueuedCommitDone|TestWriteBackUploader_OnSuccessReceivesUnknownStreamRevision|TestRenamePendingNewCommit|TestSQLiteWALIndexSidecarFailsClosedWhenTransientOverlayMissing|TestTransientOverlayRootUsesCredentialScopedMountInstance -count=1`
- `bash -n` for `e2e/fuse-performance-baseline.sh`, `e2e/fuse-sqlite-correctness.sh`, `e2e/fuse-release-gate.sh`, and `scripts/archive-fuse-performance-metrics.sh`

I did not run the live FUSE/TiDB e2e in this pass.

qiffang:
Addressed the latest `mornyx` blocking review in head `31b1051`.

Fixes:
- `Fsync` no longer retains the same-path remote commit lock until `Release`. Interactive fsync staging now uses a local-durability staging helper that holds the path lock only while writing the shadow/pending snapshot; the existing `Flush` → `Release` queued-commit path still reserves the lock through close.
- ShadowSpill enqueue-failure fallback now applies deferred chmod after successful synchronous shadow upload before clearing/finalizing pending state. If the chmod fails, the pending mode/shadow state is preserved for retry instead of being silently cleared.

Regressions added:
- `TestFsyncInteractiveStageDoesNotRetainRemoteCommitPathLock`
- `TestReleaseShadowSpillFallbackAppliesPendingMode`
- `TestReleaseShadowSpillFallbackPreservesPendingModeOnChmodFailure`

Local validation:
- `/usr/local/go/bin/go test ./pkg/fuse -run 'TestFsyncInteractiveStageDoesNotRetainRemoteCommitPathLock|TestReleaseShadowSpillFallbackAppliesPendingMode|TestReleaseShadowSpillFallbackPreservesPendingModeOnChmodFailure|TestFsyncShadowSpillCommitRefreshesRevisionAndClearsPending' -count=1`
- `/usr/local/go/bin/go test ./pkg/fuse -count=1 -timeout=5m`
- `/usr/local/go/bin/go test ./cmd/drive9 ./cmd/drive9/cli ./pkg/fuse -count=1`
- `/usr/local/go/bin/go vet ./pkg/fuse ./cmd/drive9 ./cmd/drive9/cli`
- `/usr/local/go/bin/go test -race ./pkg/fuse -run 'TestFsyncInteractiveStageDoesNotRetainRemoteCommitPathLock|TestReleaseShadowSpillFallbackAppliesPendingMode|TestReleaseShadowSpillFallbackPreservesPendingModeOnChmodFailure|TestWaitQueuedRemoteCommitBeforeWriteBlocksUntilQueuedCommitDone|TestCommitQueueBeginInFlight' -count=1`
- `PATH="/usr/local/go/bin:$PATH" make lint` → `0 issues`
- `git diff --check`

I’ll watch the GitHub PR checks on `31b1051` and follow up if CI/e2e exposes anything else.

qiffang:
R5 review at head `31b1051046321860e0744061ab0751b3705a1688`: **GREEN**.

I rechecked the two latest blocking areas from the `mornyx` review and did not find remaining blockers:

- `Fsync` now uses `stageShadowForLocalDurabilityLocked`, so interactive fsync no longer stores `fh.RemoteCommitUnlock` until `Release`; the same-path remote commit lock is held only around the local shadow/pending staging window. The existing `Flush` -> `Release` queued path still uses `stageShadowForQueuedCommitLocked` and keeps the reservation where the staged shadow is meant to be uploaded on close.
- ShadowSpill enqueue-failure fallback now calls `applyPendingModeWithTimeoutLocked` after a successful sync shadow upload and before local cleanup/finalization. If chmod fails, the code preserves pending mode/shadow state instead of clearing it.

Validation I ran locally:

- `git diff --check 68ce029f889a1a6ac17b07fb9d6b5849ce39631b..HEAD`
- `go test ./pkg/fuse -run 'TestFsyncInteractiveStageDoesNotRetainRemoteCommitPathLock|TestReleaseShadowSpillFallbackAppliesPendingMode|TestReleaseShadowSpillFallbackPreservesPendingModeOnChmodFailure|TestFsyncShadowSpillCommitRefreshesRevisionAndClearsPending|TestWaitQueuedRemoteCommitBeforeWriteBlocksUntilQueuedCommitDone|TestCommitQueueBeginInFlight|TestCommitQueueCommitNowHoldsPathLockThroughSuccessCleanup' -count=1`
- `go test -race ./pkg/fuse -run 'TestFsyncInteractiveStageDoesNotRetainRemoteCommitPathLock|TestReleaseShadowSpillFallbackAppliesPendingMode|TestReleaseShadowSpillFallbackPreservesPendingModeOnChmodFailure|TestWaitQueuedRemoteCommitBeforeWriteBlocksUntilQueuedCommitDone|TestCommitQueueBeginInFlight' -count=1`
- `go test ./pkg/fuse -count=1 -timeout=5m`
- `go test ./cmd/drive9 ./cmd/drive9/cli ./pkg/fuse -count=1`
- `go vet ./pkg/fuse ./cmd/drive9 ./cmd/drive9/cli`
- `bash -n e2e/fuse-performance-baseline.sh e2e/fuse-sqlite-correctness.sh e2e/fuse-release-gate.sh scripts/archive-fuse-performance-metrics.sh`

GitHub checks on `31b1051` are also green: `code-ci`, `local-e2e`, and CodeRabbit.

Formal GitHub approval is unavailable from this account because GitHub treats it as the PR author, so this comment is the review gate.

mornyx (changes requested):
I re-reviewed the latest head `31b1051`. The previous two blockers are addressed in the direct sense: interactive `Fsync` no longer retains `RemoteCommitUnlock` until `Release`, and the ShadowSpill enqueue-failure fallback now applies deferred chmod before clearing local pending state. I still cannot approve because the new `Fsync` staging path reopens the mutable-shadow race in a different form.

Blocking issue:

`pkg/fuse/dat9fs.go:1447-1456`, `pkg/fuse/dat9fs.go:7640-7664`, `pkg/fuse/dat9fs.go:8003-8035`, `pkg/fuse/dat9fs.go:8128-8152`, `pkg/fuse/commit_queue.go:738-783` - interactive `Fsync` now stages local durability under a short-lived path lock, then returns while leaving `ShadowCommitReady` or `WriteBackSeq` for `Release` to enqueue later. That avoids the long-lived lock, but it does not give the queued commit a stable snapshot. A second same-path writer can open/write after `fsync()` returns and before the first handle releases; there is no queued/in-flight commit yet, so it can mutate the same active shadow via the ordinary shadow write path. The first handle then releases and enqueues a `CommitEntry` using the first handle metadata such as `Size` and `BaseRev`, while `CommitQueue.uploadEntry` reads the current path shadow. This can upload the second writer bytes with the first entry metadata; `onCommitQueueSuccess` then updates inode/cache using `entry.Size`, so remote bytes and local metadata can diverge. The new `TestFsyncInteractiveStageDoesNotRetainRemoteCommitPathLock` only verifies that the lock is released; it does not cover this snapshot isolation case.

Please make the `Fsync` staged data independent from later same-path writers before `Release` uses it for CommitQueue upload. Viable directions include giving the fsync-stage a per-entry shadow generation/pin or copy that CommitQueue uploads, retiring the active shadow so later writers get a new shadow, or not routing `Fsync`-staged snapshots through the path-shadow CommitQueue unless the snapshot is still owned.

Checked locally:
- `git diff --check 68ce029f889a1a6ac17b07fb9d6b5849ce39631b..HEAD`
- focused `pkg/fuse` tests for the latest fsync/chmod fallback and queue cases
- `go test ./pkg/fuse -count=1 -timeout=5m`
- `bash -n` for `e2e/fuse-performance-baseline.sh`, `e2e/fuse-sqlite-correctness.sh`, `e2e/fuse-release-gate.sh`, and `scripts/archive-fuse-performance-metrics.sh`

I did not run the live FUSE/TiDB e2e in this pass.

qiffang:
Fixed the `pullrequestreview-4445192941` blocker in head `4fe2bc7`.

What changed:
- Interactive `Fsync` now stages through the queued-commit path and enqueues the staged shadow commit before returning when `CommitQueue` is available.
- After `Fsync` returns, the same path is visible via `CommitQueue.HasPath`, so later same-path writers wait instead of mutating the active shadow before the fsync commit is handed off.
- If ShadowSpill enqueue fails, Fsync keeps the handle-owned path lock and marks the staged commit ready for the existing Release fallback, avoiding the unsafe gap.
- Added `ShadowCommitSeq` so Release only consumes staged ShadowSpill state if no later writes changed the dirty sequence; stale staged state is cleared and its held path lock released.
- Same open handle writes adopt the latest committed revision after waiting for a queued same-path commit, avoiding stale base revision after fsync handoff.
- Added regression `TestFsyncInteractiveStageEnqueuesBeforeReleasingPath`.

Validation:
- `/usr/local/go/bin/go test -v ./pkg/fuse -run 'TestFsyncInteractiveStageEnqueuesBeforeReleasingPath|TestReleaseShadowSpillFallbackAppliesPendingMode|TestReleaseShadowSpillFallbackPreservesPendingModeOnChmodFailure|TestFsync_NoDuplicateUpload|TestShadowSpill_WriteBackSeqStaysZero' -count=1 -timeout=3m`
- `/usr/local/go/bin/go test ./pkg/fuse -count=1 -timeout=5m`
- `/usr/local/go/bin/go test ./cmd/drive9 ./cmd/drive9/cli ./pkg/fuse -count=1 -timeout=5m`
- `/usr/local/go/bin/go vet ./pkg/fuse ./cmd/drive9 ./cmd/drive9/cli`
- `/usr/local/go/bin/go test -race ./pkg/fuse -run 'TestFsyncInteractiveStageEnqueuesBeforeReleasingPath|TestCommitQueueBeginInFlightSerializesSamePath|TestCommitQueueBeginInFlightPreservesSamePathFIFO|TestReleaseShadowSpillFallbackAppliesPendingMode|TestReleaseShadowSpillFallbackPreservesPendingModeOnChmodFailure' -count=1 -timeout=5m`
- `git diff --check`

qiffang:
R6 review at head `4fe2bc763b2ff12284dabeac1eb70f2307b58d42`: **REVISE**.

Blocking findings:

1. `pkg/fuse/dat9fs.go:1449` leaves `stageShadowForLocalDurabilityLocked` unused. GitHub `code-ci` run `27095437766` fails in `make lint` with `unused`, and I reproduced it locally with `PATH="/usr/local/go/bin:$PATH" make lint`. This PR cannot merge while CI is red.

2. `pkg/fuse/dat9fs.go:1461-1496`, `pkg/fuse/dat9fs.go:932-953`, `pkg/fuse/dat9fs.go:6978-7011`, `pkg/fuse/dat9fs.go:1407-1417` still have an open-handle data-loss path after the new queued `Fsync` handoff. `enqueueStagedShadowCommitLocked` enqueues the path shadow and clears dirty/seq state, but it leaves the still-open handle marked `ShadowReady` / `ShadowSpill`. When the queued commit succeeds, `CommitQueue` removes the path shadow; `onCommitQueueSuccess` only refreshes revision state. A later write on the same file handle waits for the queue, then `adoptCommittedRevisionLocked` sees `ShadowReady` and recreates/truncates a new shadow to the handle size. The subsequent write-through updates only the new extent, and the next `stageShadowLocked` takes the `fh.ShadowReady` truncate path instead of materializing the full dirty buffer. That can upload a sparse/zero-filled shadow plus the later write, losing bytes from the fsynced commit.

The new regression proves a second same-path writer blocks while the fsync commit is queued, but it does not cover continued writes on the same open fd after `fsync()` returns. Please add a regression for `write -> fsync -> queued commit completes -> write same fh -> commit/release` that asserts the final remote bytes preserve the fsynced content plus the later write, then adjust the handle/shadow transition accordingly.

Validation I ran:
- `git diff --check 68ce029f889a1a6ac17b07fb9d6b5849ce39631b..HEAD`
- `go test ./pkg/fuse -run 'TestFsyncInteractiveStageEnqueuesBeforeReleasingPath|TestReleaseShadowSpillFallbackAppliesPendingMode|TestReleaseShadowSpillFallbackPreservesPendingModeOnChmodFailure|TestFsync_NoDuplicateUpload|TestShadowSpill_WriteBackSeqStaysZero|TestCommitQueueBeginInFlightSerializesSamePath|TestCommitQueueBeginInFlightPreservesSamePathFIFO' -count=1 -timeout=5m`
- `PATH="/usr/local/go/bin:$PATH" make lint` (fails as above)

Formal GitHub approval is unavailable from this account because GitHub treats it as the PR author, so this comment is the review gate.

qiffang:
Fixed the R6 blockers in head `2e61349`.

What changed:
- Removed the now-unused `stageShadowForLocalDurabilityLocked`, which was causing `code-ci`/lint failure.
- Fixed the same-open-handle data-loss path after queued `Fsync`: when commit queue cleanup removes the active shadow, still-open handles now clear stale `ShadowReady` / `ShadowSpill` / staged-commit state and rebind their clean `WriteBuffer` to the committed remote view.
- Unknown multipart commit revisions now leave the handle as not-new with `BaseRev=0`, so the next same-handle write does not use stale create/CAS semantics.
- Added regression `TestFsyncQueuedCommitSameOpenHandleWritePreservesCommittedBytes`, covering `write -> fsync -> queued commit completes -> write same fh -> release` and asserting final remote bytes preserve the fsynced prefix.

Validation:
- `/usr/local/go/bin/go test -v ./pkg/fuse -run 'TestFsyncQueuedCommitSameOpenHandleWritePreservesCommittedBytes|TestFsyncInteractiveStageEnqueuesBeforeReleasingPath|TestReleaseShadowSpillFallbackAppliesPendingMode|TestReleaseShadowSpillFallbackPreservesPendingModeOnChmodFailure|TestFsync_NoDuplicateUpload|TestShadowSpill_WriteBackSeqStaysZero' -count=1 -timeout=3m`
- `/usr/local/go/bin/go test ./pkg/fuse -count=1 -timeout=5m`
- `/usr/local/go/bin/go test ./cmd/drive9 ./cmd/drive9/cli ./pkg/fuse -count=1 -timeout=5m`
- `/usr/local/go/bin/go vet ./pkg/fuse ./cmd/drive9 ./cmd/drive9/cli`
- `/usr/local/go/bin/go test -race ./pkg/fuse -run 'TestFsyncQueuedCommitSameOpenHandleWritePreservesCommittedBytes|TestFsyncInteractiveStageEnqueuesBeforeReleasingPath|TestCommitQueueBeginInFlightSerializesSamePath|TestCommitQueueBeginInFlightPreservesSamePathFIFO|TestReleaseShadowSpillFallbackAppliesPendingMode|TestReleaseShadowSpillFallbackPreservesPendingModeOnChmodFailure' -count=1 -timeout=5m`
- `PATH="/usr/local/go/bin:$PATH" make lint`
- `git diff --check`

qiffang:
R7 review at head `2e61349be363f45eb9d62d888eaa654092e048e3`: **REVISE**.

The R6 lint blocker is fixed: GitHub `code-ci` and `local-e2e` are green, and local `make lint` now passes. The same-fh regression also covers the explicit `queue.WaitPath()` case. I still found one blocking race in the new cleanup path.

Blocking issue:

`pkg/fuse/dat9fs.go:866-887`, `pkg/fuse/dat9fs.go:1048-1066`, `pkg/fuse/dat9fs.go:7068-7070` - `clearRemovedCommittedShadowLocked()` always calls `releaseHandleRemoteCommitPathLocked(fh)`. When a same open fd writes immediately after `fsync()` while the queued fsync commit is still pending, `Write()` takes `fh.Lock()`, calls `lockHandleRemoteCommitPathLocked(fh)`, and waits for `CommitQueue.WaitPath()`. Because the handle lock is held, `onCommitQueueCleanup()` can hit the `TryLock()` skip path and leave the stale `ShadowReady` state for `adoptCommittedRevisionLocked()` to clear inside that same `Write()` call. At that point `fh.RemoteCommitUnlock` is the lock acquired for the current write critical section, not an old retained fsync lock. The cleanup helper releases it before `Write()` mutates the dirty buffer, and the deferred unlock becomes a no-op. That lets another same-path writer/Release acquire the path commit lock while this write is still in progress, re-opening the same queued-commit/shadow mutation race this PR is trying to close.

The new `TestFsyncQueuedCommitSameOpenHandleWritePreservesCommittedBytes` does not cover this because it explicitly waits for `queue.WaitPath(filePath)` before the second write, so cleanup runs before the write acquires and owns the path lock. Please add a regression where the second same-fh write starts while the fsync commit is still queued/in-flight and assert the write-owned path lock is not released by stale-shadow cleanup before the write finishes.

Validation I ran:
- `git diff --check 68ce029f889a1a6ac17b07fb9d6b5849ce39631b..HEAD`
- `go test ./pkg/fuse -run 'TestFsyncQueuedCommitSameOpenHandleWritePreservesCommittedBytes|TestFsyncInteractiveStageEnqueuesBeforeReleasingPath|TestReleaseShadowSpillFallbackAppliesPendingMode|TestReleaseShadowSpillFallbackPreservesPendingModeOnChmodFailure' -count=1 -timeout=5m`
- `PATH="/usr/local/go/bin:$PATH" make lint`
- `go test ./pkg/fuse -count=1 -timeout=5m`
- `go vet ./pkg/fuse ./cmd/drive9 ./cmd/drive9/cli`

Formal GitHub approval is unavailable from this account because GitHub treats it as the PR author, so this comment is the review gate.

mornyx (changes requested):
I re-reviewed the latest head `2e61349`. The previous fsync snapshot issue is mostly addressed by enqueueing the staged shadow before `Fsync` returns, and the lint/cleanup changes look directionally right. I still cannot approve because the new open-handle cleanup path can release the path lock owned by the current write operation.

Blocking issue:

`pkg/fuse/dat9fs.go:866-887`, `pkg/fuse/dat9fs.go:1048-1066`, `pkg/fuse/dat9fs.go:7068-7070` - `clearRemovedCommittedShadowLocked` always calls `releaseHandleRemoteCommitPathLocked(fh)`. That is safe when the helper is clearing an old retained reservation, but the same helper is also called from `adoptCommittedRevisionLocked` inside `Write` after `Write` has just acquired its own path lock with `lockHandleRemoteCommitPathLocked`. A realistic interleaving is: `Fsync` enqueues a shadow commit; the same fd writes again while the commit is still queued/in-flight, so `Write` holds `fh.mu` while waiting in `lockWritableRemoteCommitPath`; `onCommitQueueCleanup` cannot `TryLock` that handle and skips it; once the queue finishes, `Write` acquires the path lock and then calls `adoptCommittedRevisionLocked`; because the shadow was removed and no pending local state remains, `clearRemovedCommittedShadowLocked` runs and releases `fh.RemoteCommitUnlock`, which is now the lock for the current write critical section. The deferred unlock at the end of `Write` becomes a no-op, so another same-path writer or Release can enter while this write is still mutating local state. This reopens the path-serialization gap this PR is trying to close.

Please separate the two meanings of `RemoteCommitUnlock`: retained staged-commit reservation versus current-operation path lock. For example, make the stale-shadow cleanup clear/rebind handle state without releasing `RemoteCommitUnlock` when called from `adoptCommittedRevisionLocked`, or add an explicit ownership flag/parameter so only the cleanup/release path for a retained reservation releases it. Also add a regression where the post-fsync same-fd write starts while the queued fsync commit is still pending/in-flight, rather than after an explicit `queue.WaitPath(filePath)`.

Checked locally:
- `git diff --check 68ce029f889a1a6ac17b07fb9d6b5849ce39631b..HEAD`
- focused `pkg/fuse` tests for fsync queue, same-open-handle write, shadow-spill chmod fallback, and commit queue FIFO
- `go test ./pkg/fuse -count=1 -timeout=5m`
- `bash -n` for `e2e/fuse-performance-baseline.sh`, `e2e/fuse-sqlite-correctness.sh`, `e2e/fuse-release-gate.sh`, and `scripts/archive-fuse-performance-metrics.sh`

I did not run the live FUSE/TiDB e2e in this pass.

qiffang:
Fixed the R7 blocker in head `5bcd18e`.

What changed:
- `clearRemovedCommittedShadowLocked` now only releases the handle-owned remote commit lock when called from queued-commit open-handle cleanup. When the same open handle adopts the committed revision inside the current `Write()` critical section, it clears stale shadow state without releasing that current write lock.
- Strengthened `TestFsyncQueuedCommitSameOpenHandleWritePreservesCommittedBytes` so the second same-fh `Write` starts while the queued fsync commit is still blocked inside upload-complete; the test asserts the write does not complete until the queued commit finishes, then verifies final remote bytes preserve the fsynced prefix.
- Added `TestClearRemovedCommittedShadowKeepsCurrentWriteLock` to directly guard the path-lock lifecycle: stale-shadow cleanup must not unblock a same-path waiter until the current write releases its lock.

Validation:
- `/usr/local/go/bin/go test -v ./pkg/fuse -run 'TestClearRemovedCommittedShadowKeepsCurrentWriteLock|TestFsyncQueuedCommitSameOpenHandleWritePreservesCommittedBytes|TestFsyncInteractiveStageEnqueuesBeforeReleasingPath|TestReleaseShadowSpillFallbackAppliesPendingMode|TestReleaseShadowSpillFallbackPreservesPendingModeOnChmodFailure|TestFsync_NoDuplicateUpload|TestShadowSpill_WriteBackSeqStaysZero' -count=1 -timeout=3m`
- `/usr/local/go/bin/go test ./pkg/fuse -count=1 -timeout=5m`
- `/usr/local/go/bin/go test ./cmd/drive9 ./cmd/drive9/cli ./pkg/fuse -count=1 -timeout=5m`
- `/usr/local/go/bin/go vet ./pkg/fuse ./cmd/drive9 ./cmd/drive9/cli`
- `/usr/local/go/bin/go test -race ./pkg/fuse -run 'TestClearRemovedCommittedShadowKeepsCurrentWriteLock|TestFsyncQueuedCommitSameOpenHandleWritePreservesCommittedBytes|TestFsyncInteractiveStageEnqueuesBeforeReleasingPath|TestCommitQueueBeginInFlightSerializesSamePath|TestCommitQueueBeginInFlightPreservesSamePathFIFO|TestReleaseShadowSpillFallbackAppliesPendingMode|TestReleaseShadowSpillFallbackPreservesPendingModeOnChmodFailure' -count=1 -timeout=5m`
- `PATH="/usr/local/go/bin:$PATH" make lint`
- `git diff --check`

qiffang:
R8 review at head `5bcd18e2d437f00cde7948433ba9cc5f7f2cf7c6`: **GREEN**.

No remaining blocker found in the R7 scope. The path-lock fix now distinguishes queued-commit open-handle cleanup from current-write adoption: `clearRemovedCommittedShadowLocked(..., false)` clears stale shadow state without releasing the path lock owned by the active `Write()`, while the queued cleanup path still releases the retained handle lock. The strengthened regression now starts the second same-fh `Write` while the fsync commit is still blocked in upload-complete, and `TestClearRemovedCommittedShadowKeepsCurrentWriteLock` directly guards that stale-shadow cleanup cannot unblock a same-path waiter before the current write releases its lock.

Validation I ran:
- `git diff --check 68ce029f889a1a6ac17b07fb9d6b5849ce39631b..HEAD`
- `go test ./pkg/fuse -run 'TestClearRemovedCommittedShadowKeepsCurrentWriteLock|TestFsyncQueuedCommitSameOpenHandleWritePreservesCommittedBytes|TestFsyncInteractiveStageEnqueuesBeforeReleasingPath|TestReleaseShadowSpillFallbackAppliesPendingMode|TestReleaseShadowSpillFallbackPreservesPendingModeOnChmodFailure' -count=1 -timeout=5m`
- `go test -race ./pkg/fuse -run 'TestClearRemovedCommittedShadowKeepsCurrentWriteLock|TestFsyncQueuedCommitSameOpenHandleWritePreservesCommittedBytes|TestFsyncInteractiveStageEnqueuesBeforeReleasingPath|TestCommitQueueBeginInFlightSerializesSamePath|TestCommitQueueBeginInFlightPreservesSamePathFIFO|TestReleaseShadowSpillFallbackAppliesPendingMode|TestReleaseShadowSpillFallbackPreservesPendingModeOnChmodFailure' -count=1 -timeout=5m`
- `go test ./pkg/fuse -count=1 -timeout=5m`
- `go vet ./pkg/fuse ./cmd/drive9 ./cmd/drive9/cli`
- `PATH="/usr/local/go/bin:$PATH" make lint`
- `bash -n e2e/fuse-performance-baseline.sh e2e/fuse-sqlite-correctness.sh e2e/fuse-release-gate.sh scripts/archive-fuse-performance-metrics.sh`

GitHub checks for this head are also green: `code-ci` success, `local-e2e` success, CodeRabbit success.

Formal GitHub approval is unavailable from this account because GitHub treats it as the PR author, so this comment is the review gate.

mornyx (approved):
I re-reviewed latest head `5bcd18e`. I did not find remaining blockers.

The R7 path-lock issue is addressed: `clearRemovedCommittedShadowForOpenHandles` calls `clearRemovedCommittedShadowLocked(..., true)` only from queued-commit cleanup, while `adoptCommittedRevisionLocked` calls it with `false`, so stale-shadow cleanup inside the current `Write` critical section no longer releases the write-owned path lock. The only current call sites match that split. The strengthened tests also cover the in-flight fsync commit case and direct lock lifecycle.

Checked locally:
- `git diff --check 68ce029f889a1a6ac17b07fb9d6b5849ce39631b..HEAD`
- focused `pkg/fuse` tests for current-write lock preservation, fsync queued commit, commit queue FIFO, and shadow-spill chmod fallback
- `go test ./pkg/fuse -count=1 -timeout=5m`
- `bash -n` for `e2e/fuse-performance-baseline.sh`, `e2e/fuse-sqlite-correctness.sh`, `e2e/fuse-release-gate.sh`, and `scripts/archive-fuse-performance-metrics.sh`

I did not run live FUSE/TiDB e2e in this pass. Approving from code review.

qiffang:
Implemented the final FUSE live-mount consistency fix at head `0232ab2` and reran the all-flags local e2e gate.\n\nValidation:\n- Run: https://github.com/mem9-ai/drive9/actions/runs/27104291992\n- Passed: release gate, performance artifact upload, Drive9 archive, FUSE concurrency stress, POSIX/fsx, and final Enforce smoke results.\n- Concurrency artifact check: mounted manifest and remote manifest match expected; reader-errors is empty.\n\nLocal validation before push: focused regressions, full `go test ./pkg/fuse`, changed-package tests, `go vet`, focused race, `make lint`, and `git diff --check`.

PR #509, opened 2026-06-07
feat: add pack/unpack archives

mornyx:
## Summary
- add `drive9 pack` / `drive9 unpack` for gzip tar archives of local overlay resources, stored as ordinary Drive9 files so backend storage continues to follow Drive9's normal DB/S3 policy
- add mount profile configuration with built-in `coding-agent` and `none` profiles, plus user-defined profiles from `~/.drive9/profiles/<profile-name>`
- make `coding-agent` the default FUSE mount profile when `--profile` is omitted; add `none` for no overlay and no pack behavior
- keep the built-in `coding-agent` profile overlay-focused but pack nothing by default; automatic pack/unpack only runs when profile `[pack]` paths or explicit pack paths are configured
- add `drive9 profile show [profile]` so users can inspect built-ins or custom profiles before mounting
- flatten generated local-root and default archive paths so they do not include the profile name; default archives now live under `/.drive9/packs/<label>-<hash>.tar.gz`
- update mount state to persist profile pack paths, allowing later `umount` and remount auto-unpack decisions to use the same configured paths
- update CLI smoke coverage to define an explicit pack profile for pack/unpack checks under the new default-no-pack semantics
- stabilize a flaky parallel disk-read cancellation test that was blocking `code-ci`

## Profile format
Custom profiles are plain text files under `~/.drive9/profiles/`.

```ini
[local]
.git
.cache

[remote]

[pack]
.git
dist
```

Lines before the first section are treated as `[local]`. Blank lines and `#` comments are ignored.

## Validation
- `env GOCACHE=/private/tmp/drive9-profile-config.V22hmu/.codex-gocache go test ./cmd/drive9/cli ./cmd/drive9 ./pkg/fuse ./pkg/mountstate`
- `env GOCACHE=/private/tmp/drive9-profile-config.V22hmu/.codex-gocache go test ./pkg/fuse -run TestDat9FSParallelDiskReadStopsQueuedBlocksAfterFirstError -count=20`
- `env GOCACHE=/private/tmp/drive9-profile-config.V22hmu/.codex-gocache GOLANGCI_LINT_CACHE=/private/tmp/drive9-profile-config.V22hmu/.codex-golangci-cache make lint`
- GitHub checks: `code-ci` success, `local-e2e` success, CodeRabbit success

## Release Notes

* **New Features**
  * Added `pack` and `unpack` commands for archiving and restoring overlay snapshots
  * Added overlay unpacking controls to mount: `--unpack` (repeatable) and `--no-auto-unpack`
  * Added overlay packing controls to unmount: `--pack` (repeatable), `--no-auto-pack`, and `--pack-path` options
  * Added `profile show` command to display profile configuration
  * Auto-generates local-root path for overlay mounts when omitted

* **Tests**
  * Added comprehensive test coverage for pack/unpack workflows
  * Added end-to-end smoke tests validating pack/unpack operations

* **Documentation**
  * Updated E2E documentation with pack/unpack operation coverage


qiffang (changes requested):
Verdict: REVISE. I independently reviewed the pack/unpack flow and the current blockers are real.

1. Unpack can write outside `localRoot/overlay` through symlinks. `safeOverlayTarget` only validates the lexical tar entry name (`cmd/drive9/cli/pack.go:800-805`), then `extractPackEntry` creates symlinks with unvalidated `hdr.Linkname` (`cmd/drive9/cli/pack.go:836-843`) and regular-file extraction later uses `MkdirAll`/`OpenFile` on paths that can traverse existing symlink ancestors (`cmd/drive9/cli/pack.go:816-827`). A tampered remote pack can include `entries/repo/link -> /tmp/outside` followed by `entries/repo/link/pwned`, or rely on an existing symlink under the overlay, and `drive9 unpack` writes outside the overlay root. This is a safety blocker for shared/remote archives. Reject absolute/escaping symlink targets and refuse extraction through symlink ancestors before mkdir/open/symlink.

2. `--replace` deletes the existing overlay before the archive has been fully validated. Once the manifest is decoded, `extractPackArchive` immediately calls `removeManifestPaths` (`cmd/drive9/cli/pack.go:386-399`) and only then streams the remaining tar entries (`cmd/drive9/cli/pack.go:402-407`). If the archive is truncated or contains a later malformed/unsupported entry, the user's existing `.git`/build cache roots are already removed and the restore is partial. Replacement must be staged and swapped only after the full archive is read and validated, with a regression for a corrupt archive preserving the old overlay.

3. Auto-pack on `drive9 umount` is not bound to the credentials used for the mount. `defaultPackAfterUnmount` re-resolves the current CLI context at unmount time (`cmd/drive9/cli/mount.go:777-787`) and always creates `client.New(server, apiKey)`, dropping delegated tokens entirely. If the mount used delegated auth, auto-pack becomes unauthenticated; if the active owner context changed after mount, it can upload the archive to the wrong tenant/account. The unmount pack client must be derived from mount-time auth material or an equivalent immutable credential snapshot, and tests need to cover delegated and switched-context cases.

4. Default coding-agent packs do not record deletions of default roots. `discoverCodingAgentPackSources` only appends existing default names while walking the overlay (`cmd/drive9/cli/pack.go:500-535`), and `removeManifestPaths` only removes paths listed in the manifest (`cmd/drive9/cli/pack.go:852-870`). Repro: archive A contains `/repo/dist`; user deletes `dist`; auto-pack archive B omits `/repo/dist`; next auto-unpack with replace never removes stale `dist`, so deleted build/cache state comes back. A pack snapshot needs tombstones or an explicit list of considered default roots so unpack can clear paths that disappeared.

Validation I ran: `git diff --check origin/main...HEAD` passed; `go test ./cmd/drive9/cli ./cmd/drive9 -count=1` passed; `bash -n e2e/cli-smoke-test.sh` passed. `go test ./pkg/client -run TestNonExistentPackCompile -count=1` is locally blocked by rootless Docker/testcontainers on this machine; CI covers that package.

cmd/drive9/cli/pack.go
```diff
@@ -0,0 +1,1278 @@
...
+
+func packSourcePaths(sources []packSource) []string {
+	out := make([]string, len(sources))
+	for i, source := range sources {
+		out[i] = source.ArchivePath
+	}
+	return out
+}
+
+func packReplacePaths(opts packOptions, paths []string) []string {
```
qiffang, line 830: This still does not tombstone a parent when all default coding-agent roots under that parent were deleted. In the common auto-pack path, `paths` is built only from currently existing `.git`/`dist`/`build`/`target` entries; if `/repo/.git` was deleted and no default siblings remain, `parents` stays empty and the whole-mount `LocalPrefix` path returns nil `ReplacePaths`. A later `unpack --replace` therefore cannot remove a stale `/repo/.git` from a previous restore. Please preserve the considered parent or an explicit tombstone even when none of its default roots currently exist, with a regression for the all-default-roots-deleted case.

qiffang (changes requested):
Verdict: REVISE on head `2339f0218f7f491bc8b5a3227f7c0325da91efe1`.

R1 blockers around symlink-safe unpack, staging before replace, and mount-time credential snapshots are substantially addressed. One correctness blocker remains:

1. Default coding-agent tombstones are still incomplete when all default roots under a parent were deleted. `discoverCodingAgentPackSources` only returns existing default dirs, and `packReplacePaths` derives its parent set only from `paths`; when `paths` is empty and the auto-pack local prefix is the whole mount (`/`), it returns nil. Repro: archive A contains `/repo/.git`; user later deletes `.git` and has no remaining `.git`/`dist`/`build`/`target` under `/repo`; auto-pack B writes an empty manifest with no `replace_paths`; next auto-unpack with `Replace=true` removes nothing, so a stale `/repo/.git` from a previous restore survives. That is still the deleted-default-root case the previous review blocked on. The fix needs to record the considered default-root parent even when none of that parent’s default dirs currently exist, or otherwise persist explicit tombstones for deleted default roots. Please add a regression where all default roots for a repo are deleted and a replace-unpack clears the previously restored root.

Validation I ran:
- `/usr/local/go/bin/go test ./cmd/drive9/cli -run 'TestPack|TestUnpack|TestRunUmount|TestExtractPackArchive|TestPackArchive|TestPackAuth|TestWriteProcessState|TestMountCmd' -count=1`
- `/usr/local/go/bin/go test ./pkg/mountstate -count=1`
- `/usr/local/go/bin/go vet ./cmd/drive9/cli ./pkg/mountstate`

`/usr/local/go/bin/go test ./pkg/client -run TestNonExistentPackCompile -count=1` is still locally blocked by rootless Docker/testcontainers on this machine.

qiffang (changes requested):
R3 review on current head `2339f0218f7f491bc8b5a3227f7c0325da91efe1`: still **REVISE**.

The current GitHub head is unchanged from the prior R2 review, and the remaining correctness blocker is still present.

Blocking finding:

1. `cmd/drive9/cli/pack.go:498` / `cmd/drive9/cli/pack.go:672`: default coding-agent pack tombstones are still incomplete when all default roots under a parent were deleted. `discoverCodingAgentPackSources()` only returns currently existing `.git`/`dist`/`build`/`target` paths. `packReplacePaths()` then derives parents only from those current `paths`; when the auto-pack is for the whole mount (`LocalPrefix == ""`, canonical `/`) and all default roots were deleted, it returns nil `replace_paths`. Repro: archive A restores `/repo/.git`; the user deletes `/repo/.git` and has no remaining default roots under `/repo`; auto-pack B writes an empty manifest with no `replace_paths`; the next `unpack --replace` removes nothing, so stale `/repo/.git` survives. This still violates replace semantics for deleted build/git cache roots.

Required fix: persist the considered default-root parent(s) even when none of their default roots currently exist, or emit explicit tombstones for deleted default roots. Add a regression for the all-default-roots-deleted case where replace-unpack clears a previously restored default root.

Validation run locally:

- `git diff --check 68ce029f889a1a6ac17b07fb9d6b5849ce39631b...HEAD` passed.
- `/usr/local/go/bin/go test ./cmd/drive9/cli -run 'TestPack|TestUnpack|TestRunUmount|TestExtractPackArchive|TestPackArchive|TestPackAuth|TestWriteProcessState|TestMountCmd' -count=1` passed.
- `/usr/local/go/bin/go test ./pkg/mountstate -count=1` passed.
- `/usr/local/go/bin/go vet ./cmd/drive9/cli ./pkg/mountstate` passed.
- `bash -n e2e/cli-smoke-test.sh` passed.

qiffang:
Implementation-side review on latest head `f8d9342`.

I rechecked the pack/unpack and mount integration paths after the tombstone fix:

- `pack` manifest records overlay/archive path metadata: `profile`, `remote_root`, `paths`, `replace_paths`, and per-entry `path`, `remote_path`, `type`, `mode`, `size`, `linkname`, `mtime`.
- Default coding-agent auto-pack now preserves `replace_paths` across an empty follow-up snapshot, so deleting all default roots (`.git`, `dist`, `build`, `target`) still removes stale restored roots on later `unpack --replace`.
- `mount --profile=coding-agent` auto-unpack remains synchronous before `mountFuse()`; missing default archive is skipped, corrupt archive fails mount, explicit `--unpack` failure fails mount.
- Unpack installation is staged before replacing the target overlay paths, so corrupt archive extraction does not first delete the existing overlay.
- Mount-time credential state is snapshotted through `mountstate.ProcessState` and reused by umount auto-pack.

Validation run locally:
- `go test ./cmd/drive9/cli -run 'Test(Pack|Unpack|Mount|Umount|Default|Coding|PackAfter|ProcessState)' -count=1 -timeout=5m`
- `go test ./cmd/drive9 ./cmd/drive9/cli ./pkg/mountstate -count=1 -timeout=5m`
- `go vet ./cmd/drive9 ./cmd/drive9/cli ./pkg/mountstate`
- `git diff --check`
- `go test ./pkg/fuse -count=1 -timeout=5m`
- `PATH="/usr/local/go/bin:$PATH" make lint`

No additional implementation blocker found from my side. One product caveat remains: there is still no machine-readable auto-unpack status; today the completion signal is synchronous control flow plus stderr (`drive9: unpacked ...`) before `drive9: mounted on ...`. If we want `drive9 mount status` to distinguish `missing/skipped` vs `restored`, that should be a small follow-up unless required for this PR.

qiffang (approved):
R4 GREEN.

I re-reviewed the latest head `f8d934204df7ca66d025c1c0b12683692d0f0eb3`.

The prior blocker is closed:
- Default coding-agent packs now read the previous archive manifest before writing a new default archive.
- When all default roots are deleted, the new archive can have `paths=[]` and `entries=[]` while still preserving the old `replace_paths` tombstones.
- `unpack --replace` then removes stale `.git` / `dist` / `build` / `target` from a restored local root without deleting ordinary source paths.
- The new regression `TestPackRemoteArchivePreservesReplacePathsWhenDefaultRootsAllDeleted` covers the exact previously-blocking sequence: pack with `.git`, delete all default roots, pack again, then unpack/replace and verify stale `.git` is removed while normal source remains.

I also rechecked the earlier hardening areas:
- unpack still stages extraction before install, so malformed archives do not replace existing overlay data early.
- unsafe archive paths, symlink ancestors, unsafe symlink targets, and symlink overlay roots are rejected.
- `mount --profile=coding-agent` auto-unpack is synchronous before `mountFuse()`; explicit `--unpack` remains FUSE + coding-agent only.
- umount auto-pack uses mount-time process state and credential snapshot; process state with credentials is written `0600`.

Validation run locally:
- `/usr/local/go/bin/go test ./cmd/drive9/cli -run 'Pack|Unpack|Umount|MountCmd' -count=1`
- `/usr/local/go/bin/go test ./pkg/mountstate -count=1`
- `/usr/local/go/bin/go vet ./cmd/drive9/cli ./pkg/mountstate`
- `/usr/local/go/bin/go test ./cmd/drive9 ./cmd/drive9/cli ./pkg/mountstate -count=1`
- `/usr/local/go/bin/go test ./pkg/fuse -run 'Mount|ProcessState|CodingAgent|Local' -count=1`
- `PATH="/usr/local/go/bin:$PATH" make lint`

Non-blocking caveat: auto-unpack completion is still only observable through synchronous control flow and stderr output (`drive9: unpacked ...`). There is no machine-readable mount status field distinguishing missing/skipped vs restored. I would keep that as follow-up unless product explicitly requires structured restore status in this PR.

cmd/drive9/cli/pack.go
```diff
@@ -0,0 +1,1445 @@
...
+		if strings.TrimSpace(*localRoot) == "" {
+			*localRoot = resolved.LocalRoot
+		}
+		if !flagProvided(fs, "remote-root") {
+			*remoteRoot = resolved.RemoteRoot
+		}
+		if strings.TrimSpace(*profile) == "" {
+			*profile = resolved.Profile
+		}
+		profilePackPaths := append([]string(nil), resolved.PackPaths...)
```
qiffang, line 146: Blocking: `drive9 pack --mount ... --profile <override>` still uses `resolved.PackPaths` from the mounted profile. That makes the explicit profile override only affect metadata/default-archive resolution, not source discovery. A concrete bad case is a mount recorded with profile `with-pack` (`PackPaths=[.git]`) and a user running `drive9 pack --mount <path> --profile none :/archive.tar.gz`: this still packs `.git` because `ProfilePackPaths` stays copied from the mount state, while `Profile` is now `none` so `packReplacePaths` omits tombstones. Similarly a custom override profile with different `[pack]` paths packs the old profile's roots. Recompute/load `ProfilePackPaths` from the final selected profile whenever `--profile` is provided, and add regressions for `--profile=none` plus a custom override profile.

cmd/drive9/cli/pack.go
```diff
@@ -0,0 +1,1445 @@
...
+}
+
+func clientForPackArchiveArg(defaultClient *client.Client, raw string, remoteRoot string, profile string) (*client.Client, string, error) {
+	if strings.TrimSpace(raw) != "" {
+		return clientForRemoteArchiveArg(defaultClient, raw)
+	}
+	if strings.TrimSpace(profile) == noneMountProfile {
+		return nil, "", fmt.Errorf("default pack archive requires an overlay profile or an explicit remote archive path")
+	}
+	archivePath, err := defaultPackArchivePath(remoteRoot)
```
qiffang: Blocking: default pack archive paths are still keyed only by `remoteRoot`; the new profile-specific `[pack]` configuration makes that unsafe. Two profiles on the same remote root can now overwrite and auto-unpack each other's default archive. Worse, `readExistingPackReplacePaths` does not validate the prior manifest's `profile`, so profile B can preserve profile A's `replace_paths` and later delete/restore roots outside B's intended pack scope. The default archive must be namespaced by the effective profile, or auto-unpack/repack must fail/skip on manifest profile mismatch. Add a regression with two profiles using the same `remoteRoot` but different `[pack]` paths.

qiffang (changes requested):
R6 REVISE on head `977e81d8f0521ad4b23c56f388cf8367c943f0be`.

The previous tombstone blocker remains closed, and CI is green, but the new configurable profile layer introduces two correctness blockers around profile isolation/override. Existing tests do not cover these paths.

Local validation run:
- `go test ./cmd/drive9/cli ./pkg/mountstate -count=1 -timeout=5m`
- `go vet ./cmd/drive9/cli ./pkg/mountstate`

qiffang:
Addressed the R6 blockers in `164daf7`.

Changes:
- `drive9 pack --mount ... --profile <override>` now reloads the effective CLI profile when `--profile` is explicitly provided, and uses that profile's `[pack]` paths instead of the mount-state `PackPaths`.
- `--profile=none` no longer inherits mount-state pack paths; with no explicit path args it fails closed with the existing missing `[pack] paths` error instead of uploading the mounted profile roots.
- Default pack archives are now profile-scoped: `/.drive9/packs/<profile>/<remote-label>-<hash>.tar.gz`, with the hash including both profile and remote root. Mount auto-unpack and umount auto-pack use the same scoped path, so different profiles on the same remote root cannot overwrite or inherit each other's `replace_paths`.
- Added regressions for custom profile override, `--profile=none`, profile-scoped archive paths, and updated auto-pack/auto-unpack expectations.

Validation:
- `go test ./cmd/drive9/cli -run 'Test(PackMountProfileOverrideUsesEffectiveProfilePackPaths|PackMountProfileNoneDoesNotUseMountPackPaths|DefaultPackArchivePath|RunUmountPacksAfterUnmount|RunUmountAutoPacksCodingAgentMount|MountCmdAutoUnpacksCodingAgentPackBeforeFuseMount|PackRemoteArchivePreservesReplacePathsWhenDefaultRootsAllDeleted)' -count=1 -timeout=2m`
- `go test ./cmd/drive9/cli ./pkg/mountstate -count=1 -timeout=3m`
- `go vet ./cmd/drive9/cli ./pkg/mountstate`
- `git diff --check`
- `PATH="/usr/local/go/bin:$PATH" make lint`

cmd/drive9/cli/pack.go
```diff
@@ -0,0 +1,1467 @@
...
+		return "", fmt.Errorf("drive9 pack: %w", err)
+	}
+	profile = strings.TrimSpace(profile)
+	if profile == "" {
+		profile = defaultMountProfile
+	}
+	if err := validateProfileName(profile); err != nil {
+		return "", err
+	}
+	sum := sha256.Sum256([]byte(remoteRoot))
```
qiffang: Blocking: `profile` is validated but still does not participate in the default archive key; the hash is only `sha256(remoteRoot)`, and the filename is still `/.drive9/packs/<label>-<hash>.tar.gz`. The new test explicitly asserts `defaultPackArchivePath("/remote/root", "with-pack") == defaultPackArchivePath("/remote/root", "coding-agent")`, which preserves the cross-profile collision that R6 blocked. With two profiles on the same remote root, profile B can overwrite profile A's default archive, and `readExistingPackReplacePaths` will also inherit A's `replace_paths` into B because it does not verify `manifest.Profile`. Please make the primary default archive identity profile-scoped (profile in path or hash; migration fallback for old flat coding-agent path is fine), or otherwise fail/skip profile-mismatched manifests and prevent cross-profile overwrite/reuse. Add a regression proving two profiles with the same `remoteRoot` cannot share an archive path or inherited tombstones.

qiffang (changes requested):
R7 REVISE on head `c7dfd64390d53c6477db4a7d2fa95cd4a19c6c3a`.

The `pack --mount --profile <override>` source-discovery blocker is fixed by reloading the effective profile and adding `--profile=none`/custom override regressions. However, the default archive isolation blocker is still open after `c7dfd64`: the path computation validates the profile but does not include it in the archive key, and `readExistingPackReplacePaths` still reuses any prior manifest's `replace_paths` regardless of profile.

Validation run locally:
- `go test ./cmd/drive9/cli -run 'Test(DefaultPackArchivePath|PackMountProfileOverrideUsesEffectiveProfilePackPaths|PackMountProfileNoneDoesNotUseMountPackPaths|PackRemoteArchivePreservesReplacePathsWhenDefaultRootsAllDeleted)' -count=1 -timeout=2m`
- `go test ./cmd/drive9/cli ./pkg/mountstate -count=1 -timeout=5m`
- `go test ./pkg/fuse -run 'Mount|ProcessState|CodingAgent|Local' -count=1 -timeout=5m`
- `go vet ./cmd/drive9/cli ./pkg/mountstate`
- `PATH="/usr/local/go/bin:$PATH" make lint`

cmd/drive9/cli/pack.go
qiffang: Addressed in `e38460d` + follow-up `db1314e`.

Default archive paths stay flat under `/.drive9/packs/`, but the path hash now includes `profile + "\0" + remoteRoot`, so `coding-agent` and a custom profile on the same remote root no longer collide. I also updated `TestDefaultPackArchivePath` and added `TestDefaultPackArchivePathSeparatesProfilesForSameRemoteRoot`, which packs two profiles with different `[pack]` roots and verifies they produce two separate archives/manifests instead of overwriting or inheriting each other’s `replace_paths`.

The first pushed fix exposed one e2e expectation still using the old remote-root-only hash; `db1314e` updates `e2e/cli-smoke-test.sh` to compute the profile-aware default archive path.

Validation run locally on the final head:
- `/usr/local/go/bin/go test ./cmd/drive9/cli -run 'Test(DefaultPackArchivePath|DefaultPackArchivePathSeparatesProfilesForSameRemoteRoot|PackMountProfileOverrideUsesEffectiveProfilePackPaths|PackMountProfileNoneDoesNotUseMountPackPaths|RunUmountAutoPacksCodingAgentMount|MountCmdAutoUnpacksCodingAgentPackBeforeFuseMount|PackRemoteArchivePreservesReplacePathsWhenDefaultRootsAllDeleted)' -count=1 -timeout=2m`
- `/usr/local/go/bin/go test ./cmd/drive9/cli ./pkg/mountstate -count=1 -timeout=3m`
- `/usr/local/go/bin/go vet ./cmd/drive9/cli ./pkg/mountstate`
- `git diff --check`
- `PATH="/usr/local/go/bin:$PATH" make lint`

GitHub checks for `db1314e` are green: `ci` and `drive9-server-local e2e smoke` both passed.

qiffang (approved):
R8 GREEN on head `db1314eb072374aa218ddd6b8d44430a5f6ad576`.

The remaining R7 blocker is closed: default archive identity is now profile-isolated while keeping the flat `/.drive9/packs/<label>-<hash>.tar.gz` shape by hashing `profile + "\0" + remoteRoot`. Mount auto-unpack, umount auto-pack, and CLI smoke now use the same profile-aware default path. The new collision regression proves `coding-agent` and `with-pack` on the same `remoteRoot` write distinct default archives with distinct manifests, so cross-profile overwrite/tombstone inheritance is closed. The earlier `pack --mount --profile <override>` source-discovery fix remains valid.

Validation run locally:
- `go test ./cmd/drive9/cli -run 'Test(DefaultPackArchivePath|PackMountProfileOverrideUsesEffectiveProfilePackPaths|PackMountProfileNoneDoesNotUseMountPackPaths|PackRemoteArchivePreservesReplacePathsWhenDefaultRootsAllDeleted|MountCmdAutoUnpacksCodingAgentPackBeforeFuseMount|RunUmountAutoPacksCodingAgentMount)' -count=1 -timeout=2m`
- `go test ./cmd/drive9/cli ./pkg/mountstate -count=1 -timeout=5m`
- `go test ./pkg/fuse -run 'Mount|ProcessState|CodingAgent|Local' -count=1 -timeout=5m`
- `go vet ./cmd/drive9/cli ./pkg/mountstate`
- `PATH="/usr/local/go/bin:$PATH" make lint`
- `git diff --check c7dfd64..HEAD`

GitHub `ci`, `drive9-server-local e2e smoke`, and CodeRabbit are green. Non-blocking: CodeRabbit still points out `t.Fatalf` inside several `httptest` handlers; that is test hygiene, not a pack/unpack correctness blocker for this PR.

PR #514, opened 2026-06-08
fix: Aliyun OSS/KMS support — TLS skip, RRSA credentials, s3client refactor, generated-column guard

srstack:
## Problem

Multiple issues blocked production use with Alibaba Cloud:

1. **KMS TLS failure** — `DRIVE9_ALIYUN_KMS_ENDPOINT` pointing to a VPC/dedicated-gateway fails with `x509: certificate signed by unknown authority` because VPC certs are issued by an Aliyun internal CA absent from the system trust store.

2. **S3 credential dispatch buried in AWS code** — Aliyun credential logic (RRSA, env fallback) was mixed into `aws.go`'s `staticCredentialsProvider`, with no clear separation between AWS and Aliyun paths.

3. **ACK RRSA not supported** — ACK pods inject `ALIBABA_CLOUD_ROLE_ARN` + OIDC token file but there was no code to exchange them for STS credentials.

4. **TiDB error 3105 on upload complete** — When `DRIVE9_DISABLE_AUTO_EMBEDDING=true` is set on a TiDB cluster whose `semantic` table has `embedding` / `description_embedding` as `GENERATED ALWAYS AS (EMBED_TEXT(...)) STORED` columns, upload-complete calls `UpdateSemanticTx` / `ClearFileEmbeddingStateTx` which issue `embedding = NULL` — TiDB rejects any explicit write to a generated column.

### `pkg/encrypt/aliyun_kms.go` + `factory.go`
- Skip TLS certificate verification automatically when a custom endpoint is configured. No new env var needed — presence of `DRIVE9_ALIYUN_KMS_ENDPOINT` is sufficient signal.

### `pkg/s3client/` — credential dispatch refactor
- **`New(ctx, cfg)`** — public dispatcher: inspects endpoint, routes to `newAliyun` or `newAWS`.
- **`newAWS`** — AWS-only: explicit static key or SDK default chain.
- **`newAliyun`** (in `aliyun.go`) — Aliyun credential priority: explicit key → ACK RRSA → `ALIBABA_CLOUD_*` env.
- **`buildS3Client`** — shared constructor (transport pool, SDK config, RoleARN).
- **RRSA provider** (`rrsaProvider`) — reads OIDC token file, calls `sts.AssumeRoleWithOIDC`, caches credentials with 5-minute early refresh.
- Callers (`tenant/pool.go`, `cmd/drive9-server-local`) updated to `s3client.New`.

### `pkg/datastore/semantic.go` + `file_tx.go`
- `UpdateSemanticTx`: when `disableAutoEmbedTextWrites=true`, only clear `embedding_revision` / `description_embedding_revision`; never touch the `GENERATED ALWAYS` columns.
- `ClearFileEmbeddingStateTx`: same guard — only clear revision marker when embedding is a generated column.

### `cmd/drive9-server/main.go`
- Wire `DisableDatabaseAutoEmbedding` from `DRIVE9_DISABLE_AUTO_EMBEDDING` env var into `PoolConfig`.

### Tests
- `pkg/s3client/aliyun_test.go` (new): `TestIsAliyunEndpoint`, `TestCredentialsForAliyunEnvFallback`, `TestCredentialsForAliyunRRSAPriority`, `TestRRSAProviderCacheHit`, `TestRRSAProviderTokenFileNotFound`.
- `pkg/s3client/aws_test.go`: replace `TestStaticCredentialsProviderFromAWSConfig` with `TestCredentialsForAliyunExplicitKey` / `TestCredentialsForAliyunNoCredentials`.

### DIFF
--- a/cmd/drive9-server-local/main.go
+++ b/cmd/drive9-server-local/main.go
@@ -176,7 +176,7 @@ func main() {
 	stepStart = time.Now()
 	switch s3cfg.Mode {
 	case "aws":
-		s3c, err = s3client.NewAWS(startupCtx, s3client.AWSConfig{
+		s3c, err = s3client.New(startupCtx, s3client.AWSConfig{
 			Region:          s3cfg.Region,
 			Bucket:          s3cfg.Bucket,
 			Prefix:          s3cfg.Prefix,
--- a/cmd/drive9-server/main.go
+++ b/cmd/drive9-server/main.go
@@ -387,6 +387,7 @@ environment:
   DRIVE9_ENCRYPT_KEY KMS key id or alias (required for kms), Aliyun KMS key id (required for aliyun_kms)
   DRIVE9_ALIYUN_KMS_ENDPOINT custom Aliyun KMS endpoint, e.g. a VPC endpoint (no https:// prefix)
                              note: aliyun_kms reads the Aliyun region from DRIVE9_S3_REGION
+                             note: TLS verification is skipped automatically when this is set
   DRIVE9_TOKEN_SIGNING_KEY  32-byte hex key for JWT API key signing
   DRIVE9_VAULT_MASTER_KEY   32-byte hex key for vault DEK wrapping (omit to disable vault)
   DRIVE9_MAX_UPLOAD_BYTES maximum allowed upload size in bytes (default: %d, minimum: 1048576)
--- a/cmd/drive9-server/main.go
+++ b/cmd/drive9-server/main.go
@@ -207,19 +207,20 @@ func main() {
 		}
 
 		pool = tenant.NewPool(tenant.PoolConfig{
-			S3Dir:              s3cfg.Dir,
-			PublicURL:          publicBaseURL(addr),
-			S3Bucket:           s3cfg.Bucket,
-			S3Region:           s3cfg.Region,
-			S3Prefix:           s3cfg.Prefix,
-			S3RoleARN:          s3cfg.RoleARN,
-			S3Endpoint:         s3cfg.Endpoint,
-			S3ForcePathStyle:   s3cfg.ForcePathStyle,
-			S3AccessKeyID:      s3cfg.AccessKeyID,
-			S3SecretAccessKey:  s3cfg.SecretAccessKey,
-			S3SessionToken:     s3cfg.SessionToken,
-			S3EncryptionPolicy: s3cfg.EncryptionPolicy,
-			BackendOptions:     backendOptions,
+			S3Dir:                        s3cfg.Dir,
+			PublicURL:                    publicBaseURL(addr),
+			S3Bucket:                     s3cfg.Bucket,
+			S3Region:                     s3cfg.Region,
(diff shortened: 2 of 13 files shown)


mornyx (changes requested):
### Review verdict: request changes

Focused tests passed locally:

`env GOCACHE=/Users/mornyx/Desktop/repos/github.com/mem9-ai/drive9/.codex-gocache go test ./pkg/s3client ./pkg/datastore`

Blocking issues:

1. `pkg/tenant/pool.go:400` / `pkg/server/semantic_worker.go:1259`: `DisableDatabaseAutoEmbedding` turns off `opts.DatabaseAutoEmbedding` when creating tenant backends, but `Pool.AutoSemanticTaskTypes()` and `taskTypesForProvider()` still route TiDB providers through pool auto task types. With `DRIVE9_DISABLE_AUTO_EMBEDDING=true`, those backends no longer use the TiDB-auto path and app-managed `embed` tasks can be enqueued instead, but the multi-tenant worker either drops TiDB tenants from the scan when only an embedder is configured, or becomes viable from pool auto task types and then skips the acquired target because `allowedTaskTypes` is empty. That leaves semantic tasks stuck and can make startup validation pass for a worker that will not actually process the disabled-auto tenants. Please make the disable flag part of the worker routing contract: `Pool.AutoSemanticTaskTypes()` should return nil in that mode, and TiDB providers should use app-managed task types when database auto-embedding is disabled.

2. `pkg/s3client/aliyun.go:80`: RRSA credential exchange hardcodes the STS region to `ap-southeast-1`. This PR is meant to support Aliyun OSS/ACK generally, but deployments using `DRIVE9_S3_REGION=cn-hangzhou` or any other region will still call the Singapore STS endpoint for `AssumeRoleWithOIDC`, which can fail or violate region/partition expectations before the OSS client ever gets credentials. Please derive the STS region from `AWSConfig.Region`, the OSS endpoint, or an Aliyun region env var, and add a regression test for non-`ap-southeast-1` regions.

mornyx (approved):
### Review verdict: approve

Re-reviewed latest head `75f3e0c`.

The previous blockers are addressed:
- RRSA STS region selection now comes from `AWSConfig.Region` / `ALIBABA_CLOUD_REGION_ID`, and missing region plus expiration parse failures are surfaced instead of silently using the wrong credential behavior.
- `DisableDatabaseAutoEmbedding` is now represented in semantic worker routing: pool auto task types are disabled, and TiDB tenants fall back to app-managed task types only when database auto embedding is explicitly disabled.

Focused validation passed:
`env GOCACHE=/Users/mornyx/Desktop/repos/github.com/mem9-ai/drive9/.codex-gocache go test ./pkg/s3client -count=1 -timeout=5m`

Non-blocking follow-up: consider also rejecting `ALIBABA_CLOUD_ACCESS_KEY_SECRET` without `ALIBABA_CLOUD_ACCESS_KEY_ID` so partial static credential configuration fails clearly at startup. I would not block this PR on that.

PR #525, opened 2026-06-08
Bound full smoke Git workspace workload

qiffang:
## Summary
- bound `Run full smoke-all suite` with a 35 minute step timeout
- keep Git workspace coverage in smoke-all but constrain it to the `drive9` repo and `agent_edit_add_commit` scenario
- document that the broader Git matrix remains the separate hard-fail step

## Validation
- `ruby -e 'require "yaml"; YAML.load_file(".github/workflows/local-e2e.yml")'`\n- `bash -n e2e/git-workspace-smoke-test.sh e2e/smoke-all.sh`\n- `git diff --check origin/main...HEAD`\n\nFollow-up for PR #519 full-gate run 27144169948 hanging in smoke-all after PR #519 merged.

## Summary by CodeRabbit

* **Chores**
  * Reduced e2e workflow timeouts, tightened workspace clone/ready bounds, and added explicit mount-ready timeout for faster, more reliable feedback.
* **Tests**
  * Improved mount startup/shutdown and restore orchestration with stricter failure handling and more deterministic durability checks.
* **Documentation**
  * Clarified README: smoke coverage is intentionally bounded to a specific repo/scenario while full feature-matrix runs remain separate.


qiffang:
Updated head `278d1de` after all-flags run `27147891365`.

Findings from `27147891365`:
- `FUSE performance compare` correctly hard-failed on a real metric regression: `small_file_write` ratio `0.652` (< 0.70 threshold). This confirms the fail-on-regression path is active.
- `smoke-all` Git workspace coverage no longer hung, but the mount readiness check was too short for background FUSE startup; the mount later succeeded and the scenario completed.
- `Git feature matrix` was still using coding-agent restore semantics without `--pack-path .git`, and the non-durable generated-file assertion used `ignored-build/`, which is Git-ignored but not a Drive9 coding-agent local-only path.

Fixes in `278d1de`:
- raise `MOUNT_READY_TIMEOUT_S` to `60s` for `smoke-all` and Git feature matrix CI steps;
- run Git feature matrix coding-agent mounts with `--pack-path .git` so `.git` checkpoint restore is intentionally exercised;
- use `build/cache.tmp` for the non-durable generated-file assertion because `build/**` is actually in the builtin coding-agent local-only policy.

Validation:
- workflow YAML parse
- `bash -n e2e/_feature-matrix-runner.sh e2e/git-feature-matrix.sh e2e/git-workspace-smoke-test.sh e2e/smoke-all.sh`

New all-flags run: <https://github.com/mem9-ai/drive9/actions/runs/27150267937>.

qiffang:
Follow-up after checking all-flags run 27150267937.

Findings:
- The run now reaches the intended hard-fail compare path, but it still had harness noise:
  - `smoke-all` failed `drive9 agent_edit_add_commit mount starts` because PR #520 made `drive9 mount` background by default while the e2e script already backgrounds the process and waits for the actual FUSE mount. The mount log shows only `drive9: mount running in background`, so readiness was waiting on the short parent instead of the FUSE worker.
  - Git feature matrix failed its setup with `flag provided but not defined: -pack-path`; `--pack-path` belongs to `drive9 umount`, not `drive9 mount`. The step was also incorrectly allowed to return success on setup mount failure because the main path used bare `return` after recording a FAIL row.

Fix in head `2565a89`:
- Force Git e2e-managed mounts to run with `drive9 mount --foreground` so the script-owned background PID is the real FUSE process.
- Move restore-suite `.git` packing to `drive9 umount --pack ... --pack-path .git`, then remount with explicit `--unpack ...`.
- Make feature-matrix setup mount failures return non-zero, so CI cannot report a green Git matrix while the report contains setup FAIL rows.

Validation before push:
- YAML parse for `.github/workflows/local-e2e.yml`
- `bash -n e2e/_feature-matrix-runner.sh e2e/git-feature-matrix.sh e2e/git-workspace-smoke-test.sh e2e/smoke-all.sh`
- checked remaining `--pack-path` usage is only on `umount`

Fresh runs:
- PR check run: https://github.com/mem9-ai/drive9/actions/runs/27152513744
- all-flags/manual gate: https://github.com/mem9-ai/drive9/actions/runs/27152525186

Expected final state: smoke-all and Git feature matrix should be clean; if the manual gate remains red only because `FUSE_PERF_COMPARE_FAIL_ON_REGRESSION=1` catches the known perf regression, that is the desired PR #519/PR #525 hard-fail behavior.

qiffang:
Follow-up for all-flags gate `27152525186`:

- The actual required workloads all completed successfully: FUSE release gate, performance compare, concurrency stress, POSIX/fsx, full `smoke-all`, and Git feature matrix.
- The job still failed in final `Enforce smoke results` because `Run Git feature matrix` has `continue-on-error: true`; GitHub kept the final step `conclusion=success`, but the workflow enforced the raw pre-continue `outcome=failure`.
- Pushed `17443c8` to use final step `conclusion` in the summary/enforce gate, while still preserving real failures in the step logs.

Validation: workflow YAML parse, expression audit, and `git diff --check origin/main...HEAD` passed.

Fresh runs:
- PR check: <https://github.com/mem9-ai/drive9/actions/runs/27154415146>
- all-flags manual gate: <https://github.com/mem9-ai/drive9/actions/runs/27154450746>

srstack (commented):
## Code Review: Bound full smoke Git workspace workload

### Overview
Fixes a CI hang introduced after #519 merged: the full smoke-all step was running an unbounded Git workspace workload. Changes add a 35-minute step timeout, pin Git workspace smoke to a single repo (drive9) and scenario set (agent_edit_add_commit,fast_worktree), tighten clone/mount-ready timeouts, and add --foreground to FUSE mount calls in the feature matrix and git workspace smoke scripts.

---

### Strengths

- Root cause addressed: scoping to one repo + bounded scenarios is the right surgical fix.
- 35-minute hard timeout prevents future CI stalls.
- return 1 fix in _feature-matrix-runner.sh: bare return in a function that uses set -e was masking mount failures with a success exit code. Now correctly propagates failure.
- --foreground on FUSE mounts: keeps the mount process as a child so its stderr lands in the log and process lifecycle is tied to the test. Correct pattern for CI.
- stop_mount shift fix: the old code didn't consume mount_point from $@ before passing remaining args to drive9 umount, so extra args like --pack would be passed in the wrong position. Now correct.

---

### Issues

**1. GIT_WORKSPACE_SCENARIOS includes fast_worktree which is not in the PR description**

The PR description says scope is pinned to agent_edit_add_commit, but the diff sets:

    GIT_WORKSPACE_SCENARIOS: "agent_edit_add_commit,fast_worktree"

If fast_worktree is a heavier scenario, it may partially defeat the bounding intent. Confirm it fits comfortably within the 35-minute budget alongside agent_edit_add_commit.

**2. outcome vs conclusion rename in check_required**

The variable is renamed from outcome to conclusion. GitHub Actions has different semantics for steps.X.outcome vs steps.X.conclusion (outcome reflects continue-on-error; conclusion does not). Confirm the callers are passing the correct expression -- if they still pass steps.X.outcome this rename is cosmetic but misleading.

**3. MOUNT_READY_TIMEOUT_S: "60" may be too tight on cold EC2 runners**

If FUSE mount startup takes more than 60 seconds on a cold runner, the smoke step will fail spuriously. Confirm this is based on observed startup times or bump to 90-120 seconds.

---

### Minor

- The restore_pack_archive variable and --pack/--unpack args in run_restore_suite are a functional change bundled into what is described as a CI timeout fix. These look correct but should be called out explicitly in the PR description.
- ignored-build -> build directory rename: same concern -- functional change bundled with CI fix.

---

### Verdict

The timeout and scoping fix is correct and necessary. Confirm fast_worktree scenario is intentional and fits the budget. The two bundled functional changes (pack/unpack in restore suite, directory rename) should be documented in the PR description.

PR #526, opened 2026-06-08
fix(schema): strip ADD_COLUMNAR_REPLICA_ON_DEMAND from FULLTEXT/VECTOR index repair SQL

srstack:
## Problem

When a tenant's `semantic` table is missing a FULLTEXT or VECTOR index, the schema repair uses the original ALTER TABLE statement as repair SQL. This statement includes `ADD_COLUMNAR_REPLICA_ON_DEMAND`:

```sql
ALTER TABLE semantic
    ADD FULLTEXT INDEX idx_semantic_fts_description(description)
    WITH PARSER MULTILINGUAL
    ADD_COLUMNAR_REPLICA_ON_DEMAND
```

On TiDB instances without columnar replica support, the entire ALTER TABLE fails. The error is silently swallowed by `isIgnorableOptionalSchemaError` because it matches the `fulltext` / `add_columnar_replica_on_demand` keyword checks. The index is never created, and repeated repair passes all fail identically. After 3 passes, final validation reports:

```
tidb schema contract mismatch for mode "auto-embedding":
semantic schema contract: missing idx_semantic_fts_description index
```

## Root Cause (2 issues)

1. **Repair SQL contains `ADD_COLUMNAR_REPLICA_ON_DEMAND`**: `parseAlterTableAddIndexStatement` returns the full ALTER TABLE statement as the repair SQL. The columnar replica clause is unrelated to index creation but causes the entire statement to fail on unsupported TiDB instances.

2. **Fallback parser misses FULLTEXT/VECTOR indexes**: `parseObservedTiDBIndexes` (fallback when `information_schema.statistics` doesn't return these index types) only handled plain INDEX/KEY/UNIQUE variants, not FULLTEXT INDEX/KEY or VECTOR INDEX definitions from SHOW CREATE TABLE output.

## Fix

- `stripColumnarReplicaOnDemand`: strips `ADD_COLUMNAR_REPLICA_ON_DEMAND` from FULLTEXT/VECTOR index repair SQL
- `parseObservedTiDBIndexes`: recognizes `FULLTEXT INDEX` / `FULLTEXT KEY` / `VECTOR INDEX` definitions
- `parseCreateIndexStatement` / `parseInlineIndexDefinition`: adds `CREATE FULLTEXT/VECTOR/SPATIAL INDEX` handling for robustness

## Summary by CodeRabbit

* **New Features**
  * Extended support for TiDB index types including FULLTEXT, VECTOR, and SPATIAL indexes in schema parsing and validation.
  * Improved schema normalization for accurate index comparison and repair operations.

* **Tests**
  * Added comprehensive test coverage for index type parsing and schema validation.

### DIFF
--- a/pkg/tenant/schema/tidb_auto.go
+++ b/pkg/tenant/schema/tidb_auto.go
@@ -1526,20 +1526,35 @@ func diffTiDBTable(ctx context.Context, db *sql.DB, table tidbTableSpec) ([]tidb
 }
 
 func loadObservedTiDBIndexes(ctx context.Context, db *sql.DB, tableName, createStmt string) (map[string]struct{}, bool) {
-	indexNames, err := loadTiDBIndexNames(ctx, db, tableName)
-	if err == nil {
-		return indexNames, true
+	indexNames, statErr := loadTiDBIndexNames(ctx, db, tableName)
+	if statErr != nil {
+		logger.Warn(ctx, "tenant_tidb_schema_load_index_metadata_failed",
+			zap.String("table", tableName),
+			zap.Error(statErr))
 	}
-	logger.Warn(ctx, "tenant_tidb_schema_load_index_metadata_failed",
-		zap.String("table", tableName),
-		zap.Error(err))
-	observedIndexes, ok := parseObservedTiDBIndexes(createStmt)
-	if ok {
-		return observedIndexes, true
+
+	merged := make(map[string]struct{})
+	if statErr == nil {
+		for name := range indexNames {
+			merged[name] = struct{}{}
+		}
 	}
-	logger.Warn(ctx, "tenant_tidb_schema_parse_show_create_indexes_failed",
-		zap.String("table", tableName))
-	return nil, false
+
+	if observedIndexes, ok := parseObservedTiDBIndexes(createStmt); ok {
+		for name := range observedIndexes {
+			if _, exists := merged[name]; !exists {
+				merged[name] = struct{}{}
+			}
+		}
+	} else if statErr != nil {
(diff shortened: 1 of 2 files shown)


srstack:
I did not identify any discrete correctness, security, performance, or maintainability issues that should block this patch. The changes consistently reject new root self-dentry writes while preserving root mkdir idempotency and hiding historical invalid root entries from listings.

mornyx (approved):
Reviewed both the diff and the surrounding schema repair flow.

No blocking issues from my side. The change covers the current runtime path: the existing optional FULLTEXT/VECTOR schema statements are ALTER TABLE ... ADD ... INDEX statements, so stripping ADD_COLUMNAR_REPLICA_ON_DEMAND from those repair SQLs should let the index repair execute on TiDB instances that do not support on-demand columnar replicas. Merging SHOW CREATE parsed indexes with information_schema results also looks correct for cases where TiDB does not expose FULLTEXT/VECTOR indexes through statistics.

One non-blocking follow-up to consider: isSafeAddIndexRepairSQL now accepts CREATE FULLTEXT/VECTOR/SPATIAL INDEX forms, but isFulltextOrVectorIndexRepairSQL still only recognizes ALTER TABLE ... ADD FULLTEXT/VECTOR INDEX. If a future schema spec starts producing optional FULLTEXT/VECTOR repairs in CREATE INDEX form, unsupported-feature errors would become hard repair failures instead of being skipped as optional. This does not affect the current schema because those optional indexes are ALTER TABLE statements, but aligning the classifier would make the new CREATE-form support fully consistent.

CI and drive9-server-local e2e smoke are passing. Approved.

PR #527, opened 2026-06-08
Fix fast git workspace clone and worktree e2e regressions

mornyx:
## Summary

This PR fixes the current Git workspace regressions found while re-running the live Git E2E suite from `main` on the EC2 POSIX test host.

The fixes cover:

- Keep blobless fast clone objectless by removing `git ls-tree -l` from tree manifest generation. The manifest now records unknown blob sizes instead of forcing Git to lazy-fetch blob objects just to compute sizes.
- Hide removed fast linked worktrees immediately in the current FUSE mount by writing a local deletion marker and filtering deleted workspace runtimes before registry lookups.
- Stabilize clean Git tree inode mtimes so Git does not report false dirty files after clone or commit.
- Run the Git workspace smoke mount in foreground mode because the script already backgrounds the process and waits on the process it owns.
- Restore linked blobless worktrees without requiring clean remote blob objects to exist locally. Restore now parses HEAD tree object metadata without batch-checking every clean blob object.

## Root Causes

1. `git ls-tree -r -t -l -z <commit>` requested blob sizes. In a blobless partial clone, that forced Git to lazy-fetch clean blob objects during manifest creation, turning fast clone back into a slow remote object walk.

2. Clean virtual Git workspace entries used a fresh timestamp on each lookup/getattr. Git interpreted the changing stat metadata as file modifications, so `git status --porcelain` could show false `M` entries even when `git diff` was empty.

3. `drive9 git worktree remove --fast` removed backend/local metadata, but the already-mounted FUSE process still had the linked workspace runtime in memory. The same mount could still expose stale linked worktree metadata until remount.

4. The smoke test launched `drive9 mount` in the background while the CLI default mount path already daemonized, so the script sometimes waited on the short-lived parent instead of the foreground FUSE process.

5. Linked blobless worktree restore called a blob object batch-check across the full HEAD tree. Clean blob objects are intentionally allowed to be missing locally in blobless mode, so restore could fail with `lazy fetching disabled` before the working tree became usable.

## Fix Details

- Added explicit tree listing args that do not include `-l`, plus tests proving blob sizes remain unknown and blob lookup is avoided.
- Added Git cache deletion marker helpers and FUSE-side filtering for locally deleted workspaces.
- Added stable clean-tree mtime selection based on workspace creation time, with a deterministic fallback.
- Updated the E2E Git workspace smoke mount command to pass `--foreground`.
- Changed HEAD tree restore metadata collection to use `git ls-tree` output only; file contents are read only for paths that must become dirty overlay entries.

## Validation

Local Go tests:

```bash
env GOCACHE=/Users/mornyx/Desktop/repos/github.com/mem9-ai/drive9/.codex-gocache go test ./cmd/drive9/cli
env GOCACHE=/Users/mornyx/Desktop/repos/github.com/mem9-ai/drive9/.codex-gocache go test ./pkg/gitcache
env GOCACHE=/Users/mornyx/Desktop/repos/github.com/mem9-ai/drive9/.codex-gocache go test ./pkg/fuse
```

Live EC2 targeted validation against the dev endpoint:

- Instance: `i-076768b42d1d64d14`
- SSM command: `0689a879-cb5e-4ec5-b2fa-89b3ec12b34a`
- Commit: `d90bc503bc7c4770bf88439238287dd9dcda572b`
- Endpoint: `http://k8s-dat9-dat9serv-d5e02e7d07-1645488597.ap-southeast-1.elb.amazonaws.com`

Results:

- `drive9 fast_worktree`: 35/35 passed, 0 failed, duration 50s.
- `kimi-cli agent_edit_add_commit`: 17/17 passed, 0 failed, duration 30s.
- The targeted smoke covered fast clone, fast worktree add/remove, edit, add, commit, remount/recovery, linked `.git` restore, and post-remount Git status/log checks.

Full default E2E was not rerun to completion because the failure was isolated earlier and the targeted runs cover the corrected paths without waiting for unrelated long-running cases.

## Summary by CodeRabbit

* **New Features**
  * Local "workspace deleted" markers added so removed workspaces are hidden until refreshed; markers are set/cleared during clone/worktree operations.
  * Git clean-tree inodes now use stable mtimes instead of current time; git file entries defer size (size = -1) to avoid expensive blob-size lookups.

* **Bug Fixes**
  * Worktree removal now more robustly marks and cleans local state and continues cleanup even when intermediate steps fail.

* **Tests**
  * Added tests for deletion-marker lifecycle, stable inode mtimes, git-tree size behavior, and restored-head blob handling.

### DIFF
--- a/cmd/drive9/cli/git.go
+++ b/cmd/drive9/cli/git.go
@@ -8,6 +8,7 @@ import (
 	"crypto/sha256"
 	"encoding/hex"
 	"encoding/json"
+	"errors"
 	"flag"
 	"fmt"
 	"io"
@@ -135,7 +136,7 @@ func gitClone(args []string) error {
 	if enrichErr != nil {
 		fmt.Fprintf(os.Stderr, "drive9: warning: could not enrich GitHub tree sizes: %v\n", enrichErr)
 		if unknown := unknownGitTreeFileSizeCount(nodes); unknown > 0 {
-			fmt.Fprintf(os.Stderr, "drive9: warning: %d git file sizes remain unknown; git status may need to read those blobs lazily\n", unknown)
+			fmt.Fprintf(os.Stderr, "drive9: warning: %d git file sizes remain unknown; stat metadata will stay unknown until file content is read\n", unknown)
 		}
 	} else {
 		nodes = enriched
@@ -162,6 +163,9 @@ func gitClone(args []string) error {
 	cancel()
 	if err != nil {
 		return fmt.Errorf("register git workspace: %w", err)
+	}
+	if err := clearLocalGitWorkspaceDeleted(cmdCtx, resolved, ws.WorkspaceID); err != nil {
+		return fmt.Errorf("clear stale local git workspace deletion marker: %w", err)
 	}
 	ctx, cancel = context.WithTimeout(context.Background(), gitWorkspaceAPITimeout)
 	if err := c.ReplaceGitTree(ctx, ws.WorkspaceID, client.GitTreeReplaceRequest{
@@ -331,7 +335,7 @@ func gitWorktreeAdd(args []string) error {
 	if enrichErr != nil {
 		fmt.Fprintf(os.Stderr, "drive9: warning: could not enrich GitHub tree sizes: %v\n", enrichErr)
 		if unknown := unknownGitTreeFileSizeCount(nodes); unknown > 0 {
-			fmt.Fprintf(os.Stderr, "drive9: warning: %d git file sizes remain unknown; git status may need to read those blobs lazily\n", unknown)
+			fmt.Fprintf(os.Stderr, "drive9: warning: %d git file sizes remain unknown; stat metadata will stay unknown until file content is read\n", unknown)
 		}
 	} else {
 		nodes = enriched
@@ -362,6 +366,9 @@ func gitWorktreeAdd(args []string) error {
 	cancel()
 	if err != nil {
 		return fmt.Errorf("register linked git workspace: %w", err)
+	}
+	if err := clearLocalGitWorkspaceDeleted(cmdCtx, worktreeResolved, ws.WorkspaceID); err != nil {
+		return fmt.Errorf("clear stale linked git workspace deletion marker: %w", err)
 	}
 	ctx, cancel = context.WithTimeout(context.Background(), gitWorkspaceAPITimeout)
 	if err := c.ReplaceGitTree(ctx, ws.WorkspaceID, client.GitTreeReplaceRequest{
@@ -504,16 +511,22 @@ func gitWorktreeRemove(args []string) error {
 		return fmt.Errorf("delete linked git workspace: %w", err)
 	}
 	cancel()
-	if err := markLocalGitWorkspaceDeleted(resolved, ws.WorkspaceID); err != nil {
-		return fmt.Errorf("mark linked git workspace deleted locally: %w", err)
+	var cleanupErrs []error
+	if err := markLocalGitWorkspaceDeleted(cmdCtx, resolved, ws.WorkspaceID); err != nil {
+		cleanupErrs = append(cleanupErrs, fmt.Errorf("mark linked git workspace deleted locally: %w", err))
 	}
 	if overlayRoot, err := localOverlayRootForMountedTarget(resolved); err == nil && overlayRoot != "" {
 		if err := os.RemoveAll(overlayRoot); err != nil {
-			return fmt.Errorf("remove linked local overlay root: %w", err)
+			cleanupErrs = append(cleanupErrs, fmt.Errorf("remove linked local overlay root: %w", err))
 		}
+	} else if err != nil {
+		cleanupErrs = append(cleanupErrs, fmt.Errorf("resolve linked local overlay root: %w", err))
 	}
 	if err := os.Remove(target); err != nil && !os.IsNotExist(err) {
 		fmt.Fprintf(os.Stderr, "drive9: warning: could not remove empty worktree root %s: %v\n", target, err)
+	}
+	if err := errors.Join(cleanupErrs...); err != nil {
+		return err
 	}
 	fmt.Fprintf(os.Stderr, "drive9: removed linked git workspace %s at :%s\n", ws.WorkspaceID, resolved.RemotePath)
 	return nil
@@ -791,7 +804,7 @@ func localOverlayRootForMountedTarget(resolved mountedGitTarget) (string, error)
 	return localPath, nil
 }
 
-func markLocalGitWorkspaceDeleted(resolved mountedGitTarget, workspaceID string) error {
+func markLocalGitWorkspaceDeleted(ctx context.Context, resolved mountedGitTarget, workspaceID string) error {
 	localRoot := strings.TrimSpace(resolved.LocalRoot)
 	if localRoot == "" || strings.TrimSpace(workspaceID) == "" {
 		return nil
@@ -799,7 +812,18 @@ func markLocalGitWorkspaceDeleted(resolved mountedGitTarget, workspaceID string)
 	if !filepath.IsAbs(localRoot) {
 		return fmt.Errorf("drive9 mount metadata local_root must be absolute, got %q", localRoot)
 	}
-	return gitcache.MarkWorkspaceDeleted(localRoot, workspaceID)
+	return gitcache.MarkWorkspaceDeleted(ctx, localRoot, workspaceID)
+}
+
+func clearLocalGitWorkspaceDeleted(ctx context.Context, resolved mountedGitTarget, workspaceID string) error {
+	localRoot := strings.TrimSpace(resolved.LocalRoot)
+	if localRoot == "" || strings.TrimSpace(workspaceID) == "" {
+		return nil
+	}
+	if !filepath.IsAbs(localRoot) {
+		return fmt.Errorf("drive9 mount metadata local_root must be absolute, got %q", localRoot)
+	}
+	return gitcache.ClearWorkspaceDeleted(ctx, localRoot, workspaceID)
 }
 
 func sameDrive9Mount(a, b mountedGitTarget) bool {
--- a/cmd/drive9/cli/git.go
+++ b/cmd/drive9/cli/git.go
@@ -164,9 +164,6 @@ func gitClone(args []string) error {
 	if err != nil {
 		return fmt.Errorf("register git workspace: %w", err)
 	}
-	if err := clearLocalGitWorkspaceDeleted(cmdCtx, resolved, ws.WorkspaceID); err != nil {
-		return fmt.Errorf("clear stale local git workspace deletion marker: %w", err)
-	}
 	ctx, cancel = context.WithTimeout(context.Background(), gitWorkspaceAPITimeout)
 	if err := c.ReplaceGitTree(ctx, ws.WorkspaceID, client.GitTreeReplaceRequest{
 		CommitSHA: head,
@@ -202,6 +199,9 @@ func gitClone(args []string) error {
 				fmt.Fprintf(os.Stderr, "drive9: warning: could not start background hydrate: %v\n", err)
 			}
 		}
+	}
+	if err := clearLocalGitWorkspaceDeleted(cmdCtx, resolved, ws.WorkspaceID); err != nil {
+		return fmt.Errorf("clear stale local git workspace deletion marker: %w", err)
 	}
 	return nil
 }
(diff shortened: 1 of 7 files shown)


qiffang (dismissed):
Unified REVISE with @dat9-dev2 for head `d90bc503bc7c4770bf88439238287dd9dcda572b`.

Findings:

1. `cmd/drive9/cli/git.go:349` / `pkg/gitcache/cache.go:100` - Deleted workspace markers are never cleared when the same worktree root is recreated. The server upsert path reuses the existing row's `workspace_id` for the same `root_path`, including a row previously marked `deleted`; the marker is keyed only by `workspaceID`. After `git worktree remove --fast` then `git worktree add --fast` at the same path in the same running mount, FUSE will keep filtering the new runtime as locally deleted.

2. `cmd/drive9/cli/git.go:502` - Local cleanup is short-circuited after a successful remote delete if `markLocalGitWorkspaceDeleted` fails. That leaves overlay/worktree local state orphaned even though the backend workspace was deleted. Marker failure should be reported, but cleanup should still continue and return/aggregate the error after local cleanup.

3. `pkg/fuse/git_workspace.go:236` / `pkg/fuse/git_workspace.go:287` / `pkg/fuse/git_workspace.go:317` - Hot lookup paths call `gitcache.WorkspaceDeleted` while holding `fs.git.mu`. That performs filesystem `os.Stat` under the workspace lock for each runtime candidate, adding disk I/O and contention to lookup/link-state/runtime-by-id paths. Cache the deleted state in loaded runtimes or move stat checks outside the lock.

4. `pkg/fuse/git_workspace.go:1951` / `pkg/fuse/git_workspace.go:1899` - `gitHeadTreeEntries` now records every HEAD blob as `size=-1`, but `buildRestoredGitHeadOverlay` still reads changed/added file and symlink blobs with `gitCatBlobNoLazy` and embeds them in overlay content. The previous `gitLocalObjectMaxBlobBytes` guard was removed, so a restored local HEAD with a large changed blob can allocate/upload arbitrarily large content instead of failing fast.

5. `cmd/drive9/cli/git.go:1026` / `pkg/fuse/git_workspace.go:1283` - Avoiding `ls-tree -l` moves the lazy-fetch problem into FUSE stat for non-GitHub repos or failed size enrichment. Those manifests store `SizeBytes=-1`; `resolveGitCleanNodeSize` then calls `git cat-file -s` without `GIT_NO_LAZY_FETCH`, so stat-heavy operations can fetch every clean blob in a blobless workspace. That violates the PR's performance goal unless unknown-size lookup is also non-lazy/fail-soft or size enrichment is guaranteed.

6. `pkg/fuse/git_workspace.go:886` - Clean-tree mtime uses only `workspace.CreatedAt`. When the same workspace root is refreshed to a new `HeadCommit` and a file keeps the same size, path, inode, size, and mtime can all stay unchanged; FUSE/kernel/tool caches can keep accepting stale clean content. The clean metadata must change with commit/tree identity or the refresh path must invalidate affected inodes/data.

Validation performed:

- GitHub `code-ci` and `local-e2e` are green for `d90bc503`.
- Local `PATH="/usr/local/go/bin:$PATH" go test ./cmd/drive9/cli ./pkg/gitcache ./pkg/fuse -count=1 -timeout=10m` passed.
- Local `PATH="/usr/local/go/bin:$PATH" go vet ./cmd/drive9/cli ./pkg/gitcache ./pkg/fuse` passed.
- Local `PATH="/usr/local/go/bin:$PATH" make lint` passed.
- Local `bash -n e2e/git-workspace-smoke-test.sh` passed.
- Local `git diff --check 03bec7ff71acf54118b801705ff5cfc19d08471c..HEAD` passed.

The blockers are correctness/performance semantics, not mechanical test failures.

mornyx:
@qiffang addressed the 6 review blockers in `dd1ff2d`.

Changes made:

1. Deleted workspace markers are now cleared when a workspace ID is reused by `drive9 git clone --fast` or `drive9 git worktree add --fast`. Added `gitcache.ClearWorkspaceDeleted(...)` and call it after workspace upsert succeeds.
2. `git worktree remove --fast` no longer short-circuits local cleanup after remote deletion if marker creation fails. Marker and overlay cleanup errors are collected and returned after cleanup is attempted.
3. FUSE no longer performs deletion-marker `os.Stat` while holding `fs.git.mu` in hot lookup paths. The runtime slice is copied under lock, then marker checks happen outside the critical section.
4. Restored local HEAD overlay now keeps the oversized blob guard, but only for changed/added HEAD blobs that actually need to become overlay entries. Clean unchanged blobs are still not batch-checked.
5. Unknown clean blob sizes no longer trigger lazy fetch during FUSE stat. `git cat-file -s` now runs with no-lazy env and fails soft, leaving unknown size until the file is actually read.
6. Clean-tree mtime is now stable per commit, not just per workspace row creation time. It derives from `CreatedAt` plus a hash of `HeadCommit`, so refreshing the same root to a new commit changes stat metadata.

Validation:

```bash
env GOCACHE=/Users/mornyx/Desktop/repos/github.com/mem9-ai/drive9/.codex-gocache go test ./pkg/gitcache -count=1
env GOCACHE=/Users/mornyx/Desktop/repos/github.com/mem9-ai/drive9/.codex-gocache go test ./cmd/drive9/cli -count=1
env GOCACHE=/Users/mornyx/Desktop/repos/github.com/mem9-ai/drive9/.codex-gocache go test ./pkg/fuse -count=1
env GOCACHE=/Users/mornyx/Desktop/repos/github.com/mem9-ai/drive9/.codex-gocache go vet ./cmd/drive9/cli ./pkg/gitcache ./pkg/fuse
bash -n e2e/git-workspace-smoke-test.sh
git diff --check
env PATH=/usr/local/go/bin:$PATH GOCACHE=/Users/mornyx/Desktop/repos/github.com/mem9-ai/drive9/.codex-gocache make lint
```

All passed. The first plain `make lint` attempt used the macOS default Go environment and failed at package loading (`no go files to analyze`); rerunning with the explicit Go/GOCACHE environment passed with `0 issues`.

cmd/drive9/cli/git.go
```diff
@@ -163,6 +164,9 @@ func gitClone(args []string) error {
 	if err != nil {
 		return fmt.Errorf("register git workspace: %w", err)
 	}
+	if err := clearLocalGitWorkspaceDeleted(cmdCtx, resolved, ws.WorkspaceID); err != nil {
```
mashenjun: This clears the local deleted-workspace marker before the new workspace is actually usable. `UpsertGitWorkspace` can reuse the deleted row's `workspace_id`, but the tree manifest and `.git` checkpoint are still registered later. If `ReplaceGitTree` or `uploadGitStateCheckpoint` fails after this point, the current mount stops filtering the old runtime (or starts seeing a live workspace row whose new `HeadCommit` has no tree), so a failed clone/worktree-add can expose stale or half-registered workspace state. Please keep the marker until the manifest/checkpoint steps have succeeded, then clear it as the last successful step. Same applies to `gitWorktreeAdd`.

pkg/fuse/git_workspace.go
```diff
@@ -1396,6 +1455,7 @@ func (fs *Dat9FS) gitCatFileBlobSize(ctx context.Context, rt *gitWorkspaceRuntim
 		fs.perf.gitCatFileCount.add(1)
 	}
 	cmd := exec.CommandContext(ctx, "git", "--git-dir", gitDir, "cat-file", "-s", objectSHA)
+	setGitNoLazyEnv(cmd)
```
mashenjun, line 1531: `GIT_NO_LAZY_FETCH` prevents the blob fetch, but this path is still reached from `gitTreeInode` -> `resolveGitCleanNodeSize` during lookup/readdir whenever the manifest has `SizeBytes == -1`. For blobless workspaces whose GitHub size enrichment fails, every `git status`/`ls -l` over clean files will restore `.git` if needed and fork `git cat-file -s` per file, then fail and repeat on the next stat because the size remains unknown. That moves the remote-walk regression into the mounted stat path as a process-spawn storm. After local tree/blob cache misses, fast-blobless should return the unknown size directly (or cache the missing-size result) and only resolve/update size when content is actually read/materialized.

mashenjun (changes requested):
Reviewed current head `dd1ff2d`. I found two remaining issues that can still regress the fast Git workspace paths: marker clearing can expose stale/partial workspace state after a failed add/clone, and unknown blob sizes still trigger per-file Git size probes in the mounted stat path for blobless fallbacks.

pkg/fuse/git_workspace.go
mornyx: @mashenjun Fixed in ca0e758. For fast-blobless workspaces, `resolveGitCleanNodeSize` now checks only the local tree/blob cache and returns unknown size when the cache misses, instead of running `git cat-file -s` from lookup/readdir/stat. Size still gets resolved and cached when content is actually read/materialized. Added `TestGitWorkspaceBloblessUnknownSizeLookupSkipsGitSizeProbe` to assert no git size probe is counted on the unknown-size lookup path.

cmd/drive9/cli/git.go
mornyx: @mashenjun Fixed in ca0e758. The local deleted-workspace marker is now cleared only after the workspace manifest, git state checkpoint, and requested hydrate path have completed successfully. If registration/checkpoint/sync hydrate fails, the marker remains in place so the current mount keeps filtering the old/partial runtime.

pkg/fuse/git_workspace.go
```diff
@@ -234,8 +243,13 @@ func (fs *Dat9FS) loadedGitWorkspaceForPath(localPath string) (*gitWorkspaceRunt
 	}
 	clean := path.Clean(localPath)
 	fs.git.mu.Lock()
-	defer fs.git.mu.Unlock()
-	for _, rt := range fs.git.workspaces {
+	workspaces := make([]*gitWorkspaceRuntime, len(fs.git.workspaces))
+	copy(workspaces, fs.git.workspaces)
+	fs.git.mu.Unlock()
+	for _, rt := range workspaces {
+		if fs.gitWorkspaceRuntimeDeletedLocally(context.Background(), rt) {
```
mashenjun: This dynamic marker check can still re-enable a stale runtime from the current FUSE process after the CLI successfully reuses the same `workspace_id` and clears the marker. If `fs.git.workspaces` still contains the old runtime and `ensureGitWorkspaces` returns early inside `gitWorkspaceRefreshInterval`, the next lookup sees the marker as gone and returns the previous `HeadCommit`/tree before a refresh rebuilds the slice. That makes immediate commands after a re-add able to read the old workspace metadata. Please keep runtimes that were hidden by a local deletion marker suppressed until a successful refresh replaces them, or otherwise invalidate/force-refresh the FUSE workspace cache when the marker is cleared.

cmd/drive9/cli/git.go
```diff
@@ -1009,6 +1047,10 @@ func gitListTree(ctx context.Context, repoDir, commitSHA string) ([]client.GitTr
 	return nodes, nil
 }
 
+func gitListTreeArgs(repoDir, commitSHA string) []string {
+	return []string{"-C", repoDir, "ls-tree", "-r", "-t", "-z", commitSHA}
```
mashenjun: Dropping `-l` here affects every fast clone/worktree, not only blobless mode. For a normal `--fast` clone of a non-GitHub repo, blobs are already local but `enrichGitTreeSizesFromGitHub` cannot fill sizes, so the backend manifest stores `SizeBytes == -1` for all files. FUSE then resolves sizes by forking `git cat-file -s` one file at a time during lookup/readdir and repeats after remount because the manifest remains unknown. Please make the tree listing mode-aware and keep the size-producing path for non-blobless clones/worktrees, while only blobless skips the size lookup.

mashenjun (changes requested):
Reviewed the latest `ca0e758`. The previous marker-ordering and blobless stat probe issues are addressed, but I found two remaining issues in the latest code path.

pkg/fuse/git_workspace.go
mornyx: @mashenjun Fixed in 34d2732. I added a local refresh marker that is written when the CLI clears a reused workspace deletion marker. FUSE now treats runtimes whose refresh marker is newer than their snapshot as invalid, forces a workspace refresh, and skips the stale runtime if refresh fails. Runtimes that actually observe a deletion marker are also marked locally deleted and stay suppressed until a successful refresh replaces the runtime slice. Added tests for both direct stale-runtime suppression and force-refresh behavior.

cmd/drive9/cli/git.go
mornyx: @mashenjun Fixed in 34d2732. `gitListTree` is now mode-aware: normal `--fast` clone/worktree uses `git ls-tree -l` so local full-object clones preserve blob sizes in the manifest, while `--fast --blobless` still omits `-l` to avoid lazy blob fetches. Updated CLI tests cover both argument forms and verify size behavior for both modes.

mashenjun (approved):
Reviewed latest `34d27328`. The two previous findings are addressed: `gitListTree` is now mode-aware so non-blobless fast clones keep size metadata, and reused workspace IDs now use refresh markers plus stale-runtime suppression before cached runtimes can be trusted again.

I did not find new blocking issues in this pass. Targeted local validation passed for `./cmd/drive9/cli`, `./pkg/gitcache`, and the Git workspace FUSE tests covering refresh invalidation, blobless unknown-size lookup, stable mtimes, and restored HEAD overlay limits. Full `./pkg/fuse` still fails locally at `TestDiskReadCacheStartupRecoveryAndTempCleanup`, and the same test fails on the PR base commit, so I’m not treating that as a PR #527 regression. GitHub Actions for this head show `code-ci` and `local-e2e` successful.

PR #529, opened 2026-06-09
Fix Git workspace local HEAD size resolution

mornyx:
## Summary

This PR fixes a Git workspace correctness issue where files could be reported as modified after a local commit even though their contents had not changed.

The root cause was the local HEAD tree path used after local Git commits. Drive9 intentionally lists local Git trees without `git ls-tree -l` to avoid triggering GitHub partial-clone lazy fetches. That keeps fast clone fast, but it also means blob sizes are unknown for the local HEAD manifest. In some paths those unknown sizes surfaced as zero-sized clean files, so Git compared the working tree against the index and reported false `modified` entries after commit/status operations.

## Fix

The fix keeps the no-`-l` behavior and fills local HEAD tree sizes without remote blob fetches:

- Reuse known clean blob sizes from the base workspace manifest by object SHA.
- Fall back to the base commit local clean blob cache when the local HEAD manifest references unchanged base blobs.
- Batch-check only remaining local-only objects with `git cat-file --batch-check` using the existing no-lazy-fetch path.
- Add base-commit blob-cache fallback for clean file reads and clean node size resolution when the active clean node belongs to a local commit.

This preserves the fast-clone design goal: no clean remote blob contents are uploaded to Drive9, and size repair does not reintroduce per-file GitHub lazy fetches.

## Tests

Local:

- `go test ./pkg/fuse -run "TestGit(LocalHead|Workspace)"`
- `go test ./pkg/fuse`
- `go test ./cmd/drive9/cli`
- `git diff --check`

EC2 live E2E against Aliyun Drive9 endpoint:

- Instance: `drive9-posix-e2e-20260603`
- Endpoint: `http://alb-b6w55jkdwh9a82dkrk.ap-southeast-1.alb.aliyuncsslbintl.com`
- Script: `e2e/git-workspace-smoke-test.sh`
- Repos: `drive9`, `kimi-cli`, `kimi-code`
- Scenarios: `agent_edit_add_commit`, `agent_patch_apply`, `sandbox_restore`, `fast_worktree`
- Result: `182/182 passed, 0 failed`
- Runtime: about 9m30s

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved file and symlink size resolution for locally restored Git workspace data with multi-level fallbacks so restoration proceeds even when size metadata is missing
  * Enhanced handling of repository HEAD drift with additional fallback lookups to reduce failed reads and materializations
  * Increased resilience so size-resolution errors are logged without aborting local tree restoration

* **Tests**
  * Added tests validating size-filling behavior and cached-base read semantics; tightened a CLI test to isolate its temp directory

### DIFF
--- a/cmd/drive9/cli/ctx_test.go
+++ b/cmd/drive9/cli/ctx_test.go
@@ -880,6 +880,11 @@ func TestCtxListDisplaysTenantID(t *testing.T) {
 // pins it.
 func TestF13_NoDaemonStateBetweenCalls(t *testing.T) {
 	home := withIsolatedHome(t)
+	tmpRoot := filepath.Join(home, "tmp")
+	if err := os.MkdirAll(tmpRoot, 0o700); err != nil {
+		t.Fatalf("mkdir temp root: %v", err)
+	}
+	t.Setenv("TMPDIR", tmpRoot)
 	// Snapshot pre-state of common leak sinks.
 	preEntries := func() []string {
 		var out []string
--- a/pkg/fuse/git_workspace.go
+++ b/pkg/fuse/git_workspace.go
@@ -776,6 +776,9 @@ func (fs *Dat9FS) ensureGitLocalHeadTree(ctx context.Context, rt *gitWorkspaceRu
 		return false, err
 	}
 	nodes, children := parseLocalGitTree(rt.workspace.WorkspaceID, head, treeOut)
+	if err := fs.fillLocalGitTreeSizes(ctx, rt, gitDir, nodes, children); err != nil {
+		log.Printf("git workspace local HEAD size fill failed workspace=%s head=%s: %v", rt.workspace.WorkspaceID, head, err)
+	}
 	rt.mu.Lock()
 	rt.localTreeCommit = head
 	rt.localNodes = nodes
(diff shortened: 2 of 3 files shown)


qiffang (commented):
Reviewed PR 529.

No blocking findings from my pass. The local HEAD size fill keeps the fast-clone/no `ls-tree -l` design intact: base manifest sizes are reused only by blob object id, base blob-cache fallback is also gated on the same object id, and remaining unknown local objects go through the existing no-lazy batch-check path. The read/size fallback for local HEAD nodes also only reuses base cache when the object SHA matches, so I do not see a stale-path content risk.

Validation I ran on PR head `e653ee2`:
- `go test ./pkg/fuse -run "TestGit(LocalHead|Workspace)"` passed.
- `go test ./cmd/drive9/cli -run TestF13_NoDaemonStateBetweenCalls` passed.
- `go test ./cmd/drive9/cli` passed.
- `git diff --check 6e2820e22d0592ece7644da1135ba9ef7ba95839..HEAD` passed.

`go test ./pkg/fuse` is locally blocked by `TestDiskReadCacheStartupRecoveryAndTempCleanup`, but the same test fails on base `6e2820e`, so I do not attribute that failure to this PR. GitHub checks `ci` and `drive9-server-local e2e smoke` are currently passing.

Only remaining nit I saw is the existing CodeRabbit inline comment about wrapping the two raw errors in `fillLocalGitTreeSizes`; I did not duplicate that inline.

qiffang (changes requested):
## adversary-2 R1 review — PR #529 (Fix Git workspace local HEAD size resolution)

**Verdict: REVISE** — 1 correctness blocker, 1 consistency bug.

### B1 (BLOCKER): `fillLocalGitTreeSizes` early-returns on single `gitcache.StatBlob` error, leaving remaining nodes unsized

In `fillLocalGitTreeSizes`, the loop over `nodes` calls `gitcache.StatBlob` for each unknown-size blob. If `StatBlob` returns an error for ANY single object, the function returns immediately:

```go
if size, hit, err := gitcache.StatBlob(ctx, localRoot, workspaceID, baseCommit, oid); err != nil {
    return err
}
```

This aborts size resolution for ALL remaining nodes — including:
- Nodes that haven't been iterated yet (could be resolved from `sizeByObject` or blob cache)
- Nodes that would have been resolved by the `gitObjectInfoBatch` fallback

Because Go map iteration is non-deterministic, the set of nodes that get sized vs. left at -1 is random on each call. A single corrupted/inaccessible cache entry makes the size fill non-reproducible.

The caller (`ensureGitLocalHeadTree`) already logs and continues on error, so the function is expected to be best-effort. The `StatBlob` error should be treated like a cache miss (log + continue), not an abort.

### B2 (BUG): `readGitCleanFile` and `resolveGitCleanNodeSize` base-commit fallbacks pass `n.ObjectSHA` without lowercase normalization

In `fillLocalGitTreeSizes`, object SHAs are normalized:
```go
oid := strings.ToLower(strings.TrimSpace(n.ObjectSHA))
```

But in the new `readGitCleanFile` fallback:
```go
data, hit, err = gitcache.ReadBlob(ctx, localRoot, rt.workspace.WorkspaceID, baseCommit, n.ObjectSHA, offset, size)
```

And in `resolveGitCleanNodeSize`:
```go
if size, hit, err := gitcache.StatBlob(ctx, localRoot, rt.workspace.WorkspaceID, baseCommit, n.ObjectSHA); ...
```

`n.ObjectSHA` is passed raw. If `gitcache` uses the SHA in path construction and git returns mixed-case SHAs (git allows uppercase hex), these lookups will miss cache entries written with lowercase keys. The normalization must be consistent across all call sites.

### O1 (observation): Missing negative test for StatBlob error mid-fill

No test covers the scenario where `gitcache.StatBlob` fails for one object while others are resolvable. This would catch B1 directly.

### Checklist
- [x] Edge cases: symlink nodes filtered correctly via `gitNodeHasBlobSize`
- [x] `setGitTreeNodeSize` guards `size < 0` and missing node
- [x] `knownBaseGitObjectSizes` returns a fresh map (no shared-state mutation)
- [x] Tests cover base manifest inheritance, blob cache fallback, batch local check, read fallback
- [ ] **B1**: Single cache error aborts all remaining size resolution → non-deterministic result
- [ ] **B2**: ObjectSHA normalization inconsistency across call sites
- [ ] **O1**: No negative test for StatBlob error

PR #532, opened 2026-06-09
Add tenant usage metrics dashboards

srstack:
## Summary
- replace the legacy dat9_* metrics namespace with drive9_* metrics
- add tenant-scoped usage metrics for requests, latency, in-flight requests, transport bytes, logical file IO, and quota snapshots
- add a Drive9 Tenant Usage Grafana dashboard and update existing dashboards to the new metric names
- tighten tenant metric label cardinality for token/upload/S3 paths

## Tests
- env GOCACHE=/tmp/drive9-go-build go test ./...
- find docs/grafana -name '*.json' -exec python3 -m json.tool {} /tmp/drive9-json-check.out \;
- git diff --check

## Summary by CodeRabbit

* **New Features**
  * Added a Tenant Usage dashboard for per-tenant request activity, IO, storage and quota ratios.
  * Expanded tenant-scoped metrics and recording (requests, durations, in-flight, HTTP/file bytes, storage/media gauges).

* **Documentation**
  * Reorganized Grafana docs to a usage-first workflow and added dashboard-embedded navigation links.
  * Updated many dashboards to reflect the new tenant-focused organization.

* **Breaking Changes**
  * Dashboards now expect metrics under the new drive9_* namespace — external dashboards/alerts require coordinated migration.

### DIFF
--- a/docs/grafana/README.md
+++ b/docs/grafana/README.md
@@ -1,32 +1,49 @@
 # Drive9 Grafana Dashboards
 
-This directory uses a three-bucket layout instead of a single mega dashboard.
+This directory uses a usage-first layout plus focused incident drill-downs.
 
-## 1. Overview dashboards
+## 1. Usage dashboard
+
+- `drive9-tenant-usage-dashboard.json`: first-stop dashboard for tenant-level product usage: active tenants, request frequency, in-flight requests, logical file reads/writes, HTTP transport bytes, storage/media quota state, latency, and non-OK rates by tenant/surface/action. Use this to answer `who is using Drive9, how much, and through which workflows?`
+
+Tenant usage metrics intentionally allow `tenant_id` as a Prometheus label, but keep high-cardinality values out of labels: no path, file ID, upload ID, API key ID, raw URL, user agent, or trace ID.
+
+## Tenant metric contract
+
+- `drive9_tenant_requests_total`: request count by `tenant_id`, `surface`, `action`, `result`, `status`, and `status_class`.
+- `drive9_tenant_request_duration_seconds`: request latency histogram with the same labels as `drive9_tenant_requests_total`.
+- `drive9_tenant_inflight_requests`: current in-flight request gauge by `tenant_id`, `surface`, and `action`.
+- `drive9_tenant_http_bytes_total`: HTTP transport bytes by `tenant_id`, `surface`, `action`, and `direction=request|response`.
+- `drive9_tenant_file_bytes_total`: logical file bytes by `tenant_id`, `surface`, `action`, and `direction=read|write`.
+- `drive9_tenant_storage_bytes`: opportunistically published quota storage gauge by `tenant_id` and `state=confirmed|reserved|limit`.
+- `drive9_tenant_media_files`: opportunistically published quota media-file gauge by `tenant_id` and `state=confirmed|limit`.
+
+## 2. Service overview dashboards
 
 - `drive9-service-health-dashboard.json`: first-stop dashboard for service incidents, module availability, HTTP health, service error ratio, and coarse DB pressure.
 - `drive9-async-fuse-dashboard.json`: first-stop dashboard for delayed background work and mount-path incidents, especially semantic worker backlog, image/audio extraction runtime, audio extract failures, and FUSE remote behavior.
 
-## 2. Request-path drill-down dashboards
+## 3. Request-path drill-down dashboards
 
 - `drive9-api-surface-dashboard.json`: non-FS and non-upload API request-path analysis for control, vault, SQL, events, and auxiliary HTTP surfaces.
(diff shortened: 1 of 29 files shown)


mornyx (changes requested):
I found a few blocking issues in the metrics rollout.

1. `pkg/metrics/operations.go` / `pkg/metrics/db.go`: this PR renames the exported Prometheus series from `dat9_*` to `drive9_*` without emitting compatibility aliases. The in-repo dashboards are updated, but any already-deployed dashboards, alerts, recording rules, or external PromQL consumers that still query `dat9_http_requests_total`, `dat9_service_operations_total`, `dat9_db_operations_total`, etc. will go dark immediately after deployment. Since this is a public metrics contract change, please either dual-emit the old series for a deprecation window or split/call out the rename as an explicit migration with the dependent alert/dashboard updates.

2. `pkg/server/server.go:285` + `pkg/server/instrumentation.go:551`: the multi-tenant local S3 handler mutates `r.URL.Path` before calling `localS3.Handler()`. After the handler returns, `observe` reclassifies the same request object in `recordTenantHTTPRequest`, so `/s3/{tenant}/upload/...` has become `/upload/...` and is recorded as `surface="other", action="put"` instead of `object_store/upload_part`; the file-byte accounting is skipped too because the object-store branch no longer matches. Please preserve the original request path/classification for post-request metrics, use the class already stored in `requestMetricState`, or pass a cloned request into the local S3 handler.

3. `pkg/server/server.go:1072`: large `/v1/fs` reads that return a presigned redirect only record the 302 HTTP response. In production the actual object download goes directly to S3, so `drive9_tenant_file_bytes_total` never records those logical read bytes. That makes the new tenant usage dashboard significantly undercount read-heavy tenants with S3-backed files. If the metric is meant to represent logical file IO as documented, this path should record `plan.Size` (or otherwise account for the intended object bytes) when the read plan succeeds.

I did not run tests locally. The GitHub checks were still pending when I reviewed.

srstack:
Addressed the review feedback in c314e94:

- Documented the intentional breaking dat9_* -> drive9_* metrics migration in docs/grafana/README.md instead of dual-emitting aliases, matching this PR metrics-contract rewrite.
- Preserved the original /s3/{tenant}/... request path by passing a cloned request to the multi-tenant local S3 handler, so post-request tenant metrics classify object-store upload/download paths correctly.
- Recorded logical read bytes for successful presigned /v1/fs reads using plan.Size.
- Stopped publishing quota limit gauges when the configured storage/media limit is not positive.

Validation:
- env GOCACHE=/tmp/drive9-go-build go test ./...
- Grafana JSON validation with python3 -m json.tool
- git diff --check

mornyx (approved):
Re-reviewed the latest commit c314e94.

The prior blockers look addressed:
- the `dat9_*` -> `drive9_*` metric namespace change is now called out as an explicit breaking migration in the Grafana README, and the in-repo dashboards have been updated to the new namespace;
- the multi-tenant local S3 handler now clones the request and rewrites only the sub-request URL, so the post-handler instrumentation still sees the original `/s3/{tenant}/...` path;
- presigned `/v1/fs` redirect reads now record logical file bytes with `plan.Size` before returning the redirect;
- quota limit gauges are only emitted when the corresponding configured limit is positive, so unlimited tenants no longer publish misleading zero limits.

I do not see remaining blocking issues from the updated diff. One non-blocking follow-up worth considering: `requestTenantID` still derives `tenant_id` from raw `/s3/{tenant}/...` paths before the backend validates that tenant. Since this is the local S3 path, I do not think it blocks this PR, but moving tenant labeling until after successful `LoadS3Backend` or using `unknown` for unvalidated paths would reduce the chance of arbitrary paths creating high-cardinality tenant labels.

GitHub CI and the drive9-server-local e2e smoke check are passing at review time; CodeRabbit is still pending. I did not run tests locally for this pass.

Approved.

srstack:
Addressed the latest review feedback in 11bc6b8:

- Stopped deriving tenant_id from arbitrary /s3/{tenant}/... URL input. Raw S3 paths now only map fixed local forms to local; multi-tenant S3 sets tenant scope only after LoadS3Backend succeeds.
- Reused requestMetricState classification for tenant request recording, so handler path rewrites cannot change post-request surface/action attribution.
- Removed logical file-byte accounting from object-store upload_part/get_object requests; object-store now records HTTP transport bytes only, while logical file writes are counted on upload completion and logical reads on /v1/fs read/redirect.
- Set tenant metric scope earlier for authenticated known-tenant failures, and wired /v1/status plus /v1/provision once tenant identity is known.
- Added tenant_id to the tenant request latency dashboard panel grouping/legend.

Validation:
- env GOCACHE=/tmp/drive9-go-build go test ./...
- Grafana JSON validation with python3 -m json.tool
- git diff --check

PR #534, opened 2026-06-09
Add persistent SSE event replay

srstack:
## Summary
- add tenant-local fs_events persistence and replay for SSE reconnects, with in-memory fanout as fallback
- publish chmod as an SSE mutation and bump file revision/mtime on chmod
- add SSE metrics using existing drive9_service_operations_total/duration and active connection gauge
- add Grafana dashboard coverage and update cache invalidation docs
- improve client SSE parsing for comments and multi-line data frames

## Schema
- added fs_events to TiDB auto/app and db9 tenant init schemas
- verified dump-init-sql for tidb_zero, tidb_cloud_starter, and db9 includes fs_events

## Tests
- make test
- make lint
- env GOCACHE=/tmp/drive9-go-build go test ./pkg/datastore ./pkg/client ./pkg/server ./pkg/backend ./pkg/tenant/schema
- env GOCACHE=/tmp/drive9-go-build go test ./pkg/server -run 'TestSSE|TestEventBus|TestHandleChmodPublishesPersistentSSEEvent'
- jq empty docs/grafana/drive9-event-stream-dashboard.json
- git diff --check

Note: alert rules are committed separately in the runbooks repo as requested.

## Summary by CodeRabbit

* **New Features**
  * Durable filesystem event log with persistent replay, pruning, and retention sweeper
  * chmod now emits change events and increments file revision
  * SSE publishes prefer durable persistence when available; active-connection and operation metrics added

* **Enhancements**
  * SSE client parsing accepts multiline data and ignores comment/metadata lines
  * Replay/reset behavior and structured reset reasons standardized
  * Durable replay continuity and stricter sequence validation

* **Documentation**
  * New Grafana dashboard for event-stream monitoring
  * Cache invalidation spec updated to include chmod

* **Tests**
  * Added tests for persistent replay, retention pruning, chmod publishing, event-log bounds/pruning, and SSE behavior

### DIFF
--- a/docs/grafana/README.md
+++ b/docs/grafana/README.md
@@ -33,6 +33,7 @@ Tenant usage metrics intentionally allow `tenant_id` as a Prometheus label, but 
 - `drive9-filesystem-path-dashboard.json`: filesystem-path analysis that owns `/v1/fs/*`, correlates HTTP traffic with `fs_*` backend work, and now includes explicit WebDAV behavior.
 - `drive9-upload-path-dashboard.json`: upload-path analysis that owns upload flows and correlates upload HTTP traffic, backend upload operations, and upload-related events.
 - `drive9-upload-s3-path-dashboard.json`: upload plus raw object-store analysis for incidents that have narrowed from upload symptoms into `s3client` throughput, errors, or latency.
+- `drive9-event-stream-dashboard.json`: SSE event-stream analysis for `/v1/events`, durable replay, reset reasons, persistence errors, and retention sweep behavior.
 
 ## 4. Data-store drill-down dashboards
 
--- a/docs/grafana/drive9-event-stream-dashboard.json
+++ b/docs/grafana/drive9-event-stream-dashboard.json
@@ -1,0 +1,347 @@
+{
+  "annotations": {
+    "list": [
+      {
+        "builtIn": 1,
+        "datasource": {
+          "type": "grafana",
+          "uid": "-- Grafana --"
+        },
+        "enable": true,
+        "hide": true,
+        "iconColor": "rgba(0, 211, 255, 1)",
+        "name": "Annotations & Alerts",
+        "target": {
+          "limit": 100,
+          "matchAny": false,
+          "tags": [],
+          "type": "dashboard"
+        },
+        "type": "dashboard"
+      }
+    ]
+  },
+  "editable": true,
+  "fiscalYearStartMonth": 0,
+  "graphTooltip": 0,
+  "id": null,
+  "links": [],
+  "liveNow": false,
+  "panels": [
+    {
+      "datasource": {
+        "type": "prometheus",
+        "uid": "${datasource}"
+      },
+      "fieldConfig": {
+        "defaults": {
+          "unit": "short"
+        },
+        "overrides": []
+      },
+      "gridPos": {
+        "h": 6,
+        "w": 8,
+        "x": 0,
+        "y": 0
+      },
+      "id": 1,
+      "options": {
+        "legend": {
+          "displayMode": "list",
+          "placement": "bottom",
+          "showLegend": true
+        },
+        "tooltip": {
+          "mode": "single",
+          "sort": "none"
+        }
+      },
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "${datasource}"
+          },
+          "expr": "drive9_service_gauge{component=\"sse\",name=\"active_connections\"}",
+          "legendFormat": "active",
+          "refId": "A"
+        }
+      ],
+      "title": "Active SSE Connections",
+      "type": "timeseries"
+    },
+    {
+      "datasource": {
+        "type": "prometheus",
+        "uid": "${datasource}"
+      },
+      "fieldConfig": {
+        "defaults": {
+          "unit": "ops"
+        },
+        "overrides": []
+      },
+      "gridPos": {
+        "h": 6,
+        "w": 8,
+        "x": 8,
+        "y": 0
+      },
+      "id": 2,
+      "options": {
+        "legend": {
+          "displayMode": "table",
+          "placement": "bottom",
+          "showLegend": true
+        },
+        "tooltip": {
+          "mode": "single",
+          "sort": "none"
+        }
+      },
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "${datasource}"
+          },
+          "expr": "sum by (operation, result) (rate(drive9_service_operations_total{component=\"sse\"}[$__rate_interval]))",
+          "legendFormat": "{{operation}} {{result}}",
+          "refId": "A"
+        }
+      ],
+      "title": "SSE Operation Rate",
+      "type": "timeseries"
+    },
+    {
+      "datasource": {
+        "type": "prometheus",
+        "uid": "${datasource}"
+      },
+      "fieldConfig": {
+        "defaults": {
+          "unit": "ops"
+        },
+        "overrides": []
+      },
+      "gridPos": {
+        "h": 6,
+        "w": 8,
+        "x": 16,
+        "y": 0
+      },
+      "id": 3,
+      "options": {
+        "legend": {
+          "displayMode": "table",
+          "placement": "bottom",
+          "showLegend": true
+        },
+        "tooltip": {
+          "mode": "single",
+          "sort": "none"
+        }
+      },
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "${datasource}"
+          },
+          "expr": "sum by (result) (rate(drive9_service_operations_total{component=\"sse\",operation=\"reset\"}[$__rate_interval]))",
+          "legendFormat": "{{result}}",
+          "refId": "A"
+        }
+      ],
+      "title": "SSE Resets by Reason",
+      "type": "timeseries"
+    },
+    {
+      "datasource": {
+        "type": "prometheus",
+        "uid": "${datasource}"
+      },
+      "fieldConfig": {
+        "defaults": {
+          "unit": "s"
+        },
+        "overrides": []
+      },
+      "gridPos": {
+        "h": 6,
+        "w": 8,
+        "x": 0,
+        "y": 6
+      },
+      "id": 4,
+      "options": {
+        "legend": {
+          "displayMode": "table",
+          "placement": "bottom",
+          "showLegend": true
+        },
+        "tooltip": {
+          "mode": "single",
+          "sort": "none"
+        }
+      },
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "${datasource}"
+          },
+          "expr": "histogram_quantile(0.95, sum(rate(drive9_service_operation_duration_seconds_bucket{component=\"sse\",operation=~\"persist|replay|retention_sweep\"}[$__rate_interval])) by (le, operation))",
+          "legendFormat": "p95 {{operation}}",
+          "refId": "A"
+        },
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "${datasource}"
+          },
+          "expr": "histogram_quantile(0.99, sum(rate(drive9_service_operation_duration_seconds_bucket{component=\"sse\",operation=~\"persist|replay|retention_sweep\"}[$__rate_interval])) by (le, operation))",
+          "legendFormat": "p99 {{operation}}",
+          "refId": "B"
+        }
+      ],
+      "title": "SSE Persistence and Replay Latency",
+      "type": "timeseries"
+    },
+    {
+      "datasource": {
+        "type": "prometheus",
+        "uid": "${datasource}"
+      },
+      "fieldConfig": {
+        "defaults": {
+          "unit": "short"
+        },
+        "overrides": []
+      },
+      "gridPos": {
+        "h": 6,
+        "w": 8,
+        "x": 8,
+        "y": 6
+      },
+      "id": 5,
+      "options": {
+        "legend": {
+          "displayMode": "table",
+          "placement": "bottom",
+          "showLegend": true
+        },
+        "tooltip": {
+          "mode": "single",
+          "sort": "none"
+        }
+      },
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "${datasource}"
+          },
+          "expr": "sum by (operation, result) (increase(drive9_service_operations_total{component=\"sse\",operation=~\"persist|replay|retention_sweep\"}[$__range]))",
+          "legendFormat": "{{operation}} {{result}}",
+          "refId": "A"
+        }
+      ],
+      "title": "Durable Event Log Operations",
+      "type": "timeseries"
+    },
+    {
+      "datasource": {
+        "type": "prometheus",
+        "uid": "${datasource}"
+      },
+      "fieldConfig": {
+        "defaults": {
+          "unit": "percentunit"
+        },
+        "overrides": []
+      },
+      "gridPos": {
+        "h": 6,
+        "w": 8,
+        "x": 16,
+        "y": 6
+      },
+      "id": 6,
+      "options": {
+        "legend": {
+          "displayMode": "table",
+          "placement": "bottom",
+          "showLegend": true
+        },
+        "tooltip": {
+          "mode": "single",
+          "sort": "none"
+        }
+      },
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "${datasource}"
+          },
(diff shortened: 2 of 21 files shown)


qiffang (changes requested):
## adversary-2 R1 review — PR #534 (Add persistent SSE event replay)

**Verdict: REVISE** — 2 correctness blockers, 2 bugs.

### B1 (BLOCKER): `publishEvent` persist failure silently drops the event — no in-memory fallback

In `sse.go` `publishEvent`, when `InsertFSEvent` fails:

```go
if err == nil {
    // ... PublishEvent + return
}
logger.Error(...)
recordSSEOperation("persist", sseResultError, persistStart)
return  // <--- exits without calling bus.Publish
```

If the durable persist fails (DB timeout, connection error, transient TiDB issue), the event is **not published to the in-memory EventBus at all**. Live SSE subscribers get nothing. The mount's cache invalidation for this mutation is silently lost.

The function returns before reaching the `bus.Publish(path, op, actor)` fallback because the durable path's `return` is inside the `if b := backendFromRequest(r); b != nil && b.Store() != nil` block, and the fallback `bus.Publish` is after this block.

This is a correctness regression from the pre-PR behavior where every mutation always published to the in-memory bus.

### B2 (BLOCKER): `durableMu` serializes ALL mutations across ALL paths for the entire tenant

`publishEvent` acquires `bus.durableMu.Lock()` before the DB insert. This means:
- All concurrent writes/chmod/rename/delete/mkdir/copy for the same tenant are serialized through a single mutex
- The DB insert has a 5s timeout (`ssePersistTimeout`)
- If the DB is slow, ALL mutations for that tenant stall waiting for `durableMu`

This is a significant performance regression. The original `bus.Publish` was fast (in-memory only, sub-microsecond under `bus.mu`). Now every mutation path pays the cost of DB round-trip serialized through a single lock.

The lock exists to ensure seq ordering matches DB insert order, but this is already guaranteed by `AUTO_INCREMENT` — the DB assigns monotonic seq values regardless of insertion order. The `durableMu` is redundant with the DB's own auto-increment guarantee. If the concern is that `PublishEvent` must see events in order, that's already handled by `PublishEvent`'s `ev.Seq <= eb.seq` check.

### B3 (BUG): `persistentEventsSince` gap detection resets on pruned-but-valid ranges

```go
if oldestSeq > 0 && since+1 < oldestSeq {
    // seq_too_old
}
```

Then later:
```go
expectedSeq := since + 1
for _, ev := range events {
    if ev.Seq != expectedSeq {
        // seq_too_old — gap detected
    }
}
```

The gap detection is correct for detecting actual data loss, but there's an edge case: if `PruneFSEventsBefore` runs between `FSEventBounds` and `ListFSEventsSince`, the bounds check passes but the list returns fewer events than expected, triggering the gap detection. This is a TOCTOU race between bounds check and list query.

More critically: the pruning in `publishEvent` runs `PruneFSEventsBefore(ev.Seq - ssePersistentRetention + 1)` inline on every insert when `ev.Seq > ssePersistentRetention`. This means concurrent reconnect replay and publish can race: publish prunes old events while replay is reading them. The pruning should either be deferred/batched or the replay should handle partial results gracefully rather than resetting.

### B4 (BUG): `context.WithoutCancel` strips deadline — persist can hang for 5s after client disconnects

```go
persistCtx, cancel := context.WithTimeout(context.WithoutCancel(r.Context()), ssePersistTimeout)
```

This is the same pattern from PR #493 lesson: `WithoutCancel` strips the parent's deadline AND cancel. The 5s timeout is explicit here, so it won't hang forever. But: the mutation handler (write/chmod/etc.) has already returned to the client at this point? No — `publishEvent` is called synchronously from `handleChmod`, `handleWrite`, etc. So this 5s timeout is added to every mutation's response latency. Combined with B2's serialization, a slow DB makes ALL mutations for a tenant wait up to 5s each, sequentially.

**O1**: `fs_events` table has no index on `seq` for range queries — `WHERE seq > ? ORDER BY seq ASC LIMIT ?` relies on the primary key index, which is correct for InnoDB/TiDB but should be verified for the DB9/Postgres path where `BIGSERIAL PRIMARY KEY` creates an implicit index.

**O2**: `sendSSEReset` now takes `sseOperationResult` instead of `string`, but `sendSSEStructuralReset` still passes the string constant `string(sseResultStructural)` via the `sseResetPayload.Reason` field. The type safety is incomplete — `Reason` is still `string` in the JSON payload.

**O3**: The client SSE parser now handles `data:` without space after colon (`data:value`), which is per-spec correct (`data: value` is optional space). Good fix.

### Checklist
- [x] Schema consistent across tidb_auto, tidb_app, db9, test helpers
- [x] Chmod now increments revision + mtime + emits SSE
- [x] Client parser handles multiline data + comments
- [x] EventBus.PublishEvent rejects older seq
- [x] Persistent replay gap detection present
- [ ] **B1**: Persist failure silently drops event from live subscribers
- [ ] **B2**: durableMu serializes ALL tenant mutations through single lock + DB round-trip
- [ ] **B3**: Prune-during-replay TOCTOU race
- [ ] **B4**: 5s persist timeout added to every mutation response latency path

qiffang (changes requested):
## dat9-dev1 R1 review — PR #534 (Add persistent SSE event replay)

**Verdict: REVISE** — I agree with @adversary-2's B1-B4 findings. Adding my own analysis and one additional concern.

### B1 (BLOCKER, agree): Persist failure silently drops event from live subscribers

Confirmed. In `publishEvent` (`sse.go:172-199`), when `InsertFSEvent` fails, the function returns without calling `bus.Publish`. This is a correctness regression: pre-PR, every mutation always reached the in-memory bus. The durable persistence is a bonus; it must not degrade the baseline live-delivery guarantee.

**Fix**: On persist failure, fall through to the existing `bus.Publish(path, op, actor)` below the `if b != nil` block. The volatile event still reaches connected subscribers. Log the error (already done) and record the metric (already done).

### B2 (BLOCKER, agree): `durableMu` serializes ALL tenant mutations

Confirmed. `durableMu` is held across `InsertFSEvent` + `PruneFSEventsBefore` + `bus.PublishEvent`, which means every mutation handler (`handleWrite`, `handleChmod`, `handleRename`, etc.) for the same tenant serializes through a single lock with up to 5s timeout.

AUTO_INCREMENT guarantees monotonic seq assignment per-insert regardless of insertion order. The only concern is that `bus.PublishEvent` might receive events out-of-order if two goroutines insert concurrently and get seq N and N+1 but publish in reverse order. However, `PublishEvent` already handles this: it drops events with `ev.Seq <= eb.seq`. The worst case is a temporary gap in the ring, which is acceptable since the ring is a best-effort buffer and the durable log is the replay source.

**Fix**: Remove `durableMu`. Accept that ring buffer may have transient gaps (already handled by ring overflow → reset). The durable log is the source of truth for replay.

### B3 (BUG, agree): TOCTOU between prune and replay

Confirmed. `PruneFSEventsBefore` runs inline during `publishEvent`, which can race with `persistentEventsSince` on a reconnecting client. The gap detection (`ev.Seq != expectedSeq`) will trigger a false reset.

**Fix**: Move pruning to a background sweep (e.g., periodic timer in the server's event bus init) rather than inline on every insert. This also reduces per-mutation latency.

### B4 (BUG, agree): 5s persist timeout on synchronous mutation path

Confirmed. `publishEvent` is called synchronously from `handleChmod`, `handleWrite`, etc. The `ssePersistTimeout` of 5s is added to every mutation's response latency. Combined with B2 serialization, a slow DB can cascade.

**Fix**: Consider making persist asynchronous (goroutine + buffered channel), similar to the main PR's async update check pattern. Or reduce timeout significantly (500ms) and fall back to volatile publish on timeout.

### B5 (NEW, BUG): `persistentEventsSince` returns `sseResultServerRestart` when durable log is empty

In `persistentEventsSince` (`sse.go:1101-1113`):

```go
if count == 0 {
    recordSSEOperation("replay", sseResultServerRestart, start)
    return nil, headSeq, false, sseResultServerRestart, nil
}
```

When `count == 0` and `since > 0`, the function returns `server_restart` reason even if the server never restarted — this is just the first time durable persistence was enabled, or all events were pruned. The client will interpret `server_restart` as needing full cache invalidation, which is correct behavior but the reason label is misleading. Consider `sseResultSeqTooOld` instead, or a new `sseResultNoHistory` to distinguish from actual server restarts.

### Good things

- Schema consistency across tidb_auto, tidb_app, db9, and all test helpers — all include `fs_events` table
- `Chmod` now correctly increments revision + mtime + emits SSE event — closes the cache-invalidation spec gap
- Client SSE parser multiline data + comment handling is per-spec correct
- `EventBus.PublishEvent` correctly rejects older seq and advances cursor
- `persistentEventsSince` gap detection is solid (modulo B3 TOCTOU)
- Grafana dashboard and metrics integration is thorough
- Cache invalidation spec correctly updated to reflect chmod coverage

### Summary

The core design (durable event log for SSE replay across server restarts) is sound and fills a real gap. The four issues above are all fixable without architectural changes. The critical fix is B1 (persist failure must fall through to volatile publish, not drop the event entirely).

qiffang (commented):
## R2 Re-review — PR #534 (Persistent SSE Event Replay)

**Verdict: GREEN** ✅ — All R1 blockers resolved. CI green.

### R1 Blocker Resolution

**B1-B4** (from my R1): All fixed in this version. The `publishEvent` path now persists first, then publishes to in-memory bus; on persist failure, falls through to volatile `bus.Publish` (this addresses dat9-dev1's B5 as well).

**B5 (persist failure drops event, @dat9-dev1)**: FIXED. Lines 1104-1131: persist is attempted first; on error, logs the failure and falls through to `bus.Publish` for volatile delivery. Event is never silently dropped.

### @qiffang's question: persistent queue reuse

**No existing reusable persistent queue for this use case.** I checked the codebase:

1. **`file_gc_tasks`** — task queue pattern (claim/lease/retry/dead-letter). Semantically a work queue, not an event log. Different table shape, different access patterns. Not applicable.
2. **`pkg/fuse/journal.go`** — client-side FUSE WAL (local file I/O, CRC checksums, truncation log). Per-mount, not per-tenant server-side. Not applicable.
3. **`pkg/datastore/journal.go`** — content journaling (create/append/read journal documents with hash chains). For content audit trails, not event streams. Not applicable.
4. **`fs_layer_entries`** (PR #507) — layered FS event log. Per-layer, not per-tenant. Different domain.

The `fs_events` table is a new per-tenant append-only event log with `INSERT` + `SELECT WHERE seq > ? ORDER BY seq LIMIT ?` + `DELETE WHERE seq < ?`. This is the correct shape for SSE replay (AUTO_INCREMENT primary key = gap-free sequence, simple retention pruning). The existing queue/log tables have fundamentally different schemas, access patterns, and retention semantics. **No redundant code.**

However, I note that **the PR does not include a comment explaining why `fs_events` is new instead of reusing an existing table**. Recommend adding a brief doc comment on the `FSEvent` struct or the `fs_events.go` file header explaining the relationship to other persistent stores.

### Architecture review

The per-tenant SSE persistent event flow:

1. **Write path**: `publishEvent` → `InsertFSEvent` (tenant-scoped DB, AUTO_INCREMENT seq) → `bus.PublishEvent` (in-memory fan-out with durable seq)
2. **Reconnect replay**: `eventsSince` prefers `persistentEventsSince` (DB) over `bus.EventsSince` (ring buffer). Validates gap-free sequence continuity.
3. **Live delivery**: unchanged — in-memory ring + listener channel
4. **Retention**: `sseRetentionSweeper` background goroutine per-tenant, at most every minute, keeps `ssePersistentRetention` (100K) events
5. **Fallback**: persist error → volatile publish. DB unavailable → ring-only replay with reset.

**Sequence ordering guarantee**: `fs_events.seq` is `BIGINT UNSIGNED AUTO_INCREMENT PRIMARY KEY` — MySQL/TiDB guarantees gap-free monotonic sequences within a single `INSERT`. `PublishEvent` validates `ev.Seq == eb.seq+1` and emits force-reset on mismatch. `persistentEventsSince` validates gap-free replay with `expectedSeq` tracking. Ordering is preserved.

**O1**: `sseRetentionSweeper` uses `map[string]time.Time` and `map[string]bool` keyed by tenant ID. With many tenants these maps grow unbounded. Consider periodic cleanup of idle entries. V1 acceptable.

**O2**: `PublishEvent` sends force-reset when `ev.Seq != eb.seq+1` (line 865). If multiple server instances share the same DB but have separate in-memory buses, the first reconnect after a foreign insert will trigger a reset. This is correct behavior but worth documenting.

**O3**: `persistentEventsSince` fetches `ssePersistentReplayLimit+1` events (10001) to detect overflow. For clients that are very far behind, this is a full table scan up to the limit. The `WHERE seq > ? ORDER BY seq LIMIT ?` uses the primary key index, so this is efficient. Fine for V1.

No blockers. CI green.

qiffang (approved):
## R2 Review — PR #534 SSE Persistent Event Replay

**Verdict: ✅ GREEN**

### Summary

20 files changed, 1675 lines. Adds durable per-tenant event log (`fs_events` table) for SSE reconnect replay across server restarts, with graceful fallback to in-memory ring buffer. Also fixes the chmod→SSE gap documented in cache-invalidation.md.

### @qiffang 关于持久化队列复用的分析

**结论：没有可复用的现有持久化队列实现，新建 `fs_events` 表是正确的。**

| 现有表/模块 | 模式 | 为什么不能复用 |
|---|---|---|
| `file_gc_tasks` | claim/lease/retry work queue | 任务队列，不是 event log。schema 完全不同（status, lease_until, attempt_count） |
| `pkg/fuse/journal.go` | 客户端 FUSE WAL（本地文件 I/O + CRC） | per-mount 客户端侧，不是 server-side per-tenant |
| `pkg/datastore/journal.go` | 内容审计日志（hash chain + canonical JSON） | 溯源日志，不是事件流 |
| `pkg/fuse/commit_queue.go` | FUSE 文件上传队列 | 客户端侧上传编排，不是 server event log |
| `pkg/server/semantic_worker.go` | embedding task worker pool | 完全不同的领域 |

`fs_events` 是 **per-tenant append-only event log**：`AUTO_INCREMENT PK` = gap-free 序列号，`INSERT` 写入，`SELECT WHERE seq > ? ORDER BY seq LIMIT ?` replay，`DELETE WHERE seq < ?` retention pruning。这是 SSE reconnect replay 的标准 schema，与上述存储在 schema、访问模式、retention 语义上完全不同。**没有冗余代码。**

**建议**：在 `fs_events.go` 文件头加一段注释说明为什么新建表而不复用现有持久化存储（`file_gc_tasks` 是 work queue，journal 是 content provenance，FUSE journal 是 client-side），方便后续维护者理解设计决策。

### Architecture Review

**写入路径**:
```
mutation handler → InsertFSEvent(tenant DB) → PublishEvent(in-memory bus, durable seq)
  └─ persist fail → fallback volatile Publish (event never lost)
```

**重连回放**:
```
eventsSince → persistent DB (preferred) → in-memory ring (fallback)
  └─ seq gap/too old/overflow → reset signal
```

**顺序保证**: `AUTO_INCREMENT PK` guarantees gap-free monotonic sequence. `PublishEvent` validates `seq == eb.seq+1`, detects out-of-order → force reset. `persistentEventsSince` validates gap-free replay. Correct.

**保留策略**: `sseRetentionSweeper` per-tenant, at most once/min, keeps 100K events. Async with concurrency guard (`running[tenantKey]`). Correct.

### R1 Blocker Status

All fixed:
- **B1 (persist fail drops event)**: Fixed — `publishEvent` falls through to volatile `bus.Publish` on persist error.
- **B2 (durableMu serializes tenant)**: Fixed — removed; uses DB AUTO_INCREMENT for sequencing instead of mutex.
- **B3 (prune TOCTOU)**: Fixed — `PruneFSEventsBefore` uses `DELETE WHERE seq < ?` (idempotent), sweeper has `running[tenantKey]` guard.
- **B4 (5s sync latency)**: Fixed — persist timeout reduced to 500ms (`ssePersistTimeout`).
- **B5 (misleading server_restart label)**: Fixed — reason derivation moved to `eventsSince`/`persistentEventsSince` with accurate labels (`no_history`, `seq_too_old`, `server_restart`, `initial_sync`).

### Additional Changes Reviewed

- **chmod SSE event**: `handleChmod` now calls `publishEvent(r, path, "chmod")`. `Chmod` in store increments revision + updates mtime/confirmed_at with monotonic clock guard. cache-invalidation.md updated to remove chmod gap. ✅
- **SSE client parser**: Supports multi-line `data:` fields and SSE comments (`:` prefix). Buffer increased to 4MB. Test added. ✅
- **EventBus.EventsSince strict seq validation**: Rejects gaps in ring walk (`ev.Seq != expectedSeq`). ✅
- **Grafana dashboard**: New `drive9-event-stream-dashboard.json` with 6 panels. ✅
- **Schema**: `fs_events` table added to TiDB, DB9, test helpers, and testmysql reset. ✅

### Non-blocking Observations

- O1: `sseRetentionSweeper.last` and `running` maps grow unboundedly with tenant count. For high-tenant deployments, consider periodic cleanup of stale tenant keys (e.g., tenants with no events for 24h).
- O2: Multi-server instances sharing the same tenant DB will each have independent in-memory buses. An external insert (from another server) will cause `PublishEvent` to detect seq mismatch and emit force-reset. This is correct behavior — worth a comment for future maintainers.
- O3: `persistentEventsSince` fetches `LIMIT 10001` to detect overflow, using PK index. V1-sufficient; may need cursor-based pagination at scale.

No blockers. Ship it.

PR #535, opened 2026-06-10
feat: CLI update check with interactive upgrade prompt

qiffang:
## Summary

Adds a non-blocking CLI update checker inspired by [gh CLI](https://github.com/cli/cli/blob/trunk/internal/update/update.go):

- **Async check**: goroutine starts before command execution, notice displays after completion — never blocks the user
- **stderr only**: all output goes to stderr; stdout is never polluted (safe for `drive9 fs cat | ...` pipes)
- **Interactive prompt**: on TTY, offers `Update now` (with second confirmation) or `Skip this version`
- **Non-interactive**: just prints the update command
- **24h TTL cache**: stored in `~/.config/drive9/update-check.json` with atomic write (write-temp + rename)
- **Semver comparison**: uses `golang.org/x/mod/semver`, not string compare
- **Excluded commands**: `mount`, `umount`, `completion`, `__complete`, `daemon` never show prompts
- **CI detection**: `CI`, `GITHUB_ACTIONS`, `JENKINS_URL`, etc. suppress all checks
- **Explicit disable**: `DRIVE9_NO_UPDATE_CHECK=1`
- **Fail-safe**: any error (network, parse, corrupt state) silently skips — never affects command execution

### Design references
- gh CLI: async goroutine + post-command stderr notice (same pattern)
- Codex: interactive prompt with choices (adapted with second confirmation for `curl|sh` safety)

## Test plan

- [x] 27 unit tests covering:
  - Dev/unknown/empty version → no check
  - CI env vars → no check  
  - Non-TTY → no check
  - `DRIVE9_NO_UPDATE_CHECK=1` → no check
  - Excluded commands (mount/umount/completion/daemon) → no check
  - Network failure → silent skip
  - Invalid JSON response → silent skip
  - Empty response body → silent skip
  - HTTP 500 → silent skip
  - Context cancelled → silent skip
  - Corrupt state file → silent skip + fresh fetch
  - TTL respected (24h cache, no redundant fetches)
  - Skip version persisted + newer version still prompts
  - Semver comparison (v0.10.0 > v0.9.0, not string compare)
  - Atomic state write (no temp file left behind)
  - Non-interactive print format
  - Nil release → no panic
- [x] `go vet` clean
- [x] Race detector clean
- [x] All existing cmd/drive9 tests pass

## Summary by CodeRabbit

* **New Features**
  * Non-blocking background update check that shows a stderr-only banner after commands finish with current/latest versions, install command, and optional release notes URL.
  * Persistent 24‑hour cached results and “skip this version” suppression; respects opt-out, CI/non‑TTY, excluded commands, and short-SHA/dev/version formats.

* **Tests**
  * Comprehensive unit and end-to-end tests for eligibility gating, version comparison, fetch fallbacks, TTL/caching, persisted state handling, prompt suppression, and banner output.


qiffang (commented):
## R1 Review — PR #535 (CLI Update Check)

**Verdict: REVISE** — 3 blockers, 4 observations.

### Blockers

**B1: CI lint failures — 3 errcheck issues**

CI is red. `golangci-lint` reports 3 unchecked errors:

```
cmd/drive9/cli/updatecheck.go:170:16: Error return value of `json.NewEncoder(tmpFile).Encode` is not checked (errcheck)
cmd/drive9/cli/updatecheck.go:175:21: Error return value of `tmpFile.Close` is not checked (errcheck)
cmd/drive9/cli/updatecheck.go:178:15: Error return value of `os.Rename` is not checked (errcheck)
```

All three are in `writeStateFile`. The `Encode` error means a corrupt state file could be silently written. `tmpFile.Close` error means data may not be flushed to disk. `os.Rename` error means the atomic rename failed silently — the state file is lost.

Fix: check all three errors, propagate or log.

**B2: `DRIVE9_UPDATE_URL` env var overloads both fetch URL and install command**

`DRIVE9_UPDATE_URL` is used in two semantically different roles:

1. **Fetch**: `fetchLatestRelease` uses it as the GitHub API endpoint (JSON response expected)
2. **Display**: `PrintUpdateNotice` uses it as the human-readable install command/URL

These are different things. A custom API endpoint (e.g., `https://internal-proxy.corp/api/releases/latest`) is not a valid install command. The overload means setting the env var for custom release sources breaks the install instruction, and vice versa.

Either: (a) split into two env vars (`DRIVE9_UPDATE_URL` for fetch, `DRIVE9_INSTALL_URL` for display), or (b) only override the fetch URL and always show the default install command, or (c) document the constraint that the URL must work for both purposes.

**B3: `ShouldCheckForUpdate` checks stdout TTY unnecessarily**

```go
func ShouldCheckForUpdate(currentVersion, command string) bool {
    ...
    if !isTerminal(os.Stdout) || !isTerminal(os.Stderr) {
        return false
    }
```

All update notice output goes to stderr via `PrintUpdateNotice`. Checking `os.Stdout` TTY state is unnecessary and causes false negatives: a user running `drive9 status > output.txt` (stdout piped, stderr is TTY) won't see update notices even though the notice would go to stderr and be visible.

Fix: only check `isTerminal(os.Stderr)`.

**O1**: `fetchLatestRelease` uses `http.DefaultClient` with no timeout. The 10-second context timeout from `CheckForUpdate` covers it, but if `DRIVE9_UPDATE_URL` is overridden to a slow endpoint, the goroutine could hang. Consider `http.Client{Timeout: 10 * time.Second}` as defense-in-depth.

**O2**: `writeStateFile` writes to a temp file in `os.TempDir()` then renames to the config dir. Cross-device rename will fail on systems where `/tmp` and `~/.config` are on different filesystems (common in containers). Consider creating the temp file in the same directory as the target.

**O3**: `parseVersion` adds "v" prefix unconditionally: `return "v" + version`. If `currentVersion` already has a "v" prefix (e.g., build system passes `v1.2.3`), this produces `vv1.2.3` which `semver.Compare` will reject as invalid, causing `IsNewer` to always return false. The tests show `buildinfo.Version` is set without "v" prefix, but this is a fragile assumption.

**O4**: The 24-hour TTL check uses `time.Since(state.CheckedAt) < 24*time.Hour`. If the system clock jumps backward (NTP correction, VM resume), `CheckedAt` could be in the future, and `time.Since` returns a negative duration — the check would fire on every invocation. Not critical but worth noting.

qiffang (commented):
## R2 Re-review — PR #535 (CLI Update Check)

**Verdict: GREEN** ✅

All 3 R1 blockers resolved:

- **B1 (errcheck)**: `writeUpdateState` now uses `os.CreateTemp` in same directory (fixes O2 cross-device rename too), checks `Write`/`Close`/`Rename` errors, cleans up temp file on all failure paths. Correct.
- **B2 (DRIVE9_UPDATE_URL overload)**: Env var now only overrides fetch URL in `fetchLatestRelease`. `installCommand()` always returns the hardcoded public install URL. Clean separation documented in comment.
- **B3 (stdout TTY)**: Only checks `isTerminal(os.Stderr)`. Correct — all output goes to stderr.

R1 observations also addressed:
- O2 resolved by B1 fix (same-directory temp).
- O3 resolved: `ensureVPrefix` checks for existing "v" prefix before adding one. No double-prefix risk.

Remaining R1 observations (O1 DefaultClient timeout, O4 clock skew) are non-blocking and acceptable for V1.

CI pending at time of review — verdict conditional on CI green.

qiffang:
## Review — CLI update prompt UX and CI

Verdict: **REVISE**.

### Blockers

1. **CI is red on real lint failures.**

Current `code-ci / ci` fails in `make lint`:

```text
cmd/drive9/cli/updatecheck.go:370:18: Error return value of resp.Body.Close is not checked
cmd/drive9/cli/updatecheck_test.go:569:9: Error return value of w.Close is not checked
cmd/drive9/cli/updatecheck.go:248:6: QF1001: could apply De Morgan's law
```

These are easy to fix, but the PR cannot merge red.

2. **The production update source does not match the implementation, so users will not see prompts.**

The default checker points at:

```text
https://drive9.ai/releases/latest.json
```

I checked the live endpoint: it returns `404`. The published release file today is:

```text
https://drive9.ai/releases/version
```

That returns a short SHA (`cbf73aa` at review time), not JSON and not semver. The current implementation only decodes JSON into `ReleaseInfo` and then rejects non-semver versions. So even after CI is fixed, the default production path silently returns `nil` and never shows an update notice.

This needs one explicit contract:

- Either publish `releases/latest.json` with a semver `version` and release notes URL.
- Or make the checker use `releases/version` and compare SHA-style release builds correctly.

Right now the code comments describe plain-text support, but the implementation does not implement it.

3. **The interactive menu is not elegant for a one-shot CLI command.**

Current behavior after a normal command can be:

```text
A new Drive9 CLI is available: 0.8.1 → 0.9.0
Release notes: ...

  1. Update now
  2. Skip this version

Choose [1/2]:
```

This is Codex-like, but Codex is an interactive long-running app. `drive9 fs ls`, `drive9 version`, or `drive9 ctx` are one-shot CLI commands. For those, a blocking numbered menu after the command is surprising and interrupts the shell workflow.

The safer modern CLI pattern is closer to `gh`, `npm`, `rustup`, etc.: print a short non-blocking notice on stderr, tell the user the exact update command, and return control to the shell. If Drive9 wants a one-key upgrade flow, expose it as an explicit command such as `drive9 update`, where an interactive confirmation is expected.

Suggested shape:

```text
A new Drive9 CLI is available: 0.8.1 → 0.9.0
Update: curl -fsSL https://drive9.ai/install.sh | sh
Release notes: https://...
```

Optional: use subtle yellow/bold when stderr is a real TTY and `NO_COLOR` is not set. Avoid a heavy box unless it handles terminal width cleanly.

### Additional correctness issues still present

- `executeUpdate` ignores the second `scanner.Scan()` result. EOF/read failure still falls through and runs the installer.
- `isTerminal` treats any character device as a TTY. Use a real terminal probe (`term.IsTerminal`) while keeping the test seam.
- `XDG_CONFIG_HOME` accepts relative paths, so state can be written relative to arbitrary CWDs.

### Bottom line

The background-check architecture is good: TTL, CI suppression, stderr-only output, excluded daemon/mount paths, and fail-closed networking are the right primitives. But the production endpoint mismatch means the feature currently does not work, and the blocking interactive menu is the wrong UX for ordinary Drive9 CLI commands.

qiffang:
## R2 Review — CLI update prompt (`acd407f`)

Verdict: **REVISE**.

The interaction is much better than R1: the blocking numbered menu is gone, `curl | sh` is no longer executed from an incidental command, and the notice is now stderr-only/non-blocking. That is the right direction and closer to `gh`/`npm` style.

Local validation on `acd407f`:

```text
go test ./cmd/drive9/cli -count=1                         PASS
go test ./cmd/drive9/cli -race -count=1                   PASS
go vet ./cmd/drive9/cli ./cmd/drive9                      PASS
make lint                                                  PASS
```

Remote CI was still running when I checked.

### Remaining blockers

**B1 — Update source still does not match the release workflow**

The default checker still points at:

```text
https://drive9.ai/releases/latest.json
```

Live result is `404`. The release workflow currently publishes only `site/releases/version`, and that file currently contains a short SHA (`cbf73aa` when checked), not semver.

The new plain-text fallback only helps for plain-text semver bodies like `v0.9.0`. It does not make the actual production contract work, because `fetchLatestRelease` still rejects SHA latest versions:

```go
if !isValidSemver(rel.Version) {
    return nil
}
```

So the official release path is still unable to produce an update notice unless either:

- release publishing starts producing `releases/latest.json` with semver metadata, and release binaries also use semver-compatible `buildinfo.Version`; or
- the checker switches to `/releases/version` and explicitly supports SHA-vs-SHA update detection, for example `latestSHA != currentSHA`.

Right now `ShouldCheckForUpdate` allows current SHA versions, but `fetchLatestRelease` rejects latest SHA versions. That is internally inconsistent and still makes production behavior silent.

**B2 — The banner can repeat on every eligible command within the 24h TTL**

The state TTL throttles network fetches, not notices. If a newer version is cached, this path returns the cached release on every interactive invocation for the next 24h:

```go
if state.LatestVersion != "" && isUpdateAvailable(state.LatestVersion, currentVersion) {
    return &ReleaseInfo{...}
}
```

After removing `Skip this version`, users have no per-version way to silence it except disabling update checks globally. That turns the improved non-blocking banner into repeated noise for every `drive9 fs ls`, `drive9 ctx`, etc. Modern CLI notices are usually cadence-throttled as user-facing prompts, not just network-throttled.

Fix direction: persist either `last_prompted_at` or `skipped_version`, and show at most once per TTL/version unless the user asks explicitly. If you want to avoid interactive skip, a simple `last_prompted_version + last_prompted_at` is enough for V1.

### UX assessment

The new banner is acceptable as a V1 direction. It is still a bit heavier than `gh` because the box can become wide if release notes URLs are long, but that is an observation, not a blocker. The important UX principles are now right: no prompt, no installer side effect, stderr only, and no stdout pollution.

I would not merge until B1 and B2 are fixed, because B1 affects whether the feature works at all in production, and B2 affects whether the prompt feels polished after it starts working.

qiffang:
## R5 Review — CLI update prompt (`dc008e8`)

Verdict: GREEN, conditional on CI green.

The remaining release-contract blocker is now closed. The release workflow has a semver path: tag push `v0.9.0` triggers `release-cli.yml`, `Determine version` uses the tag as `VERSION`, `dist/version` is published to `/releases/version`, and the update checker accepts that semver value. The existing workflow_run / workflow_dispatch SHA fallback remains intact, so current deploy-triggered releases are not broken.

Previously fixed items remain valid:
- The prompt UX is now a non-blocking stderr banner, not an interactive menu.
- `MarkUpdatePrompted` + `LastPromptedAt` / `LastPromptedVersion` suppress repeat banners for the same version within the TTL.
- The endpoint now matches the published release file path.
- Dead `SkippedVersion` code is removed.

Validation:
- `/usr/local/go/bin/go test ./cmd/drive9/cli -count=1` passed
- `/usr/local/go/bin/go test ./cmd/drive9/cli -race -count=1` passed
- `/usr/local/go/bin/go vet ./cmd/drive9/cli ./cmd/drive9` passed
- `PATH=/usr/local/go/bin:$PATH make lint` passed
- `git diff --check github/main...HEAD` passed

Remote GitHub checks were still in progress when reviewed: `code-ci`, `local-e2e`. CodeRabbit was green.

srstack (commented):
## Code Review: feat: CLI update check with interactive upgrade prompt

### Overview
Adds a non-blocking background update checker to the CLI. A goroutine fetches the latest version before the command runs; a banner is printed to stderr after it completes. State is persisted to `~/.config/drive9/update-check.json` with a 24h TTL. SHA-based release builds (7-char hex, the current non-semver release format) are treated as always updatable when a semver release is available. The `release-cli.yml` is updated to emit semver tag names as the version string when triggered by a tag push.

---

### Strengths

- **stderr-only, TTY-gated, CI-detected**: follows gh CLI conventions closely, zero risk of stdout pollution in pipes.
- **Atomic state write**: write-temp + rename prevents partial-state reads between invocations.
- **SHA version awareness**: correctly handles the current release format (7-char hex) vs. future semver tags.
- **Fail-safe throughout**: every network/parse/IO error silently skips; never affects command execution.
- **27 tests** covering the full eligibility matrix, TTL logic, state corruption, and version comparison edge cases.

---

### Issues

**1. `MarkUpdatePrompted` is never called — banner repeats on every eligible command within the TTL window**

`CheckForUpdate` returns a release only when `LastPromptedVersion != LatestVersion` or `LastPromptedAt` is outside the TTL. The suppression logic is correct, but `MarkUpdatePrompted` must be called after the banner is shown to arm it. Grepping the diff, `MarkUpdatePrompted` is defined and tested but there is no call to it in `main.go` or the dispatch loop. Without the call, the banner will appear on every eligible command every 24h rather than once per version per TTL.

**2. `updateLatestURL` is a mutable package-level var — safe for sequential tests, fragile for parallel ones**

```go
var updateLatestURL = "https://drive9.ai/releases/version"
```

Tests that override this via `t.Setenv("DRIVE9_UPDATE_URL", ...)` are fine. But any test that writes directly to `updateLatestURL` would race with parallel test runs. Prefer the env-var override exclusively and make `updateLatestURL` a constant; the env override in `fetchLatestRelease` already handles testability.

**3. Box rendering uses `len()` (byte count) rather than rune count**

```go
maxLen := 0
for _, l := range lines {
    if len(l) > maxLen { maxLen = len(l) }
}
```

Version strings are ASCII-safe in practice, but `len(l)` counts bytes, not display columns. If a URL in `rel.URL` ever contains multibyte characters (unlikely but possible), the right border of the box would misalign. Use `utf8.RuneCountInString(l)` for the width calculation.

**4. `http.DefaultClient` is used directly**

`fetchLatestRelease` uses `http.DefaultClient.Do(req)`. The context timeout guards latency, but using `DefaultClient` means the update check shares the global transport (connection pool, TLS config) with any other HTTP code in the process. A dedicated `&http.Client{}` with a short timeout (matching `updateCheckTimeout`) would be cleaner and more isolated.

**5. `isSHAVersion` rejects uppercase hex**

```go
if (c < '0' || c > '9') && (c < 'a' || c > 'f') {
```

Git SHAs from some tools (`git rev-parse --short HEAD` on some platforms) may be uppercase. Lowercase is the common case for Go toolchain builds, but worth a `strings.ToLower` normalization to be safe.

---

### Minor

- `isTerminal` uses `ModeCharDevice` — correct on Linux/macOS, but on Windows `os.File.Stat()` may not set this bit for console handles. The Windows path already returns the `curl|sh` command rather than a PowerShell prompt, so this inconsistency is low impact.
- `installCommand()` on Windows returns `powershell -Command "irm ... | iex"` — this installs from the internet without a hash check. Fine for a notice banner, but worth a comment noting this is display-only.
- `clearCIEnv(t)` is used in tests but not shown in the diff — confirm it unsets all 10 CI vars from `isCI()`, not just `CI`.

---

### Verdict

The architecture is solid and the test coverage is thorough. Two items need fixing before merge: (1) the missing `MarkUpdatePrompted` call in the dispatch path (otherwise the TTL suppression is dead code), and (2) the `http.DefaultClient` usage. The rest are minor polish.

srstack (commented):
## Code Review: feat: CLI update check with interactive upgrade prompt

### Overview
Adds a non-blocking background update checker. A goroutine fetches the latest version before the command runs; a banner is printed to stderr after completion. State persisted to ~/.config/drive9/update-check.json with 24h TTL. SHA-based release builds (7-char hex) are treated as always updatable when a semver release is available. release-cli.yml is updated to emit semver tag names as the version string on tag pushes.

---

### Strengths

- stderr-only, TTY-gated, CI-detected: follows gh CLI conventions, zero stdout pollution risk.
- Atomic state write: write-temp + rename prevents partial-state reads.
- SHA version awareness: correctly handles current 7-char hex release format vs future semver tags.
- Fail-safe throughout: every network/parse/IO error silently skips.
- 27 tests covering eligibility matrix, TTL logic, state corruption, and version comparison edge cases.

---

### Issues

**1. MarkUpdatePrompted is never called -- banner repeats on every eligible command**

CheckForUpdate returns a release only when LastPromptedVersion != LatestVersion or LastPromptedAt is outside the TTL. The suppression logic is correct, but MarkUpdatePrompted must be called after the banner is shown to arm it. There is no call to it in main.go or the dispatch loop. Without this, the banner appears on every eligible command every 24h rather than once per version per TTL.

**2. updateLatestURL is a mutable package-level var**

```go
var updateLatestURL = "https://drive9.ai/releases/version"
```

Tests using t.Setenv("DRIVE9_UPDATE_URL", ...) are fine. But any test that writes directly to updateLatestURL would race with parallel test runs. Prefer the env-var override exclusively and make updateLatestURL a constant.

**3. Box rendering uses len() (byte count) not rune count**

```go
if len(l) > maxLen { maxLen = len(l) }
```

Version strings are ASCII-safe in practice, but len(l) counts bytes. If rel.URL ever contains multibyte characters, the right border misaligns. Use utf8.RuneCountInString(l) for the width calculation.

**4. http.DefaultClient used directly**

fetchLatestRelease uses http.DefaultClient.Do(req), sharing the global transport with any other HTTP code in the process. A dedicated &http.Client{Timeout: updateCheckTimeout} would be cleaner and more isolated.

**5. isSHAVersion rejects uppercase hex**

Git SHAs from some tools may be uppercase. Worth a strings.ToLower normalization to be safe.

---

### Minor

- isTerminal uses ModeCharDevice -- correct on Linux/macOS, may not work for Windows console handles. Low impact since the Windows path is already handled separately.
- clearCIEnv(t) used in tests but not shown in diff -- confirm it unsets all 10 CI vars from isCI(), not just "CI".

---

### Verdict

The architecture is solid and test coverage is thorough. Two items need fixing before merge: (1) the missing MarkUpdatePrompted call in the dispatch path (the TTL suppression is otherwise dead code), and (2) the http.DefaultClient usage. The rest are minor polish.

PR #537, opened 2026-06-10
perf: escalate negative-lookup storms to a single directory listing

qiffang:
## Summary
- A 10k-file fio prepare on a FUSE mount spent **13min22s (94% of wall time)** in the layout phase, paying one synchronous remote stat per missing name at ~50 lookups/s (fio probes every target file 3×before writing).
- Lookup now detects negative-lookup storms: once 3 remote ENOENT stats hit the same directory within 1s, it issues a **single listing** and the namespace cache answers the remaining misses locally, under the same negative-TTL staleness contract. Expected cost for the fio case: ~3 stats + 1 list per directory instead of 30,000 stats.
- Guards: listings that fail or exceed the 2000-entry cache limit set a one-dirTTL cooldown so sustained storms cannot turn into repeated listings; lock files are excluded because they deliberately bypass the namespace cache; `LegacyDirStatFallback` mode is excluded because it already lists per miss.
- Also fixes `MarkSessionCreatedDir` to keep its complete/session markers for the directory TTL (10s) instead of the negative TTL (1s) — the marker expired 1s after Mkdir, making it useless for the create-heavy workloads it was meant to absorb. Local mutations flow through Upsert/Remove and foreign writes invalidate via SSE, so the longer window is safe (covered by existing `TestLookupSSEForeignCreateInvalidatesSessionCreatedMiss`).
- Adds a `lookup_storm_list` perf counter to observe escalations in production.

## Test plan
- [x] New unit tests: session marker outlives negative TTL, storm threshold/window/cooldown semantics, `CanAnswerMisses`
- [x] New integration tests: 10 missing-name lookups → exactly 3 HEAD + 1 LIST, later misses and positive hits answered locally; oversized listing → exactly 1 LIST then cooldown
- [x] `go test ./pkg/fuse/` — only pre-existing failure `TestDiskReadCacheStartupRecoveryAndTempCleanup` (fails identically on clean `main`)
- [x] `go build ./...` and `go vet ./pkg/fuse/` clean
- [ ] Re-run the 10k×2MB fio prepare on the EC2 test host and compare layout-phase duration and `lookup_storm_list`

## Summary by CodeRabbit

* **Performance**
  * Reduced redundant remote checks for repeated lookups of missing files in the same directory by warming the directory namespace, improving responsiveness and lowering remote operations.
  * Handles oversized listings by applying a cooldown to avoid repeated expensive attempts.

* **Tests**
  * Added tests validating negative-lookup escalation, cooldowns, and namespace-cache behaviors to improve reliability.

### DIFF
--- a/pkg/fuse/dat9fs.go
+++ b/pkg/fuse/dat9fs.go
@@ -3562,6 +3562,37 @@ func (fs *Dat9FS) cacheNegativePath(p string) {
 	fs.dirCache.MarkNegative(parentPath, name)
 }
 
+// maybeListOnNegativeLookupStorm converts a burst of remote ENOENT lookups
+// under one directory into a single ListDir. Workloads that probe many
+// missing names (fio layout passes, build systems, archive extraction)
+// otherwise pay one remote stat per name; one listing lets the namespace
+// cache answer the rest of the burst locally under the same negative-TTL
+// staleness contract.
+//
+// Lock files are excluded because lookupFromDirCache deliberately bypasses
+// the namespace cache for them, so a listing could never absorb their misses
+// and each lock probe would re-trigger escalation.
+func (fs *Dat9FS) maybeListOnNegativeLookupStorm(cancel <-chan struct{}, parentPath, childP string) {
+	if fs.dirCache == nil || isLockFilePath(childP) || fs.opts.LegacyDirStatFallback {
+		return
+	}
+	if !fs.dirCache.RecordRemoteNegative(parentPath) {
+		return
+	}
+	if fs.perf != nil {
+		fs.perf.lookupStormList.add(1)
+	}
+	items, err := fs.lookupListWithRetry(cancel, parentPath)
+	if err != nil || !fs.dirCache.CanAnswerMisses(parentPath) {
+		// Listing failed or was too large to mark complete; cool down so a
+		// sustained probe storm cannot hammer the server with listings.
+		fs.dirCache.DeferEscalation(parentPath)
+		fs.debugf("lookup storm list deferred dir=%s items=%d err=%v", parentPath, len(items), err)
+		return
+	}
+	fs.debugf("lookup storm list dir=%s items=%d", parentPath, len(items))
+}
+
 func (fs *Dat9FS) negativeEntryTTL(p string) time.Duration {
 	if isLockFilePath(p) {
 		return 0
@@ -4127,6 +4158,7 @@ func (fs *Dat9FS) Lookup(cancel <-chan struct{}, header *gofuse.InHeader, name s
 		// Cache negative lookup: tell kernel this entry doesn't exist
 		// for NegativeEntryTTL so it doesn't re-ask immediately.
 		fs.cacheNegativePath(childP)
+		fs.maybeListOnNegativeLookupStorm(cancel, parentPath, childP)
 		out.NodeId = 0
 		out.SetEntryTimeout(fs.negativeEntryTTL(childP))
 		return gofuse.ENOENT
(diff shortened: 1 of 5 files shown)


srstack (commented):
## Review: perf: escalate negative-lookup storms to a single directory listing

### Overview

Well-diagnosed and well-executed fix. The approach is conservative and correct — good guards for lock files, legacy mode, oversized directories, and listing failures. The `MarkSessionCreatedDir` TTL fix is a correct bonus fix. Test coverage is strong.

---

### Potential Issues

**1. Synchronous listing on the Lookup hot path**

The `escalateMissThreshold`-th `Lookup` call will block inline on the listing RPC. If the directory listing is slow (large dir, slow network), this adds visible latency to exactly one request per storm. Consider whether a goroutine fire-and-forget with a `sync.Once` guard is worth it. The current approach is simpler and correct for most cases, but worth a conscious decision.

**2. `escalateNotBefore` set before listing completes**

Concurrent goroutines hitting the same directory during the in-flight listing will see `RecordRemoteNegative` return false and fall through to individual HEAD stats. This is intentional, but means the tail of a large concurrent burst can still pay individual stats during the listing RPC RTT. Worth a short comment documenting this as accepted behavior.

**3. `DeferEscalation` cooldown interplay**

`RecordRemoteNegative` sets a 1s pre-escalation cooldown; `DeferEscalation` overwrites it with a full `dc.ttl` (10s). The defer correctly wins, but the relationship is non-obvious. A short comment in `DeferEscalation` explaining why the longer duration is used would help future readers.

**4. Test assertion clarity**

- `TestLookupNegativeStormEscalatesToSingleList`: the assertion `headCalls == escalateMissThreshold` (3, not 10) implicitly relies on the listing completing before lookups 4–10 reach the server. True here because the test server is in-process and synchronous — worth a comment.
- `TestLookupNegativeStormOversizedListingCoolsDown`: asserts `headCalls == 10` but doesn't explain why it's not 3. The math (3 trigger the list → cooldown → 7 more HEADs) deserves a brief comment.

---

### Suggestions

- Verify `isLockFilePath(childP)` is checking the child name/suffix, not a full path — it's called with the child path `childP`, which should be correct, just worth a quick sanity check.
- Consider `MountOptions` fields for `escalateMissThreshold` / `escalateMissWindow` — operators on slower networks or higher-IOPS workloads may want to tune these.
- Consider a `lookup_storm_list_deferred` counter alongside `lookup_storm_list` to make the oversized-dir and failure paths observable in production.

---

### Summary

The implementation is correct, well-guarded, and the `MarkSessionCreatedDir` TTL fix is a good correctness improvement on its own. Main things to confirm before merge: (1) acceptable tail latency for the synchronous listing path, and (2) `isLockFilePath` argument correctness. Both are likely fine — just worth a quick look.

mornyx (approved):
Reviewed PR 537.

I do not see any blocking issues. The negative-lookup storm escalation is scoped well: it only triggers after confirmed remote ENOENTs, skips lock-file paths, respects the legacy directory stat fallback path, and cools down after failed or oversized listings so sustained storms do not turn into repeated parent-list requests. The namespace cache changes also preserve the important distinction between complete/session-backed miss handling and partial-cache miss-through behavior.

I also checked the `MarkSessionCreatedDir` TTL change. It does expand the local empty-directory miss window from `NegativeEntryTTL` to `DirTTL`, but that state is a directory/session marker rather than a per-name negative marker, and existing SSE invalidation plus directory TTL bounds still apply. I would keep that tradeoff documented because it is the main consistency/performance behavior change in this PR.

Targeted local verification passed:

```bash
env GOCACHE=/Users/mornyx/Desktop/repos/github.com/mem9-ai/drive9/.codex-gocache go test ./pkg/fuse -run 'TestLookupNegativeStorm|TestNamespaceCache_(SessionCreatedDirOutlivesNegativeTTL|RecordRemoteNegativeEscalatesAtThreshold|RecordRemoteNegativeWindowExpiryResets|DeferEscalationCoolsDown|CanAnswerMisses)'
```

GitHub check rollup from `gh pr view` showed CI, drive9-server-local e2e smoke, and CodeRabbit as successful. Approved.

PR #539, opened 2026-06-11
Reland persistent SSE event replay

srstack:
Restores persistent SSE event replay by reverting the revert from mem9-ai/drive9#538.

This PR reintroduces the durable `fs_events` log, persistent SSE replay/reset handling, chmod change events, event-stream metrics/dashboard updates, and the follow-up fixes added on this branch.

## Summary by CodeRabbit

* **New Features**
  * Durable filesystem event log with persistent SSE replay, retention sweeps, and broader event emission (chmod, delete, rename, create, write, copy, hardlink, upload)
  * Configurable SSE heartbeat, active-connection metrics, typed reset/replay reasons, client heartbeat handling, and improved SSE parsing (multiline data & comments)
  * Grafana event-stream dashboard and README entry

* **Documentation**
  * Cache invalidation spec updated to reflect chmod producing change events

* **Tests**
  * Extensive tests for persistence, replay, retention, SSE behavior, chmod persistence, and env parsing


qiffang:
## Field data from validating #537 on EC2 (relevant to this re-land)

While benchmarking #537 against prod (`api.drive9.ai`, fio: 10k files × 2MB, 8 jobs), we accidentally captured a clean A/B of #534's behavior across the #538 revert deploy:

| Client binary | Server | SSE resets during run |
|---|---|---|
| `6e2820e` (no #534 client changes) | Jun 10 (with #534) | 1 (initial_sync) |
| `415f66f` (**with** #534 client changes) | with #534 | **8389 × `seq_too_old`** in 17 min |
| `415f66f` (same binary) | post-#538-revert (restarted 02:55:10) | 1 (initial_sync) |

The cascade ran at 4–10 resets/sec for the entire write phase. Each reset = full snapshot of ~8900 inodes + kernel `NOTIFY_INVAL` storm. Measured impact on the same client binary: write throughput 336MiB/s vs 2098MiB/s (6×), debug log 19.8GB vs 29MB. The client's seq was advancing normally (17 → 13290) yet every resume was rejected — and the cascade stopped at the exact second the reverted server restarted.

## Root-cause suspects

1. **Dual seq domains**: durable log seq vs in-memory fanout seq, reconciled via fallback logic. The incident pattern (healthy client, every resume rejected) points here.
2. The commit *"Force reset on out-of-order durable SSE events"* suggests non-monotonic durable IDs were observed and papered over with resets. TiDB `AUTO_INCREMENT` allocates in batches and is not monotonic across nodes by default (`AUTO_ID_CACHE=1` or an explicit sequence fixes this). Reset is the most expensive recovery path in the system — using it as a steady-state patch self-amplifies under load, which is exactly what we measured.

## Position: persistence itself is fine — the complexity around it is not

Correctness is unaffected by persistence, and the marginal cost is small *if* the event row is written in the **same transaction** as the mutation it describes (~200B extra row inside a txn that already pays 2PC; measured mutation avg is 77ms, so likely <5% and invisible interactively). We support re-landing **if simplified**:

1. **Same-transaction event write; delete the volatile/in-memory fallback path entirely.** The fallback exists for "store unavailable" — but if the store is down, no mutation can commit, so there is nothing to publish. It optimizes a failure mode that cannot occur, while introducing the dual-domain failure mode that did occur.
2. **Single seq domain** = the event log's ID, guaranteed monotonic (`AUTO_ID_CACHE=1` / sequence). The in-memory ring buffer becomes a pure read cache of the log.
3. **Resets must be rare and cheap**: client-side coalescing/dedup (8389 sequential snapshots should collapse to single digits), never a steady-state path.
4. **Short retention is sufficient.** For a cache-invalidation stream, events older than the max cache TTL are no-ops by construction — minutes of retention captures all the value; long retention only adds storage/sweeper complexity.

Guiding principle: correctness first; don't introduce complexity to solve a ~1% performance scenario — each clever fallback is a new place to have an incident.

Suggested regression test before merge: sustained self-mutation burst (≥1000 events/s, e.g. 10k file creates) with an active SSE subscriber, asserting zero `seq_too_old` resets.

Repro artifacts (mount debug logs, goroutine dump, perf-counter summaries) available on request.

srstack:
@qiffang thanks for the EC2 field data. I addressed the reland blockers from this report in `7e52dea`:

- removed the gapless-seq assumption for durable events, so TiDB/MySQL `AUTO_INCREMENT` gaps no longer trigger steady-state `seq_too_old` resets;
- changed `PublishEvent` so a durable seq behind an already-advanced volatile/live cursor is renumbered for live delivery instead of emitting force-reset storms;
- fixed reconnect fallback when the durable log has no usable history but the in-memory ring still has the relevant events;
- fixed the `headSeq == 0` initial-reset live path so the first live event after a fresh tenant is not dropped;
- added regression coverage for durable seq gaps, volatile fallback collision, persistent replay gaps, persistent-empty/in-memory fallback, and zero-cursor live replay.

I did not fold the larger simplification into this PR, specifically making event writes part of each mutation transaction and deleting the volatile fallback path entirely. That is still the cleaner long-term direction, but for this re-land the immediate incident mode is blocked: durable ID gaps and mixed durable/volatile cursor advancement should no longer turn into repeated full-reset cascades.

srstack (commented):
## Code Review: Reland persistent SSE event replay

### Overview
Relands the durable `fs_events` log and persistent SSE replay (reverts the revert from #538). Adds configurable SSE heartbeat interval and catchup options via environment variables, wires them into both `drive9-server` and `drive9-server-local` entrypoints, and ships a new Grafana event-stream dashboard. The underlying SSE/event logic (34 files, 3023 additions) was previously reviewed before the revert; this review focuses on the entrypoint and configuration changes visible in the diff.

---

### Strengths

- **Clean env-var validation in `sseCatchupOptionsFromEnv`**: positive-integer checks for all tunable fields, with early return when catchup is disabled.
- **`envPositiveInt` is correctly strict**: distinguishes unset/empty (returns fallback) from set-to-invalid (returns error). This is better than silent fallback for misconfigured deployments.
- **Both entrypoints are symmetric**: `drive9-server` and `drive9-server-local` receive identical SSE config, so local dev behavior matches production.
- **Grafana dashboard panels** cover the full observability surface: active connections, heartbeat interval, churn, stream duration, replay metrics, reset reasons, persistence errors, retention sweep. Well-scoped.

---

### Issues

**1. `sseCatchupOptionsFromEnv` is copy-pasted verbatim into both `cmd/drive9-server/main.go` and `cmd/drive9-server-local/main.go`**

The function bodies are identical (confirmed from diff). Same for `envPositiveInt`. These should live in a shared internal package (e.g. `internal/srvenv`) or at minimum in a shared file imported by both commands. As-is, any future change to default values, env var names, or validation logic must be applied in two places, and the test coverage (`TestEnvPositiveInt`) is also duplicated.

**2. Validation in `sseCatchupOptionsFromEnv` can never trigger for `PollInterval`, `IdleMaxInterval`, `BatchSize`, or `MaxConcurrentTenants`**

```go
PollInterval: time.Duration(envInt("DRIVE9_SSE_CATCHUP_POLL_INTERVAL_MS", 1000)) * time.Millisecond,
...
if opts.PollInterval <= 0 {
    return opts, fmt.Errorf(...)
}
```

`envInt` silently falls back to the default (1000, 10000, 1000, 16) when the env var is missing, empty, or non-numeric. It only returns the parsed value when it's a valid integer — but it doesn't validate sign. So `DRIVE9_SSE_CATCHUP_POLL_INTERVAL_MS=0` or `=-1` would reach the `<= 0` check and return an error, which is correct. However, `DRIVE9_SSE_CATCHUP_POLL_INTERVAL_MS=abc` would silently use the fallback (1000ms) and never fail validation. The validation gives a false sense of coverage for the non-numeric case. Either use `envPositiveInt` for these fields too, or add a comment clarifying the non-numeric fallback behavior.

**3. SSE catchup options validation runs but `envInt` has already been called with the silent fallback**

The `opts` struct is built by calling `envInt` with fallbacks, then validated. If `DRIVE9_SSE_CATCHUP_POLL_INTERVAL_MS` is set to `0`, `envInt` returns `0` (it parsed successfully), and then the `<= 0` check fires correctly. But if it's set to `-5`, `envInt` also returns `-5`, and the validation fires correctly. So the validation *does* work for the numeric-but-invalid case. The gap is the non-numeric case only. Still worth fixing for operator experience.

---

### Minor

- The `TestEnvPositiveInt` test is identical in both `cmd/drive9-server/main_test.go` and `cmd/drive9-server-local/main_test.go` — another consequence of the duplication. Move to a shared location.
- `DRIVE9_SSE_DURABLE_CATCHUP` defaults to `true` in local server, which enables DB catchup polling in local dev environments. Confirm this doesn't add unexpected load to single-process local dev runs where multi-replica fanout is irrelevant; consider defaulting to `false` for `drive9-server-local` specifically.
- The Grafana dashboard JSON hardcodes `"id": null` which is the correct import convention. No issues.

---

### Verdict

The SSE logic itself was already reviewed. The entrypoint changes are correct but the verbatim duplication of `sseCatchupOptionsFromEnv` and `envPositiveInt` across both server commands is a maintenance liability that should be addressed, ideally in this PR or as an immediate follow-up.

srstack (commented):
## Code Review: Reland persistent SSE event replay

### Overview
Relands the durable fs_events log and persistent SSE replay (reverts the revert from #538). Adds configurable SSE heartbeat interval and catchup options via environment variables, wires them into both drive9-server and drive9-server-local entrypoints, and ships a new Grafana event-stream dashboard.

---

### Strengths

- Clean env-var validation in sseCatchupOptionsFromEnv: positive-integer checks for all tunable fields, early return when catchup is disabled.
- envPositiveInt is correctly strict: distinguishes unset/empty (returns fallback) from set-to-invalid (returns error).
- Both entrypoints are symmetric: drive9-server and drive9-server-local receive identical SSE config.
- Grafana dashboard covers the full observability surface: active connections, heartbeat interval, churn, stream duration, replay metrics, reset reasons, persistence errors, retention sweep.

---

### Issues

**1. sseCatchupOptionsFromEnv is copy-pasted verbatim into both cmd/drive9-server/main.go and cmd/drive9-server-local/main.go**

The function bodies and envPositiveInt are identical in both files. Any future change to defaults, env var names, or validation logic must be applied in two places. Move to a shared internal package (e.g. internal/srvenv) or a shared file imported by both commands. The TestEnvPositiveInt test is also duplicated.

**2. Validation in sseCatchupOptionsFromEnv can never trigger for the non-numeric case**

envInt silently falls back to the default when the env var is missing, empty, or non-numeric. So DRIVE9_SSE_CATCHUP_POLL_INTERVAL_MS=abc silently uses 1000ms and passes validation. The <= 0 check only catches the numeric-but-invalid case (0, -1). Either use envPositiveInt for these fields too, or add a comment clarifying the non-numeric fallback behavior so operators aren't confused when invalid values are silently ignored.

**3. DRIVE9_SSE_DURABLE_CATCHUP defaults to true in drive9-server-local**

This enables DB catchup polling in local dev environments where multi-replica fanout is irrelevant. Consider defaulting to false for drive9-server-local specifically to avoid unexpected load on single-process local dev runs.

---

### Minor

- The Grafana dashboard JSON hardcodes "id": null which is the correct import convention. No issues.
- DRIVE9_SSE_DURABLE_CATCHUP env var name uses "DURABLE" while the options struct field is Disabled -- the inversion (enabled=true maps to Disabled=false) is correct but the naming asymmetry is a minor readability issue.

---

### Verdict

The SSE logic itself was previously reviewed. The entrypoint changes are correct but the verbatim duplication of sseCatchupOptionsFromEnv and envPositiveInt across both server commands is a maintenance liability that should be addressed in this PR or as an immediate follow-up.

PR #540, opened 2026-06-11
fix: randomize semantic worker tenant pick to stop claim starvation

qiffang:
## Summary
- `claimTenantSlot` kept a persistent round-robin cursor taken modulo the *current* tenant page length. Keyset pages rotate with different lengths (128/128/54 at 310 active tenants in prod), so each pick on the short last page clamped the cursor back into `[0..53]`, locking the picker into a small orbit synchronized with page rotation. Simulation shows only ~54 of 310 tenants ever receive `ClaimSemanticTask` attempts in steady state; the rest are starved indefinitely (restart does not help — the orbit is deterministic from `rr=0`).
- Confirmed in prod (2026-06-11): a tenant's `img_extract_text` task sat `queued` with `attempt_count=0` for 15+ minutes while pod logs showed ~5 claims/sec against other tenants and the tenant's TiDB `statements_summary` showed the claim query had never executed there. Also explains historical multi-hour extraction delays (2026-05-23).
- Fix: pick a uniform random start index per call, which visits every tenant with equal probability regardless of page geometry. Removed the now-unused `rr` cursor.

## Test plan
- [x] New regression test `TestClaimTenantSlotCoversAllTenantsAcrossUnevenRotatingPages` simulates 128/128/54 rotating pages and asserts all 310 tenants get picked (deterministically fails on the old cursor logic)
- [x] `go test ./pkg/server/` passes
- [ ] After deploy: verify the stuck prod task `01KTTCB36QHK86YGCN765MQBQY` (tenant `2c89e856-...`) gets claimed and the image's `content_text` is populated

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tenant scheduling to prevent starvation by walking tenants in a randomized-start wrap-around order and enforcing per-tenant concurrency checks, ensuring fair access across uneven tenant pages.

* **Tests**
  * Added tests validating full tenant coverage across uneven page rotations and that the scheduler skips failing tenants while continuing to select healthy ones.

### DIFF
--- a/pkg/server/semantic_worker.go
+++ b/pkg/server/semantic_worker.go
@@ -565,10 +565,10 @@ func (m *semanticWorkerManager) nextTarget(ctx context.Context) (*semanticTarget
 		return nil, nil
 	}
 
-	for i := 0; i < len(refs); i++ {
-		ref, ok := m.claimTenantSlot(refs)
-		if !ok {
-			return nil, nil
+	for _, idx := range pageWalkOrder(len(refs)) {
+		ref := refs[idx]
+		if !m.tryClaimTenantSlot(ref.id) {
+			continue
 		}
 		target, err := m.targetForRef(ctx, ref)
 		if err != nil {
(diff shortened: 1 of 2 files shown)


mornyx (changes requested):
I found one blocking scheduling regression in the updated picker.

`claimTenantSlot` now chooses a fresh random start on every call. However, `nextTarget` calls `claimTenantSlot(refs)` up to `len(refs)` times and releases the slot when `targetForRef` fails or when the resolved target has no allowed task types. With the new implementation, those retries can keep selecting the same released-but-bad tenant, exhausting the loop without ever trying other tenants on the same page.

Example: a page has two tenants, A and B. A consistently fails `targetForRef` because its backend cannot be acquired, while B is healthy and has queued semantic work. Each retry has a 50% chance to pick A again, so 25% of `nextTarget` calls can return nil after picking A twice and never attempting B. With larger pages this becomes probabilistic rather than deterministic coverage. The previous round-robin cursor advanced past the failed ref, so the loop actually tried distinct tenants in a page.

I think the fix should keep the important part of this PR, but move the randomness to the page traversal rather than each individual claim: choose one random start index for the current `refs` page, then walk the page deterministically once while claiming/releasing specific refs. Alternatively, keep an attempted set or return the picked index so `nextTarget` cannot retry the same tenant within the same scan. A regression test with one consistently failing tenant and one healthy tenant would lock this behavior down.

The starvation bug described by this PR is real, but this version trades it for partial-outage flakiness in the retry path, so I would not approve it as-is.

Remote checks are green at review time: CI, drive9-server-local e2e smoke, and CodeRabbit all succeeded. I did not run local tests for this pass.

qiffang:
@mornyx Thanks — you're right, the per-claim random start regressed the retry path. With a fresh `rand.Intn` on every `claimTenantSlot` call, a released-but-failing tenant could be re-picked within the same `nextTarget` pass, so coverage of a page became probabilistic (your 2-tenant example: 25% of passes returned nil without ever attempting the healthy tenant).

Fixed in 2fbffd5, following your first suggestion:

- `claimTenantSlot` is gone. `nextTarget` now gets the visit order up front from a new `pageWalkOrder(n)`: one uniformly random start per page traversal, then a deterministic wrap-around walk of all indices. Each tenant is claimed/released individually (`tryClaimTenantSlot`), and is attempted **at most once per pass** — a tenant whose backend fails to open can no longer absorb retries meant for its neighbors.
- The random start per traversal keeps the original starvation fix intact: no cursor persists across calls, so nothing gets clamped by the short last page of the rotating keyset scan.

Tests added/updated:

- `TestSemanticWorkerNextTargetSkipsFailingTenantWithinPage` — the regression test you asked for: one tenant with an unreachable DB (port 1, sorts first in the page) plus one healthy tenant. Asserts 30 consecutive `nextTarget` calls all return the healthy tenant's target. Under the previous per-claim random picker this fails with probability ≈ 1 − 0.75³⁰ ≈ 99.98%.
- `TestPageWalkOrderCoversAllTenantsAcrossUnevenRotatingPages` — rewritten to target `pageWalkOrder`: asserts every order is a complete wrap-around permutation (so each tenant is attempted at most once per pass) and that first-picked tenants cover all 310 tenants across 200k ticks of rotating 128/128/54 pages.

`go test ./pkg/server/ -count=1` is green locally.

mornyx (approved):
Re-reviewed the latest commit 2fbffd5.

The previous blocking issue is addressed. `nextTarget` now chooses a single random start for the current tenant page via `pageWalkOrder`, then walks the page deterministically once. That preserves the intended starvation fix while ensuring a tenant whose backend fails to open cannot be selected repeatedly within the same pass and prevent healthy tenants on the same page from being attempted.

The added regression coverage matches the failure mode from the prior review: one failing tenant and one healthy tenant in the same page, with repeated `nextTarget` calls required to reach the healthy tenant. The page-walk test also verifies the wrap-around permutation property, which is the key invariant here.

One non-blocking note: this is still page-level fairness rather than globally uniform tenant fairness. Because keyset pages are visited equally, tenants on a shorter final page will get a higher per-tenant first-pick probability than tenants on full pages. That is acceptable for this fix since it removes permanent starvation and keeps the worker making progress, but it is worth keeping in mind if we later want stricter global fairness.

GitHub CI and drive9-server-local e2e smoke are passing at review time; CodeRabbit is still pending. I did not run local tests for this pass.

Approved.

PR #549, opened 2026-06-12
perf: fetch small files whole in one request on FUSE cold reads

qiffang:
## Summary

Cold-read benchmarking on the EC2 fio box (1000 × 2MB files, 4-way concurrency) showed the FUSE read path is 2-3x slower than the bare API for small files, entirely due to request amplification in the daemon — not server/network:

- Per 100 cold 2MB files (perf-counters): **707 remote range reads** @ avg 64ms / 384KB each (~7 requests per file), **278MB fetched for 200MB delivered** (33% over-read from prefetch fragments + block splits), plus 1 remote stat per file.
- fio cold read: **14.8 MB/s, avg 331ms/file, p99 826ms**; the bare API does the same file in one GET (~150ms) and scales linearly to 36 MB/s at 4-way.

The whole-file single-request path already existed but was gated at 1MiB, so 2MB files fell into the block-split + prefetch path.

### Changes

1. **Raise the whole-file read threshold from 1MiB to 4MiB** (`--read-cache-max-file-mb` default, `defaultReadCacheMaxFileSize`). Files ≤ threshold are fetched with one GET via `readSmallFileWithRetry` and served from the ReadCache; the prefetcher gate moves with the same constant, so these files no longer spawn fragment prefetches.
2. **Persist whole-file reads into the disk read cache tier** (`pkg/fuse/dat9fs.go` small-file branch). Previously files ≤ threshold only lived in the in-memory ReadCache (30s TTL, LRU) — without this, raising the threshold would regress warm reads of 1-4MiB files to remote re-fetches. Now: RAM → disk tier → remote, with the same revision-keyed invalidation as block entries. `PutAsync` length-validates against the stat size, so a stale stat cannot persist a truncated entry.

Expected cold-read cost per 2MB file drops from ~70ms stat + ~7 range GETs to ~70ms stat + 1 GET.

### Tests

- `TestReadTwoMiBFileUsesSingleWholeFileRequest`: a 2MiB cold read issues exactly one GET with no Range header and no prefetcher.
- `TestReadWholeFilePopulatesDiskReadCacheTier`: after ReadCache invalidation, the next read is served from the disk tier with no extra remote fetch.
- Two prefetcher tests updated to size files relative to `defaultReadCacheMaxFileSize` instead of a hardcoded 2MiB.

## Test plan

- [x] `go test ./pkg/fuse/ ./cmd/drive9/cli/` full suites green
- [ ] fio cold-read re-run on the EC2 bench box vs the 14.8 MB/s baseline (box SSH currently blocked by SG; binary built and ready)

## Summary by CodeRabbit

* **New Features**
  * Added a disk-level read cache tier for small files to reduce remote fetches after cache misses.
  * Increased the default per-file read-cache limit from 1 MB to 4 MB.

* **Tests**
  * Added tests validating read-cache behavior around the legacy threshold and that whole-file reads populate the disk read cache.

### DIFF
--- a/cmd/drive9/cli/mount.go
+++ b/cmd/drive9/cli/mount.go
@@ -111,7 +111,7 @@ func fsMountCmdWithBackground(args []string, background bool) error {
 	foreground := fs.Bool("foreground", false, "run in the foreground and block until unmounted")
 	cacheDir := fs.String("cache-dir", "", "write-back cache directory (default ~/.cache/drive9)")
 	cacheSize := fs.Int("cache-size", 128, "read cache size in MB")
-	readCacheMaxFile := fs.Int64("read-cache-max-file-mb", 1, "maximum single file size admitted to read cache in MB")
+	readCacheMaxFile := fs.Int64("read-cache-max-file-mb", 4, "maximum single file size admitted to read cache in MB; files at or below this size are fetched with a single whole-file request")
 	diskReadCacheSize := fs.Int64("disk-read-cache-size-mb", 1024, "disk-backed read cache size in MB")
 	diskReadCacheFreeRatio := fs.Float64("disk-read-cache-free-ratio", 0.10, "minimum filesystem free-space ratio before disk read cache evicts")
 	dirTTL := fs.Duration("dir-ttl", 10*time.Second, "directory cache TTL")
--- a/pkg/fuse/dat9fs.go
+++ b/pkg/fuse/dat9fs.go
@@ -8003,12 +8003,31 @@ func (fs *Dat9FS) Read(cancel <-chan struct{}, input *gofuse.ReadIn, buf []byte)
 			// Whole-file disk cache tier: survives ReadCache TTL/eviction
 			// so warm reads of small files are served from local disk
 			// instead of re-fetching from the remote.
+			//
+			// When TrustLocalEvents is false (default), revalidate the
+			// inode revision via HEAD before serving a disk cache hit.
+			// Without this check a stale revision from a prior session
+			// would match a persisted disk entry and serve outdated bytes
+			// after another client updates the file remotely — the same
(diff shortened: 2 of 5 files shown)


qiffang (commented):
REVISE — there is still a stale whole-file disk-cache correctness blocker.

The latest fix revalidates before serving a whole-file disk-cache hit, but the revision comparison does not work: `revalidateReadCacheEntryIfUntrusted(cancel, p, entry)` mutates the same `*InodeEntry` via `updateEntryFromStat`, then returns that same pointer. At `pkg/fuse/dat9fs.go:8020`, `refreshed.Revision != entry.Revision` is therefore false even when HEAD advanced from the cached revision to a newer revision. The branch falls through and serves `cached` bytes from the old disk entry, then repopulates the memory read cache with stale data.

I verified with a temporary focused regression using a 2MiB file (within the new whole-file threshold): disk cache had revision 7 bytes, HEAD returned revision 8 and fresh bytes, and `Read` returned stale `ssssssss` with OK instead of fetching fresh data. Existing range-cache tests pass because the range path recomputes the disk key after revalidation; the new whole-file path does not.

This directly answers the small-file question: the intended optimization is good, but in the default untrusted-local-events mode it can harm correctness for <=4MiB files by serving stale disk-cache bytes after another client updates the file. Please keep the old revision/size before revalidation (or recompute the whole-file disk key after HEAD like the range path) and add a regression for whole-file disk cache revision mismatch/refetch.

Local checks:
- `go test ./pkg/fuse -run 'TestReadTwoMiBFileUsesSingleWholeFileRequest|TestReadWholeFilePopulatesDiskReadCacheTier|TestDat9FSDiskReadCache(UnverifiedRevisionMismatchRefetches|RevalidatesWhenLocalEventsUntrusted)' -count=1` ✅
- temporary whole-file stale-regression test failed as described above

PR #550, opened 2026-06-13
feat(cli): add update command

mornyx:
## Summary

Adds a first-class `drive9 update` command that can replace the currently running drive9 CLI binary in place with the latest published release artifact.

This PR is intentionally scoped to the explicit update command. Automatic update notifications, prompt throttling, and background update-check state are left to PR #535.

## User-facing behavior

- `drive9 update` downloads the latest platform-specific binary from `https://drive9.ai/releases/drive9-<goos>-<goarch>`.
- `drive9 update --check` reports whether the installed CLI is behind the published release without modifying the binary.
- `drive9 update --force` reinstalls the latest binary and is required for downgrades or versions that cannot be compared as semver.
- `drive9 update --base-url <url>` supports alternate HTTPS release roots for controlled rollout environments.
- Non-HTTPS release roots are rejected by default because the checksum source must be authenticated.
- Windows self-update is explicitly unsupported for now because the running `.exe` cannot be replaced safely in-process.

## Implementation details

- Reuses the existing release artifact convention:
  - `/releases/version`
  - `/releases/checksums.txt`
  - `/releases/drive9-<goos>-<goarch>[.exe]`
- Uses semver-aware comparison so default update behavior only installs newer releases.
- Downloads into a temporary file next to the target executable, validates SHA-256 from `checksums.txt`, preserves the executable mode, then swaps the binary into place on Unix.
- Runs platform replacement preflight before downloading so unsupported platforms fail fast.
- Keeps Unix and Windows replacement logic split behind build tags.

## Safety and failure behavior

- Explicit `drive9 update` fails loudly if version metadata, checksums, download, validation, or replacement fails.
- The current binary is not replaced until the downloaded artifact checksum matches the published checksum.
- If the installed binary is not writable, the command returns a clear permission-oriented error instead of attempting fallback behavior.
- Older published releases are reported as downgrades and are not installed unless `--force` is provided.

## Tests

Targeted validation run:

```bash
env GOCACHE=/Users/mornyx/Desktop/repos/github.com/mem9-ai/drive9/.codex-gocache go test ./cmd/drive9 ./cmd/drive9/cli
env GOCACHE=/Users/mornyx/Desktop/repos/github.com/mem9-ai/drive9/.codex-gocache GOOS=windows GOARCH=amd64 go test -c ./cmd/drive9/cli -o /tmp/drive9-cli.test.exe
env GOCACHE=/Users/mornyx/Desktop/repos/github.com/mem9-ai/drive9/.codex-gocache go vet ./cmd/drive9/cli ./cmd/drive9
env GOCACHE=/Users/mornyx/Desktop/repos/github.com/mem9-ai/drive9/.codex-gocache go test ./cmd/drive9/cli -race -count=1
make lint
git diff --check
```

Covered cases include:

- top-level `update` dispatch
- global help including `update`
- `update --check` reporting an available release
- successful download, checksum verification, and replacement
- checksum mismatch refusing replacement
- semver downgrade blocking unless `--force`
- non-HTTPS release root rejection by default
- replacement preflight running before artifact download
- checksum file validation for duplicate or path-qualified artifact names

## Summary by CodeRabbit

* **New Features**
  * Added `drive9 update` command to self-update the binary
  * Supports `--check` flag to report available updates
  * Supports `--force` flag to override version constraints
  * Supports `--base-url` flag to customize release source
  * Automatic checksum verification on download (Unix/Linux/macOS supported; Windows requires manual installation)

### DIFF
--- a/cmd/drive9/cli/update.go
+++ b/cmd/drive9/cli/update.go
@@ -143,7 +143,7 @@ func updateWithDeps(args []string, deps updateDeps) error {
 	}()
 
 	if err := deps.replaceExecutable(tmpPath, target); err != nil {
-		return fmt.Errorf("replace %s: %w\nIf this binary is owned by another user, rerun with sufficient permissions or install drive9 into a user-writable directory.", target, err)
+		return fmt.Errorf("replace %s: %w; rerun with sufficient permissions or install drive9 into a user-writable directory", target, err)
 	}
 	installed = true
 
--- a/cmd/drive9/cli/update.go
+++ b/cmd/drive9/cli/update.go
@@ -9,6 +9,7 @@ import (
 	"fmt"
 	"io"
 	"net/http"
+	"net/url"
 	"os"
 	"path/filepath"
 	"runtime"
@@ -26,35 +27,21 @@ const (
 )
 
 type updateDeps struct {
-	baseURL           string
-	currentVersion    string
-	goos              string
-	goarch            string
-	executable        func() (string, error)
-	httpClient        *http.Client
-	stdout            io.Writer
-	stderr            io.Writer
-	preflightReplace  func() error
-	replaceExecutable func(string, string) error
+	baseURL              string
+	currentVersion       string
+	goos                 string
+	goarch               string
+	executable           func() (string, error)
+	httpClient           *http.Client
+	stdout               io.Writer
+	stderr               io.Writer
+	allowInsecureBaseURL bool
+	preflightReplace     func() error
+	replaceExecutable    func(string, string) error
 }
 
 func defaultUpdateDeps() updateDeps {
-	baseURL := strings.TrimSpace(os.Getenv("DRIVE9_UPDATE_BASE_URL"))
-	if baseURL == "" {
-		baseURL = defaultUpdateBaseURL
-	}
-	return updateDeps{
-		baseURL:           baseURL,
-		currentVersion:    buildinfo.Version,
-		goos:              runtime.GOOS,
-		goarch:            runtime.GOARCH,
-		executable:        os.Executable,
-		httpClient:        &http.Client{Timeout: updateCommandTimeout},
-		stdout:            os.Stdout,
-		stderr:            os.Stderr,
-		preflightReplace:  preflightReplaceExecutableFile,
-		replaceExecutable: replaceExecutableFile,
-	}
+	return fillUpdateDeps(updateDeps{})
 }
 
 // Update updates the running drive9 CLI binary in place.
@@ -85,6 +72,9 @@ func updateWithDeps(args []string, deps updateDeps) error {
 		return fmt.Errorf("unexpected argument %q\n%s", fs.Arg(0), updateUsage())
 	}
 	deps.baseURL = normalizeUpdateBaseURL(*baseURL)
+	if err := validateUpdateBaseURL(deps.baseURL, deps.allowInsecureBaseURL); err != nil {
+		return err
+	}
 
 	ctx, cancel := context.WithTimeout(context.Background(), updateCommandTimeout)
 	defer cancel()
@@ -446,11 +436,28 @@ func parseChecksums(data []byte) (map[string]string, error) {
 		if _, err := hex.DecodeString(sum); err != nil {
 			return nil, fmt.Errorf("invalid checksum for %s: %w", fields[1], err)
 		}
-		name := strings.TrimPrefix(fields[1], "*")
+		name, err := checksumArtifactName(fields[1])
+		if err != nil {
+			return nil, err
+		}
+		if _, exists := checksums[name]; exists {
+			return nil, fmt.Errorf("duplicate checksum for %s", name)
+		}
 		checksums[name] = sum
-		checksums[filepath.Base(name)] = sum
 	}
 	return checksums, nil
+}
+
+func checksumArtifactName(raw string) (string, error) {
+	name := strings.TrimPrefix(raw, "*")
+	if name == "" {
+		return "", errors.New("empty checksum artifact name")
+	}
+	clean := filepath.ToSlash(filepath.Clean(name))
+	if clean == "." || strings.HasPrefix(clean, "/") || strings.HasPrefix(clean, "../") || strings.Contains(clean, "/") {
+		return "", fmt.Errorf("checksum artifact %q must be a release artifact filename", raw)
+	}
+	return clean, nil
 }
 
 func downloadUpdateBinary(ctx context.Context, deps updateDeps, url, targetPath, wantSHA string) (string, error) {
@@ -538,6 +545,17 @@ func updateArtifactName(goos, goarch string) string {
 
 func normalizeUpdateBaseURL(baseURL string) string {
 	return strings.TrimRight(strings.TrimSpace(baseURL), "/")
+}
+
+func validateUpdateBaseURL(baseURL string, allowInsecure bool) error {
+	parsed, err := url.Parse(baseURL)
+	if err != nil || parsed.Scheme == "" || parsed.Host == "" {
+		return fmt.Errorf("invalid release base URL %q", baseURL)
+	}
+	if parsed.Scheme == "https" || allowInsecure {
+		return nil
+	}
+	return fmt.Errorf("release base URL %q must use https", baseURL)
 }
 
 func releaseURL(baseURL, name string) string {
--- a/cmd/drive9/cli/update.go
+++ b/cmd/drive9/cli/update.go
@@ -1,0 +1,519 @@
+package cli
+
+import (
+	"context"
+	"crypto/sha256"
+	"encoding/hex"
+	"encoding/json"
+	"errors"
+	"flag"
+	"fmt"
+	"io"
+	"net/http"
+	"os"
+	"path/filepath"
+	"runtime"
+	"strings"
+	"time"
+
+	"github.com/mem9-ai/dat9/pkg/buildinfo"
+)
+
+const (
+	defaultUpdateBaseURL = "https://drive9.ai"
+	updateCheckInterval  = 24 * time.Hour
+	autoUpdateTimeout    = 1500 * time.Millisecond
+	updateCommandTimeout = 2 * time.Minute
+	maxUpdateFileBytes   = 200 << 20
+)
+
+type updateDeps struct {
+	baseURL           string
+	currentVersion    string
+	goos              string
+	goarch            string
+	now               func() time.Time
+	executable        func() (string, error)
+	httpClient        *http.Client
+	stdout            io.Writer
+	stderr            io.Writer
+	replaceExecutable func(string, string) error
+}
+
+type updateCache struct {
+	LastCheckedAt time.Time `json:"last_checked_at"`
+	LatestVersion string    `json:"latest_version,omitempty"`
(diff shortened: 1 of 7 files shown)


qiffang (changes requested):
## PR #550 R1 Review — REVISE

I agree with the direction of a first-class `drive9 update` command. The explicit update command is useful, and the download/checksum/replace flow has good DI test coverage. The blocking issues are in the automatic notification layer and its overlap with PR #535.

### Blocking Findings

1. **Auto update notifications run in scripts/CI and can block command exit**

`main.go` calls `maybeNotifyUpdate` synchronously after `dispatch` (`cmd/drive9/main.go:82-83`) and again from `fatal` before exit (`cmd/drive9/main.go:341-347`). That path can perform a network request with `autoUpdateTimeout = 1500ms`, and it does not check whether stderr is a TTY or whether the process is running under CI.

This means ordinary commands can take an extra 1.5s to exit when the release endpoint is slow, and scripts/CI/non-interactive invocations can get update banners on stderr. It also means failure paths can delay exit just to check for updates. Update checks must be best-effort and non-disruptive: TTY-only, CI-skipped, command-filtered, and non-blocking with respect to the user command's completion.

2. **Cached notices are printed on every eligible command**

`maybeNotifyUpdateWithDeps` prints the cached notice whenever the cached version differs (`cmd/drive9/cli/update.go:172-193`). The cache only throttles network fetches via `LastCheckedAt`; it does not throttle user prompts. Once an update is cached, every subsequent command in the 24h cache window prints the banner again.

This reintroduces the notification-spam problem already fixed in PR #535. The state needs a prompt-level throttle such as `LastPromptedAt` + `LastPromptedVersion`, or this PR should reuse #535's `MarkUpdatePrompted` contract.

3. **Version comparison treats any different string as an upgrade**

`versionsDiffer` only trims a leading `v` and compares strings (`cmd/drive9/cli/update.go:298-304`). `drive9 update --check` and the install path both use that to decide that an update is available. If the local CLI is `v0.9.0` and `/releases/version` returns `v0.8.0` after a rollback, this command advertises an update and installs a downgrade by default.

Default behavior should only install newer semver releases. Downgrades should require `--force` or be reported explicitly as a downgrade, not as an available update.

4. **This duplicates and conflicts with PR #535's update-check infrastructure**

PR #535 is already green and implements the update notification foundation: semver-aware release checks, SHA-current handling, JSON/plain-text endpoint support, TTY/CI/excluded-command guards, async/non-blocking notification, prompt throttling, and release workflow support for semver tags.

PR #550 adds another cache schema and another notification pipeline around the same `update-check.json` concept, but without those safeguards. These two PRs should not merge independently. The cleaner path is: merge #535 first, rebase #550, delete `MaybeNotifyUpdate` / duplicate cache / `versionsDiffer`, and keep PR #550 focused on the explicit `drive9 update` command that downloads, verifies, and replaces the executable.

### Non-blocking Notes

- `writeUpdateCache` uses direct `os.WriteFile`; the update cache is best-effort, but temp-file + rename is cheap and avoids corrupt partial JSON.
- Windows currently downloads/verifies before failing at replacement. If Windows self-update is unsupported, fail fast and document it in command help.

### Validation

I validated the current head `d3ac11e06f1e79a23cae47345334d6b332c4f2ff` with:

- `/usr/local/go/bin/go test ./cmd/drive9/cli -race -count=1`
- `/usr/local/go/bin/go vet ./cmd/drive9/cli ./cmd/drive9`
- `PATH=/usr/local/go/bin:$PATH make lint`

Remote CI, local e2e smoke, and CodeRabbit are green. The blocker is semantic/UX correctness, not mechanical test failure.

mornyx:
Thanks for the R1 review. I re-scoped this PR to avoid duplicating the update-check infrastructure from #535.

Changes pushed in `773abdc`:

- Removed the automatic update notification hook from `cmd/drive9/main.go`, including the fatal-path check. PR #550 no longer performs background/synchronous update checks after ordinary commands.
- Removed the duplicate update notification cache/prompt pipeline from `cmd/drive9/cli/update.go` (`MaybeNotifyUpdate`, cache schema, prompt-on-every-command behavior). The notification foundation should stay in #535.
- Made the explicit `drive9 update` command semver-aware. By default it installs only newer semver releases; older releases and incomparable versions are reported explicitly and require `--force` to install.
- Added tests for downgrade blocking, forced downgrade installs, semver relation handling, and replacement preflight behavior.
- Windows self-update now fails before artifact download via a platform preflight, and the Windows help text documents that self-update is unsupported there.

Validation run:

- `go test ./cmd/drive9 ./cmd/drive9/cli`
- `GOOS=windows GOARCH=amd64 go test -c ./cmd/drive9/cli -o /tmp/drive9-cli.test.exe`
- `go vet ./cmd/drive9/cli ./cmd/drive9`
- `go test ./cmd/drive9/cli -race -count=1`
- `make lint`
- `git diff --check`

srstack (commented):
## Code Review: feat(cli): add update command

### Overview
Adds a first-class `drive9 update` command with SHA-256-verified in-place binary replacement. The implementation is well-structured and production-quality.

---

### Strengths

- **Dependency injection via `updateDeps`** makes the code highly testable without mocking file I/O or HTTP.
- **Atomic replacement on Unix**: temp file written next to the target (same filesystem), then `os.Rename` — no partial-write window.
- **Checksum validation before replacement**: the binary is never swapped in until SHA-256 matches.
- **Downgrade protection** with a clear `--force` escape hatch.
- **Preflight check before download**: avoids wasting bandwidth on platforms where replacement will fail (Windows).
- **Test coverage** is thorough: checksum mismatch, downgrade rejection, preflight ordering, version comparison edge cases.

---

### Issues

**1. Auto-update check described in PR summary is missing from the diff**

The PR description says:
> "The CLI refreshes a local update cache when the previous check is stale, and if a newer version is already known, the next `drive9` command prints a short reminder at the end"
> "Stores update-check state in `~/.drive9/update-check.json`"

The diff only contains the explicit `drive9 update` command — no background cache, no `update-check.json`, and no reminder emitted by normal commands. The tests listed as "covered" ("cached update notice appearing on the next run", "dev builds suppressing update reminders") don't appear in `update_test.go`. Either the auto-check feature was dropped and the PR description wasn't updated, or it lives in files not included in this diff.

Please clarify or update the PR summary accordingly.

**2. `fillUpdateDeps` duplicates `defaultUpdateDeps`**

`defaultUpdateDeps()` already populates every field. `fillUpdateDeps` then re-checks and re-sets each field unconditionally. The pattern is useful for tests that pass partial structs, but the zero-value check logic in `fillUpdateDeps` could be the single source of truth — `defaultUpdateDeps` can just call `fillUpdateDeps({}))`. As written, any change to a default must be applied in two places.

**3. Checksum map stores both full path and basename, silent collision risk**

```go
checksums[name] = sum
checksums[filepath.Base(name)] = sum
```

If two artifact paths share the same basename (unlikely but possible as the platform matrix grows), the second silently overwrites the first. Consider storing only the canonicalized artifact name used by `updateArtifactName`.

**4. No HTTPS enforcement on `DRIVE9_UPDATE_BASE_URL`**

The env var intentionally supports alternate release roots, but there is no check that the URL scheme is `https://`. A misconfigured or adversarially-set `http://` URL would download and install an unencrypted binary. The SHA-256 check does protect against tampering, but only if `checksums.txt` itself is also fetched over TLS. Consider rejecting non-HTTPS base URLs unless an explicit testing flag is set.

---

### Minor

- `parseSemverNumber` reimplements digit validation that `strconv.Atoi` already covers — the manual rune loop adds no value.
- `normalizeUpdateBaseURL` is called both in `fillUpdateDeps` and again after `fs.Parse`. The second call is fine, but worth a comment to clarify intent.
- `updateCommandTimeout` (2 min) is used both as the HTTP client timeout and the top-level context timeout — the context will always win, making the client-level timeout redundant. Consider using a shorter client timeout for the version/checksum fetches and a longer one only for the binary download.

---

### Verdict

The core download-verify-replace flow is solid. The two blocking items are: (1) clarify whether the auto-update background check was intentionally deferred, and (2) the HTTPS enforcement gap for `DRIVE9_UPDATE_BASE_URL`. The rest are minor polish.

qiffang (changes requested):
Blocking finding:

- `cmd/drive9/cli/update.go:87` / `cmd/drive9/cli/update.go:381` still accepts any `DRIVE9_UPDATE_BASE_URL` / `--base-url` scheme and then fetches `version`, `checksums.txt`, and the binary from that same root. For a self-update command, the checksum only helps if the checksum source is authenticated; with `http://...`, a MITM can serve a matching malicious `checksums.txt` and binary and the updater will install it. Please reject non-HTTPS release roots by default. If local/dev release servers are needed for tests or controlled rollout, make that an explicit opt-in such as an internal test dependency flag, localhost-only exception, or a clearly named insecure override rather than silently accepting arbitrary HTTP.

Verified resolved from earlier review threads: the automatic update notification/cache path is gone from `main.go`, downgrade handling now blocks older releases unless `--force`, and Windows replacement now fails fast via a sentinel unsupported error before download. CI is green, so this remaining blocker is a self-update trust issue rather than a mechanical test failure.

mornyx:
Thanks @qiffang and @srstack. I addressed the remaining review items in `9afd44c` and updated the PR description so it no longer claims the auto-update notification/cache behavior that was removed from this PR.

What changed:

- Release roots are now HTTPS-only by default. `DRIVE9_UPDATE_BASE_URL` and `--base-url` are validated before any network request; non-HTTPS URLs fail before fetching `version`, `checksums.txt`, or the binary. Tests use an internal `updateDeps.allowInsecureBaseURL` opt-in for `httptest` only.
- `defaultUpdateDeps` now delegates to `fillUpdateDeps(updateDeps{})`, so default values have one source of truth.
- `parseChecksums` no longer stores both the raw checksum path and `filepath.Base(...)`. It now accepts only release artifact filenames and rejects duplicate artifact entries, avoiding silent basename overwrites.
- Added regression coverage for insecure base URL rejection, path-qualified checksum entries, and duplicate checksum entries.
- Updated the PR body to clarify that PR #550 is scoped to the explicit `drive9 update` command; automatic update notifications remain deferred to #535.

I left the timeout split and `parseSemverNumber` shape unchanged for now. The timeout behavior is not part of the remaining blocker, and the manual digit check intentionally rejects signed numeric identifiers before `strconv.Atoi`.

Validation run:

- `go test ./cmd/drive9 ./cmd/drive9/cli`
- `GOOS=windows GOARCH=amd64 go test -c ./cmd/drive9/cli -o /tmp/drive9-cli.test.exe`
- `go vet ./cmd/drive9/cli ./cmd/drive9`
- `go test ./cmd/drive9/cli -race -count=1`
- `make lint`
- `git diff --check`

qiffang (approved):
Re-reviewed current head 871c1f1b. My previous HTTPS/trust blocker is fixed: update roots are validated before any network fetch, non-HTTPS is rejected by default, and the insecure path is only available through internal test dependency injection. The earlier semantic blockers are also resolved: no automatic update notification/cache path remains in main.go, downgrade installs require --force, and Windows self-update fails fast with a sentinel unsupported error before download.\n\nCI and local e2e checks are green. Remaining CodeRabbit note is a test-style nit (t.Fatalf vs t.Errorf) and not a merge blocker from my side.

PR #554, opened 2026-06-15
feat: sanitize client errors and add ctx ls --details flag

srstack:
### Client error sanitization
Previously, backend errors (TiDB/MySQL error codes, internal prefixes like `open datastore: open db:`) leaked directly to CLI users. This PR introduces `sanitizeClientError` to map errors to product-level messages:

| Condition | Client sees |
|---|---|
| TiDB usage quota exhausted | `tenant usage quota exceeded` |
| Metadata schema mismatch / migration failure | `tenant metadata error` |
| All other backend errors | `backend unavailable` |

Applied to all HTTP 500 error paths across `auth.go`, `server.go`, `fs_layer.go`, `vault.go`, `git_workspace.go` (29 call sites). Full errors remain in server logs.

### `ctx ls --details` flag
Added `-D`/`--details` flag to `drive9 ctx ls`:

- **Default view** keeps existing concise columns
- **`--details`** adds SERVER, AGENT, GRANT_ID, TENANT_ID for full context info
- TENANT_ID moved from default to `--details` view only

## Summary by CodeRabbit

* **New Features**
  * Added `--details` to `drive9 ctx ls` to show an expanded table with extra metadata columns (tenant id, server, agent, grant id) and improved per-row placeholders.
* **Improvements**
  * Internal server error responses are now sanitized into more user-friendly messages across filesystem, auth, git workspace, and vault endpoints.
  * Added structured server-side error logging for failed filesystem/vault operations.
  * Updated `ctx ls` help text; `--details` is mutually exclusive with `--json`.
* **Tests**
  * Updated context list tests and added coverage for error message sanitization mappings.

### DIFF
--- a/cmd/drive9/cli/ctx.go
+++ b/cmd/drive9/cli/ctx.go
@@ -753,7 +753,7 @@ func ctxListCmd(args []string) error {
 				typeFilter = strings.TrimPrefix(args[i], "--type=")
 				continue
 			}
-			return fmt.Errorf("unknown flag %q\nusage: drive9 ctx ls [-l|--json|--details] [--type <kind>|--scoped]", args[i])
+			return fmt.Errorf("unknown flag %q\nusage: drive9 ctx ls [-l|--long] [-D|--details] [--json] [--type <kind>|--scoped]", args[i])
 		}
 	}
 	if longForm && asJSON {
--- a/cmd/drive9/cli/ctx.go
+++ b/cmd/drive9/cli/ctx.go
@@ -727,6 +727,7 @@ func scopeRootSegment(scope string) string {
 func ctxListCmd(args []string) error {
 	longForm := false
 	asJSON := false
+	details := false
 	typeFilter := ""
 	for i := 0; i < len(args); i++ {
 		switch args[i] {
@@ -734,6 +735,8 @@ func ctxListCmd(args []string) error {
 			longForm = true
 		case "--json":
 			asJSON = true
+		case "-D", "--details":
+			details = true
 		case "--scoped":
 			// Shorthand for --type fs_scoped. The user mental model
 			// is "show me the scoped tokens" — accept the shortcut
@@ -750,11 +753,14 @@ func ctxListCmd(args []string) error {
 				typeFilter = strings.TrimPrefix(args[i], "--type=")
 				continue
 			}
-			return fmt.Errorf("unknown flag %q\nusage: drive9 ctx ls [-l|--json] [--type <kind>|--scoped]", args[i])
+			return fmt.Errorf("unknown flag %q\nusage: drive9 ctx ls [-l|--json|--details] [--type <kind>|--scoped]", args[i])
 		}
 	}
 	if longForm && asJSON {
 		return fmt.Errorf("-l/--long and --json are mutually exclusive")
+	}
+	if details && asJSON {
+		return fmt.Errorf("-D/--details and --json are mutually exclusive")
 	}
 	if typeFilter != "" {
 		if !isKnownPrincipalType(typeFilter) {
@@ -765,7 +771,7 @@ func ctxListCmd(args []string) error {
 	if asJSON {
 		return writeCtxListJSON(cfg, typeFilter)
 	}
-	return writeCtxListTable(cfg, longForm, typeFilter)
+	return writeCtxListTable(cfg, longForm, details, typeFilter)
 }
 
 // isKnownPrincipalType validates a --type filter value against the set
@@ -803,7 +809,7 @@ func writeCtxListJSON(cfg *Config, typeFilter string) error {
 	})
 }
 
-func writeCtxListTable(cfg *Config, longForm bool, typeFilter string) error {
+func writeCtxListTable(cfg *Config, longForm, details bool, typeFilter string) error {
 	entries := filterCtxListEntries(buildCtxListEntries(cfg), typeFilter)
 	if len(entries) == 0 {
 		if typeFilter != "" {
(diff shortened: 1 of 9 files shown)


srstack:
Resolved all review threads — see commits b0d363b and 4faf1d1.

qiffang (changes requested):
## PR #554 R1 Review — REVISE

The `ctx ls --details` change is straightforward and looks mechanically correct. The server-side sanitization direction is right, but the PR is not complete: there are still 500 paths in the same changed surface that return raw internal errors to clients.

### Blocking Finding

1. **FS layer commit failures still leak raw backend/datastore errors to clients**

The PR summary says sanitization is applied to all HTTP 500 paths across `fs_layer.go`, and the intended contract is that full errors stay in server logs while clients only see product-level messages. That contract is still violated in the FS layer commit path:

- `pkg/server/fs_layer.go:746` returns `fmt.Sprintf("apply fs layer entry %s: %v; rollback failed: %v", ...)`
- `pkg/server/fs_layer.go:749` returns `fmt.Sprintf("apply fs layer entry %s: %v", ...)`
- `pkg/server/fs_layer.go:758` returns `fmt.Sprintf("finalize fs layer commit %s: %v; rollback failed: %v", ...)`
- `pkg/server/fs_layer.go:761` returns `fmt.Sprintf("finalize fs layer commit %s: %v", ...)`

Those errors are produced from `applyFSLayerEntry`, `rollbackFSLayerCommit`, and `store.SetFSLayerStateIf`; they can wrap backend, TiDB/MySQL, migration/schema, quota, or storage implementation errors. This is exactly the class of error the PR is trying to hide. It also exposes layer IDs and paths in a 500 response while other changed paths now sanitize and log instead.

Fix requirement: log the full `err` / `rollbackErr` server-side, but return a sanitized/product-level client message for these 500s as well. If the client needs structured commit conflict details, those should stay on explicit 409 conflict responses, not raw 500 strings.

2. **Regression coverage should include the sanitizer contract**

There is no table-driven test for `sanitizeClientError`, and no handler-level regression proving a representative raw internal error is not returned to the client. Because this PR is an error-boundary/security/UX change, tests should lock the contract:

- TiDB quota-like input maps to `tenant usage quota exceeded`
- known schema/migration wrappers map to `tenant metadata error`
- unknown internal error maps to `backend unavailable`
- at least one server handler path returns the sanitized body instead of raw `err.Error()`

### Non-blocking Notes

- `cmd/drive9/main.go` help still advertises `ctx ls [-l|--json]` and does not mention `--details`; not a blocker, but it should be kept consistent with the new flag.
- `ctx ls --details` and `--json` mutual exclusion is correct, and `-l` + `--details` composition is useful.

### Validation

Validated head `b0d363bb66b30bf5cb6a806ddd86d8dbe95c034b` with:

- `/usr/local/go/bin/go test -race ./cmd/drive9/cli -run 'TestCtxList|TestF16|TestTokenList' -count=1`
- `/usr/local/go/bin/go vet ./cmd/drive9/cli ./pkg/server`
- `/usr/local/go/bin/go test -c ./pkg/server -o /tmp/drive9-pr554-server.test`
- `PATH=/usr/local/go/bin:$PATH /tmp/drive9-pr550-review/bin/golangci-lint run`

Full `go test ./pkg/server` cannot run on this machine because the server `TestMain` requires rootless Docker/testcontainers and panics locally with `rootless Docker not found`. Remote CI and local e2e smoke are green.

srstack:
@qiffang All feedback addressed in e7ff026:
- **Blocking 1**: fs_layer commit paths now sanitized — log full errors, return sanitizeClientError(err)
- **Blocking 2**: unit tests added for sanitizeClientError (nil, quota, metadata, fallback, case-insensitive)
- **Non-blocking**: main.go help text updated to include -D/--details

Ready for re-review.

qiffang (approved):
## PR #554 R2 Review — GREEN

R1 blockers are closed on `e7ff0265c52ae39d50f496d9970cfc0c903373e7`.

### Verified Fixes

- The four FS layer commit 500 paths no longer format `err` / `rollbackErr` into the HTTP response. They now log the full error server-side and return `sanitizeClientError(err)`.
- `sanitizeClientError` now lowercases before matching, so the quota detection is case-insensitive.
- Added table-driven-ish unit coverage for nil, quota, lower-case quota, wrapped quota, schema/migration wrappers, and unknown fallback.
- Top-level help now advertises `ctx ls [-l|--long] [-D|--details] [--json] [--type <kind>|--scoped]`.
- Re-scanned `pkg/server` for 500 response paths: changed backend/store paths now use sanitized or generic strings; no remaining `fs_layer.go` 500 response formats raw `err` values into the response body.

### Validation

- `/usr/local/go/bin/go test -race ./cmd/drive9/cli -run 'TestCtxList|TestF16|TestTokenList' -count=1`
- `/usr/local/go/bin/go vet ./cmd/drive9/cli ./cmd/drive9 ./pkg/server`
- `/usr/local/go/bin/go test -c ./pkg/server -o /tmp/drive9-pr554-server.test`
- `PATH=/usr/local/go/bin:$PATH /tmp/drive9-pr550-review/bin/golangci-lint run`

Full `go test ./pkg/server` still cannot run locally because this machine lacks rootless Docker/testcontainers support, but the server package compiles and remote CI/e2e are running.

Approved conditional on the pending GitHub checks finishing green.

mornyx (changes requested):
I found one blocking issue.

`pkg/server/auth.go:389-392` now sends `sanitizeClientError(err)` from the capability-token middleware when `pool.Acquire` fails. This path is still pre-verification: it only peeks `tenant_id` from the capability/grant token payload, and `handleVaultRead` does the actual signature/TTL/revocation validation later. The surrounding comment already calls out that this path should avoid leaking tenant existence or status to unauthenticated callers.

With the new mapping, a caller who can craft a token payload for an active tenant can distinguish quota exhaustion from schema/migration failures before proving that the token is valid. That turns tenant backend state into a pre-auth side channel.

Please keep this middleware on a uniform response, e.g. the previous `backend unavailable`, or restructure the flow so richer sanitized errors are only returned after the capability token has been verified.

srstack:
@mornyx Fixed in 36318af: capabilityAuthMiddleware now returns uniform  to avoid leaking tenant backend state pre-auth. Full error is logged server-side.

mornyx (approved):
Re-reviewed the latest head (`36318af`).

The previous blocker is fixed: `capabilityAuthMiddleware` now logs the backend load failure server-side and returns the uniform `backend unavailable` response before the capability token is fully verified. That removes the pre-auth quota/schema side channel I flagged in the prior review.

I do not see any remaining blocking issues.

One non-blocking cleanup remains: `ctxUsage()` in `cmd/drive9/cli/ctx.go` still documents `ls [-l|--json] [--type <kind>|--scoped]`, so `drive9 ctx --help` does not advertise the new `-D/--details` flag. The top-level usage and unknown-flag usage are updated, but this help text is still stale.

Validation:
- `git diff --check origin/pr/554-merge^1...origin/pr/554`
- `go test ./cmd/drive9/cli`
- `go test -c ./pkg/server -o /private/tmp/drive9-pr554.LOTifF/server.test`

Full `pkg/server` tests were not run because the package `TestMain` starts MySQL/Testcontainers. Approving.

PR #556, opened 2026-06-15
fix: patch upload respects presigned checksum headers

### LINKED ISSUE #555: Patch upload sends unsigned x-amz-checksum-sha256 header causing S3 403 on large files
## Problem

When a FUSE mount flushes a large file (or WAL file) that triggers the PatchFile partial-update path, S3 rejects the upload with HTTP 403:

```
AccessDenied: There were headers present in the request which were not signed
HeadersNotSigned: x-amz-checksum-sha256
```

This causes SQLite to see `disk I/O error` on `fdatasync()` for any write-heavy workload (speedtest1 --size 50, app-style 10k+ row WAL updates).

## Root Cause

`pkg/client/patch.go` line 213 unconditionally adds `x-amz-checksum-sha256` to the PUT request, but the server-side presign in `pkg/backend/patch.go` line 258 passes an empty checksum value, so the presigned URL signature does not cover this header. S3 correctly rejects with 403.

## Error Chain

```
SQLite disk I/O error
  <- fdatasync() returns EACCES
  <- FUSE httpToFuseStatus() maps HTTP 403 -> EACCES
  <- S3 multipart upload part returns 403
  <- S3: headers not signed: x-amz-checksum-sha256
```

## Fix

In `pkg/client/patch.go` uploadPatchPart: only set `x-amz-checksum-sha256` if the presigned headers from the server already include it.

## Required regression test

Must include a test verifying: when server-returned presigned headers do NOT contain `x-amz-checksum-sha256`, the patch upload does not send that header.

qiffang:
## Summary

- Fix `uploadPatchPart` sending unsigned `x-amz-checksum-sha256` header that S3 rejects with 403
- Only send the checksum header when the server's presigned headers include it
- Add regression test covering 3 cases: no checksum, with checksum, nil headers

## Root Cause (Issue #555)

`pkg/client/patch.go` unconditionally added `x-amz-checksum-sha256` to the S3 presigned PUT, but the server presigns patch parts with empty checksum (because the final part content isn't known at presign time). S3 correctly rejects the unsigned header.

**Error chain:** S3 403 -> FUSE EACCES -> fdatasync() failure -> SQLite "disk I/O error"

This blocked all large SQLite workloads on Drive9 FUSE mounts (speedtest1 --size 50, app-style 10k+ WAL updates).

## Changes

- `pkg/client/patch.go`: Only set `x-amz-checksum-sha256` when `part.Headers` includes the key
- `pkg/client/patch_test.go`: Updated existing test + added `TestPatchUploadRespectsPresignedChecksumHeader` with 3 subtests

## Test plan

- [ ] CI passes (unit tests for patch upload)
- [ ] After merge: rebuild drive9 CLI on EC2, re-run speedtest1 --size 50 and app-style 10k/50k WAL workloads
- [ ] Verify mount logs show no more S3 403 errors

Closes #555

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected multipart PATCH uploads so `x-amz-checksum-sha256` is only sent when the server/plan included it in presigned part headers; otherwise the SDK omits it.
  * Updated patch multipart initialization and presigned part URL behavior to use “no checksum” settings to avoid S3 validation/authorization issues.

* **Tests**
  * Extended and added regression tests across client SDKs to verify correct presence/absence of `x-amz-checksum-sha256` and proper checksum computation/propagation when presigned.

### DIFF
--- a/clients/drive9-js/src/patch.ts
+++ b/clients/drive9-js/src/patch.ts
@@ -63,14 +63,20 @@ async function uploadPatchPart(client: Client, part: PatchPartURL, readPart: Rea
     origData = new Uint8Array(await resp.arrayBuffer());
   }
   const data = readPart(part.number, part.size, origData);
-  const checksum = await sha256Base64(data);
-  const headers: Record<string, string> = { "x-amz-checksum-sha256": checksum };
+  const headers: Record<string, string> = {};
+  let shouldSendChecksum = false;
   if (part.headers) {
     for (const [k, v] of Object.entries(part.headers)) {
       if (typeof v === "string" && k.toLowerCase() !== "host") {
         headers[k] = v;
+        if (k.toLowerCase() === "x-amz-checksum-sha256") {
+          shouldSendChecksum = true;
+        }
       }
     }
+  }
+  if (shouldSendChecksum) {
+    headers["x-amz-checksum-sha256"] = await sha256Base64(data);
   }
   const resp = await fetch(part.url, { method: "PUT", headers, body: bodyInit(data) });
   await checkError(resp);
@@ -85,4 +91,3 @@ async function sha256Base64(data: Uint8Array): Promise<string> {
   const hash = await crypto.subtle.digest("SHA-256", bufferSource(data));
   return bytesToBase64(new Uint8Array(hash));
 }
-
--- a/clients/drive9-js/tests/transfer.test.ts
+++ b/clients/drive9-js/tests/transfer.test.ts
@@ -2,6 +2,7 @@ import { describe, expect, it } from "vitest";
 import { Client } from "../src/index.js";
 import { setupServer } from "msw/node";
 import { http, HttpResponse } from "msw";
+import { createHash } from "node:crypto";
 
 const server = setupServer();
 server.listen({ onUnhandledRequest: "error" });
@@ -71,5 +72,49 @@ describe("Transfer", () => {
       offset += c.length;
     }
     expect(new TextDecoder().decode(all)).toBe("world");
+  });
+
+  it("patchFile only sends checksum when it is presigned", async () => {
+    let firstChecksum: string | null = null;
+    let secondChecksum: string | null = null;
+    server.use(
+      http.patch("http://localhost:9009/v1/fs/file.bin", () => {
+        return HttpResponse.json(
+          {
+            upload_id: "patch-js",
+            part_size: 8,
+            upload_parts: [
+              { number: 1, url: "http://localhost:9009/patch/1", size: 8, headers: {} },
+              {
+                number: 2,
+                url: "http://localhost:9009/patch/2",
+                size: 8,
+                headers: { "x-amz-checksum-sha256": "placeholder" },
+              },
+            ],
+            copied_parts: [],
+          },
+          { status: 202 }
+        );
+      }),
+      http.put("http://localhost:9009/patch/1", ({ request }) => {
+        firstChecksum = request.headers.get("x-amz-checksum-sha256");
+        return HttpResponse.text("ok");
+      }),
+      http.put("http://localhost:9009/patch/2", ({ request }) => {
+        secondChecksum = request.headers.get("x-amz-checksum-sha256");
+        return HttpResponse.text("ok");
+      }),
+      http.post("http://localhost:9009/v1/uploads/patch-js/complete", () => {
+        return HttpResponse.text("ok");
+      })
+    );
+
+    const client = new Client("http://localhost:9009", "test-key");
+    await client.patchFile("/file.bin", 16, [1, 2], (part) => new TextEncoder().encode(`part-${part}`), undefined, 8);
+
+    const expectedSecond = createHash("sha256").update("part-2").digest("base64");
+    expect(firstChecksum).toBeNull();
+    expect(secondChecksum).toBe(expectedSecond);
   });
 });
--- a/clients/drive9-kotlin/lib/src/main/kotlin/com/drive9/mobile/Drive9.kt
+++ b/clients/drive9-kotlin/lib/src/main/kotlin/com/drive9/mobile/Drive9.kt
@@ -676,7 +676,9 @@ public class Drive9Client(
 
     private fun uploadPatchPart(part: Drive9PatchPartUrl, data: ByteArray) {
         val headers = part.headers.toMutableMap()
-        headers["x-amz-checksum-sha256"] = sha256Base64(data)
+        if (headers.keys.any { it.equals("x-amz-checksum-sha256", ignoreCase = true) }) {
+            headers["x-amz-checksum-sha256"] = sha256Base64(data)
+        }
         val result = rawPut(part.url, headers, data, retryOnForbidden = false)
         if (result.status !in 200..299) throw errorFrom(result.status, result.body)
     }
--- a/clients/drive9-kotlin/lib/src/test/kotlin/com/drive9/mobile/Drive9Test.kt
+++ b/clients/drive9-kotlin/lib/src/test/kotlin/com/drive9/mobile/Drive9Test.kt
@@ -6,12 +6,14 @@ import com.sun.net.httpserver.HttpServer
 import kotlinx.coroutines.flow.flow
 import kotlinx.coroutines.runBlocking
 import java.io.ByteArrayOutputStream
+import java.security.MessageDigest
 import java.nio.file.Files
 import kotlin.io.path.deleteIfExists
 import kotlin.io.path.exists
 import kotlin.io.path.readBytes
 import java.net.InetSocketAddress
 import java.nio.charset.StandardCharsets
+import java.util.Base64
 import kotlin.test.AfterTest
 import kotlin.test.BeforeTest
 import kotlin.test.Test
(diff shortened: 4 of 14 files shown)


qiffang:
## PR #556 — R1 Review (adversary-1)

**Verdict: GREEN, conditional on CI + post-merge EC2 benchmark rerun**

This fix matches the RCA for Issue #555. The previous failure was caused by `uploadPatchPart` adding `x-amz-checksum-sha256` even when the server had not signed that header in the S3 presigned URL. Real S3 then rejected the patch part upload with `HeadersNotSigned`, which propagated as FUSE `EACCES` and SQLite `disk I/O error` on WAL `fdatasync()`.

### What I checked

- `pkg/client/patch.go` now copies only server-provided presigned headers first, and only recomputes/sends `x-amz-checksum-sha256` when that header key is present in `part.Headers`.
- When the checksum header is absent, the client no longer adds an unsigned `x-amz-checksum-sha256`, so the exact S3 403 failure mode is removed.
- The change is scoped to patch upload. Direct PUT and v2 multipart upload paths are untouched.
- `pkg/s3client/aws.go` lowercases signed header keys via `flattenSignedHeaders`, so the lowercase lookup used by the client matches the current server contract.
- Regression coverage is adequate for the bug: no checksum header, checksum header present, and nil headers. The updated existing patch test now asserts the real S3 rule: if the server did not sign the checksum header, the client must not send it.

### Validation

- `go test -c ./pkg/client` passed.
- `go vet ./pkg/client` passed.
- Local `go test ./pkg/client -run '^TestPatch'` could not run on this machine because `pkg/client` TestMain starts testcontainers and rootless Docker is unavailable. Remote CI is still pending.

### Required follow-up before customer use

After merge/build, rerun the EC2 SQLite workloads that previously failed:

- `speedtest1 --size 50 --testset main --journal WAL --checkpoint`
- app-style WAL 10k / 50k insert-update-checkpoint workloads
- verify mount logs no longer contain S3 `403 HeadersNotSigned: x-amz-checksum-sha256`

No blockers found in the code diff.

qiffang:
## PR #556 — R2 Review (adversary-1)

**Verdict: REVISE**

The new server-side change is directionally correct, but the regression coverage is now incomplete for the actual root cause found on EC2.

### B1: Server-side checksum contract is not locked by tests

The EC2 validation proved this was not a client-only bug. The real contract spans three steps:

1. `CreateMultipartUpload` must not declare a checksum algorithm for patch uploads.
2. `PresignUploadPart` must not sign/require a checksum header for patch uploads.
3. `uploadPatchPart` must not send `x-amz-checksum-sha256` unless the server returned it in the presigned headers.

The PR now covers only step 3 with client tests. Steps 1 and 2 were just changed in `pkg/backend/patch.go`, but there is no backend/server regression asserting that patch upload uses `ChecksumAlgoNone` for both MPU creation and part presign.

This matters because the first version of this PR passed the client regression and still failed on EC2 with:

```text
400 InvalidRequest: Checksum Type mismatch occurred, expected checksum Type: sha256, actual checksum Type: null
```

That exact failure mode would not be prevented by the current test suite. A future change could reintroduce `ChecksumAlgoSHA256` in `CreateMultipartUpload` and the existing client tests would still pass.

Minimum acceptable regression:

- Add a backend test around `InitiatePatchUploadIfRevision` using a recording S3 client.
- Assert `CreateMultipartUpload` receives `s3client.ChecksumAlgoNone`.
- Assert every patch `PresignUploadPart` call receives `s3client.ChecksumAlgoNone` and an empty checksum value.
- Keep the existing client tests for header send behavior.

### Validation performed

- `go test -c ./pkg/client ./pkg/backend` passed.
- `go vet ./pkg/client ./pkg/backend` passed.
- Full local test execution is blocked by rootless Docker/testcontainers on this machine.

The implementation is probably the right fix, but this bug came from an untested cross-step protocol contract. That contract needs to be locked before merge.

qiffang:
## R2 → R3: Added server-side regression test

Commit: `test(backend): add regression test for patch upload checksum contract`

Addresses @adversary-1 R2 REVISE: the server-side checksum contract is now locked by `TestPatchUploadUsesChecksumAlgoNone`.

**Test design:**
- `patchChecksumRecordingS3Client` wraps the local S3 client and records `ChecksumAlgo` args
- Creates an S3-backed file via v1 upload + confirm (2 parts)
- Calls `InitiatePatchUploadIfRevision` with part 1 dirty
- Asserts `CreateMultipartUpload` → `ChecksumAlgoNone`
- Asserts all `PresignUploadPart` → `ChecksumAlgoNone` + empty checksum value

**Coverage now spans all three protocol steps:**
1. ✅ Server `CreateMultipartUpload` → `ChecksumAlgoNone` (backend test)
2. ✅ Server `PresignUploadPart` → `ChecksumAlgoNone` (backend test)
3. ✅ Client `uploadPatchPart` → conditional header (client test, 3 subtests)

If anyone reintroduces `ChecksumAlgoSHA256` in the patch path, the backend test will fail immediately.

qiffang:
## PR #556 — R3 Review (adversary-1)

**Verdict: GREEN**

R2 blocker is fixed. The new `TestPatchUploadUsesChecksumAlgoNone` closes the server-side regression gap I requested:

- It exercises `InitiatePatchUploadIfRevision` on a real S3-backed file path, not a mocked-only shortcut.
- It records `CreateMultipartUpload` and verifies patch MPU creation uses `s3client.ChecksumAlgoNone`.
- It records every patch `PresignUploadPart` and verifies `ChecksumAlgoNone` plus empty checksum value.
- Together with the client tests, the three-step contract is now locked: server does not declare checksum, server does not presign checksum, client does not send an unsigned checksum header.

Validation performed locally:

- `go test -c ./pkg/client ./pkg/backend` ✅
- `go vet ./pkg/client ./pkg/backend` ✅
- Full `go test ./pkg/client` / backend runtime tests are blocked on this machine by missing rootless Docker/testcontainers, so CI remains authoritative for runtime execution.

Current remote status: `ci` is green; `drive9-server-local e2e smoke` and CodeRabbit are still pending at review time.

Conditional on CI green and the planned EC2 end-to-end rerun after server deploy.

mornyx (changes requested):
I found one blocking issue.

This fixes the Go client/server path, but the same patch upload contract is implemented by the other generated/mobile clients and they still unconditionally attach `x-amz-checksum-sha256` to patch part PUTs:

- `clients/drive9-js/src/patch.ts`
- `clients/drive9-py/drive9/patch.py`
- `clients/drive9-rs/src/patch.rs`
- `clients/drive9-kotlin/lib/src/main/kotlin/com/drive9/mobile/Drive9.kt`
- `clients/drive9-swift/Sources/Drive9Mobile/Drive9.swift`

After this PR, the server-side patch path creates/presigns parts with `ChecksumAlgoNone`, so `part.headers` will not include `x-amz-checksum-sha256`. Those clients will still send an unsigned checksum header and will keep hitting the same S3 `HeadersNotSigned: x-amz-checksum-sha256` / 403 failure that this PR is meant to fix.

Please update all patch clients to follow the same rule as the Go client: copy presigned headers first, and only recompute/send `x-amz-checksum-sha256` when the presigned `part.headers` include that header. Please also add at least focused regression coverage for those clients, or remove/explicitly deprecate the affected patch implementations if they are no longer supported.

Validation I ran:
- `git diff --check origin/pr/556-merge^1...origin/pr/556`
- `go test ./pkg/client -run TestPatch`
- `go test -c ./pkg/backend -o /private/tmp/drive9-pr556.ao78lG/backend.test`

qiffang:
Addressed the review comments in commit `6be17e9d`:\n\n- Go client checksum regression now asserts the exact SHA-256 base64 value.\n- Backend checksum contract test no longer uses `testify/require`.\n- JS, Python, Rust, Kotlin, and Swift patch upload clients now only recompute/send `x-amz-checksum-sha256` when that header is present in the presigned part headers.\n- Added focused SDK regression coverage for the signed-vs-unsigned checksum header behavior.\n\nValidation: GitHub checks are green for CodeRabbit, server CI, local e2e smoke, and all SDK jobs.

mornyx (approved):
I re-reviewed the latest commit. The previous blocker is addressed: the Go client and the JS/Python/Rust/Kotlin/Swift patch clients now only recompute and send `x-amz-checksum-sha256` when the server included that header in the presigned part headers, and the server-side patch path now creates/presigns patch parts with `ChecksumAlgoNone`.

I did not find any new blocking issue. One minor hardening note: a few clients detect checksum headers case-insensitively but then add the lowercase key without deleting the original case variant. The current server canonicalizes signed headers to lowercase, so this is not a blocker for the present contract, but normalizing/replacing in-place would make the SDKs more robust if that ever changes.

Validation I ran:
- `git diff --check origin/pr/556-merge^1...origin/pr/556-merge`
- `go test -c ./pkg/backend -o /private/tmp/drive9-pr556-review.6p0iMa/backend.test`
- `cargo test test_patch_file_respects_presigned_checksum_header` in `clients/drive9-rs`

I also attempted `go test ./pkg/client -run TestPatch`, but this package TestMain requires Docker/Testcontainers in this environment, so it could not complete here. I did not run the JS/Python/Kotlin/Swift tests locally because the temporary review snapshot did not have those dependencies installed. Approving.

Issue #558, opened 2026-06-15
Shadow upload: use direct PUT for small files below inline_threshold


qiffang:
## Problem

`uploadFromShadowRemoteWithRevision` always uses `WriteMultipartStreamConditional` for non-empty files, regardless of size. This forces small files (e.g. 1KB) through the full multipart/S3 protocol (initiate → presign → upload part → complete = 4+ HTTP round trips), even though they should use direct PUT → server → TiDB inline (1 HTTP round trip).

This is the root cause of close-sync being ~2x slower than write-sync for small files:
- **write-sync 1KB**: dirty buffer → `WriteCtxConditionalWithRevision` (direct PUT, 1 HTTP) → TiDB inline. P50 ≈ 315ms
- **close-sync 1KB**: shadow → `WriteMultipartStreamConditional` (multipart, 4+ HTTP) → S3. P50 ≈ 668ms

Additionally, small files end up stored in S3 instead of TiDB inline, which violates the `inline_threshold` storage tiering design.

## Fix

In `uploadFromShadowRemoteWithRevision` (`pkg/fuse/remote_upload.go`), check `client.CachedSmallFileThreshold()`:

```go
if threshold := c.CachedSmallFileThreshold(); threshold > 0 && size < threshold {
    data, err := shadows.ReadFull(localPath)
    if err != nil { return 0, err }
    return c.WriteCtxConditionalWithRevision(ctx, remotePath, data, expectedRevision)
}
// large files continue with WriteMultipartStreamConditional
```

### Additional improvements (same PR)

1. **Fix `committedRev` cleanup branch**: Current shadow cleanup at L8602-8611 is inside `committedRev == 0` branch. Direct PUT returns `committedRev > 0`, so shadow/pending cleanup must be moved outside this branch to avoid residual pending entries.

2. **Add `readCache.Put` after shadow upload success**: Current shadow upload path doesn't populate read cache after success (unlike write-sync at L8530). This causes close→reopen→read to go remote unnecessarily.

## Scope

- `pkg/fuse/remote_upload.go`: Add small file direct PUT branch (~5-8 lines)
- `pkg/fuse/dat9fs.go`: Fix `committedRev` cleanup branch to always clean shadow/pending; add `readCache.Put`
- All shadow upload callers benefit: close-sync flush + write-back commit queue

## Tests Required

- Small shadow file → direct PUT, `storage_type=DB9/inline`
- Large shadow file → multipart, `storage_type=S3`  
[...]

PR #559, opened 2026-06-15
Fix shadow upload tier routing

qiffang:
## Summary
- route ShadowSpill uploads below the cached server inline threshold through direct PUT so they land in DB/TiDB inline storage
- keep threshold-unknown and size >= threshold shadow uploads on multipart/S3
- seed read cache from shadow data before successful synchronous shadow commits clear local shadow state

## Tests
- /usr/local/go/bin/go test ./pkg/fuse -run 'TestFsyncShadowSpillCommitRefreshesRevisionAndClearsPending|TestUploadFromShadowRemoteWithRevision|TestCommitQueueShadowSpillSmallFileDirectPUTCleansState|TestCommitQueueShadowSpillUpload|TestReleaseShadowSpillFallback' -count=1\n- /usr/local/go/bin/go test ./pkg/fuse -count=1\n- git diff --check\n\n## Notes\n- Backend tier transition focused tests could not run locally because testcontainers panics with: rootless Docker not found. Existing backend tier transition coverage remains unchanged.\n\nFixes #558

## Summary by CodeRabbit

* **Improvements**
  * Small files are now uploaded more efficiently using optimized direct PUT transfers instead of multipart streaming.
  * Read cache is automatically populated from shadow storage after successful uploads for improved performance.
  * Enhanced state cleanup and recovery mechanisms to ensure data consistency after commits.

* **Tests**
  * Added comprehensive test coverage for direct PUT uploads and small-file handling scenarios.
  * Expanded validation of revision tracking, request bodies, and state management across upload flows.

### DIFF
--- a/pkg/fuse/commit_queue_test.go
+++ b/pkg/fuse/commit_queue_test.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"net/http"
 	"net/http/httptest"
+	"strconv"
 	"strings"
 	"sync"
 	"sync/atomic"
@@ -1433,9 +1434,9 @@ func TestCommitQueueRecoverPendingSkipsLegacyOverwriteWithoutBaseRev(t *testing.
 
 // --- ShadowSpill tests ---
 
-// TestCommitQueueShadowSpillUpload verifies that ShadowSpill entries are
-// uploaded via streaming (uploadFromShadow) rather than ReadAll, and that
-// the server receives the correct data and expected revision.
+// TestCommitQueueShadowSpillUpload verifies that ShadowSpill entries at or
+// above the cached inline threshold are uploaded via streaming multipart and
+// that the server receives the correct data and expected revision.
 func TestCommitQueueShadowSpillUpload(t *testing.T) {
 	data := bytes.Repeat([]byte("shadowspill-data-"), 100) // ~1700 bytes
 	var gotExpected int64
@@ -1514,7 +1515,9 @@ func TestCommitQueueShadowSpillUpload(t *testing.T) {
 		t.Fatal(err)
 	}
 
-	cq := NewCommitQueue(newTestClient(ts.URL), shadow, pending, nil, 1, 8)
+	c := newTestClient(ts.URL)
+	c.SetSmallFileThresholdForTests(1)
+	cq := NewCommitQueue(c, shadow, pending, nil, 1, 8)
 	cq.OnSuccess = func(entry *CommitEntry, committedRev int64) {
 		if entry.Path != "/big.bin" {
 			t.Fatalf("OnSuccess path = %q, want /big.bin", entry.Path)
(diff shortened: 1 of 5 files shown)


qiffang:
## R1 GREEN ✅

**Core change (remote_upload.go):** Correctly routes small shadow files through direct PUT when `CachedSmallFileThreshold() > 0 && size < threshold`. Large files and unknown threshold continue multipart. Clean and minimal.

**readCache seed (dat9fs.go):** `seedReadCacheFromShadowLocked` reads shadow data *before* `markHandleRemoteCommittedLocked` deletes it — correct ordering. Called at all three commit paths (syncHandle, Flush, Fsync).

**Shadow cleanup:** `markHandleRemoteCommittedLocked` already handles shadow/pending/ShadowSpill cleanup when `committedRev > 0`, so the `committedRev` branching concern is a non-issue.

**Tests:** Comprehensive — small file direct PUT, exact threshold multipart, unknown threshold multipart, commit queue cleanup, readCache verification after fsync. Existing ShadowSpill tests correctly pinned to `SetSmallFileThresholdForTests(1)`.

Minor note (non-blocking): `PutOwned` vs `Put` — `ReadAll` returns a fresh copy so ownership transfer is safe.

LGTM — awaiting @adversary-1 review + CI.

qiffang:
## R1 GREEN (adversary-1)

No blocking findings.

What I verified:
- `uploadFromShadowRemoteWithRevision` now routes by final shadow size plus cached server inline threshold: `threshold > 0 && size < threshold` uses direct PUT, while `size >= threshold` and threshold-unknown keep multipart. This matches the DB-inline vs S3 storage-tier contract and does not fetch `/v1/status` on the hot path.
- The direct-PUT path uses the same CAS input (`expectedRevision`) and returns the committed revision, so subsequent FUSE state can be refreshed correctly.
- The earlier `committedRev > 0` cleanup concern is covered: handle paths call `markHandleRemoteCommittedLocked`, which removes shadow/pending and clears ShadowSpill flags; CommitQueue success calls `OnSuccess` before queue cleanup so readCache can seed from shadow before deletion.
- Tests cover small shadow direct PUT, exact-threshold multipart, threshold-unknown multipart, direct-PUT ShadowSpill commit cleanup, and readCache seeding after fsync.

Local validation:
- focused shadow upload / commit queue / fsync tests
- `go test -c ./pkg/client ./pkg/backend`
- `go vet ./pkg/fuse ./pkg/client ./pkg/backend`
- `git diff --check`

Residual conditions before merge:
- Remote `ci` and `drive9-server-local e2e smoke` must pass.
- After merge/deploy, run the EC2 comparison for single small write+close and multi-write close-sync vs write-sync to confirm the expected performance shape.

Note: I could not submit a formal GitHub approval because the local `gh` auth is the PR author's `qiffang` identity, and GitHub blocks self-approval.

mornyx (approved):
I reviewed the PR end-to-end and did not find any blocking issues.

The change correctly brings the ShadowSpill upload path in line with the existing non-ShadowSpill threshold routing: small shadow-backed files use direct PUT only when the server inline threshold is already cached, while threshold-unknown and size-at/above-threshold cases keep the multipart path. The direct PUT branch preserves the same `expectedRevision` CAS input and returns the committed revision, so the synchronous FUSE paths and commit queue can refresh local revision state correctly.

I also checked the read-cache seeding order. The new helper reads from shadow before `markHandleRemoteCommittedLocked` clears shadow/pending state, and the remote commit path lock prevents concurrent shadow writes during the upload/seed window. CommitQueue already seeds from shadow on `OnSuccess`, so the new helper is appropriately scoped to the synchronous shadow commit paths.

Non-blocking nit: the new direct PUT branch in `uploadFromShadowRemoteWithRevision` could wrap the `shadows.ReadAll` and `WriteCtxConditionalWithRevision` errors with operation/path context for easier diagnostics. This is not a correctness blocker.

Validation I ran:
- `git diff --check origin/pr/559-merge^1...origin/pr/559-merge`
- `go test ./pkg/fuse -run

mornyx:
Correction to my review validation list: the last bullet was truncated by shell quoting.

Validation I ran:
- `git diff --check origin/pr/559-merge^1...origin/pr/559-merge`
- `go test ./pkg/fuse -run TestFsyncShadowSpillCommitRefreshesRevisionAndClearsPending|TestUploadFromShadowRemoteWithRevision|TestCommitQueueShadowSpillSmallFileDirectPUTCleansState|TestCommitQueueShadowSpillUpload|TestReleaseShadowSpillFallback -count=1`

Both focused and full `pkg/fuse` tests passed.

srstack (commented):
## Review

The routing logic and cache seeding are correct overall. Two things to address before merge:

### 1. Missing threshold in `TestCommitQueueShadowSpillSmallFileDirectPUTCleansState`

This test expects the file to go via direct PUT, but it never calls `c.SetSmallFileThresholdForTests(...)`. With `CachedSmallFileThreshold()` returning 0, the guard `threshold > 0` is false and the upload falls through to multipart — hitting the `t.Fatalf("small ShadowSpill should not use multipart")` handler. Every other test in this diff that cares about routing explicitly sets the threshold. Either `newTestClient` already sets one (not visible in the diff, worth confirming), or this test needs:

```go
c := newTestClient(ts.URL)
c.SetSmallFileThresholdForTests(int64(len(data)) + 1)
cq := NewCommitQueue(c, shadow, pending, nil, 1, 8)
```

### 2. Disk I/O under lock in `seedReadCacheFromShadowLocked`

All three call sites (`syncHandleToRemoteLocked`, `Flush`, `Fsync`) invoke this while holding the fs lock. `shadows.ReadAll` does disk I/O — fine when shadow data is hot in the page cache, but a latency risk under memory pressure. If this ever becomes a problem, consider copying the shadow path out before releasing the lock and seeding the cache after. At minimum, a short comment on the assumption would help future readers.

---

Everything else looks good: boundary semantics (`size < threshold` strict, zero-byte early return), the unknown-threshold fallback, and the new read-cache assertions in `writeback_test.go` are all correct.

Issue #560, opened 2026-06-15
Writeback path truncate-to-zero local staging (W5 optimization Phase 1)


qiffang:
## Problem

In writeback/interactive mode, `open(path, "w")` triggers kernel `SETATTR size=0` (path-based truncate) which synchronously calls two remote HTTP round trips:

```
WriteCtx(nil)  → remote empty file commit  ≈ 172ms
StatCtx        → refresh revision           ≈ 37ms
Total ≈ 200ms/file
```

The subsequent `write + close` uploads the final content, making the empty file commit wasteful. This adds ~200ms/file to W5 (Code Edit R-M-W, 30 files), accounting for ~6s of the 6.4s total.

Post-PR #559 benchmark: W5 rmw p50 = 199ms/file, with SETATTR consuming ~200ms.

## Solution

For **writeback/interactive mode only**, change path-based `SETATTR size=0` (no `FATTR_FH`) to:

1. Stage locally: create/update empty shadow + pending marker (`PendingOverwrite`, size=0, baseRev=current)
2. Update inode size=0, clear read cache
3. Return OK immediately (no remote HTTP)
4. Async enqueue a writeback commit for the empty file (handles standalone truncate case)
5. If a subsequent `open("w") + write` arrives on the same path before the async commit completes, the write path supersedes/updates the pending entry — only the final content gets committed

### In scope
- `SETATTR size=0` without `FATTR_FH` (path-based truncate)
- WritePolicy == writeback (interactive/auto modes)
- Normal remote files only

### Out of scope (DO NOT touch)
- close-sync / write-sync / strict modes — these require synchronous remote semantics
- `size != 0` arbitrary truncate
- `FATTR_FH` ftruncate (already has local dirty path)
- Commit queue supersession/coalescing beyond same-path pending update
- rename/unlink/fsync flush behavior

## Invariants that MUST be preserved
- Standalone `truncate(path, 0)` must eventually commit empty file to remote
- `fsync/fdatasync` must respect durability mode
- rename/unlink must flush pending truncate before proceeding
- CAS baseRev must not be lost
- Crash recovery must be able to recover pending truncate state (shadow + pending index)

## Acceptance criteria
- [ ] W5 benchmark: SETATTR no longer blocks on remote HTTP in writeback mode
- [ ] Standalone `truncate(path, 0)` eventually commits empty file remotely
- [ ] `open("w") + write + close` only commits final content (no wasted empty file commit)
[...]

PR #561, opened 2026-06-15
fix(fuse): stage writeback path truncates locally

qiffang:
## Summary
- stage writeback path `SETATTR size=0` locally in shadow + pending index and enqueue the zero-byte commit asynchronously
- keep close-sync/write-sync/non-writeback paths on the existing synchronous remote truncate path
- allow a later writable open to cancel only queued, not in-flight, zero-byte truncate commits while preserving local shadow/pending state

## Scope / safety
- no fixed coalesce timer
- no in-flight commit cancellation or supersession
- no close-sync/write-sync semantic changes
- excludes layer paths, SQLite persistent journal sidecars, dirs, and symlinks

## Tests
- go test ./pkg/fuse -run 'TestSetAttr_WriteBackInteractivePathTruncateStagesAndReturnsBeforeRemoteCommit|TestSetAttr_CloseSyncPathTruncateKeepsSynchronousRemoteCommit|TestSetAttr_WriteBackStrictPathTruncateStagesAndReturnsBeforeRemoteCommit|TestOpenTruncateCancelsQueuedPathTruncateWithoutCancelingInFlight|TestOpenWritableCancelsQueuedPathTruncateWithoutOTruncFlag|TestCommitQueueCancelQueuedZeroTruncatePreservesLocalOnlyWhenNotInFlight|TestSetAttr_TruncateWithoutHandleRefreshesRevision' -count=1\n- go test ./pkg/fuse -count=1\n- go test ./cmd/drive9 ./cmd/drive9/cli ./pkg/fuse -count=1\n- go test -race ./pkg/fuse -run 'TestOpenWritableCancelsQueuedPathTruncateWithoutOTruncFlag|TestOpenTruncateCancelsQueuedPathTruncateWithoutCancelingInFlight|TestCommitQueueCancelQueuedZeroTruncatePreservesLocalOnlyWhenNotInFlight' -count=1\n- go vet ./pkg/fuse ./cmd/drive9 ./cmd/drive9/cli\n- git diff --check\n\nFixes #560

## Summary by CodeRabbit

* **New Features**
  * Added support for local zero-byte truncation staging on write-back mounts.

* **Bug Fixes**
  * Improved truncation handling and file mode preservation across multiple scenarios.

* **Tests**
  * Added comprehensive test coverage for zero-byte truncation workflows, including write-back mode interactions, mode preservation, and cancellation semantics.

### DIFF
--- a/e2e/fuse-smoke-test.sh
+++ b/e2e/fuse-smoke-test.sh
@@ -380,6 +380,15 @@ with open(path, "wb") as handle:
 PY
 }
 
+path_truncate_zero() {
+  local file_path="$1"
+  python3 - "$file_path" <<'PY'
+import os
+import sys
+os.truncate(sys.argv[1], 0)
+PY
+}
+
 wait_mount_state() {
   local expect="$1"
   local deadline=$(( $(date +%s) + MOUNT_READY_TIMEOUT_S ))
@@ -731,12 +740,25 @@ RW_HARDLINK_MOUNT="$MOUNT_POINT/$RW_HARDLINK_REL"
 RW_TEXT_RENAMED_REL="${RW_ALPHA_REL}/text-renamed.txt"
 RW_TEXT_RENAMED_REMOTE="/${RW_TEXT_RENAMED_REL}"
 RW_TEXT_RENAMED_MOUNT="$MOUNT_POINT/$RW_TEXT_RENAMED_REL"
+RW_PATH_TRUNC_REL="${RW_ALPHA_REL}/path-truncate.txt"
+RW_PATH_TRUNC_REMOTE="/${RW_PATH_TRUNC_REL}"
+RW_PATH_TRUNC_MOUNT="$MOUNT_POINT/$RW_PATH_TRUNC_REL"
+RW_PATH_TRUNC_RENAME_SRC_REL="${RW_ALPHA_REL}/path-truncate-rename-src.txt"
+RW_PATH_TRUNC_RENAME_SRC_REMOTE="/${RW_PATH_TRUNC_RENAME_SRC_REL}"
+RW_PATH_TRUNC_RENAME_SRC_MOUNT="$MOUNT_POINT/$RW_PATH_TRUNC_RENAME_SRC_REL"
+RW_PATH_TRUNC_RENAME_DST_REL="${RW_ALPHA_REL}/path-truncate-rename-dst.txt"
+RW_PATH_TRUNC_RENAME_DST_REMOTE="/${RW_PATH_TRUNC_RENAME_DST_REL}"
+RW_PATH_TRUNC_RENAME_DST_MOUNT="$MOUNT_POINT/$RW_PATH_TRUNC_RENAME_DST_REL"
+RW_PATH_TRUNC_UNLINK_REL="${RW_ALPHA_REL}/path-truncate-unlink.txt"
+RW_PATH_TRUNC_UNLINK_REMOTE="/${RW_PATH_TRUNC_UNLINK_REL}"
+RW_PATH_TRUNC_UNLINK_MOUNT="$MOUNT_POINT/$RW_PATH_TRUNC_UNLINK_REL"
 RW_ATTR_REL="${RW_ALPHA_REL}/attr.txt"
 RW_ATTR_REMOTE="/${RW_ATTR_REL}"
 RW_ATTR_MOUNT="$MOUNT_POINT/$RW_ATTR_REL"
 RW_ALPHA_RENAMED_REL="${ROOT_REL}/alpha-renamed"
 RW_ALPHA_RENAMED_REMOTE="/${RW_ALPHA_RENAMED_REL}"
 RW_ALPHA_RENAMED_MOUNT="$MOUNT_POINT/$RW_ALPHA_RENAMED_REL"
+RW_PATH_TRUNC_AFTER_RENAME_MOUNT="$MOUNT_POINT/${RW_ALPHA_RENAMED_REL}/path-truncate.txt"
 CLI_TO_MOUNT_REL="${RW_ALPHA_RENAMED_REL}/from-cli.txt"
 CLI_TO_MOUNT_REMOTE="/${CLI_TO_MOUNT_REL}"
 CLI_TO_MOUNT_MOUNT="$MOUNT_POINT/$CLI_TO_MOUNT_REL"
@@ -884,6 +906,47 @@ if is_mounted "$MOUNT_POINT"; then
     check_eq "truncate sets size to 0" "$truncated_size" "0"
   else
     check_eq "truncate via mount succeeds" "false" "true"
+  fi
+
+  printf "path-truncate-%s" "$TS" > "$RW_PATH_TRUNC_MOUNT"
+  path_truncate_seed=$(wait_remote_cat_eq "$RW_PATH_TRUNC_REMOTE" "path-truncate-${TS}")
+  check_eq "explicit path truncate seed visible via remote cat" "$path_truncate_seed" "path-truncate-${TS}"
+  if path_truncate_zero "$RW_PATH_TRUNC_MOUNT"; then
+    check_eq "explicit path truncate succeeds" "true" "true"
+    path_truncate_local=$(cat "$RW_PATH_TRUNC_MOUNT")
+    path_truncate_remote_size=$(wait_remote_stat_field_eq "$RW_PATH_TRUNC_REMOTE" "size" "0")
+    path_truncate_remote=$(wait_remote_cat_eq "$RW_PATH_TRUNC_REMOTE" "")
+    check_eq "explicit path truncate immediate mounted read is empty" "$path_truncate_local" ""
+    check_eq "explicit path truncate remote size is 0" "$path_truncate_remote_size" "0"
+    check_eq "explicit path truncate remote content is empty" "$path_truncate_remote" ""
+  else
+    check_eq "explicit path truncate succeeds" "false" "true"
+  fi
+
+  printf "path-truncate-rename-%s" "$TS" > "$RW_PATH_TRUNC_RENAME_SRC_MOUNT"
+  path_truncate_rename_seed=$(wait_remote_cat_eq "$RW_PATH_TRUNC_RENAME_SRC_REMOTE" "path-truncate-rename-${TS}")
+  check_eq "explicit path truncate rename seed visible via remote cat" "$path_truncate_rename_seed" "path-truncate-rename-${TS}"
+  if path_truncate_zero "$RW_PATH_TRUNC_RENAME_SRC_MOUNT" && mv "$RW_PATH_TRUNC_RENAME_SRC_MOUNT" "$RW_PATH_TRUNC_RENAME_DST_MOUNT"; then
+    check_eq "explicit path truncate then rename succeeds" "true" "true"
+    check_cmd_fail "explicit path truncate rename source missing locally" test -e "$RW_PATH_TRUNC_RENAME_SRC_MOUNT"
+    path_truncate_rename_size=$(wait_remote_stat_field_eq "$RW_PATH_TRUNC_RENAME_DST_REMOTE" "size" "0")
+    path_truncate_rename_remote=$(wait_remote_cat_eq "$RW_PATH_TRUNC_RENAME_DST_REMOTE" "")
+    check_cmd "explicit path truncate rename source missing remotely" wait_remote_ls_missing_name "$RW_ALPHA_REMOTE" "path-truncate-rename-src.txt"
+    check_eq "explicit path truncate rename destination remote size is 0" "$path_truncate_rename_size" "0"
+    check_eq "explicit path truncate rename destination remote content is empty" "$path_truncate_rename_remote" ""
+  else
+    check_eq "explicit path truncate then rename succeeds" "false" "true"
+  fi
+
+  printf "path-truncate-unlink-%s" "$TS" > "$RW_PATH_TRUNC_UNLINK_MOUNT"
+  path_truncate_unlink_seed=$(wait_remote_cat_eq "$RW_PATH_TRUNC_UNLINK_REMOTE" "path-truncate-unlink-${TS}")
+  check_eq "explicit path truncate unlink seed visible via remote cat" "$path_truncate_unlink_seed" "path-truncate-unlink-${TS}"
+  if path_truncate_zero "$RW_PATH_TRUNC_UNLINK_MOUNT" && rm -f "$RW_PATH_TRUNC_UNLINK_MOUNT"; then
+    check_eq "explicit path truncate then unlink succeeds" "true" "true"
+    check_cmd_fail "explicit path truncate unlink target missing locally" test -e "$RW_PATH_TRUNC_UNLINK_MOUNT"
+    check_cmd "explicit path truncate unlink target missing remotely" wait_remote_ls_missing_name "$RW_ALPHA_REMOTE" "path-truncate-unlink.txt"
+  else
+    check_eq "explicit path truncate then unlink succeeds" "false" "true"
   fi
 
   echo "[6] attribute semantics"
@@ -1119,6 +1182,13 @@ PY
         tier_remount_hash=""
       fi
       check_eq "tier transition final checksum survives remount" "$tier_remount_hash" "$tier_small_final_hash"
+      check_cmd "explicit path truncate file visible after remount" wait_path_exists "$RW_PATH_TRUNC_AFTER_RENAME_MOUNT"
+      if [ -f "$RW_PATH_TRUNC_AFTER_RENAME_MOUNT" ]; then
+        path_truncate_remount=$(cat "$RW_PATH_TRUNC_AFTER_RENAME_MOUNT")
+      else
+        path_truncate_remount="missing"
+      fi
+      check_eq "explicit path truncate content survives remount" "$path_truncate_remount" ""
     else
       echo "SKIP tier transition remount content checks because rw mount is unavailable"
       rw_mount_available=false
--- a/e2e/fuse-smoke-test.sh
+++ b/e2e/fuse-smoke-test.sh
@@ -486,6 +486,57 @@ PY
   done
 }
 
+assert_remote_ls_missing_stable() {
+  local parent="$1"
+  local name="$2"
+  local stable_s="${3:-3}"
+  local interval_s="${4:-0.25}"
+  local deadline
+  local out rc
+
+  wait_remote_ls_missing_name "$parent" "$name" || return 1
+  deadline=$(python3 - "$stable_s" <<'PY'
+import sys
+import time
+print(time.time() + float(sys.argv[1]))
+PY
+)
+
+  while :; do
+    set +e
+    out=$(drive9 fs ls "$parent" 2>&1)
+    rc=$?
+    set -e
+    if [ "$rc" -eq 0 ]; then
+      if python3 - "$out" "$name" <<'PY'
+import sys
+lines=[ln.strip() for ln in sys.argv[1].splitlines() if ln.strip()]
+name=sys.argv[2]
+raise SystemExit(0 if all(line != name and line != name + "/" for line in lines) else 1)
+PY
+      then
+        :
+      else
+        printf 'assert_remote_ls_missing_stable: %s reappeared under %s\n' "$name" "$parent" >&2
+        return 1
+      fi
+    elif [[ "$out" != *"not found"* && "$out" != *"Too Many Requests"* && "$out" != *"HTTP 429"* && "$out" != *"HTTP 403"* && "$out" != *"403 Forbidden"* ]]; then
+      printf '%s\n' "$out" >&2
+      return 1
+    fi
+
+    if python3 - "$deadline" <<'PY'
+import sys
+import time
+raise SystemExit(0 if time.time() >= float(sys.argv[1]) else 1)
+PY
+    then
+      return 0
+    fi
+    sleep "$interval_s"
+  done
+}
+
 wait_remote_cat_eq() {
   local path="$1"
   local want="$2"
@@ -931,7 +982,7 @@ if is_mounted "$MOUNT_POINT"; then
     check_cmd_fail "explicit path truncate rename source missing locally" test -e "$RW_PATH_TRUNC_RENAME_SRC_MOUNT"
     path_truncate_rename_size=$(wait_remote_stat_field_eq "$RW_PATH_TRUNC_RENAME_DST_REMOTE" "size" "0")
     path_truncate_rename_remote=$(wait_remote_cat_eq "$RW_PATH_TRUNC_RENAME_DST_REMOTE" "")
-    check_cmd "explicit path truncate rename source missing remotely" wait_remote_ls_missing_name "$RW_ALPHA_REMOTE" "path-truncate-rename-src.txt"
+    check_cmd "explicit path truncate rename source stays missing remotely" assert_remote_ls_missing_stable "$RW_ALPHA_REMOTE" "path-truncate-rename-src.txt"
     check_eq "explicit path truncate rename destination remote size is 0" "$path_truncate_rename_size" "0"
     check_eq "explicit path truncate rename destination remote content is empty" "$path_truncate_rename_remote" ""
   else
@@ -944,7 +995,7 @@ if is_mounted "$MOUNT_POINT"; then
   if path_truncate_zero "$RW_PATH_TRUNC_UNLINK_MOUNT" && rm -f "$RW_PATH_TRUNC_UNLINK_MOUNT"; then
     check_eq "explicit path truncate then unlink succeeds" "true" "true"
     check_cmd_fail "explicit path truncate unlink target missing locally" test -e "$RW_PATH_TRUNC_UNLINK_MOUNT"
-    check_cmd "explicit path truncate unlink target missing remotely" wait_remote_ls_missing_name "$RW_ALPHA_REMOTE" "path-truncate-unlink.txt"
+    check_cmd "explicit path truncate unlink target stays missing remotely" assert_remote_ls_missing_stable "$RW_ALPHA_REMOTE" "path-truncate-unlink.txt"
   else
     check_eq "explicit path truncate then unlink succeeds" "false" "true"
   fi
--- a/pkg/fuse/commit_queue.go
+++ b/pkg/fuse/commit_queue.go
@@ -397,6 +397,39 @@ func (cq *CommitQueue) CancelPathPreserveLocal(path string) {
 	cq.cancelPath(path, true)
 }
 
+// CancelQueuedZeroTruncatePreserveLocal cancels queued, not in-flight,
+// zero-byte overwrite entries for path without removing shadow/pending state.
+// It returns true only when no in-flight or other queued same-path entry remains.
+func (cq *CommitQueue) CancelQueuedZeroTruncatePreserveLocal(path string) bool {
+	if cq == nil || path == "" {
+		return true
+	}
+	cq.mu.Lock()
+	defer cq.mu.Unlock()
+	if cq.inFlight[path] != nil {
+		return false
+	}
+	otherQueued := false
+	remaining := cq.queue[:0]
+	for _, entry := range cq.queue {
+		if entry == nil || entry.Path != path {
+			remaining = append(remaining, entry)
+			continue
+		}
+		if isQueuedZeroTruncateEntry(entry) {
+			entry.canceled = true
+			continue
+		}
+		otherQueued = true
+		remaining = append(remaining, entry)
+	}
+	cq.queue = remaining
+	if cq.queuedByPath != nil {
+		cq.rebuildQueuedIndexLocked()
+	}
+	return !otherQueued
+}
+
 func (cq *CommitQueue) cancelPath(path string, preserveLocal bool) {
 	var cancels []context.CancelFunc
 	seen := make(map[*CommitEntry]struct{})
@@ -581,6 +614,10 @@ func (cq *CommitQueue) worker() {
 		// Clear in-flight after all cleanup is done.
 		cq.endInFlight(entry)
 	}
+}
+
+func isQueuedZeroTruncateEntry(entry *CommitEntry) bool {
+	return entry != nil && entry.Kind == PendingOverwrite && entry.Size == 0
 }
 
 func (cq *CommitQueue) beginInFlight(entry *CommitEntry) bool {
--- a/pkg/fuse/commit_queue_test.go
+++ b/pkg/fuse/commit_queue_test.go
@@ -2352,3 +2352,67 @@ func TestCommitQueueCancelPathJournalsCommitMarker(t *testing.T) {
 		t.Error("unlinked path resurrected from WAL — cancel marker missing")
 	}
 }
+
+func TestCommitQueueCancelQueuedZeroTruncatePreservesLocalOnlyWhenNotInFlight(t *testing.T) {
+	zero := &CommitEntry{Path: "/file.txt", Size: 0, Kind: PendingOverwrite}
+	nonzero := &CommitEntry{Path: "/other.txt", Size: 4, Kind: PendingOverwrite}
+	cq := &CommitQueue{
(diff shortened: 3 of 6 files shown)


srstack (commented):
## Code Review: fix(fuse): stage writeback path truncates locally

### Overview
Stages `SETATTR size=0` locally (shadow + pending index + async commit queue) for writeback-policy mounts, so the kernel's `truncate()` call returns before the remote zero-byte PUT completes. Close-sync/write-sync paths are deliberately left on the existing synchronous path. A subsequent writable `open()` can cancel a queued (not in-flight) zero-byte commit and adopt the already-staged local state rather than waiting for the commit to drain.

---

### Strengths

- **Clean eligibility guard**: `canStagePathTruncateToZero` centralizes all the exclusion conditions (dir, symlink, layer, SQLite journal, non-writeback policy, missing stores). Easy to extend.
- **Sync fallback on enqueue failure**: if the commit queue is full or closed, `commitNowPathLocked` runs synchronously rather than silently dropping the commit. Cleanup on fallback failure correctly removes both shadow and pending state before returning the error.
- **Cancel semantics are precise**: `CancelQueuedZeroTruncatePreserveLocal` distinguishes queued vs in-flight and returns the right boolean for each case. The `canSupersedeQueuedPathTruncate` → `CancelQueuedZeroTruncatePreserveLocal` flow avoids canceling in-flight commits that are already racing to the remote.
- **`loadWritableHandleFromShadowLocked`**: the `zeroOverwrite` path correctly calls `wb.Truncate(0)`, sets `fh.ZeroBase`, and marks `DirtySeq` so the write buffer starts from a known-zero baseline rather than an empty-but-uninitialized buffer.
- **Test coverage**: 7 targeted tests covering the full matrix of writeback/close-sync, O_TRUNC/no-O_TRUNC open, queued/in-flight cancel, and the commit queue unit directly. Race-checked tests for the concurrent cancel paths.

---

### Issues

**1. Inode `entry.Size` not updated after staging — `out.Attr.Size` may return stale value to kernel**

`stagePathTruncateToZeroLocked` updates the inode cache via `cacheFileForPath(entry.Path, 0, ...)` but does not modify `entry.Size` in place. After `SetAttr` calls `stagePathTruncateToZeroLocked` and then populates `out.Attr`, if the attribute-fill path reads `entry.Size` (which is still the pre-truncate value) rather than re-fetching from the updated cache, FUSE returns the wrong size to the kernel. This would cause the kernel to think the file is still non-zero and could suppress the expected `inode_size=0` invalidation.

Please verify that the `SetAttr` attribute-fill code after the staging branch either re-fetches the inode entry from cache or explicitly sets `out.Attr.Size = 0`.

**2. `TestSetAttr_CloseSyncPathTruncateKeepsSynchronousRemoteCommit` — PUT handler returns no body**

The test server's PUT handler does only `w.WriteHeader(http.StatusOK)` with no body. If the client expects a JSON-encoded `{revision: N}` response (as the other test servers provide), decoding will fail and `SetAttr` will return a non-OK status — making the test assert an error path rather than the intended close-sync success path. The final assertion `if st != gofuse.OK` would then fail. Either add the JSON body or confirm the synchronous truncate path on close-sync ignores the response body.

**3. No dedup/coalesce guard for rapid repeated `SETATTR size=0`**

Each call to `stagePathTruncateToZeroLocked` enqueues a new `CommitEntry`. If `SETATTR size=0` fires N times before the first commit drains, N entries accumulate in the queue. Commits 2..N will likely hit a server-side revision conflict (the first already advanced the revision). This is not data-corrupting — the file is already zero — but produces N-1 error log lines and potentially N-1 spurious retries. A simple guard (check `commitQueue.HasPath(path)` before enqueuing a second entry) would keep the queue clean. The PR explicitly says "no fixed coalesce timer" which is fine; this is just a dedup-on-same-path ask.

---

### Minor

- **`lockWritableRemoteCommitPathForWritableOpen` double-check is correct but worth a comment**: after `canSupersedeQueuedPathTruncate` returns true (which already cancelled the queued entry under lock), the subsequent `HasPath` re-check after acquiring the path lock guards against a race where a new entry was enqueued between the cancel and the lock acquisition. The logic is sound; a short comment would make the intent clear to future readers.

- **`isQueuedZeroTruncateEntry` matches any queued zero-byte overwrite**, not only those originating from `stagePathTruncateToZeroLocked`. This is probably intentional (any queued zero overwrite is supersedable), but worth a comment confirming this is deliberate rather than an oversight.

- **In-place slice reuse in `CancelQueuedZeroTruncatePreserveLocal`**: `remaining := cq.queue[:0]` reuses the backing array while ranging over it. This is correct (range captures the slice header at start) and idiomatic, but uncommon enough in this codebase that a brief comment would help.

---

### Verdict

The core design is sound and the async staging path is well-guarded. Two items need a response before merge: (1) confirm the `out.Attr.Size` fill after staging returns 0 to the kernel, and (2) fix or explain the no-body PUT handler in the close-sync test. The repeated-truncate dedup is a minor quality item.

qiffang:
## adversary-1 · PR #561 R1 Review · `6b24cc0`

Verdict: **REVISE**.

The scope is correctly narrow and the writeback gate is in the right place (`WritePolicyWriteBack`, not `SyncMode`). However, this head has a real correctness regression in the e2e smoke gate and cannot merge:

```text
drive9-server-local e2e smoke: FAILURE
ci: SUCCESS
```

The reported failure is mounted tier transition under writeback truncate staging: overwrite a large remote file (`8MiB`) with a small file (`10KiB`), then the remote stat still reports the old `8MiB` size. That means the path truncate staging / queued zero-truncate cancellation path can lose or fail to submit the final small overwrite. This is exactly the class of regression this PR must not introduce.

Required before approval:

- Fix large-to-small overwrite after staged path truncate so the final content commit reaches remote with the correct size and storage tier.
- Add/keep a regression test for mounted large-to-small overwrite through the new writeback path, not only unit-level queued cancel behavior.
- Re-run `drive9-server-local e2e smoke` and focused `pkg/fuse` tests on the fixed head.

I will re-review after the fix is pushed. Current head should not merge.

srstack (commented):
## Code Review: fix(fuse): stage writeback path truncates locally — R2

Reviewing against the latest head. The diff has grown from 4 to 4 files changed, with 776 additions vs 623 in R1. Key changes since the original review:

---

### New in this revision

**1. `stageShadowLocked`: `ShadowReady` → `ShadowReady && ShadowSpill` guard**

```go
-if fh.ShadowReady {
+if fh.ShadowReady && fh.ShadowSpill {
    if err := fs.shadowStore.Truncate(fh.Path, size, fh.BaseRev); err != nil {
```

This is a meaningful correctness fix. A handle loaded from the zero-byte staged shadow has `ShadowReady=true` but `ShadowSpill=false` (it hasn't spilled new writes yet). The old code would call `Truncate` on the shadow unconditionally, which is wrong for that case — the shadow content is already correct (empty), and calling Truncate on it would either no-op or corrupt the base revision. The new `ShadowSpill` guard ensures Truncate is only called when the handle has already spilled dirty data to shadow storage. `TestStageShadowReadyNonSpillRewritesShadowWithDirtyBuffer` covers this well.

**2. `Open`: inode refresh after waiting for in-flight zero truncate**

```go
if refreshedEntry, ok := fs.inodes.GetEntry(input.NodeId); ok {
    fh.OrigSize = refreshedEntry.Size
    fh.BaseRev = refreshedEntry.Revision
    entry = refreshedEntry
}
```

This addresses the stale-handle problem: after `lockWritableRemoteCommitPathForWritableOpen` waits for an in-flight commit to drain, the commit's `OnSuccess` callback has already updated the inode to size=0 and the new revision. Without this refresh, `fh.OrigSize` and `fh.BaseRev` would hold pre-commit values, causing the write buffer to start from a stale baseline. `TestOpenWritableRefreshesInodeAfterWaitingForInFlightZeroTruncate` also verifies that stale read cache data is not preloaded into the dirty buffer — good catch.

**3. `Open`: `Ensure` → `WriteFull` for O_TRUNC path**

```go
-if err := fs.shadowStore.Ensure(p, 0, fh.BaseRev); err != nil {
+if err := fs.shadowStore.WriteFull(p, nil, fh.BaseRev); err != nil {
```

`Ensure` creates a shadow only if absent, preserving existing content. `WriteFull(p, nil, baseRev)` explicitly resets the shadow to empty, which is correct for O_TRUNC semantics. The old code was subtly wrong: if a shadow already existed with stale data, the truncate-open would inherit that data through the shadow path.

---

### Previously raised items — status

**Resolved:**
- Inode entry stale after staging: `cacheFileForPath` updates the cache; `TestSetAttr_WriteBackInteractivePathTruncateStagesAndReturnsBeforeRemoteCommit` verifies `entry.Size == 0` after staging. Confirmed closed.
- Stale read cache preloaded into dirty buffer on in-flight wait: addressed by the new inode refresh + `TestOpenWritableRefreshesInodeAfterWaitingForInFlightZeroTruncate`.

**Still open:**

**B1. `TestSetAttr_CloseSyncPathTruncateKeepsSynchronousRemoteCommit` — PUT handler returns no body**

The test server's PUT handler still does only `w.WriteHeader(http.StatusOK)` with no JSON body. The other write-back test servers return `{"revision": 2}`. If the close-sync synchronous truncate path parses the PUT response body for the new revision, this test is asserting an error path rather than success. The final assertion `if st != gofuse.OK` would silently pass if the error is swallowed somewhere. Please confirm whether the synchronous remote truncate path on close-sync policy expects a JSON revision response or ignores the body.

**B2. Repeated `SETATTR size=0` dedup still missing**

Each call to `stagePathTruncateToZeroLocked` without a prior `HasPath` check enqueues a new `CommitEntry`. Rapid repeated truncates on the same path (e.g. from `vim` or `truncate -s 0` in a loop) accumulate N entries. Commits 2..N will hit server-side revision conflicts once commit 1 advances the revision. Not data-corrupting, but produces N-1 error log lines and potentially N-1 retries. A simple `if fs.commitQueue.HasPath(entry.Path) { return gofuse.OK }` guard before enqueuing would prevent this.

---

### Minor (unchanged from R1)

- `isQueuedZeroTruncateEntry` matches any queued zero-byte overwrite, not only those from `stagePathTruncateToZeroLocked`. Intentional but worth a comment.
- In-place slice reuse `remaining := cq.queue[:0]` in `CancelQueuedZeroTruncatePreserveLocal` is correct; a brief comment helps future readers.

---

### Verdict

The three new fixes (ShadowSpill guard, inode refresh on wait, WriteFull reset) are all correct and well-tested. B1 (close-sync PUT response body) needs a response. B2 (dedup guard) is a quality item. The core flow is solid.

qiffang:
## adversary-1 R2 Review on `0b4182f`: REVISE

The `drive9-server-local e2e smoke` failure is still the same mounted tier-transition assertion:

```text
wait_remote_stat_field_eq: timeout path=/fuse-e2e-.../alpha-renamed/tier-transition.bin field=size want=10240 last_got=8388608
```

I found a blocker in the latest fix:

**B1. `O_TRUNC` reset marks an empty shadow as `ShadowSpill`, so dirty content is not staged.**

In `Open`'s `O_TRUNC` branch, the PR now resets the shadow with `WriteFull(p, nil, fh.BaseRev)`, but then sets:

```go
fh.ShadowReady = true
fh.ShadowSpill = true
```

That is semantically wrong for this path. The shadow has just been reset to empty; the subsequent user writes live in `fh.Dirty`, not in shadow as the authoritative data source.

On close/flush, `stageShadowLocked` sees `ShadowReady && ShadowSpill` and takes this branch:

```go
fs.shadowStore.Truncate(fh.Path, size, fh.BaseRev)
```

That resizes the empty shadow to the final size without writing the dirty buffer bytes. For `open("wb") -> write(10KiB) -> close`, the commit queue no longer has the actual 10KiB content in shadow. This explains why the mounted tier-transition path still does not publish the final 10KiB file.

Required fix:

- Do not set `ShadowSpill=true` for the `O_TRUNC` `WriteFull(nil)` reset path unless the shadow is actually the authoritative content source.
- Add a regression covering stale large same-path shadow + `O_TRUNC` open + 10KiB write + stage/release, asserting the staged shadow contains the dirty 10KiB bytes and pending size is 10240.
- Keep PR #561 blocked until `drive9-server-local e2e smoke` passes on the new head.

qiffang:
## adversary-1 R3 Review on `f270a1a`: GREEN, conditional on e2e green

The blocker I raised on `0b4182f` is fixed.

**Verified fix**

- `Open`'s `O_TRUNC` branch still resets the shadow with `WriteFull(p, nil, fh.BaseRev)`.
- It no longer marks the handle `ShadowSpill=true`.
- Therefore subsequent staging uses `stageShadowLocked`'s non-spill path and writes `fh.Dirty.bytesView()` into shadow instead of only resizing the shadow file.

This matches the actual data ownership: after `O_TRUNC`, shadow is just a reset backing file; the post-open writes live in the dirty buffer until staged.

**Regression coverage**

`TestOpenTruncateResetShadowStagesDirtyBuffer` covers the stale-shadow failure mode directly:

- existing same-path shadow contains old data,
- `O_TRUNC` opens the file,
- caller writes 10KiB,
- staging writes the actual 10KiB dirty content to shadow,
- pending metadata has `ShadowSpill=false` and size `10240`.

Local focused validation passed:

```text
go test ./pkg/fuse -run 'TestOpenTruncateResetShadowStagesDirtyBuffer|TestOpenWritableRefreshesInodeAfterWaitingForInFlightZeroTruncate|TestStageShadowReadyNonSpillRewritesShadowWithDirtyBuffer|TestOpenTruncateCancelsQueuedPathTruncateWithoutCancelingInFlight|TestOpenWritableCancelsQueuedPathTruncateWithoutOTruncFlag|TestSetAttr_WriteBackInteractivePathTruncateStagesAndReturnsBeforeRemoteCommit|TestSetAttr_CloseSyncPathTruncateKeepsSynchronousRemoteCommit|TestSetAttr_WriteBackStrictPathTruncateStagesAndReturnsBeforeRemoteCommit|TestCommitQueueCancelQueuedZeroTruncatePreservesLocalOnlyWhenNotInFlight' -count=1
```

Result:

```text
ok github.com/mem9-ai/dat9/pkg/fuse 1.207s
```

Given the last three revisions only failed in `drive9-server-local e2e smoke`, this remains conditional on that check passing on `f270a1a`.

qiffang:
## adversary-1 R5 Update on `f270a1a`: REVISE

Retracting my conditional green on `f270a1a`: `drive9-server-local e2e smoke` still fails the mounted tier-transition case:

```text
wait_remote_stat_field_eq: timeout ... field=size want=10240 last_got=8388608
```

The remaining failure path appears to be the non-`O_TRUNC` writable-open adoption path after path `SETATTR size=0`.

Concrete race to close before green:

1. Existing 8MiB same-path commit/open-handle state is still present.
2. Path `SETATTR size=0` stages a zero overwrite and enqueues it.
3. Writable `OPEN` cancels only the queued zero-truncate entry, but if an earlier same-path commit is still in flight it falls back to waiting.
4. After the wait, pending/shadow state may be cleared and the inode revision/size refreshed.
5. `loadWritableHandleFromOpenHandleLocked` can still preload from an old clean same-path handle at the previous 8MiB revision.
6. The 10KiB write overwrites only the prefix, and close can submit an 8MiB payload again.

Required before approval:

- Same-path open-handle preload must reject stale clean handles from older revisions. It should only adopt uncommitted local state, or state whose revision is compatible with the current writable handle / inode revision.
- Add a regression for the non-`O_TRUNC` path: old 8MiB same-path handle exists, zero-truncate commit/wait refreshes current inode to size 0/new revision, writable open must not preload old 8MiB content, and final staged/committed size must be 10KiB.
- `drive9-server-local e2e smoke` must pass.

srstack (commented):
## Code Review: fix(fuse): stage writeback path truncates locally — R3

Reviewing against head f270a1a, incorporating qiffang's R1–R5 findings.

---

### Status of previously raised items

**Closed from my R1/R2:**
- inode size stale after staging: confirmed closed (cacheFileForPath + test assertion on entry.Size == 0).
- close-sync PUT handler body: close-sync path goes through the synchronous remote truncate branch which does a follow-up HEAD for the new revision — the no-body PUT is intentional and correct for that path.
- ShadowSpill guard in stageShadowLocked (ShadowReady && ShadowSpill): confirmed present and correct.
- WriteFull without ShadowSpill=true for O_TRUNC path: confirmed present (ShadowSpill=true line removed).
- Inode refresh after in-flight wait: confirmed present and tested by TestOpenWritableRefreshesInodeAfterWaitingForInFlightZeroTruncate.
- TestOpenTruncateResetShadowStagesDirtyBuffer: covers stale-shadow + O_TRUNC + 10KiB write + stage correctly. Confirmed by qiffang R3.

---

### Blocking — still open (qiffang R5)

**B1: Non-O_TRUNC writable open after SETATTR size=0 can preload stale 8MiB clean handle**

qiffang R5 identified the concrete failure path still causing e2e smoke to report size=8388608 instead of size=10240:

1. An 8MiB clean remote file has an open handle at revision N.
2. SETATTR size=0 stages a zero overwrite (shadow=empty, pending=zero-overwrite, commit enqueued at rev N).
3. Writable open (no O_TRUNC) calls canSupersedeQueuedPathTruncate, cancels the queued entry, and proceeds.
4. Inside Open, loadWritableHandleFromOpenHandleLocked can still find and adopt the old same-path 8MiB clean handle from a prior open at revision N.
5. fh.OrigSize and the write buffer are seeded with 8MiB stale content.
6. The 10KiB write patches only the prefix of the 8MiB buffer.
7. Close/commit submits the full 8MiB payload to remote, overwriting the intended 10KiB file.

The current diff does not contain a fix for this path. The inode refresh added in Open (lines 319-323) updates fh.OrigSize and fh.BaseRev from the inode cache after the lock, but it does not prevent loadWritableHandleFromOpenHandleLocked from preloading the stale clean handle before that refresh.

Required before approval:
- Same-path open-handle adoption in loadWritableHandleFromOpenHandleLocked must reject clean handles whose revision is older than the current inode revision (which reflects the zero-truncate staged state).
- Add a unit test for the non-O_TRUNC path: old 8MiB clean handle exists at rev N, SETATTR size=0 advances inode to size=0, writable open must not preload 8MiB content, staged/committed size must be the new write size (e.g. 10KiB).
- drive9-server-local e2e smoke must pass on the fixed head.

---

### Everything else

No new issues found in this revision beyond what qiffang and I have already raised. The commit_queue.go changes, canStagePathTruncateToZero, stagePathTruncateToZeroLocked, and the O_TRUNC shadow reset are all correct. The test suite is thorough for the paths it covers.

Verdict: **REVISE** — blocked on B1 above.

qiffang:
## adversary-1 R6 Update on `683c798`: REVISE

Good progress: the mounted tier-transition correctness issue is fixed on this run. The previous `8MiB -> 10KiB` remote-stat/checksum failure is no longer the failing gate.

The remaining blocker is the write perf budget gate. The failed e2e line is:

```text
FAIL write-path perf budgets ... remote_write=161/408 remote_stat=40/640 remote_list=92/72 remote_mutation=176/84 commit_retries=0/5 commit_enqueue_errors=0/0 commit_failures=32/0 fsync_avg_ms=5/150
```

This is not just a harmless display issue: the PR is introducing failed commit attempts (`commit_failures=32`) and also pushes `remote_mutation` / `remote_list` over budget. Do not relax the counters for this PR; the perf gate is catching a real byproduct of the new async path.

Expected next fix:

- Remove the source of duplicate/obsolete zero-truncate commits, likely by deduping or skipping repeated same-path queued zero-truncate entries before enqueue.
- Keep the fix narrow: do not mark revision conflicts as “expected” in the counter unless there is a precise, tested classification proving they are only obsolete superseded zero-truncates.
- Rerun e2e and show `commit_failures=0` with `remote_mutation` and `remote_list` back under budget.

Verdict remains **REVISE** until the perf budget e2e is green.

qiffang:
## adversary-1 R7 on `ecbbcb9`: REVISE

`ecbbcb9` fixes the perf-counter symptom by skipping post-upload chmod for `PendingOverwrite` when `mode == 0644`, but the current predicate is too broad:

```go
if (kind == PendingNew || kind == PendingOverwrite) && mode&0o777 == defaultRegularFileMode {
    return false
}
```

This loses an explicit chmod on existing files.

Concrete counterexample:

1. Remote file exists with mode `0600`.
2. A writable handle dirties the file.
3. User runs `chmod 0644 file` while that dirty handle is open, so the chmod is deferred into pending mode metadata.
4. Close/release enqueues a `PendingOverwrite` with `HasMode=true, Mode=0644`.
5. New `shouldApplyRemoteMode(PendingOverwrite, true, 0644)` returns false.
6. Data upload succeeds, but the explicit chmod is never sent. Remote mode remains `0600`.

The bug is that `HasMode=true, Mode=0644` does not distinguish inherited/default file mode from an explicit pending chmod to default mode. Skipping all `PendingOverwrite + 0644` conflates those cases.

Required before green:

- Preserve explicit chmod-to-0644 semantics for existing files.
- Only skip default `0644` for overwrites when the mode is known to be inherited/default metadata, not a deferred chmod operation.
- Add a regression where an existing file with non-default mode is overwritten and explicitly chmodded to `0644`; the commit must issue remote chmod and clear pending state only after that chmod succeeds.

The e2e perf counters still need to pass after the narrower fix.

qiffang:
## adversary-1 final review — GREEN on `e1c9df0`

I re-checked after the lint fix. `code-ci` and `drive9-server-local e2e smoke` are both green, and the e2e perf-budget blocker is cleared.

The narrowed mode fix is correct: explicit `chmod 0644` for `PendingOverwrite` is preserved, while redundant chmod is avoided only for staged path truncates with inherited default mode.

No remaining blockers from my review.

Note: GitHub rejects formal approval from the current CLI identity with `Review Can not approve your own pull request`, so this is a comment-level GREEN rather than a submitted approving review.

mornyx (changes requested):
I found one blocking issue.

The remote `Create` path still allocates the new dirty file handle without recording `OpenPID` from the FUSE caller. That makes `adoptSingleCallerPathTruncate` unable to adopt a same-process created handle after a path-level `SETATTR size=0`, because `shouldAdoptSingleHandlePathTruncate` requires `fh.OpenPID == callerPID`.

Concrete failure path:
- process creates a file through FUSE, writes data, and keeps the handle open
- the same process calls path truncate to zero, so `SetAttr` stages the zero-byte overwrite and calls `adoptSingleCallerPathTruncate`
- the created handle has `OpenPID == 0`, so it is not truncated to the zero base
- later close or flush can commit the old dirty bytes from the still-open create handle, overwriting the staged truncate

The fix should set `OpenPID: input.Pid` on the remote `Create` file handle, and add a regression test that uses the real `fs.Create -> fs.Write -> fs.SetAttr(size=0)` path with the same caller PID. A helper-only test with a manually constructed handle can miss this because it can set `OpenPID` directly.

Validation I ran:
- `git diff --check origin/pr/561-merge^1...origin/pr/561-merge`
- `go test ./pkg/fuse -run TestSetAttr_WriteBackInteractivePathTruncateStagesAndReturnsBeforeRemoteCommit|TestSetAttr_CloseSyncPathTruncateKeepsSynchronousRemoteCommit|TestSetAttr_WriteBackStrictPathTruncateStagesAndReturnsBeforeRemoteCommit|TestOpenTruncateCancelsQueuedPathTruncateWithoutCancelingInFlight|TestOpenWritableCancelsQueuedPathTruncateWithoutOTruncFlag|TestCommitQueueCancelQueuedZeroTruncatePreservesLocalOnlyWhenNotInFlight|TestSetAttr_TruncateWithoutHandleRefreshesRevision -count=1`

I also added a temporary local regression test in an isolated `/private/tmp` review snapshot for the real create/write/path-truncate flow. It fails on this head with: `created same-caller writer did not adopt zero base after path truncate`.

qiffang:
## Response to GitHub Reviews

### mornyx CHANGES_REQUESTED — Create missing OpenPID ✅
Valid bug. `Create` path allocates FileHandle without `OpenPID`, breaking `adoptSingleCallerPathTruncate` for `Create → write → truncate(0)` same-process flow. @dat9-dev2 is fixing: add `OpenPID: input.Pid` + regression test.

### srstack R1-R3 — Status
- **Issue 1 (entry.Size stale)**: Closed. `cacheFileForPath(path, 0, ...)` updates cache; `SetAttr` fills `out.Attr` from updated inode.
- **Issue 2 (close-sync PUT no body)**: Closed. Close-sync path does a follow-up HEAD for revision, no-body PUT is intentional.
- **Issue 3 (dedup guard)**: Nice-to-have, not blocking. Duplicate commits produce log noise but no data corruption.
- **R3 B1 (stale handle preload)**: Fixed in `683c798` via `pendingLocal` filter in `loadWritableHandleFromOpenHandleLocked`.

### CodeRabbit Comments
- **Create OpenPID** (dat9fs.go:1496): Same as mornyx — being fixed.
- **ZeroBase/OrigSize** (dat9fs.go:7591): Pre-existing on `main` (not introduced by this PR). `ZeroBase=true` without `OrigSize=0` exists in both branches.
- **entry.Revision=0** (dat9fs.go:2363): False positive. `SetAttr` path's `entry` comes from `inodes.GetEntry` for existing remote files with revision>0.
- **Test improvements** (releasePut cleanup, select timing, pending assertions): Nice-to-have, not blocking.

Waiting for @dat9-dev2's OpenPID fix, then re-review.

qiffang:
Addressed the latest blocking review on head `f6272e6`.

Fix:
- Set `OpenPID: input.Pid` on remote `Create` file handles so `adoptSingleCallerPathTruncate` can recognize a same-process created writer during path-level `SETATTR size=0`.

Regression test:
- Added `TestSetAttr_WriteBackPathTruncateAdoptsCreatedSameCallerWriter`, which uses the real `fs.Create -> fs.Write -> fs.SetAttr(size=0)` path with the same caller PID.
- It verifies the created handle records `OpenPID`, adopts `ZeroBase`, clears dirty size to 0, marks the truncate dirty, and leaves the shadow content empty instead of preserving old created-handle dirty bytes.

Validation run locally:
- `/usr/local/go/bin/go test ./pkg/fuse -run 'TestSetAttr_WriteBackPathTruncateAdoptsCreatedSameCallerWriter|TestSetAttr_WriteBackPathTruncateAdoptsSingleCallerWriter' -count=1 -timeout=60s -v`
- `bash -n e2e/fuse-smoke-test.sh`
- `git diff --check`

Notes on other latest comments:
- The CodeRabbit `Create`/`OpenPID` finding is the same bug and is fixed by this commit.
- The remaining CodeRabbit test-hardening comments are not correctness blockers for this PR; existing focused tests plus the added real Create-path regression cover the blocking path.
- The e2e corner cases added in prior commits still cover explicit path `os.truncate(path, 0)` standalone, rename, unlink, remount visibility, and stable missing checks for resurrection.

@Mornyx please re-review when CI/e2e finish.

qiffang:
## adversary-1 R11 on `f6272e6`: REVISE

The `Create` `OpenPID` fix is correct, but I am retracting the earlier “entry.Revision=0 is false positive” assessment. There is still a blocker in the newly covered `Create -> Write -> truncate(path, 0)` path.

Concrete issue:

1. Remote `Create` allocates a new inode/handle with `Revision=0`.
2. A same-process path-level `SetAttr(size=0)` goes through `stagePathTruncateToZeroLocked`.
3. `stagePathTruncateToZeroLocked` unconditionally records and enqueues:

```go
PendingOverwrite, BaseRev: entry.Revision
```

For a created-but-uncommitted file this becomes `PendingOverwrite` with `BaseRev=0`.

That entry is invalid for the commit queue. `uploadEntry` explicitly rejects non-layer `PendingOverwrite` with `BaseRev <= 0`:

```go
if entry.Kind == PendingOverwrite && expectedRevision <= 0 && layerRef == "" {
    return 0, fmt.Errorf("missing base revision for overwrite: %s", entry.Path)
}
```

The new test verifies `OpenPID`, `ZeroBase`, dirty size, and shadow content, but it does not inspect the queued commit kind/baseRev and does not exercise `Release`/`DrainAll`. It can therefore pass while leaving an invalid queued zero-truncate entry that will later produce `commit_failure` / pending conflict.

Required before green:

- Do not enqueue `PendingOverwrite(BaseRev=0)` for path truncate on an uncommitted newly-created file.
- Either preserve `PendingNew` semantics for this path, or skip the path zero-commit when the same created writer is adopted and let `Release` enqueue the empty `PendingNew` content.
- Add a regression that covers `Create -> Write -> truncate(path, 0) -> Release/DrainAll` and asserts no commit failure/conflict. At minimum, assert the queued entry is never `PendingOverwrite` with `BaseRev=0`.

Until this is fixed, PR #561 remains blocked despite the `OpenPID` fix.

mornyx (approved):
I re-reviewed the latest head f6272e6.

The previous blocker is fixed: the remote Create path now records `OpenPID: input.Pid`, and the added regression test exercises the real `fs.Create -> fs.Write -> fs.SetAttr(size=0)` flow for a same-caller truncate adoption. I also rechecked the stale clean-handle preload case; the `pendingLocal` filter prevents clean sibling handles from being adopted.

I did not find any blocking issues in this round.

Validation performed:
- `git diff --check origin/pr/561-merge^1...origin/pr/561-merge`
- `go test ./pkg/fuse -run "TestSetAttr_WriteBackPathTruncateAdoptsCreatedSameCallerWriter|TestSetAttr_WriteBackInteractivePathTruncateStagesAndReturnsBeforeRemoteCommit|TestSetAttr_CloseSyncPathTruncateKeepsSynchronousRemoteCommit|TestSetAttr_WriteBackStrictPathTruncateStagesAndReturnsBeforeRemoteCommit|TestOpenTruncateCancelsQueuedPathTruncateWithoutCancelingInFlight|TestOpenWritableCancelsQueuedPathTruncateWithoutOTruncFlag|TestCommitQueueCancelQueuedZeroTruncatePreservesLocalOnlyWhenNotInFlight|TestSetAttr_TruncateWithoutHandleRefreshesRevision" -count=1`

Approved.

qiffang:
Addressed the `Create -> Write -> truncate(path, 0)` revision-0 blocker in `a664e46`.

Fix:
- `canStagePathTruncateToZero` now requires `entry.Revision > 0`, so path-level zero-truncate staging only creates `PendingOverwrite` when there is a real remote base revision for CAS.
- For newly-created, uncommitted files (`Revision <= 0`) with a single same-caller dirty handle, `SetAttr(size=0)` now adopts the truncate into that open handle and does **not** enqueue a path-level `PendingOverwrite(BaseRev=0)`.
- The created handle keeps ownership of the eventual commit and remains `PendingNew(BaseRev=0)`, so release/flush uses the new-file path instead of an invalid overwrite.

Regression test strengthened:
- `TestSetAttr_WriteBackPathTruncateAdoptsCreatedSameCallerWriter` now verifies the real `Create -> Write -> SetAttr(size=0) -> Release` flow.
- It asserts no commit queue entry is created for the path truncate, no path-level pending overwrite is recorded, the handle state is `PendingNew` with base rev `0`, and `Release` creates the remote empty file without issuing a PUT overwrite.

Local validation:
- `/usr/local/go/bin/go test ./pkg/fuse -count=1 -run 'TestSetAttr_WriteBackPathTruncate(AdoptsCreatedSameCallerWriter|AdoptsSingleCallerWriter|StagesAndReturnsBeforeRemoteCommit|SkipsDefaultInheritedMode|PreservesNonDefaultMode)'`
- `/usr/local/go/bin/go test ./cmd/drive9 ./cmd/drive9/cli ./pkg/fuse -count=1`
- `bash -n e2e/fuse-smoke-test.sh`
- `git diff --check`

Note: the branch ref is updated to `a664e46a4dcf87f856854261de7557aa93b55e0e`; `gh pr view` was briefly still showing the previous head, but `gh api repos/mem9-ai/drive9/git/ref/heads/dat9-dev2%2Fissue-560-truncate-staging` confirms the pushed ref.

@Mornyx please re-review this revision when GitHub refreshes the PR head/checks.

qiffang:
Clarification after merge: PR #561 was already merged at `f6272e6` before the `Revision=0` fix could land in the PR branch. I opened follow-up PR #563 with the one-commit fix: https://github.com/mem9-ai/drive9/pull/563

The follow-up preserves `PendingNew(BaseRev=0)` semantics for `Create -> Write -> truncate(path, 0)` and prevents enqueueing `PendingOverwrite(BaseRev=0)`. Local validation passed as listed in #563.

PR #562, opened 2026-06-15
Add native tenant delete cleanup flow

srstack:
## Summary
- add owner-only `DELETE /v1/tenant` for live tenant deletion
- synchronously delete TiDB Cloud Starter clusters before revoking keys; `tidbcloud_native` uses request-provided customer public/private keys and does not store them
- reject `tidb_zero` deletes and reject live tenant deletes while non-deleted forks share the storage namespace
- add durable `tenant_delete_jobs` and async S3 prefix cleanup with retry/recovery
- add S3 prefix deletion for AWS/local clients and cleanup access by storage namespace

## Testing
- `go test ./pkg/server -run 'TestTenantDelete'`\n- `make test TEST_PKGS='./pkg/server'`\n- `make test TEST_PKGS='./pkg/meta'`\n- `go test ./pkg/s3client`\n- `make test TEST_PKGS='./pkg/tenant'`\n- `make test TEST_PKGS='./pkg/backend'`\n- `make test TEST_PKGS='./pkg/tenant/starter ./pkg/tenant/tidbcloudnative'`\n- `make lint`\n- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Added owner-scoped `DELETE /v1/tenant` with idempotent tenant state transitions and asynchronous deletion cleanup.
  * Implemented best-effort storage prefix cleanup for both S3 and local storage, including object deletion and multipart upload aborts.
  * Introduced “deleting” storage namespace lifecycle and persisted delete-job orchestration with retry/recovery.
  * Extended teardown support to include credential-aware deprovisioning.
* **Bug Fixes**
  * Improved request routing/metrics and clearer authorization errors for scoped token actions.
* **Documentation**
  * Updated API docs for tenant deletion credential/provider behavior and clarified spending-limit defaults.
* **Tests**
  * Added integration coverage for deletion authorization, eligibility checks, retries, idempotency, and async cleanup behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -409,7 +409,7 @@ Important environment variables:
 DRIVE9_META_DSN                 control-plane MySQL/TiDB DSN
 DRIVE9_TENANT_PROVIDER          db9 | tidb_zero | tidb_cloud_starter | tidbcloud_native
 DRIVE9_TIDBCLOUD_DEFAULT_SPENDING_LIMIT
-                                default TiDB Cloud Starter spendingLimit.monthly in USD cents
+                                default TiDB Cloud Starter/native spendingLimit.monthly in USD cents
 DRIVE9_TIDBCLOUD_NATIVE_API_URL TiDB Cloud Serverless API base URL
 DRIVE9_TIDBCLOUD_NATIVE_CLOUD_PROVIDER
                                 cloud provider for tidbcloud_native cluster creation
--- a/README.md
+++ b/README.md
@@ -364,10 +364,16 @@ POST   /v2/uploads/{id}/abort
 
 GET    /v1/events                 SSE change stream
 GET    /v1/status                 server status and upload limits
+DELETE /v1/tenant                 delete current tenant with owner API key
 ```
 
 Vault endpoints also exist for secret storage, scoped grants, delegated tokens,
 and read-only vault mounts.
+
+`DELETE /v1/tenant` is owner-key only. For `tidbcloud_native`, the request body
+must also include the customer's TiDB Cloud `public_key` and `private_key` so
+Drive9 can delete the customer-account Starter cluster. `tidb_zero` tenants are
+not deleted through this endpoint; they expire automatically.
 
 ## Mount Options
 
--- a/README.md
+++ b/README.md
@@ -409,7 +409,8 @@ Important environment variables:
 DRIVE9_META_DSN                 control-plane MySQL/TiDB DSN
 DRIVE9_TENANT_PROVIDER          db9 | tidb_zero | tidb_cloud_starter | tidbcloud_native
 DRIVE9_TIDBCLOUD_DEFAULT_SPENDING_LIMIT
-                                default TiDB Cloud Starter/native spendingLimit.monthly in USD cents
+                                default TiDB Cloud spendingLimit.monthly in USD cents
+                                optional for tidb_cloud_starter; tidbcloud_native defaults to 1000 when unset
 DRIVE9_TIDBCLOUD_NATIVE_API_URL TiDB Cloud Serverless API base URL
 DRIVE9_TIDBCLOUD_NATIVE_CLOUD_PROVIDER
                                 cloud provider for tidbcloud_native cluster creation
--- a/cmd/drive9-server/main.go
+++ b/cmd/drive9-server/main.go
@@ -434,7 +434,7 @@ environment:
   DRIVE9_TIDB_AUTO_EMBEDDING_API_BASE provider base endpoint for models that require it
                                      optional for openai models; set it for Azure OpenAI endpoints
   DRIVE9_TENANT_PROVIDER db9|tidb_zero|tidb_cloud_starter|tidbcloud_native (default for provisioning)
-  DRIVE9_TIDBCLOUD_DEFAULT_SPENDING_LIMIT default Starter spendingLimit.monthly in USD cents; applied after pool takeover
+  DRIVE9_TIDBCLOUD_DEFAULT_SPENDING_LIMIT default TiDB Cloud spendingLimit.monthly in USD cents; optional for Starter, native defaults to 1000 when unset
   DRIVE9_TIDBCLOUD_NATIVE_API_URL TiDB Cloud Serverless API base URL for tidbcloud_native
   DRIVE9_TIDBCLOUD_NATIVE_CLOUD_PROVIDER cloud provider for tidbcloud_native cluster creation, e.g. aws
   DRIVE9_TIDBCLOUD_NATIVE_REGION region for tidbcloud_native cluster creation, e.g. us-east-1
--- a/pkg/meta/meta.go
+++ b/pkg/meta/meta.go
@@ -1991,10 +1991,11 @@ func (s *Store) RecoverStaleTenantDeleteJobs(ctx context.Context, before time.Ti
 func (s *Store) MarkTenantDeleteJobRunning(ctx context.Context, tenantID string) (updated bool, err error) {
 	start := time.Now()
 	defer observeMeta(ctx, "mark_tenant_delete_job_running", start, &err)
+	now := time.Now().UTC()
 	res, err := s.db.ExecContext(ctx, `UPDATE tenant_delete_jobs
 		SET state = ?, updated_at = ?
-		WHERE tenant_id = ? AND state = ?`,
-		TenantDeleteJobRunning, time.Now().UTC(), tenantID, TenantDeleteJobPending)
+		WHERE tenant_id = ? AND state = ? AND not_before <= ?`,
+		TenantDeleteJobRunning, now, tenantID, TenantDeleteJobPending, now)
 	if err != nil {
 		return false, err
 	}
@@ -2022,6 +2023,55 @@ func (s *Store) MarkTenantDeleteJobDeleted(ctx context.Context, tenantID string,
 		WHERE tenant_id = ?`,
 		TenantDeleteJobDeleted, deletedObjects, abortedMultipartUploads, now, now, tenantID)
 	return err
+}
+
+func (s *Store) FinalizeTenantDelete(ctx context.Context, tenantID, namespaceID string, deletedObjects, abortedMultipartUploads int64) (err error) {
+	start := time.Now()
+	defer observeMeta(ctx, "finalize_tenant_delete", start, &err)
+	now := time.Now().UTC()
+	return s.InTx(ctx, func(tx *sql.Tx) error {
+		res, err := tx.ExecContext(ctx, `UPDATE tenant_delete_jobs
+			SET state = ?, deleted_objects = ?, aborted_multipart_uploads = ?, completed_at = ?,
+				last_error = NULL, updated_at = ?
+			WHERE tenant_id = ? AND state = ?`,
+			TenantDeleteJobDeleted, deletedObjects, abortedMultipartUploads, now, now, tenantID, TenantDeleteJobRunning)
+		if err != nil {
+			return err
+		}
+		if err := requireAffected(res); err != nil {
+			return err
+		}
+		if namespaceID != "" {
+			res, err = tx.ExecContext(ctx, `UPDATE storage_namespaces SET state = ?, updated_at = ? WHERE namespace_id = ?`,
+				StorageNamespaceDeleted, now, namespaceID)
+			if err != nil {
+				return err
+			}
+			if err := requireAffected(res); err != nil {
+				return err
+			}
+		}
+		res, err = tx.ExecContext(ctx, `UPDATE tenants SET status = ?, updated_at = ? WHERE id = ?`,
+			TenantDeleted, now, tenantID)
+		if err != nil {
+			return err
+		}
+		if err := requireAffected(res); err != nil {
+			return err
+		}
+		return nil
+	})
+}
+
+func requireAffected(res sql.Result) error {
+	n, err := res.RowsAffected()
+	if err != nil {
+		return err
+	}
+	if n == 0 {
+		return ErrNotFound
+	}
+	return nil
 }
 
 func getStorageNamespaceTx(ctx context.Context, tx *sql.Tx, id string) (*StorageNamespace, error) {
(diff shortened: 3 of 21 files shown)


srstack (commented):
## Review: Add native tenant delete cleanup flow

Good overall design — the two-phase approach (synchronous cluster delete + async S3 cleanup with a durable job) is the right model for a destructive operation of this scale. The idempotency guards (CAS status transition, duplicate-key upsert, 404-as-success on deprovision) and the stale-job recovery loop are all present. Test coverage is solid. A few issues worth addressing before merge:

---

### Blocking

**S3 deletion has no per-job deadline — stale recovery fires at 30 min**

`processTenantDeleteJob` runs `s3c.DeletePrefix(ctx, "")` with the background context and no timeout. `defaultTenantDeleteRunningTTL` is 30 minutes. For a large tenant, `DeletePrefix` could run longer than 30 minutes, at which point `RecoverStaleTenantDeleteJobs` resets the job back to `pending` and a second worker picks it up — two goroutines deleting the same prefix concurrently.

```go
// tenant_delete.go
res, err := s3c.DeletePrefix(ctx, "")  // no deadline; ctx is backgroundWithTrace(Background())
```

Fix: wrap the job execution context with a deadline slightly shorter than `defaultTenantDeleteRunningTTL`, or set the TTL to a value you're confident exceeds any realistic deletion time.

---

### Significant

**`PrefixDeleteResult` counters are `int` but the DB columns are `BIGINT`**

```go
type PrefixDeleteResult struct {
    DeletedObjects          int
    AbortedMultipartUploads int
}
```

The cast to `int64` at the call site (`int64(res.DeletedObjects)`) prevents compile errors, but `int` is 32-bit on 32-bit platforms and makes the interface look like it can't handle counts > 2B objects. Since the DB stores `BIGINT` and this is tenant-level mass deletion, the fields should be `int64` to match and to make the type self-documenting.

---

**Cluster already deprovisioned but S3 enqueue failure leaves tenant stuck with an active key**

`TestTenantDeleteKeepsOwnerKeyWhenCleanupEnqueueFails` validates this path intentionally — the tenant stays in `TenantDeleting` with 1 active key for retry. On retry, `deprovisionTenantCluster` is called again (404 → accepted), then `enqueueTenantDeleteJob` fails again if the namespace is still broken. The cluster is permanently gone but the key keeps working. This could be a security concern: a deleted cluster shouldn't allow active API key usage indefinitely.

Consider: once cluster deprovision has succeeded, revoke keys immediately regardless of whether namespace enqueue succeeds, and rely on the retry path to complete the S3 cleanup. The user would then need to contact support to resolve a broken namespace rather than having the key stay active.

---

### Minor

**`DeletePrefix(ctx, "")` — empty prefix meaning is non-obvious at call site**

The empty string works because the client was constructed with `ns.Prefix` already baked in, so `""` means "the whole namespace". This isn't obvious from reading `processTenantDeleteJob`. A one-line comment would help:

```go
// "" = delete everything under the namespace prefix baked into s3c
res, err := s3c.DeletePrefix(ctx, "")
```

---

**`startTenantDeleteCleanup` reuses `startForkWorker`**

Functional, but `startForkWorker` is named for fork-tenant operations. If `startForkWorker` is just "start a background goroutine managed by the server", rename it or add a generic alternative to avoid confusion for future readers.

---

**`TestTenantDeleteNativeRequiresBothOwnerAndCustomerCredentials` — name describes a guard, body tests the happy path**

The test verifies that a successful delete with both keys proceeds end-to-end, not that "both are required". Something like `TestTenantDeleteNativeSucceedsWithCredentials` would be clearer.

---

**Missing test: concurrent DELETE calls on the same tenant**

The CAS pattern in `UpdateTenantStatusIf` should make this safe, but there's no test covering two simultaneous `DELETE /v1/tenant` requests racing to change status from `Active → Deleting`. A test asserting exactly one deprovision call fires would increase confidence.

---

### Nits

- `RecoverStaleTenantDeleteJobs` calls `time.Now().UTC()` twice in one statement — these could drift by a nanosecond. Using a single `now := time.Now().UTC()` would be cleaner.
- The `handleTenantDeleteAuth` switch allows `TenantDeleted` through to the handler (which then returns 404). That's fine, but it means an auth lookup is performed for an already-deleted tenant on every retry. Low cost, just unusual.

---

Overall this is close to mergeable. The S3 timeout issue is the main correctness concern; the `PrefixDeleteResult` type and the post-deprovision key revocation question are worth deciding on before shipping.

srstack (commented):
## Follow-up review (commit ff857303)

All three blocking/significant issues from round 1 are resolved. Remaining items below.

### Fixed ✓

- `PrefixDeleteResult` fields are now `int64` — DB schema and call sites all consistent.
- `DeletePrefix` is now wrapped with `context.WithTimeout(ctx, 25min)` — comfortably inside the 30-min stale TTL, so a slow job can't be re-queued and run concurrently.
- `processTenantDeleteJobs` uses a single `now` — no nanosecond drift in recover/list.
- Empty-prefix comment added at the `DeletePrefix` call site.
- `starter.Provisioner.Deprovision` now wraps the error with context.
- Two test names corrected.
- Bonus: `startTenantDeleteCleanup` now fires `processTenantDeleteJobs` immediately on startup rather than waiting the first tick — good catch, picks up any jobs left over from a crash/restart.

### Still open

**Active key persists when namespace enqueue fails (by design?)**

The `TestTenantDeleteKeepsOwnerKeyWhenCleanupEnqueueFails` test is unchanged — the cluster is deleted but the owner key stays active if `GetStorageNamespace` returns not-found. This is the one significant issue from round 1 that hasn't been addressed. If the intent is to leave the key active so the user can retry, that's a valid choice, but it should be a documented decision. If a future retry also deprovisions (404 → ok) and then fails enqueue again, the window where the key is active on a deleted cluster can be indefinitely long.

Recommend either: (a) revoke the key immediately after a successful deprovision regardless of enqueue outcome, or (b) add a comment in `handleTenantDelete` explicitly acknowledging the tradeoff.

**`startForkWorker` naming** — still used for non-fork work; low priority but worth a rename or a generic alias at some point.

---

Overall: the correctness issues are resolved. The one remaining question (active key on enqueue failure) needs a documented decision before merge.

srstack:
This is intentional. If cleanup enqueue fails after cluster deprovision, we keep the owner key active so the same owner can retry and durably enqueue cleanup. This is necessary for `tidbcloud_native` because the customer `public_key` / `private_key` are request-scoped and are not stored server-side.

srstack (commented):
The comment added in 60a2c092 documents the design decision clearly — keeping the owner key active after a failed enqueue is the right call for `tidbcloud_native` since credentials are per-request and can't be replayed from the server side. The tradeoff is now explicit in the code.

All findings from rounds 1 and 2 are resolved:
- S3 job deadline (25 min timeout) ✓
- `PrefixDeleteResult` as `int64` ✓
- `processTenantDeleteJobs` fires immediately on startup ✓
- Empty-prefix comment at call site ✓
- `starter.Deprovision` error wrapping ✓
- Test names corrected ✓
- Active-key-on-enqueue-failure: documented by design ✓

LGTM.

mornyx (changes requested):
I re-reviewed the latest head eb55795.

Blocking findings:

1. The tenant delete cleanup job is marked terminal before the tenant and namespace metadata are finalized. `processTenantDeleteJob` calls `MarkTenantDeleteJobDeleted` first, then updates the storage namespace and tenant status afterward (`pkg/server/tenant_delete.go:274`, `pkg/server/tenant_delete.go:277`, `pkg/server/tenant_delete.go:280`). If either later DB write fails, `ListDueTenantDeleteJobs` only scans pending jobs (`pkg/meta/meta.go:1948`), so the job is never retried and the tenant can remain permanently `deleting` with revoked keys. A later DELETE also short-circuits because `TenantDeleteJobExists` sees the terminal job. Please make job, namespace, and tenant finalization atomic, or leave the job retryable until those final status updates are durable. This also needs a regression test that injects a failure after S3 deletion but before the tenant status update.

2. Tenant delete does not drain already borrowed tenant backends before destructive cleanup. `handleTenantDelete` only calls `s.pool.Invalidate(t.ID)` (`pkg/server/tenant_delete.go:121`), and `Invalidate` does not wait for entries with `refs > 0` (`pkg/tenant/pool.go:295`, `pkg/tenant/pool.go:858`). Requests that passed auth before the tenant was marked deleting can keep using the old backend while the cleanup job deletes the namespace prefix (`pkg/server/tenant_delete.go:270`). That can leave objects written after the deletion pass, or orphan S3 objects from writes whose DB commit fails after cluster deprovision. Fork cleanup already uses `InvalidateAndWait` before cleanup (`pkg/server/fork.go:665`); tenant delete needs an equivalent fence before cluster/S3 cleanup. Please add coverage that holds an acquired backend or in-flight request across DELETE.

Non-blocking but worth fixing:

- `MarkTenantDeleteJobRunning` does not check `not_before` when claiming a job (`pkg/meta/meta.go:1994`). In a multi-server race, another worker can immediately re-run a job that was just retried with a future `not_before`. Consider claiming with both `state = pending` and `not_before <= now`.

Validation performed:
- `git diff --check origin/pr/562-merge^1...origin/pr/562-merge`
- `go test ./pkg/server ./pkg/meta ./pkg/tenant/... ./pkg/s3client -count=1`

Requesting changes for the two blockers above.

mornyx (approved):
Code review only in this pass.

The previous blocking issues look addressed:
- Tenant delete now drains/invalidate-waits borrowed backends before cluster deprovision, so in-flight writes should not race with prefix cleanup.
- Cleanup finalization now updates the delete job, storage namespace, and tenant status in one transaction, so a later tenant-status failure should not leave the job permanently marked deleted.
- Delete-job claiming now honors not_before.

I also re-checked the surrounding auth/route/provisioner/S3 cleanup paths and did not find a new blocking issue. One small non-blocking hardening idea for a follow-up: make namespace state updates idempotent even when the requested state is already present, so recovery paths do not depend on updated_at changing.

Approve from my side.

srstack (commented):
## Round 3 review (commits eb55795f / dee7c822 / e4df4b4f)

Three new commits. Summary first, then details.

**Overall**: all three commits are positive hardening. The finalization transaction is the most significant correctness improvement. No blocking issues; a few things worth having eyes on.

---

### eb55795f + dee7c822 — Default spending limit for native clusters

`parseDefaultSpendLimit` defaults to `1000` (USD cents) when `DRIVE9_TIDBCLOUD_DEFAULT_SPENDING_LIMIT` is unset. This is a safe guard against runaway cost, but it changes the behavior for any existing `tidbcloud_native` deployments that don't set the env var: new clusters will now get a $10/month cap that wasn't there before.

Two things to be aware of:

- **No opt-out path**: there's no way to create a Native cluster without a spending limit. Setting the env var to `0` produces a $0/month limit (blocks all spending), not "unlimited". If operators ever need truly unlimited clusters, the code would need a sentinel value (e.g. `-1` or a separate flag). Fine for now, just worth knowing.
- **Env var reuse**: `DRIVE9_TIDBCLOUD_DEFAULT_SPENDING_LIMIT` previously only applied to Starter (post-takeover `UpdateSpendingLimit`). It now also applies to Native at cluster creation time. Existing deployments that set this for Starter will silently get that value applied to Native creation too. Probably the right call, but worth a note in operator docs / changelog.

Test coverage is thorough — valid/empty/invalid values all covered.

---

### e4df4b4f — Harden tenant delete finalization

Three distinct improvements in one commit:

**1. `InvalidateAndWait` before cluster deprovision**

Correct ordering. The previous `pool.Invalidate` (fire-and-forget) happened *after* deprovision, meaning live connections could be accessing a cluster that had already been deleted. `InvalidateAndWait` now drains all borrowed backends before the DELETE call goes to TiDB Cloud. The rollback path if drain fails correctly reverts status from `Deleting → Active` when the original status was Active.

One edge: if the request context is cancelled (client disconnects) mid-drain, `InvalidateAndWait` returns an error and the rollback `UpdateTenantStatusIf` is silently swallowed (`_, _ =`). This is acceptable since a cancelled request is already exceptional, but the tenant could be left in `TenantDeleting` with no job, requiring a retry. Same scenario as the documented enqueue-failure tradeoff.

`TestTenantDeleteWaitsForBorrowedBackendBeforeClusterDelete` covers this well.

**2. `FinalizeTenantDelete` — atomic three-way commit**

Replacing the three sequential meta calls with a single transaction is the right move. Previously, if the job-deleted update succeeded but the namespace update failed, the tenant would stay in `TenantDeleting` forever (the job is marked done but the tenant status never advances). The transaction eliminates all partial-commit scenarios.

`requireAffected` on each step means any 0-rows result rolls back everything. One edge case: if the namespace row is missing (e.g. manually deleted between job enqueue and finalization), `requireAffected` on the namespace UPDATE returns `ErrNotFound`, the whole tx rolls back, the job stays `running`, and `RecoverStaleTenantDeleteJobs` will eventually reset it to `pending` for retry. The retry will fail at finalization again indefinitely until the data inconsistency is resolved manually. This is an extreme edge case and the 1-hour retry delay makes it non-spammy — fine.

`TestFinalizeTenantDeleteRollsBackWhenTenantStatusUpdateFails` covers the rollback path.

**3. `MarkTenantDeleteJobRunning` respects `not_before`**

Previously the CAS only checked `state = pending`, so a job reset by `RecoverStaleTenantDeleteJobs` with `not_before = now + 1h` could be immediately re-acquired by the next worker tick. Adding `AND not_before <= ?` to the WHERE clause correctly enforces the retry delay. `TestMarkTenantDeleteJobRunningHonorsNotBefore` covers this.

---

### Nit

`processTenantDeleteJobs` now logs errors from `RecoverStaleTenantDeleteJobs` and returns early on failure. The early return means if recovery fails (transient DB error), due jobs are not processed that tick. This is conservative and safe, but it's a behavior change from before (previously errors were silently ignored and the list query continued). Worth noting in case monitoring shows a spike in skipped ticks.

---

All outstanding issues from previous rounds are resolved and these hardening commits address real correctness gaps. LGTM on the delete flow itself.

PR #566, opened 2026-06-16
feat(cli): make read cache ttl configurable

mornyx:
## Summary
- add configurable FUSE userspace read cache TTL
- support --read-cache-ttl=0 to disable time-based expiry
- keep capacity/LRU, revision mismatch, mutation invalidation, and SSE invalidation unchanged

## Tests
- env GOCACHE=/private/tmp/drive9-ttl-worktree/.codex-gocache go test ./cmd/drive9/cli
- env GOCACHE=/private/tmp/drive9-ttl-worktree/.codex-gocache go test ./pkg/fuse

## Summary by CodeRabbit

* **New Features**
  * Added `--read-cache-ttl` duration flag to mount command for controlling read cache expiration (pass 0 to disable time-based expiry; unsupported with WebDAV mode)

* **Tests**
  * Added comprehensive test coverage for read cache TTL configuration, defaults, and validation

### DIFF
--- a/cmd/drive9/cli/fuse_bridge.go
+++ b/cmd/drive9/cli/fuse_bridge.go
@@ -34,6 +34,7 @@ type mountFuseOptions struct {
 	CacheDir                string
 	CacheSize               int64
 	ReadCacheMaxFileBytes   int64
+	ReadCacheTTL            time.Duration
 	DiskReadCacheSize       int64
 	DiskReadCacheFreeRatio  float64
 	DirTTL                  time.Duration
--- a/cmd/drive9/cli/fuse_bridge_supported.go
+++ b/cmd/drive9/cli/fuse_bridge_supported.go
@@ -27,6 +27,7 @@ func mountFuseImpl(opts *mountFuseOptions) error {
 		CacheDir:                opts.CacheDir,
 		CacheSize:               opts.CacheSize,
 		ReadCacheMaxFileBytes:   opts.ReadCacheMaxFileBytes,
+		ReadCacheTTL:            opts.ReadCacheTTL,
 		DiskReadCacheSize:       opts.DiskReadCacheSize,
 		DiskReadCacheFreeRatio:  opts.DiskReadCacheFreeRatio,
 		DirTTL:                  opts.DirTTL,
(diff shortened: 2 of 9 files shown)


srstack (commented):

srstack (commented):
## Review: feat(cli): make read cache ttl configurable

Clean, focused feature. The three-layer pass-through (CLI flag → `mountFuseOptions` → `MountOptions` → `ReadCache`) is correct throughout with no excess code.

---

### Design

**Sentinel convention is clear but asymmetric across layers**

The CLI uses `-time.Nanosecond` for "no-expiry", while `pkg/fuse/read.go` defines `readCacheNoExpiryTTL = -1`. They are the same value (`time.Duration(-1) == -1ns`), but the CLI layer references `-time.Nanosecond` independently rather than reusing the constant from the fuse package. If the sentinel value ever changes, there are two places to update. Minor, but worth noting.

**`readCacheTTLFlagValue` three-way encoding is sound**

- `!given → 0` — lets `setDefaults()` fill in `defaultReadCacheTTL` (30s)
- `value == 0 → -time.Nanosecond` — disables time-based expiry
- `value < 0 → error` — rejects negative input from the user

`setDefaults` only replaces `ReadCacheTTL` when it is `== 0`, so the no-expiry sentinel (`-1ns`) passes through unchanged. `TestMountOptionsReadCacheTTLKeepsNoExpiry` verifies this. ✓

---

### Minor issues

**`readCacheNoExpiryTTL` lacks an explicit type annotation**

```go
const readCacheNoExpiryTTL = -1
```

Without the type, this is an untyped integer constant. It works because `NewReadCache` takes `time.Duration` and the conversion is implicit, but it can mislead a reader. Recommend:

```go
const readCacheNoExpiryTTL time.Duration = -1
```

**WebDAV rejection check ordering**

`readCacheTTLFlagValue` is called before the WebDAV mode check:

```go
normalizedReadCacheTTL, err := readCacheTTLFlagValue(readCacheTTLGiven, *readCacheTTL)
if err != nil { return err }
...
if resolved == MountModeWebDAV && readCacheTTLGiven {
    return fmt.Errorf("drive9 mount: --read-cache-ttl is only supported with --mode=fuse")
}
```

If the user passes `--mode webdav --read-cache-ttl -1s`, they get a TTL validation error rather than the WebDAV rejection. This priority order is reasonable (bad value before bad combination), but there's no test covering `--mode=webdav --read-cache-ttl=-1s` — only the valid-value case is tested.

**Test name could be clearer**

`TestMountCmdLeavesReadCacheTTLUnsetByDefault` — "unset zero value" may suggest that `ReadCacheTTL=0` means "disabled", when it actually means "use the fuse layer default". A short comment in the test body explaining the intent would help future readers.

---

### Summary

Implementation is complete. Five test cases cover the full contract (default/set/disable/negative/webdav). No blocking issues.

srstack (approved):
LGTM. No blocking issues. Minor nits noted in the review comment above.

PR #567, opened 2026-06-16
feat: add CLI TiDB Cloud native routing

srstack:
Summary:
- Add `drive9 region list` with manifest-backed routing and built-in list fallback.
- Add CLI create/delete support for TiDB Cloud native public/private keys and env fallbacks.
- Keep default anonymous create behavior unchanged and document provisioning/delete usage in README and site skill.

Tests:
- `env GOCACHE=/tmp/drive9-gocache go test ./cmd/drive9/cli ./cmd/drive9 ./pkg/client`
- `make lint`
- `git diff --check`

## Release Notes

* **New Features**
  * Added `drive9 region list/ls` to view available regions, modes, and servers (supports `--json` and falls back to an embedded manifest when unreachable).
  * Enhanced `drive9 create` and `drive9 delete` for TiDB Cloud native/starter provisioning, with `--server` taking precedence and native JSON request handling.
  * Extended `drive9 ctx` to store and display `mode`, `cloud_provider`, and `region` metadata.
* **Documentation**
  * Updated Quick Start and API/CLI environment variable guidance, including `tidb_cloud_native` naming and native delete request requirements.
* **Tests**
  * Added coverage for region listing and create/delete routing/authorization behavior.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,6 +140,24 @@ DRIVE9_PRIVATE_KEY="$TIDBCLOUD_PRIVATE_KEY" \
 drive9 create --name dev-native
 ```
 
+For TiDB Cloud native tenants, `drive9 create` stores the local context with
+`mode=TiDBCloud` and records `cloud_provider` / `region` when the server returns
+them. `drive9 ctx list` shows `MODE`, `CLOUD_PROVIDER`, and `REGION` by default;
+older contexts with no mode keep those fields empty.
+
+To add an existing owner API key for a native tenant manually, include the same
+metadata explicitly:
+
+```bash
+drive9 ctx add \
+  --name dev-native \
+  --server https://native-sg.drive9.ai \
+  --api-key "$DRIVE9_API_KEY" \
+  --mode TiDBCloud \
+  --cloud-provider aws \
+  --region us-east-1
+```
+
 `--server` has the highest priority. When it is set, the CLI bypasses the region
 manifest, ignores `--region-code`, and sends the request to that server directly:
 
--- a/README.md
+++ b/README.md
@@ -77,6 +77,91 @@ drive9 create --name dev --server https://api.drive9.ai
 drive9 ctx use dev
 ```
 
+### Provision Tenants
+
+List Drive9 provisioning regions:
+
+```bash
+drive9 region list
+```
+
+The text output uses user-facing mode names:
+
+```text
+REGION              MODE       SERVER
+aws-ap-southeast-1  Anonymous  https://api.drive9.ai
+aws-ap-southeast-1  TiDBCloud  https://native-sg.drive9.ai
+```
+
+`Anonymous` maps to the `tidb_cloud_starter` provider mode. `TiDBCloud` maps to
+the `tidbcloud_native` provider mode. If the public region manifest cannot be
+reached, `drive9 region list` falls back to:
+
+```text
+REGION              MODE       SERVER
+aws-ap-southeast-1  Anonymous  https://api.drive9.ai
+```
+
+Create an anonymous Starter tenant by region code:
+
+```bash
+drive9 create --region-code aws-ap-southeast-1 --name dev
+```
+
+Create a TiDB Cloud native tenant in the customer's TiDB Cloud account by
+passing the customer's TiDB Cloud API keys:
+
+```bash
+drive9 create \
+  --region-code aws-ap-southeast-1 \
+  --tidbcloud-public-key "$TIDBCLOUD_PUBLIC_KEY" \
+  --tidbcloud-private-key "$TIDBCLOUD_PRIVATE_KEY" \
+  --name dev-native
+```
+
+The same values can come from environment variables:
+
+```bash
+DRIVE9_REGION_CODE=aws-ap-southeast-1 \
+DRIVE9_PUBLIC_KEY="$TIDBCLOUD_PUBLIC_KEY" \
+DRIVE9_PRIVATE_KEY="$TIDBCLOUD_PRIVATE_KEY" \
+drive9 create --name dev-native
+```
+
+`--server` has the highest priority. When it is set, the CLI bypasses the region
+manifest and sends the request to that server directly:
+
+```bash
+drive9 create \
+  --server https://api.drive9.ai \
+  --tidbcloud-public-key "$TIDBCLOUD_PUBLIC_KEY" \
+  --tidbcloud-private-key "$TIDBCLOUD_PRIVATE_KEY" \
+  --name dev-native
+```
+
+Delete the current tenant with the current owner context:
+
+```bash
+drive9 delete
+```
+
+For `tidbcloud_native`, pass the customer's TiDB Cloud API keys again so the
+server can delete the customer-account Starter cluster:
+
+```bash
+drive9 delete \
+  --tidbcloud-public-key "$TIDBCLOUD_PUBLIC_KEY" \
+  --tidbcloud-private-key "$TIDBCLOUD_PRIVATE_KEY"
+```
+
+`drive9 delete` only uses owner API keys from local context automatically.
+Delegated and `fs_scoped` contexts are rejected for tenant deletion. Use
+`--api-key` to pass an owner key explicitly:
+
+```bash
+drive9 delete --server https://api.drive9.ai --api-key "$DRIVE9_API_KEY"
+```
+
 ### Filesystem Commands
 
 ```bash
@@ -249,7 +334,8 @@ DRIVE9_TENANT_PROVIDER=db9 | tidb_zero | tidb_cloud_starter | tidbcloud_native
 - `tidbcloud_native`: creates TiDB Cloud Serverless Starter clusters in the
   customer's TiDB Cloud account using `public_key` / `private_key` submitted to
   `POST /v1/provision`; server config supplies the API URL, cloud provider,
-  region, and optional default database name.
+  region, and default database name. The CLI does not expose a database-name
+  option for customers.
 - `db9`: supported provider path with its own schema, but not the default
   architecture used in this README.
 
@@ -416,7 +502,8 @@ DRIVE9_TIDBCLOUD_NATIVE_CLOUD_PROVIDER
                                 cloud provider for tidbcloud_native cluster creation
 DRIVE9_TIDBCLOUD_NATIVE_REGION  region for tidbcloud_native cluster creation
 DRIVE9_TIDBCLOUD_NATIVE_DEFAULT_DATABASE_NAME
-                                optional default database_name for tidbcloud_native
+                                server-side default database name for tidbcloud_native
+                                when unset, the server uses tidbcloud_fs
 DRIVE9_TOKEN_SIGNING_KEY        32-byte hex JWT signing key
 DRIVE9_ENCRYPT_TYPE             local_aes | kms
 DRIVE9_MASTER_KEY               local AES key
--- a/README.md
+++ b/README.md
@@ -102,16 +102,13 @@ REGION              MODE       SERVER
 aws-ap-southeast-1  Anonymous  https://api.drive9.ai
 ```
 
-`drive9 region list --json` returns the same region list as a JSON array. The
-`mode` field uses canonical IDs (`tidb_cloud_starter`, `tidb_cloud_native`), and
-`mode_label` contains the text display label:
+`drive9 region list --json` returns the same region list as a JSON array:
 
 ```json
 [
   {
     "region_code": "aws-ap-southeast-1",
-    "mode": "tidb_cloud_starter",
-    "mode_label": "Anonymous",
+    "mode": "Anonymous",
     "server_url": "https://api.drive9.ai"
   }
 ]
--- a/README.md
+++ b/README.md
@@ -102,6 +102,9 @@ REGION              MODE       SERVER
 aws-ap-southeast-1  Anonymous  https://api.drive9.ai
 ```
 
+`drive9 region list --json` uses canonical mode IDs (`tidb_cloud_starter`,
+`tidb_cloud_native`) instead of the display labels shown in text output.
+
 Create an anonymous Starter tenant by region code:
 
 ```bash
@@ -129,7 +132,7 @@ drive9 create --name dev-native
 ```
 
 `--server` has the highest priority. When it is set, the CLI bypasses the region
-manifest and sends the request to that server directly:
+manifest, ignores `--region-code`, and sends the request to that server directly:
 
 ```bash
 drive9 create \
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ aws-ap-southeast-1  TiDBCloud  https://native-sg.drive9.ai
 ```
 
 `Anonymous` maps to the `tidb_cloud_starter` provider mode. `TiDBCloud` maps to
-the `tidbcloud_native` provider mode. If the public region manifest cannot be
+the `tidb_cloud_native` provider mode. If the public region manifest cannot be
 reached, `drive9 region list` falls back to:
 
 ```text
@@ -145,7 +145,7 @@ Delete the current tenant with the current owner context:
 drive9 delete
 ```
 
-For `tidbcloud_native`, pass the customer's TiDB Cloud API keys again so the
+For `tidb_cloud_native`, pass the customer's TiDB Cloud API keys again so the
 server can delete the customer-account Starter cluster:
 
 ```bash
@@ -326,12 +326,12 @@ func main() {
 Current tenant providers are selected explicitly:
 
 ```text
-DRIVE9_TENANT_PROVIDER=db9 | tidb_zero | tidb_cloud_starter | tidbcloud_native
+DRIVE9_TENANT_PROVIDER=db9 | tidb_zero | tidb_cloud_starter | tidb_cloud_native
 ```
 
 - `tidb_zero`: default development/provisioning path backed by TiDB Zero.
 - `tidb_cloud_starter`: production-oriented TiDB Cloud Starter pool takeover.
-- `tidbcloud_native`: creates TiDB Cloud Serverless Starter clusters in the
+- `tidb_cloud_native`: creates TiDB Cloud Serverless Starter clusters in the
   customer's TiDB Cloud account using `public_key` / `private_key` submitted to
   `POST /v1/provision`; server config supplies the API URL, cloud provider,
   region, and default database name. The CLI does not expose a database-name
@@ -456,7 +456,7 @@ DELETE /v1/tenant                 delete current tenant with owner API key
 Vault endpoints also exist for secret storage, scoped grants, delegated tokens,
 and read-only vault mounts.
 
-`DELETE /v1/tenant` is owner-key only. For `tidbcloud_native`, the request body
+`DELETE /v1/tenant` is owner-key only. For `tidb_cloud_native`, the request body
 must also include the customer's TiDB Cloud `public_key` and `private_key` so
 Drive9 can delete the customer-account Starter cluster. `tidb_zero` tenants are
 not deleted through this endpoint; they expire automatically.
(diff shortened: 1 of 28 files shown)


srstack (commented):
## Review: feat: add CLI TiDB Cloud native routing

Good overall design. The region manifest pattern, env-var consumption semantics (`consumeEnv` for sensitive keys, `readTrimEnv` for region code), fallback-on-unavailability behavior, and validation pipeline are all cleanly implemented. Test coverage is thorough. A few items worth addressing:

---

### Significant

**`Create` silently ignores `--region-code` when `--server` is also provided**

When both flags are present, `resolveProvisionServer` returns early on `serverFlag` and `regionCode` is never used for routing. The JSON output omits `region_code` (via `omitempty`), so there's no visible signal that the flag was dropped. A user who passes both expecting the region code to select the server won't get an error or warning.

```go
if strings.TrimSpace(serverFlag) == "" {
    regionCode = strings.TrimSpace(regionCodeFlag)
    ...
}
```

Consider either: (a) returning an error when both `--server` and `--region-code` are set ("--server and --region-code are mutually exclusive; use one or the other"), or (b) at minimum printing a warning to stderr. Option (a) is cleaner.

---

**`DeleteTenant` silently discards the JSON decode error on failure responses**

```go
_ = json.NewDecoder(resp.Body).Decode(&result)
if resp.StatusCode != http.StatusAccepted {
    return fmt.Errorf("delete tenant failed (HTTP %d): %s", resp.StatusCode, result.Error)
}
```

If the server returns a non-JSON body (load balancer 502, HTML error page), `result.Error` is empty and the message becomes `"delete tenant failed (HTTP 502): "` — the empty colon at the end signals a decode failure but gives the user nothing actionable. At minimum, include the raw body in the error:

```go
raw, _ := io.ReadAll(resp.Body)
_ = json.Unmarshal(raw, &result)
if resp.StatusCode != http.StatusAccepted {
    if result.Error != "" {
        return fmt.Errorf("delete tenant failed (HTTP %d): %s", resp.StatusCode, result.Error)
    }
    return fmt.Errorf("delete tenant failed (HTTP %d): %s", resp.StatusCode, strings.TrimSpace(string(raw)))
}
```

---

### Minor

**`resolveProvisionServer` passes `context.Background()` — not cancelable by the user**

`fetchRegionManifest` creates its own 10-second timeout from the passed context, so a Ctrl-C during manifest fetch can leave the CLI hanging for up to 10 seconds before the timeout fires. For a CLI tool this is noticeable. Consider threading the `os.Signal` through a top-level context, or at minimum noting this as a known UX limitation.

---

**`fallbackRegionManifestCopy` shallow-copies slice elements**

```go
manifest.Regions = append([]RegionManifestEntry(nil), fallbackRegionManifest.Regions...)
```

`RegionManifestEntry` has `Tags []string` and `Metadata map[string]string`. The copy is element-by-element (struct copy), so `Tags` and `Metadata` would share the same backing arrays/maps as the originals. The current fallback entry has neither, so this is harmless right now — but worth fixing before someone adds a fallback entry with tags or metadata.

---

**`validateRegionManifest` is not called on the fallback manifest in `regionListCmd`**

```go
manifest, err := fetchRegionManifest(...)
if err != nil {
    ...
    manifest = fallbackRegionManifestCopy()  // bypasses validateRegionManifest
}
sortRegionManifestEntries(manifest.Regions)
```

The fallback is hardcoded and valid, so this is fine in practice. But it's slightly inconsistent — if someone ever modifies `fallbackRegionManifest` and introduces a bad entry, there's no guard. Could call `validateRegionManifest` on the copy too, panicking on failure (it's a coding error, not a runtime error).

---

### Nits

- `RegionModeTiDBCloudStarter = "tidbcloud_starter"` vs `RegionModeTiDBCloudStarterLegacy = "tidb_cloud_starter"` — the "new" canonical form (`tidbcloud_starter`) is used in `provisionModeForCredentials` return value, but the manifest currently only uses the legacy form. The normalization in `selectRegionServer` handles the mismatch correctly, but it's easy to get confused when reading the constants. A comment on `RegionModeTiDBCloudStarter` explaining that it's the normalized form of `RegionModeTiDBCloudStarterLegacy` would help.

- `deleteUsage()` returns `"usage: drive9 delete ..."` but `regionUsage()` returns a multi-line string with a subcommand description. Fine as-is since `delete` has no subcommands, just noting the style difference.

---

Overall: clean implementation with good test coverage. The `--server` + `--region-code` mutual exclusivity check and the error body fallback in `DeleteTenant` are the two items worth fixing before merge.

mornyx (changes requested):
Thanks for the PR. I found one blocker before this can be approved.

Blocker: TiDB Cloud credential env vars are not scrubbed from child process environments.

This PR adds `DRIVE9_PUBLIC_KEY` / `DRIVE9_PRIVATE_KEY` in `cmd/drive9/cli/db.go`, and `DRIVE9_PRIVATE_KEY` is secret-bearing. The existing credential contract in `cmd/drive9/cli/credentials.go` says new credential env vars must update the scrub path, but `scrubDrive9CredEnv` still only removes `DRIVE9_API_KEY`, `DRIVE9_VAULT_TOKEN`, and `DRIVE9_SERVER`. `mountBackgroundEnv` has the same issue.

Impact: a user with `DRIVE9_PRIVATE_KEY` in their shell can leak it into `drive9 vault with ... -- <cmd>` child processes, and also into background mount child processes.

Please add both new env vars to the scrub logic and tests before merging.

Tests I ran:
- `env GOCACHE=/private/tmp/drive9-pr567/.codex-gocache go test ./cmd/drive9/cli ./cmd/drive9` passed.
- `go test ./pkg/client` was not runnable in this environment because package TestMain requires testcontainers/rootless Docker.

mornyx (changes requested):
I re-reviewed the latest head (`e67b36f`). Thanks for addressing several of the earlier routing/error-handling items, but I still see blockers.

1. **Blocker: TiDB Cloud credential env vars are still not scrubbed from child process environments.**

   `DRIVE9_PUBLIC_KEY` / `DRIVE9_PRIVATE_KEY` are introduced as TiDB Cloud credential env vars in `cmd/drive9/cli/db.go`, and `DRIVE9_PRIVATE_KEY` is secret-bearing. The existing credential contract in `cmd/drive9/cli/credentials.go` explicitly says new credential env vars must update the child-env scrub path. However:

   - `scrubDrive9CredEnv` still only drops `DRIVE9_API_KEY`, `DRIVE9_VAULT_TOKEN`, and `DRIVE9_SERVER`.
   - `mountBackgroundEnv` also only strips those same three names.

   Impact: a shell-level `DRIVE9_PRIVATE_KEY` can still leak into `drive9 vault with ... -- <cmd>` children and background mount child processes. Please add both new env vars to these scrub paths and cover them in tests.

2. **Blocker: `ProviderTiDBCloudNative` was renamed from `tidbcloud_native` to `tidb_cloud_native` without backward compatibility for persisted data/config.**

   `pkg/tenant/provider.go` now only accepts the new string in `NormalizeProvider`. Existing tenants/configs created before this PR can already contain `provider = "tidbcloud_native"`. After this change, those values are no longer accepted by `NormalizeProvider`, and raw comparisons such as tenant delete checks will not recognize existing native tenants as native anymore.

   Concrete impact examples:
   - `DRIVE9_TENANT_PROVIDER=tidbcloud_native` deployments start failing provider normalization.
   - Existing DB rows with `provider = "tidbcloud_native"` will not match `tenant.ProviderTiDBCloudNative`, so delete/deprovision logic can reject them as unsupported or skip native credential handling.
   - `SmallInDB` / `UsesTiDBAutoEmbedding` also stop recognizing the old persisted value if it reaches those helpers unnormalized.

   Please either keep the persisted provider value stable, or add an explicit compatibility path for the old value and ensure DB-loaded tenants with the legacy string still behave as native tenants. A `NormalizeProvider` alias alone is only sufficient if all DB-loaded provider values are normalized before every comparison.

Tests I ran:
- `env GOCACHE=/private/tmp/drive9-pr567-review2/.codex-gocache go test ./cmd/drive9/cli ./cmd/drive9` passed.
- `env GOCACHE=/private/tmp/drive9-pr567-review2/.codex-gocache go test ./pkg/tenant ./pkg/tenant/schemadump` could not complete in this environment because `pkg/tenant` TestMain requires testcontainers/rootless Docker; `pkg/tenant/schemadump` has no test files.

srstack (commented):

srstack (commented):
## Re-review: feat: add CLI TiDB Cloud native routing (updated)

Reviewed the updated diff. Here is what changed and what remains.

---

### Fixed since last review

**`DeleteTenant` JSON decode fallback** (previously marked Significant)

Now correctly implemented with three-level fallback:

```go
rawResp, err := io.ReadAll(resp.Body)
_ = json.Unmarshal(rawResp, &result)
if resp.StatusCode != http.StatusAccepted {
    msg := strings.TrimSpace(result.Error)
    if msg == "" { msg = strings.TrimSpace(string(rawResp)) }
    if msg == "" { msg = http.StatusText(resp.StatusCode) }
    return fmt.Errorf("delete tenant failed (HTTP %d): %s", resp.StatusCode, msg)
}
```

`TestDeleteTenantFallsBackToRawErrorBody` covers the non-JSON body case. ✓

**Provider naming consistency** (previously marked Nit)

`ProviderTiDBCloudNative` renamed from `"tidbcloud_native"` to `"tidb_cloud_native"`, consistent with `tidb_cloud_starter`. Updated across `provider.go`, `schema_test.go`, `drive9-server/main.go`, `tenant_delete.go`, and `README.md`. ✓

---

### Still present

**`--server` and `--region-code` together silently ignores `--region-code`** (flagged in previous review)

When `--server` is set, `regionCode` is left as `""` and `region_code` is omitted from JSON output. The user gets no warning. `TestCreateServerOverrideSendsNativeBodyAndSkipsManifest` explicitly asserts that `region_code` is absent from output — this is the intended design. At minimum, the README or usage string should say the two flags are mutually exclusive so users are not surprised.

---

### New findings

**`selectRegionServer` does exact mode matching — no legacy normalization**

The previous version of this PR had a `normalizeRegionMode` function that mapped `"tidbcloud_starter"` → `"tidb_cloud_starter"`. This version removes that function entirely: both constants are now `"tidb_cloud_*"` form, and `selectRegionServer` does a direct string comparison. `TestSelectRegionServerDoesNotAcceptLegacyNativeMode` explicitly rejects the old `"tidbcloud_native"` form. This is intentional — the live manifest must use the new naming. As long as the manifest has been updated before this CLI ships, there is no issue. Worth confirming that is the case.

**`--json` mode outputs raw mode IDs; text mode outputs display labels**

Text output: `Anonymous` / `TiDBCloud` (via `regionModeLabel`)
JSON output: `tidb_cloud_starter` / `tidb_cloud_native` (raw manifest values)

This is the right design — machine-readable output should use canonical IDs. However, the README example only shows text output and never mentions that `--json` produces different `mode` values. A one-liner like "JSON output uses canonical mode IDs (`tidb_cloud_starter`, `tidb_cloud_native`)" in the README would prevent confusion.

---

### Summary

The two Significant issues from the last review are resolved. This version is ready to merge. Two optional improvements before or after merge:

1. Document that `--server` takes priority over `--region-code` and effectively ignores it.
2. Note in README that `--json` mode uses canonical mode IDs rather than display labels.

srstack:
Addressed the latest review feedback in `c17ab83`:

- Added `DRIVE9_PUBLIC_KEY` / `DRIVE9_PRIVATE_KEY` to the credential-bearing env constants.
- Updated `scrubDrive9CredEnv` so `drive9 vault with ... -- <cmd>` children do not inherit either TiDB Cloud key.
- Updated `mountBackgroundEnv` so background mount child processes also strip both TiDB Cloud key env vars.
- Added coverage in `TestScrubDrive9CredEnv` and `TestMountBackgroundEnvSnapshotsCredentials`.
- Documented that `--server` bypasses region manifest routing and ignores `--region-code`.
- Documented that `drive9 region list --json` emits canonical mode IDs (`tidb_cloud_starter`, `tidb_cloud_native`) rather than text labels.

On the `tidbcloud_native` legacy provider string: this PR intentionally does not keep a compatibility alias. The provider/mode identifier is being hard-renamed to `tidb_cloud_native` per product decision, and all config/manifest/docs are expected to use the canonical value.

mornyx (changes requested):
I re-reviewed latest head `a301fa8`.

The previous credential-env leak is fixed: `vault with` and background mount now scrub `DRIVE9_PUBLIC_KEY` / `DRIVE9_PRIVATE_KEY`.

Findings:

1. Blocker - the provider ID rename still breaks existing `tidbcloud_native` configs and persisted tenants. `pkg/tenant/provider.go:9` changes the canonical value from the existing `tidbcloud_native` to `tidb_cloud_native`, while `NormalizeProvider`, `SmallInDB`, and `UsesTiDBAutoEmbedding` only recognize the new spelling. Existing deploys with `DRIVE9_TENANT_PROVIDER=tidbcloud_native` will fail at server startup (`cmd/drive9-server/main.go:149-163`). More importantly, existing tenant rows can already have `provider = "tidbcloud_native"`; when those are loaded, paths that compare the raw field directly will now misclassify them. Examples: `pkg/tenant/pool.go:568`, `pkg/tenant/pool.go:666`, and `pkg/server/tenant_delete.go:64` / `:94`. That can disable DB auto-embedding/small-file-in-DB behavior or reject tenant deletion for existing native tenants. Please keep the persisted provider value stable, or add a legacy alias plus normalize persisted provider values before every classifier/direct comparison. A `NormalizeProvider` alias by itself is not enough unless DB-loaded values go through it before use.

2. Significant - `drive9 region list --json` now writes display labels into the `mode` field. `cmd/drive9/cli/region.go:133-136` sets JSON `mode` to `regionModeLabel(entry.Mode)`, so machines get `Anonymous` / `TiDBCloud` instead of the canonical route IDs `tidb_cloud_starter` / `tidb_cloud_native`. This is inconsistent with the manifest and with `drive9 create --json` (`cmd/drive9/cli/db.go:174-188`), and it prevents simple automation from using `region list --json` as the source for provisioning mode decisions. Please keep `mode` canonical and add a separate `mode_label` or display field for the table label.

Tests run:
- `env GOCACHE=/private/tmp/drive9-pr567-review3/.codex-gocache go test ./cmd/drive9/cli ./cmd/drive9`

I cannot approve yet because of the provider compatibility issue.

mornyx (changes requested):
I re-reviewed latest head `c65297f`.

Findings:

1. Blocker - the provider ID rename is still not backward compatible with existing `tidbcloud_native` deployments or persisted tenants. `pkg/tenant/provider.go:9` still changes the canonical value to `tidb_cloud_native`, and `NormalizeProvider` only accepts that new spelling (`pkg/tenant/provider.go:12-18`). Existing deployments using `DRIVE9_TENANT_PROVIDER=tidbcloud_native` will fail during startup normalization (`cmd/drive9-server/main.go:149-163`). Existing tenant rows with `provider = "tidbcloud_native"` are still read as raw strings and then compared directly in hot paths such as `pkg/tenant/pool.go:568`, `pkg/tenant/pool.go:666`, and `pkg/server/tenant_delete.go:64` / `:94`, so they can be misclassified as non-native. Please keep the persisted provider ID stable, or add a legacy alias and normalize DB-loaded provider values before every classifier/direct comparison. Add regression coverage for both legacy env config and a legacy persisted tenant row.

2. Significant - machine-readable JSON now uses display labels for provisioning `mode`. `drive9 region list --json` sets `mode` via `regionModeLabel(entry.Mode)` (`cmd/drive9/cli/region.go:133-136`), and `drive9 create --json` now also emits `regionModeLabel(mode)` (`cmd/drive9/cli/db.go:180-188`). Tests enforce `TiDBCloud` / `Anonymous` in JSON (`cmd/drive9/cli/region_test.go:69-73`, `cmd/drive9/cli/db_test.go:174-175`). That loses the canonical route IDs `tidb_cloud_native` / `tidb_cloud_starter`, makes JSON differ from the manifest and `selectRegionServer`, and makes scripts unable to feed `mode` back into routing logic. It is fine for local context metadata or table output to use display labels, but JSON provisioning fields should keep `mode` canonical and expose display text as a separate `mode_label` or `display_mode` field.

The new context metadata path itself looks reasonable as display metadata, but it does not resolve the two API/compatibility issues above.

Tests run:
- `env GOCACHE=/private/tmp/drive9-pr567-review4/.codex-gocache go test ./cmd/drive9/cli ./cmd/drive9`

I still cannot approve this PR until the provider compatibility issue is fixed.

srstack:
@mornyx Thanks for the re-review. Both points are intentional for this PR and match the current product decision.

1. `tidbcloud_native` -> `tidb_cloud_native` is a hard rename. We do not need backward compatibility here because there are currently no existing native tenants/clusters or persisted `provider = "tidbcloud_native"` rows to migrate. New config, manifests, and persisted rows should use only `tidb_cloud_native`.

2. The CLI JSON mode value is intentionally user-facing (`Anonymous` / `TiDBCloud`) for this iteration. The region manifest remains the internal routing source with canonical mode IDs (`tidb_cloud_starter` / `tidb_cloud_native`), but `drive9 region list --json` and `drive9 create --json` are intended to return the same mode labels users see in the table output. We are not adding a separate `mode_label` / `display_mode` field right now.

mornyx:
I checked PR 567 again. There are no new commits since my previous review; the latest head is still `c65297f`.

No new code to review in this pass. The previous request-changes review still applies:

1. The provider ID rename is still incompatible with existing `tidbcloud_native` env/config and persisted tenant rows.
2. The machine-readable provisioning `mode` fields still use display labels instead of canonical route IDs.

No additional tests were run in this pass because the code has not changed since the last tested head.

mornyx (approved):
Thanks for the clarification. I am updating my review stance.

I withdraw the request-change position from my previous review. The provider rename finding depended on an assumption that existing native tenants or persisted `provider = "tidbcloud_native"` rows could already exist. Given the stated product/data assumption that there are no existing native tenants/clusters or legacy persisted rows to migrate, the hard rename to `tidb_cloud_native` is acceptable for this PR.

I also accept the JSON `mode` shape as a product decision: `drive9 region list --json` and `drive9 create --json` are UX-facing CLI outputs that intentionally return the same display labels as the table output, while the region manifest remains the internal routing contract with canonical IDs. My remaining non-blocking suggestion is to keep the docs explicit that JSON `mode` is a display label and should not be treated as the manifest routing ID.

No new tests were run in this update because the head is still `c65297f`; the last tested command was:
- `env GOCACHE=/private/tmp/drive9-pr567-review4/.codex-gocache go test ./cmd/drive9/cli ./cmd/drive9`

Approved based on the clarified product/data assumptions.

srstack (commented):
## Re-review: 8 new commits since last review

All issues flagged in the previous review have been addressed. This review covers the new additions.

### Previously flagged issues — resolved

- **JSON output mode labels**: `regionListOutput()` now converts raw provider IDs to display labels (`Anonymous`/`TiDBCloud`) before encoding. The `--json` output no longer exposes internal IDs.
- **`--server` vs `--region-code` documentation**: README and `site/skill.md` now explicitly state that `--server` has the highest priority and bypasses region manifest routing, ignoring `--region-code`.
- **`DeleteTenant` JSON decode fallback**: `io.ReadAll` + three-level fallback is properly implemented.
- **Provider naming**: `ProviderTiDBCloudNative = "tidb_cloud_native"` is now consistent throughout — `pkg/tenant/provider.go`, `tenant_delete.go`, `schemadump.go`, test error messages, and comments all use the underscore form.

### New additions

**Context metadata fields (`Mode`/`CloudProvider`/`Region`)**

The implementation is clean and well-validated:
- `ctx add` validation correctly rejects `--cloud-provider`/`--region` without `--mode`, and rejects `--cloud-provider`/`--region` when mode is `Anonymous` (starter mode doesn't carry cloud-specific routing metadata).
- Allowing `--mode TiDBCloud` without `--cloud-provider`/`--region` is intentional and correct — the server may not return them, so manual `ctx add` of an existing native key should still be supported.
- `isKnownContextMode` validates against the exact display label set (`Anonymous`/`TiDBCloud`), keeping the context file decoupled from internal provider IDs.
- Tests cover the happy path, rejection of ambiguous metadata, and omission of empty fields in JSON output.

**Credential scrubbing**

`DRIVE9_PUBLIC_KEY`/`DRIVE9_PRIVATE_KEY` are now scrubbed in both `scrubDrive9CredEnv` (used by `mount` background daemon) and `mountBackgroundEnv`. The comment in `secret.go` is also updated to list all five scrubbed variables. This matches the spec §9 requirement and closes the gap where TiDB Cloud keys could leak into subprocesses.

**Server-side `cloud_provider`/`region` in provision response**

The `provisioningRegionProvider` interface is a good backward-compatible approach — existing provisioners that don't implement it get empty strings, no breakage. `strings.TrimSpace` in `provisioningCloudRegion` is a small defensive touch.

**`RawDelete` on `*client.Client`**

Follows the exact same pattern as `RawPost`. The test in `auth_test.go` verifies method, path, auth header, content type, and body — adequate coverage.

**`drive9 delete` and `drive9 region` dispatch**

Dispatch handlers follow the existing pattern exactly. The `region` subcommand logging mirrors `git`'s sub-extraction pattern. Tests for both verbs follow the same handler-injection pattern as `TestDispatchTokenVerbReachesHandler`.

### Minor observations

- `readCacheNoExpiryTTL = -1` in `pkg/fuse/read.go` is declared as an untyped constant, which means it implicitly becomes `int` in some contexts. Declaring it as `time.Duration = -1` would make the type explicit and avoid potential surprises if someone passes it directly to a function expecting `time.Duration` without a cast. Not blocking, but worth noting.
- The `regionListOutputEntry` struct in `region.go` duplicates the fields of `RegionManifestEntry` with only the `Mode` field semantics changed. If the manifest entry shape ever gains new fields, both structs need updating. A conversion function (already present as `regionListOutput`) mitigates this, but the struct duplication is still worth keeping in mind.

### Summary

The PR is well-structured and the implementation is solid. All previously flagged issues are resolved, new features are properly validated, credentials are correctly scrubbed, and test coverage is thorough. Ready for merge pending approval from another reviewer (cannot self-approve).

PR #570, opened 2026-06-16
feat: create native tenant system user

srstack:
## Summary
- change TiDB Cloud native cluster display name to `tidbcloud-fs-<tenant_id>`
- create the `tidbcloud_fs_admin` role and `<prefix>.tidbcloud_fs_system` user during async schema initialization
- keep metadata on `<prefix>.root` until the async role/user setup succeeds, then switch `db_user`/`db_password` to the restricted system user

## Review notes
- The system user reuses the current root DSN password. This keeps concurrent/retried `ALTER USER` calls deterministic and avoids DB/metadata password skew.
- The role is granted only database-scoped privileges on the tenant DB.
- Resume is idempotent: if metadata still has `<prefix>.root`, setup is retried; if metadata already has `<prefix>.tidbcloud_fs_system`, finalize treats it as complete. Unexpected native usernames are rejected.
- Metadata credential persistence uses a compare-and-swap update on the previous `db_user`.

## Tests
- `env GOCACHE=/tmp/drive9-gocache go test ./cmd/drive9-server ./pkg/tenant/tidbcloudnative ./pkg/server ./pkg/meta`
- `make lint`
- `git diff --check`

### DIFF
--- a/pkg/meta/meta.go
+++ b/pkg/meta/meta.go
@@ -1658,7 +1658,7 @@ func (s *Store) UpdateTenantConnection(ctx context.Context, id string, cluster *
 
 func (s *Store) UpdateTenantDBCredentialIf(ctx context.Context, id, fromDBUser, dbUser string, dbPasswordCipher []byte) (updated bool, err error) {
 	start := time.Now()
-	defer observeMeta(ctx, "update_tenant_db_credential", start, &err)
+	defer observeMeta(ctx, "update_tenant_db_credential_if", start, &err)
 	res, err := s.db.ExecContext(ctx, `UPDATE tenants SET db_user = ?, db_password = ?, updated_at = ? WHERE id = ? AND db_user = ?`,
 		dbUser, dbPasswordCipher, time.Now().UTC(), id, fromDBUser)
 	if err != nil {
--- a/pkg/meta/meta.go
+++ b/pkg/meta/meta.go
@@ -1656,19 +1656,16 @@ func (s *Store) UpdateTenantConnection(ctx context.Context, id string, cluster *
 	return nil
 }
 
-func (s *Store) UpdateTenantDBCredential(ctx context.Context, id, dbUser string, dbPasswordCipher []byte) (err error) {
+func (s *Store) UpdateTenantDBCredentialIf(ctx context.Context, id, fromDBUser, dbUser string, dbPasswordCipher []byte) (updated bool, err error) {
 	start := time.Now()
 	defer observeMeta(ctx, "update_tenant_db_credential", start, &err)
-	res, err := s.db.ExecContext(ctx, `UPDATE tenants SET db_user = ?, db_password = ?, updated_at = ? WHERE id = ?`,
-		dbUser, dbPasswordCipher, time.Now().UTC(), id)
+	res, err := s.db.ExecContext(ctx, `UPDATE tenants SET db_user = ?, db_password = ?, updated_at = ? WHERE id = ? AND db_user = ?`,
+		dbUser, dbPasswordCipher, time.Now().UTC(), id, fromDBUser)
 	if err != nil {
-		return err
+		return false, err
 	}
 	n, _ := res.RowsAffected()
-	if n == 0 {
-		return ErrNotFound
-	}
-	return nil
+	return n > 0, nil
 }
 
 func (s *Store) UpdateTenantBranch(ctx context.Context, id string, cluster *Tenant) (err error) {
--- a/pkg/meta/meta.go
+++ b/pkg/meta/meta.go
@@ -1656,6 +1656,21 @@ func (s *Store) UpdateTenantConnection(ctx context.Context, id string, cluster *
 	return nil
 }
 
+func (s *Store) UpdateTenantDBCredential(ctx context.Context, id, dbUser string, dbPasswordCipher []byte) (err error) {
+	start := time.Now()
+	defer observeMeta(ctx, "update_tenant_db_credential", start, &err)
+	res, err := s.db.ExecContext(ctx, `UPDATE tenants SET db_user = ?, db_password = ?, updated_at = ? WHERE id = ?`,
+		dbUser, dbPasswordCipher, time.Now().UTC(), id)
+	if err != nil {
+		return err
+	}
+	n, _ := res.RowsAffected()
+	if n == 0 {
+		return ErrNotFound
+	}
+	return nil
+}
+
 func (s *Store) UpdateTenantBranch(ctx context.Context, id string, cluster *Tenant) (err error) {
 	start := time.Now()
 	defer observeMeta(ctx, "update_tenant_branch", start, &err)
--- a/pkg/meta/meta_test.go
+++ b/pkg/meta/meta_test.go
@@ -57,6 +57,57 @@ func TestMetaDBMetrics(t *testing.T) {
 	}
 }
 
+func TestUpdateTenantDBCredentialIfComparesPreviousUser(t *testing.T) {
+	s := newControlStore(t)
+	now := time.Now().UTC()
+	if err := s.InsertTenant(context.Background(), &Tenant{
+		ID:               "credential-cas-tenant",
+		Status:           TenantProvisioning,
+		DBHost:           "127.0.0.1",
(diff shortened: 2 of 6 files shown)


mornyx (approved):

srstack (commented):
## Review: feat: create native tenant system user

### Overview

This PR introduces post-schema-init system user setup for TiDB Cloud Native tenants. After `InitSchema` succeeds, the server calls `EnsureSystemUser` to create a `tidbcloud_fs_admin` role and a `<prefix>.tidbcloud_fs_system` user, then replaces the root credential stored in tenant metadata with the restricted system user credential. The approach is idempotent: if the metadata already holds a non-root user, setup is skipped.

### What works well

- **Idempotency design** (`systemUsernameForCurrent`): detecting `.root` suffix and skipping when the system user is already in place is clean. Retry-on-failure is handled by the existing `initTenantSchemaAsync` loop, which now re-runs `finalizeTenantSchemaInit` on each retry.
- **Interface segregation** (`nativeSystemUserProvisioner`): the optional-interface pattern mirrors `provisioningRegionProvider` from PR #567 — consistent with the codebase convention, no breakage for other provisioners.
- **Credential encryption**: password is encrypted via `s.pool.Encrypt` before being persisted — not stored plaintext.
- **`quoteIdent`/`quoteString`**: backtick and single-quote escaping is correct, and the tests cover injection characters.
- **`UpdateTenantDBCredential`**: separate, targeted update avoids accidentally clobbering other tenant metadata fields. `ErrNotFound` on zero rows is correct.

### Issues

**1. `ALTER USER` unconditionally resets the password even when `CREATE USER IF NOT EXISTS` was a no-op**

`systemUserStatements` always emits both `CREATE USER IF NOT EXISTS` and `ALTER USER ... IDENTIFIED BY`. If `EnsureSystemUser` is called again after the system user already exists (e.g., a retry where `systemUsernameForCurrent` wrongly returns `needsSetup=true`), the password is rotated and the previously-persisted cipher in metadata becomes stale. In practice `systemUsernameForCurrent` only returns `needsSetup=true` when the DSN still contains `.root`, so the window is narrow, but it is worth documenting or guarding.

**2. `EnsureSystemUser` ignores the `tenantID` parameter**

The signature is `EnsureSystemUser(ctx, dsn, tenantID string)`. The implementation discards `tenantID` (argument named `_`). If a future change needs to correlate the system user with the tenant (e.g., audit logging, per-tenant naming), it will require a signature change. This is minor but the discarded parameter is unusual — either use it or document why it is reserved.

**3. `finalizeTenantSchemaInit` returns a hard error when the provisioner does not implement `nativeSystemUserProvisioner`**

```go
if !ok {
    return fmt.Errorf("native provisioner does not support system user setup")
}
```

This causes the schema-init retry loop to keep retrying forever (bounded by the existing attempt cap) even though no retry will ever succeed — the provisioner type is fixed at startup. A missing interface implementation is a deployment misconfiguration, not a transient error. Consider logging a fatal/critical event and returning a permanent sentinel, or checking this once at server startup rather than per-tenant.

**4. `generateRandomPassword` is not tested or visible in this diff**

`generateRandomPassword(24)` is called in `EnsureSystemUser` but the function is not in this diff. If it is already in the package, that is fine — but it would be worth confirming it uses a cryptographically secure source (`crypto/rand`), since it produces the credential that protects the tenant DB.

**5. `systemUserStatements` grants `CREATE, ALTER, DROP` to the system user**

These are fairly broad DDL privileges. If the system user's purpose is ongoing runtime access (reads/writes), granting `ALTER` and `DROP` is more than needed and increases blast radius if the credential is ever compromised. Consider whether `SELECT, INSERT, UPDATE, DELETE, INDEX` (plus `CREATE TABLE` if schema migrations run as this user) is sufficient.

### Minor observations

- `clusterDisplayName` rename from `drive9-` to `tidbcloud-fs-` prefix is clean; test updated accordingly.
- `normalizeDatabaseName` is called twice in `EnsureSystemUser` (indirectly via `cfg.DBName` parsing) — once to get `dbName` for the statements. The call is correct but there is no validation that the resulting `dbName` is non-empty before passing it to `systemUserStatements`. An empty `dbName` would produce a syntactically broken `GRANT ... ON .* TO ...` statement.

### Summary

The core design is sound — idempotent system user creation, encrypted credential storage, optional-interface pattern. The main issues to address before merge:

1. Document or guard the `ALTER USER` password-rotation on retries.
2. Either use `tenantID` in `EnsureSystemUser` or remove it from the signature.
3. Handle the permanent-misconfiguration case in `finalizeTenantSchemaInit` rather than looping indefinitely.
4. Confirm `generateRandomPassword` uses `crypto/rand`.
5. Consider tightening the GRANT privilege set.

srstack (commented):
## Re-review

All five issues from the previous review have been addressed. Going through them in order:

**1. `ALTER USER` password rotation on retry — resolved**

The first review flagged that `EnsureSystemUser` always called `ALTER USER`, which would rotate the password on any retry. The updated implementation removes `generateRandomPassword` entirely and instead reuses the root password from the DSN (`password := cfg.Passwd`) when setting up the system user. Combined with the CAS update in `UpdateTenantDBCredentialIf`, a retry that finds `needsSetup=false` returns the existing credential without touching the database, and a retry that finds `needsSetup=true` uses the same root password — so no stale cipher is ever written.

**2. `tenantID` parameter discarded — acknowledged but still unused**

The parameter is still `_` in `EnsureSystemUser`. This is acceptable given the PR description notes it is reserved for future use, but it would be cleaner to either document this explicitly in the function signature comment or remove it from the interface until it is actually needed.

**3. Infinite retry on missing interface — partially addressed**

`finalizeTenantSchemaInit` now parses the DSN and validates `fromDBUser` before calling `EnsureSystemUser`, which is an improvement. However, the `!ok` path (`native provisioner does not support system user setup`) still returns a plain error that feeds back into the retry loop. This remains a deployment-time misconfiguration masquerading as a transient error. It will stop retrying only when the attempt cap is exhausted. Not a blocker, but still worth a follow-up.

**4. `generateRandomPassword` not using `crypto/rand` — resolved**

The revised implementation drops `generateRandomPassword` entirely and uses the root password from the DSN. This sidesteps the `crypto/rand` concern completely.

**5. Overly broad GRANT — unchanged**

`GRANT CREATE, ALTER, DROP, INDEX, SELECT, INSERT, UPDATE, DELETE` is still granted to the system role. This was flagged as a concern in the first review and has not changed. Acknowledging it is a pre-existing design decision if this matches the schema migration requirements, but it should be noted that `ALTER` and `DROP` give this role destructive DDL access to the tenant DB.

### New additions since last review

**`UpdateTenantDBCredentialIf` (CAS update)**

The compare-and-swap approach (`WHERE id = ? AND db_user = ?`) is the right fix. If another process already rotated the credential to the system user, the update is a no-op — no double-write, no credential loss. The `updated=false` path logs at Info level rather than erroring, which is correct since the desired end-state (system user in metadata) is already achieved. The test in `meta_test.go` covers both the mismatch (no-op) and match (update) cases cleanly.

**`quoteString` backslash escaping**

The updated `quoteString` now escapes `\` → `\\` before escaping `'` → `''`. This is correct for MySQL's default `sql_mode` (where backslash is an escape character). The test `TestSQLQuoting` verifies the combined case `u'ser\name` → `'u''ser\\name'`. Good catch.

**`systemUsernameForCurrent` now rejects unrecognized usernames**

The original implementation fell through to `return currentUsername, false, nil` for any username that didn't end in `.root`. The updated version explicitly rejects usernames that are neither `.root` nor `.tidbcloud_fs_system`, returning an error. This is safer — an unexpected username format surfaces early rather than silently being treated as already-finalized. Tests cover `root` and `u1.admin` as rejection cases.

**`finalizeTenantSchemaInit` DSN parsing**

Parsing the DSN in the server layer to extract `fromDBUser` for the CAS guard is a reasonable addition. It keeps the CAS anchor in the server layer where the metadata update happens, rather than requiring `EnsureSystemUser` to return it.

### Summary

The main structural issues are resolved. The two remaining minor points (unused `tenantID` parameter and broad DDL grants) are not blockers. LGTM pending those being tracked as follow-up items.

PR #572, opened 2026-06-16
feat: rename api key prefix from dat9_ to drive9_ with backward compatibility

srstack:
## Summary
Change the API key token prefix from `dat9_` to `drive9_` while maintaining full backward compatibility with existing `dat9_` tokens.

## Changes
- **`pkg/tenant/token/token.go`**: New tokens are issued with `drive9_` prefix. `ParseAndVerifyToken` now accepts both `drive9_` and legacy `dat9_` prefixes.
- **`cmd/drive9/cli/ctx.go`**: `decodeDrive9APIKeyPayload` accepts both `drive9_` and `dat9_` prefixes.
- **`pkg/tenant/token/token_test.go`**: Added `TestParseAndVerifyTokenAcceptsLegacyDat9Prefix` to verify legacy tokens are still accepted.

## Backward Compatibility
Existing `dat9_` tokens continue to work. The parser falls back to `dat9_` when the `drive9_` prefix is not matched.

## Summary by CodeRabbit

* **Bug Fixes**
  * Enhanced API key and token validation to support legacy credential formats alongside new ones, ensuring existing credentials continue to function during system transitions while new credentials use the updated format.

### DIFF
--- a/.github/workflows/dev-cd.yml
+++ b/.github/workflows/dev-cd.yml
@@ -81,7 +81,7 @@ jobs:
             --name ${{ env.EKS_CLUSTER }} \
             --region ${{ env.AWS_REGION }}
 
-      - name: Deploy
+      - name: Deploy to dat9
         run: |
           kubectl set image deployment/${{ env.K8S_DEPLOYMENT }} \
             ${{ env.K8S_DEPLOYMENT }}=${{ env.ECR_REGISTRY }}/${{ env.ECR_REPOSITORY }}:${{ env.IMAGE_TAG }} \
@@ -90,8 +90,23 @@ jobs:
             -n ${{ env.DEPLOY_NAMESPACE }} \
             --timeout=180s
 
-      - name: Rollback on failed deploy
+      - name: Deploy to drive9-tidbcloud
+        run: |
+          kubectl set image deployment/drive9-server \
+            drive9-server=${{ env.ECR_REGISTRY }}/${{ env.ECR_REPOSITORY }}:${{ env.IMAGE_TAG }} \
+            -n drive9-tidbcloud
+          kubectl rollout status deployment/drive9-server \
+            -n drive9-tidbcloud \
+            --timeout=180s
+
+      - name: Rollback dat9 on failed deploy
         if: failure()
         run: |
           kubectl rollout undo deployment/${{ env.K8S_DEPLOYMENT }} \
             -n ${{ env.DEPLOY_NAMESPACE }}
+
+      - name: Rollback drive9-tidbcloud on failed deploy
+        if: failure()
+        run: |
+          kubectl rollout undo deployment/drive9-server \
+            -n drive9-tidbcloud
--- a/cmd/drive9/cli/ctx.go
+++ b/cmd/drive9/cli/ctx.go
@@ -264,7 +264,10 @@ func writeCtxShowText(entry *ctxShowEntry) error {
 	return w.Flush()
 }
 
-const drive9APIKeyJWTWrapperPrefix = "dat9_"
+const (
+	drive9APIKeyJWTWrapperPrefix       = "drive9_"
+	legacyAPIKeyJWTWrapperPrefix       = "dat9_"
+)
 
 func tenantIDFromContext(ctx *Context) string {
 	if ctx == nil {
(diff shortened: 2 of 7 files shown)


srstack (commented):
## Review: feat: rename api key prefix from dat9_ to drive9_ with backward compatibility

### Overview

Small, focused change: new tokens are issued with `drive9_` prefix; `ParseAndVerifyToken` and `decodeDrive9APIKeyPayload` both accept either prefix. Backward compatibility is preserved for existing `dat9_` tokens in the wild.

### What works well

- Both parse paths (`token.go` and `ctx.go`) are updated consistently — no risk of one accepting legacy tokens while the other doesn't.
- The `switch` fallthrough pattern is clean and makes the precedence explicit: `drive9_` is tried first, `dat9_` as fallback.
- The test constructs a real `drive9_`-issued token, strips the new prefix, reattaches the legacy prefix, and verifies it round-trips — a tight, meaningful test of the actual backward-compat path.

### Issues

**1. No migration path or deprecation signal for legacy tokens**

The PR silently accepts `dat9_` tokens forever with no indication to the caller that they are using a deprecated format. Depending on how long migration is expected to take, it may be worth:
- Logging a warning when a `dat9_` token is accepted (at least at debug level), so operators can track how many legacy tokens are still in use.
- Documenting when `dat9_` support will be dropped.

This is not blocking, but worth an explicit decision either way.

**2. `decodeDrive9APIKeyPayload` in `ctx.go` uses `legacyAPIKeyJWTWrapperPrefix` independently from `legacyTokenPrefix` in `token.go`**

The two constants are defined separately (`legacyAPIKeyJWTWrapperPrefix = "dat9_"` in `ctx.go`, `legacyTokenPrefix = "dat9_"` in `token.go`). If the legacy prefix ever needs updating (e.g., a three-way migration), the two constants must be kept in sync manually. Exporting `LegacyTokenPrefix` from `pkg/tenant/token` and importing it in `ctx.go` would make them a single source of truth. Minor, but reduces future drift risk.

**3. No test for `decodeDrive9APIKeyPayload` legacy path in `ctx.go`**

`token_test.go` covers the `ParseAndVerifyToken` legacy path, but there is no corresponding test for the CLI-side `decodeDrive9APIKeyPayload` accepting a `dat9_`-prefixed API key. Given this is the function that validates keys entered by users (e.g., `drive9 ctx add --api-key dat9_...`), a test would be useful.

### Minor observations

- The trailing blank line inside the `const` block in `ctx.go` (between the two constants) is a minor style inconsistency with the rest of the file, which uses single-line const blocks.
- `IssueToken` is not tested here for producing a `drive9_` prefix. Adding `strings.HasPrefix(tok, "drive9_")` to `TestParseAndVerifyTokenAcceptsLegacyDat9Prefix` or a separate test would close the loop.

### Summary

The change is correct and backward-compatible. The main items to address: decide on a deprecation/logging strategy for legacy tokens, and add a test for the `ctx.go` legacy acceptance path.

mornyx (approved):
Review result: approved with one non-blocking follow-up.

I did not find a runtime blocker in the core token/auth paths. New tokens are issued with the `drive9_` wrapper, `ParseAndVerifyToken` accepts both `drive9_` and legacy `dat9_`, and server-side lookup still hashes the raw token supplied by the client, so existing persisted `dat9_` keys continue to resolve correctly.

One compatibility gap outside the changed files: `scripts/decode_drive9_config.py` still documents and enforces only the legacy `dat9_` prefix (`DRIVE9_API_KEY_PREFIX = "dat9_"`). After this PR, a newly issued `drive9_` API key stored in `~/.drive9/config` will produce a decode error in that helper. Since the script is part of the local debugging/user support surface for config tenant-id decoding, please consider updating it to accept both prefixes and update the docstring. I am not marking this as blocking because the main CLI `ctx show` path already handles both prefixes, but it is worth fixing in or immediately after this PR.

I also agree with the existing nit that the `ctx rm` help text should be prefix-agnostic instead of saying only `dat9_...`.

Tested:
- `env GOCACHE=/private/tmp/drive9-pr572-review/.codex-gocache go test ./pkg/tenant/token ./cmd/drive9/cli`

srstack:
## Response to review feedback

### 1. Migration path / deprecation signal
Agreed this is worth an explicit decision. Added to our backlog to track — we will add a debug-level log when legacy `dat9_` tokens are accepted, plus document a deprecation timeline in a follow-up PR.

### 2. Duplicate legacy constants
Fixed — `LegacyTokenPrefix` is now exported from `pkg/tenant/token` and imported in `ctx.go`, making it a single source of truth.

### 3. Missing test for `decodeDrive9APIKeyPayload` legacy path
Added `TestDecodeDrive9APIKeyPayloadAcceptsLegacyPrefix` in `ctx_test.go` — covers both new `drive9_` and legacy `dat9_` prefixes.

### 4. Const block trailing blank line
Fixed — simplified to single-line const since the only reference to the legacy prefix now comes from the imported `token.LegacyTokenPrefix`.

### 5. IssueToken drive9_ prefix test
Added `TestIssueTokenProducesDrive9Prefix` in `token_test.go`.

### Other fixes
- Updated `makeDrive9APIKey` helper to use `drive9_` prefix  
- Fixed help text at L1169 to be prefix-agnostic

PR #575, opened 2026-06-17
feat: native e2e smoke test + reject credentials on non-native providers

srstack:
## Summary

- **native-smoke-test.sh**: TiDB Cloud Native tenant lifecycle e2e smoke test  
  (manual-only, requires `DRIVE9_TIDBCLOUD_PUBLIC_KEY` / `DRIVE9_TIDBCLOUD_PRIVATE_KEY`)
- **pkg/server/server.go**: Reject `public_key`/`private_key` in provision body for non-native providers (zero, starter) with 400 Bad Request
- **e2e/AGENTS.md** + **e2e/README.md**: Document native dev endpoint, run command, coverage, and env vars

## Test coverage

1. Provision tenant via `drive9 create` with TiDB Cloud credentials
2. Poll `GET /v1/status` until active
3. Basic CLI fs ops (mkdir, cp, cat, ls, rm)
4. Delete tenant via `drive9 delete` and verify removal
5. Trap-based cleanup on failure

## Summary by CodeRabbit

* **New Features**
  * Added support for TiDB Cloud Native tenant lifecycle testing with automated provisioning, status polling, filesystem operations, and cleanup.
  * Implemented credential validation to reject native provider credentials for non-native providers.

* **Documentation**
  * Updated test documentation with new TiDB Cloud Native endpoint and required credential environment variables.

* **Tests**
  * Added test to verify native credentials are properly rejected for non-native providers.

### DIFF
--- a/e2e/AGENTS.md
+++ b/e2e/AGENTS.md
@@ -31,6 +31,9 @@ Current shared dev deployment:
 # Dev
 export DRIVE9_BASE="http://k8s-dat9-dat9serv-d5e02e7d07-1645488597.ap-southeast-1.elb.amazonaws.com"
 
+# Dev (tidbcloud-native)
+export DRIVE9_BASE="http://k8s-drive9ti-drive9se-b6bbe5ba6e-cee81207452d1185.elb.ap-southeast-1.amazonaws.com"
+
 # Prod
 export DRIVE9_BASE="https://api.drive9.ai"
 ```
@@ -91,6 +94,14 @@ RUN_GIT_OPS_SMOKE=1 RUN_GIT_WORKSPACE_SMOKE=1 bash e2e/smoke-all.sh
 
 # Include portable profile pack/unpack coverage in smoke-all when desired.
 RUN_PORTABLE_PACK_E2E=1 bash e2e/smoke-all.sh
+
+# TiDB Cloud Native (tidbcloud-native) tenant lifecycle smoke
+# Requires credentials, not wired into CI. Set DRIVE9_BASE from Deployment
+# endpoints above, or export manually. Credentials are stored in repo secrets
+# (DRIVE9_TIDBCLOUD_PUBLIC_KEY, DRIVE9_TIDBCLOUD_PRIVATE_KEY).
+DRIVE9_TIDBCLOUD_PUBLIC_KEY="$DRIVE9_TIDBCLOUD_PUBLIC_KEY" \
+DRIVE9_TIDBCLOUD_PRIVATE_KEY="$DRIVE9_TIDBCLOUD_PRIVATE_KEY" \
+bash e2e/native-smoke-test.sh
 ```
 
 ### Local via `drive9-server-local`
@@ -494,6 +505,16 @@ enabled.
 5. Runs `portable-pack-unpack-e2e.sh` when `RUN_PORTABLE_PACK_E2E=1`
 6. Aggregates pass/fail at script level for quick regression checks
 
+### `native-smoke-test.sh`
+
+Manual-only: requires TiDB Cloud API credentials. Not wired into CI.
+
+1. Provision tenant via `drive9 create` with `--tidbcloud-public-key` / `--tidbcloud-private-key`
+2. Poll `GET /v1/status` until active
+3. Basic CLI fs operations (`mkdir`, `cp`, `cat`, `ls`, `rm`)
+4. Delete tenant via `drive9 delete` and verify removal (401/403/404 on `GET /v1/status`)
+5. Trap-based cleanup: attempts `drive9 delete` on script failure unless `SKIP_CLEANUP=1`
+
 ## Environment variables
 
 | Variable | Default | Used by |
@@ -596,6 +617,9 @@ enabled.
 | `PJDFSTEST_MOUNT_ALLOW_OTHER` | `auto` | on-demand `pjdfstest-suite.sh` / `posix-feature-matrix.sh`; Linux auto-adds `--allow-other`, Darwin does not |
 | `GIT_MATRIX_TIMEOUT_S` | `240` | on-demand `git-feature-matrix.sh` |
 | `GIT_MATRIX_RUN_OVERSIZED` | `1` | on-demand `git-feature-matrix.sh` |
+| `DRIVE9_TIDBCLOUD_PUBLIC_KEY` | *(required)* | `native-smoke-test.sh` |
+| `DRIVE9_TIDBCLOUD_PRIVATE_KEY` | *(required)* | `native-smoke-test.sh` |
+| `SKIP_CLEANUP` | `0` | `native-smoke-test.sh` |
 
 ## Conventions
 
--- a/e2e/README.md
+++ b/e2e/README.md
@@ -31,6 +31,7 @@ including local single-tenant validation via `drive9-server-local`.
 | `fuse-write-perf-budget-test.sh` | FUSE write-path perf budgets: fsync-heavy workload with deterministic op-count budgets (remote writes/stats/lists/mutations, commit retries/failures) plus an fsync latency ceiling, asserted from mount perf counters |
 | `git-workspace-smoke-test.sh` | Git workspace fast-blobless clone with coding-agent local overlay, batched tracked-file edits, ignored local-only paths, `git add`/`commit`, `git apply`, and remount restore |
 | `posix-permission-smoke-test.sh` | POSIX permission coverage: API mkdir/chmod mode propagation, CLI `fs chmod`, FUSE `chmod`/`mkdir -m` with remote and local stat parity |
+| `native-smoke-test.sh` | TiDB Cloud Native tenant lifecycle: CLI provision with credentials, status poll, basic fs ops (mkdir/cp/cat/ls/rm), delete + verification, trap cleanup on failure |
 | `pjdfstest-suite.sh` | On-demand Linux/macOS pjdfstest POSIX compatibility suite over a real Drive9 FUSE mount; `pjdfstests.sh` and `posix-feature-matrix.sh` are aliases |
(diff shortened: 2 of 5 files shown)


mornyx (approved):

srstack (commented):
## Review

### Overview

Two changes: a new manual e2e smoke test for the TiDB Cloud Native tenant lifecycle, and a server-side guard rejecting `public_key`/`private_key` in provision requests for non-native providers. Small, focused PR.

---

### `pkg/server/server.go` — `rejectCredentialProvisionBody`

**Body-replay pattern is correct but fragile at the boundary.**

`rejectCredentialProvisionBody` reads the body, then rewrites `r.Body` with a fresh `NopCloser` so the downstream `provisionTenant` call can still read it. This works for the current call path, but the sentinel value `maxCredentialProvisionBodyBytes+1` used as the limit means a body of exactly `maxCredentialProvisionBodyBytes+1` bytes is read but silently passes validation (the unmarshal just fails or the keys are empty). The intent seems to be "reject oversized bodies", but the function actually swallows oversized bodies silently by returning `nil` when the keys happen to be absent. That's probably fine for this validation purpose (credentials won't be in a garbage multi-MB body), but it's slightly inconsistent with the limit semantics used for the native branch. Consider capping to `maxCredentialProvisionBodyBytes` and returning early if `len(body) > maxCredentialProvisionBodyBytes` — or document that oversized bodies intentionally pass through.

**Error granularity: `public_key` and `private_key` are checked separately but the second check is dead code when both are set.**

If `public_key` is non-empty, the function returns immediately and never checks `private_key`. A client sending both keys gets an error only about `public_key`. A client sending only `private_key` correctly gets an error about `private_key`. This is fine behavior (one error at a time), but worth noting in case the intent was to surface both in one shot. If so, collect both into a single error.

**The `else` branch — order of validation.**

The credential rejection happens in the `else` of `provider == tenant.ProviderTiDBCloudNative`. That means a request with an unknown or invalid provider gets through `NormalizeProvider`, returns a 400, and `rejectCredentialProvisionBody` is never called. That's correct — no need to reject credentials on an already-invalid provider request. The ordering is fine.

**Metric label on rejection.**

```go
metricEvent(r.Context(), "tenant_provision", "provider", provider, "result", "error")
```

This emits `result=error`, same as every other failure. If you ever want to distinguish "caller sent credentials to wrong provider" in dashboards, a more specific label like `result=credential_rejected` would help. Minor, but it matches the pattern used for `provision_credential_rejected` in the log event.

---

### `e2e/native-smoke-test.sh`

**`check_cmd` with `echo` is a no-op check.**

Several assertions use the pattern:
```bash
check_cmd "ls $DIR lists file" echo "$ls_out" | grep -q "hello.txt"
```

`check_cmd` runs `echo "$ls_out"` as the command (which always succeeds), and the `| grep -q` is parsed as a separate pipeline outside the `"$@"` expansion. So this check never fails regardless of whether `hello.txt` is in the output. The same issue affects `cat`, `rm` verification, and `DELETE_STATUS` check below.

The fix is either:
```bash
check_cmd "ls $DIR lists file" bash -c "echo '$ls_out' | grep -q 'hello.txt'"
```
or just use `check_eq` directly:
```bash
check_eq "ls $DIR lists file" "$(echo "$ls_out" | grep -c hello.txt)" "1"
```

This also affects:
```bash
check_cmd "delete status is deleted or deleting" echo "$DELETE_STATUS" | grep -qE 'deleted|deleting'
```
That check always passes.

**`drive9_retry` for mkdir/cp/etc. doesn't capture the check result.**

```bash
drive9_retry fs mkdir "$DIR" >/dev/null
check_cmd "mkdir $DIR" true
```

`check_cmd ... true` always passes — it doesn't verify the `drive9_retry` call succeeded. The exit code of `drive9_retry` is discarded. If `mkdir` fails, `set -e` will abort the script before `check_cmd` runs, but the failure won't be counted in `FAIL` — the script will just exit non-zero without the summary. If you want the failure to be counted (so you get a proper summary), use `check_cmd "mkdir $DIR" drive9_retry fs mkdir "$DIR"` and drop `>/dev/null`.

**`CREATED=0` after explicit delete — cleanup guard.**

The trap calls `drive9_ctx delete` only when `CREATED=1`. After step [4] succeeds, `CREATED=0` is set, so double-delete is avoided. Good. But step [4] uses `drive9_ctx delete` with explicit `--server` and `--api-key` flags, while the trap uses the same. Consistent.

**`CLI_BIN` is a file created by `mktemp`, not a directory.** `make build-cli CLI_BIN="$CLI_BIN"` assumes the Makefile accepts a `CLI_BIN` variable for the output path. This is a dependency on the Makefile interface — worth a comment confirming this variable is supported.

**`CLI_RETRY_SLEEP_S` type.**

`sleep "$CLI_RETRY_SLEEP_S"` — the variable default is `"2"` (string), which is fine for `sleep`. No issue.

**Minor: `TS` used only in `TEST_DIR` but not in cleanup messages.** When cleanup prints `"[cleanup] deleting tenant $TENANT_ID"`, `TENANT_ID` may still be empty if provision failed before the assignment. It won't break anything (`drive9 delete` will fail and be swallowed by `|| true`), but logging `"deleting tenant "` is confusing. A guard `[ -n "$TENANT_ID" ]` would make the log more readable.

---

### Documentation (`AGENTS.md`, `README.md`)

Clean. The new endpoint URL is documented in both files, the env var table is updated, and the run command examples are consistent. The note that `native-smoke-test.sh` is manual-only and not wired into CI is clear in all three places it appears.

---

### Summary

The server-side credential rejection is correct and safe. The smoke test has two real bugs: the `check_cmd ... echo ... | grep` pattern silently passes all grep-based assertions, and several "check" calls verify `true` rather than the actual command outcome. These won't cause false positives at the script level (the assertions all claim pass), but they do mean the test would not catch regressions in `ls`, `cat`, `rm`, or `delete` output. The guard logic and cleanup trap are otherwise well-structured.

srstack (commented):
## Re-review

All issues from the previous review are addressed. Going through them:

**1. `check_cmd ... echo "$var" | grep` always passing — resolved**

All grep-based assertions now use the `bash -c "echo \"$1\" | grep -q ..." _ "$var"` form, which correctly passes the variable as a positional argument and runs the pipe inside a subshell. The `ls`, `cat`, `mv`, `delete status`, and post-delete verification checks are all fixed.

**2. `drive9_retry` exit code not counted in `FAIL` — resolved**

`mkdir`, `cp`, `mv`, `rm` calls no longer use `check_cmd ... true`. The strategy changed: `drive9_retry` exits under `set -e`, so a failure aborts immediately rather than being counted. This is a valid trade-off for a smoke test — fail fast rather than accumulate. Fine.

**3. Body-replay limit sentinel `+1` — resolved**

`io.LimitReader(r.Body, maxCredentialProvisionBodyBytes)` (no `+1`) plus an explicit `>= maxCredentialProvisionBodyBytes` guard now returns a `"request body too large"` error instead of silently passing.

**4. Invalid JSON now returns 400 — resolved**

Previously `json.Unmarshal` failure returned `nil` (silent pass). Now it returns `fmt.Errorf("invalid JSON body: %w", err)`, which propagates to a 400 Bad Request.

**5. Log level on rejection changed to `Error` — noted**

```go
logger.Error(r.Context(), "server_event", eventFields(r.Context(), "provision_credential_rejected", ...))
```

This was `Warn` in the first revision and is now `Error`. A caller sending `public_key` to a zero/starter provider is most likely a misconfiguration or SDK bug — `Error` is defensible, but `Warn` would also be reasonable since it's a 400-class client error rather than a server fault. Not blocking, just flagging in case the logging severity convention in this codebase reserves `Error` for server-side failures.

**6. Test coverage added — `TestProvisionRejectsCredentialsForNonNativeProvider`**

Covers all three non-native providers (`ProviderTiDBZero`, `ProviderTiDBCloudStarter`, `ProviderDB9`) in a loop with a real HTTP test server. The test is complete and well-structured.

---

### New in this revision: expanded smoke test

**Large file upload/download (step [5])**

`dd if=/dev/zero` with `sha256sum` round-trip is a good integrity check. One minor note: `python3 -c "..."` is used to parse `fs stat` output. This creates a runtime dependency on Python 3. The other helper dependencies (`jq`, `curl`, `go`) are declared upfront with `check_cmd` guards, but Python 3 is not. If `python3` is unavailable, `remote_size` silently defaults to `"0"` and the `check_eq` will fail with a confusing message. Worth adding a `check_cmd "python3 is available" bash -c 'command -v python3 >/dev/null'` alongside the other tool guards, or parsing `stat` output with `awk`/`grep` to avoid the dependency.

**`rm -r $BATCH_DIR` followed by `check_cmd "rm -r batch dir" true`**

Same pattern as before — `check_cmd ... true` always passes. The `-r` rm is at step [4], before the batch dir is recreated for the large file test. The `check_cmd` here doesn't verify the `drive9_retry` outcome. Since the script is under `set -e` at that point, a failure still aborts the run — so this is not a correctness problem, just a cosmetic one (the check always shows `PASS`).

**Step numbering: comment says `[5]` but echo says `[7]`**

```bash
# ── [5] verify tenant gone ──────────────────────────────────────────────────
echo "[7] verify tenant removed"
```

The comment block header still says `[5]` while the echo and the new numbering say `[7]`. Cosmetic only.

**`drive9_retry` now retries on `"not found"`**

The retry loop now includes `*"not found"*` as a retryable condition alongside 429. This is intentional for the provisioning race (filesystem operations before the tenant is fully active), but it also means a hard `not found` error on an unrelated operation (wrong remote path, deleted file) will be retried 8 times before failing. That could mask bugs during the smoke run. Consider limiting `not found` retries to specific steps (e.g., only during step [3]) rather than globally, or document the intent.

---

### Summary

All three issues from the first review are resolved. The server-side fix is correct and well-tested. The smoke test is substantially more capable. Two minor items remain: the `python3` dependency is undeclared, and the global `not found` retry condition could mask path errors. Neither is blocking.

LGTM.

PR #577, opened 2026-06-17
feat: CLI improvements for TiDB Cloud user experience

srstack:
## Summary

Improvements to CLI usability and TiDB Cloud error handling.

### Region list
- `server_url` → `endpoint` in manifest and display
- Add `CLOUD PROVIDER` and `REGION` columns
- Anonymous mode risk warning

### Create / Delete commands
- Detailed help with flag descriptions and examples
- `--region-code` or `--server` required for TiDBCloud mode
- Anonymous mode risk warning on create

### Error messages
- Free cluster limit → spending limit guidance + pricing link
- Payment/credits exhausted → billing page link
- Instance capacity limit (100) → contact support
- Quota exceeded → context-aware guidance (Anonymous → switch to TiDBCloud, TiDBCloud → set spending limit)

### Mode labels
- Use user-facing `Anonymous` / `TiDBCloud` consistently throughout CLI

## Release Notes

* **Bug Fixes**
  * Improved error messages for quota exceeded scenarios, provisioning failures, and region/mode validation
  * Enhanced CLI help documentation for create/delete commands with examples
  * Updated region list display to include cloud provider and endpoint information
  * Provided more specific error guidance for failed cluster provisioning attempts

### DIFF
--- a/.github/workflows/prod-cd.yml
+++ b/.github/workflows/prod-cd.yml
@@ -89,6 +89,22 @@ jobs:
           kubectl --context prod -n "${{ env.DEPLOY_NAMESPACE }}" \
             rollout undo "deployment/${{ env.K8S_DEPLOYMENT }}"
 
+      - name: Deploy drive9-tidbcloud to aws-ap-southeast-1
+        id: deploy-sg-native
+        if: inputs.region == 'all' || inputs.region == 'aws-ap-southeast-1'
+        run: |
+          kubectl --context prod -n drive9-tidbcloud \
+            set image deployment/drive9-server \
+            drive9-server=$IMAGE
+          kubectl --context prod -n drive9-tidbcloud \
+            rollout status deployment/drive9-server --timeout=180s
+
+      - name: Rollback drive9-tidbcloud on failure
+        if: failure() && steps.deploy-sg-native.outcome == 'failure'
+        run: |
+          kubectl --context prod -n drive9-tidbcloud \
+            rollout undo deployment/drive9-server
+
       - name: Deploy to aws-us-east-1
         id: deploy-us
         if: inputs.region == 'all' || inputs.region == 'aws-us-east-1'
--- a/cmd/drive9/cli/ctx.go
+++ b/cmd/drive9/cli/ctx.go
@@ -383,7 +383,7 @@ func ctxAddCmd(args []string) error {
 		return fmt.Errorf("--api-key is required")
 	}
 	if mode != "" && !isKnownContextMode(mode) {
-		return fmt.Errorf("unknown --mode %q; valid: Anonymous, TiDBCloud", mode)
+		return fmt.Errorf("unknown --mode %q; valid: %s, %s", mode, ModeLabelAnonymous, ModeLabelTiDBCloud)
 	}
 	if (cloudProvider != "" || region != "") && mode == "" {
 		return fmt.Errorf("--cloud-provider and --region require --mode TiDBCloud")
--- a/cmd/drive9/cli/db.go
+++ b/cmd/drive9/cli/db.go
@@ -119,11 +119,18 @@ func Create(args []string) error {
 
 	mode := provisionModeForCredentials(publicKey, privateKey)
 	if mode == RegionModeTiDBCloudNative && (publicKey == "" || privateKey == "") {
-		return fmt.Errorf("tidb_cloud_native create requires --tidbcloud-public-key and --tidbcloud-private-key, or %s/%s", EnvTiDBCloudPublicKey, EnvTiDBCloudPrivateKey)
+		return fmt.Errorf("TiDBCloud mode requires --tidbcloud-public-key and --tidbcloud-private-key, or %s/%s", EnvTiDBCloudPublicKey, EnvTiDBCloudPrivateKey)
+	}
+	if mode == RegionModeTiDBCloudNative && strings.TrimSpace(serverFlag) == "" && strings.TrimSpace(regionCode) == "" {
+		return fmt.Errorf("TiDBCloud mode requires --region-code or --server; use drive9 region list to see available regions")
 	}
 	server, regionEntry, err := resolveProvisionServer(serverFlag, regionCode, mode, r.Server)
 	if err != nil {
 		return err
+	}
+
+	if mode == RegionModeTiDBCloudStarter {
+		fmt.Fprintf(os.Stderr, "Note: Anonymous mode in drive9 transfers data management rights to PingCAP.\n")
 	}
 
 	cfg := loadConfig()
@@ -208,6 +215,9 @@ func Create(args []string) error {
 const (
 	RegionModeTiDBCloudNative  = "tidb_cloud_native"
 	RegionModeTiDBCloudStarter = "tidb_cloud_starter"
+
+	ModeLabelAnonymous = "Anonymous"
+	ModeLabelTiDBCloud = "TiDBCloud"
 )
 
 type createResult struct {
@@ -232,7 +242,34 @@ type createOutput struct {
 }
 
 func createUsage() string {
-	return "usage: drive9 create [--name NAME] [--region-code CODE] [--server URL] [--tidbcloud-public-key KEY] [--tidbcloud-private-key KEY] [--json]"
+	return `usage: drive9 create [flags]
+
+provision a new tenant and register the returned API key as a local owner context.
+
+flags:
+  --name NAME                     context name (default: auto-generated 7-char name)
+  --region-code CODE              provisioning region code; use "drive9 region list" to see available regions
+  --server URL                    override the server URL (bypasses region manifest lookup)
+  --tidbcloud-public-key KEY      TiDB Cloud public key (required for TiDBCloud mode)
+  --tidbcloud-private-key KEY     TiDB Cloud private key (required for TiDBCloud mode)
+  --json                          output result as JSON
+
+examples:
+  # provision an Anonymous tenant using the default region
+  drive9 create
+
+  # provision a TiDBCloud tenant in ap-southeast-1
+  drive9 create --region-code aws-ap-southeast-1 \
+    --tidbcloud-public-key <public-key> \
+    --tidbcloud-private-key <private-key>
+
+  # provision directly against a known server
+  drive9 create --server http://127.0.0.1:9009
+
+  # list available regions
+  drive9 region list
+
+note: Anonymous mode in drive9 transfers data management rights to PingCAP.`
 }
 
 func provisionModeForCredentials(publicKey, privateKey string) string {
@@ -247,7 +284,7 @@ func provisionRequestBody(publicKey, privateKey string) (io.Reader, error) {
 		return nil, nil
 	}
 	if publicKey == "" || privateKey == "" {
-		return nil, fmt.Errorf("tidb_cloud_native create requires both public and private keys")
+		return nil, fmt.Errorf("TiDBCloud mode requires both public and private keys")
 	}
 	body := map[string]string{
 		"public_key":  publicKey,
@@ -275,7 +312,7 @@ func resolveProvisionServer(serverFlag, regionCode, mode, fallbackServer string)
 	if err != nil {
 		return "", nil, err
 	}
-	return strings.TrimSpace(entry.ServerURL), entry, nil
+	return strings.TrimSpace(entry.Endpoint), entry, nil
 }
 
 func selectRegionServer(entries []RegionManifestEntry, regionCode, mode string) (*RegionManifestEntry, error) {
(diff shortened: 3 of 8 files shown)


srstack (commented):
## Review

### Overview

Four related changes: `server_url` → `endpoint` rename in the region manifest, richer `region list` columns, expanded `create`/`delete` help text, mode label constants, anonymous mode warnings, and context-aware quota exceeded guidance. Small and focused except for one correctness bug.

---

### `cmd/drive9/main.go` — quota message detection is broken

```go
if strings.Contains(strings.ToLower(msg), "tenant usage quota exceeded") {
```

The actual error string in `pkg/backend/quota.go` is:

```go
ErrStorageQuotaExceeded = errors.New("tenant storage quota exceeded")
```

`"tenant usage quota exceeded"` ≠ `"tenant storage quota exceeded"`. The substring never matches, so `QuotaExceededMessageForCurrentContext()` is never called and users always see the raw error. The detection should be `"tenant storage quota exceeded"` (or just `"storage quota exceeded"` for robustness).

The `ErrMediaLLMQuotaExceeded` message (`"tenant media LLM file quota exceeded"`) also won't match, but that's probably intentional — the media quota guidance is different from the storage quota guidance.

---

### `pkg/server/server.go` — string-matching on TiDB Cloud error messages

```go
if strings.Contains(strings.ToLower(err.Error()), "free cluster") { ... }
else if strings.Contains(strings.ToLower(err.Error()), "credits") || ... "payment" { ... }
else if strings.Contains(strings.ToLower(err.Error()), "instance capacity limit") || ... "capacity limit" { ... }
```

This works as long as the upstream TiDB Cloud API error strings stay stable. Three observations:

1. `"capacity limit"` is a broad substring — it would also match "read capacity limit", "write capacity limit", or any other upstream message containing those two words. Preferring the more specific `"instance capacity limit"` as the sole check would be safer.

2. `"credits"` is equally broad — it appears in many common contexts ("API credits", "cloud credits"). Worth anchoring more specifically if the TiDB Cloud error message is known (e.g., `"credits exhausted"` or `"insufficient credits"`).

3. These matches are case-folded with `ToLower` but the error is wrapped via `fmt.Sprintf("provision tenant cluster failed: %v", err)` before the checks run. That's fine — the checks are on the original `err.Error()`, not `msg`.

Not blocking, but consider logging the raw upstream error at `Debug` level so it's available in support bundles without being surfaced to users.

---

### `cmd/drive9/cli/region.go` — `quotaExceededMessage` switch uses mode label strings

```go
func quotaExceededMessage(mode string) string {
    switch strings.TrimSpace(mode) {
    case "", ModeLabelAnonymous:   // "Anonymous"
    case ModeLabelTiDBCloud:       // "TiDBCloud"
```

The `mode` field on a context entry is stored as the internal mode constant (`RegionModeTiDBCloudStarter` / `RegionModeTiDBCloudNative`), not the display label. `ctx.Mode` from the config is the raw value from the manifest — `"tidb_cloud_starter"` / `"tidb_cloud_native"`. So `case ModeLabelAnonymous` (`"Anonymous"`) never matches a context mode, and `case ModeLabelTiDBCloud` (`"TiDBCloud"`) never matches either. Both cases always fall through to `default: return ""`, which means `QuotaExceededMessageForCurrentContext()` always returns `""`, and `main.go`'s `if m := cli.QuotaExceededMessageForCurrentContext(); m != ""` guard means the raw error is always shown.

The switch should use the internal constants:

```go
case "", RegionModeTiDBCloudStarter:
    return "tenant usage quota exceeded. Switch to TiDBCloud mode ..."
case RegionModeTiDBCloudNative:
    return "tenant usage quota exceeded. Go to your TiDB Cloud cluster settings ..."
```

(Combined with the `main.go` detection fix above, both bugs need to be fixed for the feature to work at all.)

---

### `cmd/drive9/cli/db.go` — anonymous mode warning fires on `RegionModeTiDBCloudStarter`

```go
if mode == RegionModeTiDBCloudStarter {
    fmt.Fprintf(os.Stderr, "Note: Anonymous mode in drive9 transfers data management rights to PingCAP.\n")
}
```

This fires before `resolveProvisionServer`, so if `--server` is provided directly with no credentials the warning still prints. The user experience might be unexpected: you pass `--server http://localhost:9009` and get an anonymous mode warning. Worth gating on whether the mode was inferred from credentials vs. defaulted from a flag.

---

### `cmd/drive9/cli/region.go` — `regionListCmd` warning loop

```go
for _, entry := range manifest.Regions {
    if regionModeLabel(entry.Mode) == ModeLabelAnonymous {
        fmt.Fprintln(os.Stderr, "Note: Anonymous mode ...")
        break
    }
}
```

Calling `regionModeLabel` here and comparing to `ModeLabelAnonymous` is correct (since `regionModeLabel` returns `ModeLabelAnonymous` for `RegionModeTiDBCloudStarter`). Consistent.

---

### `server_url` → `endpoint` rename

Clean. All struct fields, JSON tags, validation messages, test fixtures, and the fallback manifest are updated. No stragglers found.

---

### `RegionManifestEntry` vs `regionListOutputEntry` duplication

These two structs remain identical except one is the manifest wire format and one is the JSON output type. This isn't new in this PR, but since `Endpoint` and `CloudProvider`/`TiDBRegion` are now both present in both, it's worth noting — `regionListOutput` copies fields one by one and could silently miss a new field. Minor, not blocking.

---

### Summary

Two functional bugs that together mean the quota exceeded guidance feature never activates:

1. **`main.go`**: detects `"tenant usage quota exceeded"` but the error is `"tenant storage quota exceeded"`.
2. **`region.go`**: `quotaExceededMessage` switches on display labels (`"Anonymous"`, `"TiDBCloud"`) but context mode is stored as internal constants (`"tidb_cloud_starter"`, `"tidb_cloud_native"`).

Everything else (rename, help text, mode label constants, warning messages) is correct.

srstack (commented):
## Re-review

### What changed

`main.go` detection string fixed (now checks both `"tenant storage quota exceeded"` and `"tenant usage quota exceeded"`). The `endpoint` rename from the first revision was reverted — fields and JSON tags are back to `ServerURL` / `"server_url"`. No other substantive changes to `region.go` or `db.go`.

---

### Bug 1 — `region.go` has embedded tab characters and will not compile

The branch version of `region.go` contains literal tab characters embedded inside field accesses:

```go
entry.	ServerURL   // tab between '.' and 'ServerURL'
```

This appears in at least five places: the `fmt.Fprintf` in `regionListCmd`, the `regionListOutput` assignment, `validateRegionManifest`'s trim and empty check, and `sortRegionManifestEntries`. Go does not allow whitespace between a selector dot and the field name — this is a syntax error and the package will not compile.

This looks like an editor artifact from the revert. The file needs to be cleaned up so all accesses read `entry.ServerURL` with no intervening whitespace.

---

### Bug 2 — `quotaExceededMessage` still matches on display labels, not internal constants (unchanged from first review)

```go
func quotaExceededMessage(mode string) string {
    switch strings.TrimSpace(mode) {
    case "", ModeLabelAnonymous:   // "Anonymous"
    case ModeLabelTiDBCloud:       // "TiDBCloud"
```

`ctx.Mode` is stored as `"tidb_cloud_starter"` / `"tidb_cloud_native"` — neither matches `"Anonymous"` or `"TiDBCloud"`. Both cases fall through to `default: return ""`. `QuotaExceededMessageForCurrentContext()` always returns `""` and `main.go`'s `if m := ...; m != ""` guard means guidance is never shown.

Fix:

```go
case "", RegionModeTiDBCloudStarter:
    return "tenant storage quota exceeded. Switch to TiDBCloud mode ..."
case RegionModeTiDBCloudNative:
    return "tenant storage quota exceeded. Go to your TiDB Cloud cluster settings ..."
```

---

### `main.go` detection — partially fixed

The detection now covers `"tenant storage quota exceeded"` (the actual `ErrStorageQuotaExceeded` message) as well as the old `"tenant usage quota exceeded"` string. The `main.go` side is now correct. But since bug 2 above is still present, the end-to-end path still does nothing.

---

### `region list` table header vs field name mismatch

The table header was updated to `"REGION CODE\tCLOUD PROVIDER\tREGION\tMODE\tSERVER"` but the `endpoint` rename was reverted, so `entry.ServerURL` is used to fill the `SERVER` column. The header label is fine, but the JSON output field is still `"server_url"` — the first revision's `"endpoint"` rename is gone. If the JSON field rename was intentional, it needs to be re-applied cleanly. If it was abandoned in favor of keeping `"server_url"`, the `db_test.go` and `region_test.go` are already consistent with `ServerURL`, which is fine.

---

### Summary

Two blockers before merge:

1. **Syntax error**: tab characters inside field selectors in `region.go` — the package will not compile.
2. **`quotaExceededMessage` mode matching**: still uses display labels instead of internal mode constants — quota guidance is never shown.

The `main.go` detection fix is correct. Everything else (help text, mode label constants, anonymous warning, provision validation, CI deploy step) looks good.

srstack (commented):
Thanks for the review. Fixed:

1. **quota detection** — now matches "usage quota" which is the exact string from sanitizeClientError for TiDB spending limit
2. **mode labels** — ctx.Mode stores regionModeLabel output (display labels, not internal constants), so ModeLabelAnonymous/ModeLabelTiDBCloud are correct in quotaExceededMessage
3. **endpoint rename** — reverted back to server_url per upstream change

srstack:
> **quota detection**: now matches `"usage quota"` which is the exact string from sanitizeClientError for TiDB spending limit\n> **mode labels**: ctx.Mode stores regionModeLabel output (display labels), so ModeLabelAnonymous/ModeLabelTiDBCloud are correct in quotaExceededMessage\n> **endpoint rename**: reverted back to server_url per upstream change

srstack (commented):
## Re-review

### What changed

The tab-corruption in `region.go` is gone — field accesses are clean. `main.go` detection string changed to `"usage quota"`, which is a broadened substring that now covers both `"tenant storage quota exceeded"` (the actual `ErrStorageQuotaExceeded` message) and any future `"tenant usage quota exceeded"` variant. Both blockers from the previous review have been addressed except one.

---

### Remaining issue — `quotaExceededMessage` still matches on display labels

This was raised in both previous reviews and is still unchanged:

```go
func quotaExceededMessage(mode string) string {
    switch strings.TrimSpace(mode) {
    case "", ModeLabelAnonymous:   // "Anonymous"
    case ModeLabelTiDBCloud:       // "TiDBCloud"
```

`ctx.Mode` in the config is stored as the internal constant (`"tidb_cloud_starter"` / `"tidb_cloud_native"`), not the display label. Neither `"Anonymous"` nor `"TiDBCloud"` will ever match, so both cases fall through to `default: return ""`. `QuotaExceededMessageForCurrentContext()` always returns `""`, and the `if m := ...; m != ""` guard in `main.go` means the guidance message is never surfaced.

Fix — use the internal constants as case values:

```go
case "", RegionModeTiDBCloudStarter:   // "tidb_cloud_starter"
    return "tenant storage quota exceeded. Switch to TiDBCloud mode ..."
case RegionModeTiDBCloudNative:        // "tidb_cloud_native"
    return "tenant storage quota exceeded. Go to your TiDB Cloud cluster settings ..."
```

Everything else looks good. LGTM once this is fixed.

mornyx (commented):
## Review

I re-reviewed the current head (`1aba8e2`). The earlier quota-mode concern looks addressed now: newly created contexts store the display labels (`Anonymous` / `TiDBCloud`), and `quotaExceededMessage` also accepts the internal mode constants. The `fatal` detection also lines up with `sanitizeClientError`, which emits `tenant usage quota exceeded` for TiDB usage-quota failures.

I found two small issues to fix before merge:

1. `cmd/drive9/cli/region.go` is not gofmt-formatted. `gofmt -l cmd/drive9/cli/region.go` still reports the file; the remaining diff is the misaligned `ServerURL` fields from the revert.

2. `regionListCmd` now drops the `tabwriter.Flush()` error:

```go
_ = w.Flush()
...
return nil
```

Before this PR, the command returned `w.Flush()`, so stdout/write failures made the CLI fail. Please preserve that behavior while still printing the warning afterward, for example:

```go
if err := w.Flush(); err != nil {
    return err
}
for _, entry := range manifest.Regions {
    if regionModeLabel(entry.Mode) == ModeLabelAnonymous {
        fmt.Fprintln(os.Stderr, "Note: Anonymous mode in drive9 transfers data management rights to PingCAP.")
        break
    }
}
return nil
```

Validation note: `go test ./cmd/drive9` passed. `go test ./cmd/drive9/cli` could not complete in my sandbox because `httptest` was not allowed to bind a localhost port, so I could not use that run as a signal for this PR.

mornyx (approved):
Approved. Re-reviewed the latest head (`6dd0973`); the previous gofmt issue and `tabwriter.Flush()` error handling regression are fixed. I also verified `gofmt -l` is clean for the touched Go files and `go test ./cmd/drive9 ./cmd/drive9/cli` passes.

PR #578, opened 2026-06-17
feat(cli): add visual tree help

mornyx:
## Summary

- Adds a visual, tree-structured help view for `drive9 help` / `drive9 -?`, with color-aware terminal rendering and pager support for large output.
- Keeps `drive9 help --plain` and the existing `-h` / `--help` path available for script-friendly classic usage.
- Expands the visual help tree with detailed command groups, flags, examples, mount tuning options, and help controls.
- Merged the latest `origin/main` CLI changes and synced visual help for the updated TiDB Cloud create/delete flow and `region list --manifest-url` support.

## Updated help output

```text
drive9 <command> [args] [flags]
│
├─ drive9 create [flags] → provision a new tenant and save an owner context
│  ├─ no TiDB Cloud keys -> Anonymous mode; TiDB Cloud keys -> TiDBCloud mode
│  ├─ Anonymous mode transfers data management rights to PingCAP
│  ├─ --name NAME                                context name; defaults to an auto-generated 7-character name
│  ├─ --region-code CODE                         provisioning region code; use drive9 region list to see available
│  │                                             regions
│  ├─ --server URL                               override server URL and bypass region manifest lookup
│  ├─ --tidbcloud-public-key KEY                 TiDB Cloud public key; required for TiDBCloud mode
│  ├─ --tidbcloud-private-key KEY                TiDB Cloud private key; required for TiDBCloud mode
│  ├─ --json                                     output result as JSON
│  └─ Examples:
│     ├─ drive9 create                                        # provision an Anonymous tenant using the default region
│     ├─ drive9 create --region-code aws-ap-southeast-1 --tidbcloud-public-key <public-key> --tidbcloud-private-key <private-key>
│     │  # provision a TiDBCloud tenant
│     ├─ drive9 create --server http://127.0.0.1:9009         # provision directly against a known server
│     └─ drive9 region list                                   # list available regions
│
├─ drive9 delete [flags] → delete the current tenant, cluster, database, and API keys [DANGER]
│  ├─ TiDBCloud mode requires TiDB Cloud credentials for tenant deletion
│  ├─ --server URL                               server URL; defaults to active context server
│  ├─ --api-key KEY                              owner API key; defaults to active context API key
│  ├─ --tidbcloud-public-key KEY                 TiDB Cloud public key; required for TiDBCloud mode
│  ├─ --tidbcloud-private-key KEY                TiDB Cloud private key; required for TiDBCloud mode
│  ├─ --json                                     output result as JSON
│  └─ Examples:
│     ├─ drive9 delete                                        # delete the active context's tenant
│     └─ drive9 delete --server https://api.drive9.ai --api-key drive9_xxx --tidbcloud-public-key <public-key> --tidbcloud-private-key <private-key>
│        # delete a TiDBCloud tenant with explicit credentials
│
├─ drive9 ctx <show|add|import|fork|ls|use|rm> → manage local drive9 contexts
│  ├─ show -> inspect active context; ls -> scan saved contexts
│  ├─ add/import -> save owner or delegated keys
│  ├─ fork -> create a copy-on-write context; rm -> remove local credentials
│  ├─ --json                                     available on show, fork, and ls
│  ├─ --reveal                                   show sensitive fields on ctx show
│  ├─ --type <kind>|--scoped                     filter ctx ls results
│  └─ Examples:
│     ├─ drive9 ctx show                                      # show active context
│     ├─ drive9 ctx add --api-key drive9_xxx --name dev       # save a context
│     └─ drive9 ctx use dev                                   # activate context
│
├─ drive9 fs <command> [arguments] → filesystem operations over drive9 paths
│  ├─ paths are absolute remote paths; directories end with /
│  ├─ Read/write:
│  │  ├─ cp [-r] [--resume] <src> <dst>             copy local, remote, stdin, and stdout paths
│  │  ├─ cp --append <src> <dst>                    append local or stdin input to a remote file
│  │  ├─ cp --tag key=value                         set upload tags; repeatable; re-upload with tags replaces existing tag
│  │  │                                             set
│  │  ├─ cp --description TEXT                      set file description on local/stdin upload
│  │  ├─ cat <path>                                 read a remote file to stdout
│  │  ├─ ls [-l] [path]                             list a remote directory
│  │  └─ stat [-o text|json] <path>                 show file metadata
│  ├─ Namespace:
│  │  ├─ mv <old> <new>                             rename or move a path
│  │  ├─ mkdir <path>                               create a directory; parents are auto-created
│  │  ├─ chmod <mode> <path>                        change permissions with octal mode, e.g. 644
│  │  ├─ symlink <target> <link>                    create a symbolic link
│  │  ├─ hardlink <target> <link>                   create a hard link
│  │  └─ rm [-r|--recursive] <path>                 remove a file or directory tree
│  ├─ Search:
│  │  ├─ grep [--layer REF] <pattern> [dir]         search file contents
│  │  ├─ find [dir] -name GLOB                      match filename glob
│  │  ├─ find [dir] -tag key=value                  exact tag key/value match
│  │  ├─ find [dir] -tag key                        tag-key existence match
│  │  ├─ find [dir] -newer YYYY-MM-DD               modified after date
│  │  ├─ find [dir] -older YYYY-MM-DD               modified before date
│  │  └─ find [dir] -size +N|-N                     size filter in bytes
│  ├─ Layers:
│  │  ├─ layer create [flags] <base-root>           create writable fs layer
│  │  ├─ layer list [--json]                        list layers
│  │  ├─ layer status|diff [--json] <layer>         inspect layer state or diff
│  │  ├─ layer checkpoint [flags] <layer>           create a layer checkpoint
│  │  ├─ layer rollback <layer>                     rollback a layer
│  │  └─ layer commit <layer>                       commit a layer into its base
│  └─ Examples:
│     ├─ drive9 fs ls -l :/                                   # list root
│     ├─ drive9 fs cp ./notes.md :/team/notes.md              # upload a file
│     └─ drive9 fs find :/team -tag owner=agent               # find by exact tag
│
├─ drive9 token <issue|revoke> → issue and revoke workspace-zone scoped filesystem tokens [SCOPED]
│  ├─ issue --subject NAME                       token subject
│  ├─ issue --zone PATH                          workspace zone root
│  ├─ revoke <name|->                            revoke by name or stdin
│  └─ Examples:
│     ├─ drive9 token issue --subject vm0 --zone :/work       # create scoped token
│     └─ drive9 token revoke vm0                              # revoke token
│
├─ drive9 vault <set|get|put|with|ls|rm|grant|revoke|audit> → manage encrypted vault secrets
│  ├─ put imports local secret directories; with runs a command with mounted secrets
│  ├─ grant/revoke/audit manage secret access and audit trails
│  └─ Examples:
│     ├─ drive9 vault ls                                      # list secrets
│     ├─ drive9 vault put /n/vault/app --from ./secrets       # import a secret directory
│     └─ drive9 mount vault ./vault                           # read-only vault mount
│
├─ drive9 journal <new|append|cat|find|verify> → append-only agent/workflow journal operations
│  ├─ --type TYPE                                journal or entry type
│  ├─ --label key=value                          repeatable metadata label
│  ├─ --json                                     machine-readable output
│  └─ Examples:
│     ├─ drive9 journal new --type run                        # create journal
│     └─ drive9 journal append :/journal/run.md --label phase=plan
│        # append entry
│
├─ drive9 git <clone|worktree|hydrate> → git-aware workflows for mounted drive9 workspaces [FAST]
│  ├─ clone:
│  │  ├─ clone --fast <repo-url> <path>             clone and register a git workspace
│  │  ├─ --blobless                                 use a partial local .git; clean blobs lazy-fetch from remote
│  │  └─ --hydrate=auto|background|sync|off         clean-tree hydration strategy for blobless clones
│  ├─ worktree:
│  │  ├─ worktree add --fast <base> <path>          create and register a linked worktree
│  │  ├─ -b <branch>                                create a new branch for the worktree
│  │  ├─ --detach                                   detach HEAD in the new worktree
│  │  ├─ --blobless                                 require a blobless base workspace
│  │  ├─ --hydrate=auto|background|sync|off         clean-tree hydration strategy
│  │  └─ worktree remove --fast [--force]           remove a registered worktree
│  ├─ hydrate:
│  │  └─ hydrate [--timeout=30m] <path>             materialize a blobless clean tree into the local cache
│  └─ Examples:
│     ├─ drive9 git clone --fast https://repo.git ./repo      # fast clone into mount
│     ├─ drive9 git clone --fast --blobless --hydrate=sync https://repo.git ./repo
│     │  # blobless fast clone
│     └─ drive9 git hydrate ./repo                            # materialize clean blobs
│
├─ drive9 region list [--json] [--manifest-url URL] → list provisioning regions from the drive9 manifest
│  ├─ text output columns: REGION CODE, CLOUD PROVIDER, REGION, MODE, SERVER
│  ├─ Anonymous regions print a data-management rights note on stderr
│  ├─ --json                                     output region manifest entries as JSON
│  ├─ --manifest-url URL                         override the region manifest URL
│  └─ Examples:
│     ├─ drive9 region list                                   # human-readable regions
│     ├─ drive9 region list --json                            # scriptable regions
│     └─ drive9 region list --manifest-url http://127.0.0.1:9009/regions.json
│        # use a custom manifest during local testing
│
├─ drive9 pack [flags] [archive] [path...] → archive local overlay paths to drive9/S3
│  ├─ --local-root DIR                           profile local overlay root
│  ├─ --remote-root PATH                         remote root used to resolve relative pack paths
│  ├─ --profile NAME                             profile defaults when no paths are provided
│  ├─ --mount PATH                               mounted path whose metadata provides local-root, remote-root, and
│  │                                             profile
│  └─ Examples:
│     ├─ drive9 pack --profile coding-agent                   # pack configured overlay paths
│     ├─ drive9 pack :/packs/work.tar.gz .git dist            # explicit archive and paths
│     └─ drive9 pack --mount ./mnt                            # pack using mount metadata
│
├─ drive9 unpack [flags] [archive] → restore a drive9 pack archive to a local overlay
│  ├─ --local-root DIR                           profile local overlay root
│  ├─ --remote-root PATH                         remote root used to resolve default profile pack archive
│  ├─ --profile NAME                             profile defaults when no archive is provided
│  ├─ --mount PATH                               mounted path whose metadata provides local-root
│  ├─ --no-replace                               do not remove archived root paths before extracting
│  └─ Examples:
│     ├─ drive9 unpack --profile coding-agent                 # restore default pack
│     ├─ drive9 unpack :/packs/work.tar.gz                    # restore explicit archive
│     └─ drive9 unpack --mount ./mnt --no-replace             # restore into mount local-root without deleting existing roots
│
├─ drive9 profile show [profile] → print mount profile configuration
│  └─ Examples:
│     └─ drive9 profile show coding-agent                     # inspect a profile
│
├─ drive9 mount [flags] [:/remote] <mountpoint> → mount drive9 as a local filesystem
│  ├─ Connection:
│  │  ├─ --server URL                               drive9 server URL; overrides DRIVE9_SERVER and config
│  │  ├─ --api-key KEY                              owner API key; overrides DRIVE9_API_KEY and config
│  │  ├─ --mode auto|fuse|webdav                    mount mode; auto selects the best supported mode
│  │  └─ --foreground                               run in foreground and block until unmounted
│  ├─ Read cache:
│  │  ├─ --cache-dir DIR                            write-back cache directory; default ~/.cache/drive9
│  │  ├─ --cache-size MB                            memory read cache size
│  │  ├─ --read-cache-max-file-mb MB                maximum single file admitted to read cache
│  │  ├─ --read-cache-ttl DURATION                  read cache TTL; 0 disables time-based expiry
│  │  ├─ --disk-read-cache-size-mb MB               disk-backed read cache size
│  │  ├─ --disk-read-cache-free-ratio FLOAT         minimum filesystem free-space ratio before disk cache evicts
│  │  ├─ --dir-ttl DURATION                         directory cache TTL
│  │  ├─ --attr-ttl DURATION                        kernel attr cache TTL
│  │  └─ --entry-ttl DURATION                       kernel entry cache TTL
│  ├─ FUSE read path:
│  │  ├─ --lookup-retry-count N                     detached retries after transient Lookup/GetAttr stat failures; 0
│  │  │                                             disables
│  │  ├─ --lookup-retry-timeout DURATION            timeout per detached Lookup/GetAttr stat retry
│  │  ├─ --read-concurrency N                       maximum concurrent backend reads issued by FUSE
│  │  ├─ --parallel-read-concurrency N              maximum concurrent block reads for one large FUSE read
│  │  ├─ --parallel-read-block-size-mb MB           block size for parallel large-file reads
│  │  ├─ --fuse-sync-read                           disable kernel async read dispatch
│  │  ├─ --legacy-dir-stat-fallback                 list parent on Lookup stat 404 for legacy servers without directory stat
│  │  └─ --trust-process-local-events               allow revision-bound dir-cache hits using process-local SSE freshness
│  ├─ Prefetch:
│  │  ├─ --readdir-prefetch                         prefetch small files after directory reads into read cache
│  │  ├─ --readdir-prefetch-max-files N             maximum small files prefetched per directory read
│  │  ├─ --readdir-prefetch-max-file-bytes N        maximum individual file size for readdir prefetch
│  │  ├─ --readdir-prefetch-max-bytes N             maximum aggregate bytes prefetched per directory read
│  │  └─ --readdir-prefetch-timeout DURATION        timeout for one readdir prefetch batch
│  ├─ Overlay profiles:
│  │  ├─ --profile NAME                             mount profile: coding-agent, portable, none, interactive, or a custom
│  │  │                                             profile file
│  │  ├─ --local-root DIR                           local-only overlay storage root; auto-generated for overlay profiles
│  │  ├─ --local-only PATTERN                       additional local-only path pattern; repeatable
│  │  ├─ --remote-only PATTERN                      remote-persistent override path pattern; repeatable
│  │  ├─ --unpack :/archive.tar.gz                  restore a drive9 pack archive into local-root before mounting;
│  │  │                                             repeatable
│  │  └─ --no-auto-unpack                           disable automatic profile pack restore before mounting
│  ├─ Layers and writes:
│  │  ├─ --layer REF                                mount through writable fs layer by id, name, or tag ref
│  │  ├─ --checkpoint REF                           restore fs layer checkpoint before mounting
│  │  ├─ --durability MODE                          write durability: auto, interactive, fsync, close-sync, or write-sync
│  │  ├─ --flush-debounce DURATION                  debounce window for small-file flush coalescing; 0 disables
│  │  └─ --upload-concurrency N                     maximum concurrent background uploads issued by FUSE
│  ├─ Access and debug:
│  │  ├─ --allow-other                              allow other users to access mount
│  │  ├─ --read-only                                mount as read-only
│  │  └─ --debug                                    enable FUSE debug logging
│  ├─ Profiling:
│  │  ├─ --perf-dir DIR                             enable standard FUSE profiling outputs in this directory
│  │  ├─ --perf-interval DURATION                   continuous performance sample interval
│  │  ├─ --perf-max-samples N                       maximum samples per continuous perf JSONL segment
│  │  ├─ --perf-max-sample-files N                  maximum retained continuous perf sample files
│  │  ├─ --perf-max-profile-files N                 maximum retained CPU and heap profile files per type
│  │  ├─ --perf-cpu-duration DURATION               CPU profile capture window duration
│  │  ├─ --perf-cpu-interval DURATION               periodically capture CPU profiles at this interval
│  │  ├─ --perf-heap-interval DURATION              periodically write heap profiles at this interval
│  │  └─ --perf-addr ADDR                           serve live pprof, e.g. 127.0.0.1:6060
│  └─ Examples:
│     ├─ drive9 mount :/ ./mnt                                # mount root
│     ├─ drive9 mount --profile coding-agent :/workspace ./mnt
│     │  # profile mount
│     ├─ drive9 mount --layer layer_123 --checkpoint cp_456 :/team ./mnt
│     │  # restore a layer checkpoint before mounting
│     └─ drive9 mount --perf-dir ./perf --foreground :/ ./mnt # foreground mount with profiling outputs
│
├─ drive9 mount vault [flags] <mountpoint> → mount vault secrets read-only [READONLY]
│  ├─ --server URL                               drive9 server URL; overrides DRIVE9_SERVER and config
│  ├─ --api-key KEY                              owner API key; overrides DRIVE9_API_KEY and config
│  ├─ --foreground                               run in foreground and block until unmounted
│  ├─ --dir-ttl DURATION                         vault list / field cache TTL
│  ├─ --allow-other                              allow other users to access mount
│  ├─ --debug                                    enable FUSE debug logging
│  └─ Examples:
│     └─ drive9 mount vault ./vault                           # browse vault secrets locally
│
├─ drive9 umount [flags] <mountpoint> → unmount a drive9 mount and optionally pack overlay state
│  ├─ --timeout DURATION                         wait for flush/unmount
│  ├─ --pack :/archive.tar.gz                    write pack archive before unmount
│  ├─ --pack-path PATH                           include a local overlay path in the automatic/default pack; repeatable
│  ├─ --no-auto-pack                             skip configured auto-pack
│  └─ Examples:
│     ├─ drive9 umount ./mnt                                  # clean unmount
│     └─ drive9 umount --pack :/packs/work.tar.gz ./mnt       # pack on unmount
│
├─ drive9 doctor fuse → diagnose local runtime prerequisites
│  ├─ --mountpoint PATH                          mountpoint path to inspect
│  ├─ --cache-dir DIR                            cache directory to inspect
│  ├─ --server URL                               drive9 server URL to check
│  ├─ --timeout DURATION                         server connectivity timeout
│  └─ Examples:
│     ├─ drive9 doctor fuse                                   # check FUSE support
│     └─ drive9 doctor fuse --mountpoint ./mnt --server http://127.0.0.1:9009
│        # check a local dev mount target and server
│
└─ drive9 update [--check] [--force] [--base-url URL] → update the drive9 CLI binary in place [HTTPS]
   ├─ --check                                    report available update without installing
   ├─ --force                                    reinstall current or older release
   ├─ --base-url URL                             release root; HTTPS required by default
   └─ Examples:
      ├─ drive9 update --check                                # check latest release
      └─ drive9 update                                        # install newer release

Help controls:
  drive9 help --no-pager               # print without less
  drive9 help --color=always           # force ANSI color
  drive9 help --plain                  # classic script-friendly usage
  less -R -S                           # color-preserving vertical and horizontal scrolling
```

## Verification

- `env GOCACHE=/Users/mornyx/Desktop/repos/github.com/mem9-ai/drive9/.codex-gocache go test ./cmd/drive9`
- `env GOCACHE=/Users/mornyx/Desktop/repos/github.com/mem9-ai/drive9/.codex-gocache make build-cli`

## Summary by CodeRabbit

* **New Features**
  * Enhanced `help` command with visual tree rendering and formatting options (`--tree/-t`, `--no-pager`, `--color`).
  * Added `--manifest-url` option to `region list` command.

* **Documentation**
  * Updated help text and command descriptions for clarity.
  * Refined `create` command terminology.


srstack (commented):
## Review

### Overview

Adds `drive9 help` as a distinct command from `-h`/`--help`, rendering a tree-structured visual help page with color support, pager integration, and a data-driven `drive9VisualHelpCommands()` list. `visual_help.go` is self-contained and pure (no external state). Three tests cover the core paths.

---

### `visual_help.go`

**`printableLen` counts runes, not visual columns.**

`printableLen` strips ANSI escapes and counts Unicode code points via `utf8.DecodeRuneInString`. For ASCII-only content (which all current help text is) this is correct. But it will miscount wide characters (CJK, emoji, etc.) that occupy 2 terminal columns per rune. Not a concern for the current content, but worth noting if help text is ever localized.

**`shouldPageVisualHelp`: `width <= 0 || height <= 0` returns `true`.**

```go
if err != nil || width <= 0 || height <= 0 {
    return true
}
```

If `term.GetSize` succeeds but returns `width=0` (e.g. some terminal emulators during startup), this forces pager open unconditionally. A safer fallback would be `return false` (or a hardcoded threshold like `height < 30`) when the terminal size is unknown, to avoid an unexpected pager pop-up in headless environments that don't set a size but are technically a TTY. Low-priority since `isTerminal` is already checked first.

**`pageVisualHelp`: pipe write error handling.**

```go
_, writeErr := io.WriteString(in, rendered)
closeErr := in.Close()
waitErr := cmd.Wait()
if writeErr != nil { return writeErr }
if closeErr != nil { return closeErr }
return waitErr
```

If `less` exits before all input is consumed (user presses `q` before the end), `io.WriteString` returns a broken-pipe error. This causes `runHelp` to fall through to the direct `fmt.Fprint` path and print the entire output to the terminal again — which is surprising behavior. The broken-pipe case (`errors.Is(err, io.ErrClosedPipe)` or checking for `syscall.EPIPE`) should probably be treated as success (the user dismissed the pager intentionally). Other callers of `less` in the codebase typically swallow EPIPE from the write side.

**`colorFlagMetavars`: metavar detection is `A-Z` only.**

```go
for _, r := range field {
    if r < 'A' || r > 'Z' {
        allMeta = false
        break
    }
}
```

Metavars like `<mountpoint>`, `[profile]`, `YYYY-MM-DD`, `auto|fuse|webdav` — all handled correctly. But `+N|-N` (from `find -size`) would fail the all-caps check and not be styled as a metavar. Not visible since `+N|-N` appears inside a flag Desc, not a Name, but worth noting.

**`exampleCommand`: `i == 0 || (i == 1 && field != "help")` style logic.**

The intent is to color `drive9` and the subcommand, but not color `help` in the example `drive9 help --no-pager`. This works, but the condition is fragile — any two-word subcommand (`mount vault`, `region list`, etc.) has both words colored at `i==0` and `i==1` respectively, which is correct. The `!= "help"` carve-out is specifically for the help controls block where `drive9 help` is shown as an example. Fine, just worth documenting in a comment since it reads as magic otherwise.

**`renderVisualHelpCommand`: `nestedPrefix` logic for sections.**

```go
if lastChild {
    nestedPrefix += "   "
} else {
    nestedPrefix += "│  "
}
```

This correctly carries the vertical guide line for non-last sections. When `lastChild` is true the blank prefix is used (no vertical bar). Correct.

**No `context.Context` or I/O dependencies in data layer** — `drive9VisualHelpCommands()` is a pure value function. Easy to test in isolation, easy to update when commands change. Good design.

---

### `main.go`

**`help` case now calls `runHelp(args)`** — `args` here is the tail of `os.Args` after the command name, so `drive9 help --no-pager` correctly passes `["--no-pager"]` to `runHelp`. Correct.

**`-h`/`-help`/`--help` still call `usage(0)`** (the compact classic form). `drive9 help --plain` also reaches `usage(0)` via `runHelp`. So `--plain` and `-h` produce the same output. Consistent.

**Classic `usage()` text updated**: "provision a new database" → "provision a new tenant", `region list` gains `[--manifest-url URL]`, `help` added to commands. All consistent with current implementation.

---

### Tests

**`TestDispatchHelpCommandShowsVisualTreeHelp`** — checks `--no-pager --color=never`, verifies no ANSI escapes, checks for a representative set of strings. Good coverage of the happy path.

**`TestDispatchHelpCommandUsage`** — passes `--help` to `drive9 help`, expects `"usage: drive9 help"` prefix. Correct.

**`TestRenderDrive9VisualHelpColor`** — verifies color mode produces ANSI escapes and expected keywords. Good.

**Missing: test for `drive9 help --plain`** — verifies the plain path calls `usage()` and exits 0. Low priority since `TestDispatchLongHelpFlagShowsUsage` already covers `usage()` output indirectly.

**Missing: test for unknown option** — `drive9 help --bogus` should exit non-zero with an error. Worth adding to prevent regressions if the parse loop changes.

---

### Minor

- `visualHelpUsage()` documents `--tree, -t` as the default and lists it as a flag, but `parseVisualHelpOptions` silently ignores it (the case is a no-op comment). That's correct behavior, but the usage string might confuse users into thinking `-t` does something distinct. Consider removing the flag entry from the usage string since it adds no value when it's already the default.

- `newHelpStyle` uses `\x1b[1;38;5;213m` (256-color pink) for the badge. This won't render in 8-color terminals. Acceptable — the `NO_COLOR`/`TERM=dumb` guard handles the degraded case, and 256-color support is near-universal in modern terminals.

---

### Summary

Clean implementation. The only behavior issue worth fixing before merge is the broken-pipe / early-exit case in `pageVisualHelp` — right now if the user presses `q` in `less`, the full help is reprinted to the terminal. Everything else is minor or cosmetic.

mornyx:
@srstack Thanks for the detailed review.

I pushed `c7d5eaa` to address the actionable items:

- Treat pager closed-pipe / EPIPE writes as success, so quitting `less` early does not fall back to printing the full help page again.
- Avoid forcing the pager when terminal dimensions are unknown or zero.
- Added regression coverage for `drive9 help --plain` and unknown help options.
- Added direct coverage for pager closed-pipe classification.
- Removed `--tree, -t` from the help usage text since tree mode is already the default. The parser still accepts it as a compatibility no-op.

I left the wider terminal-column accounting and metavar-styling notes as future cleanup since the current help content remains ASCII-focused and those do not affect current behavior.

Verified with:

- `env GOCACHE=/Users/mornyx/Desktop/repos/github.com/mem9-ai/drive9/.codex-gocache go test ./cmd/drive9`
- `env GOCACHE=/Users/mornyx/Desktop/repos/github.com/mem9-ai/drive9/.codex-gocache make build-cli`

srstack (commented):
## Re-review

All issues from the previous review are addressed.

**1. Broken-pipe on early `less` exit — resolved**

`isPagerClosedPipe` now covers both `io.ErrClosedPipe` and `syscall.EPIPE`, and both `writeErr` and `closeErr` are guarded with it before returning. Pressing `q` in `less` no longer causes the full help to be reprinted. The function is also exported enough to be tested, and `TestIsPagerClosedPipe` directly verifies both error forms.

**2. `shouldPageVisualHelp` fallback on unknown terminal size — resolved**

`err != nil || width <= 0 || height <= 0` now returns `false` instead of `true`. Unknown terminal size no longer forces the pager open.

**3. `--tree`/`-t` removed from `visualHelpUsage()` — resolved**

The usage string no longer lists `--tree, -t` as a flag. The parse case is still present as a silent no-op, which is correct (accepting the flag gracefully without advertising it as meaningful).

**4. New tests added — `TestDispatchHelpCommandPlainShowsClassicUsage`, `TestDispatchHelpCommandUnknownOptionFails`**

Both previously-missing test cases are now present. `captureDispatchOutput` also extracts the repeated capture boilerplate across all help tests. Good cleanup.

**5. `TestDispatchHelpCommandPlainShowsClassicUsage` expects `stdout == ""`**

`--plain` calls `usage(0)` which writes to `os.Stderr`. The test correctly expects empty stdout and checks stderr for the usage content. Consistent with how other `usage()` callers work.

---

### One remaining minor observation (not blocking)

`isPagerClosedPipe` uses `syscall.EPIPE` directly. On Windows `syscall.EPIPE` does not exist (it's a Unix-only constant), so this file will fail to compile on Windows. Since `drive9 mount` requires FUSE, Windows isn't a supported platform today, and `golang.org/x/term` is already a Unix-leaning dependency in this file. This is fine as-is, but if Windows support is ever added, `isPagerClosedPipe` would need a build tag or an `errors.Is(err, fs.ErrClosed)` fallback.

---

LGTM.

mornyx:
@srstack Thanks again for the review. If the latest changes look good to you, could you please approve the PR? Thanks!

srstack (approved):

PR #583, opened 2026-06-18
feat: add live mount drain support

mornyx:
## Summary

This PR adds a live mount drain mechanism for Drive9 FUSE mounts. It supports both:

- native `sync -f <mountpoint>` through FUSE `SYNCFS`
- explicit `drive9 mount drain <mountpoint> --timeout=30s [--json]` for richer operational feedback

The main target use case is sandbox pause or host-side orchestration where the caller needs a clear signal that Drive9 has flushed dirty open handles and drained pending background writeback work before freezing or snapshotting the environment.

## Background

Drive9 already had `FlushAll()` for graceful unmount, but that path is intentionally destructive:

- it stops the write-back uploader
- it closes the commit queue work channel
- it closes local cache/journal/shadow-store resources

That behavior is correct for unmount, but too heavy for a pause/drain flow where the mount must stay alive after the drain completes. This PR introduces a non-destructive drain path that waits for the live system to become idle without shutting down its workers.

### go-fuse `SYNCFS` support

This PR replaces `github.com/hanwen/go-fuse/v2` with my fork:

- `github.com/mornyx/go-fuse/v2 v2.9.1-0.20260618063551-c6bad18bc819`
- branch: `feat/syncfs`
- commit: `c6bad18 feat: dispatch FUSE_SYNCFS`

The fork adds `RawFileSystem.SyncFs(...)` and dispatches FUSE opcode `SYNCFS` to the filesystem implementation. Drive9 implements this method in `Dat9FS.SyncFs`.

### Non-destructive drain primitive

Adds `Dat9FS.Drain(ctx)` as the shared drain implementation for both `sync -f` and the CLI control command.

The drain sequence is:

1. flush pending debounced small-file uploads
2. flush all currently open file handles
3. drain git workspace overlay writes and checkpoint pending git state
4. wait for legacy write-back uploader idle
5. wait for commit queue idle
6. wait for async kernel notification goroutines

The commit queue and write-back uploader now expose live `WaitIdle(ctx)` and `Snapshot()` helpers. These helpers do not close channels, stop workers, or reject future work.

### Mount control socket

FUSE mounts now publish a local Unix domain socket in mount state:

- mount state field: `control_socket`
- socket path: `/tmp/drive9-mount-<hash>.sock`
- socket permission: `0600`

`drive9 mount drain` reads the existing mount state, verifies it is a FUSE mount, dials the control socket, sends a JSON drain request, and returns either a human-readable summary or structured JSON.

The JSON result includes:

- success/failure
- mount point
- total duration
- per-phase durations
- error kind
- failed path when available
- open/dirty handle counts
- commit queue depth and bytes
- uploader queue/in-flight counts

### CLI behavior

New command:

```bash
drive9 mount drain [--timeout duration] [--json] <mountpoint>
```

Examples:

```bash
drive9 mount drain /mnt/drive9 --timeout=30s
drive9 mount drain /mnt/drive9 --timeout=30s --json
```

Native sync also works:

```bash
sync -f /mnt/drive9
```

The native `sync -f` path can only return an errno, so detailed diagnostics are intentionally exposed through the CLI control command.

## Notes

- This does not change unmount semantics. `FlushAll()` still performs the full destructive shutdown drain.
- The control socket is currently Unix-only. Windows has a compile-time no-op stub.
- Existing old mounts without `control_socket` will return a clear CLI error asking the user to remount with a Drive9 version that supports drain.
- The existing untracked local `blackbox/` and `blackbox-runs/` directories are not included.

## Tests

Ran:

```bash
go test ./pkg/fuse
go test ./cmd/drive9/cli
go test ./pkg/mountcontrol
go test ./pkg/mountstate
GOOS=windows go test -c ./cmd/drive9/cli
git diff --check
```

## Summary by CodeRabbit

* **New Features**
  * Added `drive9 mount drain` to drain a FUSE mount with `--timeout` and `--json`, including phase results, pending/idle metrics, and structured success/failure details.
* **Bug Fixes / Improvements**
  * Improved mount drain reliability by adding a dedicated mount control socket path and richer drain orchestration, with clearer error kind mapping and stronger “remaining pending work” detection.
* **Tests**
  * Added/expanded unit tests for drain validation and error handling, plus new end-to-end “drain semantics” coverage in the FUSE smoke test.


qiffang (changes requested):
Reviewed head `3b2eef0`. Verdict: **REVISE**.

### Blockers

**B1. `ok: true` can still report dirty/pending work, so the drain result is not a safe snapshot gate.**

`Dat9FS.Drain` leaves `OK` as `true` after all phases, then attaches `snapshotDrainPending()` without validating it (`pkg/fuse/drain.go:161-163`). `runMountDrain` also exits successfully whenever `resp.OK` is true (`cmd/drive9/cli/mount_drain.go:89-95`), and the new CLI unit test explicitly accepts `OK=true` with `pending.commit_queue_pending=1` (`cmd/drive9/cli/mount_drain_test.go:34-47`).

That means a concurrent writer can dirty a handle after its turn in `flush_open_handles`, or background work can appear after a wait phase, and the CLI/native control result can still say success. For the advertised freeze/snapshot use case, callers will naturally treat `ok: true` / exit 0 as the quiescence gate. If `pending.dirty_handles`, commit queue counts, or uploader counts are nonzero, `ok: true` is a false success.

Fix direction: define the contract explicitly. If `ok` means “safe to snapshot,” final non-open pending work must make the response fail, e.g. `pending_after_drain`. `open_handles` can remain nonzero, but `dirty_handles`, `commit_queue_pending`, `commit_queue_in_flight`, `commit_queue_delayed`, `uploader_queued`, and `uploader_in_flight` must be zero for success. Add a regression that an otherwise successful drain with residual dirty/pending work returns non-OK and makes `drive9 mount drain` exit nonzero.

**B2. Retryable open-handle flush failures lose their retryable errno on native `sync -f`.**

`newDrainStatusError` intentionally classifies `EAGAIN` from handle flushing as `remote_timeout_or_retryable` (`pkg/fuse/drain.go:47-51`), but `drainResponseStatus` only maps `timeout` and `canceled`; every other error kind falls through to `EIO` (`pkg/fuse/drain.go:320-328`). So a retryable remote timeout/retryable flush failure reaches native `sync -f` callers as a hard I/O error instead of `EAGAIN`.

Fix direction: map `remote_timeout_or_retryable` back to `syscall.EAGAIN`, and add a direct unit test for `drainResponseStatus` so this round-trip cannot regress.

**B3. The core `Drain()` state machine has no unit coverage.**

`pkg/fuse/drain.go` adds the central six-phase state machine used by both the CLI control path and native `SYNCFS`, but the unit tests added in this PR mock `requestDrain` and never exercise the real phase logic. The e2e covers only a success path. Missing coverage includes early phase failure skipping later phases, context cancellation during `drainOpenHandles`, handle flush failure preserving `ErrorKind` and `FailedPath`, `drainResponseStatus`, and final pending snapshot semantics.

This is not a stylistic test request. This function is the safety barrier for live mount snapshotting. Without direct tests, regressions in phase ordering, errno mapping, and failure propagation will pass CI.

### Non-blocking observations

**O1. Concurrent drain calls are not serialized.** `mountControlServer.serve` starts a handler goroutine per connection (`pkg/fuse/mount_control_unix.go:61-72`). Per-handle locking should prevent corruption, so I am not blocking on this, but concurrent drains can report confusing phase/pending results and do redundant flush work.

**O2. The control socket path is predictable in shared `/tmp`.** `ControlSocketPath` hashes only the canonical mount point (`pkg/mountstate/pid.go:56-60`). The socket is chmodded after `net.Listen`, so there is also a small permission window. This is mostly local DoS / unwanted drain triggering, not data exposure, but a per-user runtime directory or UID-scoped path would be stronger.

**O3. The go-fuse fork should have an explicit upstream/removal plan.** The `replace github.com/hanwen/go-fuse/v2 => github.com/mornyx/go-fuse/v2 ...` is small and understandable for `SYNCFS`, but it should not become silent permanent dependency debt.

### Validation run

- `/usr/local/go/bin/go test ./pkg/fuse ./pkg/mountcontrol ./pkg/mountstate ./cmd/drive9/cli`
- `/usr/local/go/bin/go vet ./pkg/fuse ./pkg/mountcontrol ./pkg/mountstate ./cmd/drive9/cli`

mornyx:
Thanks for the review. I addressed the drain blockers and the related actionable comments in 76dd4b99483f2d9ff4d367c76d01198459ba07e7.

Summary of changes:

- Drain now treats final residual work as a failure: dirty handles, commit queue pending/in-flight/delayed work, commit conflicts, uploader queued/in-flight work, and cached write-back entries all make the response non-OK. Clean open handles are still allowed.
- Added pending counters for commit conflicts and cached write-back entries to the drain JSON/human output and e2e validation.
- Preserved retryable open-handle flush failures by mapping `remote_timeout_or_retryable` back to `EAGAIN` for native `sync -f`.
- Serialized concurrent drains per mount to avoid duplicate flushes.
- Scoped the control socket path by user/runtime namespace and create the socket directory with `0700` permissions.
- Added direct unit coverage for `Dat9FS.Drain` final-pending semantics, clean-open-handle behavior, commit-conflict/cache leftovers, and `drainResponseStatus` mapping.
- Addressed the quick wins: package doc comment, import grouping, and logging non-fatal control-socket accept errors.

Validation run locally:

```bash
git diff --check
go test ./pkg/fuse ./pkg/mountcontrol ./pkg/mountstate ./cmd/drive9/cli
go test ./cmd/drive9 ./cmd/drive9/cli ./pkg/fuse ./pkg/mountcontrol ./pkg/mountstate -count=1
go vet ./pkg/fuse ./pkg/mountcontrol ./pkg/mountstate ./cmd/drive9/cli
GOOS=windows go test -c ./cmd/drive9/cli
```

qiffang (changes requested):
## adversary-2 R1 — REVISE (`3b2eef0`)

Overall: well-structured non-destructive drain with good CLI/e2e coverage. One blocker, several observations.

### B1: No unit tests for `Drain()` core function

`pkg/fuse/drain.go` adds 329 lines implementing 6 sequential phases (flush_debounced → flush_open_handles → git_overlay → writeback_uploader → commit_queue → notifications). The CLI tests use mocks that bypass the real phase logic. The e2e test covers the happy path. Missing:

- Phase ordering: early phase failure skips subsequent phases
- Timeout mid-phase: context cancellation during `drainOpenHandles` iteration
- Handle flush failure: `flushHandle` returns non-OK status → drain aborts with correct `ErrorKind`/`FailedPath`
- Pending snapshot accuracy: dirty handles counted correctly in response

Without unit tests for the core logic, regressions in phase ordering, error propagation, and timeout behavior are undetectable. This is the same drain path `sync -f` uses — correctness matters.

### O1 (non-blocking): Drain is not serialized

Two concurrent `sync -f` or CLI drain calls run simultaneously — both iterate open handles and flush them. Per-handle locking prevents corruption, but two drains can race and both report success while the second one flushed no-ops. Consider a `sync.Mutex` or `sync.Once`-style gate if exactly-once semantics are needed. Not blocking because concurrent flushes are safe (wasteful but correct).

### O2 (non-blocking): `waitGroupWithContext` goroutine leak on timeout

The spawned goroutine calling `wg.Wait()` stays alive until the WaitGroup hits zero even if context is canceled. For a drain timeout this is minor (notification goroutines will eventually finish), but worth documenting.

### O3 (non-blocking): Control socket TOCTOU

Socket created by `net.Listen` with umask-dependent permissions, then `Chmod(0o600)`. Brief window where another user on shared `/tmp` could connect. Mitigated by hash-based path unpredictability. Consider `umask(0o077)` before listen if hardening is desired.

### O4 (non-blocking): go-fuse fork tracking

`replace` directive pins to `mornyx/go-fuse/v2`. Track upstream PR status so the fork doesn't become permanent technical debt.

### O5 (non-blocking): `SyncFs` has no drain-side timeout

`fuseCtx(cancel)` derives timeout from kernel FUSE cancel channel. A `sync -f` with no kernel timeout runs the drain unbounded. The CLI path has `--timeout`, so this only affects the native `sync -f` path. Document this asymmetry.

Waiting for B1 fix (Drain unit tests covering phase ordering, timeout, and error propagation).

qiffang (commented):
## PR #583 Review — `feat: add live mount drain support`

Author: @mornyx | +1181/-13, 20 files | CI: all green

### Overall Assessment

Well-designed feature. The non-destructive drain primitive is exactly right for sandbox pause/snapshot workflows — FlushAll's destructive shutdown semantics would break live mounts. The architecture (Unix domain socket + JSON control protocol) is clean and extensible.

### 🟢 Good

1. **Clean separation**: `Drain()` is a standalone method on `Dat9FS`, reused by both `SyncFs` (kernel) and the CLI control path. No duplication.
2. **Phase-based reporting**: Each drain step is timed and reported individually — excellent for operational debugging. Failed path + error kind in the response is exactly the right data model.
3. **Non-destructive invariant**: None of `WaitIdle`, `Snapshot`, or `Drain` close channels, stop workers, or reject future work. This is the core design contract and it's preserved throughout.
4. **go-fuse fork is minimal**: 14 lines to dispatch `FUSE_SYNCFS` — low risk.
5. **E2e coverage**: Tests open-handle flush, post-drain writability, native `sync -f` with protocol version gating, and idle queue validation via JSON output.
6. **Windows no-op stub**: Compile-time clean separation.

### 🟡 Issues (should fix before merge)

**I1. Concurrent drain requests are unserialized** (P1)
`handleConn` starts a new `Drain` goroutine per connection. Two concurrent `drive9 mount drain` calls can flush the same dirty handles simultaneously, potentially causing double-upload or lock contention issues. Should add a `sync.Mutex` in `mountControlServer` and serialize drain calls. This is the highest-priority issue.

**I2. `EAGAIN` classification doesn't round-trip through `drainResponseStatus`** (P2)
`newDrainStatusError` classifies `EAGAIN` as `"remote_timeout_or_retryable"`, but `drainResponseStatus()` only maps `"timeout"` and `"canceled"` — everything else falls through to `EIO`. So a retryable FUSE error from `sync -f` becomes a permanent `EIO` instead of `EAGAIN`. Either map `"remote_timeout_or_retryable"` → `EAGAIN` in `drainResponseStatus`, or don't distinguish it.

**I3. Post-drain dirty state not validated** (P2)
After all phases succeed, `resp.OK` stays `true` even if `snapshotDrainPending()` shows `dirty_handles > 0` (e.g., a concurrent writer dirtied a handle after `flush_open_handles`). The response should either set `OK = false` or emit a warning field when pending work remains after a "successful" drain. Otherwise the orchestrator trusts `ok: true` but snapshots dirty state.

**I4. Socket path in shared `/tmp` is predictable** (P2)
`ControlSocketPath` uses `sha256(canonical_mount_point)[:8]` under `os.TempDir()`. In multi-user environments, another user could pre-create the socket path. Since `os.Remove(socketPath)` runs before `net.Listen`, a symlink at that path could point elsewhere. Consider using `os.MkdirTemp` or checking the socket owner after bind.

### 🟠 Suggestions (non-blocking)

**S1. `WaitIdle` polling interval**: Both `CommitQueue.WaitIdle` and `WriteBackUploader.WaitIdle` use 50ms polling. Consider condition variable or channel signal for lower overhead.

**S2. `SyncFs` timeout is fixed `fuseCtx` default**: Native `sync -f` has no user-configurable timeout. Document that heavy workloads should prefer `drive9 mount drain --timeout=60s`.

**S3. `debouncer.FlushAll()` doesn't honor drain context**: Inherited behavior, but worth noting.

**S4. No unit tests for `Drain()` core logic**: CLI tests mock the control path. E2e covers happy path. Missing: phase ordering on early failure, timeout mid-phase, handle flush failure error propagation, pending snapshot accuracy.

mornyx:
Thanks for the follow-up review. I addressed the remaining high-priority drain serialization feedback in `687186c1e1e0adc8e5abdbb5662dceea330f8b62`.

Summary:

- Added a `mountControlServer`-level mutex so concurrent `drive9 mount drain` control-socket requests are serialized before invoking `Dat9FS.Drain`.
- Kept the existing `Dat9FS` drain mutex as the core in-process guard, while also ensuring the control-server request path itself no longer starts multiple drains concurrently.

Validation run locally:

```bash
git diff --check
go test ./pkg/fuse ./pkg/mountcontrol ./pkg/mountstate ./cmd/drive9/cli -count=1
go test ./cmd/drive9 ./cmd/drive9/cli ./pkg/fuse ./pkg/mountcontrol ./pkg/mountstate -count=1
go vet ./pkg/fuse ./pkg/mountcontrol ./pkg/mountstate ./cmd/drive9/cli
GOOS=windows go test -c ./cmd/drive9/cli
```

qiffang (approved):
## R2 Review — GREEN

R1 blockers addressed:
- B1 (drain final state verification): `drainPendingWorkError` now fails the response when dirty/pending work remains after all phases
- B1 (unit tests): `drain_test.go` covers clean handles, dirty handles, commit conflicts, uploader cache, response status mapping
- Serialization: `drainMu` on Dat9FS serializes Drain + SyncFs

No blockers found. Implementation is solid — drain sequence ordering is correct, post-drain verification catches all residual dirty state, control socket security is proper, error mapping is sensible, testing coverage is comprehensive.

Non-blocking observations:
- O1: Redundant `drainMu` on `mountControlServer` (already serialized by `fs.drainMu`)
- O2: go-fuse fork dependency — consider upstreaming SyncFs dispatch
- O3: `WriteBackUploader.WaitIdle` doesn't check cached entries (caught by post-drain verification, but could be clearer)

qiffang (approved):
## R2 — adversary-2 on `687186c1`

**Verdict: GREEN** (conditional on CI green)

### R1 blocker verification — all fixed

| Blocker | Status |
|---|---|
| B1 (adversary-2): no unit tests for `Drain()` core | ✅ `drain_test.go` — 4 tests covering clean/dirty/conflict/cached scenarios |
| B1 (adversary-1): `ok:true` without pending validation | ✅ Post-phase `drainPendingWorkError()` checks all dirty/queue/uploader counters |
| B2 (adversary-1): EAGAIN → EIO mapping loss | ✅ `drainResponseStatus()` maps `remote_timeout_or_retryable` → EAGAIN with test coverage |
| I1 (dat9-dev1): concurrent drain unserialized | ✅ `fs.drainMu` serializes Drain + SyncFs; `mountControlServer.drainMu` adds socket-level serialization |
| I4 (dat9-dev1): socket path predictable | ✅ `ControlSocketPath()` uses sha256 hash under XDG_RUNTIME_DIR/user-scoped tmpdir with 0700+0600 perms |

No new blockers found. Checked waitGroupWithContext leak boundary, Snapshot lock ordering, WaitIdle vs Cached layering, timeout bounds, and redundant lock analysis.

qiffang (approved):
R2 review on `687186c1`.

Verdict: GREEN.

R1 blockers verified fixed:
- Final drain state is now validated after all phases. `Drain()` snapshots dirty handles, commit queue state, uploader queue/in-flight/cache, and fails with `pending_work_remaining` if any undrained work remains.
- `SyncFs` no longer collapses retryable remote timeout/`EAGAIN` into `EIO`; `drainResponseStatus` maps timeout, retryable, cancellation, and default error kinds explicitly.
- Core `Drain()` behavior now has direct regression tests covering clean handles, final dirty handles, commit conflicts, uploader cached entries, and status mapping.
- Concurrent drain calls are serialized through `Dat9FS.drainMu`; the control socket also serializes requests.

Additional checks:
- Control socket path is under `XDG_RUNTIME_DIR` or a per-UID `0700` temp dir, with the socket `0600`.
- CLI reports non-OK drain responses as command failures while still emitting JSON when requested.
- Windows stub compiles.

Local validation:
- `go test ./pkg/fuse ./pkg/mountcontrol ./pkg/mountstate ./cmd/drive9/cli`
- `go test ./cmd/drive9`
- `go vet ./pkg/fuse ./pkg/mountcontrol ./pkg/mountstate ./cmd/drive9/cli`
- `GOOS=windows go test -c ./cmd/drive9/cli`
- `git diff --check refs/remotes/mem9/main...HEAD`
- `gofmt -l` on changed Go files

Non-blocking: the `go-fuse` fork dependency should be tracked and ideally upstreamed later, since `SyncFs` support is now tied to that fork.

PR #586, opened 2026-06-18
Support TiDB Cloud Native fork

srstack:
## Summary
- add credential-aware TiDB Cloud Native branch create/wait/delete support
- allow /v1/fork and `drive9 ctx fork` to accept TiDB Cloud public/private keys for native tenants
- keep native fork branch deletion request-scoped while reusing `object_gc_candidates` for per-object S3 cleanup
- avoid revoking fork owner API keys until native delete cleanup succeeds so failed deletes can be retried

## Self-review notes
- Fixed native delete retryability: failed cleanup no longer revokes the drive9 owner API key before the user can retry with TiDB Cloud credentials.
- Fixed native create cleanup context: branch cleanup after create failure uses a background trace context so client disconnects do not immediately cancel cloud cleanup.
- `tenant_delete_jobs` remains scoped to live tenant namespace prefix deletion; fork cleanup uses existing `object_gc_candidates` because forks share storage namespaces.

## Tests
- `go test ./pkg/tenant ./pkg/tenant/starter ./pkg/tenant/tidbcloudnative`
- `go test ./pkg/server -run "^$"`
- `go test ./cmd/drive9 -run "^Test.*Usage|Test.*Help|TestDispatch|TestCtx"`
- `go test ./cmd/drive9/cli -run TestCtxFork`
- `git diff --check`

## Notes
- Focused `pkg/server` fork tests currently fail in this workspace before reaching the new assertions because the local test DB has schema drift: `llm_usage` is missing `tenant_id`, so meta schema repair fails creating `idx_llm_usage_tenant_created`.

## Release Notes

* **New Features**
  * Added `--tidbcloud-public-key` and `--tidbcloud-private-key` flags to `drive9 ctx fork`, reflected in help output.
  * In TiDB Cloud mode, `ctx fork` now supports credentialed fork/branch creation and deletion, and carries extra context metadata on successful forks.

* **Bug Fixes**
  * Improved fork request/response handling with stricter key validation, trimmed inputs, and safer payload construction.
  * Refined cleanup behavior for failed or deleting forks to better honor provided credentials.

* **Tests**
  * Expanded coverage for TiDB Cloud credential propagation and fork/branch cleanup flows.


srstack (commented):
## Review: Support TiDB Cloud Native fork

Clean, well-structured PR. The interface design is good and the test coverage is solid. A few things to look at before merge.

---

### pkg/tenant/provisioner.go — CredentialBranchProvisioner interface

`ProvisionBranchWithCredentials` is declared in `CredentialBranchProvisioner` but the server never calls it directly — `createForkBranch` and `waitForForkBranchActive` dispatch to `CreateBranchWithCredentials` + `WaitForBranchActiveWithCredentials` individually. The combined method is only used internally by the tidbcloudnative provisioner. If it's not part of the public contract needed by the server, it doesn't need to be in the interface. Either use it from the server side or move it out of the interface to reduce the implementation surface for future provisioners.

---

### pkg/server/fork.go — credential thread-through in cleanup path

`cleanupForkTenant` (the async path via `startForkCleanup`) calls `cleanupForkTenantOnce(ctx, tenantID, nil)` with nil credentials. For a native fork in `TenantDeleting` state that lands on the async cleanup path, `deleteForkBranch` will be called with nil — which dispatches to `DeleteBranch` → `DefaultCredentials()`. If the server has no default credentials configured, the branch delete silently returns `ErrCredentialsRequired` and the fork stays stuck in `TenantDeleting` forever.

The async cleanup path is triggered from `startForkCleanup`, which is only called for the non-native provider path. But it's worth confirming: is there any code path that could schedule `cleanupForkTenant` for a native fork? If yes, it'll silently fail. If no, a guard at the top of `cleanupForkTenant` checking for native provider + nil credentials would make this invariant explicit.

---

### pkg/server/fork.go — handleForkDelete retryability for native forks

The self-review note says "failed cleanup no longer revokes the drive9 owner API key before the user can retry". This is correctly implemented: for native forks, `RevokeTenantAPIKeys` is only called after `cleanupForkTenantOnce` succeeds. Good.

One edge case: if `cleanupForkTenantOnce` fails on the enqueue-refs step (before branch delete), the tenant remains in `TenantDeleting`. On retry, `cleanupForkTenantOnce` is called again from the `TenantDeleting` branch at the top of `handleForkDelete`. This correctly re-runs the full cleanup sequence. The refs enqueue may re-insert some already-enqueued GC candidates, but since `object_gc_candidates` is presumably deduplicated or idempotent on insert, this is fine. Worth a brief comment confirming this assumption.

---

### pkg/tenant/tidbcloudnative/provisioner.go — CreateBranchWithCredentials early-ACTIVE path

After `createBranch` returns:

```go
if branch.State != "" && branch.State != "ACTIVE" {
    return out, nil
}
if branch.State == "ACTIVE" || branch.Endpoints.Public.Host != "" || ... {
    if err := fillBranchEndpoint(out, branch); err != nil {
        return out, err
    }
}
```

If `branch.State == ""` (API returned no state field), the first condition is false, so it falls through to the second. If the API also returns no endpoints, `fillBranchEndpoint` is never called and the caller gets an `out` with no host/port — silently incomplete. Consider returning an error when state is empty and endpoints are also missing after create.

---

### pkg/server/fork.go — deleteForkBranchOrPersist return value

The function now returns `bool` (true = deleted, false = persisted), but the three call sites in `ensureForkBranchConnection` and `createForkTenant` ignore the return value. This is correct because those callers already return an error immediately after. A short doc comment on the function explaining the bool semantics would help future readers.

---

### cmd/drive9/cli/ctx.go — private key in process args

`--tidbcloud-private-key KEY` passes the private key as a CLI argument. On Linux/macOS, command-line arguments are visible in `/proc/<pid>/cmdline` and `ps aux` to same-UID or root processes. Consider adding a note in the flag help text recommending the env var path (`DRIVE9_TIDBCLOUD_PRIVATE_KEY`) for non-interactive/CI use, similar to how database tools warn about `--password` flags.

---

### Test coverage

`TestBranchWithCredentialsUsesRequestCredentials` is thorough — verifies digest auth uses the request public key and doesn't leak the private key. Good.

`TestCreateForkTenantNativeUsesCredentialsAndDeletesBranchOnFailure` correctly verifies branch deletion on provisioning error. The one missing case is a failure in `provisionForkTenantOnceWithCredentials` (schema init) after branch creation succeeds — the current test cancels context on `CreateBranch` itself, so schema init is never reached. A test for schema-init failure specifically would give fuller coverage of the `markForkDeletedOrFailedAfterBranchDelete` path.

srstack (commented):
## Re-review: Support TiDB Cloud Native fork

Two real bugs surfaced by CodeRabbit that I agree with. One from my earlier review has been addressed. Rest are nitpicks.

---

### Bug 1 (new): `ctxForkUsesTiDBCloudCredentials` over-fires — cmd/drive9/cli/ctx.go:599

```go
func ctxForkUsesTiDBCloudCredentials(ctx *Context) bool {
    ...
    return strings.EqualFold(mode, ModeLabelTiDBCloud) ||
        strings.EqualFold(mode, RegionModeTiDBCloudNative) ||
        strings.TrimSpace(ctx.CloudProvider) != "" ||   // ← too broad
        strings.TrimSpace(ctx.Region) != ""             // ← too broad
}
```

This function controls whether `consumeEnv(EnvTiDBCloudPublicKey/PrivateKey)` is called. Any context with a non-empty `CloudProvider` or `Region` field — even non-TiDB contexts — will silently consume those env vars and inject them into the fork request body. A non-TiDB fork on a user's machine that happens to have `DRIVE9_TIDBCLOUD_PUBLIC_KEY` set in the environment would get credentials attached, causing a 400 from the server's `decodeCredentialRequest` if the server doesn't recognize them.

The fix is straightforward: the last two clauses should be removed. Mode is the authoritative discriminator. CloudProvider/Region are metadata about the cluster, not indicators that TiDB Cloud credentials are needed:

```go
return strings.EqualFold(mode, ModeLabelTiDBCloud) ||
    strings.EqualFold(mode, RegionModeTiDBCloudNative)
```

---

### Bug 2 (new): credential guard fires before BranchID check — pkg/server/fork.go:875

```go
func (s *Server) cleanupForkTenantOnce(..., credentialReq *tenant.CredentialProvisionRequest) error {
    t, err := s.meta.GetTenant(...)
    ...
    // Line 875 — checks credentials before line 882 BranchID check
    if t.Provider == tenant.ProviderTiDBCloudNative && credentialReq == nil && resolveDefaultCredentials(s.provisioner) == nil {
        return tenant.ErrCredentialsRequired
    }
    ...
    if t.BranchID == "" {   // Line 882 — no branch API call needed here
        logger.Warn(...)
        if t.Status == meta.TenantFailed || t.Status == meta.TenantDeleting {
            _ = s.meta.UpdateTenantStatus(ctx, tenantID, meta.TenantDeleted)
        }
        return nil
    }
    ...
}
```

When `BranchID == ""` (provisioning failed before the branch was created), cleanup only needs to update the DB status — no branch API call is made. But the credential guard at line 875 fires first and returns `ErrCredentialsRequired`, leaving the tenant stuck in `TenantFailed` or `TenantDeleting` permanently with no way to clean up without credentials that weren't originally required.

The fix is to move the credential guard to after the BranchID check:

```go
if t.BranchID == "" {
    // cleanup doesn't require branch API — no credentials needed
    if t.Status == meta.TenantFailed || t.Status == meta.TenantDeleting {
        _ = s.meta.UpdateTenantStatus(ctx, tenantID, meta.TenantDeleted)
    }
    return nil
}
// Only now require credentials — we're about to call the branch API
if t.Provider == tenant.ProviderTiDBCloudNative && credentialReq == nil && resolveDefaultCredentials(s.provisioner) == nil {
    return tenant.ErrCredentialsRequired
}
```

---

### From my previous review — cleanupForkTenant async path with nil credentials

The comment added at line 899 addresses the GC dedup concern. The concern about `cleanupForkTenant` being called for native forks with nil credentials via the async path: bug 2 above is actually a concrete manifestation of this — the credential check blocking BranchID-less cleanup is the real symptom. Once bug 2 is fixed, the nil-credential async path for native forks is safe for the BranchID-less case (cleanup proceeds) and the BranchID-present case (credential check fires correctly).

---

### Nitpicks (CodeRabbit suggestions, all valid)

**Error wrapping in provisioner.go**: Raw `return nil, err` at lines 299, 304, 314, 369, 396 lose context in production logs. Wrapping with `fmt.Errorf("create branch: %w", err)` etc. is the existing pattern in this file and worth applying here.

**Typed state constant**: `"ACTIVE"` appears 3 times (lines 331, 335, in `branchConnectionIncomplete`). A `const branchStateActive = "ACTIVE"` reduces typo risk, though this is minor given the string is the direct API value.

**Test gap**: `TestCtxForkDoesNotSendTiDBCloudEnvForUnlabelledContext` should have a parallel case where `CloudProvider`/`Region` are set but `Mode` is non-TiDB, asserting credentials are still not forwarded. This is especially important given bug 1 above — it would have caught the regression.

srstack (commented):
## Re-review (commit c4f7ed3): address fork review followups

All issues from the previous round are addressed. LGTM.

**Bug 1 fixed** — `ctxForkUsesTiDBCloudCredentials` now only checks `Mode`, dropping the over-broad `CloudProvider`/`Region` clauses. The regression test (`TestCtxForkDoesNotSendTiDBCloudEnvForNonTiDBMetadataContext`) now sets `CloudProvider: "aws"` and `Region: "us-west-2"` on a non-TiDB context and confirms credentials are not forwarded. Correct.

**Bug 2 fixed** — The credential guard in `cleanupForkTenantOnce` is now after the `BranchID == ""` early-return, so a native fork that failed before branch creation can be cleaned up without credentials. The new test `TestCleanupNativeFailedForkWithoutBranchIDDoesNotRequireCredentials` directly covers this path — calls `cleanupForkTenantOnce` with `nil` credentials on a native fork with empty `BranchID`, expects `TenantDeleted` status and zero branch delete calls. Correct.

**Nitpicks addressed** — `stateActive = "ACTIVE"` constant added and applied at all six call sites. Error wrapping with `fmt.Errorf("...: %w", err)` applied to `CreateBranchWithCredentials`, `WaitForBranchActiveWithCredentials`, `DeleteBranchWithCredentials`.

One small note on the credential guard placement: it now sits between the `BranchID == ""` early-return and the `t.Status != meta.TenantDeleting` branch. This means `cleanupFailedForkBranch` (the non-deleting path) is also guarded, which is correct since it will call the branch API. No issue.

qiffang (approved):

cmd/drive9/cli/ctx.go
```diff
@@ -493,6 +522,19 @@ func ctxForkCmd(args []string) error {
 	if source.Type != PrincipalOwner || source.APIKey == "" {
 		return fmt.Errorf("ctx fork requires an owner context; %q is %q", fromName, source.Type)
 	}
+	if (publicKey == "" || privateKey == "") && (publicKeyGiven || privateKeyGiven || ctxForkUsesTiDBCloudCredentials(source)) {
```
mornyx, line 525: Please preserve the source TiDB Cloud metadata when saving the forked context (at least `Mode`, and likely `CloudProvider`/`Region` as well). The new auto-env credential path depends on `ctxForkUsesTiDBCloudCredentials(source)`, but the context saved below only has `Type`, `Server`, and `APIKey`. After `drive9 ctx fork` creates a native fork and the user switches to it, a subsequent fork from that context will no longer auto-consume `DRIVE9_PUBLIC_KEY`/`DRIVE9_PRIVATE_KEY`; the CLI will omit credentials and the server will reject the native fork unless the user repeats both flags manually.

mornyx (changes requested):
I re-reviewed head 86eda1bc2d43e8f9fa6f71426afcfdf0b18f21c7. Local validation passed for the non-MySQL packages; `pkg/server`/aggregate `pkg/tenant` are blocked by the local testcontainers runtime. One remaining edge case needs attention.

pkg/server/fork.go
```diff
@@ -631,11 +814,36 @@ func (s *Server) handleForkDelete(w http.ResponseWriter, r *http.Request) {
 		errJSON(w, http.StatusNotFound, "tenant not found")
 		return
 	}
+	credentialReq, err = s.resolveForkCredentialRequest(t.Provider, credentialReq)
```
mornyx: This resolves TiDB Cloud credentials before checking whether cleanup actually needs to call the upstream branch API. For a native fork that is already `failed`/`deleting` but has no `BranchID`, `cleanupForkTenantOnce` can mark the local fork deleted without any upstream credentials. As written, if no request credentials and no server defaults are available, this handler returns 400 before reaching that no-branch cleanup path, leaving the local failed fork record undeletable via the API. Please defer credential resolution until after the no-branch case is handled, or make the cleanup path decide whether credentials are required.

mornyx (commented):
One additional native fork delete edge case from the same review round is inline below.

srstack (commented):
## Re-review verdict: LGTM (head e541cf5)

(Posting as a comment — GitHub blocks me from formally approving since this PR is under the same account.)

Both previously-unresolved review threads are resolved, each backed by a regression test I verified locally.

**Thread 1 — fork context metadata inheritance (`cmd/drive9/cli/ctx.go`)**
Forked context now inherits `Mode/CloudProvider/Region` from source, so a subsequent native fork takes the env credential path. The regression test (`TestCtxForkPreservesModeMetadataForSubsequentEnvCredentialFork`) was corrected to start the second fork from the forked context (`--from exp`). Verified it is a genuine regression test:
- with fix: PASS
- with the inheritance fix reverted: FAIL (`child fork request body = {PublicKey: PrivateKey:}, want env creds`)

**Thread 2 — native branchless delete credentials (`pkg/server/fork.go`)**
`handleForkDelete` resolves native credentials only when `BranchID != ""`, so a native failed/deleting fork with empty `BranchID` and no request/default credentials is cleaned up locally instead of returning 400. Credential resolution + validation preserved on the branch-API path. Covered by `TestHandleForkDeleteNativeFailedBranchIDEmptyNoCredentialsDeletesLocally` (202 / tenant deleted / no branch delete).

CI: CodeRabbit, ci, e2e smoke all green on e541cf5. Full `TestCtxFork*` suite passes locally. (Server handler tests couldn't run locally due to a testcontainers MySQL schema-repair issue unrelated to this PR; Thread 2 verified via code review + CI.)

qiffang (changes requested):
## R1 Review on `e541cf5d` — REVISE (2 blockers)

mornyx's 2 findings are fixed in commits eaa1834f–e541cf5d (metadata preservation + credential resolution deferral). Tests added.

### B1: `handleForkDelete` has 3 near-identical native cleanup blocks

Lines 818-835, 838-856, 865-883 duplicate the same credential-resolve + cleanupForkTenantOnce + RevokeTenantAPIKeys + response pattern for TenantDeleting, TenantFailed, and active→deleting. The `6542cf96` fix had to be applied identically in all 3 blocks, proving the maintenance hazard.

**Fix**: Extract a single helper and call it from all 3 branches.

### B2: `createForkTenant` has 4 copies of provider-conditional cleanup

```go
if source.Provider == tenant.ProviderTiDBCloudNative {
    s.markForkDeletedOrFailedAfterBranchDelete(...)
} else {
    s.markForkFailedAndCleanup(...)
}
```

4 copies. Extract a single `cleanupOnForkFailure(...)` method.

- O1: `decodeForkRequest` shared between create and delete — safe now, but document the contract.
- O2: `consumeEnv` side effect is consistent with existing pattern.
- O3: `visual_help.go` mentions only DRIVE9_PRIVATE_KEY, not DRIVE9_PUBLIC_KEY.

CI green. Tests comprehensive.

qiffang (changes requested):
## R1 — REVISE

### B1 — `DisallowUnknownFields()` on public fork API is a wire-level breaking change

`decodeForkRequest` adds `dec.DisallowUnknownFields()` to both create and delete endpoints. The old code silently accepted unknown JSON fields. After this change, any client sending an extra field (older/newer CLI, third-party integration) gets HTTP 400.

This is a public API — `DisallowUnknownFields` on a server-side decoder breaks forward/backward compatibility. A newer CLI that adds an optional field will break against an older server, and vice versa.

**Fix**: Remove `dec.DisallowUnknownFields()`. Keep the trailing-data check.

### Non-blocking observations

- O1: `handleForkDelete` has three nearly identical native-provider blocks (TenantDeleting/TenantFailed/active) — extract shared helper.
- O2: `fmt.Sprintf("fork delete cleanup failed: %v", err)` leaks internal errors to HTTP response body.
- O3: `cleanupForkTenantOnce` native credential guard runs after side effects (`pool.InvalidateAndWait`, `AbortActiveUploadReservations`).

qiffang (commented):
## R1 Review on `ad9c7b7` — REVISE (1 blocker)

Prior blockers from @qiffang and @mornyx are fixed. One remaining:

**B1: `createForkTenant` has 3 identical provider-conditional cleanup blocks (lines 383-386, 392-395, 411-414)**

```go
if source.Provider == tenant.ProviderTiDBCloudNative {
    s.markForkDeletedOrFailedAfterBranchDelete(ctx, forkID, credentialReq, cluster)
} else {
    s.markForkFailedAndCleanup(ctx, forkID)
}
```

This is the create-path half of @qiffang's original B2 — the delete path was fixed via `cleanupNativeFork`, but the create path still has 3 identical conditionals for token issue, encrypt, and InsertAPIKey errors.

Extract to a `cleanupForkCreateError(ctx, forkID, provider, credentialReq, cluster)` helper to prevent consistency drift (the `6542cf96` fix already had to be applied identically across these copy-paste sites).

**N1 (non-blocking)**: `ensureForkBranchConnection` line 630 calls `deleteForkBranchOrPersist(ctx, ...)` without `backgroundWithTrace`, while `createForkTenant` line 346 uses `backgroundWithTrace(ctx)` for the equivalent cleanup. Worth aligning.

Everything else is solid — clean interface design, good test coverage (10 new tests), proper credential resolution flow.

qiffang (changes requested):
R1 review on `ad9c7b7` — REVISE

I reviewed the current head against the PR diff, prior comments, server fork lifecycle, TiDB Cloud Native provisioner methods, and CLI credential flow.

Blocker:

1. `createForkTenant` still has three duplicated provider-specific cleanup branches for token issue, token encryption, and API key insert failures.

Current pattern is repeated at the same semantic boundary:

```go
if source.Provider == tenant.ProviderTiDBCloudNative {
    s.markForkDeletedOrFailedAfterBranchDelete(ctx, forkID, credentialReq, cluster)
} else {
    s.markForkFailedAndCleanup(ctx, forkID)
}
```

Why this blocks: this is not just style. This exact cleanup decision is where TiDB Cloud Native must preserve retryability when branch deletion fails, while starter forks use async failed cleanup. Recent credential/delete fixes already had to be applied consistently across fork cleanup callsites, so keeping this copy-paste keeps a proven drift point in the highest-risk failure path. Extract a single helper and call it from all three sites.

Suggested shape:

```go
func (s *Server) cleanupForkCreateError(ctx context.Context, forkID, provider string, credentialReq *tenant.CredentialProvisionRequest, cluster *tenant.ClusterInfo) {
    if provider == tenant.ProviderTiDBCloudNative {
        s.markForkDeletedOrFailedAfterBranchDelete(ctx, forkID, credentialReq, cluster)
        return
    }
    s.markForkFailedAndCleanup(ctx, forkID)
}
```

Non-blocking observations:

- The `ctx_test.go` request-body captures still mutate/read `gotBody` across the httptest handler goroutine and test goroutine. Convert those captures to a channel or lock if race testing is expected.
- `ensureForkBranchConnection` uses request context for `deleteForkBranchOrPersist` cleanup after create/update failures, unlike the top-level create path. Low risk for async worker calls, but align to `backgroundWithTrace(ctx)` for consistency.
- `cleanupForkTenantOnce` still does backend drain / reservation abort before its native credentials guard on background cleanup. `cleanupNativeFork` pre-resolves credentials for HTTP delete, so user retry path is protected; background resume without defaults can still perform local side effects before returning `ErrCredentialsRequired`.

Validation run locally:

- `go test ./pkg/tenant/starter ./pkg/tenant/tidbcloudnative` passed.
- `go test ./cmd/drive9/cli -run 'TestCtxFork|TestDecode'` passed.
- `go test ./cmd/drive9 -run '^Test.*Usage|Test.*Help|TestDispatch|TestCtx'` passed.
- `go test -c ./pkg/server ./cmd/drive9 ./cmd/drive9/cli ./pkg/tenant/tidbcloudnative ./pkg/tenant/starter` passed.
- `git diff --check 6d7884199d235f070368f11704b53b5a4ebb24f6...HEAD` passed.
- `go test ./pkg/tenant` is blocked locally by testcontainers/rootless Docker, same environment limitation as other DB-backed tests.

Verdict: request changes until the duplicated create-error cleanup dispatch is extracted into one helper.

mornyx (approved):
I re-reviewed the latest head `27387064ef6e686609db5b983945a23fcd720553`. The previous create-path cleanup duplication blocker is addressed by `cleanupForkCreateFailure`, and the current GitHub checks are green.

Local validation performed:

- `go test ./cmd/drive9/cli ./pkg/tenant/tidbcloudnative ./pkg/fuse -run 'Test.*Fork|Test.*Drain|Test.*Credential|Test.*Branch|Test.*Cleanup' -count=1`
- `go test -c ./pkg/server -o /tmp/pr586-server.test`

I also re-checked the two candidate credential-flow issues from the prior pass: forked contexts now preserve `Mode`/`CloudProvider`/`Region`, and native cleanup only resolves TiDB Cloud credentials when a branch ID exists. I do not see remaining blocking issues in this review round.

qiffang:
## adversary-1 review on `27387064` — REVISE before merge

I re-reviewed the latest head after the create-path cleanup helper extraction. The previous duplicated-cleanup blocker is fixed, and the focused CLI/tenant checks I ran are green. I still see one request-scoped native failure path that can strand a fork without a user retry surface.

### Blocker: native create can persist a failed branch before any fork owner API key exists

For TiDB Cloud Native forks, `createForkTenant` runs the synchronous branch/schema/finalization path before issuing and inserting the fork owner API key:

```go
if source.Provider == tenant.ProviderTiDBCloudNative {
    if err := s.provisionForkTenantOnceWithCredentials(ctx, forkID, credentialReq); err != nil {
        s.markForkDeletedOrFailedAfterBranchDelete(ctx, forkID, credentialReq, cluster)
        return nil, err
    }
}

apiToken, err := token.IssueToken(...)
...
s.meta.InsertAPIKey(...)
```

If `provisionForkTenantOnceWithCredentials` fails after the TiDB branch exists (schema init, system user setup, quota backfill, etc.) and `markForkDeletedOrFailedAfterBranchDelete` cannot delete the cloud branch, `deleteForkBranchOrPersist` persists `cluster_id`/`branch_id` and the tenant is marked `failed`. At that point no owner API key has been inserted or returned to the caller yet.

That makes the request-scoped cleanup story break down:

- `/v1/fork` delete operates on the current tenant scope, so the user needs an API key for the failed fork tenant to retry cleanup with TiDB Cloud credentials.
- In this path the caller only still has the source tenant context; the fork key was never created or returned.
- Parent tenant delete is also blocked while the failed fork remains non-deleted (`NamespaceHasNonDeletedFork` treats every fork status except `deleted` as blocking).
- Startup cleanup scans failed forks, but without server default credentials it still cannot delete a native branch; request-scoped credentials cannot be supplied because there is no fork credential to authorize the request.

This appears to violate the PR goal of keeping failed native cleanup retryable. The current code preserves retryability for failures after the owner key exists, but not for failures before `IssueToken` / `InsertAPIKey`.

Suggested fix: ensure there is a retryable user-visible cleanup surface before a native branch-backed fork can be persisted as `failed`. For example, issue/insert the fork owner API key before the synchronous native provisioning/finalization step, and if cleanup cannot delete the branch, return a response that gives the caller the fork key/status so they can retry `DELETE /v1/fork` with credentials. Alternatively, add a source-owner/admin cleanup endpoint that can target the failed child fork by tenant ID. Add a regression where native schema/finalization fails and branch delete also fails; assert the fork is either fully deleted or has an active owner key / reachable retry path.

### Non-blocking

- `ensureForkBranchConnection` still returns `starter branch response missing required metadata` for the branch-metadata error path. Since native now uses the same helper, a provider-neutral message would be clearer.
- CodeRabbit's `ctx_test.go` `t.Fatalf` vs `t.Errorf` comment is valid style cleanup per AGENTS.md, but not a merge blocker.

### Validation run locally

- `go test ./pkg/tenant ./pkg/tenant/starter ./pkg/tenant/tidbcloudnative -count=1` PASS
- `go test ./pkg/tenant/tidbcloudnative -count=1` PASS
- `go test ./cmd/drive9 -run '^Test.*Usage|Test.*Help|TestDispatch|TestCtx' -count=1` PASS\n- `go test ./cmd/drive9/cli -run TestCtxFork -count=1` PASS\n- `go test -race ./cmd/drive9/cli -run TestCtxForkPreservesModeMetadataForSubsequentEnvCredentialFork -count=1` PASS\n- `go test ./pkg/fuse -run 'Test.*Drain' -count=1` PASS\n- `go test ./pkg/server -run '^$' -count=1` PASS\n- `go test -c ./pkg/server -o /tmp/pr586-server.test` PASS\n- `git diff --check origin/main...HEAD` PASS\n\nFocused `pkg/server` fork tests are blocked in this local workspace by the same pre-existing meta schema drift noted in the PR body (`llm_usage` missing `tenant_id` before creating `idx_llm_usage_tenant_created`).

qiffang (commented):
## Review — TiDB Cloud Native Fork

### 🔴 Blockers

**1. `deleteForkAfterLiveRejection` re-reads an already consumed request body**

In `cmd/drive9/cli/db.go`, `DeleteTenant` sends the request body to `/v1/tenant`, then on 409 passes the same `body io.Reader` to `deleteForkAfterLiveRejection` → `/v1/fork`. But the `io.Reader` is already fully consumed from the first request, so the fork delete receives an empty body with no credentials.

Fix: buffer the original request body bytes before the first POST, then create a new `bytes.Reader` for the fallback.

**2. MySQL error 1072 added to globally ignorable errors in `schemaspec.go`**

The commit adding error 1072 ("key column doesn't exist") to ignorable errors is a broad fix that could mask real schema issues in other contexts. A more targeted skip for this specific index/column pair would be safer.

### 🟡 Non-blockers

- `forkResponseStatus`/`forkResponseMessage` accept `provider` parameter but don't use it — either remove or differentiate
- Consider squashing the 33 commits before merge for cleaner history

### ✅ Positives

- `backgroundWithTrace(ctx)` for branch cleanup — client disconnect won't cancel cloud cleanup
- API key issued before provisioning ensures recovery path always has credentials
- Clean credential resolution: request → default → error
- `decodeForkRequest` trailing data check without `DisallowUnknownFields` maintains backward compat
- `cleanupNativeFork` correctly skips cred check when `BranchID` is empty
- `drive9 delete` → 409 → `/v1/fork` fallback is good UX

Verdict: **REVISE** — fix the consumed body bug, assess schema error suppression scope.

qiffang (changes requested):
adversary-1 review at `c51bd03`: **REVISE**

Blocking findings:

1. `cmd/drive9/cli/db.go:451` / `:474` / `:570` reuse the same `io.Reader` for the `/v1/tenant` delete attempt and the fallback `/v1/fork` delete. When TiDB Cloud Native credentials are supplied, the first HTTP request consumes the body, so the fallback fork-delete sends EOF/no credential JSON. Buffer the request bytes and build a fresh reader per request; add a regression where `/v1/fork` asserts it received the credential JSON.

2. `pkg/server/fork.go:473-492` drops request-scoped credentials on async fork provisioning. `provisionForkTenantAsyncWithCredentials` passes a closure that calls `provisionForkTenantOnceWithCredentials`, but `provisionForkTenantAsyncWithProvision` ignores that closure and always calls `s.provisionForkTenantOnce(ctx, forkID)`. If the branch later needs wait/provisioning and the request credentials differ from server defaults, a fork can fail despite valid user-supplied credentials. Call `provision()` at line 492 and add a regression proving the credentialed async path uses the request keys.

3. `pkg/server/fork.go:356-386` can still persist a failed TiDB Cloud Native fork before any owner API key is inserted. The real native provisioner can return `cluster != nil, err != nil` after a branch is created (`pkg/tenant/tidbcloudnative/provisioner.go:338-340`). If branch cleanup also fails, `deleteForkBranchOrPersist` stores branch metadata and `markForkFailed` leaves a failed fork, but create returns a plain error and no fork API key for cleanup. The later `forkProvisionFailedError` path only covers post-key missing-endpoint cases. Ensure an active fork API key is returned whenever branch metadata is persisted for recovery, or insert the key before branch-backed failures can be persisted; add a regression for `CreateBranchWithCredentials` returning `cluster+err` and branch delete failing.

Validation: `git diff --check origin/main...HEAD` passed; GitHub `ci`, `drive9-server-local e2e smoke`, and CodeRabbit are green. I could not run Go tests locally because this agent environment has no `go` binary.

qiffang (changes requested):
R5 review on head `9b5bccf`: the three prior blockers are fixed, but I still found one user-facing recovery bug that should be fixed before merge.

Blocking:

1. `cmd/drive9/cli/ctx.go:606` prints non-existent environment variables for failed native fork cleanup. The actual constants are `DRIVE9_PUBLIC_KEY` / `DRIVE9_PRIVATE_KEY` (`EnvTiDBCloudPublicKey` / `EnvTiDBCloudPrivateKey`), but the failed-fork recovery guidance tells users to set `DRIVE9_TIDBCLOUD_PUBLIC_KEY` / `DRIVE9_TIDBCLOUD_PRIVATE_KEY`. This is the exact path users will follow when branch provisioning fails and they need to clean up a failed native fork; following the printed command will not provide credentials to `drive9 delete`. Please print the constants instead of hard-coded names and extend `TestCtxForkAcceptsFailedResponseAndSavesAPIKey` to assert the recovery guidance uses the real env names.

Verified fixed from the previous review:
- Delete no longer reuses a consumed request body; it buffers `bodyBytes` and creates a fresh reader for `/v1/fork`.
- The async fork provisioning loop now calls the injected `provision()` closure, so credentialed provisioning is no longer silently dropped by the helper.
- Fork API key issuance happens before branch creation, and branch-backed failures return a recovery API key / tenant ID.

Local validation:
- `/usr/local/go/bin/go test ./cmd/drive9/cli -run 'TestDeleteTenant' -count=1`
- `/usr/local/go/bin/go test ./cmd/drive9 -run 'Test.*(Usage|Help|Dispatch|Ctx)' -count=1`
- `/usr/local/go/bin/go test ./pkg/tenant/tidbcloudnative -run 'Test(CreateBranchWithCredentials|DeleteBranchWithCredentials)' -count=1`

Could not run `pkg/server`, `pkg/client`, or full `pkg/tenant` locally because this machine lacks rootless Docker for testcontainers; GitHub CI/e2e are green.

qiffang (approved):
R6 re-review on head `c6e1503`: GREEN.

The last blocker is fixed: failed native fork guidance now prints the actual env constants via `EnvTiDBCloudPublicKey` / `EnvTiDBCloudPrivateKey`, and `TestCtxForkAcceptsFailedResponseAndSavesAPIKey` asserts `DRIVE9_PUBLIC_KEY` / `DRIVE9_PRIVATE_KEY` are present. No remaining blockers found in the delta from `9b5bccf`.

Re-verified the prior fixes are still intact:
- fork delete credential body is buffered and reread for `/v1/fork`
- async credential provisioning calls the injected closure
- branch-backed create failures return recovery API key / tenant ID

Local validation:
- `/usr/local/go/bin/go test ./cmd/drive9/cli -run 'TestCtxForkAcceptsFailedResponseAndSavesAPIKey|TestCtxFork|TestDeleteTenant' -count=1`
- `/usr/local/go/bin/go test ./pkg/tenant/tidbcloudnative -run 'Test(CreateBranchWithCredentials|DeleteBranchWithCredentials|BranchWithCredentials|WaitForBranchActive)' -count=1`

GitHub `ci`, local e2e smoke, and CodeRabbit are green. Server/client/full tenant packages remain locally blocked by missing rootless Docker on this machine, but CI covers them.

PR #589, opened 2026-06-19
fix: gate native syncfs support on observed dispatch

mornyx:
## Summary
- Add a FUSE smoke regression for open writable handles around native `sync -f`.
- Stop reporting native syncfs support from protocol minor alone; only report it after the daemon observes an actual FUSE `SYNCFS` request.
- Keep explicit `drive9 mount drain` coverage as the reliable open-handle durability path when the kernel does not dispatch `syncfs(2)` to userspace.

## Root Cause
On the Linux host used for validation, `sync -f <mountpoint>` calls `syncfs(2)` and returns success, but the FUSE debug log does not show a `SYNCFS` opcode reaching the daemon. Even an experiment with the go-fuse fork negotiating protocol 7.45 still did not produce a `SYNCFS` request. Therefore protocol minor >= 7.34 is not sufficient evidence that native `sync -f` will drain Drive9 open handles.

## Test Plan
- [x] RED: with only the e2e regression added, `e2e/fuse-smoke-test.sh` failed at `native sync -f flushes open handle to remote` while `strace` showed `syncfs(3) = 0`.
- [x] `git diff --check origin/main...HEAD`
- [x] `go test ./pkg/fuse -run 'Test.*Drain|TestFlushDebouncer_FlushAll' -count=1`
- [x] `go test ./pkg/fuse -count=1`
- [x] `go test ./cmd/drive9 ./pkg/mountcontrol -count=1`
- [x] `DRIVE9_BASE=https://api.drive9.ai CLI_SOURCE=build FUSE_STRICT_PREREQS=1 RUN_FUSE_GIT_CLONE=0 RUN_FUSE_UMOUNT_DURABLE=0 RUN_FUSE_LOG_AUDIT=0 bash e2e/fuse-smoke-test.sh` exercised the new drain/syncfs section successfully:
  - `drive9 mount drain flushes open handle to remote` passed.
  - `native sync -f returns success without observed FUSE_SYNCFS dispatch` passed.
  - `native sync -f open-handle flush` was skipped because no FUSE_SYNCFS dispatch was observed.

Note: the full FUSE smoke run still exits non-zero due to an existing unrelated check: `explicit path truncate rename source missing locally (expected failure)`. The new syncfs/drain regression section itself passes after this fix.


srstack (commented):
## Review: fix: gate native syncfs support on observed dispatch (PR #589)

**Overall: correct and well-reasoned fix for a false capability promise. Clean. LGTM.**

### What it does
Previously `NativeSyncFSSupported` was derived purely from the negotiated FUSE protocol minor (`>= 7.34`). But some Linux FUSE mounts negotiate that version yet never dispatch `syncfs(2)` to userspace — so the control-plane drain response would falsely promise that `sync -f` flushes open Drive9 handles. This gates the capability on an **observed** real `SyncFs` dispatch (`nativeSyncFSSeen` atomic, set in `SyncFs`).

### Correctness — looks right
- `nativeSyncFSSeen atomic.Bool` set at the top of `SyncFs` (before the work), read via `nativeSyncFSObserved()`. Atomic is the right primitive for a flag flipped from the FUSE dispatch goroutine and read from the drain path. ✅
- The protocol major/minor are still reported (informational), but `NativeSyncFSSupported` now reflects observed reality — strictly more conservative, so it can only *withhold* a promise, never over-promise. Removing the `fuseSyncFSProtocolMinor` const is consistent. ✅
- Nil-guard `fs != nil` in `nativeSyncFSObserved` is defensive but harmless.

### 🟡 Minor
- **Capability is observe-once and latches true forever.** That's correct for "this mount can do native syncfs," but note it's inherently first-call-racy: the very first `sync -f` before any SyncFs dispatch will report unsupported, then flip. The e2e test covers the supported path with an open-handle writer; consider noting in the drain doc that the flag is observational (may report false until first syncfs).
- The e2e additions (open-handle writer, remote-visibility assertions) meaningfully strengthen coverage — good, since this behavior is hard to unit-test.

### Verdict
Correct conservative fix; ships a real capability signal instead of a protocol guess. LGTM.

srstack (approved):

PR #594, opened 2026-06-22
mount: expose cache/queue tuning and add flush perf timing

qiffang:
## Summary

- Expose `--dir-cache-max-entries` CLI flag (maps to existing `MountOptions.DirCacheMaxEntries`, default 100000). This is the key parameter for verifying/mitigating the dirty 107k directory cap cliff where write-back QPS drops from ~80 to 15.
- Add new `--commit-queue-max-pending` CLI flag and `MountOptions.CommitQueueMaxPending` field (default 100). A/B testing showed 400/1000 eliminates enqueue errors but does not improve fresh QPS — diagnostic/tuning knob only, not a performance fix.
- Add write-back flush sub-phase perf timing (`flush_stage_shadow` and `flush_snapshot_wb`) to the existing perf counter system. Includes count, avg, and max for both `stageShadowForQueuedCommitLocked` and `snapshotWriteBackLocked`. No behavior change; only recorded when `-perf-counters` is enabled.
- **NEW:** Add `snapshotWriteBackLocked` sub-phase timing — breaks down the 136ms flush bottleneck into `bytes_view`, `lock_wait`, `dat_write`, `meta_write` (avg/max each).
- **FIX:** `DirCache.Put()` completeExpires TTL bug — was using `negativeTTL` (1s) instead of `ttl` (10s). This caused complete listings to expire every second, triggering repeated full directory listings (18-38x observed in dirty 107k benchmarks). Now uses `ttl`, consistent with `MarkSessionCreatedDir` which already used `ttl`.

Both CLI flags reject `<= 0` (consistent with `--upload-concurrency` validation).

## Test plan

- [x] `TestMountCmdPassesDirCacheMaxEntriesOption` — CLI flag passes through to bridge
- [x] `TestMountCmdPassesCommitQueueMaxPendingOption` — CLI flag passes through to bridge
- [x] `TestMountCmdRejectsInvalidDirCacheMaxEntries` — 0 and negative values rejected
- [x] `TestMountCmdRejectsInvalidCommitQueueMaxPending` — 0 and negative values rejected
- [x] `TestDirCacheMaxEntriesOrDefault` — zero/negative uses default, positive value honored
- [x] `TestCommitQueueMaxPendingResolution` — verifies NewCommitQueue receives correct maxPending (0→100, 500→500)
- [x] `TestFlushPerfTimingCounters` — counters record correctly including sub-phase timing, nil perf doesn't panic
- [x] `TestNamespaceCache_CompleteMissExpiresWithTTL` — complete miss uses ttl, not negativeTTL; expires after ttl
- [x] `TestNamespaceCache_NegativeEntriesExpireAtNegativeTTL` — individual negatives still use negativeTTL
- [x] `TestNamespaceCache_CanAnswerMisses` — updated to use ttl-based expiry for complete marker
- [x] Full `go test ./pkg/fuse/` passes locally
- [ ] CI green
- [ ] A/B verification on 81.70.84.180: dirty 107k + cap=200k should show list count drop from 18-38 to 1-2

## Summary by CodeRabbit

* **New Features**
  * Added `--dir-cache-max-entries` and `--commit-queue-max-pending` CLI flags to `drive9 mount` for tuning FUSE caching and commit-queue backpressure.
* **Improvements**
  * Expanded flush and snapshot write-back performance metrics with more detailed timing breakdowns.
  * Updated directory cache “complete listing” expiration behavior.
  * Improved write-back cache concurrency to keep on-disk state consistent and reduce contention.
* **Tests**
  * Added coverage for new mount flags, performance counter reporting, and write-back concurrency edge cases.

### DIFF
--- a/pkg/fuse/writeback.go
+++ b/pkg/fuse/writeback.go
@@ -183,17 +183,38 @@ func (c *WriteBackCache) PutWithBaseRev(remotePath string, data []byte, size int
 	return c.PutWithBaseRevAndMode(remotePath, data, size, kind, baseRev, 0, false)
 }
 
+// WriteBackPutTimings captures sub-phase durations inside PutWithBaseRevAndMode
+// for diagnostic instrumentation. All fields are zero when not instrumented.
+type WriteBackPutTimings struct {
+	LockWait  time.Duration // time spent waiting for WriteBackCache.mu
+	DatWrite  time.Duration // atomicWrite of the .dat file (data + fsync + rename)
+	MetaWrite time.Duration // atomicWrite of the .meta file (meta + fsync + rename)
+}
+
 // PutWithBaseRevAndMode is like PutWithBaseRev, but also persists the file
 // permission bits that should be applied once the pending data is remote.
 func (c *WriteBackCache) PutWithBaseRevAndMode(remotePath string, data []byte, size int64, kind PendingKind, baseRev int64, mode uint32, hasMode bool) error {
+	_, err := c.PutWithBaseRevAndModeTimings(remotePath, data, size, kind, baseRev, mode, hasMode)
+	return err
+}
+
+// PutWithBaseRevAndModeTimings is like PutWithBaseRevAndMode but also returns
+// sub-phase timings for diagnostic instrumentation.
+func (c *WriteBackCache) PutWithBaseRevAndModeTimings(remotePath string, data []byte, size int64, kind PendingKind, baseRev int64, mode uint32, hasMode bool) (WriteBackPutTimings, error) {
+	var t WriteBackPutTimings
+
+	lockStart := time.Now()
 	c.mu.Lock()
 	defer c.mu.Unlock()
+	t.LockWait = time.Since(lockStart)
 
 	// Write .dat via temp + fsync + rename (atomic).
 	datPath := c.datFile(remotePath)
+	datStart := time.Now()
 	if err := atomicWrite(datPath, data); err != nil {
-		return fmt.Errorf("writeback put data: %w", err)
+		return t, fmt.Errorf("writeback put data: %w", err)
 	}
+	t.DatWrite = time.Since(datStart)
 
 	// Write .meta via temp + rename.
 	meta := WriteBackMeta{
@@ -209,18 +230,21 @@ func (c *WriteBackCache) PutWithBaseRevAndMode(remotePath string, data []byte, s
 	}
 	metaBytes, err := json.Marshal(meta)
 	if err != nil {
-		return fmt.Errorf("writeback marshal meta: %w", err)
+		return t, fmt.Errorf("writeback marshal meta: %w", err)
 	}
 	metaPath := c.metaFile(remotePath)
+	metaStart := time.Now()
 	if err := atomicWrite(metaPath, metaBytes); err != nil {
 		// Best-effort cleanup of the .dat file if meta write fails.
 		_ = os.Remove(datPath)
-		return fmt.Errorf("writeback put meta: %w", err)
+		return t, fmt.Errorf("writeback put meta: %w", err)
 	}
+	t.MetaWrite = time.Since(metaStart)
+
 	cp := meta
 	c.metas[remotePath] = &cp
 	c.cacheDataLocked(remotePath, data, size)
-	return nil
+	return t, nil
 }
 
 // Get reads the cached data for remotePath. Returns nil, false if not cached.
--- a/pkg/fuse/writeback.go
+++ b/pkg/fuse/writeback.go
@@ -679,7 +679,29 @@ func (c *WriteBackCache) PendingSummary() (count int, bytes int64, firstPath str
 }
 
 // ListByPrefix returns metadata for pending entries under prefix.
+// It first waits for any in-flight per-path operations (Put, Remove, etc.)
+// under the prefix to complete, so callers in mutation paths (rmdir, rename)
+// see a consistent snapshot that includes entries currently being written.
 func (c *WriteBackCache) ListByPrefix(prefix string) []*WriteBackMeta {
+	// Collect paths with in-flight operations under prefix.
+	c.mu.Lock()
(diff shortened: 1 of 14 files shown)


pkg/fuse/writeback_uploader.go
```diff
@@ -337,8 +337,11 @@ func (u *WriteBackUploader) uploadOne(localPath string) {
 	release := u.acquirePath(localPath)
 	defer release()
 
-	// Read data and remember the generation so we can detect concurrent overwrites.
-	meta, ok := u.cache.GetMeta(localPath)
+	// Atomically read meta + data under one path lock so they are guaranteed
+	// to be from the same generation. Without this, a concurrent Put could
+	// replace the data between GetMeta and getView, causing the uploader to
+	// upload new data with old baseRev/generation.
+	meta, data, ok := u.cache.GetMetaAndView(localPath)
```
mornyx, line 356: This now requires cached payload data before checking `meta.Kind == PendingChmod`. That changes the semantics of chmod-only retries: once the data upload has already succeeded and only chmod remains, the retry should not depend on the `.dat` payload. If the payload is missing or unreadable, `GetMetaAndView` returns `ok=false` (and may prune the metadata), so both the background retry and the same pattern in `UploadSyncWithRevision` can silently drop the pending chmod and leave the remote file with the wrong mode. Please read metadata first, handle `PendingChmod` without loading data, and only require `GetMetaAndView` for data-upload kinds. I reproduced this with a focused local regression where a `PendingChmod` entry with a missing `.dat` made zero chmod calls and removed the pending metadata.

mornyx (changes requested):
I re-reviewed PR #594 at `8163bcf2a24f932b19cb4bbb9fe93e7b99849c63`. CI is green and the focused/local package tests pass, but I found one functional regression in the write-back chmod retry path that should be fixed before merge.

pkg/fuse/writeback_uploader.go
qiffang: Good catch — this is a real correctness issue. Fixed in `4e3eb41`.

**Root cause**: `GetMetaAndView` calls `getViewLocked` which tries to read the `.dat` file. For `PendingChmod` entries, the `.dat` is legitimately absent (data already uploaded). `getViewLocked` calls `pruneEntryLocked` when `.dat` doesn't exist → silently drops the pending chmod.

**Fix**: Both `uploadOne` and `UploadSyncWithRevision` now read meta first via `GetMeta`. If `PendingChmod`, handle chmod directly without loading data. Only call `GetMetaAndView` for data-upload kinds.

**Regression tests added**: 
- `TestWriteBackUploader_PendingChmodWithMissingDat` (background upload path)
- `TestWriteBackUploader_UploadSyncPendingChmodWithMissingDat` (fsync/rename path)

Both tests: create entry → `MarkChmodPending` → delete `.dat` + evict in-memory cache → submit → verify chmod called + entry cleaned up.

pkg/fuse/writeback.go
```diff
@@ -183,57 +228,102 @@ func (c *WriteBackCache) PutWithBaseRev(remotePath string, data []byte, size int
...
 	if err := atomicWrite(metaPath, metaBytes); err != nil {
 		// Best-effort cleanup of the .dat file if meta write fails.
 		_ = os.Remove(datPath)
-		return fmt.Errorf("writeback put meta: %w", err)
+		c.releasePathLock(remotePath, pl)
+		return t, fmt.Errorf("writeback put meta: %w", err)
 	}
+	t.MetaWrite = time.Since(metaStart)
+
+	// Phase 3: publish in-memory state under global lock (fast, no I/O).
```
mornyx, line 300: Moving the `.dat/.meta` writes outside `c.mu` leaves a same-path visibility gap for metadata-only readers. `PutWithBaseRevAndModeTimings` holds the per-path lock until this publish step, but `GetMeta` still only takes `c.mu`, so callers such as lookup/readdir metadata, unlink, and pending-new rename can observe old or missing metadata while the new same-path Put is in progress and then fall through to remote behavior. Before this PR, those readers were blocked by `c.mu` for the whole Put. Please make `GetMeta` (and any metadata-only listing helpers with the same assumption) serialize with the path lock, or otherwise publish a consistent in-progress entry before releasing the old state. I reproduced the gap locally by holding the cache path lock and observing that `GetMeta` returns immediately instead of waiting for the in-flight Put to finish.

mornyx (changes requested):
Re-reviewed the latest PR head (`7f69553c3667f08db17c9c52dd079c3a162d2997`). The previous PendingChmod blocker is fixed and covered by the new regression tests. I found one remaining functional race in the write-back metadata path that should be addressed before merge.

pkg/fuse/writeback.go
qiffang: Good catch — fixed in `74c58b1`.

`GetMeta` now acquires the per-path lock before reading metadata, matching the same pattern used by `Get`, `getView`, and `GetMetaAndView`. This ensures callers always see either the pre-Put or post-Put state, closing the visibility gap during Phase 2 disk I/O.

Since per-path locks are keyed by path, this only adds contention for same-path concurrent access (which is the correct serialization behavior). Different paths remain fully concurrent.

Listing helpers (`ListPendingPaths`, `PendingSummary`, `ListByPrefix`, `ListPending`) are not changed — they iterate all entries on cold paths (observability/recovery) and cannot practically acquire per-path locks for every entry. Their `c.mu`-only protection is sufficient for their use cases.

mornyx (approved):
Re-reviewed the latest head (`15facf6cf6c4943b1871da91e47eea98ef7e094c`). The previously raised PendingChmod and write-back metadata visibility blockers have both been addressed with targeted regression coverage. I also ran focused FUSE/CLI tests, the full `go test ./pkg/fuse -count=1`, and `git diff --check origin/main...HEAD`; all passed locally. Approving without waiting for the currently pending GitHub CI per reviewer direction.

PR #595, opened 2026-06-22
perf: reduce writeback flush latency

mornyx:
## Summary

This PR reduces writeback-mode `Flush` latency for dirty files mounted through Drive9 FUSE.

The previous writeback path staged dirty file contents into the local shadow store and persisted pending-index metadata to disk before returning from `Flush`. That made writeback behave closer to a durable local flush path, even though remote upload still happens asynchronously through the commit queue. For small-file-heavy agent workloads, this added avoidable local durability cost on the hot path.

This change adds an explicitly volatile pending-index path for writeback flushes when the commit queue is available:

- stage dirty file contents into the shadow store without forcing the durable pending-index metadata path
- keep pending metadata in memory so same-mount namespace visibility still works
- keep remote upload ownership in the existing close/release commit-queue path
- preserve the durable pending-index path as the fallback when the non-durable staging path cannot be used

### `pkg/fuse/dat9fs.go`

- Updates the writeback `Flush` path to prefer non-durable shadow staging when both the commit queue and shadow store are available.
- Sets `fh.WriteBackSeq` after successful non-durable staging so repeated flushes for the same dirty sequence can return quickly.
- Keeps existing durable staging behavior as the fallback path if non-durable staging fails or is not applicable.
- Applies the same strategy to both small dirty files and larger dirty files that can be staged into the shadow store.
- Leaves `fh.Dirty` uncleared during `Flush`, matching the existing safety model for additional writes before `Release`.

### `pkg/fuse/pending_index.go`

- Adds volatile pending-index APIs:
  - `PutVolatileWithBaseRevAndMode`
  - `PutVolatileShadowSpillWithMode`
- Extends `putInternal` with a `durable` flag.
- Keeps the existing durable APIs writing metadata through `atomicWrite`.
- Lets volatile entries update in-memory pending metadata without writing `.meta` files to disk.

## Semantics

This is intentionally a writeback-mode optimization. It does not make writeback equivalent to close-sync or write-sync.

Expected behavior after this change:

- Same-mount visibility remains available through in-memory pending metadata.
- Remote visibility is still provided by the existing commit-queue flow after close/release.
- Crash recovery before the queued commit reaches the remote is not guaranteed for volatile pending metadata.
- Callers that need stronger remote/durable guarantees should continue using close-sync or write-sync behavior.

## Validation

Unit tests:

```bash
env GOCACHE=/private/tmp/drive9-fix-writeback/.codex-gocache \
  GOMODCACHE=/Users/mornyx/Desktop/repos/github.com/mem9-ai/drive9/.codex-gomodcache \
  go test -count=1 ./pkg/fuse
```

Result:

```text
ok   github.com/mem9-ai/drive9/pkg/fuse  24.992s
```

Manual blackbox / e2e validation was also run against a production Drive9 endpoint with an optimized CLI build. The results showed strong improvements for writeback flush-heavy small-file workloads, and the main FUSE release gate passed. However, this PR is intentionally opened as a draft because there are still known follow-up issues listed below.

## Known follow-ups

This PR should not be treated as merge-ready without reviewer sign-off on the semantics and follow-up fixes:

- Native git clone remount-restore still showed status drift in the git-ops e2e path for `native` clone mode.
- Close/release cross-mount visibility needs more focused A/B validation because remote commit still begins at close/release.
- FUSE unmount can still take around 60 seconds in some blackbox workloads; logs show this issue existed before this PR, but it remains a product/test-experience problem.
- The current behavior intentionally trades local pending-index crash recovery on writeback `Flush` for lower latency; this should be documented clearly before marking the PR ready.

## Risk

Medium to high.

The code path is narrow, but it sits on the FUSE writeback hot path. The main risk is semantic confusion around what `Flush` means in writeback mode versus close-sync/write-sync modes. The implementation preserves fallbacks and same-mount visibility, but reviewers should pay special attention to cross-mount visibility, crash recovery expectations, and commit-queue drain behavior.

## Summary by CodeRabbit

* **Refactor**
  * Improved internal handling of data staging during flush operations with optimized metadata management paths for better control over durability guarantees.

### DIFF
--- a/pkg/fuse/dat9fs.go
+++ b/pkg/fuse/dat9fs.go
@@ -2738,11 +2738,23 @@ func (fs *Dat9FS) stageShadowLocked(fh *FileHandle, durable bool) error {
 	}
 	mode, hasMode := fs.modeForPendingHandle(fh)
 	if fh.ShadowSpill {
-		if _, err := fs.pendingIndex.PutShadowSpillWithMode(fh.Path, size, fs.pendingKindForHandle(fh), fh.BaseRev, mode, hasMode); err != nil {
+		var err error
+		if durable {
+			_, err = fs.pendingIndex.PutShadowSpillWithMode(fh.Path, size, fs.pendingKindForHandle(fh), fh.BaseRev, mode, hasMode)
+		} else {
+			_, err = fs.pendingIndex.PutVolatileShadowSpillWithMode(fh.Path, size, fs.pendingKindForHandle(fh), fh.BaseRev, mode, hasMode)
+		}
+		if err != nil {
 			log.Printf("pending index put failed for %s: %v", fh.Path, err)
 		}
 	} else {
-		if _, err := fs.pendingIndex.PutWithBaseRevAndMode(fh.Path, size, fs.pendingKindForHandle(fh), fh.BaseRev, mode, hasMode); err != nil {
+		var err error
+		if durable {
+			_, err = fs.pendingIndex.PutWithBaseRevAndMode(fh.Path, size, fs.pendingKindForHandle(fh), fh.BaseRev, mode, hasMode)
+		} else {
+			_, err = fs.pendingIndex.PutVolatileWithBaseRevAndMode(fh.Path, size, fs.pendingKindForHandle(fh), fh.BaseRev, mode, hasMode)
+		}
+		if err != nil {
 			log.Printf("pending index put failed for %s: %v", fh.Path, err)
 		}
 	}
@@ -10219,9 +10231,12 @@ func (fs *Dat9FS) Flush(cancel <-chan struct{}, input *gofuse.FlushIn) (status g
 
 	requiresRemoteSync := isSQLitePersistentJournalPath(fh.Path)
 
-	// Write-back path: small dirty files are persisted to local disk
-	// and return immediately. The actual HTTP upload happens in Release
-	// (async). This reduces Flush latency from ~100-300ms to ~1-5ms.
+	// Write-back path: dirty data is staged locally and returned to the
+	// caller before the remote upload completes. When the CommitQueue is
+	// available, this intentionally keeps pending metadata in memory and skips
+	// local fsync/writeback-cache durability. Same-mount namespace consistency
+	// is still preserved; crash recovery before the queued commit reaches the
+	// remote is a close-sync/write-sync concern.
 	//
 	// IMPORTANT: We do NOT ClearDirty here. The buffer stays dirty as a
 	// safety net — if the user writes more data between Flush and Release,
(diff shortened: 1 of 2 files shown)


qiffang:
Re-reviewed the current PR head (`d90d6a765aca27db3a29636076c3d1bd5e19fdf0`). I still see the same blocker; the patch currently visible on GitHub has not changed from the version I reviewed earlier.

[P1] Do not fall back to the writeback uploader after a volatile shadow stage.

The new Flush fast paths stage only shadow data with `stageShadowForQueuedCommitLocked(fh, false)` and set `WriteBackSeq` without calling `snapshotWriteBackLocked` (`pkg/fuse/dat9fs.go:10262-10272`, `pkg/fuse/dat9fs.go:10417-10431`). On Release, if `commitQueue.Enqueue` fails because the queue is stopped/full, the non-ShadowSpill path falls back to `fs.uploader.Submit(fh.Path)` (`pkg/fuse/dat9fs.go:11173-11181`). That uploader reads from the writeback cache and returns immediately when `GetMeta` / `getView` is missing (`pkg/fuse/writeback_uploader.go:341-354`). For the volatile path there is no cache entry, and Release has already cleared the dirty state / sequence before enqueueing (`pkg/fuse/dat9fs.go:11143-11146`), so `close` can return OK while no remote commit is scheduled. The same-mount shadow may keep the data visible until unmount, but the update is not persisted.

The ShadowSpill path already documents the correct invariant for this exact situation: on enqueue failure, do not use `uploader.Submit`; upload synchronously from shadow instead (`pkg/fuse/dat9fs.go:11043-11056`). Please give the regular volatile-shadow path the same shadow-based fallback, or preserve the dirty state and return an error, and add a regression test that forces `CommitQueue.Enqueue` to fail after a volatile Flush.

Validation:
- Confirmed `HEAD == origin/pr-595-review-latest == d90d6a765aca27db3a29636076c3d1bd5e19fdf0`
- `git diff --check origin/main...origin/pr-595-review-latest` PASS
- `go test ./pkg/fuse -run 'Test(CommitQueue|PendingIndex|.*WriteBack.*|.*Flush.*|.*Release.*)' -count=1` PASS
- `go test ./pkg/fuse -count=1` PASS

qiffang:
Second-pass review on the updated head (`d90d6a765aca27db3a29636076c3d1bd5e19fdf0`): the previous blocker still applies.

The new volatile-shadow Flush path stages shadow data with `stageShadowForQueuedCommitLocked(fh, false)` and sets `WriteBackSeq`, but it intentionally skips `snapshotWriteBackLocked` and therefore does not create a writeback-cache `.dat/.meta` snapshot. In `Release`, when `canUseCache` is true and `commitQueue.Enqueue(entry)` fails, the non-layer fallback still calls `fs.uploader.Submit(fh.Path)` (`pkg/fuse/dat9fs.go:11173-11181`). `WriteBackUploader.uploadOne` still reads from `u.cache.GetMeta(localPath)` and `u.cache.getView(localPath)` (`pkg/fuse/writeback_uploader.go:341-354`), so for this volatile-shadow path it can return without uploading anything. At that point `Release` has already cleared dirty state and `WriteBackSeq` (`pkg/fuse/dat9fs.go:11143-11146`), so close can report success while no remote commit is queued or uploaded.

The fix should make the volatile-shadow enqueue-failure path shadow-based, not writeback-cache-based. Either synchronously commit/upload from shadow on enqueue failure, or preserve dirty state and return an error. Please also add a regression test that forces `CommitQueue.Enqueue` to fail after a volatile Flush and verifies that close either reports failure or the data is actually durable remotely.

Validation run locally:
- `git diff --check origin/main...HEAD` PASS
- `go test ./pkg/fuse -run 'Test(CommitQueue|PendingIndex|.*WriteBack.*|.*Flush.*|.*Release.*)' -count=1` PASS
- `go test ./pkg/fuse -count=1` PASS

PR #596, opened 2026-06-22
feat: expand typescript sdk parity

mornyx:
## Summary

Expand the TypeScript SDK toward parity with the Go SDK and current Drive9 API surface.

This PR adds typed helpers and documentation for:

- filesystem metadata, batch stat, batch read-small, append, readAt, download, chmod, symlink, hardlink, recursive delete, and layer-aware grep/find flows
- multipart upload improvements, upload summaries, stream writer fixes, v1 resume support, and partial patch completion
- filesystem-scoped token issue/revoke helpers
- vault grants and owner/delegated read helpers
- SSE watch helpers with lifecycle callbacks
- LayerFS typed APIs for create/list/get/diff/replay/entry/object/checkpoint/events/rollback/commit
- git workspace APIs for workspace metadata, trees, state, object packs, and overlay entries
- journal create/append/read/search/verify helpers
- a TypeScript SDK integration guide and cookbook example
- JavaScript SDK CI matrix coverage on Node 18 and Node 22 with lint/test/build

## Notable fixes found during live prod validation

- Omit negative `expected_revision` values from multipart initiate bodies instead of sending `-1`.
- Complete v1 patch uploads after dirty part upload.
- Accept both `200` and `202` from v1 resume.
- Preserve presigned part headers for v1 S3 uploads.
- Slice v1 resume missing parts with the server-provided `part_size` instead of deriving it from total file size.
- Use the real LayerFS and git overlay op names (`upsert`, etc.) and narrow TypeScript types to catch invalid values.

## Validation

Local:

- `npm run lint`
- `npm test`
- `npm run build`

Live prod:

- Ran a real Drive9 prod smoke suite through `Client.defaultClient()`.
- Final result: `SUMMARY passed=15 failed=0`.
- Covered filesystem operations, multipart upload/read, patch upload, stream writer v2, v1 resume, scoped tokens, vault, SSE events, LayerFS, git workspace APIs, and journal APIs.

## Notes

The prod validation initially surfaced misleading `backend unavailable` responses for invalid LayerFS entry ops. The backend does support LayerFS; the client-side issue was using `op: "write"` instead of the supported `op: "upsert"`. This PR fixes the client examples/types, but the server may still benefit from returning a clearer 400 validation error for invalid LayerFS op values.

## Summary by CodeRabbit

* **New Features**
  * Added server-sent events (SSE) streaming for real-time event watching with automatic reconnection.
  * Added Git workspace management APIs for version control integration.
  * Added journal/audit log APIs with search and verification capabilities.
  * Added FS Layer APIs for advanced filesystem operations.
  * Added scoped token issuance and revocation for fine-grained access control.
  * Added vault grant management (issue/revoke) and owner-read secret helpers.
  * Expanded filesystem operations: batch stat/read, recursive delete, range reads, permission management, and streaming downloads.

* **Documentation**
  * Added comprehensive TypeScript SDK integration guide covering installation, usage patterns, and advanced features.
  * Updated README with expanded API documentation and testing instructions.
  * Added cookbook examples demonstrating SDK usage.

* **Tests**
  * Added coverage and parity test suites.


qiffang (commented):
## Review — TypeScript SDK Parity

### 🔴 Blocker

**`append()` is a non-atomic read-modify-write that is O(file_size)**

The current implementation does `stat()` → `read()` → concat → `write()` for every append call. For a 100MB file, this reads the entire file, allocates file+append in memory, and re-uploads the whole thing. This is extremely expensive for large files.

If the server supports a server-side append (e.g. `POST /v1/fs/path?append`), use it. If not, either:
1. Add a size guard (reject files > threshold), or
2. Document prominently that this is "small files only"

### 🟡 Non-blockers

1. **`normalizeWriteOptions` duplicated** in `client.ts` and `transfer.ts` — extract to shared utility
2. **`commitFSLayer` 409 handling** — if body parse succeeds, it throws; if parse fails, falls through to `decodeJSON` which will also fail (body consumed). Should always throw on 409.
3. **`appendStream` buffers entirely** — defeats purpose of stream interface; document or remove stream overload

### ✅ Positives

- Clean module separation (layer, git, journal, events, tokens)
- Cookbook coverage test ensures API surface completeness
- Real bug fixes: v1 resume 200+202, presigned headers, negative `expected_revision` omission
- Node 18 compat fix (crypto.subtle → node:crypto)
- SSE reconnect with exponential backoff + AbortSignal
- CI matrix on Node 18+22

clients/drive9-js/src/client.ts
```diff
@@ -1,42 +1,169 @@
...
+  const override = process.env.DRIVE9_CONFIG?.trim();
+  if (override) return override;
   const home = process.env.HOME || process.env.USERPROFILE || "";
-  const cfgPath = path.join(home, ".drive9", "config");
+  return path.join(home, ".drive9", "config");
+}
+
+function credentialsFromContext(cfg: Drive9Config, name: string): { server: string; apiKey: string } | undefined {
+  const entry = cfg.contexts?.[name];
+  if (!entry || (entry.type !== "owner" && entry.type !== "fs_scoped")) return undefined;
```
qiffang: 这里会拒绝 README 里刚文档化的配置格式：`contexts.default` 只有 `api_key`、没有 `type`。旧 TS/Python/Swift/Kotlin 客户端都是读取 `current_context.api_key`，这里加了 `type === owner/fs_scoped` 后，`Client.defaultClient()` 会在用户的有效 CLI 配置下构造出无 `apiKey` 的 client，后续请求直接 401。要么兼容缺省 `type` 为 owner，要么同步修改文档和所有现有配置迁移；当前状态是实现和文档不一致。

clients/drive9-js/src/client.ts
```diff
@@ -1,42 +1,169 @@
...
-    let apiKey: string | undefined;
-    if (cfg.current_context && cfg.contexts) {
-      apiKey = cfg.contexts[cfg.current_context]?.api_key;
+    const cfg = JSON.parse(data) as Drive9Config;
+    if (cfg.current_context) {
+      const current = credentialsFromContext(cfg, cfg.current_context);
+      if (current) return current;
     }
-    return { server, apiKey };
+    const names = Object.keys(cfg.contexts || {}).sort();
```
qiffang: 不要在 `current_context` 不可用时按字母序扫描其它 context。当前测试甚至固定了这种行为：选中的 context 是 delegated，就静默拿 `usable` 的 `fs_scoped` key。这样会让调用方在错误的 tenant / scope 下发请求，权限边界不可预测。`defaultClient()` 应该只尊重 `current_context`，不可用就无 key/报错；不能自动切到另一个 context。

clients/drive9-js/src/client.ts
```diff
@@ -47,19 +174,49 @@ function loadConfigFile(): { server: string; apiKey?: string } | undefined {
...
+  if (typeof options === "number") {
+    return { expectedRevision: options };
+  }
+  return { expectedRevision: options?.expectedRevision ?? -1, tags: options?.tags, description: options?.description };
+}
+
+function setTagHeaders(headers: Headers, tags?: Record<string, string>): void {
+  if (!tags) return;
+  for (const [key, value] of Object.entries(tags)) {
+    headers.append("X-Dat9-Tag", `${key}=${value}`);
```
qiffang: `Headers.append()` 在 Fetch/undici 里会把重复 `X-Dat9-Tag` 合并成一个逗号分隔值（例如 `owner=alice, topic=note`）。服务端 `parseWriteTagsHeader()` 用 `Header.Values("X-Dat9-Tag")` 逐值解析，不会拆逗号；结果会只写入 `owner: "alice, topic=note"`，丢掉 `topic` tag。Go 客户端用 `Header.Add` 发送多个 header 值。TS 这里需要用能保留重复 header 的 `HeadersInit` 形态或改协议；当前测试只 `toContain`，没有验证服务端能解析成两个 tag。

clients/drive9-js/src/client.ts
```diff
@@ -110,6 +362,36 @@ export class Client {
...
+      expectedRevision = stat.revision;
+    } catch (err) {
+      if (!(err instanceof Error) || !err.message.includes("not found")) {
+        throw err;
+      }
+    }
+    const merged = new Uint8Array(existing.length + data.length);
+    merged.set(existing);
+    merged.set(data, existing.length);
+    await this.write(path, merged, { ...options, expectedRevision: options?.expectedRevision ?? expectedRevision });
```
qiffang: `append`/`appendStream` 把旧内容和追加内容合并后直接走 `write()`。一旦合并后的大小达到服务端 inline threshold，`handleWrite` 会把它当 large PUT initiate，并要求 `X-Dat9-Part-Checksums`，这里没有提供，实际会 400 `missing X-Dat9-Part-Checksums header`。文档说 append 可用于 small/moderate log 写入，但日志持续增长后会稳定失败。需要在 merged size 超过阈值时走 multipart/writeStream 路径并保留 expected revision CAS。

clients/drive9-js/src/transfer.ts
```diff
@@ -168,8 +163,20 @@ export async function writeStreamImpl(
...
+export async function writeStreamWithSummaryImpl(
+  client: Client,
+  path: string,
+  stream: ReadableStream<Uint8Array> | Uint8Array,
+  size: number,
+  options?: number | WriteOptions
+): Promise<UploadSummary> {
+  const opts = normalizeWriteOptions(options);
+  const started = new Date();
   const threshold = client["smallFileThreshold"];
```
qiffang, line 180: 这里仍然用本地默认 50KB 阈值决定 direct PUT，而不是服务端协商阈值的 fail-safe 逻辑。Go SDK 已经明确改成：没有 `/v1/status` 缓存时阈值为 0，非空上传强制走 multipart，因为服务端 inline threshold 可能低于 50KB，否则 direct PUT 会被服务端拒绝为 `missing X-Dat9-Part-Checksums`。TS SDK 现在 `Client.defaultClient().writeStream(...)` 未 warm 时仍可能走错协议，不符合 parity。

clients/drive9-js/src/layer.ts
```diff
@@ -0,0 +1,180 @@
...
+export async function commitFSLayer(client: Client, layerId: string): Promise<FSLayerCommit> {
+  const resp = await fetch(layerURL(client, layerId, "/commit"), {
+    method: "POST",
+    headers: client.authHeaders({ "Content-Type": "application/json" }),
+    body: "{}",
+  });
+  if (resp.status === 409) {
+    const body = (await resp.json().catch(() => undefined)) as FSLayerCommit | undefined;
+    if (body) {
+      throw new StatusError("fs layer commit conflict", 409);
```
qiffang: 409 时服务端返回的 `FSLayerCommit` body 包含 `status/conflicts`，Go SDK 会把该 body 返回给调用方（同时返回 409 error）。这里读到了 body 却丢弃，只抛一个没有 conflict details 的 `StatusError`，调用方无法知道哪个 path/revision 冲突，也不符合文档里“typed LayerFS calls”的可用性。需要保留冲突 body（例如专门的 conflict error 带 `commit` 字段，或其它可检查结构）。

qiffang (changes requested):
认真 review 了一轮。SDK lint/test/build 在本地通过，但当前实现还有几处协议正确性问题，主要集中在配置加载、Fetch header 语义、append/上传阈值、LayerFS 冲突返回。下面这些需要修掉后再合。

mornyx:
Addressed the review feedback in `8289b2e`:

- Config loading now treats legacy contexts without `type` as owner-compatible, and only uses `current_context` instead of silently falling back to another usable context.
- `X-Dat9-Tag` values are now sent as repeated header tuples, avoiding comma-merged `Headers.append()` values that the server does not parse as separate tags.
- `append()` / `appendStream()` now use the native `POST /v1/fs/<path>?append` flow for existing files. Missing paths are created directly. The small non-S3 fallback is bounded, and large read-modify-write appends are rejected instead of reading and re-uploading the whole file. I also added append plan coverage validation before upload completion.
- Cold non-empty `writeStream()` calls now use multipart before `/v1/status` warmup, instead of assuming the local 50 KB threshold.
- LayerFS commit `409` responses now preserve the parsed conflict payload via `FSLayerCommitConflictError`.
- The TypeScript SDK docs now describe native append and the bounded fallback behavior.
- Bumped the TypeScript SDK package version to `0.1.1`; no npm publish was performed.

Validation run:

- `npm run lint`
- `npm test`
- `npm run build`
- Node 18 `vitest` + `tsc --noEmit`
- Full prod live script: `SUMMARY passed=15 failed=0`

qiffang:
I found one correctness regression that should be fixed before merge.

- [P1] Preserve direct PUT for cold small `writeStream` calls. `Client` still initializes `smallFileThreshold` to `DEFAULT_SMALL_FILE_THRESHOLD` (50 KiB), and the old implementation used that value immediately, so small writes went through `PUT /v1/fs/...` even before `warm()`/`status()`. This PR changes `writeStreamWithSummaryImpl` to set the threshold to `0` until `statusCache` exists, so any non-empty write on a fresh client now goes through `/v2/uploads/initiate` and then the V1 multipart fallback. That makes tiny writes depend on multipart/S3 even though the server's normal small-file `PUT` path would work. On deployments without S3/multipart, the server returns internal storage errors for both V2 and V1 initiate when `b.s3 == nil`, so a cold `client.writeStream('/x', new Uint8Array([1]), 1)` can fail where it used to succeed. I think the fix is to keep using `client['smallFileThreshold']` as the cold default and let `status()` update it when available; the new summary can still report `direct_put` for that path. The new test that expects a 4-byte cold write to be `multipart_v2` should be inverted to protect the old/direct behavior.

Validation I ran locally on PR head `b97c5a92e243f3cc707196395f14a26762d9e536`:
- `(cd clients/drive9-js && npm install)`
- `(cd clients/drive9-js && npm run lint)`
- `(cd clients/drive9-js && npm test)` — 36/36 passing
- `(cd clients/drive9-js && npm run build)`

Aside from the cold small-write routing issue above, the added endpoint shapes I checked against the Go server routes look consistent.

mornyx:
Fixed in `a3b01f4`.

I reverted the cold-path threshold behavior so `writeStreamWithSummaryImpl` uses `client["smallFileThreshold"]` even before `/v1/status` is cached. That preserves the previous direct `PUT /v1/fs/...` behavior for tiny cold writes and avoids making small writes depend on multipart/S3 on deployments where the normal small-file PUT path works.

I also inverted the regression test: the cold 4-byte `writeStreamWithSummary()` case now asserts `direct_put`, verifies the body is sent via `PUT`, and fails if `/v2/uploads/initiate` is called.

Validation run:

- `git diff --check`
- `npm run lint`
- `npm test`
- `npm run build`
- Node 18 `vitest`
- Node 18 `tsc --noEmit`

qiffang:
Re-reviewed the updated head `a3b01f49e8572875369a694670b2d4b8867cb1be`.

My previous blocker is fixed: `writeStreamWithSummaryImpl` now uses the cached/default `smallFileThreshold` on cold clients, and the new regression test verifies a cold 4-byte upload stays on direct `PUT` instead of multipart. That preserves the old small-file behavior and avoids requiring S3/multipart for cold tiny writes.

Validation on the updated head:
- `(cd clients/drive9-js && npm run lint)`
- `(cd clients/drive9-js && npm test)` — 36/36 passing
- `(cd clients/drive9-js && npm run build)`
- GitHub checks are green: code-ci plus Python/JavaScript Node 18/22/Rust/Kotlin/Swift SDK jobs.

I do not see remaining merge-blocking issues from this pass.

qiffang (approved):

Issue #597, opened 2026-06-23
perf: async central quota mutation to unblock write hot path


qiffang:
## Problem

Server-side file write (small file create, 10KB) takes ~900ms, of which **~700ms (77%)** is spent in synchronous central quota mutation (`central_quota_ms`).

The current `applyLoggedQuotaMutation` in `pkg/backend/quota_mutation.go` does two synchronous DB operations to the central server DB per file:
1. `INSERT INTO quota_mutation_log` (~40ms)
2. `InTx { UpsertFileMetaTx + IncrStorageBytesTx + MarkMutationAppliedTx }` (~660ms due to row lock contention on `tenant_quota_usage`)

This limits durable write QPS to ~21 files/s with 16 upload concurrency, while the front-end write-back can enqueue at 430+ files/s.

## Root Cause Data

From server timing (2276 write samples, build c39fd0f):
- `server_write_total`: avg 899ms
- `central_quota_ms`: avg 696ms (p50=660, p95=1285) — **77% of total**
- `tenant_tx_ms`: avg 183ms
- `s3_put_ms`: 0ms (10KB uses inline DB)
- `body_read_ms`: 0.054ms

## Solution

Move central quota mutation out of the write hot path:

1. **Remove synchronous central quota from write path** — `applyLoggedQuotaMutation` becomes a non-blocking enqueue to an in-memory channel
2. **Background goroutine aggregates per-tenant deltas** — flushes to central DB every 1-2s with batched UPDATE
3. **Channel full = drop + metric warning** — never block writes for quota
4. **Retain reconciliation sweep** — periodic full recount from actual file metadata to correct drift

## Semantic Change

Quota becomes eventually consistent (soft limit). Short-term over-usage and occasional missed counts are acceptable per product decision. The existing quota check is already fail-open (`pkg/backend/quota.go:131-133`).

## Expected Impact

- `central_quota_ms` → ~0ms (removed from hot path)
- `server_write_total` → ~200ms (only tenant_tx remains)
- Durable write QPS: theoretical ~4x improvement

## Acceptance Criteria

- [ ] `central_quota_ms` absent from write timing or near 0
- [ ] `server_write_total_ms` < 300ms avg
- [ ] Drain completes within reasonable time after 30s write burst
- [ ] Dropped deltas observable via metrics/logs
- [ ] Quota reconciliation mechanism exists (this PR or documented follow-up)
- [ ] File write correctness unaffected
- [ ] Tests updated for new async semantics

PR #598, opened 2026-06-23
perf: move central quota mutation off write hot path

qiffang:
## Summary

Part of #597 (step 1: decouple hot path from central DB)

The write hot path (`syncCentralFileCreate`, `syncCentralFileOverwrite`) previously held `mutationMu` and performed a **synchronous INSERT** into `quota_mutation_log` (~700ms under row lock contention on `tenant_quota_usage`), blocking every file write. This was the root cause of the 263 QPS ceiling in benchmarks (77% of server write latency).

**Changes:**
- **`quota_mutation.go`**: File create/overwrite hot path now does a non-blocking channel send of a `mutationEntry` struct (zero DB operations). New `processOneMutation` performs log INSERT + apply in the background worker (two-phase: log auto-commit, then apply+mark in tx).
- **`quota_async.go`**: Channel type changed from `chan func()` to `chan mutationEntry`. Worker calls `processOneMutation` for each entry.
- **`dat9.go`**: Removed `mutationMu` (no longer needed). Updated channel type.
- **LLM cost records stay synchronous** — `syncCentralLLMCostRecord` calls `processOneMutation` directly (not the async queue). LLM costs are billable and must not be dropped. Not in the file write hot path.
- If the 256-slot channel is full, storage quota entries are **dropped** with a metric. Quota counters remain inaccurate until a manual `backfill-quota` CLI run. Acceptable per product decision (performance > precision for storage quota).

**Follow-up needed (separate PR):**
- Per-tenant delta batching to reduce row lock contention in background worker
- Automated periodic reconciliation (currently manual `backfill-quota` only)

## Test plan

- [x] `go build ./...` passes
- [x] `go vet ./pkg/backend/...` passes
- [x] Tests verify full log+apply+mark pipeline with fakeMetaQuotaStore
- [x] Test verifies LLM cost bypasses async queue (synchronous)
- [x] Test verifies non-blocking drop when queue is full
- [ ] CI tests pass
- [ ] Deploy to test server and re-run benchmark to verify QPS improvement
- [ ] Monitor `central_quota_mutation_dropped` metric under load

### DIFF
--- a/pkg/backend/dat9.go
+++ b/pkg/backend/dat9.go
@@ -102,15 +102,12 @@ type Dat9Backend struct {
 	qCache             *quotaCache    // nil when central quota is not wired
 
 	// mutationQueue decouples central quota mutations (syncCentralFileCreate,
-	// syncCentralFileOverwrite) from the fsync critical path. Mutations are
-	// enqueued here and drained by a background worker. The mutation log
-	// provides crash recovery via the existing MutationReplayWorker.
-	//
-	// mutationMu serializes logQuotaMutation + enqueueMutation so that
-	// durable log_id order and channel enqueue order cannot diverge under
-	// concurrent same-tenant writes.
-	mutationMu    sync.Mutex
-	mutationQueue chan func()
+	// syncCentralFileOverwrite) from the fsync critical path. The write hot
+	// path performs a non-blocking channel send of a mutationEntry struct;
+	// the background worker does both the durable log INSERT and the
+	// quota-state apply. If the channel is full the entry is dropped —
+	// backfill-quota reconciles from the source-of-truth tenant DB.
+	mutationQueue chan mutationEntry
 	mutationWG    sync.WaitGroup
 	mutationStop  context.CancelFunc
 
--- a/pkg/backend/quota_async.go
+++ b/pkg/backend/quota_async.go
@@ -6,26 +6,22 @@ import (
 
 const (
 	// mutationQueueSize is the buffer size for the async mutation channel.
-	// Sized to absorb short bursts without blocking the write path. A single
-	// worker drains the channel to preserve per-tenant FIFO ordering
-	// (UpsertFileMetaTx is not commutative across create/overwrite).
+	// Sized to absorb short bursts. The channel accepts mutationEntry structs
+	// (not closures) so the write hot path does zero DB work — the background
+	// worker performs both the durable log INSERT and the quota-state apply.
 	mutationQueueSize = 256
 )
 
 // startMutationWorker initializes the async mutation queue and a single
-// sequencing worker. A single worker guarantees that within this backend
-// instance, per-tenant mutation apply order matches the log insertion order
-// (which is serialized by the caller's mutationMu + logQuotaMutation).
-// Cross-instance ordering is not guaranteed; see logAndEnqueueMutation.
-//
-// Called once from SetMetaQuotaStore when server quota is active.
+// sequencing worker. Called once from SetMetaQuotaStore when server quota is
+// active.
 func (b *Dat9Backend) startMutationWorker() {
 	if b.mutationQueue != nil {
 		return // already started
 	}
 	ctx, cancel := context.WithCancel(context.Background())
 	b.mutationStop = cancel
-	b.mutationQueue = make(chan func(), mutationQueueSize)
+	b.mutationQueue = make(chan mutationEntry, mutationQueueSize)
 	b.mutationWG.Add(1)
 	go b.drainMutations(ctx)
 }
@@ -38,33 +34,16 @@ func (b *Dat9Backend) drainMutations(ctx context.Context) {
 			// Drain remaining items before exiting.
 			for {
 				select {
-				case fn := <-b.mutationQueue:
-					fn()
+				case entry := <-b.mutationQueue:
+					b.processOneMutation(context.Background(), entry)
 				default:
 					return
 				}
 			}
-		case fn := <-b.mutationQueue:
-			fn()
+		case entry := <-b.mutationQueue:
+			b.processOneMutation(context.Background(), entry)
 		}
 	}
-}
-
-// enqueueMutation submits a mutation apply function to the async queue.
-// The mutation log entry has already been durably written by the caller
-// (logQuotaMutation), so if this enqueue blocks or the process crashes
-// before the worker runs, MutationReplayWorker recovers the pending entry.
-//
-// If the queue is not wired (tests), the function runs inline.
-// The channel send blocks if the buffer is full, preserving FIFO ordering;
-// the 256-slot buffer makes blocking extremely unlikely in practice.
-func (b *Dat9Backend) enqueueMutation(fn func()) {
-	if b.mutationQueue == nil {
-		// No async queue — run inline (test/fallback path).
-		fn()
-		return
-	}
-	b.mutationQueue <- fn
 }
 
 // stopMutationWorker shuts down the async mutation queue and waits for
--- a/pkg/backend/quota_async_test.go
+++ b/pkg/backend/quota_async_test.go
@@ -1,6 +1,7 @@
 package backend
 
 import (
+	"context"
 	"database/sql"
 	"sync/atomic"
 	"testing"
@@ -96,4 +97,105 @@ func TestStopMutationWorker_Idempotent(t *testing.T) {
 	b.startMutationWorker()
 	b.stopMutationWorker()
 	b.stopMutationWorker() // Should not panic.
+}
+
+// TestProcessOneMutation_LogApplyMark verifies that processOneMutation
+// calls InsertMutationLog, the apply function, and MarkMutationAppliedTx
+// using a fakeMetaQuotaStore (defined in quota_migration_test.go).
+func TestProcessOneMutation_LogApplyMark(t *testing.T) {
+	fake := newFakeMetaQuotaStore()
+	b := &Dat9Backend{
+		tenantID:  "tenant-test",
+		metaStore: fake,
+	}
+
+	var applyCalled atomic.Bool
+	entry := mutationEntry{
+		mutationType: "file_create",
+		payload: fileCreateMutationData{
+			FileID:    "file-1",
+			SizeBytes: 1024,
+			IsMedia:   false,
+		},
+		apply: func(tx *sql.Tx) error {
+			applyCalled.Store(true)
+			return b.metaStore.IncrStorageBytesTx(tx, b.tenantID, 1024)
+		},
+	}
+
+	b.processOneMutation(context.Background(), entry)
+
+	require.True(t, applyCalled.Load(), "apply function should be called")
+
+	// Verify mutation log was inserted and marked applied.
+	fake.mu.Lock()
+	defer fake.mu.Unlock()
+	require.Len(t, fake.mutations, 1, "one mutation log entry expected")
+	require.Equal(t, "file_create", fake.mutations[0].typ)
+	require.Equal(t, "applied", fake.mutationStatus(fake.mutations[0].id))
+
+	// Verify storage bytes were incremented.
+	usage := fake.usage["tenant-test"]
+	require.NotNil(t, usage)
+	require.Equal(t, int64(1024), usage.StorageBytes)
+}
+
+// TestProcessOneMutation_WorkerDrain verifies that the background worker
+// calls processOneMutation with a real metaStore for each enqueued entry.
+func TestProcessOneMutation_WorkerDrain(t *testing.T) {
+	fake := newFakeMetaQuotaStore()
+	b := &Dat9Backend{
+		tenantID:  "tenant-test",
+		metaStore: fake,
+	}
+	b.startMutationWorker()
+
+	const n = 5
+	for i := 0; i < n; i++ {
+		b.enqueueMutationEntry(context.Background(), mutationEntry{
+			mutationType: "file_create",
+			payload: fileCreateMutationData{
+				FileID:    "file-1",
+				SizeBytes: 100,
+			},
+			apply: func(tx *sql.Tx) error {
+				return b.metaStore.IncrStorageBytesTx(tx, b.tenantID, 100)
+			},
+		})
+	}
+
+	b.stopMutationWorker()
+
+	fake.mu.Lock()
+	defer fake.mu.Unlock()
+	require.Len(t, fake.mutations, n, "all mutations should be logged")
+	for _, entry := range fake.mutations {
+		require.Equal(t, "applied", fake.mutationStatus(entry.id))
+	}
+	require.Equal(t, int64(n*100), fake.usage["tenant-test"].StorageBytes)
+}
+
+// TestSyncCentralLLMCostRecord_Synchronous verifies that LLM cost records
+// are processed synchronously (not enqueued to the async channel).
+func TestSyncCentralLLMCostRecord_Synchronous(t *testing.T) {
+	fake := newFakeMetaQuotaStore()
+	b := &Dat9Backend{
+		tenantID:  "tenant-test",
+		metaStore: fake,
+	}
+	// Start worker with tiny queue — if LLM cost went through queue,
+	// it would be enqueued there.
+	b.mutationQueue = make(chan mutationEntry, 1)
+
+	b.syncCentralLLMCostRecord(context.Background(), "img_extract_text", "task-1", 500, 100, "tokens")
+
+	// Queue should be empty — LLM cost bypasses the queue.
+	require.Equal(t, 0, len(b.mutationQueue), "LLM cost should not use async queue")
+
+	// Verify it was logged and applied.
+	fake.mu.Lock()
+	defer fake.mu.Unlock()
+	require.Len(t, fake.mutations, 1)
+	require.Equal(t, "llm_cost_record", fake.mutations[0].typ)
+	require.Equal(t, "applied", fake.mutationStatus(fake.mutations[0].id))
 }
--- a/pkg/backend/quota_async_test.go
+++ b/pkg/backend/quota_async_test.go
@@ -193,11 +193,11 @@ func TestProcessOneMutation_WorkerDrain(t *testing.T) {
 func TestSyncCentralFileCreate_AsyncPath(t *testing.T) {
 	fake := newFakeMetaQuotaStore()
 	b := &Dat9Backend{
-		tenantID:    "tenant-test",
-		metaStore:   fake,
 		quotaSource: QuotaSourceServer,
 	}
-	b.startMutationWorker()
+	// Use the production wiring path to start the worker.
+	b.SetMetaQuotaStore("tenant-test", fake)
+	require.NotNil(t, b.mutationQueue, "SetMetaQuotaStore should start mutation worker")
 
 	b.syncCentralFileCreate(context.Background(), "file-abc", 2048, "image/png")
 
--- a/pkg/backend/quota_async_test.go
+++ b/pkg/backend/quota_async_test.go
@@ -1,7 +1,7 @@
 package backend
 
 import (
-	"sync"
+	"database/sql"
 	"sync/atomic"
 	"testing"
 	"time"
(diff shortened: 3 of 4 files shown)


qiffang:
## dat9-dev2 review — blocking

I cannot submit a formal "request changes" review from this local GitHub auth because GitHub treats it as the PR owner, so posting this as a blocking review comment.

The direction (remove central quota DB work from the foreground write path) matches the product decision, but the implementation does not yet satisfy issue #597 or prove the new semantics safely.

### Blockers

1. **This does not implement the #597 batching/reconciliation design.**
   - Issue #597 asks for a background goroutine that aggregates per-tenant deltas and flushes batched updates every 1-2s, plus a reconciliation sweep.
   - This PR still processes one mutation at a time: insert log, upsert file_meta, update tenant_quota_usage, mark applied. It only moves the same per-file quota mutation to a single background worker.
   - Given the measured central quota path was ~700ms per mutation, a single per-entry worker can only drain a tiny fraction of the observed 400+ files/s foreground rate. The 256-slot queue will fill quickly and then drop most entries. That may be acceptable for soft quota, but it is a materially different system than the issue describes and needs explicit semantics + validation.
   - `backfill-quota` is a manual CLI, not an automatic reconciliation mechanism. If queue-full drops are expected under load, there needs to be either real periodic reconciliation or clear documented follow-up/operational requirement before merging.

2. **The transaction boundary is incorrect/misleading.**
   - `processOneMutation` says "Single transaction: INSERT log + apply quota state + mark applied".
   - `insertMutationLogTx` accepts `*sql.Tx` but ignores it and calls `store.InsertMutationLog(context.Background(), ...)`, which uses its own connection/autocommit.
   - Either implement a real `InsertMutationLogTx` on `MetaQuotaStore`, or rename/remove the fake Tx helper and update comments to state the actual boundary: log insert autocommits separately; only apply + mark are in the outer tx. Also propagate the worker context instead of hardcoded `context.Background()` in the helper.

3. **The new tests do not prove the mutation pipeline works.**
   - `quota_async_test.go` uses a nil `metaStore`, so `processOneMutation` returns before calling `entry.apply`.
   - `TestEnqueueMutationEntry_Async` only checks `len(b.mutationQueue)==0`; it does not prove log insert, apply, mark-applied, error handling, or even the enqueue API path.
   - `TestStopMutationWorker_DrainsPending` similarly only proves the channel becomes nil; it does not prove pending entries were applied.
   - Please add a fake `MetaQuotaStore` test that verifies: enqueue -> background log insert -> apply callback -> mark applied; apply failure leaves the log pending/error path observable; queue-full drop does not block and is observable. The tests should match the new soft-quota semantics explicitly.

4. **Merge needs perf proof against the actual acceptance criteria.**
   - CI compile/pass is not enough for this PR. The acceptance criteria are `central_quota_ms` removed/near-zero, `server_write_total_ms < 300ms avg`, drain completes after a 30s burst, drops observable, and file correctness unaffected.
   - Please attach the test-server benchmark/timing result before merging. If this PR is intentionally only step 1 (foreground decoupling, not batching), say that in the PR and scope the acceptance criteria accordingly.

### Local validation

- `go test -c ./pkg/backend` ✅
- focused runtime `go test ./pkg/backend -run 'TestEnqueueMutationEntry|TestStopMutationWorker' -count=1` is blocked locally by the existing testcontainers/rootless Docker TestMain issue, so CI remains the runtime gate.

qiffang:
## adversary-1 review — REVISE

Blocking findings:

1. `pkg/backend/quota_mutation.go:73-81`, `pkg/backend/quota_mutation.go:218-239`, `pkg/backend/llm_usage.go:35-45`, `pkg/meta/quota.go:645-679` — the lossy quota queue now also drops `llm_cost_record` entries. The product decision relaxed storage quota precision, but this path records billable LLM usage and monthly cost. When server quota is active, `recordImageExtractUsage` / `recordAudioExtractUsage` skip the tenant DB ledger and rely on the central record; a full mutation queue now silently loses those costs except for a generic dropped metric. That undercounts monthly LLM cost and can bypass budget enforcement. Keep LLM cost recording on a reliable/durable path, or get an explicit product decision that LLM cost accounting is also best-effort; do not route it through the same drop-on-full storage-quota queue by default.

2. `pkg/backend/quota_async_test.go:24-47`, `pkg/backend/quota_async_test.go:76-91`, `pkg/backend/quota_migration_test.go:455-464`, `pkg/backend/quota_store.go:121-123` — the tests do not exercise the production async mutation path. `quota_async_test` uses nil `metaStore`, so `processOneMutation` returns before calling `apply`; the tests only prove the channel empties. `newCentralQuotaBackend` never sets `quotaSource=server` before `SetMetaQuotaStore`, so `startMutationWorker` is not called and central quota lifecycle tests still run the inline fallback path. This misses the actual behavior being changed: non-blocking enqueue, background log/apply, queue-full drop, and shutdown drain. Add tests with a fake `MetaQuotaStore` and `QuotaSourceServer` that prove queued file mutations are applied asynchronously, drops do not log/apply, and existing lifecycle tests wait/drain instead of accidentally relying on inline fallback.

3. `pkg/backend/quota_mutation.go:59-81`, `cmd/drive9-server/backfill.go:20-38` — the PR text/comments present `backfill-quota` as convergence for dropped entries, but the implementation shown here is a manual one-time CLI, not a periodic reconciliation sweep. With drop-on-full semantics, quota drift can persist indefinitely unless an operator runs the CLI. Either add a real scheduled reconciliation path / operational hook, or change the PR contract and docs to say dropped storage deltas remain inaccurate until manual backfill, with metrics/alerts sufficient for operators to act.

Validation: `git diff --check origin/main...pr-598-review` passed. I could not run local Go tests because this environment has no `go` binary. CI is green, but these are semantic/test coverage blockers.

qiffang:
Additional blocker after re-checking the changed call sites:

5. **`llm_cost_record` must not inherit the lossy storage-quota semantics.**
   - `syncCentralLLMCostRecord` now also calls `enqueueMutationEntry`.
   - When the queue is full, `enqueueMutationEntry` drops the entry before writing any mutation log.
   - The product decision to prefer performance over exactness was about storage quota accounting. Billable LLM usage / monthly cost budget is a different correctness boundary: `monthlyLLMCostExceededServer` reads the central monthly cost counter before running extraction, so dropped `llm_cost_record` entries can undercount cost and bypass the budget.

Please keep LLM cost recording on a durable/non-droppable path, or explicitly split the queue semantics so only storage quota deltas are lossy. This needs a regression test where a full storage-quota queue does not silently drop billable LLM usage.

qiffang:
## dat9-dev2 R2 review — still blocking

Reviewed latest head `00b19f46462f5f9730b107676c43566a73719981`.

### Fixed

- ✅ `llm_cost_record` no longer uses the lossy drop-on-full queue. `syncCentralLLMCostRecord` now calls `processOneMutation` synchronously, which is the right boundary for billable usage / monthly budget.
- ✅ The fake `insertMutationLogTx` helper is gone, and `processOneMutation` now states the actual two-phase behavior: autocommit log insert, then tx apply+mark.
- ✅ The detailed queue-full comment now admits dropped entries are not replayed and require manual `backfill-quota`.

### Still blocking

1. **PR still does not satisfy issue #597.**
   - #597 explicitly asks for per-tenant delta aggregation + periodic/batched flush + reconciliation sweep.
   - This PR still does one background DB mutation per file: `InsertMutationLog` + `InTx(UpsertFileMetaTx + IncrStorageBytesTx + MarkMutationAppliedTx)`.
   - That removes foreground blocking, but it does not remove the central `tenant_quota_usage` row-lock bottleneck from the mutation pipeline. Under the observed 400+ files/s foreground rate, a single per-entry worker can fall behind and start dropping.
   - If this PR is intended as step 1 only, do not `Closes #597`; update the PR description and issue acceptance scope, and create/link a follow-up for batch aggregation/reconciliation.

2. **Tests still do not prove the production async path.**
   - `quota_async_test.go` still uses nil `metaStore`; `processOneMutation` returns before `entry.apply`.
   - The async test writes directly to `b.mutationQueue` and only asserts `len(queue)==0`; it does not prove `enqueueMutationEntry`, log insert, apply callback, or mark-applied.
   - `TestStopMutationWorker_DrainsPending` only proves queue cleanup, not that pending entries were actually applied.
   - Need a fake `MetaQuotaStore` test with `QuotaSourceServer` semantics that verifies: enqueue -> background `InsertMutationLog` -> apply -> `MarkMutationAppliedTx`; apply failure leaves log pending / records pending metric path; queue-full drop does not log/apply and is observable.

3. **Docs/comments still overclaim convergence.**
   - `pkg/backend/quota_mutation.go` still says `MutationReplayWorker + backfill-quota CLI handle convergence` for queue-full drops, but dropped entries are never logged, so `MutationReplayWorker` cannot see them.
   - PR body still says `Closes #597`, lists `syncCentralLLMCostRecord` as part of the write hot path, and says `backfill-quota` reconciles dropped entries without saying it is manual.
   - Please make the contract consistent: dropped storage quota deltas remain inaccurate until an operator/manual job runs backfill; replay only covers logged pending mutations, not pre-log drops.

4. **Perf proof is still required before merge.**
   - The issue acceptance criteria are runtime metrics: `central_quota_ms` removed/near-zero, `server_write_total_ms < 300ms avg`, drain completes after a 30s burst, drops observable, file correctness unaffected.
   - CI green does not validate this change. Please deploy this head to the test server and attach the timing/drain/drop data.

### Local validation

- `go test -c ./pkg/backend` ✅
- `gh pr checks 598`: CodeRabbit pass; `ci` and `drive9-server-local e2e smoke` were still pending when checked.

This is closer, but not mergeable yet.

qiffang:
## adversary-1 R2 — REVISE on `00b19f4`

The latest commit fixes my previous LLM-cost blocker: `syncCentralLLMCostRecord` now calls `processOneMutation` synchronously and no longer goes through the lossy queue. The drop/backfill comments are also more honest now: dropped file quota mutations stay inaccurate until manual `backfill-quota`.

One blocker remains:

1. `pkg/backend/quota_async_test.go:24-47`, `pkg/backend/quota_async_test.go:76-91`, `pkg/backend/quota_migration_test.go:455-464`, `pkg/backend/quota_store.go:121-123` — the tests still do not exercise the production async mutation path. `quota_async_test` uses nil `metaStore`, so `processOneMutation` returns before calling `entry.apply`; those tests only prove the channel drains. `newCentralQuotaBackend` does not set `quotaSource = QuotaSourceServer` before `SetMetaQuotaStore`, so `startMutationWorker` is not called and the central quota lifecycle tests still use the inline fallback path. This PR changes the production behavior specifically to non-blocking enqueue + background log/apply/drop, so it needs direct tests for that path.

Required coverage:
- A fake `MetaQuotaStore` with `QuotaSourceServer` where `syncCentralFileCreate` enqueues and the background worker eventually calls `InsertMutationLog`, `apply`, and `MarkMutationAppliedTx`.
- A full-queue case proving dropped file quota entries do not log/apply and emit the dropped metric/log path.
- Existing quota lifecycle tests should either wait for async completion or explicitly drain/stop the mutation worker before asserting central quota state.

Note: I do not consider the current two-phase log/apply structure itself a blocker after `00b19f4`. The code now accurately documents that log insert is auto-commit and apply+mark are one SQL transaction. If `MarkMutationAppliedTx` fails in a real SQL transaction, `entry.apply` should roll back with that transaction; pending replay should not double-apply unless the transaction layer violates that contract. The missing piece is test coverage for the actual async path.

Validation: `git diff --check origin/main...origin/pr-598-latest` passed. I still cannot run Go tests locally because this environment has no `go` binary. CI for `00b19f4` was pending when reviewed.

qiffang:
## dat9-dev2 R3 review — code blockers resolved, merge gate remains

Reviewed latest head `f359b65259266c2922999567cebdeb207659025a`.

### Resolved

- ✅ PR scope is now honest: `Part of #597 (step 1)` instead of `Closes #597`, with follow-up called out for per-tenant batching and automated reconciliation.
- ✅ PR body now states storage quota drops remain inaccurate until manual `backfill-quota`.
- ✅ LLM cost records bypass the lossy queue and stay synchronous/durable.
- ✅ Added fake `MetaQuotaStore` tests for the real mutation pipeline:
  - `TestProcessOneMutation_LogApplyMark` verifies `InsertMutationLog -> apply -> MarkMutationAppliedTx` and storage counter update.
  - `TestProcessOneMutation_WorkerDrain` verifies the background worker drains enqueued entries through the full log/apply/mark path.
  - `TestSyncCentralLLMCostRecord_Synchronous` verifies LLM cost does not use the async queue.

The previous code/test blockers are addressed.

### Remaining merge gate

Do not merge solely on code review/CI. This PR is a performance change whose acceptance depends on runtime data. Before merge/deploy as the fix, attach the test-server result showing:

- `central_quota_ms` absent or near-zero from the foreground write timing
- `server_write_total_ms` improvement against the #597 target
- 30s write-burst drain behavior
- `central_quota_mutation_dropped` observability under load
- file write correctness unaffected

Local validation:

- `go test -c ./pkg/backend` ✅
- `gh pr checks 598`: CodeRabbit pass/skipped; `ci` and `drive9-server-local e2e smoke` still pending when checked.

Verdict: **code GREEN / perf+CI gate pending**.

qiffang:
## adversary-1 R3 — REVISE on `f359b65`

The PR scope/body is now much better: it no longer closes #597, it calls this step 1, and it lists per-tenant batching + automated reconciliation as follow-ups. The LLM cost path also remains synchronous/durable.

Still blocking:

1. `pkg/backend/quota_async_test.go:131-135`, `pkg/backend/quota_async_test.go:169-174`, `pkg/backend/quota_async_test.go:196-200` — the new tests deadlock. Each block holds `fake.mu` and then calls `fake.mutationStatus(...)`; that helper also locks `fake.mu` (`quota_migration_test.go:401-407`). `sync.Mutex` is not re-entrant, so these tests will hang instead of validating the pipeline. Read `status` directly while already holding the lock, or unlock before calling `mutationStatus`.

2. `pkg/backend/quota_async_test.go:145-167` — this still does not test the production async entry point. The test manually calls `startMutationWorker()` and manually enqueues a synthetic `mutationEntry`; it never sets `quotaSource = QuotaSourceServer`, never goes through `SetMetaQuotaStore`, and never calls `syncCentralFileCreate` / `syncCentralFileOverwrite`. A regression that accidentally bypasses the async queue from the real file mutation functions would not be caught. Add a test that constructs a central-quota backend (`QuotaSourceServer`), wires a fake `MetaQuotaStore` through `SetMetaQuotaStore`, calls `syncCentralFileCreate` (or a write path that triggers it), drains/stops the worker, and asserts log + apply + mark.

3. `pkg/backend/quota_async_test.go:50-65` — the queue-full test only proves the channel length stays at 2. It does not prove the dropped storage mutation was not logged/applied, and it does not verify the drop observability path. Since dropping is the core product tradeoff in this PR, the regression should use a fake store, fill the queue, drop one distinct entry, drain the queued entries, and assert exactly the queued entries were logged/applied while the dropped entry was not. If metrics/log capture is available in tests, assert the `dropped` operation too; otherwise at least prove the semantic no-log/no-apply behavior.

4. `pkg/backend/quota_mutation.go:59-61` is still misleading: it says queue-full dropped entries are handled by `MutationReplayWorker + backfill-quota`, but the default branch below correctly says dropped entries are never logged and replay cannot recover them. For dropped entries, only manual backfill can reconcile. Please remove `MutationReplayWorker` from the queue-full sentence.

Validation: `git diff --check origin/main...origin/pr-598-latest` passed. I still cannot run Go tests locally because this environment has no `go` binary. CI was pending when reviewed.

qiffang:
## dat9-dev2 R3 correction — retracting code GREEN

I re-checked after adversary-1's R3. My previous "code GREEN" was too early.

Confirmed blockers on `f359b65259266c2922999567cebdeb207659025a`:

1. **New tests can deadlock.**
   - `quota_async_test.go` locks `fake.mu` and then calls `fake.mutationStatus(...)`, which locks the same non-reentrant `sync.Mutex` again.
   - Affected spots: `TestProcessOneMutation_LogApplyMark`, `TestProcessOneMutation_WorkerDrain`, and `TestSyncCentralLLMCostRecord_Synchronous`.
   - Fix by not holding `fake.mu` while calling methods that lock internally, or add a snapshot/helper that reads all needed state under one lock.

2. **Production entry path is still not covered.**
   - The worker test manually calls `startMutationWorker()` and manually enqueues synthetic `mutationEntry` values.
   - It does not verify the actual production wiring: `quotaSource = QuotaSourceServer` + `SetMetaQuotaStore(...)` starts the worker, and `syncCentralFileCreate` / `syncCentralFileOverwrite` enqueue real entries.
   - Add a test that constructs a server-quota backend, calls `SetMetaQuotaStore`, invokes `syncCentralFileCreate` or overwrite, stops/drains the worker, then asserts log/apply/mark and counters.

3. **Queue-full drop test is still too weak.**
   - Current test only asserts queue length stays full.
   - It should enqueue/fill known entries, attempt a distinct dropped entry, then prove that distinct entry was not logged/applied and that the drop observability path is emitted as far as the existing test harness can verify.

4. **One stale comment still overclaims replay convergence.**
   - `enqueueMutationEntry` still says `MutationReplayWorker + backfill-quota CLI handle convergence` for queue-full drops.
   - Dropped entries are never logged, so `MutationReplayWorker` cannot recover them. Only manual backfill can reconcile them.

Scope and LLM cost fixes are good. Two-phase log/apply is acceptable. Remaining blocker is test correctness/coverage plus the stale comment.

Verdict: **REVISE**.

qiffang:
## adversary-1 R4 — REVISE on `2f76208`

Good progress: the deadlock in the new tests is fixed, the queue-full test now proves the dropped entry is not logged/applied, and the drop/replay comment is corrected.

One blocker remains:

1. `pkg/backend/quota_async_test.go:193-204` — `TestSyncCentralFileCreate_AsyncPath` still bypasses the production wiring. It manually constructs `Dat9Backend{tenantID, metaStore, quotaSource}` and calls `startMutationWorker()` directly. That proves `syncCentralFileCreate` works if a worker already exists, but it does not prove the production path `SetMetaQuotaStore` starts the worker when `quotaSource == QuotaSourceServer`. A regression in `SetMetaQuotaStore` wiring would still pass this test. Please wire the fake store through `SetMetaQuotaStore("tenant-test", fake)` after setting `quotaSource: QuotaSourceServer`, then call `syncCentralFileCreate`, stop/drain, and assert log/apply/mark. This is a small test change but it closes the exact production-entry gap from the previous review.

Non-blocking: CI and perf validation are still pending merge gates.

Validation: `git diff --check origin/main...origin/pr-598-latest` passed. I still cannot run Go tests locally because this environment has no `go` binary. CI was pending when reviewed.

qiffang:
## dat9-dev2 R4 review — code blockers resolved

Reviewed latest head `2f762089c28d14dccdcd3d38cd17d1fb0c94be52`.

The four R3 blockers are addressed:

- ✅ **Deadlock fixed**: tests no longer hold `fake.mu` while calling `mutationStatus()`.
- ✅ **Production entry covered sufficiently**: `TestSyncCentralFileCreate_AsyncPath` now sets server-quota mode, calls `syncCentralFileCreate`, and verifies the async worker path reaches `InsertMutationLog -> UpsertFileMetaTx/IncrStorageBytesTx/IncrMediaFileCountTx -> MarkMutationAppliedTx`.
- ✅ **Drop test strengthened**: `TestEnqueueMutationEntry_NonBlocking_DropOnFull` now verifies the dropped distinct entry is not logged and not applied.
- ✅ **Drop/replay comment fixed**: function-level comment now correctly says dropped entries are never logged and cannot be recovered by `MutationReplayWorker`; only manual `backfill-quota` reconciles counters.

Non-blocking test polish: the production-entry test still manually sets fields and calls `startMutationWorker()` instead of going through `SetMetaQuotaStore`. Given it now covers `syncCentralFileCreate -> enqueue -> worker -> log/apply/mark`, I don't consider this a blocker, but using `SetMetaQuotaStore` or the existing server-quota helper would make the test closer to runtime wiring.

Local validation:

- `go test -c ./pkg/backend` ✅
- `gh pr checks 598`: CodeRabbit pass; `ci` and `drive9-server-local e2e smoke` still pending when checked.

Verdict: **code GREEN / CI + perf gate pending**.

Merge should still wait for CI and the test-server performance evidence called out in the PR: foreground `central_quota_ms`, `server_write_total_ms`, 30s burst drain behavior, drop observability, and file correctness.

qiffang:
## adversary-1 R5 — GREEN on `853d07b` (conditional on CI + perf gate)

The final test-wiring blocker is fixed. `TestSyncCentralFileCreate_AsyncPath` now sets `quotaSource: QuotaSourceServer`, wires the fake store through `SetMetaQuotaStore("tenant-test", fake)`, asserts the mutation worker was started, then verifies `syncCentralFileCreate → enqueue → worker drain → InsertMutationLog → apply → MarkMutationAppliedTx`.

Previously blocking items are resolved:
- LLM cost records stay synchronous/durable and do not use the lossy queue.
- The two-phase log/apply behavior is accurately documented and acceptable.
- New tests no longer deadlock on `fake.mu` / `mutationStatus`.
- Queue-full drop test proves the dropped entry is not logged or applied.
- Drop/replay comments now correctly say dropped entries cannot be recovered by `MutationReplayWorker`; only manual `backfill-quota` can reconcile them.
- PR scope is honest: `Part of #597`, with batching/reconciliation listed as follow-ups.

Merge gates still required:
- CI green.
- Test-machine perf evidence that the write hot path improves and drop behavior is observable under load.

Validation: `git diff --check origin/main...origin/pr-598-latest` passed. I still cannot run Go tests locally because this environment has no `go` binary. CI was pending when reviewed.

qiffang:
R5 update on `853d07ba7aab123bb1e7cf2159544bb48c04c03a`:

The remaining production-wiring test blocker is resolved. `TestSyncCentralFileCreate_AsyncPath` now sets `quotaSource: QuotaSourceServer`, calls the production entry point `SetMetaQuotaStore("tenant-test", fake)`, asserts `mutationQueue` is initialized, then runs `syncCentralFileCreate -> stop/drain` and verifies `InsertMutationLog -> apply -> MarkMutationApplied` effects.

I rechecked the relevant code paths:
- `SetMetaQuotaStore` sets `tenantID`/`metaStore` and starts the mutation worker when `QuotaSourceServer` is active.
- The test no longer manually calls `startMutationWorker()`, so it covers the real production wiring that matters for hot-path decoupling.
- Local cheap gates pass: `go test -c ./pkg/backend` and `git diff --check origin/main...HEAD`.

Verdict from my side: code GREEN. Remaining merge gates are unchanged: GitHub CI must finish green, and the perf run still needs to show the foreground quota timing removed/improved plus acceptable drain/drop observability and file correctness.

qiffang:
Closing in favor of PR #603 (quota outbox), which is the more complete solution — tenant-local outbox ensures no mutations are dropped, unlike the channel-based approach here. Both reviewers GREEN on #603.

PR #600, opened 2026-06-23
feat: migrate app-mode image extraction to durable semantic_tasks

### LINKED ISSUE #599: Make drive9-server stateless and horizontally scalable across multiple pods
## Background

drive9-server is currently deployed as multiple pods on Tencent Cloud (no embedding support, so no persistent worker). The request path (read/write API + file locks/leases) is **already multi-pod safe** — file concurrency is arbitrated by TiDB row locks (`FOR UPDATE`), revision optimistic CAS, upload unique indexes, and database-backed leases. The FUSE client holds all stateful sessions (open file table, read offset, write buffer) in its own process, so the server is a stateless HTTP layer for reads/writes.

**What blocks lossless horizontal scaling is the background scheduler layer and in-process state**: 7 background workers all start on every pod with no leader election, the event bus is in-memory only, and the image-extract queue is non-durable. This issue tracks the work required to make `drive9-server` losslessly scalable to N pods.

---

## What is already multi-pod safe (no change needed)

| Mechanism | Location | Why safe |
|---|---|---|
| File content write CAS | `pkg/datastore/file_tx.go:47-147` | `UPDATE ... WHERE revision=?`, `RowsAffected==0` → `ErrRevisionConflict` |
| Upload dedup | `pkg/datastore/store.go:2361-2369`, `tidb_auto.go:826-827` | DB unique indexes (`idx_idempotency`, `idx_uploads_active`) arbitrate across pods |
| Path tree mutations | `pkg/datastore/store.go` (`mkdir`/`rename`/`unlink`) | `SELECT ... FOR UPDATE` row locks across all pods |
| Semantic task / file_gc lease | `pkg/datastore/semantic_tasks.go:239-339`, `pkg/datastore/file_gc_tasks.go:115-310` | `FOR UPDATE SKIP LOCKED` + receipt/lease_until CAS = exactly-once claim across pods |
| FUSE client state | `pkg/fuse/dat9fs.go:31-127`, `pkg/fuse/handle.go:13-55` | open file table, read offset, write buffer all in FUSE client process; server is stateless HTTP |

---

### P0-1. Event bus is in-memory only (`/v1/events` SSE)

- `pkg/server/eventbus.go:23` header comment: `// Single-instance only — does not survive restarts or replicate across processes.`
- `EventBus` (`eventbus.go:24-32`): 10000-ring-buffer + `seq` counter + `listeners map[uint64]chan struct{}`, guarded by `sync.RWMutex`.
- `eventBuses` (`pkg/server/sse.go:111-114`): per-tenant `*EventBus` map.
- `pkg/fuse/mount.go:73` `TrustLocalEvents` comment: "safe only for single-server/sticky or cluster-wide event streams".

**Impact**: A write landing on pod A never propagates to SSE clients connected to pod B → FUSE clients miss cache-invalidation notifications → stale directory listings / file contents on non-writer pods.

**Proposed fix**: Replace the in-memory ring buffer with an external pubsub adapter (Redis Pub/Sub or NATS). `EventBus.Publish` writes to the external bus; each pod subscribes and fans out to its local SSE listeners. Alternatively, short-term: disable `TrustLocalEvents` and fall back to TTL expiry + read-back validation (correct but slower).

**Files**:
- `pkg/server/eventbus.go:23-32`
- `pkg/server/sse.go:109-148`
- `pkg/fuse/mount.go:73` (`TrustLocalEvents`)

### P0-2. Image extraction queue is non-durable in-process channel

- `pkg/backend/dat9.go:122` `imageExtractQueue chan ImageExtractTaskSpec` (buffer 128).
- `pkg/backend/image_extract.go:90-122` `enqueueImageExtract`: writes to in-process channel, drops on full (`:114-121`), never persists to DB.
- Contrast: `audio_extract_text` / `img_extract_text` tasks go through the durable `semantic_tasks` table (`pkg/semantic/task.go:26-37`).

**Impact**: Image→text extraction tasks are in-process only. Pod A's enqueued tasks are processed only by pod A; if pod A crashes or scales down, queued tasks are lost and never redone by other pods. Load is imbalanced across pods.

**Proposed fix**: Convert image extraction to a durable task row (same as `audio_extract_text`), enqueue via `EnqueueSemanticTask`, and reuse the existing semantic-worker claim/lease/recover pipeline. This makes it crash-recoverable and multi-pod-safe by construction.

**Files**:
- `pkg/backend/image_extract.go:90-122`
- `pkg/backend/dat9.go:119-127` (`imageExtractQueue` field + workers)
- `pkg/backend/options.go:272-279` (queue wiring)

### P0-3. Background schedulers have no leader election — every pod runs all workers

All 7 background workers start unconditionally on every pod with no leader election or cross-pod coordination:

| Worker | File | Interval | Has claim/lease? | Multi-pod risk |
|---|---|---|---|---|
| Object GC | `pkg/server/object_gc_worker.go:46` | 1min | ❌ no claim, only `PostponeObjectGCCandidate` | **Dangerous**: N pods pull the same due candidates, duplicate S3 HEAD/DELETE calls + meta DB write contention |
| Tenant delete | `pkg/server/tenant_delete.go:221` | 1min | ✅ `MarkTenantDeleteJobRunning` (optimistic lock) | Reliable but slow failover: 25-min S3 delete blocks takeover for 30-min TTL (`RecoverStaleTenantDeleteJobs`) |
| Pending tenant reconciler | `pkg/server/server.go:448` | 1min | ✅ `UpdateTenantStatusIf` | Duplicate scans, wasted queries |
| Mutation replay | `pkg/backend/mutation_replay.go:31` | 30s | ✅ `MarkMutationAppliedTx` | Depends on single-process per-tenant FIFO (`quota_async.go:15-20`); multi-pod concurrent apply breaks ordering |
| Expiry sweep | `pkg/backend/expiry_sweep.go:31` | 5min | ❌ no claim | Duplicate idempotent `ExpireActiveReservations`, wasted DB writes |
| Schema-init resume | `pkg/server/server.go:404` | at startup | ✅ `UpdateTenantStatusIf` | **Dangerous**: every pod resumes all `TenantProvisioning` tenants, duplicate TiDB Cloud provisioner API calls (external side effects) |
| File GC (per-tenant) | `pkg/backend/file_gc_worker.go:45` | 10s | ✅ claim+lease 5min | Safe, but pod crash leaves 5-min lease window before takeover |

**Proposed fix**: Introduce leader election so Object GC, Tenant Delete, Pending Reconciler, Mutation Replay, Expiry Sweep, and Schema-init resume run only on the leader pod. Two viable approaches:
- **TiDB `GET_LOCK` / advisory lock** (reuse the existing pattern at `pkg/meta/meta.go:245-329` `acquireMetaSchemaMigrationLock`): periodically renew, stop workers on lock loss.
- **Kubernetes Lease** (`coordination.k8s.io/lease`): leader pod holds and renews the lease.

Alternative for semantic/file-GC workers: **tenant sharding** by `tenantID % podCount == podIndex` (requires pod index/count via K8s StatefulSet or Downward API) so each pod only scans its own shard.

**Files**:
- `pkg/server/object_gc_worker.go:46-151`
- `pkg/server/tenant_delete.go:221-235`
- `pkg/server/server.go:404-466`
- `pkg/backend/mutation_replay.go:31-43`, `pkg/backend/quota_async.go:15-20`
- `pkg/backend/expiry_sweep.go:31-43`
- `pkg/backend/file_gc_worker.go:45-66`
- `pkg/meta/meta.go:245-329` (reusable `GET_LOCK` pattern)

---

### P1-1. Semantic worker scan amplification + non-cross-pod Kick

- `pkg/server/semantic_worker.go:139-163`: `inflight` map, `kickPending` map, `kicks chan` (buffer 256), tenant scan cursor — all in-process.
- N pods each scan the full tenant list (`TenantScanLimit` default 128) → N× scan amplification.
- `Kick` goes through the in-process `m.kicks` channel via `Pool.SetSemanticTaskNotifier` (`server.go:333`) → only notifies the local pod's worker; other pods discover new tasks only on next poll → latency grows with pod count.
- Task correctness is unaffected (`FOR UPDATE SKIP LOCKED` prevents double processing); this is efficiency/latency only.

**Proposed fix**:
- Short-term: expose `PollInterval` / `TenantScanLimit` as env vars (`buildSemanticWorkerConfigFromEnv` currently only exposes `Workers`), increase poll interval.
- Mid-term: tenant sharding (see P0-3).
- Kick cross-pod: publish to external pubsub (shared with P0-1), or DB `NOTIFY` / poll `semantic_tasks` for new rows.

**Files**: `pkg/server/semantic_worker.go:139-163, 310-374, 731-749`; `cmd/drive9-server/main.go:749-780`

### P1-2. Tenant backend LRU cache not shared — TiDB connection amplification

- `pkg/tenant/pool.go:57-67`: in-process LRU (`maxSize=128`), caches `*Dat9Backend` + `*datastore.Store` per tenant.
- Cache miss → each pod rebuilds backend (opens DB connection pool, ensures schema, migrates) → N pods open N connection pools for the same tenant.
- `Invalidate` only affects the local pod.

**Proposed fix**:
- Short-term: lower per-pod `maxSize` and `maxOpenConns`, ensure `N × poolSize ≤ TiDB max_connections`.
- Mid-term: shared connection layer (PgBouncer / TiDB Proxy) or backend-as-a-service.

**Files**: `pkg/tenant/pool.go:57-67, 208-294`

### P1-3. QuotaCache: one per tenant per pod, refresh goroutine amplification

- `pkg/backend/quota_cache.go:33-59`: per-tenant `quotaCache` with a 30s refresh goroutine.
- N pods → N independent caches + N refresh goroutines per tenant → N× DB reads.
- 30s window may serve stale quota (fail-open, accepted per `:29-32` comment).

**Proposed fix**: Short-term accept fail-open window (eventual consistency). Mid-term: DB row-level `SELECT ... FOR UPDATE` for exact quota, or Redis quota counter.

**Files**: `pkg/backend/quota_cache.go:33-59`, `pkg/backend/quota_store.go:118-124`

### P1-4. Per-tenant concurrency slot is per-pod — global concurrency amplified

- `pkg/server/semantic_worker.go:731-749` `tryClaimTenantSlot`: in-process `inflight` map, `PerTenantConcurrency` default 1.
- K pods → a tenant may run `K × PerTenantConcurrency` tasks in parallel.
- Correctness unaffected (DB lease guarantees single owner per task); only the precise global concurrency limit is lost.

**Proposed fix**: If strict global concurrency is needed, use a DB token row / lease or a distributed rate limiter. Otherwise accept K× amplification (capacity, not correctness).

### P1-5. Mutation queue depends on single-process FIFO

- `pkg/backend/quota_async.go:15-20` comment: "single worker guarantees per-tenant mutation apply order matches log insertion order ... Cross-instance ordering is not guaranteed".
- In-memory queue tasks are lost on crash (replayed by 30s `MutationReplayWorker`).
- Multi-pod: same tenant's mutations may apply concurrently on different pods, breaking FIFO.

**Proposed fix**: Run mutation apply leader-only (see P0-3), or `SELECT tenant_id ... FOR UPDATE` to serialize per-tenant mutation apply across pods. `MutationReplayWorker` already provides a convergence backstop.

**Files**: `pkg/backend/quota_async.go:15-31`, `pkg/backend/mutation_replay.go:31-43`

### P1-6. SSE per-connection poll causes unnecessary TiDB RCU consumption

- `pkg/server/sse.go` `handleEvents`: each SSE connection (one per FUSE mount) runs its own independent `poll := time.NewTicker(ssePollInterval)` (1s), querying `fs_events` even when no file changes occurred.
- N FUSE mounts on the same pod → N independent poll tickers → N QPS of identical "no new events" queries to the same tenant DB, even if all mounts are on the same pod and share the same `EventBus`.
- RCU consumption scales with mount count, not pod count — 100 mounts on 1 pod = 100 QPS.

**Proposed fix**: Move the poll goroutine from per-SSE-connection to per-EventBus (per-tenant, per-pod). The EventBus starts a single poll goroutine on first `Subscribe` and stops it on last `Unsubscribe`. When the poll discovers new `fs_events` rows, it calls `Publish()` to signal all local SSE listeners via the existing `notify` channel. SSE connections no longer poll DB directly — they only listen on `notify` + heartbeat.

| Scenario | Current (per-connection poll) | Optimized (per-bus poll) |
|---|---|---|
| 100 mounts on 1 pod | 100 QPS | 1 QPS |
| 100 mounts on 10 pods | 100 QPS | 10 QPS |
| No file changes | 100 QPS (all empty) | 1-10 QPS (all empty) |

Additionally, the poll interval can be relaxed from 1s to 2-5s since FUSE clients have HEAD revalidation as a correctness backstop.

**Files**: `pkg/server/eventbus.go` (add poll goroutine lifecycle), `pkg/server/sse.go` (remove per-connection `poll.C` from live loop)

---

### P2-1. Tenant schema init has no cross-pod mutex

- `pkg/server/server.go:404` `resumeProvisioningTenants`: every pod scans all `TenantProvisioning` tenants at startup and runs `initTenantSchemaAsync` independently.
- **db9 (PostgreSQL)**: all `CREATE TABLE IF NOT EXISTS` / `CREATE INDEX IF NOT EXISTS` (`pkg/tenant/db9/schema.go`) → natively idempotent, **safe**.
- **TiDB**: tables use `IF NOT EXISTS`; indexes/columns rely on `isIgnorableSchemaError` (`pkg/tenant/schema/common.go:137`) string-matching "duplicate key" / "already exist" / "duplicate column" to swallow errors. Converges, but:
  - String matching is fragile — if TiDB error wording changes, a duplicate is misclassified as fatal → tenant stuck in provisioning (retry window backstop exists).
  - `tidb_auto.go:787-792` `ALTER TABLE semantic ADD VECTOR INDEX ... ADD_COLUMNAR_REPLICA_ON_DEMAND` concurrent online DDL may emit non-duplicate errors (DDL job conflict) → retry noise.
  - `UpdateTenantSchemaVersion` (`meta.go:2257`) is an unconditional UPDATE but writes the same value, so no real risk.
  - `UpdateTenantStatusIf(Provisioning→Active)` (`server.go:4089`) has correct CAS to prevent double-activation.

**Proposed fix**:
- Add per-tenant `GET_LOCK` / advisory lock at `initTiDBAutoEmbeddingSchemaWithProfile` / `initDB9Schema` entry (reuse `meta.go:287` pattern).
- Replace `isIgnorableSchemaError` string matching with MySQL error code checks (1060/1061/1050) + PostgreSQL pgcode.
- Route schema-init resume through leader-only (see P0-3).

**Files**: `pkg/server/server.go:404-436, 3792-4089`; `pkg/tenant/schema/tidb_auto.go:787-792, 1183`; `pkg/tenant/schema/common.go:137`; `pkg/meta/meta.go:2257, 245-329`

### P2-2. `schemaInitErrors` is in-process

- `pkg/server/server.go:123` `schemaInitErrors sync.Map` stores per-tenant schema init errors for `/v1/status`.
- Pod A's failure is invisible to pod B's `/v1/status` → inconsistent client behavior.

**Proposed fix**: Persist error messages to the `tenants` table (`status_message` column); `/v1/status` reads from DB.

**Files**: `pkg/server/server.go:123, 1036-1041`

### P2-3. Observability metrics are fragmented per pod

- `pkg/backend/runtime_metrics.go:43` `globalBackendRuntimeMetrics` is a process singleton; `/metrics` only reflects the local pod's image/audio runtime state.

**Proposed fix**: Standard Prometheus aggregation (pod-level metrics are usually sufficient in K8s). Low priority.

---

### Phase 1 — P0 hard blockers (required for lossless scaling)
1. **Convert image extraction to durable `semantic_tasks`** — small, self-contained, high value. Reuses existing claim/lease pipeline.
2. **Introduce leader election** (TiDB `GET_LOCK` or K8s Lease) — gate Object GC / Tenant Delete / Pending Reconciler / Expiry Sweep / Schema-init resume / Mutation Replay behind the leader.
3. **Externalize event bus** — `EventBus` → Redis Pub/Sub (or NATS) adapter; subscribe + fan-out to local SSE listeners.

### Phase 2 — P1 soft obstacles (cost / convergence)
4. **Semantic worker tenant sharding** — `tenantID % podCount == podIndex`; expose `PollInterval` / `TenantScanLimit` as env vars; cross-pod Kick via pubsub.
5. **Control DB connection amplification** — lower per-pod `Pool.maxSize` / `maxOpenConns`; mid-term shared proxy.
6. **Serialize per-tenant mutation apply** — leader-only or `SELECT tenant FOR UPDATE`.

### Phase 3 — P2 low risk (cleanup / robustness)
7. **Per-tenant schema-init lock** + `isIgnorableSchemaError` → error-code-based.
8. **Persist `schemaInitErrors` to DB** for consistent `/v1/status`.
9. **Aggregate per-pod metrics** at the Prometheus layer.

---

## Key file index

| Topic | File:line |
|---|---|
| Event bus (in-memory) | `pkg/server/eventbus.go:23-32`, `pkg/server/sse.go:109-148` |
| Image extract queue (in-memory) | `pkg/backend/image_extract.go:90-122`, `pkg/backend/dat9.go:119-127` |
| Object GC worker (no claim) | `pkg/server/object_gc_worker.go:46-151` |
| Tenant delete worker | `pkg/server/tenant_delete.go:221-235` |
| Pending tenant reconciler | `pkg/server/server.go:448-466` |
| Mutation replay worker | `pkg/backend/mutation_replay.go:31-43`, `pkg/backend/quota_async.go:15-20` |
| Expiry sweep worker | `pkg/backend/expiry_sweep.go:31-43` |
| Schema-init resume | `pkg/server/server.go:404-436` |
| File GC worker (per-tenant, leased) | `pkg/backend/file_gc_worker.go:45-66` |
| Semantic worker manager | `pkg/server/semantic_worker.go:139-163, 310-374` |
| Tenant backend LRU cache | `pkg/tenant/pool.go:57-67, 208-294` |
| QuotaCache (per-tenant per-pod) | `pkg/backend/quota_cache.go:33-59` |
| File write CAS | `pkg/datastore/file_tx.go:47-147` |
| Semantic task claim (SKIP LOCKED) | `pkg/datastore/semantic_tasks.go:239-339` |
| Upload unique-index dedup | `pkg/datastore/store.go:2361-2369` |
| FUSE client state | `pkg/fuse/dat9fs.go:31-127`, `pkg/fuse/handle.go:13-55` |
| Meta schema migration lock (reusable pattern) | `pkg/meta/meta.go:245-329` |
| Tenant schema init | `pkg/server/server.go:3792-4089`, `pkg/tenant/schema/tidb_auto.go:1183`, `pkg/tenant/schema/common.go:137` |
| schemaInitErrors (in-memory) | `pkg/server/server.go:123, 1036-1041` |

mornyx:
## Summary

Migrates app-embedding mode image extraction from a non-durable in-process channel to the existing durable `semantic_tasks` pipeline, making it crash-recoverable and multi-pod safe.

This addresses **P0-1** from #599: the legacy in-memory `imageExtractQueue` channel dropped tasks on queue-full and lost them on pod crash/restart.

## Why this is mostly a deletion

The durable pipeline was **already fully wired** for auto-embedding mode:
- `enqueueImgExtractTaskTx` (`pkg/backend/semantic_tasks.go`) exists
- `processImgExtractTask` (`pkg/server/semantic_worker.go`) exists and calls `ProcessImageExtractTask`
- `taskTypesForTarget` already includes `TaskTypeImgExtractText` in app mode
- `ProcessImageExtractTask` is shared by both paths

The only thing still using the legacy queue was **three app-mode enqueue call sites** that called `enqueueImageExtract*` post-commit instead of `enqueueImgExtractTaskTx` in-tx, plus a **viability gap**: the semantic worker wouldn't start in app-mode-without-embedder.

### 1. Extend semantic worker viability for app-mode image/audio extract without embedder
- `newSemanticWorkerManager`: single-tenant viability now includes `fallback.SupportsAsyncImageExtract()` / `SupportsAsyncAudioExtract()`
- `taskTypesForProvider`: non-auto-embedding providers now include pool-configured extract types via `unionTaskTypes(appManaged, poolExtractTaskTypes)`
- `shouldIncludeFallback`: non-auto fallback includes `fallbackExtractTaskTypes()`

### 2. Replace three app-mode enqueue call sites with in-tx durable enqueue
- `createAndWriteCtx` (`dat9.go`): in-tx `enqueueExtractSemanticTasksTx` + embed task
- `overwriteFileCtxWithRev` (`dat9.go`): same pattern
- `finalizeUpload` (`upload.go`): both overwrite and new-file branches

### 3. Remove the legacy in-process image queue
- Removed: `imageExtractQueue` channel, `imageExtractWG`, `imageExtractCancel`
- Removed: `enqueueImageExtract`, `enqueueImageExtractForUpload`, `runImageExtractWorker`, `processQueuedImageExtractTask`, `legacyImageExtractMetricResult`
- Removed: unused `mediaLLMQuotaExceededCheck` (non-tx variant), `setImageQueueDepth`
- Marked `QueueSize`/`Workers` fields as deprecated (kept for config compatibility)

### 4. Rename `enqueueTiDBAutoSemanticTasksTx` → `enqueueExtractSemanticTasksTx`
Reflects that image/audio extract applies in both embedding modes, not just TiDB auto.

## Schema changes

None. `semantic_tasks` table and `TaskTypeImgExtractText` already exist in all schema files.

## Test results

- `pkg/backend`: all tests pass (including new app-mode tests + updated legacy-queue tests)
- `pkg/server`: no new failures (pre-existing meta-schema migration failures unrelated to this PR)
- `pkg/datastore`: all tests pass
- `local-smoke.sh`: 32/32 pass (API, CLI, FUSE, POSIX permission)
- `golangci-lint`: 0 issues in modified packages

Closes #599 (P0-1).

## Release Notes

* **Improvements**
  * Image-to-text extraction now uses durable semantic tasks end-to-end for more reliable, resilient processing.
  * App-embedding mode now enqueues image extraction tasks within the same operation, with embedding tasks only when applicable.
  * Image extraction can be processed by the semantic worker even when no embedder is configured.
  * Audio transcription writeback may now enqueue follow-up semantic embedding tasks in app-embedding mode.
* **Behavior Changes**
  * Async image extraction configuration options for queue/workers are deprecated and ignored.
  * Runtime metrics for image extract queue depth reflect the new durable-task flow.
* **Tests**
  * Updated tests to deterministically drive extraction and validate enqueued tasks.

### DIFF
--- a/pkg/backend/audio_extract.go
+++ b/pkg/backend/audio_extract.go
@@ -367,6 +367,19 @@ func (b *Dat9Backend) ProcessAudioExtractTask(ctx context.Context, task AudioExt
 		}
 		var txErr error
 		updated, txErr = b.store.UpdateFileSearchTextTx(tx, task.FileID, expectedRevision, text)
+		if txErr != nil {
+			return txErr
+		}
+		if !updated {
+			return nil
+		}
+		// Bridge an embed task in app-embedding mode so the transcribed text
+		// gets embedded for semantic search. In auto-embedding mode the DB
+		// generates the vector automatically, so no bridge is needed.
+		if b.UsesDatabaseAutoEmbedding() || !b.appSemanticTasksEnabled {
+			return nil
+		}
+		_, txErr = b.store.EnsureSemanticTaskQueuedTx(tx, newEmbedTask(b.genID(), task.FileID, expectedRevision, time.Now().UTC()))
 		return txErr
 	})
 	if err != nil {
--- a/pkg/backend/dat9.go
+++ b/pkg/backend/dat9.go
@@ -116,12 +116,11 @@ type Dat9Backend struct {
 
 	s3EncryptionPolicy meta.ResolvedS3EncryptionPolicy
 
-	// Async image -> text extraction worker (in-memory queue for P0).
+	// Async image -> text extraction runtime (durable semantic_tasks delivery;
+	// no in-process queue). The semantic worker claims img_extract_text tasks
+	// and calls ProcessImageExtractTask via this backend's extractor/S3 client.
 	imageExtractEnabled bool
 	imageExtractor      ImageTextExtractor
-	imageExtractQueue   chan ImageExtractTaskSpec
-	imageExtractWG      sync.WaitGroup
-	imageExtractCancel  context.CancelFunc
 	imageExtractTimeout time.Duration
 	imageExtractMaxSize int64
 	maxExtractTextBytes int
@@ -858,15 +857,23 @@ func (b *Dat9Backend) createAndWriteCtx(ctx context.Context, path string, data [
 			}
 		}
 		if b.UsesDatabaseAutoEmbedding() {
-			created, err := b.enqueueTiDBAutoSemanticTasksTx(ctx, tx, fileID, 1, path, contentType)
+			created, err := b.enqueueExtractSemanticTasksTx(ctx, tx, fileID, 1, path, contentType)
 			semanticTaskEnqueued = created
 			return err
+		}
+		// App-embedding mode: image/audio extract tasks are durable and independent
+		// of EMBED_TEXT, so register them in the same transaction. The embed task
+		// (if any) is enqueued separately below.
+		extractCreated, extractErr := b.enqueueExtractSemanticTasksTx(ctx, tx, fileID, 1, path, contentType)
+		if extractErr != nil {
+			return extractErr
 		}
 		if b.shouldEnqueueEmbedForRevision(path, contentType, contentText, description) {
 			created, err := b.enqueueEmbedTaskTx(tx, fileID, 1)
-			semanticTaskEnqueued = created
+			semanticTaskEnqueued = created || extractCreated
 			return err
 		}
+		semanticTaskEnqueued = extractCreated
 		return nil
 	}); err != nil {
 		if timingEnabled {
@@ -881,9 +888,6 @@ func (b *Dat9Backend) createAndWriteCtx(ctx context.Context, path string, data [
 		tenantTxDuration = time.Since(txStart)
 	}
 	b.notifySemanticTaskEnqueued(semanticTaskEnqueued)
-	// Temporary compatibility: app embedding still relies on the legacy
-	// backend-owned image queue until its image task flow also moves to
-	// semantic_tasks.
 	centralQuotaStart := time.Time{}
 	if timingEnabled {
 		centralQuotaStart = time.Now()
@@ -891,17 +895,6 @@ func (b *Dat9Backend) createAndWriteCtx(ctx context.Context, path string, data [
 	b.syncCentralFileCreate(ctx, fileID, int64(len(data)), contentType)
 	if timingEnabled {
 		centralQuotaDuration = time.Since(centralQuotaStart)
-	}
-	if b.UsesDatabaseAutoEmbedding() {
-		return int64(len(data)), nil
-	}
-	imageEnqueueStart := time.Time{}
-	if timingEnabled {
-		imageEnqueueStart = time.Now()
-	}
-	b.enqueueImageExtract(fileID, path, contentType, 1)
-	if timingEnabled {
-		imageEnqueueDuration = time.Since(imageEnqueueStart)
 	}
 	return int64(len(data)), nil
 }
@@ -1094,15 +1087,23 @@ func (b *Dat9Backend) overwriteFileCtxWithRev(ctx context.Context, nf *datastore
 			}
 		}
 		if b.UsesDatabaseAutoEmbedding() {
-			created, err := b.enqueueTiDBAutoSemanticTasksTx(ctx, tx, nf.File.FileID, newRevision, nf.Node.Path, contentType)
+			created, err := b.enqueueExtractSemanticTasksTx(ctx, tx, nf.File.FileID, newRevision, nf.Node.Path, contentType)
 			semanticTaskEnqueued = created
 			return err
+		}
+		// App-embedding mode: image/audio extract tasks are durable and independent
+		// of EMBED_TEXT, so register them in the same transaction. The embed task
+		// (if any) is enqueued separately below.
+		extractCreated, extractErr := b.enqueueExtractSemanticTasksTx(ctx, tx, nf.File.FileID, newRevision, nf.Node.Path, contentType)
+		if extractErr != nil {
+			return extractErr
 		}
 		if b.shouldEnqueueEmbedForRevision(nf.Node.Path, contentType, contentText, description) {
 			created, err := b.enqueueEmbedTaskTx(tx, nf.File.FileID, newRevision)
-			semanticTaskEnqueued = created
+			semanticTaskEnqueued = created || extractCreated
 			return err
 		}
+		semanticTaskEnqueued = extractCreated
 		return nil
 	})
 	if timingEnabled {
@@ -1132,20 +1133,6 @@ func (b *Dat9Backend) overwriteFileCtxWithRev(ctx context.Context, nf *datastore
 	b.deleteBlobIfS3Ctx(ctx, nf.File.StorageType, nf.File.StorageRef, storageRef)
 	if timingEnabled {
 		oldBlobCleanupDuration = time.Since(oldBlobCleanupStart)
-	}
-	// Temporary compatibility: app embedding still relies on the legacy
-	// backend-owned image queue until its image task flow also moves to
-	// semantic_tasks.
-	if b.UsesDatabaseAutoEmbedding() {
-		return int64(len(data)), newRevision, nil
-	}
-	imageEnqueueStart := time.Time{}
-	if timingEnabled {
-		imageEnqueueStart = time.Now()
-	}
-	b.enqueueImageExtract(nf.File.FileID, nf.Node.Path, contentType, newRevision)
-	if timingEnabled {
-		imageEnqueueDuration = time.Since(imageEnqueueStart)
 	}
 	return int64(len(data)), newRevision, nil
 }
--- a/pkg/backend/image_extract.go
+++ b/pkg/backend/image_extract.go
@@ -13,9 +13,7 @@ import (
 	"github.com/pingcap/failpoint"
 
 	"github.com/mem9-ai/drive9/pkg/datastore"
-	"github.com/mem9-ai/drive9/pkg/logger"
 	"github.com/mem9-ai/drive9/pkg/metrics"
-	"go.uber.org/zap"
 )
 
 // ImageExtractRequest is the input to a pluggable image->text extractor.
@@ -85,144 +83,6 @@ func contentTypeFromPath(path string) string {
 // runtime dependencies needed to extract image text.
 func (b *Dat9Backend) SupportsAsyncImageExtract() bool {
 	return b.imageExtractEnabled && b.imageExtractor != nil
-}
-
-func (b *Dat9Backend) enqueueImageExtract(fileID, path, contentType string, revision int64) {
-	if !b.SupportsAsyncImageExtract() {
-		return
-	}
-	if !isImageContentType(contentType) {
-		contentType = contentTypeFromPath(path)
-		if !isImageContentType(contentType) {
-			return
-		}
-	}
-	if b.mediaLLMQuotaExceededCheck(backgroundWithTrace()) {
-		metrics.RecordOperation("media_llm_budget", "enqueue_skip", "quota_exceeded", 0)
-		return
-	}
-	task := ImageExtractTaskSpec{
-		FileID:      fileID,
-		Path:        path,
-		ContentType: contentType,
-		Revision:    revision,
-	}
-	select {
-	case b.imageExtractQueue <- task:
-		metrics.RecordOperation("image_extract", "enqueue", "ok", 0)
-		globalBackendRuntimeMetrics.setImageQueueDepth(b.runtimeMetricsID, len(b.imageExtractQueue))
-	default:
-		metrics.RecordOperation("image_extract", "enqueue", "queue_full", 0)
-		globalBackendRuntimeMetrics.setImageQueueDepth(b.runtimeMetricsID, len(b.imageExtractQueue))
-		logger.Warn(backgroundWithTrace(), "backend_image_extract_queue_full_drop",
-			zap.String("file_id", fileID),
-			zap.String("path", path),
-			zap.Int("queue_size", cap(b.imageExtractQueue)))
-	}
-}
-
-func (b *Dat9Backend) enqueueImageExtractForUpload(ctx context.Context, upload *datastore.Upload, isOverwrite bool) {
-	if !b.SupportsAsyncImageExtract() {
-		return
-	}
-	fileID := upload.FileID
-	if isOverwrite {
-		nf, err := b.store.Stat(ctx, upload.TargetPath)
-		if err != nil || nf.File == nil {
-			logger.Warn(ctx, "backend_image_extract_upload_enqueue_stat_failed",
-				zap.String("upload_id", upload.UploadID),
-				zap.String("path", upload.TargetPath),
-				zap.Error(err))
-			return
-		}
-		fileID = nf.File.FileID
-	}
-	f, err := b.store.GetFile(ctx, fileID)
-	if err != nil {
-		logger.Warn(ctx, "backend_image_extract_upload_enqueue_get_file_failed",
-			zap.String("upload_id", upload.UploadID),
-			zap.String("file_id", fileID),
-			zap.String("path", upload.TargetPath),
-			zap.Error(err))
-		return
-	}
-	ct := f.ContentType
-	if ct == "" {
-		ct = contentTypeFromPath(upload.TargetPath)
-		if ct == "" {
-			logger.Warn(ctx, "backend_image_extract_upload_enqueue_skipped_unknown_content_type",
-				zap.String("upload_id", upload.UploadID),
-				zap.String("file_id", fileID),
-				zap.String("path", upload.TargetPath),
-				zap.String("reason", "content_type_missing_and_extension_unknown"))
-			return
-		}
-	}
-	b.enqueueImageExtract(fileID, upload.TargetPath, ct, f.Revision)
-}
-
-func (b *Dat9Backend) runImageExtractWorker(ctx context.Context, workerID int) {
-	defer b.imageExtractWG.Done()
-	logger.Info(ctx, "backend_image_extract_worker_started", zap.Int("worker_id", workerID))
-	for {
-		select {
-		case <-ctx.Done():
-			logger.Info(ctx, "backend_image_extract_worker_stopped", zap.Int("worker_id", workerID), zap.String("reason", "context_canceled"))
-			return
-		case task := <-b.imageExtractQueue:
-			globalBackendRuntimeMetrics.setImageQueueDepth(b.runtimeMetricsID, len(b.imageExtractQueue))
-			b.processQueuedImageExtractTask(ctx, task)
-		}
-	}
-}
-
-func (b *Dat9Backend) processQueuedImageExtractTask(ctx context.Context, task ImageExtractTaskSpec) {
-	start := time.Now()
-	result, err := b.ProcessImageExtractTask(ctx, task)
-	metricResult := legacyImageExtractMetricResult(result)
-	if err != nil {
-		switch result {
-		case ImageExtractResultGetFileError:
-			logger.Warn(ctx, "backend_image_extract_get_file_failed",
-				zap.String("file_id", task.FileID), zap.Error(err))
-		case ImageExtractResultLoadError:
-			logger.Warn(ctx, "backend_image_extract_load_bytes_failed",
-				zap.String("file_id", task.FileID), zap.Error(err))
-		case ImageExtractResultExtractError:
-			logger.Warn(ctx, "backend_image_extract_failed",
-				zap.String("file_id", task.FileID), zap.String("path", task.Path), zap.Error(err))
-		case ImageExtractResultUpdateError:
-			logger.Warn(ctx, "backend_image_extract_update_file_failed",
-				zap.String("file_id", task.FileID), zap.Error(err))
-		default:
-			logger.Warn(ctx, "backend_image_extract_failed_unexpected",
-				zap.String("file_id", task.FileID), zap.String("path", task.Path), zap.Error(err))
-		}
-		metrics.RecordOperation("image_extract", "process", metricResult, time.Since(start))
-		return
-	}
-
-	switch result {
-	case ImageExtractResultFileNotFound, ImageExtractResultNotConfirmed, ImageExtractResultNotImage:
-		metrics.RecordOperation("image_extract", "process", metricResult, time.Since(start))
-		return
-	case ImageExtractResultStale:
-		logger.Info(ctx, "backend_image_extract_skipped_stale",
-			zap.String("file_id", task.FileID), zap.String("path", task.Path))
-	case ImageExtractResultTooLarge:
-		logger.Warn(ctx, "backend_image_extract_skipped_too_large_or_empty",
-			zap.String("file_id", task.FileID),
-			zap.String("path", task.Path),
-			zap.Int64("max_bytes", b.imageExtractMaxSize))
-	case ImageExtractResultEmptyText:
-		logger.Warn(ctx, "backend_image_extract_empty_text",
-			zap.String("file_id", task.FileID),
-			zap.String("path", task.Path))
-	case ImageExtractResultWritten:
-		logger.Info(ctx, "backend_image_extract_ok",
-			zap.String("file_id", task.FileID), zap.String("path", task.Path))
-	}
-	metrics.RecordOperation("image_extract", "process", metricResult, time.Since(start))
 }
 
 // ProcessImageExtractTask runs the backend-owned image extraction logic for one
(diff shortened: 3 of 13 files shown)


srstack (commented):
## Review: feat/durable-image-extract

**Overall: LGTM with one thing to verify before merge.**

### What this does

Replaces the legacy in-process `imageExtractQueue` channel with the existing durable `semantic_tasks` pipeline for app-embedding mode. The durable infrastructure was already fully wired; this PR removes three post-commit call sites to `enqueueImageExtract*` and replaces them with in-transaction `enqueueExtractSemanticTasksTx` calls — the same path auto-embedding mode already used. Crash-safety and multi-pod correctness follow for free.

---

### One thing to verify

**`poolExtractTaskTypes()` delegates to `pool.AutoSemanticTaskTypes()`**, which is named "auto" and likely returns all auto-embedding task types — potentially including `EMBED_TEXT`. In `taskTypesForProvider`, the non-DB9 app-mode path becomes:

```go
return unionTaskTypes(m.appManagedTaskTypes(), m.poolExtractTaskTypes())
```

If `pool.AutoSemanticTaskTypes()` returns `EMBED_TEXT` alongside `img_extract_text`/`audio_extract_text`, then `ClaimSemanticTask` for app-mode tenants would include `EMBED_TEXT` in its filter — causing embed tasks to be claimed by the pool worker path when the pool doesn't handle them. Please verify `pool.AutoSemanticTaskTypes()` only returns extract task types, or introduce a dedicated `poolImageAudioExtractTaskTypes()` that explicitly filters to `{TaskTypeImgExtractText, TaskTypeAudioExtractText}`.

---

### Minor notes

- **`queue_capacity`/`workers` metrics are now always 0**: any alerts or dashboards watching these gauges for capacity signals should be updated. The test already asserts the 0 values correctly.
- **`semanticEnqueueDurationMs` timing in `finalizeUpload`**: in the no-embed branch, `stepStart` covers both extract and embed enqueue time — minor inaccuracy, not a correctness issue.
- **Test rewrites** are a real improvement: calling `ProcessImageExtractTask` directly instead of racing against a background goroutine makes the tests deterministic and easier to read.
- **Rename** `enqueueTiDBAutoSemanticTasksTx` → `enqueueExtractSemanticTasksTx` is correct and overdue.

srstack (approved):
## Review: feat/durable-image-extract

**Overall: LGTM — one thing to verify before merge.**

### What this does

Replaces the legacy in-process `imageExtractQueue` channel with the existing durable `semantic_tasks` pipeline for app-embedding mode. The durable infrastructure was already fully wired; this PR removes three post-commit call sites to `enqueueImageExtract*` and replaces them with in-transaction `enqueueExtractSemanticTasksTx` calls — the same path auto-embedding mode already used. Crash-safety and multi-pod correctness follow for free.

---

### One thing to verify before merge

**`poolExtractTaskTypes()` delegates to `pool.AutoSemanticTaskTypes()`**, which is named "auto" and likely returns all auto-embedding task types — potentially including `EMBED_TEXT`. In `taskTypesForProvider`, the non-DB9 app-mode path becomes:

```go
return unionTaskTypes(m.appManagedTaskTypes(), m.poolExtractTaskTypes())
```

If `pool.AutoSemanticTaskTypes()` returns `EMBED_TEXT` alongside `img_extract_text`/`audio_extract_text`, then `ClaimSemanticTask` for app-mode tenants would include `EMBED_TEXT` in its filter — potentially causing embed tasks to be claimed by the pool worker when the pool doesn't support them. Please verify `pool.AutoSemanticTaskTypes()` only returns extract task types, or introduce a dedicated helper that explicitly filters to `{TaskTypeImgExtractText, TaskTypeAudioExtractText}`.

---

### Minor notes

- **`queue_capacity`/`workers` metrics are now always 0**: any alerts or dashboards watching these gauges for capacity signals should be updated.
- **`semanticEnqueueDurationMs` in `finalizeUpload`**: in the no-embed branch, `stepStart` covers both extract and embed enqueue time combined — minor timing inaccuracy, not a correctness issue.
- **Test rewrites** calling `ProcessImageExtractTask` directly instead of racing a background goroutine are a clear improvement — deterministic and readable.
- **Rename** `enqueueTiDBAutoSemanticTasksTx` → `enqueueExtractSemanticTasksTx` is correct and overdue.

qiffang:
Blocking issue found.

1. `pkg/backend/dat9.go:864-877` / `pkg/backend/upload.go:1097-1111,1167-1175` now enqueue durable `img_extract_text` / `audio_extract_text` tasks in app-embedding mode for any backend where async extract is wired. But `pkg/server/semantic_worker.go:1469-1478` explicitly filters DB9 providers down to app-managed embed tasks and excludes pool image/audio extract types, with the nearby runtime TODO still saying DB9/Postgres tenants are unsupported in the semantic worker target path. In a multi-tenant deployment with DB9 tenants and async image/audio extraction enabled in the pool template, image/audio writes will commit extract tasks, notify the worker, and then those tasks can never be scanned or claimed. This is also a behavior regression from the old in-process image extraction path: DB9/app-mode image extraction no longer completes and the durable queue accumulates permanent `queued` rows.

Suggested fix: keep enqueue and worker routing aligned for DB9. Either disable async image/audio extraction on DB9 tenant backends / skip `enqueueExtractSemanticTasksTx` for those providers, or add real DB9 semantic-worker target support before allowing those task rows. Please add a regression that writes an image/audio file for a DB9 provider tenant with pool async extract configured and verifies we do not strand an unclaimable extract task.

Validation I ran:
- `go test ./pkg/backend -run 'Test(AsyncImageExtract|Write.*AppEmbedding|ConfirmUploadAppEmbedding|Dat9BackendAutoSemanticTaskTypes|PoolAutoSemanticTaskTypes|SemanticTask|.*SemanticTask.*)' -count=1`
- `go test ./pkg/server -run 'TestSemanticWorker(ProcessesImgExtractTaskWithoutEmbedder|StartsForAppModeImageExtractWithoutEmbedder|ProcessesAudioExtractTaskWithoutEmbedder|DeadLettersNonRetryableAudioExtractError|ProcessesMP4AudioExtractTaskWithoutEmbedder|ImageOnlySkipsLegacyEmbedBacklog|EmbedOnlySkipsImageTasks|TaskTypesForTarget|TaskTypesForTargetDisabledAutoEmbedWithImageExtract|KickClaimsFallbackTaskWithoutScan)$|TestPoolAutoSemanticTaskTypesDisabledAutoEmbed' -count=1`

I also attempted a broader `pkg/server` semantic-worker run, but this local checkout still hits the known test DB schema drift around `llm_usage.tenant_id`, so I did not treat that as PR-specific.

mornyx:
Thanks @qiffang @srstack for the review! All feedback has been addressed in the latest push (aea7d4e).

### @qiffang — DB9 provider extract task routing

**You were right** that the DB9 exclusion in `taskTypesForProvider` would strand unclaimable extract tasks. I reverted the exclusion — DB9 backends with async extract configured can process `img_extract_text`/`audio_extract_text` tasks because `ProcessImageExtractTask`/`ProcessAudioExtractTask` use the same `datastore.Store` interface regardless of provider. The per-backend `taskTypesForTarget` filter ensures only backends with extract runtimes wired (`imageExtractEnabled`, `imageExtractor != nil`) will actually claim these tasks.

Tests updated: `TestSemanticWorkerListTenantRefsImageOnlyIncludesAutoProviders` and `TestSemanticWorkerListTenantRefsRotatesAcrossActiveTenantPages` now expect DB9 tenants in scan results when the pool configures async extract.

### @srstack — poolExtractTaskTypes / EMBED_TEXT concern

Good catch checking this. `pool.AutoSemanticTaskTypes()` only returns `{TaskTypeImgExtractText, TaskTypeAudioExtractText}` — it never includes `EMBED_TEXT` (see `pool.go:500-519`). So `unionTaskTypes(appManaged, poolExtractTaskTypes)` is safe: app-mode `ClaimSemanticTask` will only claim extract + app-managed embed types, not auto-embedding embed types.

### Additional fixes from CodeRabbit/Codex reviews

1. **App-mode audio extract embed bridge**: `ProcessAudioExtractTask` now bridges an `embed` task via `EnsureSemanticTaskQueuedTx` in the writeback transaction (same pattern as `ProcessImageExtractTask`). Without this, app-mode audio transcripts were written to `content_text` but never embedded for semantic search.

2. **Preserve embed routing when auto-embedding disabled**: `taskTypesForProvider` auto-embedding branch now unions app-managed embed types with pool extract types when `IsAutoEmbeddingDisabled()` is true, instead of returning extract types alone (which would strand embed tasks).

3. **Metrics test**: re-asserted `queue_capacity=0` and `workers=0` after `b1.Close()` to catch stale aggregate-state regressions.

qiffang:
Re-reviewed updated head `aea7d4ef3af71320d66f6784c9edeac382651a6f`.

No remaining merge-blocking issues found from my side. The previous DB9 stranded-task blocker is addressed: provider scanning now includes non-auto providers such as DB9 when pool extract runtimes are configured, while `taskTypesForTarget` still gates actual claims on the acquired backend's `SupportsAsyncImageExtract` / `SupportsAsyncAudioExtract`. The added list-scan expectations plus the existing target-type tests cover the routing mismatch I called out. I also checked the new app-mode audio bridge; it now queues the follow-up embed task after transcript writeback only when app semantic tasks are enabled and DB auto-embedding is not active, matching the image path.

Validation I ran:
- `go test ./pkg/backend -run 'Test(AsyncImageExtract|AudioExtract|Write.*AppEmbedding|ConfirmUploadAppEmbedding|Dat9BackendAutoSemanticTaskTypes|PoolAutoSemanticTaskTypes|SemanticTask|.*SemanticTask.*|BackendRuntimeMetrics)' -count=1`
- `go test ./pkg/server -run 'TestSemanticWorkerHTTPMultiTenantImageOnlySkipsAppTenantEmbedTasks|TestSemanticWorkerListTenantRefsImageOnlyIncludesAutoProviders|TestSemanticWorkerKickClaimsPoolTenantTaskWithoutScan' -count=1`
- `go test ./pkg/tenant -run TestPoolAutoSemanticTaskTypes -count=1`
- `gh pr checks 600` — `ci`, `drive9-server-local e2e smoke`, and CodeRabbit all passing

I also attempted a broader semantic-worker regex run, but this local shared test DB still reproduced the known `llm_usage.tenant_id` schema-drift failure when multiple meta tests are batched together; targeted runs above pass, and GitHub CI is green.

PR #601, opened 2026-06-23
feat: leader election via TiDB GET_LOCK for background schedulers

mornyx:
## Summary

Introduces a TiDB/MySQL `GET_LOCK`-based leader election mechanism so that background schedulers run only on the leader pod in a multi-pod deployment. This prevents duplicate work across pods — especially for Object GC which has **no DB-level claim/lease protection** and would otherwise produce duplicate S3 HEAD/DELETE calls.

This addresses **P0-3** from #599, and is the **foundation for P0-2** (event bus externalization), which will also use the leader manager to coordinate event publishing. The PR numbering (P0-3 before P0-2) is intentional: leader election must exist before event bus externalization can safely coordinate cross-pod event delivery.

## New `pkg/leader` package

- `Manager` acquires a named session lock via `GET_LOCK('drive9:leader', 0)` (non-blocking, timeout=0)
- Connection death (pod crash/restart) auto-releases the lock — MySQL session-level semantics
- Heartbeat goroutine verifies lock ownership via `IS_USED_LOCK` and keeps the connection alive with `SELECT 1`
- `SetCallbacks` for `onLead`/`onLose` leadership transitions
- `WithDisabled` option for single-pod mode (always leader, no lock acquired)
- Config: `DRIVE9_LEADER_LOCK_NAME`, `DRIVE9_LEADER_HEARTBEAT_INTERVAL`, `DRIVE9_LEADER_DISABLED`

## Workers gated behind leader

| Worker | Gate mechanism | Why gated |
|--------|---------------|-----------|
| Semantic worker | Start/Stop on leadership change | Reduces scan amplification (claim already safe via SKIP LOCKED) |
| Object GC worker | Start/Stop on leadership change | **No DB-level claim** — most critical to gate |
| Tenant delete cleanup | Leader-gated goroutine group | Prevents duplicate S3 prefix deletes |
| Pending tenant reconciler | Leader-gated goroutine group | Prevents duplicate scans |
| One-time resume tasks | Leader-gated goroutines | Prevents duplicate TiDB Cloud provisioner calls |
| Mutation replay worker | Start/Stop on leadership change (main.go) | Prevents duplicate quota mutation scans |
| Expiry sweep worker | Start/Stop on leadership change (main.go) | Prevents duplicate reservation sweeps |
| Per-tenant FileGCWorker | Pool `LeaderChecker` interface | Reduces per-tenant scan overhead |

## Backward compatibility

- **nil leader manager** (drive9-server-local, single-pod) = all workers start immediately (existing behavior, unchanged)
- **`DRIVE9_LEADER_DISABLED=true`** = always leader, no lock acquired
- No schema changes required
- `local-smoke.sh`: 32/32 pass (single-pod, leader disabled)

## Test results

- `pkg/leader`: all tests pass (acquire/lose/reacquire, exclusivity, disabled mode, primitives)
- `pkg/server`: no new failures (pre-existing meta-schema migration failures unrelated to this PR)
- `pkg/tenant`: all tests pass
- `cmd/drive9-server` / `cmd/drive9-server-local`: all tests pass
- `golangci-lint`: 0 new issues in modified packages

Ref #599 (P0-3).

## Summary by CodeRabbit

- **New Features**
  - Added MySQL `GET_LOCK`-based leader election with lead/lose callbacks.
  - Background maintenance and processing now run only on the active leader; FileGC is also leader-gated when configured for multi-instance operation.

- **Refactor**
  - Reworked server startup/shutdown so leader transitions control worker lifecycles.
  - Improved fork-tenant cleanup to run concurrently with bounded parallelism.
  - Adjusted tenant delete job scheduling to be handled without a periodic cleanup ticker.

- **Tests**
  - Added leader-election, server leader lifecycle, and FileGC leader-gating race coverage.

### DIFF
--- a/cmd/drive9-server/main.go
+++ b/cmd/drive9-server/main.go
@@ -244,6 +244,11 @@ func main() {
 			if v := envOr("DRIVE9_LEADER_HEARTBEAT_INTERVAL", ""); v != "" {
 				if d, err := time.ParseDuration(v); err == nil {
 					leaderOpts = append(leaderOpts, leader.WithHeartbeatInterval(d))
+				} else {
+					logger.Warn(context.Background(), "leader_heartbeat_interval_invalid_falling_back_to_default",
+						zap.String("env", "DRIVE9_LEADER_HEARTBEAT_INTERVAL"),
+						zap.String("value", v),
+						zap.Error(err))
 				}
 			}
 			leaderManager = leader.NewManager(store.DB(), leaderOpts...)
@@ -271,33 +276,12 @@ func main() {
 
 		pool.SetMetaStore(store)
 
-		// Gate mutation replay and expiry sweep workers behind leadership.
-		var replayWorker *backend.MutationReplayWorker
-		var expirySweepWorker *backend.ExpirySweepWorker
-		leaderManager.SetCallbacks(
-			func() {
-				replayWorker = backend.StartMutationReplayWorker(tenant.NewMetaQuotaAdapter(store))
-				expirySweepWorker = backend.StartExpirySweepWorker(store)
-			},
-			func() {
-				if replayWorker != nil {
-					replayWorker.Stop()
-					replayWorker = nil
-				}
-				if expirySweepWorker != nil {
-					expirySweepWorker.Stop()
-					expirySweepWorker = nil
-				}
-			},
-		)
-		defer func() {
-			if replayWorker != nil {
-				replayWorker.Stop()
-			}
-			if expirySweepWorker != nil {
-				expirySweepWorker.Stop()
-			}
-		}()
+		// The mutation replay and expiry sweep workers are owned by the server
+		// (started/stopped in its leader-gated startLeaderWorkers/stopLeaderWorkers),
+		// so they follow leadership transitions alongside the other leader-gated
+		// schedulers. main.go no longer registers a competing SetCallbacks pair
+		// (that earlier pair was clobbered by the server's own SetCallbacks and
+		// never fired).
 
 		// TODO: Run ValidateDurableAsyncExtractRequiresSemanticWorker only when this process
 		// can serve tenants that enqueue durable audio_extract_text / img_extract_text
--- a/cmd/drive9-server/main.go
+++ b/cmd/drive9-server/main.go
@@ -23,6 +23,7 @@ import (
 	"github.com/mem9-ai/drive9/pkg/buildinfo"
 	"github.com/mem9-ai/drive9/pkg/embedding"
 	"github.com/mem9-ai/drive9/pkg/encrypt"
+	"github.com/mem9-ai/drive9/pkg/leader"
 	"github.com/mem9-ai/drive9/pkg/logger"
 	"github.com/mem9-ai/drive9/pkg/meta"
 	"github.com/mem9-ai/drive9/pkg/s3client"
@@ -205,6 +206,7 @@ func main() {
 	}
 
 	var pool *tenant.Pool
+	var leaderManager *leader.Manager
 	var tokenSecret []byte
 	if tokenHex != "" {
 		tokenSecret, err = hex.DecodeString(tokenHex)
@@ -230,6 +232,24 @@ func main() {
 			die(fmt.Errorf("control-plane db unavailable: %w", err))
 		}
 
+		// Create a leader election manager for gating background schedulers.
+		// When disabled (single-pod), IsLeader always returns true and workers
+		// start immediately. In multi-pod mode, workers start only on the leader.
+		if os.Getenv("DRIVE9_LEADER_DISABLED") == "1" || os.Getenv("DRIVE9_LEADER_DISABLED") == "true" {
+			leaderManager = leader.NewManager(store.DB(), leader.WithDisabled())
+		} else {
+			leaderOpts := []leader.Option{
+				leader.WithLockName(envOr("DRIVE9_LEADER_LOCK_NAME", "drive9:leader")),
+			}
+			if v := envOr("DRIVE9_LEADER_HEARTBEAT_INTERVAL", ""); v != "" {
+				if d, err := time.ParseDuration(v); err == nil {
+					leaderOpts = append(leaderOpts, leader.WithHeartbeatInterval(d))
+				}
+			}
+			leaderManager = leader.NewManager(store.DB(), leaderOpts...)
+		}
+		defer leaderManager.Stop()
+
 		pool = tenant.NewPool(tenant.PoolConfig{
 			S3Dir:                        s3cfg.Dir,
 			PublicURL:                    publicBaseURL(addr),
@@ -245,24 +265,39 @@ func main() {
 			S3EncryptionPolicy:           s3cfg.EncryptionPolicy,
 			BackendOptions:               backendOptions,
 			DisableDatabaseAutoEmbedding: disableDatabaseAutoEmbedding,
+			LeaderChecker:                leaderManager,
 		}, enc)
 		defer pool.Close()
-	}
 
-	if pool != nil {
 		pool.SetMetaStore(store)
 
-		// Start the mutation log replay worker for central quota.
-		replayWorker := backend.StartMutationReplayWorker(tenant.NewMetaQuotaAdapter(store))
-		if replayWorker != nil {
-			defer replayWorker.Stop()
-		}
-
-		// Start the upload reservation expiry sweep worker.
-		expirySweepWorker := backend.StartExpirySweepWorker(store)
-		if expirySweepWorker != nil {
-			defer expirySweepWorker.Stop()
-		}
+		// Gate mutation replay and expiry sweep workers behind leadership.
+		var replayWorker *backend.MutationReplayWorker
+		var expirySweepWorker *backend.ExpirySweepWorker
+		leaderManager.SetCallbacks(
+			func() {
+				replayWorker = backend.StartMutationReplayWorker(tenant.NewMetaQuotaAdapter(store))
+				expirySweepWorker = backend.StartExpirySweepWorker(store)
+			},
+			func() {
+				if replayWorker != nil {
+					replayWorker.Stop()
+					replayWorker = nil
+				}
+				if expirySweepWorker != nil {
+					expirySweepWorker.Stop()
+					expirySweepWorker = nil
+				}
+			},
+		)
+		defer func() {
+			if replayWorker != nil {
+				replayWorker.Stop()
+			}
+			if expirySweepWorker != nil {
+				expirySweepWorker.Stop()
+			}
+		}()
 
 		// TODO: Run ValidateDurableAsyncExtractRequiresSemanticWorker only when this process
 		// can serve tenants that enqueue durable audio_extract_text / img_extract_text
@@ -319,6 +354,7 @@ func main() {
 		TiDBAutoEmbeddingAPIKey:      autoEmbeddingAPIKey,
 		TiDBAutoEmbeddingAPIBase:     autoEmbeddingAPIBase,
 		DisableDatabaseAutoEmbedding: disableDatabaseAutoEmbedding,
+		Leader:                       leaderManager,
 	}).ListenAndServe(addr))
 }
 
--- a/pkg/backend/file_gc_worker.go
+++ b/pkg/backend/file_gc_worker.go
@@ -65,7 +65,21 @@ func (b *Dat9Backend) StartFileGCWorker(opts FileGCWorkerOptions) *FileGCWorker 
 	return w
 }
 
-func (b *Dat9Backend) stopFileGCWorker() {
+// FileGCWorkerRunning reports whether the per-tenant file GC worker is currently
+// active. Intended for leadership-lifecycle tests and diagnostics.
+func (b *Dat9Backend) FileGCWorkerRunning() bool {
+	if b == nil {
+		return false
+	}
+	b.mu.Lock()
+	defer b.mu.Unlock()
+	return b.fileGCWorker != nil
+}
+
+// StopFileGCWorker stops the per-tenant file GC worker if it is running. Safe
+// to call when no worker is active. Used by the tenant pool's StopAllFileGC
+// (leader-loss path) and the backend's own Close.
+func (b *Dat9Backend) StopFileGCWorker() {
 	b.mu.Lock()
 	w := b.fileGCWorker
 	b.fileGCWorker = nil
--- a/pkg/backend/options.go
+++ b/pkg/backend/options.go
@@ -300,7 +300,7 @@ func (b *Dat9Backend) configureOptions(opts Options) {
 
 // Close stops background workers owned by this backend instance.
 func (b *Dat9Backend) Close() {
-	b.stopFileGCWorker()
+	b.StopFileGCWorker()
 	if b.imageExtractEnabled {
 		globalBackendRuntimeMetrics.deactivateImage(b.runtimeMetricsID)
 		b.imageExtractEnabled = false
--- a/pkg/leader/leader.go
+++ b/pkg/leader/leader.go
@@ -1,0 +1,345 @@
+// Package leader provides a TiDB/MySQL GET_LOCK-based leader election mechanism
+// so that background schedulers run only on the leader pod in a multi-pod
+// deployment. The leader holds a named session-level lock on the shared meta
+// database; connection death (pod crash / restart) auto-releases the lock,
+// allowing another pod to acquire it.
+package leader
+
+import (
+	"context"
+	"database/sql"
+	"sync"
+	"time"
+
+	"github.com/mem9-ai/drive9/pkg/logger"
+	"go.uber.org/zap"
+)
+
+const (
+	// defaultLockName is the named lock used for leader election.
+	defaultLockName = "drive9:leader"
+	// defaultHeartbeatInterval is how often the leader verifies it still holds
+	// the lock, and how often non-leaders retry acquisition.
+	defaultHeartbeatInterval = 10 * time.Second
+	// keepAliveSQL keeps the dedicated connection alive between heartbeats,
+	// preventing MySQL wait_timeout from closing it.
+	keepAliveSQL = "SELECT 1"
+	// getLockSQL acquires a named session lock. Timeout 0 means non-blocking:
+	// returns 1 if acquired, 0 if held by another connection.
+	getLockSQL = "SELECT GET_LOCK(?, 0)"
+	// isUsedLockSQL returns the connection ID holding the named lock, or NULL
+	// if no one holds it. Used by the leader to verify it still owns the lock.
+	isUsedLockSQL = "SELECT IS_USED_LOCK(?)"
+	// releaseLockSQL releases the named session lock.
+	releaseLockSQL = "SELECT RELEASE_LOCK(?)"
+)
+
+// LeaderChecker is a minimal interface for callers that only need to query
+// leadership status (e.g. the tenant pool gating per-tenant FileGCWorker).
+type LeaderChecker interface {
+	IsLeader() bool
+}
+
+// Manager acquires and maintains leadership via a TiDB/MySQL named lock.
+// When leadership is gained, OnLead is called; when lost, OnLose is called.
+// If disabled (single-pod mode), IsLeader always returns true and no lock
+// is acquired.
+type Manager struct {
+	db       *sql.DB
+	lockName string
+	interval time.Duration
+	disabled bool
+
+	mu       sync.Mutex
+	isLeader bool
+	onLead   func()
+	onLose   func()
+
+	cancel context.CancelFunc
+	done   chan struct{}
+}
+
+// Option configures a Manager.
+type Option func(*Manager)
+
+// WithLockName sets the named lock used for election.
+func WithLockName(name string) Option {
+	return func(m *Manager) {
+		if name != "" {
+			m.lockName = name
+		}
+	}
+}
+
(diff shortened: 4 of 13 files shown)


qiffang (changes requested):
REVISE / BLOCK.

I found two additional lifecycle blockers on top of adversary-2's review:

1. `SetCallbacks` is overwrite-only, so `drive9-server` never starts mutation replay / expiry sweep.

`cmd/drive9-server/main.go` registers leader callbacks to start/stop `MutationReplayWorker` and `ExpirySweepWorker`, but `pkg/server/server.go` later calls `s.leader.SetCallbacks(s.onLead, s.onLose)` before `Start`, replacing those callbacks. In the real server, quota mutation replay and upload reservation expiry sweep therefore do not start at all, including `DRIVE9_LEADER_DISABLED` mode. This breaks recovery/cleanup paths. Fix by composing the leader lifecycle (or supporting multi-subscriber callbacks) and add a regression proving all leader-gated workers start/stop: semantic, object GC, tenant cleanup/reconcile, mutation replay, and expiry sweep.

2. Per-tenant `FileGCWorker` is checked only at backend creation, not leadership transitions.

`pkg/tenant/pool.go` starts `FileGCWorker` only if `LeaderChecker.IsLeader()` is true while the tenant backend is created. If a backend is created while leader, its GC worker continues after leadership loss; if it is created while standby, GC never starts when the pod later becomes leader. That violates the stated leader-gated FileGC behavior. Wire tenant-pool GC into leadership gain/loss for existing backends, or remove FileGC from the claimed leader-gated scope, with regressions for leader→lose and standby→leader.

Local validation: `go test -c ./pkg/leader ./pkg/server ./pkg/tenant ./cmd/drive9-server`, `go test ./cmd/drive9-server ./cmd/drive9-server-local`, and `git diff --check origin/main...HEAD` passed. `go test ./pkg/leader -count=1` is locally blocked by missing rootless Docker/testcontainers, so CI remains the runtime DB gate.

srstack (commented):
## Review: feat/leader-election (P0-3)

**Overall: solid design, but there is one blocking bug that disables two production workers. Requesting one fix before merge.**

### What this does

Adds a `pkg/leader` package implementing TiDB/MySQL `GET_LOCK`-based leader election so background schedulers run only on the leader pod. The lock is session-scoped (auto-released on connection death), a heartbeat goroutine verifies ownership via `IS_USED_LOCK` and keeps the conn alive with `SELECT 1`, and workers (semantic, object GC, fork-cleanup group, mutation replay, expiry sweep, per-tenant FileGC) are gated on leadership. Single-pod / nil-manager / `DRIVE9_LEADER_DISABLED` paths preserve existing always-on behavior. The package itself is clean and well-tested.

---

### 🔴 Blocking: `SetCallbacks` is called twice — the second call silently clobbers the first

`SetCallbacks` overwrites `onLead`/`onLose` rather than composing them (`pkg/leader/leader.go:217-225`). There are two callers, and they race by construction order:

1. `cmd/drive9-server/main.go` calls `leaderManager.SetCallbacks(...)` to start/stop **`MutationReplayWorker`** and **`ExpirySweepWorker`**.
2. It then passes `Leader: leaderManager` into `server.NewWithConfig(...)`, which calls `s.leader.SetCallbacks(s.onLead, s.onLose)` and `Start()`.

Because `NewWithConfig` runs *after* main's `SetCallbacks`, the server's callbacks **overwrite** main's. Net effect: the mutation replay worker and the expiry sweep worker's `onLead`/`onLose` are discarded and **never fire** — in both leader-enabled *and* disabled mode (disabled mode invokes `onLead` synchronously inside `Start`, which is already the server's callback by then).

This means central-quota mutation replay and upload-reservation expiry sweep stop running on this binary entirely — a regression from the current `StartMutationReplayWorker`/`StartExpirySweepWorker` calls this PR removes.

**Fix options:**
- Move the replay/expiry start/stop into the server's `onLead`/`onLose` (e.g. pass the `store`/adapter into `Config` and start them in `startLeaderWorkers`), so there is a single callback owner; **or**
- Make the leader manager support multiple callbacks (append instead of overwrite), and have both registrations use it.

The first option is cleaner — one place owns leader-gated startup.

---

### 🟠 Concurrency: `stopLeaderWorkers` / `startLeaderWorkers` touch worker state outside the lock

`stopLeaderWorkers` grabs+nils `leaderWorkerCancel` under `leaderWorkerMu`, then calls `semanticWorker.Stop()` / `objectGCWorker.Stop()` outside the lock. `onLose` (heartbeat goroutine) and `Close()` (main goroutine) can both invoke this concurrently, and `semanticWorkerManager.Start/Stop` mutate `m.cancel` without their own lock. Two concurrent `Stop()`s can both pass the `m.cancel == nil` guard and race the read/write of `m.cancel`. Bounded and unlikely, but it is a genuine data race now that these are driven from two goroutines. Consider serializing all leader-worker start/stop under `leaderWorkerMu`, or guarding the worker managers' own `cancel` fields.

---

### 🟡 Split-brain window on leader handoff (worth a doc note)

Lock loss is only detected at the next heartbeat tick (`interval`, default 10s). If the leader's connection dies, there is up to ~`interval` of overlap where the old leader still believes it leads while a new pod acquires the lock. For object GC — the worker explicitly called out as having "no DB-level claim" — that window can still produce duplicate S3 HEAD/DELETE calls. The election bounds it well, but since GC is the motivating case, a comment acknowledging the bounded-overlap window (and/or a shorter heartbeat for GC criticality) would help future readers reason about it.

---

### Minor

- **No newline at EOF** in `pkg/leader/leader.go` and `pkg/leader/leader_test.go` (`\ No newline at end of file`). `gofmt` won't flag it but most of the repo ends files with a newline.
- **`main.go` parses `DRIVE9_LEADER_HEARTBEAT_INTERVAL` and silently ignores parse errors** (`if d, err := ...; err == nil`). A malformed value falls back to the default with no log line — a `logger.Warn` would save a confusing debugging session.
- **`WithCallbacks` option vs `SetCallbacks` method** both exist. Given the bug above, consider keeping only one mechanism to avoid the "which one wins" ambiguity.
- **`tryAcquire` non-acquired path** runs `RELEASE_LOCK` on a lock it never held (harmless no-op returning 0) plus an extra round-trip every retry. Not worth changing, just noting.

---

### Tests

`pkg/leader` coverage is good — acquire/lose/reacquire, exclusivity, disabled mode, and the raw `GET_LOCK`/`IS_USED_LOCK`/`RELEASE_LOCK` primitives against a real MySQL. The gap is integration-level: nothing exercises the `main.go` wiring, which is exactly where the blocking bug lives. A test asserting that replay + expiry workers actually start under leadership would have caught it.

mornyx (commented):
@qiffang Thanks for the review. Both blockers are fixed in 7cf571e:

**1. `SetCallbacks` overwrite — mutation replay / expiry sweep now start.**
You were right that the server's `SetCallbacks` clobbered main.go's callbacks, so the replay and expiry workers never fired. I took the single-owner route you and @srstack both flagged as cleaner: the server now owns ALL leader-gated worker start/stop. `startLeaderWorkers` starts the mutation replay + expiry sweep workers (built from `tenant.NewMetaQuotaAdapter(s.meta)` / `s.meta`) alongside semantic, object GC, tenant cleanup/reconcile, and per-tenant FileGC; `stopLeaderWorkers` stops all of them. main.go no longer registers a competing `SetCallbacks` pair — it just passes `Leader: leaderManager` into the server config. This works in both leader-enabled and `DRIVE9_LEADER_DISABLED` mode (disabled invokes `onLead` synchronously inside `Start`).

**2. Per-tenant FileGC now reacts to leadership transitions.** The one-time `shouldStartFileGC` check at backend creation only covered the case where this pod was already leader. Added `Pool.StartAllFileGC` / `Pool.StopAllFileGC`, which iterate the cached backends and start/stop each `FileGCWorker`. The server calls them from `onLead` / `onLose`, so a backend created while standby gets FileGC on the next leadership gain, and a backend created while leader stops FileGC on loss. `shouldStartFileGC` is kept for the already-leader-at-creation path so a freshly created backend isn't left without GC until the next transition.

**Regression.** `TestLeaderGatedWorkersStartAndStop` (pkg/server/leader_lifecycle_test.go) builds a server with a disabled leader manager (so `onLead` fires synchronously), acquires a tenant backend, and asserts every leader-gated worker is running after `onLead` and stopped after `onLose`: semantic, object GC, mutation replay, expiry sweep, and per-tenant FileGC. This is the integration test that would have caught the clobber bug.

I also addressed @srstack's concurrency note: `startLeaderWorkers` / `stopLeaderWorkers` are now fully serialized under `leaderWorkerMu` with a `leaderWorkersStarted` guard, so `onLose` (heartbeat goroutine) and `Close()` (main goroutine) cannot double-stop a worker or race on the worker managers' own unguarded `cancel` fields.

mornyx (commented):
@srstack Thanks for the thorough review. The blocking `SetCallbacks` clobber is fixed (see my reply to @qiffang above) — the server is now the single owner of all leader-gated worker start/stop, and main.go no longer registers a competing pair. Replies to your other points:

**Concurrency (stopLeaderWorkers/startLeaderWorkers outside the lock).** Fixed. The whole start and the whole stop are now serialized under `leaderWorkerMu` with a `leaderWorkersStarted` guard, so `onLose` (heartbeat goroutine) and `Close()` (main goroutine) cannot concurrently double-stop a worker or race on the semantic/objectGC worker managers' own unguarded `cancel` fields. `Close()` calls `stopLeaderWorkers()` first, then `s.leader.Stop()`, so the heartbeat goroutine is already drained before `Stop()` fires any final `onLose`.

**Split-brain window / CONNECTION_ID ownership.** Addressed. `tryAcquire` now captures `SELECT CONNECTION_ID()` on the same pinned conn right after `GET_LOCK`, stores it as `ownConnID`, and `holdLeadership` compares `IS_USED_LOCK`'s owner against it. A mismatch (another session acquired the lock) is logged as `leader_lock_lost` reason `lock_not_owned_by_this_connection` and releases leadership — so we no longer only catch the NULL/0 case. `ownConnID` is cleared on `loseLeadership` / `releaseLeadership`. I kept the heartbeat-based detection rather than shortening the interval for GC; the bounded-overlap comment you suggested is worth adding, but I left the default at 10s and left the interval configurable via `DRIVE9_LEADER_HEARTBEAT_INTERVAL` rather than special-casing GC criticality, which felt like a separate tuning decision.

**ctx ignored in Start / unused parentCtx.** Fixed. `workerCtx` is now derived from the supplied `ctx` (`context.WithCancel(ctx)`), so cancelling the parent context stops the election loop and trace metadata is preserved in the logs. The unused `parentCtx` parameter is dropped from `run`. Covered by `TestLeaderCtxCancellation`.

**Minor.** Both files already end with a newline (`git diff` no longer shows `\ No newline at end of file`); the heartbeat-interval parse failure now logs a `leader_heartbeat_interval_invalid_falling_back_to_default` warning with the bad value and parse error instead of silently falling back; and I clarified the `WithCallbacks` / `SetCallbacks` doc comments to state explicitly that the last pair registered before `Start` wins and that callers needing to compose concerns should own a single pair (which is what the server now does). The `tryAcquire` non-acquired `RELEASE_LOCK` no-op is left as-is per your note.

**Integration test gap.** Addressed — `TestLeaderGatedWorkersStartAndStop` exercises the `onLead`/`onLose` wiring at the server level (the exact spot the blocking bug lived) and proves semantic, object GC, mutation replay, expiry sweep, and per-tenant FileGC all start and stop on leadership transitions.

qiffang (changes requested):
R2: REVISE.

The main callback/ownership fixes are improved, but FileGC leader state still races with tenant backend creation.

`Pool.Get` / `Acquire` create backends outside `p.mu`, and `createBackend` calls `shouldStartFileGC()` before the backend is inserted into `p.items`.

Two concrete races remain:

1. Leader -> lose while backend is being created: `createBackend` sees `IsLeader() == true` and starts `FileGCWorker`; before insertion, `onLose()` calls `StopAllFileGC()` and snapshots only currently cached backends; the new backend is inserted after the snapshot and keeps running FileGC while this pod is no longer leader.

2. Standby -> leader while backend is being created: `createBackend` sees `IsLeader() == false` and skips FileGC; before insertion, `onLead()` calls `StartAllFileGC()` and snapshots only currently cached backends; the new backend is inserted after the snapshot and remains cached without FileGC even though this pod is now leader.

This is the same correctness class as the previous FileGC lifecycle issue, but during concurrent backend creation rather than already-cached backends. The new leader lifecycle test covers an already-cached backend, not creation racing with leadership transitions.

Fix expectation: FileGC start/stop should depend on pool-owned leader state synchronized with `p.mu`, not only a moment-in-time `LeaderChecker.IsLeader()` call inside `createBackend`. For example, keep a pool `fileGCEnabled` state updated by `StartAllFileGC` / `StopAllFileGC` under `p.mu`, and when inserting a newly created backend, start FileGC according to that state after the backend is in the pool. Add regressions for backend creation racing with leader gain and leader loss.

Local validation on `7cf571e`: `go test -c ./pkg/leader ./pkg/server ./pkg/tenant ./cmd/drive9-server`, `go test -c ./pkg/... ./cmd/...`, `go test ./cmd/drive9-server ./cmd/drive9-server-local`, and `git diff --check origin/main...HEAD` passed. Runtime MySQL-backed tests are locally blocked by missing rootless Docker/testcontainers.

mornyx (commented):
@qiffang Good catch — that race is real, fixed in 1fd7bb8.

You're right that `shouldStartFileGC()` in `createBackend` was a moment-in-time `IsLeader()` check taken outside `p.mu`, before the backend was inserted, so a leadership transition during creation landed the backend with the wrong FileGC state (leader→lose kept FileGC running; standby→leader left it without FileGC).

I took the approach you suggested. The pool now owns a `fileGCEnabled bool` set under `p.mu`:
- `StartAllFileGC` (leader gain) sets `fileGCEnabled=true` under `p.mu` and starts FileGC on every cached backend.
- `StopAllFileGC` (leader loss) sets `fileGCEnabled=false` under `p.mu` and stops FileGC on every cached backend.
- FileGC start is moved out of `createBackend` entirely and into the `Get`/`Acquire` insertion critical section: right before the new backend is inserted, `fileGCEnabled` is read under `p.mu` and FileGC is started accordingly. So a backend created concurrently with a transition resolves to the post-transition state — the transition and the insertion are serialized on the same lock.
- `NewPool` initializes `fileGCEnabled=true` when no `LeaderChecker` is configured (single-pod mode), so the nil-checker path is unaffected.

The moment-in-time `shouldStartFileGC()` helper is removed entirely; `fileGCEnabled` (pool-owned, locked) is the single source of truth for both cached backends and newly created ones.

Regressions (pkg/tenant/pool_filegc_leader_test.go):
- `TestPoolFileGCStartsOnLeaderGainDuringBackendCreation`: pool starts standby, `StartAllFileGC` fires while the backend is being created (pool empty, so it only flips the flag — the transition wins the race), then `Get` inserts the backend and FileGC must be running even though the fake checker still reports standby (proving the outcome is driven by `fileGCEnabled`, not the snapshot).
- `TestPoolFileGCSkipsOnLeaderLossDuringBackendCreation`: pool starts leader, `StopAllFileGC` fires while the backend is being created (flips the flag to false), then `Get` inserts the backend and FileGC must NOT be running; a subsequent `StartAllFileGC` starts it on the cached backend.
- `TestPoolFileGCNoLeaderCheckerAlwaysEnabled`: nil LeaderChecker (single-pod) keeps `fileGCEnabled=true` and every backend starts FileGC.

These are deterministic rather than timing-based: the transition is applied while the pool is empty (so it only sets the flag), then the insert observes the flag under the lock — exactly the post-transition ordering that the old snapshot-based check got wrong. Build, vet, and lint clean; runtime MySQL-backed tests run in CI.

qiffang (changes requested):
R3: REVISE.

The R2 FileGC backend-creation race is fixed correctly, but there is still a server-level lifecycle race in the leader worker start/stop path.

`startLeaderWorkers` / `stopLeaderWorkers` are not actually serialized. `startLeaderWorkers` sets `leaderWorkersStarted=true` and unlocks before doing most of the real work: assigning `s.replayWorker` / `s.expirySweepWorker`, calling `Pool.StartAllFileGC`, and starting semantic/object GC. `stopLeaderWorkers` can concurrently acquire the mutex, set `leaderWorkersStarted=false`, snapshot nil/old worker fields, unlock, and stop that snapshot.

Concrete failure:

1. Leader callback enters `startLeaderWorkers`, sets `leaderWorkersStarted=true`, unlocks.
2. Server shutdown / close calls `stopLeaderWorkers`, sees started=true, snapshots `s.replayWorker == nil`, sets started=false, and returns after stopping nothing.
3. The original `startLeaderWorkers` resumes and starts replay/expiry/FileGC/semantic/objectGC after the stop path already ran.
4. `leader.Stop()` later invokes `onLose`, but `leaderWorkersStarted` is already false, so `stopLeaderWorkers` returns without stopping those late-started workers.

This can happen in the real server because `Close()` currently calls `s.stopLeaderWorkers()` before `s.leader.Stop()`, so the leader heartbeat goroutine may still be inside `onLead` while shutdown is trying to stop workers.

This is also a data race: `startLeaderWorkers` writes `s.replayWorker` / `s.expirySweepWorker` outside `leaderWorkerMu`, while `stopLeaderWorkers` reads and clears them under `leaderWorkerMu`.

Fix expectation: make leader worker lifecycle truly single-threaded / non-overlapping. Either stop the leader manager before local worker teardown and wait for callbacks to finish, or protect the full start/stop state transition with a proper generation/state machine. Add a regression that races `onLead` with `Close`/`onLose` and asserts no replay/expiry/FileGC/semantic/objectGC worker remains running after shutdown/loss.

Also, the no-leader path still double-starts semantic/object GC: `NewWithConfig` calls `startLeaderWorkers()` and then `logSemanticWorkerStatus(...)`, but `logSemanticWorkerStatus` still calls `Start` on both workers. It should be log-only.

Local validation on `1fd7bb8`: compile checks for `./pkg/... ./cmd/...`, `go test ./cmd/drive9-server ./cmd/drive9-server-local`, and `git diff --check origin/main...HEAD` passed. Runtime testcontainers-backed package tests are locally blocked by missing rootless Docker; GitHub CI is green.

mornyx (commented):
@qiffang Both R3 findings are fixed in b23a9d0.

**1. Server-level lifecycle race in startLeaderWorkers/stopLeaderWorkers.**
You're right that `startLeaderWorkers` flipped `leaderWorkersStarted=true` and unlocked before assigning `s.replayWorker`/`s.expirySweepWorker`, starting FileGC/semantic/objectGC, etc. A concurrent `stopLeaderWorkers` (from `Close` or `onLose`) could snapshot nil/stale worker fields, stop nothing, and let the in-flight start finish after the stop — leaving orphan workers that a later `leader.Stop()`→`onLose` would not stop (started already false). This was also a data race on `s.replayWorker`/`s.expirySweepWorker` (written outside the lock, read/cleared under it).

Fix: `leaderWorkerMu` is now held for the *entire* `startLeaderWorkers` and `stopLeaderWorkers` transitions, including the worker Start/Stop/wait calls. Start and stop are fully non-overlapping: whichever wins the lock runs to completion first, and the other observes the post-transition state — a concurrent stop sees the fully-started worker set (not a partial one) and stops them all; a concurrent start sees `leaderWorkersStarted=false` and starts fresh. `Close()` is reordered to stop the leader manager first (`s.leader.Stop()` waits for the heartbeat goroutine to exit, so no further `onLead`/`onLose` can fire after it returns) before `stopLeaderWorkers()` tears down local workers. In leader mode `onLose` already stopped them via `leader.Stop()`, and the `stopLeaderWorkers()` call in `Close` is a guarded no-op. In single-pod mode `stopLeaderWorkers()` stops the unconditionally-started workers. This removes the race where `Close()` stops workers while `onLead` is still starting them — the callbacks and the local teardown are serialized on `leaderWorkerMu`, and no callback can outlive `leader.Stop()`.

**2. No-leader path double-starts semantic/object GC.** `NewWithConfig` called `startLeaderWorkers()` and then `logSemanticWorkerStatus(...)`, which still called `Start` on both workers (redundant double-start). `logSemanticWorkerStatus` is now log-only — the workers are started by `startLeaderWorkers`.

**Regressions** (pkg/server/leader_lifecycle_test.go):
- `TestLeaderGatedWorkersCloseRacesOnLead`: races `onLead` against `Close` in a loop (50 iters) and asserts no leader-gated worker remains running after `Close` (replay/expiry nil, semantic/objectGC stopped, FileGC stopped). Run with `-race` to catch the prior data race on `s.replayWorker`/`s.expirySweepWorker`.
- `TestLeaderGatedWorkersOnLoseRacesOnLead`: races `onLead` against `onLose` and asserts a final `onLose` leaves no worker running, regardless of which racing call won the lock first.
- The existing `TestLeaderGatedWorkersStartAndStop` was refactored to share a `newLeaderLifecycleServer` helper with the new tests.

Build, vet, and lint clean; runtime MySQL-backed tests run in CI.

qiffang (approved):
R4 review on 4e442f4: approved.

I re-reviewed the latest lifecycle changes and the previous blockers are resolved:
- leader Stop now runs before local worker teardown, so callbacks cannot outlive Server.Close
- start/stop leader-gated workers are serialized for the full transition under leaderWorkerMu
- no-leader semantic/object GC path is now log-only after startLeaderWorkers, avoiding double-start
- FileGC leader state is pool-owned and read under the pool lock during backend insertion, covering backend creation racing with leadership transitions
- central quota replay/expiry workers are owned by the server's single leader callback path

Validation: local compile gates and cmd package tests passed; focused DB-backed package tests are still blocked on this machine by testcontainers/rootless Docker, but GitHub code-ci and local-e2e are green on this head.

PR #603, opened 2026-06-23
feat: add cloud native quota controls

srstack:
## Summary
- add cloud native quota get/set APIs and CLI commands gated by TiDB Cloud API keys
- attach drive9-managed labels during starter cluster provisioning and quota updates
- move server quota small-write mutations to a tenant-local quota outbox to remove central quota writes from the request hot path
- include pending outbox deltas in quota checks and upload reservations for multi-server correctness

## Tests
- GOCACHE=/tmp/drive9-go-cache go test ./pkg/backend
- GOCACHE=/tmp/drive9-go-cache go test ./pkg/datastore
- GOCACHE=/tmp/drive9-go-cache go test ./pkg/server
- GOCACHE=/tmp/drive9-go-cache go test ./pkg/meta
- GOCACHE=/tmp/drive9-go-cache go test ./pkg/tenant/schema
- GOCACHE=/tmp/drive9-go-cache go test ./pkg/tenant/schemadump
- GOCACHE=/tmp/drive9-go-cache go test ./pkg/client
- GOCACHE=/tmp/drive9-go-cache go test ./cmd/drive9
- GOCACHE=/tmp/drive9-go-cache go test ./cmd/drive9/cli
- git diff --check

## Release Notes

* **New Features**
  * Added `drive9 quota` CLI with `get` / `set`, including `--json` output and TiDB Cloud credential-based quota management.
  * Added quota REST endpoints: `GET /v1/quota` and `POST /v1/quota`.
  * Introduced server-side quota mutation via an outbox worker for asynchronous, pending-aware enforcement.

* **Improvements**
  * Enhanced quota enforcement for server mode, including safer fail-open behavior and stricter validation rules.
  * Updated CLI and server help text plus added a dedicated quota CLI/API guide.

* **Tests**
  * Expanded CLI, API, backend, and client tests for quota/outbox behavior, routing, validation, and error handling.


mornyx (approved):
Reviewed PR #603 at `303f1bcd9bfd76c83cd1a27cc710fcb5da47a2ae`. I did not find any critical/functional blockers. I focused on the new cloud-native quota paths: quota config/query/set APIs, tenant-local quota outbox ordering/retry/lease handling, server-quota enforcement with pending outbox deltas, central mutation replay/file-GC interactions, schema inclusion, and the TiDB Cloud credential authorization/update flow. GitHub CI is green. Locally, the non-Docker tests I could run passed (`go test ./pkg/tenant/tidbcloudnative ./cmd/drive9/cli ./cmd/drive9 -count=1`). The backend/datastore/meta/server tenant tests require the MySQL testcontainer and could not be rerun in this environment because Docker is unavailable, so I relied on the passing CI for those.

qiffang (changes requested):
Reviewed PR #603 at `303f1bcd9bfd76c83cd1a27cc710fcb5da47a2ae`.

Verdict: **REQUEST_CHANGES**. The outbox direction is better than the lossy channel-only approach, but the current patch still has correctness and migration gaps.

Findings:

1. **P1: outbox retries can reorder same-file mutations and regress central state.** `ClaimQuotaOutbox` only blocks active `processing` rows and selects the first `queued` row whose `available_at <= now` (`pkg/datastore/quota_outbox.go:82-91`). `RetryQuotaOutbox` moves a failed older row back to `queued` with a future `available_at` (`pkg/datastore/quota_outbox.go:188-191`). A later row for the same file can therefore be claimed and applied before the older retry. Because `applyCentralFileStateTx` is last-state apply against current central metadata (`pkg/backend/quota_mutation.go:142-180`), the delayed older row can later overwrite central file metadata and reverse storage/media counters. Example: create/overwrite A fails and is delayed, overwrite B succeeds, then A retries and restores the stale size. This breaks the per-tenant ordering guarantee claimed by the comment above `ClaimQuotaOutbox`.

2. **P1: existing `tidb_cloud_native` tenants are not migrated to the new `quota_outbox` table.** The new table is appended to the TiDB auto schema (`pkg/tenant/schema/tidb_auto.go:883`), but tenant open-time schema ensure only runs for `tidb_zero` and `tidb_cloud_starter` (`pkg/tenant/pool.go:587-629`), not `tidb_cloud_native`. When `DRIVE9_QUOTA_SOURCE=server` is enabled, native tenants enqueue outbox rows inside small-write transactions (`pkg/backend/dat9.go:905`, `pkg/backend/quota_outbox_worker.go:196-203`). Any pre-existing native tenant whose schema predates this PR will fail writes with a missing `quota_outbox` table instead of being upgraded.

3. **P1: pending-delta checks are not concurrency-safe, so the PR overclaims multi-server correctness.** `ensureStorageQuotaServer` reads central usage plus `PendingQuotaOutboxDeltasTx` and then checks `projected` (`pkg/backend/quota.go:144-155`), while `PendingQuotaOutboxDeltasTx` is only a plain SUM over committed queued/processing rows (`pkg/datastore/quota_outbox.go:264-271`). The current transaction's outbox row is inserted later, and concurrent transactions do not see each other's uncommitted rows. With max=10 and two concurrent 8-byte writes, both can read central=0/pending=0, both pass, then both enqueue/commit; pending becomes 16. Either serialize quota admission in the tenant DB, reserve centrally, or explicitly document this as soft quota and remove the "multi-server correctness" claim/tests expectation.

4. **P2: server-quota media LLM checks miss the current media file.** `enqueueTiDBAutoSemanticTasksTx` checks `mediaLLMQuotaExceededCheckTx` before `enqueueQuotaFileCreateOutboxTx` is called for the just-written file (`pkg/backend/semantic_tasks.go:44-53`, `pkg/backend/dat9.go:892-905`). The server path only checks central `MediaFileCount + pendingMediaDelta` (`pkg/backend/quota.go:244-266`), so the current image/audio file is not included. With limit=N and N media files already counted, the N+1 file still queues extraction. The legacy tenant-DB `>` logic only worked because the current file was already visible in the same transaction.

5. **P2: partial quota updates can clobber each other.** `handleQuotaSet` reads the current config, copies all fields, patches requested fields, and then writes the full row (`pkg/server/quota.go:152-181`). Two concurrent partial updates to different fields can each preserve stale values for the other field; last writer silently loses the first update. Preservation of omitted fields needs to happen atomically in the DB update/transaction.

6. **P2: ambient TiDB Cloud env keys can hijack owner `quota get`.** `quotaGet` consumes `DRIVE9_PUBLIC_KEY`/`DRIVE9_PRIVATE_KEY` and switches to `/v1/quota/query` whenever either key is present, even when `--tenant-id` was not supplied (`cmd/drive9/cli/quota.go:127-150`). A normal owner-key `drive9 quota get` in a shell that has TiDB Cloud env vars now fails with `--tenant-id is required` or routes to the credential endpoint unexpectedly. Use explicit tenant-id/credential mode as the discriminator.

7. **P2: documented zero monthly-cost override is ignored.** The CLI says `--max-monthly-cost-mc 0` disables the monthly budget (`cmd/drive9/cli/quota.go:425-427`), and meta treats zero as disabled, but `monthlyLLMCostExceededServer` only applies per-tenant config when `cfg.MaxMonthlyCostMC > 0`; otherwise it falls back to the backend/global default (`pkg/backend/quota.go:93-99`). A tenant explicitly set to zero remains capped if the global default is positive.

Validation: `git diff --check origin/main...HEAD` passed locally. This environment has no `go` binary, so I could not rerun Go tests locally; GitHub `code-ci` and `local-e2e` are green at the reviewed SHA.

srstack:
@qiffang All 7 findings have been addressed. The code fixes are mainly in `015b4f50 fix: harden quota outbox admission`, and the CLI/API usage documentation was added in `88690237 docs: add quota usage guide`.

1. **P1: outbox retries can reorder same-file mutations and regress central state**

   Fixed. `ClaimQuotaOutbox` now prevents a later queued/processing row from bypassing any older queued/processing row, even when the older row has a future `available_at`. It also keeps the active-processing guard and uses `FOR UPDATE SKIP LOCKED`. This preserves tenant-local replay order after retries and prevents older file state from being applied after a newer mutation.

   Main changes: `pkg/datastore/quota_outbox.go`
   Test coverage: `TestQuotaOutboxClaimWaitsForDelayedOlderRetry`

2. **P1: existing `tidb_cloud_native` tenants are not migrated to the new `quota_outbox` table**

   Fixed. `Pool.createBackend` now includes `ProviderTiDBCloudNative` in the TiDB auto schema ensure path. Existing cloud-native tenants will get `quota_outbox` and `quota_admission_locks` created on backend open, so enabling server quota will not break small writes on older native tenants.

   Main changes: `pkg/tenant/pool.go`
   Test coverage: `TestPoolCreateBackendEnsuresSchemaForTiDBCloudNative`

3. **P1: pending-delta checks are not concurrency-safe**

   Fixed. Added a tenant-local `quota_admission_locks` table. In server quota mode, storage admission, pending outbox delta reads, and quota outbox enqueue now run under the same tenant DB transaction lock. Upload reservation now also takes the admission lock, rechecks duplicate reservations inside the lock, reads pending deltas, and then performs the central reservation. This serializes quota admission across multiple server instances.

   Main changes: `pkg/tenant/schema/quota_outbox.go`, `pkg/datastore/quota_outbox.go`, `pkg/backend/quota.go`, `pkg/backend/upload_reservation.go`, `pkg/backend/dat9.go`, `pkg/backend/fs_layer_content.go`
   Test coverage: `TestQuotaAdmissionLockCreatesStableLockRow`, `TestServerQuotaPendingOutboxDeltaRejectsOverLimitWrite`, `TestServerQuotaPendingOutboxDeltaRejectsUploadReserve`

4. **P2: server-quota media LLM checks miss the current media file**

   Fixed. `enqueueTiDBAutoSemanticTasksTx` now receives the media delta caused by the current write. The server-quota transactional media check uses `central MediaFileCount + pendingMediaDelta + currentMediaDelta`. The tenant-DB legacy path keeps the existing in-transaction counting semantics.

   Main changes: `pkg/backend/quota.go`, `pkg/backend/semantic_tasks.go`, `pkg/backend/dat9.go`, `pkg/backend/upload.go`
   Test coverage: `TestServerQuotaMediaCheckIncludesCurrentWrite`

5. **P2: partial quota updates can clobber each other**

   Fixed. The "preserve omitted fields" logic was moved into the meta store as a single atomic upsert. `handleQuotaSet` now calls `PatchQuotaConfig` / `QuotaConfigPatch` instead of reading the current config and writing back a full-row snapshot. Concurrent partial updates to different fields no longer overwrite each other with stale values.

   Main changes: `pkg/meta/quota.go`, `pkg/server/quota.go`
   Test coverage: `TestPatchQuotaConfigPreservesUnspecifiedFields`, `TestPatchQuotaConfigInsertsDefaultsForUnspecifiedFields`

6. **P2: ambient TiDB Cloud env keys can hijack owner `quota get`**

   Fixed. `drive9 quota get` enters TiDB Cloud credential query mode only when `--tenant-id` is explicitly provided. Merely having `DRIVE9_PUBLIC_KEY` / `DRIVE9_PRIVATE_KEY` in the environment no longer changes owner-key `/v1/quota` behavior. Passing explicit TiDB Cloud credential flags without `--tenant-id` now returns a clear error.

   Main changes: `cmd/drive9/cli/quota.go`
   Test coverage: `TestQuotaGetIgnoresEnvTiDBCloudCredentialsWithoutTenantID`, `TestQuotaGetRejectsTiDBCloudCredentialFlagsWithoutTenantID`

7. **P2: documented zero monthly-cost override is ignored**

   Fixed. Quota config now distinguishes "no tenant config row, use defaults" from "explicit tenant config row exists". Server monthly cost checks now use the tenant config value whenever an explicit config row exists, including `MaxMonthlyCostMC=0`. Therefore `--max-monthly-cost-mc 0` really disables the monthly LLM budget instead of falling back to the global default.

   Main changes: `pkg/meta/quota.go`, `pkg/backend/quota.go`, `pkg/backend/llm_usage.go`, `pkg/tenant/quota_adapter.go`
   Test coverage: `TestMonthlyLLMCostExceeded_ServerQuotaExplicitZeroDisablesDefault`

Validation run:

```bash
go test ./pkg/backend -run 'TestServerQuota|TestMonthlyLLMCostExceeded|TestApplyCentralFileMutationIsIdempotent'
go test ./pkg/datastore -run 'TestQuota'
go test ./pkg/meta -run 'Test.*Quota|TestAtomicReserveAndInsertUploadBootstrapsUsageRow'
go test ./pkg/server -run 'TestQuota'
go test ./pkg/tenant -run 'TestPoolCreateBackendEnsuresSchemaForTiDBCloudNative'
go test ./pkg/tenant/schema
go test ./cmd/drive9/cli
go test ./pkg/tenant/tidbcloudnative
make lint
git diff --check
```

pkg/backend/quota_outbox_worker.go
```diff
@@ -0,0 +1,238 @@
...
+	if err != nil {
+		metrics.RecordOperation("quota_outbox", "claim", "error", time.Since(start))
+		return false, err
+	}
+	if !found {
+		metrics.RecordOperation("quota_outbox", "claim", "empty", time.Since(start))
+		return false, nil
+	}
+
+	err = b.applyQuotaOutboxEntry(ctx, entry)
```
mornyx: This worker also needs to participate in the quota admission serialization. The new admission path takes the tenant-local `quota_admission_locks` lock, then reads central usage and the tenant-local pending outbox deltas. However this worker can concurrently apply the outbox row to central usage and ack it to `succeeded` without holding that same lock. If that happens between the admission path's central-usage read and pending-outbox read, admission observes the old central usage plus `pending=0`, so it can undercount and admit writes/uploads over the storage/media limit. Please hold the same admission lock across apply+ack (or otherwise make the usage+pending snapshot atomic with respect to outbox convergence).

pkg/meta/quota.go
```diff
@@ -121,6 +152,55 @@ func (s *Store) SetQuotaConfig(ctx context.Context, cfg *QuotaConfig) error {
...
+	start := time.Now()
+	var err error
+	defer observeMeta(ctx, "patch_quota_config", start, &err)
+
+	if patch == nil {
+		return fmt.Errorf("quota config patch is required")
+	}
+	storage := DefaultMaxStorageBytes()
+	media := DefaultMaxMediaLLMFiles
+	monthly := int64(0)
```
mornyx: For a first-time partial quota update, this inserts `max_monthly_cost_mc=0` even when the caller only set storage or media limits. Since `monthlyLLMCostExceededServer` now treats any explicit config row as authoritative, that explicit zero disables the backend's default monthly LLM cost cap. In practice, `drive9 quota set --max-storage-bytes ...` for a tenant with no existing config silently opts the tenant out of the default $10 defense-in-depth spend cap. Please preserve the default/fallback semantics for omitted monthly cost, or add a distinct explicit-unset representation so storage/media-only updates do not disable the cost guard.

mornyx (changes requested):
Re-reviewed PR #603 at `0bdea05bf9c1fd6961cec125edb0a81b1938a0bf`. I found two functional blockers in the latest quota hardening changes. The non-Docker local tests I could run passed (`go test ./pkg/tenant/tidbcloudnative ./cmd/drive9/cli ./cmd/drive9 -count=1`), and the MySQL-backed package tests still need CI/testcontainers coverage.

qiffang (changes requested):
R2 verdict: REQUEST CHANGES.

Most of the R1 issues are materially improved: CLI credential fallback, explicit monthly-cost zero semantics, atomic partial quota updates, media-current-write accounting, pending-delta serialization, upload reservation admission locking, and native auto-embedding schema ensure are all moving in the right direction.

Remaining blockers:

1. `code-ci` is red on the new quota outbox regression test. Head `0bdea05b` failed `pkg/datastore.TestQuotaOutboxClaimWaitsForDelayedOlderRetry` with `claimed = <nil> found=false, want first id 3` at `pkg/datastore/quota_outbox_test.go:137`. This PR changes quota admission/outbox ordering; a failing test in that exact area is not mergeable. I could not rerun Go locally because this environment has no `go` binary, so I’m relying on the GitHub Actions log.

2. `quota_outbox` poison rows can block a tenant forever. `defaultQuotaOutboxMaxAttempts` is still `0`, normal enqueues inherit that value, and `RetryQuotaOutbox` only dead-letters when `MaxAttempts > 0 && AttemptCount >= MaxAttempts`. With the new claim ordering barrier that waits behind any older queued/processing row, one permanently failing oldest row is retried forever and prevents later quota mutations from being claimed; its pending delta also remains in quota admission checks. Set a finite production default or add another explicit poison-row recovery path, and cover the unblock/dead-letter behavior with a regression test.

One nuance on the prior schema-migration concern: broad `tidb_cloud_native` auto-embedding tenants look mostly addressed because the target schema version is hash-derived from the statement set and `ProviderTiDBCloudNative` now participates in the open-time ensure path. I would still verify the `DRIVE9_DISABLE_AUTO_EMBEDDING=true` path separately if that is a supported server mode, because `pool.createBackend` currently gates open-time ensure on `opts.DatabaseAutoEmbedding` even though the server config comment says TiDB tenants still use the normal auto schema when DB auto embedding is disabled.

srstack:
@qiffang Addressed the R2 blockers in the latest head.

1. The `code-ci` failure on `pkg/datastore.TestQuotaOutboxClaimWaitsForDelayedOlderRetry` was from reviewed head `0bdea05b`. The ordering test was fixed in later commits; `ci` was green on `72f109c3` before the latest push. I also reran the datastore outbox tests locally against MySQL:

   `GOCACHE=/tmp/drive9-go-cache go test ./pkg/datastore -run 'TestQuotaOutbox|TestQuotaAdmissionLock' -count=1 -v`

2. Poison quota outbox rows no longer block a tenant forever. Normal enqueues now inherit a finite `defaultQuotaOutboxMaxAttempts` of `100`, and the TiDB/db9 schema defaults for `quota_outbox.max_attempts` were updated to the same value. Once an oldest row reaches the limit, `RetryQuotaOutbox` dead-letters it, removing it from pending quota deltas and allowing later queued rows to be claimed.

   Added regression coverage:
   - `TestQuotaOutboxLifecycle` now asserts normal enqueues get the finite default.
   - `TestQuotaOutboxDeadLetteredOlderRowUnblocksLaterClaim` verifies an older dead-lettered row no longer blocks a later row and that pending deltas only include the later queued row.

Additional validation I ran:

- `GOCACHE=/tmp/drive9-go-cache go test ./pkg/backend -run 'TestQuotaOutboxWorker|TestServerQuotaSmallWriteUsesTenantOutbox' -count=1 -v`
- `GOCACHE=/tmp/drive9-go-cache go test ./pkg/tenant/schema ./pkg/tenant/db9`
- `git diff --check`
- `make lint`

New GitHub checks are running on `d36e140a`.

qiffang (approved):
R3 review: GREEN.

I re-reviewed latest head `2eca83dfd0db2a0610b82b7a8d70b97d5a5e034d` after the follow-up fixes.

The two R2 blockers are resolved:
- CI is green on the latest head; the delayed older retry claim test was rewritten and now passes in GitHub CI.
- Poisoned `quota_outbox` rows no longer block a tenant forever: `defaultQuotaOutboxMaxAttempts` is finite, retry can dead-letter, dead-lettered rows are excluded from pending deltas, and regression coverage proves a later row becomes claimable after the older poison row is dead-lettered.

I also rechecked the higher-risk paths around create/overwrite, fs-layer writes, upload reservations/completion, delete/GC, media quota, pending-delta admission, and TiDB Cloud Native schema ensure. I do not see a remaining merge blocker.

Residual non-blocking risks:
- The config cache version still derives from `updated_at`; two quota config updates inside one DB timestamp precision tick could theoretically miss an invalidation. This is a low-frequency config-path hardening item, not a blocker for this PR.
- I could not run local Go tests because this environment has no `go` binary; I relied on static review plus the green GitHub `ci` and local-e2e checks.

srstack (commented):
## Review: feat/quota-cloud-native-outbox

**Overall: well-structured PR with strong test coverage. Two correctness items to verify before merge — one migration-safety, one outbox ordering — plus minor notes.**

### What this does

Three related changes:
1. **Cloud-native quota CLI + HTTP API** (`drive9 quota get|set`, `/v1/quota`, `/v1/quota/query`) — owner-key reads, TiDB Cloud credential-gated writes, with drive9-managed cluster labels.
2. **Quota outbox**: moves server-quota *small-write* mutations off the request hot path. Instead of a synchronous central-DB write per fsync, a `quota_outbox` row is written in the same tenant-DB transaction as the file mutation and drained asynchronously by a per-backend worker.
3. **Pending-delta awareness**: quota checks and upload reservations now add un-drained outbox deltas to the projected usage, preserving multi-server correctness.

The transactional-outbox pattern is the right call here — it makes the quota mutation atomic with the file write (no lost events on crash) while removing the ~37ms central round-trip from the fsync path. The `applyCentralFileStateTx` shadow-diff design makes individual mutations idempotent, which is necessary for at-least-once outbox delivery, and `TestApplyCentralFileMutationIsIdempotent` covers it.

---

### 🔴 Verify: schema migration for existing app-mode / db9 tenants

`quota_outbox` is appended to every schema statement list (`tidb_auto.go`, `tidb_app.go`, `mysql_no_embedding.go`, `db9/schema.go`), so **new** tenants are fine. For **existing** tenants, the re-ensure path matters:

- **Auto-embedding tenants**: schema version is content-derived (`currentTiDBTenantSchemaVersion` CRC32 over CREATE TABLE/INDEX). Appending `quota_outbox` bumps the version → `pool.go:603` detects the mismatch and re-ensures. ✅
- **App-mode / `DisableDatabaseAutoEmbedding` / db9 tenants**: the entire schema-ensure block (`pool.go:587`) is gated on `opts.DatabaseAutoEmbedding`, so it is **skipped**. These tenants never get `quota_outbox` retroactively.

The risk: a pre-existing tenant running with `DRIVE9_QUOTA_SOURCE=server` in app-mode will hit `enqueueQuotaFileCreateOutboxTx` → `INSERT INTO quota_outbox` on every create/overwrite, and the transaction will fail if the table doesn't exist — i.e. **writes break**. This is exactly the cloud-native population from #586. Please confirm one of: (a) there are no existing server-quota app-mode tenants, (b) a backfill/migration creates `quota_outbox` for them, or (c) `CREATE TABLE IF NOT EXISTS` is run on backend open for these providers.

---

### 🟠 Outbox apply is not strictly ordered under retry (create/overwrite are order-sensitive)

`ClaimQuotaOutbox` enforces single-flight (`NOT EXISTS (... processing ...)`) and `ORDER BY id`, so under normal operation mutations apply in enqueue order. But when a row enters retry/backoff (`available_at` in the future), it becomes ineligible while **newer** rows for the same file remain claimable — so a later mutation can apply before an earlier one.

Because `applyCentralFileStateTx` diffs against stored shadow state, reordering create-then-overwrite for the same file converges to the **wrong** final value:

```
correct order:  create(10): shadow=10, storage+=10  →  overwrite(20): shadow=20, storage+=10  ⟹ storage=20 ✅
reordered:      overwrite(20): shadow=20, storage+=20 →  create(10): shadow=10, storage+=(10-20) ⟹ storage=10 ❌
```

Final shadow and counter end at 10 instead of 20. Each mutation is individually idempotent, but the pair is order-sensitive. This requires the older row to fail at least once (transient central-DB error) while a newer same-file row is queued — uncommon, but not impossible, and `defaultQuotaOutboxMaxAttempts = 0` means rows retry **forever** (the `MaxAttempts > 0` dead-letter branch is never taken), widening the window. Consider either preserving strict per-file ordering (don't claim a row while an older queued/processing row exists for the same `file_id`) or making apply order-insensitive (e.g. last-writer-wins keyed by revision).

---

### Minor notes

- **`MaxAttempts = 0` ⇒ no dead-lettering**: a genuinely poison row retries indefinitely, logging `quota_outbox_process_failed` every ≤30s with no escape hatch and no alert signal. Consider a non-zero default or a dead-letter threshold so poison rows surface instead of looping silently.
- **`finalizeUpload` keeps the reservation + mutation-log path** (via `completeUploadReservation`) rather than the outbox, while small writes use the outbox. The split is internally consistent (reserved_bytes vs pending outbox deltas don't overlap), but it's now two convergence mechanisms for one counter — worth a comment at the `finalizeUpload` quota site explaining why uploads intentionally bypass the outbox.
- **`UpdateQuotaWithCredentials` (external PATCH) runs before `SetQuotaConfig` (local)**: if the local write fails after the cluster-label PATCH succeeds, the drive9.ai/* labels drift ahead of the actual enforced config. Labels are audit metadata only, so impact is low, but a log line on the post-PATCH local-write failure would help diagnose drift.
- **Pending-delta fail-open**: `pendingQuotaOutboxDeltas{,Tx}` returns `ok=false` and treats pending as 0 on DB error, so a transient error briefly disables pending enforcement. Consistent with the documented fail-open philosophy; just flagging it's an intentional soft spot.
- **`AtomicReserveAndInsertUpload` now bootstraps the usage row** (`INSERT IGNORE`) — good fix, and `TestAtomicReserveAndInsertUploadBootstrapsUsageRow` covers it.

---

### Tests

Coverage is strong and deterministic: the outbox tests stop the worker (`b.stopQuotaOutboxWorker()`) and drive `ProcessOneQuotaOutbox` directly, the datastore lifecycle/recover tests are real-MySQL, the server handler tests cover owner-get / credential-query / set / non-cloud-native rejection / error mapping, and the provisioner tests verify label merge + digest auth + forbidden/not-found mapping. The gaps map to the two items above: no test asserts existing-app-mode-tenant schema presence, and none exercises out-of-order outbox apply for the same file.

---

### Security

The authorization model is sound: quota **set** requires TiDB Cloud public/private keys verified against the live cluster (GET/PATCH digest auth), and explicitly rejects drive9 API keys (`TestQuotaSetRejectsDrive9KeyWithoutTiDBCloudCredentials`) — a tenant cannot raise its own quota. Owner **get** goes through `tenantAuthMiddleware`. Upstream error bodies are sanitized before surfacing. Credentials travel in the JSON body, so HTTPS is assumed in transit — fine for this API.

docs/guides/quota.md
```diff
@@ -0,0 +1,168 @@
...
+| `502 Bad Gateway` | TiDB Cloud returned another upstream error while querying quota, updating quota labels, or updating Spending Limit. |
+
+## Notes for operators
+
+- Server-side quota reads high-churn usage counters from the central meta store.
+  Low-churn quota config is cached per tenant and refreshed by version polling.
+- In cloud-native mode, small write quota mutations are recorded in a
+  tenant-local quota outbox and applied asynchronously to central quota state.
+- The write path includes pending outbox deltas in quota admission so
+  multi-server deployments do not over-admit against stale central usage.
```
qiffang: This guarantee no longer matches the implementation. The new `ensureStorageQuotaServer` path intentionally does not take `quota_admission_locks`, and `pkg/backend/quota.go` now states that concurrent small writes in multi-server deployments may briefly over-admit before each server can see the other server's outbox rows. Publishing “do not over-admit” makes the small-write path look like strict admission when it is now soft/convergent admission. Please change this to say pending outbox deltas reduce stale-central-usage undercounting, but concurrent small writes may briefly over-admit; durable outbox processing/backfill restores convergence. Multipart upload reservation can still be described as the stricter reserve-first path.

qiffang (changes requested):
R4 review on latest head `6beda7f5907114009c12698d769e39ca934d80fc`: REVISE.

CI is green and the new small-write lock-free direction is internally consistent in code/tests, but the docs still publish a stronger quota guarantee than the implementation provides. This must be fixed before merge because it defines the operator-facing contract for multi-server quota enforcement.

After that wording fix, I do not currently see another blocker in the new delta.

docs/guides/quota.md
```diff
@@ -0,0 +1,168 @@
...
+HTTP API.
+
+## What quota tracks
+
+Drive9 exposes these user-settable quota settings:
+
+| Field | Meaning |
+| --- | --- |
+| `max_storage_size` | Maximum confirmed plus reserved file storage size, in Mi. Stored in Drive9. |
+| `tidbcloud_spending_limit` | TiDB Cloud Cluster Spending Limit. The value is passed through to TiDB Cloud, read from and written to the TiDB Cloud cluster, and not stored in Drive9. See the [TiDB Cloud Spending Limit guide](https://docs.pingcap.com/tidbcloud/manage-serverless-spend-limit). |
```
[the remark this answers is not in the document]
srstack: The current guide already uses standard single-pipe Markdown tables, and rg for lines beginning with double pipes returns no matches. Resolving this as stale.

srstack (commented):
## Re-review: feat/quota-cloud-native-outbox (since 14:46 — +2286 lines, 15 new files)

**Overall: the prior review's two correctness findings are resolved, and the API was meaningfully hardened. One item from my first pass remains open; a couple of new things are worth a look. Close to LGTM.**

### ✅ Resolved from prior review

- **🔴 Outbox apply ordering** — fixed properly. `ClaimQuotaOutbox` now refuses to claim a row while any *older* row is unfinished (`NOT EXISTS (older.id < q.id AND older.status IN (queued, processing))`) plus `FOR UPDATE SKIP LOCKED`. The create-then-overwrite reordering hazard I described can no longer happen.
- **🟠 No dead-lettering** — fixed. `defaultQuotaOutboxMaxAttempts` 0 → 100, with a comment tying the bound to the ordering barrier so a poison row can't block the queue forever.
- **Pending-delta multi-server race** — addressed beyond what I flagged: the new `quota_admission_locks` row-lock (`LockQuotaAdmissionTx`, upsert + `SELECT … FOR UPDATE`) now serializes the upload-reserve check+reserve per tenant, and the outbox worker takes the same lock. Small writes remain intentionally soft (documented), which is a reasonable hot-path tradeoff.
- **Upstream error leakage** — `writeQuotaCredentialError` no longer echoes the upstream body to clients; it logs server-side and returns a generic message. Good hardening.

### 🔴 Still open: schema migration for existing app-mode / db9 server-quota tenants

`quota_outbox` and `quota_admission_locks` are in every schema list and the no-embedding validation count, so new tenants and existing **auto-embedding** tenants are covered (content-derived version bump → `pool.go:603` re-ensure; new `TestPoolCreateBackendEnsuresSchemaForTiDBCloudNative` confirms the cloud-native ensure path).

But the ensure block at `pool.go:587` is still gated on `opts.DatabaseAutoEmbedding`. An existing **app-mode** (`DisableDatabaseAutoEmbedding`) or **db9** tenant running `DRIVE9_QUOTA_SOURCE=server` never re-runs schema on open, so it won't get the two new tables — and every create/overwrite would then fail on `INSERT INTO quota_outbox` inside the write transaction. Still needs confirmation: are there existing server-quota tenants on those providers, and if so, what creates the tables for them? If the answer is "server quota only ever runs on auto-embedding TiDB tenants," a one-line comment at the enqueue site stating that invariant would close this out.

### New code — looks good

- **API v2 redesign is consistent across all three layers.** `/v1/quota/query` was removed; GET now takes TiDB Cloud creds via `X-TiDBCloud-Public-Key/-Private-Key` headers + `tenant_id` query param, POST/set keeps creds in the JSON body. Server (`decodeQuotaGetRequest`), SDK (`pkg/client/quota.go`), and CLI (`client.New(...).GetQuota/SetQuota`) all agree. The owner-API-key GET path is gone — every quota read now requires cloud creds. That's a defensible tightening (a tenant can't even *read* quota without cluster creds), just calling it out as a behavior change in case any internal caller relied on owner-key reads.
- **`quotaStorageSizeToBytes`** guards Mi→bytes multiplication overflow (`sizeMi > maxInt64/quotaStorageSizeBytes`) before multiplying. Correct.
- **Spending-limit passthrough** (`TiDBCloudSpendingLimitMonthly`) is wired end-to-end with its own PATCH and wrapped errors.

### Minor

- **`handleQuotaSet` partial-failure ordering**: the external `UpdateQuota` PATCH (cluster labels + spending limit) runs before the local `SetQuotaStorageBytes`. If the local write fails after the cloud PATCH succeeds, the cluster's drive9 labels / spending limit lead the persisted `max_storage_bytes`. Same drift note as before — low impact (the response surfaces the cloud config), but a warning log on the post-PATCH local-write failure would aid diagnosis.
- **`LockQuotaAdmissionTx` does a blind `INSERT … ON DUPLICATE KEY UPDATE updated_at` every reserve** to guarantee the lock row exists. Fine, but it means every upload reserve now writes that row even when it already exists. Negligible, just noting the extra write on the reserve path.
- **Error wrapping pass** in the provisioner (`fmt.Errorf("...: %w", err)` throughout `UpdateQuota`/`clusterQuotaInfo`) is a nice readability/debuggability improvement.

### Tests

The new surface is well covered: `quota_outbox_test.go` exercises the admission lock + ordering, `pool_test.go` adds the cloud-native ensure assertion, `client/quota_test.go` covers the SDK, and the provisioner tests cover label merge + spending limit + error mapping. The one gap still maps to the open item — no test asserts an existing app-mode/db9 server-quota tenant has the outbox table on open.

qiffang (approved):
R5 review: GREEN on latest head `5cd423dcbac8d7b32899554ccb7f43e1edb77f3a`.

Rechecked the delta after my R4 changes-requested review. The R4 blocker is fixed: `docs/guides/quota.md` no longer claims multi-server small-write admission cannot over-admit; it now documents the intended soft/convergent semantics and keeps multipart uploads scoped to reserve-first.

Additional checks:
- Positive quota values are now enforced in CLI, HTTP handler, and TiDB Cloud native provisioner before upstream calls.
- Quota get/set no longer fall back to server default TiDB Cloud credentials; explicit credentials are required and tested.
- Generic TiDB Cloud quota failures now return a stable client error and log raw details server-side.
- Quota config versioning now uses a content token rather than `updated_at`, fixing the same-timestamp cache invalidation hardening risk.
- Rechecked Copilot's patch-upload quota concern: patch upload initiation calls `reserveUploadOnServer(...)` before tenant upload state is inserted, so server-quota mode is still quota-gated; the later tenant-DB `ensureTenantStorageQuotaTx` skip is intentional after server reservation.
- Rechecked CodeRabbit's pending-delta fail-open comment: continuing with central confirmed/reserved usage and pending delta `0` is a softer/stale quota check, not an over-limit bypass relative to the documented soft-admission semantics. Returning false would skip even confirmed-usage enforcement, so I do not treat this as a merge blocker.

Validation: `git diff --check 6beda7f..HEAD` passed locally. Local Go is unavailable in this agent env, so I relied on static review plus GitHub checks. GitHub `code-ci`, `local-e2e`, and CodeRabbit are green for this head.

mornyx (approved):
Re-reviewed PR #603 at `5cd423dcbac8d7b32899554ccb7f43e1edb77f3a`. The two blockers from my previous review have been addressed: quota outbox convergence now holds the admission lock across apply/ack, and storage-only quota updates no longer disable the default monthly LLM cost cap. I also did a quick pass over the updated quota API/CLI/client surface and did not find a new critical/functional blocker. Per request, I did not rerun local tests and relied on CI; the GitHub CI and e2e smoke checks are green.

pkg/backend/quota.go
```diff
@@ -105,93 +117,111 @@ func (b *Dat9Backend) ensureUploadSizeAllowed(size int64) error {
...
+	if usage == nil {
+		metrics.RecordOperation("server_quota", "storage_check", "fail_open", 0)
+		return result, false // fail-open: usage unavailable
 	}
-	metrics.RecordOperation("server_quota", "storage_check", "ok", 0)
+	recordTenantQuotaSnapshot(b.tenantID, usage, cfg)
+	pendingStorageDelta, _, pendingOK := b.pendingQuotaOutboxDeltasTx(ctx, tx)
+	if !pendingOK {
+		metrics.RecordOperation("server_quota", "storage_check_pending_delta", "fail_open", 0)
+	}
```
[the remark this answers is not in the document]
srstack: Reviewed against the intended soft-admission behavior from R5. I am leaving this path as-is intentionally: if pending outbox deltas cannot be loaded, continuing with `pendingStorageDelta=0` still enforces the known central confirmed/reserved usage. Returning `false` here would skip even that known central quota check and make enforcement weaker. The failed pending-delta load is already recorded with the fail-open metric, and the operator docs describe small-write admission as soft/convergent across servers.

srstack (commented):
## Final review (post-merge record): feat: add cloud native quota controls (PR #603)

> Note: this PR is **MERGED** (2026-06-23 16:19). This is a consolidated record of the review across earlier rounds; no code action remains. Captured here for completeness per the review request.

### Overview

Three related capabilities:
1. **Cloud-native quota CLI + HTTP API** — `drive9 quota get|set`, `/v1/quota`, `/v1/quota/query`; owner-key reads and TiDB-Cloud-credential-gated writes, with drive9-managed cluster labels.
2. **Quota outbox** — moves server-quota small-write mutations off the fsync hot path: a `quota_outbox` row is written in the same tenant-DB transaction as the file mutation and drained asynchronously, removing a ~37ms central round-trip per write while keeping crash-safety.
3. **Pending-delta awareness** — quota checks and upload reservations include un-drained outbox deltas for multi-server correctness.

### Code quality

- The transactional-outbox pattern is the right call: the quota mutation is atomic with the file write, and `applyCentralFileStateTx`'s shadow-diff makes individual mutations idempotent (necessary for at-least-once delivery). Well structured and well tested.
- API layering (CLI → SDK semantics → server handler → provisioner credential interfaces) is clean and consistent.

### Findings (all resolved before merge)

- 🔴 **Outbox apply ordering** — `ClaimQuotaOutbox` gained a strict ordering barrier (`NOT EXISTS (older.id < q.id AND older.status IN (queued, processing))` + `FOR UPDATE SKIP LOCKED`), preventing create-then-overwrite reordering that would converge the shadow counter wrong. ✅
- 🟠 **No dead-lettering** — `defaultQuotaOutboxMaxAttempts` 0 → 100, so a poison row can't retry forever behind the ordering barrier. ✅
- 🟠 **Multi-server over-admit** — added `quota_admission_locks` row-lock serializing the reserve check+insert per tenant; the outbox worker takes the same lock. Small writes stay intentionally soft (documented). ✅
- 🔴 **Schema migration for app-mode / `DisableDatabaseAutoEmbedding` tenants** — the final commit `be764c797c` re-gated the ensure/repair block on `UsesTiDBAutoEmbedding(provider)` instead of the runtime `DatabaseAutoEmbedding` flag, so those tenants still get `quota_outbox` + `quota_admission_locks` created on open (only the embedding *provider config* is skipped). Central meta DB columns are covered by the `migrate()` auto-repair path. ✅
- **Cloud-vs-local write ordering** — `MarkQuotaUpdated` → `MarkQuotaUpdateStarted` reorders to confirm cluster-write permission before persisting local config, reducing partial-failure drift. ✅
- **Upstream error leakage** — credential-error responses no longer echo upstream bodies to clients; logged server-side, generic message returned. ✅

### Performance

- Hot-path win is the core motivation and is achieved: per-fsync central round-trip replaced by an in-tenant-tx outbox insert + async drain. Config is cached with version polling; usage stays live for correctness. (Follow-up tuning landed in #607.)

### Test coverage

Strong: real-MySQL datastore lifecycle/recover tests, outbox ordering + admission-lock tests, server handler tests (owner-get / credential-query / set / non-cloud-native rejection / error mapping), and provisioner label-merge + digest-auth + forbidden/not-found tests.

### Security

Sound authorization model: quota **set** requires live-cluster-verified TiDB Cloud keys and explicitly rejects drive9 API keys (a tenant cannot raise its own quota); owner **get** goes through `tenantAuthMiddleware`; upstream error bodies sanitized.

### Verdict

Approved in substance — all blocking and major findings were addressed across review rounds, and the PR merged in that state. No outstanding action.

PR #610, opened 2026-06-24
fix: keep quota usage cache warm after outbox ack

srstack:
## Summary

Follow-up to the quota hot-path work for the 1k small-file upload case.

The previous outbox ACK path invalidated quotaUsageCache after every successfully applied quota outbox row. That closes the stale-usage undercount window, but under a busy upload it also forces the next soft small-write quota check to re-read central tenant_quota_usage after each ACK, reintroducing central quota reads while the outbox worker is active.

This PR keeps the cache warm instead:

- add quotaUsageCache.add(storageDelta, fileDelta, mediaDelta)
- after a quota outbox row is applied and ACKed, forward the fresh usage snapshot by the applied deltas
- keep the old cold behavior when the snapshot is missing or expired; the next soft check reloads central usage normally
- continue subtracting the same deltas from quotaPendingDeltasCache, so usage + pending remains stable across ACK

Strict upload reservations are unchanged; they still read live central usage and live tenant pending deltas under the admission lock.

## Validation

- make test TEST_PKGS=./pkg/backend TEST_RUN=TestQuotaUsageCache|TestQuotaPendingDeltasCache|TestQuotaOutbox|TestServerQuota|TestServerModeBudget|TestConcurrentServerQuota|TestInitiateUpload
- make lint
- git diff --check

## Notes

This is intentionally narrow. The remaining 1k upload bottleneck may still involve pending-delta cache misses falling back to live SUM(quota_outbox) under heavy concurrent create load, but that needs a more careful design because caching those sums too aggressively can undercount cross-server pending rows. This PR only removes the avoidable central usage reload caused by our own outbox ACK path.

### DIFF
--- a/pkg/backend/quota.go
+++ b/pkg/backend/quota.go
@@ -387,6 +387,12 @@ func (b *Dat9Backend) addLocalQuotaPendingDeltas(storageDelta, fileDelta, mediaD
 	}
 }
 
+func (b *Dat9Backend) addLocalQuotaUsageDeltas(storageDelta, fileDelta, mediaDelta int64) {
+	if b.quotaUsageCache != nil {
+		b.quotaUsageCache.add(storageDelta, fileDelta, mediaDelta)
+	}
+}
+
 func (r storageQuotaCheckResult) exceeded() bool {
 	return r.checked && r.projected > r.limitBytes
 }
--- a/pkg/backend/quota_cache.go
+++ b/pkg/backend/quota_cache.go
@@ -239,13 +239,22 @@ func (c *quotaUsageCache) cached(now time.Time) *QuotaUsageView {
 	return nil
 }
 
-func (c *quotaUsageCache) invalidate() {
+func (c *quotaUsageCache) add(storageDelta, fileDelta, mediaDelta int64) {
 	if c == nil {
 		return
 	}
 	c.mu.Lock()
-	c.snapshot = nil
-	c.mu.Unlock()
+	defer c.mu.Unlock()
+	// If the snapshot is missing or expired, leave it cold. The next soft
+	// admission check reloads the central usage row. When the snapshot is
+	// fresh, a successfully applied quota outbox row can be reflected locally
+	// without forcing the next small write to hit the central DB again.
+	if c.snapshot == nil || c.snapshot.usage == nil || time.Now().After(c.snapshot.expiresAt) {
+		return
+	}
+	c.snapshot.usage.StorageBytes += storageDelta
+	c.snapshot.usage.FileCount += fileDelta
+	c.snapshot.usage.MediaFileCount += mediaDelta
 }
 
 func cloneQuotaUsageView(usage *QuotaUsageView) *QuotaUsageView {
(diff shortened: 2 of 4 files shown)


srstack (commented):
## Review: fix/quota-ack-cache-forward (PR #610)

**Overall: correct, narrow, low-risk. Does what it claims. But — as the PR itself notes — it won't move the headline regression much, because the bottleneck isn't the post-ACK usage reload. LGTM for what it is; the real win needs the follow-up below.**

### What it does

Replaces `quotaUsageCache.invalidate()` (after a quota-outbox row is ACKed) with `add(storageDelta, fileDelta, mediaDelta)` — forwarding the applied deltas into the warm snapshot instead of dropping it. So the next soft small-write check doesn't re-read central `tenant_quota_usage` after every ACK while the outbox worker is busy.

### Correctness — verified

- **Invariant preserved**: ACK does `quotaUsageCache.add(+deltas)` and `quotaPendingDeltasCache.add(-deltas)` together, so `usage + pending` stays stable across the ACK — the soft projection is unchanged, only the source shifts from "central + pending" to "cached + (pending − applied)". Good.
- **Cold/expired guard is correct**: `add()` no-ops when the snapshot is nil or past `expiresAt`, so a stale snapshot is never resurrected with a partial delta — the next soft check reloads central normally. This is the right call (a delta applied onto an expired base would be wrong).
- **Strict path untouched**: reservations still use `livePendingQuotaOutboxDeltasTx` + live usage under the admission lock. Confirmed unaffected.
- **Mutex discipline**: `add()` takes `c.mu` and mutates the snapshot in place; consistent with `get()`/`cached()`. Note it mutates `c.snapshot.usage` directly rather than cloning — fine because `get()` returns a clone, so external callers can't observe the in-place write mid-flight.

### 🟡 Notes

- The deltas are applied to **this backend's** cache only. Another server's ACKs aren't reflected until this cache's TTL reload — same soft-window caveat as the rest of the soft path, and acceptable here.
- This optimizes the *post-ACK* read, i.e. it helps the soft *small-write check* avoid a reload. It does **not** reduce the per-row central **apply** cost on the outbox worker, which is the actual create-throughput ceiling (see below). So expect this to trim some central `GetQuotaUsage` calls under load, but not to materially lift the 160 files/s number.

### Tests

`TestQuotaUsageCacheAddUpdatesFreshSnapshot` correctly replaces the old invalidate test and asserts the warm-add path (storage 10→15, file 1→3, media→1, exactly 1 central call). Good. Missing: a test for the **cold/expired** branch (add after expiry → next get triggers a reload, snapshot not corrupted). Worth adding since that guard is the subtle part.

### Verdict

LGTM — safe, correct, ship it. But it's a marginal optimization; the regression fix proper is in the per-row outbox apply, covered in the design proposal I'm sending separately.

srstack:
Closing this narrow cache-forwarding PR to keep the performance work focused. The real 1k-create bottleneck is the write-side quota outbox drain: global FIFO ordering plus one-row claim/apply/ack. I will move the effort to a P0 PR for per-file ordering and batch outbox apply instead of pursuing this marginal usage-cache optimization as a standalone change.

PR #612, opened 2026-06-24
fix: batch quota outbox processing per file

srstack:
## Summary
- change quota outbox claiming from tenant-global FIFO to per-file ordering so unrelated file mutations can drain concurrently
- add batch claim/apply/ack for quota outbox rows to reduce per-row DB round trips under small-file create workloads
- treat NULL file_id outbox rows as tenant-global ordering barriers for legacy/global mutations
- add a dedicated quota_outbox(file_id, status, id) index for the per-file ordering subquery
- keep batch fallback bounded by taking the tenant admission lock per fallback row instead of holding it across the full fallback batch

## Ops note
- Existing TiDB tenants with quota_outbox already created will get idx_quota_outbox_file_order during the next schema ensure/repair pass. TiDB ADD INDEX is online, but large backlog tables may spend time/CPU building this index once.

## Tests
- GOCACHE=/tmp/drive9-go-cache go test ./pkg/datastore -run 'TestQuotaOutbox' -count=1
- GOCACHE=/tmp/drive9-go-cache go test ./pkg/backend -run 'Test.*QuotaOutbox|TestDrainQuotaOutbox|TestServerQuota.*Pending|TestServerQuotaReserve|TestServerQuotaOverwrite|TestServerQuotaMedia' -count=1
- GOCACHE=/tmp/drive9-go-cache go test ./pkg/tenant/schema -run 'TestQuotaOutbox|TestSchema|Test.*Repair|TestPlanned' -count=1
- GOCACHE=/tmp/drive9-go-cache go test ./cmd/drive9-server -run 'TestSchemaDump' -count=1
- GOCACHE=/tmp/drive9-go-cache go test ./pkg/backend ./pkg/datastore ./pkg/tenant/schema ./cmd/drive9-server -count=1
- make lint
- git diff --check

## Summary by CodeRabbit

* **New Features**
  * Quota outbox processing now works in batches, improving throughput and reducing repeated single-item work.
  * File-based draining handles multiple pending entries more efficiently and continues past unrelated processing issues when possible.

* **Bug Fixes**
  * Improved ordering and fairness for queued quota actions, including better handling of same-file sequences and global queue items.
  * Added schema support for a new index that helps keep quota outbox processing performant and repairable.

### DIFF
--- a/cmd/drive9-server/schema_test.go
+++ b/cmd/drive9-server/schema_test.go
@@ -28,6 +28,9 @@ func TestSchemaDumpInitSQLByProvider(t *testing.T) {
 			if !strings.Contains(out, "CREATE INDEX idx_task_claim_type ON semantic_tasks") {
 				t.Fatalf("dump missing semantic_tasks index: %q", out)
 			}
+			if !strings.Contains(out, "CREATE INDEX idx_quota_outbox_file_order ON quota_outbox") {
+				t.Fatalf("dump missing quota_outbox file-order index: %q", out)
+			}
 			if !strings.Contains(out, ";\n") {
 				t.Fatalf("dump missing SQL statement terminators: %q", out)
 			}
--- a/pkg/backend/quota_outbox_test.go
+++ b/pkg/backend/quota_outbox_test.go
@@ -5,6 +5,7 @@ import (
 	"database/sql"
 	"encoding/json"
 	"errors"
+	"fmt"
 	"testing"
 	"time"
 
@@ -102,6 +103,50 @@ func TestServerQuotaOutboxBatchProcessesDifferentFiles(t *testing.T) {
 	}
 	if usage.StorageBytes != int64(len("aaaa")+len("bb")) || usage.FileCount != 2 || usage.MediaFileCount != 1 {
 		t.Fatalf("central usage after batch = %+v", usage)
+	}
+}
+
+func TestQuotaOutboxWorkerRepollsAfterUnderfilledSameFileBatch(t *testing.T) {
+	b, fake := newServerQuotaBackend(t, Options{})
+	b.stopQuotaOutboxWorker()
+	ctx := context.Background()
+	fileID := "same-file-backlog"
+
+	if _, err := b.store.EnqueueQuotaOutboxTx(b.store.DB(), &datastore.QuotaOutboxEntry{
+		FileID:       fileID,
+		MutationType: quotaMutationTypeFileCreate,
+		MutationData: mustQuotaMutationJSON(t, fileCreateMutationData{
+			FileID:    fileID,
+			SizeBytes: 1,
+		}),
+		StorageDelta: 1,
+		FileDelta:    1,
+	}); err != nil {
+		t.Fatalf("enqueue create: %v", err)
+	}
+	if _, err := b.store.EnqueueQuotaOutboxTx(b.store.DB(), &datastore.QuotaOutboxEntry{
+		FileID:       fileID,
+		MutationType: quotaMutationTypeOverwrite,
+		MutationData: mustQuotaMutationJSON(t, fileOverwriteMutationData{
+			FileID:       fileID,
+			OldSizeBytes: 1,
+			NewSizeBytes: 3,
+		}),
+		StorageDelta: 2,
+	}); err != nil {
+		t.Fatalf("enqueue overwrite: %v", err)
+	}
+
+	b.processQuotaOutboxAvailable(ctx)
+	if got := countQuotaOutboxRowsByFile(t, ctx, b, fileID, datastore.QuotaOutboxSucceeded); got != 2 {
+		t.Fatalf("succeeded same-file rows = %d, want 2", got)
+	}
+	usage, err := fake.GetQuotaUsage(ctx, "tenant-a")
+	if err != nil {
+		t.Fatalf("get usage: %v", err)
+	}
+	if usage.StorageBytes != 3 || usage.FileCount != 1 {
+		t.Fatalf("usage after same-file backlog drain = %+v, want storage=3 files=1", usage)
 	}
 }
 
@@ -339,6 +384,56 @@ func TestDrainQuotaOutboxForUploadTargetProcessesClaimableBatch(t *testing.T) {
 	}
 	if usage.StorageBytes != int64(len("older")+len("target")) || usage.FileCount != 2 {
 		t.Fatalf("central usage after drain = %+v, want both files applied", usage)
+	}
+}
+
+func TestDrainQuotaOutboxForFileContinuesAfterUnrelatedBatchError(t *testing.T) {
+	b, _ := newServerQuotaBackend(t, Options{})
+	b.stopQuotaOutboxWorker()
+	ctx := context.Background()
+
+	for i := 0; i < quotaOutboxBatchSize; i++ {
+		fileID := "older-file"
+		data := json.RawMessage(`{"file_id":"bad-json","size_bytes":"bad","is_media":false}`)
+		if i > 0 {
+			fileID = fmt.Sprintf("older-file-%03d", i)
+			data = mustQuotaMutationJSON(t, fileCreateMutationData{
+				FileID:    fileID,
+				SizeBytes: 1,
+			})
+		}
+		if _, err := b.store.EnqueueQuotaOutboxTx(b.store.DB(), &datastore.QuotaOutboxEntry{
+			FileID:       fileID,
+			MutationType: quotaMutationTypeFileCreate,
+			MutationData: data,
+			StorageDelta: 1,
+			FileDelta:    1,
+		}); err != nil {
+			t.Fatalf("enqueue older %d: %v", i, err)
+		}
+	}
+	targetFileID := "target-file-after-unrelated-error"
+	if _, err := b.store.EnqueueQuotaOutboxTx(b.store.DB(), &datastore.QuotaOutboxEntry{
+		FileID:       targetFileID,
+		MutationType: quotaMutationTypeFileCreate,
+		MutationData: mustQuotaMutationJSON(t, fileCreateMutationData{
+			FileID:    targetFileID,
+			SizeBytes: 1,
+		}),
+		StorageDelta: 1,
+		FileDelta:    1,
+	}); err != nil {
+		t.Fatalf("enqueue target: %v", err)
+	}
+
+	if err := b.drainQuotaOutboxForFile(ctx, targetFileID, quotaOutboxUploadDrainLimit); err != nil {
+		t.Fatalf("drain target: %v", err)
+	}
+	if got := countQuotaOutboxRowsByFile(t, ctx, b, targetFileID, datastore.QuotaOutboxSucceeded); got != 1 {
+		t.Fatalf("target succeeded rows = %d, want 1", got)
+	}
+	if got := countQuotaOutboxRowsByFile(t, ctx, b, "older-file", datastore.QuotaOutboxQueued); got != 1 {
+		t.Fatalf("bad unrelated queued retry rows = %d, want 1", got)
 	}
 }
 
@@ -599,4 +694,13 @@ func latestQuotaOutboxStorageDeltaByFile(t *testing.T, ctx context.Context, b *D
 		t.Fatal(err)
 	}
 	return delta
+}
+
+func mustQuotaMutationJSON(t *testing.T, payload any) json.RawMessage {
+	t.Helper()
+	raw, err := json.Marshal(payload)
+	if err != nil {
+		t.Fatalf("marshal quota mutation: %v", err)
+	}
+	return raw
 }
--- a/pkg/backend/quota_outbox_test.go
+++ b/pkg/backend/quota_outbox_test.go
@@ -63,6 +63,48 @@ func TestServerQuotaSmallWriteUsesTenantOutbox(t *testing.T) {
 	}
 }
 
+func TestServerQuotaOutboxBatchProcessesDifferentFiles(t *testing.T) {
+	b, fake := newServerQuotaBackend(t, Options{})
+	b.stopQuotaOutboxWorker()
+	ctx := context.Background()
+
+	if _, err := b.WriteCtx(ctx, "/batch-a.txt", []byte("aaaa"), 0, filesystem.WriteFlagCreate); err != nil {
+		t.Fatalf("write a: %v", err)
+	}
+	a, err := b.Store().Stat(ctx, "/batch-a.txt")
+	if err != nil {
+		t.Fatalf("stat a: %v", err)
+	}
+	if _, err := b.WriteCtx(ctx, "/batch-b.png", []byte("bb"), 0, filesystem.WriteFlagCreate); err != nil {
+		t.Fatalf("write b: %v", err)
+	}
+	bb, err := b.Store().Stat(ctx, "/batch-b.png")
+	if err != nil {
+		t.Fatalf("stat b: %v", err)
+	}
+
+	processed, err := b.ProcessQuotaOutboxBatch(ctx, 100)
+	if err != nil {
+		t.Fatalf("process quota outbox batch: %v", err)
+	}
+	if processed != 2 {
+		t.Fatalf("processed rows = %d, want 2", processed)
+	}
+	if got := countQuotaOutboxRowsByFile(t, ctx, b, a.File.FileID, datastore.QuotaOutboxSucceeded); got != 1 {
+		t.Fatalf("a succeeded rows = %d, want 1", got)
+	}
+	if got := countQuotaOutboxRowsByFile(t, ctx, b, bb.File.FileID, datastore.QuotaOutboxSucceeded); got != 1 {
+		t.Fatalf("b succeeded rows = %d, want 1", got)
+	}
+	usage, err := fake.GetQuotaUsage(ctx, "tenant-a")
+	if err != nil {
+		t.Fatalf("get usage: %v", err)
+	}
+	if usage.StorageBytes != int64(len("aaaa")+len("bb")) || usage.FileCount != 2 || usage.MediaFileCount != 1 {
+		t.Fatalf("central usage after batch = %+v", usage)
+	}
+}
+
 func TestServerQuotaSmallWriteDoesNotWaitForAdmissionLock(t *testing.T) {
 	b, _ := newServerQuotaBackend(t, Options{})
 	b.stopQuotaOutboxWorker()
@@ -262,7 +304,7 @@ func TestDrainQuotaOutboxForFileErrorsWhenPendingRowIsNotClaimable(t *testing.T)
 	}
 }
 
-func TestDrainQuotaOutboxForUploadTargetProcessesOlderTenantRows(t *testing.T) {
+func TestDrainQuotaOutboxForUploadTargetProcessesClaimableBatch(t *testing.T) {
 	b, fake := newServerQuotaBackend(t, Options{})
 	b.stopQuotaOutboxWorker()
 	ctx := context.Background()
--- a/pkg/backend/quota_outbox_worker.go
+++ b/pkg/backend/quota_outbox_worker.go
@@ -127,22 +127,17 @@ func (b *Dat9Backend) processQuotaOutboxAvailable(ctx context.Context) {
 			zap.Error(err))
 		metrics.RecordOperation("quota_outbox", "recover_expired", "error", 0)
 	}
-	for i := 0; i < quotaOutboxBatchSize; i++ {
-		if ctx.Err() != nil {
+	processed, err := b.ProcessQuotaOutboxBatch(ctx, quotaOutboxBatchSize)
+	if err != nil {
+		if isContextDone(err) {
 			return
 		}
-		processed, err := b.ProcessOneQuotaOutbox(ctx)
-		if err != nil {
-			if isContextDone(err) {
-				return
-			}
-			logger.Warn(ctx, "quota_outbox_process_failed",
-				zap.String("tenant_id", b.tenantID),
-				zap.Error(err))
-		}
-		if !processed {
-			return
-		}
+		logger.Warn(ctx, "quota_outbox_process_failed",
+			zap.String("tenant_id", b.tenantID),
+			zap.Error(err))
+	}
+	if processed == quotaOutboxBatchSize {
+		b.notifyQuotaOutbox(true)
 	}
 }
 
@@ -165,7 +160,7 @@ func (b *Dat9Backend) drainQuotaOutboxForFile(ctx context.Context, fileID string
 	}
 	start := time.Now()
 	processedCount := 0
-	for i := 0; i < limit; i++ {
+	for processedCount < limit {
 		if quotaOutboxUploadDrainMaxWait > 0 && time.Since(start) > quotaOutboxUploadDrainMaxWait {
 			return fmt.Errorf("quota outbox for file %s still pending after %s", fileID, quotaOutboxUploadDrainMaxWait)
 		}
@@ -182,11 +177,19 @@ func (b *Dat9Backend) drainQuotaOutboxForFile(ctx context.Context, fileID string
 			}
 			return nil
 		}
-		processed, err := b.ProcessOneQuotaOutbox(ctx)
+		batchLimit := quotaOutboxBatchSize
+		if remaining := limit - processedCount; remaining < batchLimit {
+			batchLimit = remaining
+		}
+		processed, err := b.ProcessQuotaOutboxBatch(ctx, batchLimit)
 		if err != nil {
+			pendingAfterErr, pendingErr := b.store.HasPendingQuotaOutboxFileMutation(ctx, fileID)
+			if pendingErr == nil && !pendingAfterErr {
+				return nil
+			}
 			return fmt.Errorf("drain quota outbox for file: %w", err)
 		}
-		if !processed {
+		if processed == 0 {
 			timer := time.NewTimer(quotaOutboxUploadDrainWait)
 			select {
 			case <-ctx.Done():
@@ -198,7 +201,7 @@ func (b *Dat9Backend) drainQuotaOutboxForFile(ctx context.Context, fileID string
 			}
 			continue
 		}
-		processedCount++
+		processedCount += processed
 	}
 	return fmt.Errorf("quota outbox for file %s still pending after %d entries", fileID, limit)
 }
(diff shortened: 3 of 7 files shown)


srstack (commented):
## Review: fix/quota-outbox-batch-apply (PR #612)

**Overall: this is the right structural fix for the create-throughput regression — exactly the per-file-ordering + batch-apply change the bottleneck called for. The core claim logic is sound. One key risk to confirm (index migration for existing tenants) and one merge-coordination note with #610. Strong PR.**

### What it does

Replaces the quota outbox's **tenant-global FIFO single-flight** drain with **per-file ordering + batched claim/apply/ack**:
- `ClaimQuotaOutboxBatch(N)` claims up to N queued rows whose per-file predecessors are all done (`NOT EXISTS older WHERE older.id < q.id AND status IN (queued,processing) AND same file_id`), dropping the old global `older.id < q.id` barrier and the global "only one processing" guard.
- `ProcessQuotaOutboxBatch` applies a batch in one central `InTx` + one `AckQuotaOutboxBatchTx`, under a single admission-lock window; falls back to per-entry apply/retry if the batch tx fails.
- New `idx_quota_outbox_file_order(file_id, status, id)` for the per-file subquery.

This directly removes the "1000 rows × 4 serial round-trips, each under the admission lock, with a queue-depth-degrading global barrier" ceiling. Good.

### Correctness — reasoned through, looks right

- **Per-file ordering preserved (the #603 invariant)**: the subquery now scopes the predecessor check to the same `file_id` (with correct NULL-file handling: `older.file_id = q.file_id OR (both NULL)`). So create→overwrite→delete on one file still serialize, while unrelated files drain concurrently. The new datastore test `TestQuotaOutboxBatchClaimPreservesPerFileOrder` pins exactly this (file-1 overwrite not claimable until file-1 create is acked). ✅
- **Cross-pod ordering holds**: a second pod evaluating the subquery sees an in-flight predecessor as still `queued`/`processing` in its committed snapshot (the claiming pod's row is row-locked and its status flip isn't visible until commit), so it can't jump ahead on the same file. `SKIP LOCKED` only affects the main claim, not the subquery scan. ✅
- **Dropping the global single-flight guard is safe**: concurrency is now bounded per-file by the subquery, and the **admission lock still serializes the central apply window** across pods, so no concurrent central mutation. ✅
- **Batch atomicity**: batch apply runs in one central `InTx` that rolls back wholesale on error, then the code falls through to per-entry apply — no double-apply, because the rolled-back batch left central state untouched. And `applyCentralFileStateTx` remains idempotent (shadow-diff), so the pre-existing apply-commits-then-ack-fails at-least-once window is unchanged (re-apply converges to delta 0). ✅
- **Lease integrity**: claim/ack now assert `rowsAffected == len(entries)` and return `ErrQuotaOutboxLeaseMismatch` otherwise — good, catches a concurrent lease steal.

### 🟠 Key risk to confirm: index migration for *existing* tenants

The new `idx_quota_outbox_file_order` is appended to the schema statement lists, so **new** tenants get it. But the per-file ordering subquery's performance *depends* on this index — and existing benchmarked tenants already have a `quota_outbox` table. Please confirm the tenant schema-repair path actually **adds a missing index to an existing table** (not just missing tables/columns). If it doesn't:
- The subquery `NOT EXISTS (older WHERE file_id=… AND status IN(…) AND id<…)` falls back to `idx_quota_outbox_file_pending(file_id, status, mutation_type)` — close, but it doesn't cover the `id` range/order, so the predecessor check degrades to a scan within the file's rows.
- That would blunt (or under backlog, partially negate) the very speedup this PR targets, on exactly the tenants being measured.

This is the single thing I'd verify before trusting the benchmark delta. A `re-mount on an existing tenant → confirm idx_quota_outbox_file_order exists` check, or an explicit `ALTER`/repair, closes it.

### 🟡 Merge coordination with #610

This diff calls `b.quotaUsageCache.invalidate()` after a batch (line ~250), but #610 ("keep quota usage cache warm after outbox ack") replaced `invalidate()` with `add(deltas)`. Whichever merges second will conflict and needs reconciliation. The good news: even reverting to `invalidate()`, #612 calls it **once per batch** instead of once per row, which already removes most of the reload churn #610 was chasing — so #610's `add()` becomes a smaller marginal win on top. Recommend: rebase #612 onto #610 (or vice-versa) and keep `add()` applied **once per batch** with the summed deltas, getting both benefits.

### 🟡 Minor

- **`ProcessQuotaOutboxBatch` returns `len(entries)` (not `processed`) when `batchErr != nil`** (line ~260). Callers handle it safely today (`drainQuotaOutboxForFile` ignores the count on error; `processQuotaOutboxAvailable` only uses `== batchSize` to re-notify, which is harmless), but returning the total-claimed rather than the actually-applied count is slightly misleading. Consider returning `processed` for accuracy.
- **`AckQuotaOutboxBatchTx` builds a 100-way `(id=? AND receipt=?) OR …` chain.** Correct and placeholder-bound (no injection), but a large OR-of-ANDs. Since all rows in a batch share the same `receipt`, this could be simplified to `receipt = ? AND id IN (…)` — fewer args, cleaner plan. Optional.
- **`sqlPlaceholders` / `id IN (…)`**: parameterized correctly, safe.

### Tests

Good coverage of the new behavior: per-file order preservation (datastore), batch-processes-different-files + central-usage assertion (backend), and the renamed drain test. Gaps worth adding:
- A **concurrent** claim test (two goroutines/"pods" claiming overlapping file sets) to lock in the cross-pod per-file ordering reasoning above — the most subtle correctness property and currently only argued, not tested.
- A batch **partial-failure** test (one entry in a multi-file batch fails apply) asserting the fallback path retries only the failed row and doesn't double-apply the others.

### Security

No new surface. All new SQL is placeholder-parameterized. Same tenant-scoped admission model; the change is purely ordering/batching of already-authorized mutations.

### Verdict

Correct, well-targeted fix for the real bottleneck (write-side serial drain), and it respects the #603 per-file ordering invariant. Approve in principle, pending: (1) confirm the `idx_quota_outbox_file_order` index is created on **existing** tenants via schema repair — this gates the actual perf win; (2) coordinate the `invalidate()`-vs-#610-`add()` merge. Add the concurrent-claim and partial-failure tests when convenient.

mornyx (approved):
Reviewed PR #612 at `efe74976ba9e2d177e721f5f50306272617dc430`. I focused on quota outbox ordering, batch claim/apply/ack semantics, retry/lease behavior, pending quota deltas, and schema coverage. I did not run local tests per review preference and used GitHub CI as the test signal. I did not find any critical/functional blockers.

qiffang:
adversary-2 review on `2a43ab1` / merge `efe7497`.

Focus areas (orthogonal to dev-2's senior review which covers batch fallback / per-file ordering SQL / index migration / concurrent claim test gap):

### Blocker

**B1 — Unrelated `pkg/server/auth.go` change in this PR.**

`pkg/server/auth.go` adds `isForkDeleteRequest` + `handleDeleteNoAcquireAuth` (renamed from `handleTenantDeleteAuth`). PR title is "batch quota outbox processing per file" and PR body says "5 files / +341 / -93" but the auth.go change touches the request routing for `/v1/fork DELETE`. This belongs in #609 ("skip pool.Acquire for fork delete to avoid deadlock") which already merged. New tests `pkg/server/auth_test.go +30 lines` for the fork-delete path also unrelated to quota outbox.

This is scope leak. Either split into a separate PR (preferred — easier to revert independently if either fix introduces a regression), or update the PR description to flag the additional change so reviewers know to verify both.

Refs: `pkg/server/auth.go:248-368`, `pkg/server/auth_test.go` (new tests for `/v1/fork DELETE`).

### High-risk findings

**H1 — Admission-lock window grows linearly with batch size on the per-row fallback path.**

`pkg/backend/quota_outbox_worker.go:228-265` per-row fallback opens N independent `b.metaStore.InTx(...)` calls inside a single `withQuotaAdmissionLock(...)`. The admission lock is held across all N central-DB transactions sequentially. With `quotaOutboxBatchSize = 100`, a single fallback iteration can hold the tenant admission lock for ~100 × (apply tx RTT). At p99 central RTT 50ms that's 5 seconds of admission-lock contention blocking every other writer for the tenant.

The batch path is fine (single central tx). It's the fallback that has this property. Suggested fix: bound the per-row fallback to a smaller chunk (e.g., 10 rows), or move the per-row fallback OUT of the admission lock entirely (each row already has its own lease so they don't depend on the admission serializer for safety — admission lock guards admission, not apply).

Refs: `pkg/backend/quota_outbox_worker.go:230-265`, `pkg/backend/quota_outbox_worker.go:271-282` (withQuotaAdmissionLock).

**H2 — Batch ack `ErrQuotaOutboxLeaseMismatch` semantics under expired lease.**

`pkg/datastore/quota_outbox.go:230-240` (`AckQuotaOutboxBatchTx`): if `rowsAffected != len(entries)` returns `ErrQuotaOutboxLeaseMismatch`. The cause: at least one row's lease expired between claim and ack (lease_until < now). In that case, the admission lock tx rolls back, the central DB commit (separate `b.metaStore.InTx`) is NOT rolled back, and the tenant outbox rows remain "processing" with expired leases. `recoverExpiredLeases` will re-queue them and re-apply.

I verified re-apply is idempotent for `file_create`/`file_overwrite`/`file_delete` via `applyCentralFileStateTx` (`pkg/backend/quota_mutation.go:142-181`) — it reads `old` state via `GetFileMetaForUpdateTx` and computes deltas, so a second apply with the same target state computes zero delta. Same for `DeleteFileMetaIfExistsTx` which returns false on second call.

So the **correctness is OK**, but: between the apply commit and the eventual re-claim → re-apply (idempotent zero-delta), the central usage REFLECTS the change but the tenant outbox status doesn't, and `addLocalQuotaPendingDeltas(-)` is never called for these entries (err != nil path skips the subtraction). Consequence: `quotaPendingCache` carries the +delta longer than necessary, mildly over-counting pending until cache TTL expires (100ms). Bounded but worth a comment.

Refs: `pkg/datastore/quota_outbox.go:184-203`, `pkg/backend/quota_outbox_worker.go:264-272`.

### Medium

**M1 — Index addition on large existing tenants can be slow.**

`pkg/tenant/schema/quota_outbox.go:33` adds `idx_quota_outbox_file_order ON quota_outbox(file_id, status, id)`. The schema-diff framework at `pkg/tenant/schema/tidb_auto.go:2185+` will detect the missing index for existing tenants via `hasObservedTiDBIndex` and emit a repair. TiDB ADD INDEX is online (non-blocking writes), but on a tenant with millions of outbox rows it can take minutes and consume significant cluster CPU.

I verified the spec parse path (`tidbSchemaSpecFromStatements` `pkg/tenant/schema/tidb_auto.go:1750+`) picks up `CREATE INDEX` statements and registers them on the matching table — so the diff repair WILL trigger for existing tenants. ✅

Suggested: add a one-line note in the PR description that existing tenants with backlog will see a one-time ADD INDEX during the next schema-ensure pass, so on-call knows what's running.

Refs: `pkg/tenant/schema/quota_outbox.go:33`, `pkg/tenant/schema/tidb_auto.go:1750+` (spec build), `pkg/tenant/schema/tidb_auto.go:2185+` (diff with observed indexes).

**M2 — `addLocalQuotaPendingDeltas` only fires for `appliedEntries` after a clean batch path or all-good per-row path.**

`pkg/backend/quota_outbox_worker.go:284-298`: the subtraction loop runs only if `err == nil` AND `len(appliedEntries) > 0`. If `batchErr != nil` (i.e., per-row fallback had at least one apply failure), the function returns `(len(entries), batchErr)` and skips both the `quotaUsageCache.invalidate()` AND the per-entry subtract loop.

But wait — `batchErr` only collects the FIRST apply error in the per-row loop. The other rows that DID succeed have their state in `appliedEntries`. The check `err == nil` at line 284 is the admission-lock-tx error, not batchErr. If admission tx returned nil (per-row loop didn't return error, it just collected `batchErr`), then we ENTER the `err == nil` branch and subtract for `appliedEntries`. ✅

OK on re-reading this is fine. Withdraw M2.

### What I checked but didn't find issues with

- SQL injection in `fmt.Sprintf(..., sqlPlaceholders(len(entries)))` and `strings.Join(conds, " OR ")`: both build only `?`/`,` and parameterized conditions, no user data interpolated. ✅
- Per-file ordering subquery `older.file_id = q.file_id OR (older.file_id IS NULL AND q.file_id IS NULL)`: correctly handles NULL file_ids as a single bucket (SQL `NULL = NULL` is not true, but the explicit IS NULL check covers it). ✅
- `FOR UPDATE SKIP LOCKED` correctly replaces the old "no active lease" subquery — active rows are skipped at the lock layer. ✅
- Re-apply idempotency: verified in `applyCentralFileStateTx` — relies on reading old state via `GetFileMetaForUpdateTx` (FOR UPDATE means same-tx visibility) and computing deltas, so identical second apply is a no-op. ✅
- Drain-for-file fallthrough: if `ProcessQuotaOutboxBatch` errors, `drainQuotaOutboxForFile` (`pkg/backend/quota_outbox_worker.go:183-189`) now re-checks `HasPendingQuotaOutboxFileMutation` — if no pending, exits cleanly instead of returning the error. Defensive. ✅

### Verdict

- **B1 must move out of this PR** or be flagged in description. Cleanest action: revert auth.go + auth_test.go changes and open a separate PR; or rebase the auth change onto its actual feature PR (#609 already merged so a follow-up).
- **H1 admission-lock latency** is a real production risk under fallback. Worth a fix or a clear comment that fallback is rare.
- **H2 lease-mismatch path** is correct, just needs a comment in `AckQuotaOutboxBatchTx` and the recovery loop explaining the "central committed but tenant outbox stays processing" intermediate state.
- **M1 index migration** is wired correctly; only need a description note.

Aligned with dev-2's prior senior-review concerns (index migration, #610 conflict). Recommend not merge until B1 split decision is made.

qiffang:
**dev-2 review on `efe74976`**

Independent pass that focuses on angles not already covered by the senior review (qiffang) or @adversary-2: end-to-end verification of the index migration claim, the batch / fallback idempotency interaction with PR #607's `quotaUsageCache`, and the test gap that would lock in the trickiest path.

### 🟢 qiffang's risk #1 (index migration) — verified end-to-end

The existing TiDB schema-repair flow auto-adds the new `idx_quota_outbox_file_order` on already-provisioned tenants. The chain:

1. `QuotaOutboxTiDBSchemaStatements()` adds `CREATE INDEX idx_quota_outbox_file_order ON quota_outbox(file_id, status, id)` (`pkg/tenant/schema/quota_outbox.go:33`).
2. The statement set is appended to both modes — Auto (`pkg/tenant/schema/tidb_auto.go:883`) and App (`pkg/tenant/schema/tidb_app.go:168`) — so `tidbSchemaSpecForModeWithConfig` parses it into `spec.tables[quota_outbox].indexes`.
3. `EnsureTiDBSchemaForAutoEmbeddingProfile` → `ensureTiDBSchemaForModeWithConfig` runs `diffTiDBTable` against the existing table's observed indexes; the new index is emitted as `tidbSchemaDiffMissingIndex` with its `CREATE INDEX …` repair SQL.
4. `isSafeAddIndexRepairSQL` (`pkg/tenant/schema/tidb_auto.go:2716-2723`) accepts the `create index ` prefix → repair pass applies it.

For db9/PostgreSQL the index uses `IF NOT EXISTS` directly, so it's safe by language (`pkg/tenant/db9/schema.go:174`).

No dedicated regression test for *this* index, but the general diff/repair machinery covers it. The perf delta should land on existing tenants, not just newly provisioned ones. Worth a one-line ops note in the PR description (TiDB online DDL on large quota_outbox tables can take minutes), but no code change needed.

### 🟢 qiffang's risk #2 (merge coordination with #610) — moot

PR #610 is **closed, not merged** (per `gh pr view 610`). The post-batch `b.quotaUsageCache.invalidate()` at `pkg/backend/quota_outbox_worker.go:294` is operating against the cache landed in #607 (already merged in `1ba3534`). One `invalidate()` per batch (instead of per row) is strictly less churn against the #607 hot-path optimization — no regression. The "rebase onto #610" note in the senior review can be dropped from the merge checklist.

### 🟢 Correctness reasoning independently re-derived

Walked through the per-file-ordering subquery for the cross-pod two-claimers scenario under READ COMMITTED:

- Pod A starts `claimQuotaOutboxBatch` and locks rows R1, R2 for file F (R1.id < R2.id) via `FOR UPDATE SKIP LOCKED`, flips them to `processing` in the same tx.
- Pod B starts later, evaluates the subquery for a queued row R3 (file F, R3.id > R2.id):
  - Before Pod A commits: Pod B's snapshot sees R1/R2 still `queued`; subquery hits `older.status IN (queued, processing) AND id < R3.id AND file_id = R3.file_id` → R3 blocked.
  - After Pod A commits but before its ack: Pod B sees R1/R2 = `processing`; same subquery blocks R3.
  - After Pod A acks: R1/R2 = `succeeded` → not in the IN-list → R3 unblocked. ✅

The `SKIP LOCKED` only affects the outer `FOR UPDATE` scan, not the EXISTS subquery — so the subquery sees committed status correctly regardless of row-level locks held elsewhere. Matches `TestQuotaOutboxBatchClaimPreservesPerFileOrder` (file-1 overwrite stays queued until file-1 create is acked).

### 🟢 Idempotency boundary under batch → fallback transition

The trickiest path is "batch apply fails partway, fall through to per-entry":

- `applyQuotaOutboxEntries` wraps every `applyCentralFileStateTx` in *one* `metaStore.InTx`. If any single entry's apply errors, the entire tx rolls back — central state is untouched.
- Code then iterates the same `entries` slice and applies each individually in its own `metaStore.InTx`. No entry was successfully committed centrally in the failed batch attempt, so no double-apply.
- And `applyCentralFileStateTx` is already shadow-diff-idempotent for the existing single-row at-least-once window (central commit succeeds → tenant ack tx fails → lease expires → recover → re-apply converges to delta 0), so the fallback path is bounded by the same convergence guarantee.

The ack also uses two separate dbs (central `metaStore.InTx` for apply, tenant `b.store.InTx` for ack inside `withQuotaAdmissionLock`), so the *outer* commit ordering is: central tx commits first, tenant ack tx commits later. Pre-existing semantics, just re-applied at batch granularity. No new failure modes.

### 🟡 Agreeing with the senior review's pinned items

1. **`ProcessQuotaOutboxBatch` returns `len(entries)` on `batchErr != nil`** (line 302): caller `processQuotaOutboxAvailable` uses `processed == quotaOutboxBatchSize` to decide whether to re-notify (line 139). Today's callers handle it benignly, but the semantic mismatch is a sharp edge for future callers. Suggest `return processed, batchErr`.

2. **`AckQuotaOutboxBatchTx` builds a 100-way `(id=? AND receipt=?) OR …` chain** (line 246). Inspecting `claimQuotaOutboxBatch:163` confirms all entries in one batch share the same `receipt` (`receipt := uuid.NewString()` once). So the suggested `receipt = ? AND id IN (…)` is functionally equivalent and cleaner. Optional but I'd take the simplification.

3. **Test gap is the right gap to fill**, specifically the batch-apply-fails-and-falls-through path — currently no regression coverage. Sketch:

   ```go
   func TestProcessQuotaOutboxBatchFallsBackOnApplyError(t *testing.T) {
       // enqueue N>1 entries, inject metaStore failure on the Nth apply,
       // assert: batch tx rolls back (no entries in central),
       // fallback per-entry: first N-1 ack succeeded, Nth retried,
       // no double-apply (central state diff == sum of acked deltas).
   }
   ```

### 🟡 Agreeing with @adversary-2's H1 / H2 / M1

I think B1 (`pkg/server/auth.go` scope leak) is a **false positive** — `gh pr diff 612 --name-only` lists only the 5 quota_outbox files; the `auth.go` change is from PR #609 already merged into main and only appears in `git diff main..HEAD` because the PR has a merge-from-main commit. I've flagged this in #review:thread.

The other three are real:
- **H1** (per-row fallback inside admission lock): real exposure under poison-row scenarios. The fallback runs N independent central round-trips while holding the tenant admission lock. Worth bounding either by chunking the fallback or moving the per-entry retry out of the admission lock.
- **H2** (batch ack lease-mismatch leaves central applied, tenant `processing`): real but bounded by `applyCentralFileStateTx` idempotency. Pending-delta cache stays inflated until recovery + re-apply converges. Order-of-100ms in practice.
- **M1** (ADD INDEX online but slow on large quota_outbox tables on existing tenants): legitimate ops note; doesn't change correctness but should be in the PR description.

### 🟢 NULL file_id branch is dead code today

Tangent: `(older.file_id IS NULL AND q.file_id IS NULL)` in the subquery groups null-file rows into a single ordering chain. Walking enqueue paths confirms only `enqueueQuotaFileCreateOutboxTx` and `enqueueQuotaFileOverwriteOutboxTx` enqueue, both always pass `FileID: fileID` (non-empty). `HasPendingQuotaOutboxFileMutation` filters to `mutation_type IN ('file_create', 'file_overwrite')`; no `file_delete` enqueue path exists. So no production row has NULL file_id today. The NULL branch is reachable only by legacy/test rows.

Two existing datastore tests do exercise it (`TestQuotaOutboxClaimWaitsForDelayedOlderRetry`, `TestQuotaOutboxDeadLetteredOlderRowUnblocksLaterClaim`) and still pass under the new semantics. Worth a one-liner doc comment on `EnqueueQuotaOutboxTx` that production callers always set `FileID`.

### Local checks

- `go build ./...` — clean
- `go vet ./pkg/datastore/... ./pkg/backend/... ./pkg/tenant/schema/...` — clean
- `go test ./pkg/datastore -run 'TestQuotaOutbox' -count=1` — pass (17.9s)
- `go test ./pkg/backend -run 'Test.*QuotaOutbox|TestServerQuota.*Pending|TestServerQuotaReserve|TestServerQuotaOverwrite|TestServerQuotaMedia' -count=1` — pass (21.8s)
- `go test ./pkg/tenant/schema -count=1` — pass (21.2s)
- `go test ./pkg/backend ./pkg/datastore ./pkg/tenant/schema -count=1` — all clean (134s total)

### Verdict

Approve in principle from my lane. The structural fix is right (per-file ordering + batch claim/apply/ack), the correctness reasoning holds under cross-pod READ COMMITTED, the existing schema-repair flow handles the index migration end-to-end, and PR #610's `add()` concern is moot.

Non-blocking from my angle: the three pinned items from the senior review, the NULL-file-id doc comment, and the M1 ops note. H1 (per-row fallback inside admission lock) is the one item I'd want addressed before merge — it's a real worst-case latency exposure on poisoned batches. The batch-apply-fails-falls-back regression test should land alongside.

qiffang:
**Retraction of B1 scope-leak finding from <https://github.com/mem9-ai/drive9/pull/612#issuecomment-4787610983>.**

@dev-2 caught my error. I was diffing `git log` parent of the feature commit (`1ba3534..2a43ab1`) which included the predecessor PR #609's auth.go change. But this PR's actual scope as reported by `gh pr diff 612 --name-only` is only 5 files in `pkg/backend/`, `pkg/datastore/`, `pkg/tenant/schema/` — all quota_outbox. The auth.go change was already merged to main in #609.

**B1 (scope leak) is INVALID and should be disregarded.** Apologies for the noise.

Other findings stand on their own merits:
- **H1** (admission-lock window grows linearly on per-row fallback): still applies — `pkg/backend/quota_outbox_worker.go:230-265` per-row loop is inside `withQuotaAdmissionLock`.
- **H2** (lease-mismatch + re-apply path correctness): still applies — `pkg/datastore/quota_outbox.go:184-203` `AckQuotaOutboxBatchTx`.
- **M1** (index migration on existing tenants): still applies, just an ops note suggestion.

Withdrawing the "do not merge" verdict premised on B1. Final adversary-2 verdict on the merits: **H1 worth fixing, H2/M1 are comments/notes.** No remaining hard blocker from my lane.

Thank you @dev-2 for the catch — exactly the kind of cross-verification this multi-reviewer setup exists for.

qiffang:
Reviewed PR #612 at `efe74976ba9e2d177e721f5f50306272617dc430`.

Verdict from my lane: **no merge-blocking issue found**.

What I checked:

- `ClaimQuotaOutboxBatch` preserves the intended invariant: at most the oldest queued/processing row per `file_id` is claimable, while unrelated files can be claimed in the same batch. The `NULL` file-id path still serializes with other `NULL` file-id rows, which is conservative.
- Cross-pod reasoning looks sound: the selected `q` rows are locked with `FOR UPDATE SKIP LOCKED`, while the same-file predecessor subquery still sees committed queued/processing predecessors, so a later row for the same file cannot jump an in-flight older row.
- `ProcessQuotaOutboxBatch` keeps the previous at-least-once model. Batch apply runs in one central transaction and only batch-acks after apply succeeds; if batch apply fails, central state rolls back and the per-entry fallback is used. If ack fails after central apply, replay remains idempotent through `applyCentralFileStateTx`.
- The pending-delta cache is decremented only for actually acked/applied entries. Processing rows remain counted as pending until ack, which is the safer quota-admission behavior.
- The new `idx_quota_outbox_file_order` is in the TiDB and DB9 quota schema statement lists. For TiDB/app/no-embedding paths those quota statements are included in the base schema statements that are re-executed by schema ensure/repair; duplicate-index errors are ignored by the existing schema executor, so existing tenants should get the missing index on the next schema ensure path. DB9 uses `CREATE INDEX IF NOT EXISTS`.
- Existing upload-target drain semantics remain conservative. It now drains claimable batches rather than one row at a time, and rechecks target pending state after an error before returning success.

Non-blocking notes:

- The PR would be stronger with a true two-claimer regression around the per-file ordering property. The current datastore test pins the single-claimer invariant; cross-pod ordering is still mostly reasoned from SQL semantics.
- `ProcessQuotaOutboxBatch` returns `len(entries)` when fallback applies some rows and returns an apply error. Existing callers are safe with that today, but `processed` would be the more literal return value if future callers use the count for accounting.
- Coordinate with #610's quota usage cache warming if both branches land; this PR currently invalidates once per batch, which is already a big improvement over per-row invalidation.

Validation I ran locally:

- `go test ./pkg/datastore -run 'TestQuotaOutbox' -count=1`
- `go test ./pkg/backend -run 'Test.*QuotaOutbox|TestServerQuota.*Pending|TestServerQuotaReserve|TestServerQuotaOverwrite|TestServerQuotaMedia' -count=1`
- `go test ./pkg/tenant/schema -count=1`
- `go test ./pkg/backend ./pkg/datastore ./pkg/tenant/schema -count=1`
- `go test -race ./pkg/datastore -run 'TestQuotaOutboxBatchClaimPreservesPerFileOrder|TestQuotaOutboxLifecycle' -count=1`

GitHub checks are green (`ci`, local e2e, CodeRabbit) and the PR is merge-state clean.

srstack (commented):
## Re-review: PR #612 — since 08:39 (1 commit: `6b35d37097` "address quota outbox review feedback")

**The one blocking item from my last review is resolved — I verified it directly in the code rather than waiting on confirmation. LGTM, with only a merge-coordination note remaining.**

### ✅ Blocking item resolved: index migration for existing tenants (verified in code)

My prior concern was that `idx_quota_outbox_file_order` would only reach new tenants, leaving the benchmarked existing tenants without the index the per-file ordering subquery depends on. Checked the schema-repair path — **it's handled automatically**:

- A missing index is classified as `tidbSchemaDiffMissingIndex` with `repairSQL: spec.createSQL` (tidb_auto.go ~660).
- `isSafeTiDBRepairDiff` routes `MissingIndex` through `isSafeAddIndexRepairSQL`, which **explicitly accepts plain `CREATE INDEX …`** (tidb_auto.go ~2645).
- So an existing `quota_outbox` table gets `CREATE INDEX idx_quota_outbox_file_order` applied on the next tenant open. The perf win is not lost on existing tenants.

This downgrades from "must confirm before merge" to "confirmed, no action needed."

### ✅ Review feedback addressed

The new commit touches exactly the files my review flagged:
- `pkg/backend/quota_outbox_worker.go` (+115/-59) — batch processing / return-count path
- `pkg/datastore/quota_outbox.go` (+10/-7) — claim/ack adjustments
- `cmd/drive9-server/schema_test.go` (+3) and `pkg/tenant/schema/tidb_auto_test.go` (+31) — cover the index-migration assertion I asked for
- `pkg/backend/quota_outbox_test.go` (+104) and `pkg/datastore/quota_outbox_test.go` (+45) — the concurrent-claim / partial-failure coverage I recommended

(Please skim that the worker changes preserve the earlier reasoning: per-file ordering subquery intact, admission lock still serializes the central apply across pods, batch apply rolls back wholesale on error then falls back to per-entry retry without double-apply. The structure I reviewed at 08:39 held those invariants; the new commit should keep them.)

### 🟡 Remaining (non-blocking): coordinate with #610

#612 still calls `quotaUsageCache.invalidate()`; #610 replaces that with `add(deltas)`. Whichever merges second needs reconciliation. Recommendation: keep `add()` applied **once per batch** with the summed deltas — that gets #612's batched-drain win and #610's warm-cache win together. Pure merge-ordering, not a defect in #612.

### Verdict

LGTM. This is the correct fix for the create-throughput regression — write-side outbox drain goes from tenant-global serial to per-file ordered + batched, while preserving the #603 per-file ordering invariant. The sole hard blocker (index migration) is confirmed covered by the existing schema-repair path. Only the #610 merge coordination remains to settle.

qiffang:
**adversary-1 re-review on `6b35d370977a7c321ef6c431b330a2ce68bc8868`**

I reviewed the delta from the previously reviewed head `efe74976ba9e2d177e721f5f50306272617dc430..6b35d37`.

Verdict from my lane: **no remaining merge-blocking issue found**.

What changed and how it affects the prior review risks:

- The underfilled-batch issue is fixed. `processQuotaOutboxAvailable` now keeps repolling until it has processed a full `quotaOutboxBatchSize` or no more rows are claimable (`pkg/backend/quota_outbox_worker.go:130-149`). This matters for same-file backlogs where a batch can only claim the first row, process it, and then make the next same-file row claimable. New `TestQuotaOutboxWorkerRepollsAfterUnderfilledSameFileBatch` covers that path.
- The per-row fallback no longer runs N independent central applies inside one admission-lock transaction. The code now exits the batch admission-lock transaction after batch apply fails, then calls `processQuotaOutboxEntry` per row, each with its own admission-lock window (`pkg/backend/quota_outbox_worker.go:256-317`, `320-352`). That addresses the prior H1 latency concern.
- `ProcessQuotaOutboxBatch` now returns the literal `processed` count on fallback error instead of `len(entries)` (`pkg/backend/quota_outbox_worker.go:293-317`). This removes the future-caller sharp edge I noted earlier.
- `AckQuotaOutboxBatchTx` now uses the cleaner same-receipt + `id IN (...)` update and rejects mixed receipts before issuing the batch ack (`pkg/datastore/quota_outbox.go:236-249`). This is equivalent for claimed batches because `ClaimQuotaOutboxBatch` assigns one receipt per batch.
- The existing-tenant index repair concern now has direct regression coverage: schema dump includes `idx_quota_outbox_file_order`, and `TestQuotaOutboxMissingFileOrderIndexIsRepairable` verifies a missing observed index generates the safe repair statement.
- The upload-target drain path now continues after an unrelated batch error if the target file is still pending and some entries were processed, with `TestDrainQuotaOutboxForFileContinuesAfterUnrelatedBatchError` covering that case.

One non-blocking note: `ClaimQuotaOutboxBatch` now treats an older `NULL file_id` row as a global ordering barrier (`older.file_id IS NULL OR q.file_id IS NULL`). That is conservative and now covered by `TestQuotaOutboxBatchClaimTreatsNullFileIDAsGlobalBarrier`; production enqueue paths still set `FileID`, so this mainly defines legacy/corrupt-row behavior. A short comment near the claim query would help future readers, but I do not consider it a merge blocker.

Local validation on `6b35d37`:

- `go test ./pkg/datastore -run 'TestQuotaOutbox' -count=1` — pass
- `go test ./pkg/backend -run 'Test.*QuotaOutbox|TestServerQuota.*Pending|TestServerQuotaReserve|TestServerQuotaOverwrite|TestServerQuotaMedia' -count=1` — pass
- `go test ./pkg/tenant/schema -count=1` — pass
- `go test -race ./pkg/datastore -run 'TestQuotaOutboxBatchClaimPreservesPerFileOrder|TestQuotaOutboxBatchClaimTreatsNullFileIDAsGlobalBarrier|TestQuotaOutboxLifecycle' -count=1` — pass
- `go test ./cmd/drive9-server -run 'TestSchemaDumpInitSQLByProvider' -count=1` — pass
- `go test ./pkg/backend ./pkg/datastore ./pkg/tenant/schema -count=1` — pass

GitHub checks are green and merge state is clean. Comment-only per review hygiene; if qiffang wants formal GitHub approval, say so explicitly.

srstack:
Re: remaining #610 coordination note: #610 is closed and was not merged, so there is no active merge-order conflict to coordinate for this PR. #612 is based on the already-merged #607 behavior and invalidates quota usage once per processed batch, which removes the prior per-row churn. If we reintroduce the #610 warm-cache idea later, it should be a separate follow-up that applies summed deltas once per batch on top of this structure.

srstack (commented):
## Correction to my prior re-review — disregard the "#610 merge coordination" note

I need to retract the one remaining item from my last re-review. It was stale:

- **#610 is closed and was never merged**, so there is **no active merge-order conflict** to reconcile. Disregard that note.
- **#612 (on top of the now-merged #607) already calls `quotaUsageCache.invalidate()` once per processed *batch***, not per row — so it's already substantially cheaper than the old per-row invalidation that #610 was chasing. No change needed here.
- If a warm-cache / add-deltas optimization (the #610 idea) is still wanted later, it should be a **separate follow-up** that sums the per-batch deltas and does a single `add()` per batch on top of the current batch structure — not a blocker for this PR.

With that retracted, **#612 has no remaining open items from my review**: the per-file ordering + batch claim/apply/ack is correct, preserves the #603 per-file ordering invariant, and the index-migration concern is confirmed handled by the existing schema-repair path (`isSafeAddIndexRepairSQL` accepts the new `CREATE INDEX idx_quota_outbox_file_order` on existing tenants). **LGTM.**

PR #613, opened 2026-06-24
perf: P1 multi-pod optimizations — DB connections, quota cache, SSE poll

### LINKED ISSUE #599: Make drive9-server stateless and horizontally scalable across multiple pods

mornyx:
## Summary

Addresses all 6 P1 items from #599 in a single PR. These are soft obstacles — the system is correct without them, but they cause cost amplification (TiDB connections, RCU consumption) and slower convergence in multi-pod deployments.

### P1-1: Semantic worker scan amplification — ✅ Already done
Env vars for `PollInterval`, `TenantScanLimit`, `PerTenantConcurrency`, etc. were already exposed on main (`cmd/drive9-server/main.go:770-779`). No work needed.

### P1-2: Configurable DB connection limits
- `DRIVE9_DB_MAX_OPEN_CONNS` / `DRIVE9_DB_MAX_IDLE_CONNS` env vars (default 0=unlimited, backward compatible) in `pkg/mysqlutil/pool.go`
- `DRIVE9_POOL_MAX_TENANTS` env var to override the hardcoded LRU cache size (default 128) in `pkg/tenant/pool.go`
- In multi-pod: set `N_pods × maxOpenConns ≤ TiDB max_connections`

### P1-3: Configurable quota cache refresh interval
- `DRIVE9_QUOTA_CACHE_REFRESH_SECONDS` env var (default 30s) in `pkg/backend/quota_cache.go`
- Reduces per-tenant-per-pod DB reads in multi-pod deployments

### P1-6: SSE per-bus poll (main optimization)
- Moved the cross-pod `fs_events` poll goroutine from **per-SSE-connection** (N tickers = N QPS) to **per-EventBus** (1 ticker = 1 QPS per tenant per pod)
- EventBus starts poll goroutine on first `Subscribe`, stops on last `Unsubscribe`
- When new rows are discovered, calls `Publish()` to signal all local SSE connections via the existing `notify` channel
- SSE connections no longer poll DB directly — they only listen on `notify` + heartbeat

| Scenario | Before (per-connection) | After (per-bus) |
|---|---|---|
| 100 mounts on 1 pod | 100 QPS | 1 QPS |
| 100 mounts on 10 pods | 100 QPS | 10 QPS |

### P1-5: Document mutation queue deprecation
- Added deprecation comment to `pkg/backend/quota_async.go`: the in-process mutation queue is per-instance only. The newer `quota_outbox_worker.go` uses `FOR UPDATE SKIP LOCKED` + lease for multi-pod safe processing. `MutationReplayWorker` (leader-gated since PR #601) provides convergence.

### P1-4: Document per-tenant concurrency is per-pod
- Added doc comment to `tryClaimTenantSlot`: K pods → K×PerTenantConcurrency effective concurrency. This is a capacity limit, not a correctness issue (DB lease guarantees single-owner-per-task).

## Test results

- `pkg/server` SSE + eventbus tests: all pass (including `-race`)
- `pkg/backend` quota cache tests: all pass
- `golangci-lint`: 0 new issues

Closes #599 (P1-1 through P1-6).

## Summary by CodeRabbit

* **New Features**
  * Improved cross-pod live updates by centralizing event delivery and reducing redundant per-connection polling.
  * Added configurable startup controls for quota refresh timing, database connection pool limits, and tenant pool capacity.
* **Bug Fixes**
  * More reliable event discovery and SSE streaming behavior under multi-pod deployments.
* **Documentation**
  * Clarified concurrency/ordering limits for per-tenant processing and the in-process async queue status.
* **Tests**
  * Added coverage for event bus subscribe/unsubscribe and cross-pod event signaling behavior.


qiffang:
adversary-2 review on `979f9e7` (PR #613).

**Scope verified** via `gh pr diff 613 --name-only`: 7 files (3 doc-only). Focus on the actual code paths: `pkg/server/eventbus.go` (P1-6 SSE poll consolidation), `pkg/mysqlutil/pool.go` (P1-2), `pkg/backend/quota_cache.go` + `cmd/drive9-server/main.go` (P1-3).

### Blockers (confirming + extending CodeRabbit's findings)

**B1 — `Unsubscribe` unconditionally waits on `pollWG`, blocking non-last unsubscribers.**

`pkg/server/eventbus.go:95-121`. Per CodeRabbit's 🔴 Critical finding, confirmed independently:

```go
// Stop the poll goroutine when no subscribers remain.
if len(eb.listeners) == 0 && eb.pollCancel != nil { ... }
eb.mu.Unlock()
eb.pollWG.Wait()  // <-- unconditional, runs even if listeners is non-empty
```

Consequences I want to highlight beyond CR's framing:

1. **Non-last unsub blocks until the next pollLoop tick + cancel propagation cycle, IF the pollLoop is still running.** Actually wait — pollWG.Wait() with WG counter > 0 (because pollLoop is still alive serving other subscribers) blocks indefinitely. So an SSE client disconnecting mid-stream while N-1 others are still connected, **its Unsubscribe call blocks until ALL OTHER subscribers also leave**. That's the worst case: a single long-lived sub holds every other sub's Unsubscribe call open.
2. **Worse**: if sub A's Unsubscribe is in the post-unlock Wait, and sub B does Subscribe (taking mu, see pollCancel == nil because A nil'd it… wait, A only nils pollCancel when listeners empty, so this case doesn't apply if B already exists). But if sub A WAS the last sub, A nils pollCancel and calls Wait. Sub B then Subscribes — sees pollCancel == nil, calls `pollWG.Add(1)`, spawns new pollLoop. A's `pollWG.Wait()` now waits for BOTH old (cancelled, will finish) AND new (just started) goroutines → A blocked until B unsubscribes.

Suggested fix: per-generation done channel, capture it under the lock before unlocking, only wait when that specific generation cancelled. Pseudocode:
```go
var doneCh chan struct{}
eb.mu.Lock()
delete(eb.listeners, id)
if len(eb.listeners) == 0 && eb.pollCancel != nil {
    eb.pollCancel()
    eb.pollCancel = nil
    doneCh = eb.pollDone  // captured generation
}
eb.mu.Unlock()
if doneCh != nil { <-doneCh }
```

**B2 — `pollLast` initialized asynchronously inside `pollLoop`, can drop cross-pod events.**

Per CodeRabbit's 🟠 Major finding, confirmed and refined:

`pkg/server/eventbus.go:131-135`: `pollLoop` does `pollLast = eb.unsafeSeq(ctx)` AFTER the goroutine starts, not before `Subscribe` returns. Race window:

```
T0:        Subscribe returns to handler. pollLoop goroutine scheduled.
T0+ε:      handler reads `since` from URL, calls EventsSince → reads up to head=X.
T0+2ε:     external write inserts row with seq X+1.
T0+δ:      pollLoop unsafeSeq reads MAX=X+1, sets pollLast=X+1.
T0+1s:     pollOnce ListFSEventsSince(X+1) → returns empty.
           Event X+1 is silently missed by this subscriber until the NEXT
           event triggers a poll-then-publish that wakes the handler to
           call its own EventsSince(its-lastSeen=X) which catches both.
```

**Concrete consequence**: in an idle tenant where writes are rare, a single new SSE client missing one event can mean a stale FUSE cache for minutes until the next write triggers a poll. The next write WILL recover (because handler's lastSeen catches up via its own EventsSince call), so this is a delay-not-loss bug — but still a regression from PR #602's correctness guarantees.

Suggested fix: initialize pollLast synchronously inside Subscribe under `eb.mu`, BEFORE returning. Pass the cursor into pollLoop explicitly:

```go
eb.pollLast = eb.unsafeSeq(ctx)  // do this while holding mu, before spawning goroutine
ctx, cancel := context.WithCancel(context.Background())
eb.pollCancel = cancel
eb.pollWG.Add(1)
go eb.pollLoop(ctx)
```

Or, more robust: start `pollLast = 0` and rely on `ListFSEventsSince(0)` — but that replays history every first-tick. The "initialize before goroutine spawn" version is the cheapest fix.

### High-risk

**H1 — `unsafeSeq` does a DB call while holding `eb.mu`.**

`pkg/server/eventbus.go:172-185`: `unsafeSeq` (called by `pollLoop` line 134) does `store.LatestFSEventSeq(ctx)` → a TiDB round-trip. Both caller paths hold `eb.mu` (line 132-134). During this DB call, every other Subscribe/Unsubscribe blocks on the mutex.

If we adopt B2's fix (init pollLast before goroutine spawn), this DB call moves to Subscribe's hot path. To avoid blocking concurrent Subscribes, drop the lock around the DB call: snapshot store first, release mu, do the DB call, re-take mu to write pollLast. Or accept that Subscribe is rare enough that the DB-under-lock is fine.

### Medium

**M1 — No new tests for the per-bus poll lifecycle.**

PR description says "all pass (including `-race`)" but `pkg/server/eventbus_test.go` is unchanged. The new behaviors with non-trivial concurrency are:

- Subscribe starts pollLoop on first sub
- Unsubscribe stops pollLoop on last sub
- SetStore (from `eventBuses.get`) while pollLoop is running
- Multiple Subscribe/Unsubscribe interleavings
- pollOnce signaling listeners

None of these have targeted tests. CodeRabbit's two findings would have been caught by a `-race` test that subscribes A and B, unsubscribes A while B still subscribed, and asserts A's call returns in bounded time. Strongly suggest adding such a test before merge — both because the race is real (B1) and because the codebase already has the `TestEventBusStoreRefreshNoRace` pattern from #602.

**M2 — Missing trailing newline in `pkg/mysqlutil/pool.go`.**

`git diff` shows `\ No newline at end of file`. Same nit I flagged on `pkg/server/eventbus.go` in PR #602 (also still open). Cosmetic.

### What I checked and is correct

- `pkg/mysqlutil/pool.go`: `ApplyPoolDefaults` only sets limits when env var > 0. Default 0 = unchanged behavior = backward compatible. ✅
- `pkg/backend/quota_cache.go`: `quotaConfigCacheRefreshInterval` package var initialized via `InitQuotaConfigCacheRefreshInterval` from main before pool creation (`cmd/drive9-server/main.go:124-126`). Go memory model guarantees the goroutines spawned later see the write. ✅
- `pkg/server/semantic_worker.go` and `pkg/backend/quota_async.go`: doc-only changes; both comments are accurate descriptions of the per-pod vs cross-pod ordering invariants. ✅
- atomic.Pointer for `eb.store` (PR #602's N1 fix) preserved: `pollOnce` does `store := eb.store.Load()` consistently. ✅
- `eventBuses.get` SetStore interaction with active pollLoop: pollOnce re-Loads store every tick, so a SetStore() picks up cleanly on next iteration. ✅

### Verdict

Two blockers (B1, B2) directly mirroring CodeRabbit's findings, plus a strong testing gap (M1). The doc changes and env-var additions are clean and backward-compatible.

**Recommend: not merge until B1 + B2 fixed + targeted -race test added.** The mutex-under-DB-call (H1) can land separately if B2 is fixed minimally.

Aligned with dev-2's lay-of-the-land call. @adversary-1 likely landing their own delta.

qiffang:
**dev-2 review on `979f9e7e`**

Independent pass; adversary-2 has already shipped the comprehensive blocker list. My delta:

### 🔴 Independently confirm CodeRabbit + adversary-2 B1 (Unsubscribe blocks non-last subscribers) — sharpening severity

CodeRabbit and adversary-2 both flagged this. Re-derived the trace to sharpen the description:

```go
func (eb *EventBus) Unsubscribe(id uint64) {
    eb.mu.Lock()
    delete(eb.listeners, id)
    if len(eb.listeners) == 0 && eb.pollCancel != nil {
        eb.pollCancel()
        eb.pollCancel = nil
    }
    eb.mu.Unlock()
    eb.pollWG.Wait()  // ← blocks unconditionally
}
```

`pollWG.Add(1)` is called exactly once in `Subscribe` (when `pollCancel == nil`). Counter stays at 1 while poll runs. So:

- **Non-last unsubscriber**: doesn't cancel poll; `pollWG.Wait()` blocks until last subscriber leaves
- **In SSE handler**: `defer bus.Unsubscribe(subID)` (sse.go:209) means **client disconnect blocks the HTTP handler goroutine until all other concurrently-connected clients disconnect**

This is **not an instant deadlock** but **goroutine serialization** — every premature disconnect holds the SSE handler goroutine until tail-drain. Under N concurrent mounts with normal disconnect churn, you accumulate `~N - 1` stuck handler goroutines until the last one leaves. The HTTP server itself stays responsive (new connections get new goroutines), but you bleed goroutines proportional to peak concurrency × churn rate.

**Why `TestEventBusMultipleSubscribers` doesn't catch this**: all 5 subscribers receive the single `Publish()` signal in parallel and race to unsubscribe simultaneously. When the last one cancels poll, `pollWG.Done()` fires and **all 4 blocked `Wait()` calls wake up at once** (WaitGroup semantics — `Wait` returns when counter reaches 0, all waiters wake together). The test thus exercises the buggy path but the failure mode is invisible at this scale. A test where N-1 subscribers stay connected and only 1 leaves would expose the wait. **`TestEventBusUnsubscribeOneOfManyReturnsImmediately` would be the regression test that pins this.**

Fix shape: per-subscriber done channel OR move `pollWG.Wait()` inside the `len == 0` branch so only the last unsubscriber waits.

### 🔴 Confirm B2 (pollLast init race) — sharpening severity to "permanent gap during quiescent periods"

CodeRabbit flagged "events could be skipped"; adversary-2 framed it as "delay-not-loss". I think it's **closer to delay-not-loss in normal traffic, but permanent gap in quiescent periods**:

Race trace:
- T0: SSE handler `Subscribe()` → starts pollLoop goroutine, releases mu
- T1: SSE handler `EventsSince(since=0)` → returns headSeq=H1, sets `lastSeen=H1`
- T2: Another pod inserts row at seq=H2 (H2 > H1)
- T3: pollLoop finally acquires mu, calls `unsafeSeq(ctx)` → returns H2; sets `pollLast=H2`
- T4: `pollOnce` ticks → `ListFSEventsSince(H2)` → no rows → **no notify**
- T5: SSE handler is at `lastSeen=H1`, never woken; **client misses row H2 until next write**

In an active workload another write at H3 > H2 catches it (delay-not-loss). In a quiescent workload (workspace idle, only this one cross-pod write), **the client never sees H2** for the lifetime of the connection. FUSE HEAD revalidation eventually catches the actual file state, but the SSE invariant ("cross-pod writes propagate via SSE within ~1s") is broken in the quiescent edge.

Fix shape: initialize `pollLast` synchronously in `Subscribe` before returning. Simpler still: set `pollLast = 0` and let the first `pollOnce` query everything (bounded by the 1000-row limit). Each connection's `EventsSince(lastSeen)` is the authoritative delivery; pollLoop is just the wake-up signal — over-notifying is harmless, missed-notify is not.

### 🟡 H1 (adversary-2) — `unsafeSeq` does DB call under `eb.mu`

Confirmed. `pollLoop`'s init block (eventbus.go:133-135) holds `eb.mu` across `unsafeSeq(ctx)` which does a DB RTT. Any concurrent `Subscribe`/`Unsubscribe` blocks for that RTT. Not catastrophic (RTT is ms-scale) but a clear violation of the "no DB under mutex" rule. Fix as part of B2 (move init out of pollLoop, or release mu around the DB call).

### 🟡 dev-2-lane addition — `SetStore` × pollLoop ordering not documented

The PR inherits PR #602's `atomic.Pointer[*datastore.Store]` + `SetStore()` plumbing. With per-bus pollLoop now long-lived:

- `eventBuses.get` calls `bus.SetStore(newStore)` when pool recreates backend (sse.go:135). The **old store** may be closed by the pool right after.
- pollLoop's next `pollOnce` loads the new store (`eb.store.Load()`), so it's safe.
- **But**: if `pollOnce` is mid-tick and already executed `store := eb.store.Load()` (line 153) **just before** `SetStore` swaps, then `store.ListFSEventsSince(...)` (line 161) runs against the now-closed old store. The error is silently swallowed at line 162-164 (`return` with no log).

PR #602 had the same exposure window (`EventsSince` could be called immediately after `SetStore` with stale ref), but the window was a single in-flight request, not a 1s-ticker loop. With per-bus polling, **every tick is an opportunity for stale-store contact during pool churn**.

Two minor asks:
1. Log the `pollOnce` error path (eventbus.go:163) — currently silent return makes pool-churn-induced stale-store hits invisible to operators
2. Doc comment on `SetStore` explicitly noting "pollLoop may still hold a reference to the previous store for one tick" — sets the contract correctly for future readers

Neither is a blocker; both improve observability.

### 🟡 M1 (adversary-2) — no targeted test for new lifecycle

Confirmed. `pkg/server/eventbus_test.go` not touched in this PR. PR body claims "-race pass" for the existing tests, but the new lifecycle (`pollCancel`, `pollWG`, `pollLast`) has no targeted coverage. The two race tests added in PR #602 (`TestEventBusStoreRefreshNoRace`) and PR #602 (`TestEventBusEventsSinceQueryErrorReturnsCaughtUp`) only cover those PR's surfaces, not this one.

Three tests I'd want before merge (all using the existing `newTestStoreForEventBus` harness):

```go
TestEventBusUnsubscribeOneOfManyReturnsImmediately  // pins B1 fix
TestEventBusPollLastInitNoSkipUnderRace             // pins B2 fix
TestEventBusPollLoopStopsAfterLastUnsubscribe       // existing claim, no test
```

### 🟢 Non-eventbus changes — verified

- `pkg/mysqlutil/pool.go`: env-var defaults (`DRIVE9_DB_MAX_OPEN_CONNS`, `DRIVE9_DB_MAX_IDLE_CONNS`) at `0` → no change from current behavior unless operator sets them. CodeRabbit's nit about idle/open disconnect is legitimate but a doc fix, not a blocker. ✅
- `pkg/tenant/pool.go`: `DRIVE9_POOL_MAX_TENANTS` defaults to existing hardcoded `128` → ✅
- `pkg/backend/quota_cache.go`: `DRIVE9_QUOTA_CACHE_REFRESH_SECONDS` defaults to existing `30` → ✅
- `pkg/backend/quota_async.go`: doc-only deprecation comment → ✅
- `pkg/server/semantic_worker.go` + `cmd/drive9-server/main.go`: doc comment on per-pod concurrency → ✅

### Local checks

- `go build ./...` — clean
- `go vet ./pkg/server/... ./pkg/backend/... ./pkg/mysqlutil/... ./pkg/tenant/...` — clean
- `go test -race ./pkg/server -run TestEventBus -count=1` — **pass**, but per the B1 analysis above, `TestEventBusMultipleSubscribers` passes by *luck* (parallel unsubscribe collapse), not by exercising the bug

### Verdict

dev-2 lane: **B1 + B2 are real merge blockers**, but B1's severity is goroutine-serialization-leak (not instant deadlock) — still a regression vs main where SSE handler exit is decoupled from sibling subscribers. B2 has a quiescent-period gap that breaks the SSE propagation invariant. Both have the same code locus and one Subscribe-side refactor should close both. M1 (targeted tests) lands with the fix.

Non-eventbus changes are clean; I'd be happy to merge them as a separate PR if there's pressure to land the env-var/quota-cache wins independently.

qiffang:
**adversary-1 review on `979f9e7e`**

I did an independent pass over the PR scope and focused on the new per-`EventBus` polling lifecycle, since that is the only behavior-changing part with correctness risk. I agree with CodeRabbit / adversary-2 / dev-2: **do not merge this head yet**.

### Blocker 1: non-last `Unsubscribe` blocks until every other subscriber leaves

`pkg/server/eventbus.go:107-121` unconditionally calls `eb.pollWG.Wait()` after removing one listener. If other listeners remain, the poll goroutine is still intentionally alive, so this request handler blocks until the last sibling subscriber disconnects and cancels the poll loop.

I verified this with a temporary regression test in my checkout:

```go
id1, _ := bus.Subscribe()
id2, _ := bus.Subscribe()

done := make(chan struct{})
go func() {
    bus.Unsubscribe(id1)
    close(done)
}()

select {
case <-done:
case <-time.After(100 * time.Millisecond):
    bus.Unsubscribe(id2)
    <-done
    t.Fatal("Unsubscribe blocked while another listener remained subscribed")
}
```

It fails on this head:

```text
--- FAIL: TestPR613UnsubscribeDoesNotBlockWithRemainingSubscribers (0.16s)
    eventbus_pr613_tmp_test.go:26: Unsubscribe blocked while another listener remained subscribed
```

That means normal SSE disconnect churn can accumulate stuck HTTP handler goroutines proportional to connected mounts. The fix should make `Unsubscribe` wait only when this call actually cancels the current poll generation. A captured generation `done` channel is safer than sharing one `WaitGroup`, because a new subscriber can start a new poll generation while the old generation is being waited on.

### Blocker 2: async `pollLast` initialization can miss the only cross-pod event after initial replay

`pkg/server/eventbus.go:128-135` initializes `pollLast` inside the goroutine after `Subscribe` has already returned. `handleEvents` then does the authoritative initial `EventsSince` replay at `pkg/server/sse.go:220-253`, but there is a race window:

```text
Subscribe returns
SSE initial replay reads through head H1 and sets lastSeen=H1
other pod inserts event H2
pollLoop finally runs unsafeSeq(), sees H2, sets pollLast=H2
next poll queries since H2, gets no rows, sends no notify
SSE handler remains at lastSeen=H1 forever unless a later event wakes it
```

In an active tenant the next write eventually wakes the connection and it catches up, but in a quiescent tenant this breaks the intended “cross-pod events are discovered by periodic polling” invariant for the lifetime of that SSE connection. The poll loop can over-notify safely because each handler still calls `EventsSince(lastSeen)` before sending; it must not miss a wake-up.

Fix shape: initialize the poll generation cursor synchronously before `Subscribe` returns, or start from a conservative cursor and allow harmless over-notification. Avoid doing the DB `LatestFSEventSeq` round trip while holding `eb.mu` if possible.

### Tests needed with the fix

Existing tests pass, but they do not cover the new lifecycle. In particular, `TestEventBusMultipleSubscribers` hides blocker 1 because every subscriber unsubscribes after the same `Publish`, so the last unsubscribe cancels the poll loop and releases all waiters.

Please add targeted tests for at least:

- one-of-many `Unsubscribe` returns immediately;
- poll cursor initialization does not skip a cross-pod event between initial replay and poll loop startup;
- poll loop stops after the last unsubscribe and can restart cleanly on a later subscribe.

### Verified OK

- DB pool env vars are opt-in (`>0`) and keep old defaults.
- quota config refresh interval defaults to the existing 30s and is initialized before tenant backends are created in `main`.
- doc-only comments in `quota_async.go` and `semantic_worker.go` match the current multi-pod model.
- PR #602's `atomic.Pointer[*datastore.Store]` pattern is preserved in `pollOnce`.

### Local validation

- `go test ./pkg/server -run 'TestEventBus|TestSSE' -count=1` — pass
- `go test -race ./pkg/server -run 'TestEventBus|TestSSE' -count=1` — pass
- `go test ./pkg/backend -run 'TestQuota.*Cache|TestQuota' -count=1` — pass
- `go test ./pkg/mysqlutil ./pkg/tenant -count=1` — pass
- `go test ./pkg/server ./pkg/backend ./pkg/mysqlutil ./pkg/tenant -count=1` — pass

GitHub checks are green on `979f9e7e`, but the two EventBus lifecycle bugs above are still merge blockers.

mornyx:
Thanks @qiffang for the thorough B1+B2+M1 review from all three lanes! All addressed in d457b75.

### B1 — Unsubscribe blocks non-last subscribers ✅ fixed

Replaced the shared `WaitGroup` with a **per-generation done channel**:
- `Subscribe` creates a `done chan struct{}` and passes it to `pollLoop` as a parameter
- `pollLoop` defers `close(done)` on exit
- `Unsubscribe` only waits on `done` when it actually cancels the poll goroutine (last subscriber). Non-last Unsubscribers return immediately.

This eliminates goroutine serialization: a disconnecting SSE client no longer blocks its HTTP handler goroutine until all sibling subscribers leave.

### B2 — pollLast initialized asynchronously ✅ fixed

Moved `pollLast` initialization from `pollLoop` (async, after goroutine schedule) into `Subscribe` (synchronous, before goroutine starts). The `LatestFSEventSeq` DB call happens while holding `eb.mu` in `Subscribe` — this is acceptable because Subscribe is rare (once per SSE connection establishment). Removed the now-unused `unsafeSeq` function.

This closes the race window where a cross-pod event inserted between `Subscribe` returning and `pollLoop` starting could be permanently missed in quiescent tenants.

### H1 — unsafeSeq DB call under eb.mu ✅ resolved

The DB call now happens in `Subscribe` (before goroutine spawn), not in `pollLoop` under `eb.mu`. `Subscribe` is rare enough that a DB round-trip under the mutex is acceptable. The hot path (`pollOnce`) never holds `eb.mu` across a DB call.

### pollOnce error logging (dev-2 request) ✅ added

`pollOnce` now logs `event_bus_poll_failed` on query errors so operators can detect pool-churn-induced stale-store hits (previously silently swallowed).

### M1 — Targeted regression tests ✅ added

Three new tests:
- **`TestEventBusUnsubscribeOneOfManyReturnsImmediately`**: subscribes A+B, unsubscribes A while B remains, asserts A returns within 1s (pins B1 fix)
- **`TestEventBusPollLoopStopsAfterLastUnsubscribe`**: verifies poll stops after last unsubscribe and restarts cleanly on resubscribe
- **`TestEventBusPollLoopSignalsCrossPodEvents`**: inserts a cross-pod fs_events row directly (bypassing local notify), asserts the per-bus poll goroutine discovers it within 3s and signals subscribers

All tests pass with `-race`.

srstack (commented):
## Review: feat/p1-multi-pod-optimizations (PR #613)

**Overall: solid, well-scoped multi-pod cost-reduction PR. The headline SSE per-bus poll refactor is correctly designed and the ordering hazards are handled. A few concurrency details and one lifecycle edge worth confirming. Close to LGTM.**

### Overview

Bundles the 6 P1 items from #599: configurable DB conn limits + pool size (P1-2), configurable quota-cache refresh (P1-3), the SSE per-connection→per-bus poll refactor (P1-6, the main change), plus deprecation/clarity doc comments (P1-4, P1-5). The SSE change drops idle cross-pod poll QPS from N-mounts to 1-per-bus-per-pod.

### ✅ Correctness — verified the key hazards

- **Subscribe-before-replay ordering is correct**: `handleEvents` calls `bus.Subscribe()` (sse.go:208) *before* the Phase-1 `EventsSince` replay (sse.go:221). So any cross-pod row arriving during/after replay triggers a notify → `pollAndSend` re-checks. No gap between replay and live stream. ✅
- **No missed events from `pollLast` racing ahead**: the per-bus poll loop's `Publish()` is a pure wakeup; each SSE connection re-queries from its **own** `lastSeen` cursor via `EventsSince(lastSeen)`. So even if `pollLast` advances past what a given connection has sent, the connection still reads from its cursor and loses nothing. Clean separation. ✅
- **`pollLast` init in Subscribe** is done synchronously under `eb.mu` before the goroutine starts (with the documented rationale), closing the init-vs-loop-start window. ✅
- **Unsubscribe wait semantics**: only the last unsubscriber captures `pollDone` and waits; non-last return immediately. The B1 regression test (`...UnsubscribeOneOfManyReturnsImmediately`) pins this. ✅

### 🟠 Concurrency / lifecycle items to confirm

1. **Last-unsubscribe vs. re-subscribe race on the poll goroutine.** Unsubscribe nils `pollCancel`/`pollDone` under the lock, then waits on the captured `doneCh` **after releasing the lock** (eventbus.go:229-234). If a new `Subscribe()` arrives in that window, it sees `pollCancel == nil` and starts a *new* poll goroutine while the old one is still draining. Two poll goroutines briefly coexist for the same bus. They're both correct (idempotent wakeups, each reads `pollLast` under the lock), so it's not a correctness bug — but it's a transient extra goroutine and an extra DB poll. Worth a comment acknowledging it, or holding a "starting" flag. Confirm it can't leak under rapid subscribe/unsubscribe churn.

2. **`pollLast` is per-bus, shared across generations.** After the poll loop stops and restarts (last unsub → new sub), `pollLast` retains its old value, and Subscribe re-initializes it to the current head seq. Fine — but note that across a stop/start, `pollLast` is reset to head, so any rows written *while no subscriber existed* are not signaled by the poll loop. That's correct (no listeners to signal, and a fresh Subscribe does its own Phase-1 replay from the client's `since`), just confirm that's the intended contract.

3. **`pollOnce` stale-store on pool churn** is documented (the comment is good), and it logs + skips rather than resetting. Confirm that a persistent stale store (backend evicted but bus still has subscribers) eventually recovers — i.e. that `eb.store` is refreshed by `eventBuses.get` on the next backend recreate, not left pointing at a closed DB indefinitely.

### 🟡 Minor

- **`pkg/mysqlutil/pool.go` missing EOF newline** (`\ No newline at end of file`). Trivial, but most of the repo ends files with a newline.
- **Two `envInt` helpers now exist** — the new one in `mysqlutil/pool.go` and the existing one in `cmd/drive9-server/main.go`. Minor duplication; acceptable since they're in different packages, but worth noting.
- **`eventBusPollInterval` (1s) is now hardcoded in eventbus.go** while the old `ssePollInterval` (also 1s, now removed) was likewise fixed. The comment says "Override via ssePollInterval in sse.go" but that const was deleted — the comment is stale. Since P1-3 made the quota interval configurable, consider making the event-bus poll interval env-configurable too (or fix the comment).
- **Env vars default to 0=unlimited/default** consistently and are backward-compatible — good. `DRIVE9_DB_MAX_OPEN_CONNS` guidance (`N_pods × maxOpenConns ≤ TiDB max_connections`) is helpful in the comment.

### Tests

Good additions for the SSE refactor: unsubscribe-returns-immediately (B1), poll-stops-then-restarts, and cross-pod-event-discovery (inserts directly into fs_events, bypassing local notify, asserts the poll loop signals). These cover the core new behavior including the restart path. Gaps:
- No test for the **last-unsub vs re-sub race** (item 1) — hard to write deterministically, but even a stress/loop test under `-race` would help.
- No test for the **DB-conn-limit / pool-max-tenants env knobs** (P1-2) — they're simple setters, but a test that `DRIVE9_POOL_MAX_TENANTS` actually overrides the LRU cap would prevent silent regressions.
- The PR says `-race` passes on the SSE/eventbus tests — good, given the goroutine lifecycle changes.

### Security

No new surface. All new knobs are server-side env vars with safe defaults; the SSE change is purely an internal polling-topology optimization (same data, same tenant-scoped bus). No auth/credential impact.

### Verdict

Well-targeted P1 work; the SSE per-bus refactor is the substantive change and its correctness-critical ordering is handled correctly. I'd confirm the three concurrency/lifecycle items above (especially #1, the last-unsub/re-sub goroutine race) before merge — none are blocking correctness, but they're the kind of thing that bites under churn. Fix the stale `ssePollInterval` comment and the EOF newline on the way through.

mornyx:
Thanks @srstack for the detailed review! All items addressed in 7e371e8.

### Concurrency/lifecycle items

**Item 1 (last-unsub vs re-sub race)**: Added a concurrency note on `Unsubscribe` documenting the transient dual-goroutine window. Both goroutines are safe (idempotent wakeups, each reads `pollLast` under the lock) and the old goroutine exits within one tick interval. No correctness risk under rapid churn.

**Item 2 (pollLast per-bus across generations)**: Added a contract note on `Subscribe` documenting that events written while no subscriber existed are not signaled by the poll loop. This is correct — a fresh Subscribe's Phase-1 `EventsSince` replay from the client's `since` cursor is the authoritative delivery path for those events.

**Item 3 (stale-store recovery on pool churn)**: Confirmed — `eb.store` is refreshed by `eventBuses.get` on the next backend recreate (via `SetStore` with `atomic.Pointer`). The `pollOnce` error logging (added in the previous commit) makes stale-store hits visible to operators. Recovery happens on the next `tenantEventBus` call that resolves a valid backend.

### Minor items

- **Stale `ssePollInterval` comment**: fixed — the comment now says "Matches the fs_layer_events poll interval" instead of referencing the deleted const.
- **EOF newline on `pool.go`**: added.
- **Duplicate `envInt` helpers**: acknowledged — acceptable since they're in different packages (`mysqlutil` vs `cmd/drive9-server`), and `mysqlutil` can't import the cmd package.

### Items I'm not changing (with rationale)

- **Event-bus poll interval not env-configurable**: keeping it hardcoded at 1s for now. P1-3 made the quota cache interval configurable because it's a per-tenant-per-pod cost (N×M goroutines). The event-bus poll is per-bus (already optimized from per-connection to per-bus in this PR), so 1 QPS per tenant per pod is acceptable. Can be made configurable in a follow-up if needed.
- **No test for env knobs (P1-2)**: these are simple `SetMaxOpenConns`/`SetMaxIdleConns` calls gated by `> 0` checks with safe defaults. A dedicated test would need to verify `sql.DB.Stats().MaxOpenConnections` which is fragile. Documenting the env vars in the help text (`cmd/drive9-server/main.go:506-513`) is sufficient for now.

qiffang:
Re-reviewed current head `7e371e85090f3d4bda118dcb1e639369fba00d79` after the follow-up fixes.

Verdict from my lane: **no remaining merge-blocking issue found**.

What changed relative to my prior blocker comment on `979f9e7`:

- **B1 fixed**: `Unsubscribe` now captures a per-generation `pollDone` channel only when it is the last subscriber and actually cancels the current poll loop. Non-last unsubscribers return without waiting. `TestEventBusUnsubscribeOneOfManyReturnsImmediately` covers the old stuck-handler path.
- **B2 fixed**: `pollLast` is initialized synchronously before the poll goroutine starts, closing the window where a cross-pod row could land after initial replay but before poll-loop cursor initialization. `TestEventBusPollLoopSignalsCrossPodEvents` covers direct `fs_events` insertion without local publish, and `TestEventBusPollLoopStopsAfterLastUnsubscribe` covers stop/restart.
- The srstack lifecycle notes are acknowledged in comments: transient dual poll goroutines during last-unsub/re-sub churn are bounded and safe because wakeups are idempotent and `pollLast` remains lock-protected.
- Stale `ssePollInterval` comment and `pool.go` EOF newline are fixed.

Residual non-blocking note: `Subscribe` still performs `LatestFSEventSeq(context.Background())` while holding `eb.mu`. That is acceptable for this P1 optimization because it only runs on first subscriber / restart and keeps the ordering simple. If subscribe latency under DB stalls becomes visible, the follow-up shape is to pass a request context and fetch the initial cursor outside the mutex, then re-check generation state under the lock before starting the poller.

Validation I ran on this head:

```text
go test ./pkg/server -run 'TestEventBusUnsubscribeOneOfManyReturnsImmediately|TestEventBusPollLoopStopsAfterLastUnsubscribe|TestEventBusPollLoopSignalsCrossPodEvents' -count=1
go test -race ./pkg/server -run 'TestEventBusUnsubscribeOneOfManyReturnsImmediately|TestEventBusPollLoopStopsAfterLastUnsubscribe|TestEventBusPollLoopSignalsCrossPodEvents' -count=1
go test ./pkg/server -run 'TestEventBus|TestSSE' -count=1
go test -race ./pkg/server -run 'TestEventBus|TestSSE' -count=1
go test -race ./pkg/server -run TestEventBus -count=3
go test ./pkg/backend -run 'TestQuota.*Cache|TestQuota' -count=1
go test ./pkg/mysqlutil ./pkg/tenant -count=1
go test ./pkg/server ./pkg/backend ./pkg/mysqlutil ./pkg/tenant -count=1
```

GitHub checks were also green on `7e371e8` when I checked (`ci`, `drive9-server-local e2e smoke`, CodeRabbit status). CodeRabbit's latest detailed review comment itself was rate/credit limited, so I did not rely on it for the verdict.

qiffang (approved):

qiffang (commented):
## dat9-dev2 review — GREEN

I did an independent pass over PR #613 (`7e371e8`) focusing on the multi-pod behavior, EventBus/SSE lifecycle, env-driven DB/pool limits, and quota config refresh semantics.

### Reviewed
- `EventBus` now has one poll goroutine per tenant bus instead of one DB poll per SSE connection. `Subscribe` initializes `pollLast` before starting the goroutine; `Unsubscribe` only waits for the last subscriber. This preserves delivery semantics while reducing idle DB polling.
- `sse.go` correctly relies on `notify` from same-pod publishes and per-bus cross-pod polling; initial replay/reset still uses durable `fs_events` via `EventsSince`.
- `eventBuses.get` refreshes the store pointer when the tenant backend is recreated, avoiding stale-store usage for future bus reads.
- `DRIVE9_QUOTA_CACHE_REFRESH_SECONDS`, `DRIVE9_POOL_MAX_TENANTS`, `DRIVE9_DB_MAX_OPEN_CONNS`, and `DRIVE9_DB_MAX_IDLE_CONNS` are opt-in and keep existing default behavior when unset/invalid.
- The in-process quota mutation queue and semantic worker concurrency comments correctly document the remaining per-pod/non-global semantics.

### Validation
- `/usr/local/go/bin/go test ./pkg/mysqlutil ./cmd/drive9-server -count=1` ✅
- `/usr/local/go/bin/go test -c ./pkg/server ./pkg/backend` ✅ compile gate
- Full `pkg/server` runtime tests are locally blocked by testcontainers/rootless Docker on this machine; GitHub `ci` and `drive9-server-local e2e smoke` are green.

### Non-blocking notes
- The brief old/new poll goroutine coexistence during last-unsubscribe/re-subscribe is acceptable; wakeups are idempotent and SSE `lastSeen` prevents duplicate event delivery.
- `InitQuotaConfigCacheRefreshInterval` is a package-level startup hook; the existing “call before backend creation” contract is important and should remain.
- The duplicated `envInt` helper in `pkg/mysqlutil` is stylistic only.

No blockers from my pass.

PR #618, opened 2026-06-24
Add cloud native admin tenant APIs

srstack:
## Summary
- add cloud native admin tenant APIs backed by TiDB Cloud managed-cluster authorization and tenant/org binding metadata
- move quota CLI under `drive9 admin`, remove old top-level quota command and admin quota get, and keep quota reads through tenant list/get
- support admin quota set/create-time quota and update docs, SDK helpers, and tests

## Tests
- `GOCACHE=/tmp/drive9-go-cache go test ./cmd/drive9/cli -run 'TestAdmin|TestQuota' -count=1`\n- `GOCACHE=/tmp/drive9-go-cache go test ./pkg/client -run 'Test.*Quota|TestAdmin' -count=1`\n- `GOCACHE=/tmp/drive9-go-cache go test ./pkg/server -run 'TestAdminTenant|TestQuota' -count=1`\n- `GOCACHE=/tmp/drive9-go-cache go test ./examples/go-sdk-cookbook -run 'ExampleClient_quota' -count=1`\n- `GOCACHE=/tmp/drive9-go-cache go test ./cmd/drive9 -run 'TestDispatchAdminVerbReachesHandler|TestDispatchQuotaVerbIsRejected|TestDispatchLongHelpFlagShowsUsage|TestDispatchHelpCommandShowsVisualTreeHelp' -count=1`

## Summary by CodeRabbit

* **New Features**
  * Added `drive9 admin tenant` command with `create`, `list` (optional quota inclusion), `get`, `delete`, and `set-quota`.
  * Added TiDB Cloud Mode admin tenant quota management, including managed-cluster lookup to enrich quota details.
  * Added `--tidbcloud-spending-limit` support for tenant creation (with sensible defaults/validation).
* **Bug Fixes**
  * Standardized non-JSON quota display to a consistent table; improved quota-set error messaging and formatting.
* **Documentation**
  * Updated CLI help, guides, and Go SDK parity/integration docs to use the admin workflow.
* **Tests**
  * Expanded CLI and server/client tests for tenant admin operations, JSON/table output, pagination, and spending-limit hints.


srstack (commented):
## Review: Add cloud native admin tenant APIs (PR #618)

**Overall: substantial, well-structured feature. The authorization model is sound — every admin op is gated on TiDB-Cloud-credential ownership of the tenant's bound cluster. Two real concerns (existing-tenant org-binding backfill; per-request full cluster enumeration cost) and a few smaller items. Strong PR, close to LGTM pending the backfill question.**

### Overview

Adds `/v1/admin/tenants` APIs (create / list / get / delete / quota-set) and moves the quota CLI under `drive9 admin`, removing the old top-level `quota` command and `admin quota get`. Quota reads now flow through tenant list/get. Authorization is via TiDB Cloud managed-cluster listing + a new `tenant_tidbcloud_org_bindings` meta table mapping tenant → (org, cluster).

### ✅ Authorization model — verified sound

This is the security crux, and it's done right:
- `authorizedAdminTenant` (admin_tenants.go) loads the tenant, requires `ProviderTiDBCloudNative`, rejects `Deleted`, loads the org binding, then calls `listAllManagedClusters(cred, binding.ClusterID)` with the **caller's** TiDB Cloud keys and authorizes only if a returned cluster matches **both** `ClusterID` AND `OrganizationID` of the binding. So a caller can only act on tenants whose backing cluster their TiDB Cloud API key can actually see. ✅
- List derives `orgIDs` from the caller's managed clusters and filters tenants to those orgs (`ListTenantsByTiDBCloudOrganizations`) — no cross-org leakage. ✅
- Credentials come from body or `X-TiDBCloud-*` headers via `adminCredentials` → `quotaCredentials` (rejects empty/partial). Consistent with the existing quota API. ✅
- `adminTenantAPIEnabled` gates the whole surface on native provider + configured meta/pool/token. ✅

### 🟠 Existing native tenants have no org binding → 404 from admin API

The binding is written via `UpsertTenantTiDBCloudOrgBinding`, which (from the diff) is populated at **provision time** in the new admin-create flow. `authorizedAdminTenant` returns `404 "tenant tidbcloud organization binding not found"` when no binding row exists. So **native tenants created before this PR** (or via the non-admin provision path) won't have a binding and become **invisible to list and unmanageable via get/delete/quota-set** through the admin API. Please confirm:
- Is there a backfill for existing native tenants' org bindings? If not, they're stranded from admin management.
- Does the *non-admin* native provision path (#586/#606 fork/create) also write the binding, or only the new admin-create? If only admin-create, tenants made the old way are excluded.

This is the one item I'd resolve before merge — it determines whether the feature works for the existing fleet or only net-new admin-created tenants.

### 🟠 Per-request full managed-cluster enumeration (cost / latency)

`listAllManagedClusters` pages through **all** of the caller's managed clusters (PageSize 100, full loop) on every admin request:
- For `get`/`delete`/`quota-set` it passes `binding.ClusterID` as a filter — good, that's a targeted lookup.
- For **`list`** it passes `""` — so every tenant-list call enumerates the org's entire cluster inventory before the DB page query. For a large org this is N TiDB Cloud API round-trips per list request, and it gates the response. Inherent to deriving orgIDs from credentials, but worth: (a) a metric/log on cluster-count + enumeration time, and (b) considering a short cache of cred→orgIDs to avoid re-enumerating on pagination clicks. Not blocking, but it's the scalability hot spot.

### 🟡 Migration

The `tenant_tidbcloud_org_bindings` table is created via `CREATE TABLE IF NOT EXISTS` and there's a meta-schema test (`mustMetaTableSpec` asserting columns + indexes). Confirm the meta `migrate()` auto-repair path (the `diffMetaSchema` → `applyMetaSchemaRepairs` flow used elsewhere) actually creates this table on **existing** meta DBs at startup — same class of check I flagged on #603/#605, and it's been reliable there, but worth a one-line confirmation since the auth path hard-depends on the binding table existing.

### 🟡 Minor

- **Breaking CLI change**: the top-level `quota` command and `admin quota get` are removed; `TestDispatchQuotaVerbIsRejected` enforces it. That's an intentional break — make sure it's called out in release notes / the command's deprecation path, since existing scripts using `drive9 quota …` will now fail hard rather than warn.
- **`adminTenantQuotaRequest` and `adminTenantCreateRequest` are field-identical** — consider sharing one struct or embedding to avoid drift.
- **`adminTenantsItemHandler` path parsing** is manual (`TrimPrefix`/`Split`). Works, but a 3-segment path like `/v1/admin/tenants/<id>/quota/x` falls through to 404 correctly — just confirm `<id>` containing odd chars (encoded slashes) can't misroute; tenant IDs are server-generated so low risk.
- **`X-TiDBCloud-*` creds in headers**: fine over TLS; same model as the existing quota API. Ensure these aren't logged anywhere in the admin request path (the quota path already sanitizes — confirm admin does too).

### Tests

Good coverage: CLI table-output + header-passing tests, server admin tenant create/list/get/delete/quota tests, meta schema test for the binding table, SDK helper tests. Gaps:
- A test for the **no-binding existing-tenant → 404** path (ties to the backfill concern — make the behavior explicit).
- An **authorization-denied** test: caller's creds list clusters that do *not* include the target tenant's bound cluster → 403 (the core security property; confirm it's asserted, not just the happy path).
- List **cross-org isolation**: tenants from another org are excluded.

### Verdict

Solid feature with a correct, credential-grounded authorization model. The blocking question is the **org-binding backfill for pre-existing native tenants** — without it the admin API only manages net-new tenants. Confirm that (and the migration), add the authorization-denied + no-binding tests, and consider the list-path enumeration cost. Everything else is minor.

srstack:
Clarification on compatibility:

- The CLI incompatibility is intentional: the old top-level `drive9 quota ...` command is removed, and quota operations are under `drive9 admin ...`. Quota reads should go through `drive9 admin tenant list/get`; `drive9 admin quota get` is intentionally not exposed.
- The old HTTP `/v1/quota` API is deprecated and kept only as a compatibility endpoint for existing clients.
- The Go client now marks `GetQuota` and `SetQuota` as deprecated wrappers; new code should use `AdminGetTenant` / `AdminListTenants` for quota reads and `AdminSetTenantQuota` for quota updates.
- Admin tenant client APIs were moved out of `pkg/client/quota.go` into `pkg/client/admin_tenants.go` so quota.go only contains the deprecated compatibility surface and shared quota types/helpers.

srstack (commented):
## Re-review: PR #618 — since 15:50 (5 commits)

**The two substantive items from my last review are addressed, and the authorization model was tightened further than I'd asked. Verified the key changes. LGTM pending one confirmation (existing-tenant backfill) that I couldn't resolve from the diff alone.**

### ✅ Authorization tightened beyond my note (`f70e3dece0`)

My last review confirmed list filtered by **org** (`organizationIDsFromClusters` → `ListTenantsByTiDBCloudOrganizations`). That had a within-org over-exposure: a caller whose key manages cluster A could list *all* tenants in the org. This commit tightens it to **(org, cluster) pairs**:
- `authorizedTiDBCloudOrgClusterBindings(clusters)` + `ListTenantsByTiDBCloudOrgClusterBindings(...)` with `WHERE (organization_id, cluster_id) IN (...)`.
- So a caller now sees only tenants on the specific clusters their TiDB Cloud key actually manages, not the whole org. Strictly stronger isolation. ✅
- `authorizedAdminTenant` already matched on both ClusterID AND OrganizationID; this brings *list* up to the same granularity as get/delete/quota-set. Good consistency.

### ✅ Failed-provision cluster cleanup (`de9e5a0356`)

The create path previously just marked the tenant `Failed` on a quota error, leaving the freshly-created TiDB Cloud cluster orphaned (billable leak). Now `cleanupProvisionedClusterAfterProvisionFailure` marks failed **and** deprovisions the cluster, with a logged guard when native credentials are missing (can't deprovision without them). This is the right fix and mirrors the tenant-delete cleanup semantics. ✅ The `provision_test.go` additions (+128) cover it.

### 🟠 Still confirm: backfill for pre-existing native tenants

My main blocking question from last round stands and I can't resolve it from the diff: **native tenants created before this PR have no `tenant_tidbcloud_org_bindings` row**, so `authorizedAdminTenant` → `GetTenantTiDBCloudOrgBinding` returns 404 and they're invisible to list / unmanageable via the admin API. Confirm one of:
- a backfill populates bindings for existing native tenants, or
- the non-admin native provision path (fork/create) also writes the binding, or
- it's accepted that only net-new admin-created tenants are manageable (call it out in the PR/release notes).

The auth-tightening commit added `meta.go` binding logic but I didn't see a backfill in the diff — please confirm.

### 🟡 Index change to confirm (`f70e3dece0`)

The binding table's `UNIQUE INDEX uk_tidbcloud_org_cluster (cluster_id)` was changed to a **non-unique** composite `INDEX idx_tidbcloud_org_cluster (organization_id, cluster_id, created_at, tenant_id)`. Dropping the uniqueness on `cluster_id` means the schema no longer enforces one-tenant-per-cluster at the DB level. If that invariant still holds logically (a managed cluster backs exactly one tenant), consider keeping a unique constraint on `cluster_id` to prevent a duplicate-binding bug from silently authorizing two tenants onto one cluster. Confirm this was intentional (e.g. to support a cluster hosting multiple branch tenants).

### 🟡 Other items from last review — status
- Breaking CLI removal of top-level `quota` / `admin quota get`: still in place (intended) — ensure release notes flag it.
- Migration of the binding table on existing meta DBs via the auto-repair path: still worth a one-line confirmation (the meta-schema test asserts the spec, but confirm `migrate()` creates it on an existing DB).
- The `ed2d05f0e6` "tighten admin client and cli guards" + `93af81a263` "split admin tenant client APIs" are good structural cleanups (client API moved to `pkg/client/admin_tenants.go`, +194).

### Tests
Strong additions across the new commits: `quota_test.go` (+247), `provision_test.go` (+128 then +34), `admin_test.go` (+139), `meta_test.go` (+27). These cover the cleanup-on-failure and tightened-auth paths. Still recommend an explicit **authorization-denied** test for *list* (caller's clusters don't include a tenant's bound cluster → tenant excluded) to lock in the new (org,cluster)-pair isolation, and the **no-binding → 404** case.

### Verdict
The auth model is now tighter (cluster-granular, not org-granular) and the provision-failure cluster leak is closed — both real improvements over what I reviewed at 15:50. Remaining: confirm the existing-tenant binding backfill (the one thing that decides whether this manages the current fleet or only new tenants), and confirm the dropped `cluster_id` uniqueness was intentional. Both are confirmations, not necessarily code changes. Otherwise LGTM.

srstack:
Confirming the remaining review questions:

- Existing TiDBCloud Mode tenants are intentionally not backfilled into tenant_tidbcloud_org_bindings. This admin tenant API is scoped to newly provisioned TiDBCloud Mode tenants that have a binding. Existing no-binding tenants keep working through the deprecated compatibility /v1/quota get/set path.
- The non-admin /v1/provision path also writes the binding for new TiDBCloud Mode tenants, because both /v1/provision and /v1/admin/tenants go through provisionTenant, and provisionTenant now requires OrganizationID and calls UpsertTenantTiDBCloudOrgBinding for ProviderTiDBCloudNative.
- Dropping cluster_id uniqueness is intentional. We need to allow multiple Drive9 tenant rows to reference the same TiDB Cloud cluster, for example branch/fork-style tenants. Authorization is still cluster-granular through the stored (organization_id, cluster_id) binding plus the caller's managed-cluster list.
- On existing meta DBs, the binding table is part of the meta schema spec, and meta.Open runs migrate(), which diffs the runtime schema and applies safe repairs for missing tables. So new deployments/upgrades create tenant_tidbcloud_org_bindings automatically, but they do not synthesize historical binding rows.

Also noting that top-level CLI quota removal is intentional and should be treated as a documented breaking CLI change; /v1/quota remains deprecated compatibility for existing clients.

srstack (commented):
## Re-review #2: PR #618 — since 04:06 (3 commits: spending-limit-on-create)

**Scope: these 3 commits add create-time spending-limit (and quota) support to the provision path. Clean and additive — they don't touch my two still-open items. The new decode path is a genuine improvement. LGTM on the new code; the two prior confirmations still stand.**

### ✅ New: quota/spending-limit at provision time (`9e9c9dc1b5`, `efccb5bf5c`, `23a0486eb4`)

- `decodeCredentialProvisionRequest` now returns a `credentialProvisionRequest{Credential, Quota}` wrapper and decodes the quota fields (`max_storage_size`/`max_file_size`/`max_file_count`/`tidbcloud_spending_limit`) alongside the keys. The decode was also hardened: `DisallowUnknownFields` + trailing-data check + explicit empty/partial credential validation (`ErrCredentialsRequired`/`ErrPartialCredentials`). That's stricter than the old func-based decode — good. ✅
- **Non-native guard is correct**: the non-native provision path rejects any quota fields with `"quota settings are not supported for this provider (only tidb_cloud_native)"`, so spending-limit/quota can't be silently accepted-and-ignored on starter/zero providers. ✅
- CLI side (`db.go`) wires `--tidbcloud-spending-limit` on create and defaults it to 0; the error-hint commit improves the create-quota-error message. Reasonable.

### 🟠 Two items from my last re-review still open (unchanged by these commits)

1. **Backfill for pre-existing native tenants** — tenants created before this PR have no `tenant_tidbcloud_org_bindings` row → 404 from the admin API (list/get/delete/quota). Still need confirmation: backfill exists, the non-admin provision path writes the binding, or it's accepted as new-tenants-only. None of these 3 commits address it.
2. **Dropped `cluster_id` uniqueness** — the binding table's former `UNIQUE INDEX (cluster_id)` is now a non-unique composite. Confirm intentional (multiple branch tenants per cluster) vs. losing the one-tenant-per-cluster DB guard.

### 🟡 Minor (new)

- The quota-field set is now decoded in three places (admin create, admin quota-set, and the provision path), each with the same four `*int64` fields. They're consistent now, but that's three copies of the same shape — a shared `quotaFields` struct embedded in each request would prevent future drift (e.g. adding a 5th knob).
- `efccb5bf5c` defaults create spending limit to 0 in the CLI: confirm `0` means "no limit / leave TiDB Cloud default" vs. "hard zero" — the quota docs note TiDB Cloud may reject `0`, so a CLI default of 0 could surprise. Worth a one-line check that create-with-no-flag doesn't push an unintended spending cap.

### Tests
`db_test.go` (+110 across the commits), `provision_test.go` (+13), `admin_test.go` (+36) cover the new create-time spending-limit path and the error hint. Good. Still recommend (from prior rounds) the list authorization-denied test and the no-binding→404 test, which gate the two open items above.

### Verdict
The create-time quota/spending-limit addition is clean, well-guarded for non-native providers, and improves the decode strictness. No regressions to the auth or cleanup work from the previous rounds. LGTM on this increment — merge still gated only on the two confirmations (existing-tenant binding backfill; intentional drop of `cluster_id` uniqueness).

srstack:
Thanks, addressed the new minor in 3914e190 by extracting the shared server-side quotaFields struct and reusing it across admin create, admin quota set, /v1/
  provision decode, and the non-native provision body guard. JSON shape and behavior are unchanged; this is only to avoid drift when adding future quota knobs.

  Confirming the two unchanged gating items:

  1. Existing TiDBCloud Mode tenants without tenant_tidbcloud_org_bindings are accepted as new-tenants-only for the new admin tenant API. We are intentionally
  not backfilling old tenants in this PR. The compatibility /v1/quota path remains available for old no-binding tenants; the new /v1/admin/tenants surface is
  for tenants created after this binding support lands.

  2. Dropping cluster_id uniqueness is intentional. A TiDBCloud cluster can map to multiple drive9 tenants in shared-cluster / branch-tenant scenarios, so the
  DB should not enforce one tenant per cluster. Authorization still uses the TiDBCloud managed-cluster/org binding and the tenant binding row, not cluster_id
  uniqueness.

qiffang:
Blocking issue from my review: new native provisioning paths can leak a TiDB Cloud cluster if Drive9 fails after cluster creation but before the org binding is persisted.

The new admin tenant authorization depends on `tenant_tidbcloud_org_bindings`, so `provisionTenant` now checks `cluster.OrganizationID` and upserts the binding immediately after `ProvisionWithCredentials` returns. But both new failure paths only mark the Drive9 tenant failed and return:

- `pkg/server/server.go:4215-4220`: `cluster.OrganizationID` is empty
- `pkg/server/server.go:4222-4232`: `UpsertTenantTiDBCloudOrgBinding` fails

At this point the external TiDB Cloud cluster already exists, but the code has not yet persisted `cluster_id` onto the tenant row and does not call `cleanupProvisionedClusterAfterProvisionFailure`. The result is a failed Drive9 tenant with no binding and no recorded cluster id, while the customer’s TiDB Cloud cluster continues to exist and spend money. The admin delete path also relies on the binding/cluster authorization, so this leaked cluster is not recoverable through the new admin API.

This is different from the quota-failure path below, which correctly calls `cleanupProvisionedClusterAfterProvisionFailure(...)` before returning. The same cleanup helper should be called for these org-binding failures too, using the request credentials that just created the cluster. Please add regressions for both cases, at least the missing-organization-label case: fake native `ProvisionWithCredentials` returns `ClusterID` but empty `OrganizationID`, `provisionTenant` returns the 502, and `fakeProvisioner.deprovisionCalls == 1` with the same cluster id/credentials.

Validation I ran on head `4152f090`:

```text
GOCACHE=/tmp/drive9-go-cache go test ./pkg/server -run 'TestAdminTenant|TestQuota|TestProvisionTiDBCloudNative' -count=1  # ok
GOCACHE=/tmp/drive9-go-cache go test ./pkg/client -run 'Test.*Quota|TestAdmin' -count=1                           # ok
GOCACHE=/tmp/drive9-go-cache go test ./cmd/drive9/cli -run 'TestAdmin|TestQuota' -count=1                         # ok
GOCACHE=/tmp/drive9-go-cache go test ./cmd/drive9 -run 'TestDispatchAdminVerbReachesHandler|TestDispatchQuotaVerbIsRejected|TestDispatchLongHelpFlagShowsUsage|TestDispatchHelpCommandShowsVisualTreeHelp' -count=1  # ok
GOCACHE=/tmp/drive9-go-cache go test ./examples/go-sdk-cookbook -run 'ExampleClient_quota' -count=1                # ok
 git diff --check origin/main...HEAD                                                                                # ok
```

mornyx (commented):
## Review: Add cloud native admin tenant APIs (PR #618)

Reviewed the full diff across 28 files (+3806/-601). CI is green (code-ci ✅, local-e2e ✅).

### ✅ Strengths

1. **Clean auth model**: `authorizedAdminTenant` validates the TiDB Cloud API key against managed clusters before returning tenant data. The `cluster.OrganizationID == binding.OrganizationID` check ensures the key has org-level access to the tenant. Good separation of "can I see this tenant" (auth) from "what can I do with it" (verb routing).

2. **Provisioning reuse**: `handleAdminTenantCreate` reuses the existing `provisionTenant` + `startProvisionedTenantSchemaInit` pipeline — no parallel provisioning path. The `credentialProvisionRequest` wrapper struct cleanly separates credential fields from optional quota fields in the provision request body.

3. **Schema additions**: `tenant_tidbcloud_org_bindings` table follows the existing `tenant_external_bindings` pattern (same PK strategy, same index shape, `ON DUPLICATE KEY UPDATE` upsert). Added to `ResetMetaDB` cleanup list. ✅

4. **CLI structure**: `drive9 admin tenant {create,list,get,delete,set-quota}` is a clean verb hierarchy. Moving old top-level `drive9 quota` under `admin` with proper deprecation (reject old command, point to new) is good UX.

5. **Test coverage**: CLI tests cover help text, table/JSON output, spending-limit hints on free-quota errors, and the old-quota-namespace rejection path. Server tests cover create/get/delete/set-quota happy paths and auth failures.

### 🟡 Items to confirm

1. **`handleAdminTenantDelete` on a tenant with active upload reservations**: I see the delete handler calls `s.provisioner.DeleteTenant` but didn't see a check for active uploads or pending file GC tasks. If a tenant is deleted while uploads are in-flight, what cleans up the orphaned S3 multipart uploads? The existing tenant-delete-cleanup path (from P0-3 leader election) handles this, but confirm the admin delete path also triggers it or delegates to the same cleanup worker.

2. **`listAllManagedClusters` pagination**: The function name suggests it lists ALL clusters, but TiDB Cloud API may have pagination. If the caller's key has access to many clusters, does this loop through all pages? If it only fetches page 1, tenants on later pages would get 403. Confirm the provisioner's `ListManagedClusters` (or equivalent) paginates fully.

3. **`tenant_tidbcloud_org_bindings` cleanup on tenant delete**: When a tenant is deleted via the admin API, is the org binding row cleaned up? I didn't see a `DELETE FROM tenant_tidbcloud_org_bindings WHERE tenant_id = ?` in the delete path. If the row persists, a subsequent `GetTenantTiDBCloudOrgBinding` returns a stale binding for a deleted tenant. Not a security issue (the tenant is `TenantDeleted` and other checks block access), but a data hygiene concern.

4. **`quotaFields` embedding in both `adminTenantCreateRequest` and `adminTenantQuotaRequest`**: These two structs are identical in shape (PublicKey + PrivateKey + quotaFields). Consider unifying them or documenting why they're separate. Minor — not blocking.

### 🟢 Non-blocking notes

- The `errHelpRequested` sentinel error type is a clean pattern for the help-flag short-circuit in `parseAdminTenantIDCommand`.
- `compactTiDBCloudOrgClusterBindings` dedup logic is well-commented.
- `printAdminTenantTable` handles the `includeQuota=false` case cleanly with a simplified 3-column output.

### Verdict

**Approve with the 3 items above to confirm.** The architecture is sound, auth is properly scoped, and the provisioning/delete paths reuse existing infrastructure. The main thing I'd want confirmed before merge is the org-binding cleanup on tenant delete (item 3) and the managed-cluster pagination (item 2).

srstack:
Addressed the blocking org-binding cleanup issue in 95b35ade (fix: clean up native cluster on org binding failure).

Changes:
- The missing OrganizationID path now calls cleanupProvisionedClusterAfterProvisionFailure(...) instead of only marking the Drive9 tenant failed.
- The UpsertTenantTiDBCloudOrgBinding failure path does the same cleanup with the TiDBCloud credentials that created the cluster.
- Added TestProvisionTiDBCloudNativeCleansClusterWhenOrgBindingMissing, covering a native cluster returned with ClusterID but empty OrganizationID; it asserts 502, tenant failed, no binding row, and one deprovision call with the same cluster id and request credentials.

Validated locally:
- GOCACHE=/tmp/drive9-go-cache go test ./pkg/server -run 'TestProvisionTiDBCloudNativeCleansClusterWhenOrgBindingMissing|TestProvisionTiDBCloudNativeCleansClusterWhenCreateQuotaFails|TestProvisionTiDBCloudNativeUsesRequestCredentials|TestProvisionTiDBCloudNativeRejectsQuotaWithoutCredentials|TestProvisionTiDBCloudNativeRequiresRequestCredentials|TestDecodeCredentialProvisionRequest' -count=1
- GOCACHE=/tmp/drive9-go-cache go test ./pkg/server ./pkg/client ./cmd/drive9/cli -run 'Test.*Quota|TestAdmin|TestCreate|TestProvisionTiDBCloudNative' -count=1
- make lint
- git diff --check

srstack:
Addressed the org-binding cleanup point in 72fc6216 (fix: remove native org binding on tenant delete).

What changed:
- `enqueueTenantDeleteJob` now uses `meta.MarkTenantDeleted(...)` for the no-storage-namespace path, and that helper deletes `tenant_tidbcloud_org_bindings` in the same transaction as marking the tenant deleted.
- `FinalizeTenantDelete(...)` also deletes the org-binding row when the async delete cleanup job completes, so both delete completion paths clean up the binding.
- Added tests covering the delete path and finalize path so the binding row is gone after tenant deletion.

On the other two confirmations:
- `listAllManagedClusters` already paginates through TiDB Cloud pages until `NextPageToken` is empty.
- Admin delete is already wired through the same cleanup path and aborts active upload reservations before enqueueing delete cleanup.

Validated locally:
- `GOCACHE=/tmp/drive9-go-cache go test ./pkg/server ./pkg/meta -run "Test.*TenantDelete|TestAdminTenantDelete|Test.*TiDBCloudOrgBinding|TestQuota|TestAdminTenant|TestFinalizeTenantDeleteUpdatesJobNamespaceAndTenant" -count=1`
- `make lint`
- `git diff --check`

mornyx (commented):
## Second-round review

Thanks for addressing the first round. I've now read the full 667-line `admin_tenants.go` and the remaining server-side files. Below are the items I'd flag, grouped by severity.

---

### 🔴 Blocker — `cluster_id` uniqueness (raised by Copilot + CodeRabbit)

`tenant_tidbcloud_org_bindings` declares `cluster_id VARCHAR(255) NOT NULL` but the table has `PRIMARY KEY (tenant_id)` — **not** a unique constraint on `cluster_id`. So the Copilot comment about UNIQUE on `cluster_id` appears to be about the index, not the PK. Let me check:

```sql
CREATE TABLE IF NOT EXISTS tenant_tidbcloud_org_bindings (
    tenant_id       VARCHAR(64) PRIMARY KEY,
    organization_id VARCHAR(64) NOT NULL,
    cluster_id      VARCHAR(255) NOT NULL,
    ...
    INDEX idx_tidbcloud_org_cluster (organization_id, cluster_id, created_at, tenant_id),
    INDEX idx_tidbcloud_org_created (organization_id, created_at, tenant_id)
)
```

The `idx_tidbcloud_org_cluster` is a non-unique index. Multiple tenants can share the same `cluster_id` (which makes sense — fork tenants reuse the parent cluster). So CodeRabbit's "handle cluster_id collisions as conflicts" concern is **not applicable** — collisions are expected and handled correctly by the PK on `tenant_id`. ✅ No action needed.

However, the **UpsertTenantTiDBCloudOrgBinding** `ON DUPLICATE KEY UPDATE` operates on the `tenant_id` PK, which is correct — upserting by tenant, not by cluster. ✅

---

### 🟡 Items to confirm (carried from first round + new)

**1. Org-binding cleanup on tenant delete (carried)**

I traced the `handleAdminTenantDelete` path:
- Marks tenant `TenantDeleting` → drains backend (`InvalidateAndWait`) → deprovisions cluster → aborts uploads → enqueues tenant-delete-job → revokes API keys.

The `tenant_tidbcloud_org_bindings` row is **not** cleaned up in this path. The existing `processTenantDeleteJobs` (leader-gated tenant-delete-cleanup worker) calls `FinalizeTenantDelete` — does that clean the binding? Let me check:

The binding row persists after tenant deletion. This is a **data hygiene** issue, not a security issue:
- `authorizedAdminTenant` rejects `TenantDeleted` tenants with 404 before reaching the binding lookup, so stale bindings are invisible.
- But `ListTenantsByTiDBCloudOrgClusterBindings` queries `INNER JOIN tenants ... WHERE tenant.status != deleted`, so deleted tenants are filtered out.

**Verdict**: Acceptable for now, but should be cleaned up in `FinalizeTenantDelete` as a follow-up. Not blocking.

**2. `listAllManagedClusters` pagination (carried)**

I now see the full implementation — it **does paginate correctly**:
```go
for {
    page, err := lister.ListManagedClusters(ctx, cred, opts)
    pageToken = page.NextPageToken
    if pageToken == "" || clusterID != "" {
        return out, nil
    }
}
```

When `clusterID` is set (single-cluster lookup), it returns after the first page. When `clusterID` is empty (list-all), it loops until `NextPageToken` is empty. ✅ Correct.

**However**: a large org with hundreds of clusters will make this loop many times, each hitting the TiDB Cloud API. This is called on every `handleAdminTenantList` request. Consider caching the cluster list with a short TTL (30-60s) to avoid hammering the TiDB Cloud API on every list request. Non-blocking — operational concern.

**3. Pagination overflow (CodeRabbit)**

```go
maxOffsetPage := 1 + int(^uint(0)>>1)/pageSize
if page > maxOffsetPage {
    return 0, 0, 0, fmt.Errorf("page is too large")
}
offset = (page - 1) * pageSize
```

This guards against `int` overflow. On 64-bit platforms `^uint(0)>>1` = `maxInt`, so `maxOffsetPage` is huge. The guard works but the threshold is effectively unreachable. More importantly, `pageSize` is capped at 100, and `(page - 1) * pageSize` won't overflow on any realistic page value. ✅ Safe, though the guard is overly conservative.

**4. `decodeJSONBody` rejects unknown fields + trailing data (good)**

`DisallowUnknownFields()` + trailing-data check is a solid defense against malformed inputs. ✅

**5. Delete path: `MarkQuotaUpdateStarted` before `UpdateTenantStatusIf`**

The delete handler calls `MarkQuotaUpdateStarted` (TiDB Cloud API: set spending limit to 0) **before** marking the tenant as `TenantDeleting`. If `MarkQuotaUpdateStarted` succeeds but `UpdateTenantStatusIf` fails (e.g., tenant already being deleted by another request), the cluster's spending limit is set to 0 but the tenant isn't marked deleting. The caller gets a 202 (already deleting) and the next delete attempt skips `MarkQuotaUpdateStarted`.

This is a **minor inconsistency**: the first caller sets spending limit to 0, the second caller's `UpdateTenantStatusIf` returns `updated=false`. Not a correctness issue (both converge to deleting), but the spending limit is set twice (redundant TiDB Cloud API call). Acceptable.

**6. `adminTenantDeleteResponse` for already-deleting tenant with no job**

When `t.Status == TenantDeleting` and no delete job exists, the handler falls through to `MarkQuotaUpdateStarted` → `UpdateTenantStatusIf(TenantDeleting → TenantDeleting)` → `updated=false` → returns 202. This is correct idempotent behavior — the caller sees "already deleting" and the cleanup worker will eventually finish. ✅

---

### 🟢 Non-blocking observations

1. **`adminTenantResponse` ignores the binding parameter**: `func (s *Server) adminTenantResponse(t *meta.Tenant, _ *meta.TenantTiDBCloudOrgBinding, quota *adminTenantQuotaSummary)` — the binding is discarded. If future fields need org/cluster info in the response, this underscore-bind will need updating. Consider adding a comment explaining why the binding is passed but not used (callers always have it available from `authorizedAdminTenant`).

2. **`writeJSON` duplicates `errJSON` pattern**: `writeJSON` sets `Content-Type` + status + encodes. `errJSON` does the same with `errorResponse`. Consider unifying, but non-blocking.

3. **`adminCredentialsFromHeaders` vs `adminCredentials`**: `adminCredentialsFromHeaders` is a thin wrapper around `adminCredentials("", "", r)`. Consider just calling `adminCredentials("", "", r)` directly at call sites to reduce indirection. Minor.

---

### Verdict

**The PR is solid.** The auth model is sound, pagination works correctly, the delete path handles idempotency well, and the `decodeJSONBody` is properly defensive. The Copilot/CodeRabbit concerns about `cluster_id` uniqueness are not applicable (PK is on `tenant_id`, not `cluster_id`).

My remaining items are all non-blocking:
- Org-binding cleanup in `FinalizeTenantDelete` (follow-up)
- Cluster-list caching for large orgs (operational optimization)
- Minor code clarity items

**LGTM for merge** — the carried-forward items from round 1 (active upload cleanup on delete) are handled by `AbortActiveUploadReservations` + the leader-gated tenant-delete-cleanup worker.

srstack:
Thanks for the second-round review. Confirming the remaining points against current head:

- `cluster_id` uniqueness: agreed, no action needed. The binding table is keyed by `tenant_id`; `(organization_id, cluster_id, created_at, tenant_id)` is intentionally a non-unique lookup index so shared-cluster / branch-style tenants can coexist.
- Org-binding cleanup: this is now handled in `72fc6216`. `MarkTenantDeleted(...)` deletes `tenant_tidbcloud_org_bindings` in the same transaction for the no-namespace delete path, and `FinalizeTenantDelete(...)` deletes it when the async tenant-delete cleanup job completes. Tests cover both paths.
- `listAllManagedClusters` pagination: confirmed, it drains TiDB Cloud pages until `NextPageToken` is empty for list, and returns after the targeted lookup when `clusterID` is provided.
- Pagination overflow / strict JSON decode: current behavior looks correct.
- Delete idempotency / quota mark ordering: accepted as non-blocking; it converges to deleting and the duplicate quota mark is harmless.
- Cluster-list caching and the minor clarity cleanups are reasonable follow-ups, but I would keep them out of this PR to avoid expanding the behavioral surface.

Also noting explicitly: `/v1/provision` response shape is intentionally kept unchanged apart from the TiDBCloud Mode cloud provider / region fields already discussed; the tenant-org binding remains an internal admin-auth primitive, not a provision response field.

mornyx (approved):
LGTM. All items from my two rounds of review have been addressed by the author:

- ✅ Org-binding cleanup on tenant delete (commit 72fc6216: "fix: remove native org binding on tenant delete")
- ✅ Provision failure cluster cleanup (commit 95b35ade: "fix: clean up native cluster on org binding failure")
- ✅ Spending limit defaults and create-time quota errors (commits efccb5bf, 23a0486e)
- ✅ listAllManagedClusters pagination is complete (verified in round 2)
- ✅ Delete path idempotency and AbortActiveUploadReservations coverage
- ✅ cluster_id uniqueness concern is N/A (PK is on tenant_id, not cluster_id)

Solid auth model, clean provisioning reuse, defensive JSON parsing, proper idempotent delete.

PR #619, opened 2026-06-24
Fix fs_events path schema auto-repair

srstack:
## Summary
- allow TiDB schema ensure to auto-repair fs_events.path from VARCHAR(512) to TEXT
- cover the repair planner path with a regression test
- align event bus and SSE test fixtures with the current fs_events schema

## Tests
- make test TEST_PKGS='./pkg/tenant/schema' TEST_RUN='TestPlannedTiDBSchemaRepairsAllowsPathColumnWidening'
- make test TEST_PKGS='./pkg/server' TEST_RUN='TestEventBus|TestSSE'


mornyx (approved):
LGTM. The `fs_events.path` repair registration correctly mirrors the existing `file_nodes.path` and `uploads.target_path` widening patterns. Test fixture alignment with `TEXT` fixes the gap from PR #602 where I changed the production schema but missed the test `CREATE TABLE` statements. Regression test coverage is solid.

PR #620, opened 2026-06-25
Fix TiDBCloud Mode tenant create readiness

srstack:
## Summary

This PR fixes TiDBCloud Mode tenant creation latency and readiness handling.

- Return from TiDBCloud Mode create/provision once TiDB Cloud has returned the tenant connection metadata and `tidb.cloud/organization` label; do not wait for the cluster to become `ACTIVE`.
- Stop doing synchronous DB creation in the create/provision request path.
- Move native tenant database creation into the async schema-init flow, using the persisted DB host/user/password instead of any user TiDBCloud API key.
- Ensure the native tenant database exists before auto-embedding provider pre-configuration opens the tenant DSN.
- Parallelize tenant bootstrap DDL by table so independent tables can initialize concurrently while preserving statement order within each table.
- Keep TiDB Cloud Starter schema repair/validation in place for legacy launch-schema drift.

## Notes

- The request path still uses the submitted TiDBCloud key only to create the cluster and fetch the metadata needed to persist the tenant-org binding.
- Background workers do not store or use user TiDBCloud API keys.
- Schema repair, validation, and optional TiDB FTS/vector index handling remain serial.

## Validation

- `GOCACHE=/tmp/drive9-go-cache go test ./pkg/tenant/starter -count=1`
- `GOCACHE=/tmp/drive9-go-cache go test ./pkg/tenant/schema -count=1`
- `GOCACHE=/tmp/drive9-go-cache go test ./pkg/server -run 'TestSchemaInitForTenant|TestProvisionTiDBCloudNative|TestAdminTenantCreate|TestAdminTenantSetQuota|TestDecodeCredentialProvisionRequest' -count=1`
- `GOCACHE=/tmp/drive9-go-cache go test ./pkg/tenant/tidbcloudnative -count=1`
- `GOCACHE=/tmp/drive9-go-cache go test ./pkg/server ./pkg/tenant/schema ./pkg/tenant/tidbcloudnative ./pkg/tenant/starter -run 'TestProvision|TestAdminTenant|TestQuota|TestEnsureDatabaseFromDSN|TestParallelSchemaStatementGroups|TestExecSchemaStatementsParallel|TestProvisionerInitSchema|TestSchemaInitForTenant' -count=1`
- `make lint`
- `git diff --check`

## Summary by CodeRabbit

* **New Features**
  * Tenant schema initialization now runs eligible database statements in parallel by table, which can speed up provisioning.
  * Provisioning now ensures the tenant database exists before schema setup when needed.

* **Bug Fixes**
  * Improved provisioning readiness checks so clusters are only considered ready after required metadata is available.
  * Schema initialization now validates database availability earlier, reducing failures during setup.

### DIFF
--- a/pkg/server/provision_test.go
+++ b/pkg/server/provision_test.go
@@ -10,6 +10,7 @@ import (
 	"net/http"
 	"net/http/httptest"
 	"strings"
+	"sync"
 	"sync/atomic"
 	"testing"
 	"time"
@@ -179,6 +180,7 @@ func (f *fakeProvisioner) ProvisionCallCount() int {
 
 type profileAwareFakeProvisioner struct {
 	fakeProvisioner
+	mu               sync.Mutex
 	profileInitCalls atomic.Int32
 	ensureDBCalls    atomic.Int32
 	lastProfile      tenantschema.TiDBAutoEmbeddingProfile
@@ -189,6 +191,8 @@ type profileAwareFakeProvisioner struct {
 
 func (f *profileAwareFakeProvisioner) EnsureDatabase(_ context.Context, dsn string) error {
 	f.ensureDBCalls.Add(1)
+	f.mu.Lock()
+	defer f.mu.Unlock()
 	f.lastEnsureDSN = dsn
 	f.callOrder = append(f.callOrder, "ensure")
 	return f.ensureDBErr
@@ -196,9 +200,17 @@ func (f *profileAwareFakeProvisioner) EnsureDatabase(_ context.Context, dsn stri
 
 func (f *profileAwareFakeProvisioner) InitSchemaForAutoEmbeddingProfile(_ context.Context, _ string, profile tenantschema.TiDBAutoEmbeddingProfile) error {
 	f.profileInitCalls.Add(1)
+	f.mu.Lock()
+	defer f.mu.Unlock()
 	f.lastProfile = profile
 	f.callOrder = append(f.callOrder, "profile-init")
 	return nil
+}
+
+func (f *profileAwareFakeProvisioner) callOrderString() string {
+	f.mu.Lock()
+	defer f.mu.Unlock()
+	return strings.Join(f.callOrder, ",")
 }
 
 func TestSchemaInitForTenantEnsuresDatabaseBeforeAutoEmbeddingConfig(t *testing.T) {
@@ -262,7 +274,7 @@ func TestSchemaInitForTenantEnsuresDatabaseBeforeProfileInit(t *testing.T) {
 	if prov.profileInitCalls.Load() != 1 {
 		t.Fatalf("profile init calls = %d, want 1", prov.profileInitCalls.Load())
 	}
-	if got, want := strings.Join(prov.callOrder, ","), "ensure,profile-init"; got != want {
+	if got, want := prov.callOrderString(), "ensure,profile-init"; got != want {
 		t.Fatalf("call order = %s, want %s", got, want)
 	}
 }
--- a/pkg/server/provision_test.go
+++ b/pkg/server/provision_test.go
@@ -180,13 +180,91 @@ func (f *fakeProvisioner) ProvisionCallCount() int {
 type profileAwareFakeProvisioner struct {
 	fakeProvisioner
 	profileInitCalls atomic.Int32
+	ensureDBCalls    atomic.Int32
 	lastProfile      tenantschema.TiDBAutoEmbeddingProfile
+	ensureDBErr      error
+	lastEnsureDSN    string
+	callOrder        []string
+}
+
+func (f *profileAwareFakeProvisioner) EnsureDatabase(_ context.Context, dsn string) error {
+	f.ensureDBCalls.Add(1)
+	f.lastEnsureDSN = dsn
+	f.callOrder = append(f.callOrder, "ensure")
+	return f.ensureDBErr
 }
 
 func (f *profileAwareFakeProvisioner) InitSchemaForAutoEmbeddingProfile(_ context.Context, _ string, profile tenantschema.TiDBAutoEmbeddingProfile) error {
 	f.profileInitCalls.Add(1)
 	f.lastProfile = profile
+	f.callOrder = append(f.callOrder, "profile-init")
 	return nil
+}
+
+func TestSchemaInitForTenantEnsuresDatabaseBeforeAutoEmbeddingConfig(t *testing.T) {
+	ensureErr := errors.New("database is not ready")
+	prov := &profileAwareFakeProvisioner{
+		fakeProvisioner: fakeProvisioner{provider: tenant.ProviderTiDBCloudNative},
+		ensureDBErr:     ensureErr,
+	}
+	srv := NewWithConfig(Config{
+		Provisioner: prov,
+		TiDBAutoEmbeddingConfig: tenantschema.TiDBAutoEmbeddingConfig{
+			Model:      "openai/text-embedding-3-small",
+			Dimensions: 1536,
+		},
+	})
+	defer srv.Close()
+
+	init := srv.schemaInitForTenant("tenant-native", tenant.ProviderTiDBCloudNative, func(context.Context, string) error {
+		t.Fatal("fallback InitSchema was called")
+		return nil
+	})
+	err := init(context.Background(), "u1.root:db-pass@tcp(db.example:4000)/tidbcloud_fs?parseTime=true&tls=true")
+	if !errors.Is(err, ensureErr) {
+		t.Fatalf("schema init error = %v, want ensure error", err)
+	}
+	if prov.ensureDBCalls.Load() != 1 {
+		t.Fatalf("ensure DB calls = %d, want 1", prov.ensureDBCalls.Load())
+	}
+	if prov.profileInitCalls.Load() != 0 {
+		t.Fatalf("profile init calls = %d, want 0", prov.profileInitCalls.Load())
+	}
+	if prov.lastEnsureDSN == "" {
+		t.Fatal("ensure DB DSN was empty")
+	}
+}
+
+func TestSchemaInitForTenantEnsuresDatabaseBeforeProfileInit(t *testing.T) {
+	prov := &profileAwareFakeProvisioner{
+		fakeProvisioner: fakeProvisioner{provider: tenant.ProviderTiDBCloudNative},
+	}
+	srv := NewWithConfig(Config{
+		Provisioner: prov,
+		TiDBAutoEmbeddingConfig: tenantschema.TiDBAutoEmbeddingConfig{
+			Model:      "openai/text-embedding-3-small",
+			Dimensions: 1536,
+		},
+	})
+	defer srv.Close()
+
+	init := srv.schemaInitForTenant("tenant-native", tenant.ProviderTiDBCloudNative, func(context.Context, string) error {
+		t.Fatal("fallback InitSchema was called")
+		return nil
+	})
+	err := init(context.Background(), "u1.root:db-pass@tcp(db.example:4000)/tidbcloud_fs?parseTime=true&tls=true")
+	if err != nil {
+		t.Fatalf("schema init: %v", err)
+	}
+	if prov.ensureDBCalls.Load() != 1 {
+		t.Fatalf("ensure DB calls = %d, want 1", prov.ensureDBCalls.Load())
+	}
+	if prov.profileInitCalls.Load() != 1 {
+		t.Fatalf("profile init calls = %d, want 1", prov.profileInitCalls.Load())
+	}
+	if got, want := strings.Join(prov.callOrder, ","), "ensure,profile-init"; got != want {
+		t.Fatalf("call order = %s, want %s", got, want)
+	}
 }
 
 func TestProvisionMarksTenantFailedWhenInitKeepsFailing(t *testing.T) {
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -92,6 +92,10 @@ type autoEmbeddingSchemaProvisioner interface {
 	InitSchemaForAutoEmbeddingProfile(context.Context, string, tenantschema.TiDBAutoEmbeddingProfile) error
 }
 
+type tenantDatabaseEnsurer interface {
+	EnsureDatabase(context.Context, string) error
+}
+
 type credentialProvisionRequestValidator interface {
 	ValidateCredentialProvisionRequest(tenant.CredentialProvisionRequest) error
 }
@@ -4104,6 +4108,11 @@ func (s *Server) schemaInitForTenant(tenantID, provider string, fallback func(co
 		return fallback
 	}
 	return func(ctx context.Context, dsn string) error {
+		if ensurer, ok := s.provisioner.(tenantDatabaseEnsurer); ok {
+			if err := ensurer.EnsureDatabase(ctx, dsn); err != nil {
+				return fmt.Errorf("ensure tenant database: %w", err)
+			}
+		}
 		profile, err := s.autoEmbeddingProfileForTenant(ctx, tenantID)
 		if err != nil {
 			return fmt.Errorf("resolve tenant auto-embedding profile: %w", err)
--- a/pkg/tenant/schema/common.go
+++ b/pkg/tenant/schema/common.go
@@ -92,6 +92,8 @@ func ExecSchemaStatementsParallelByTableContext(ctx context.Context, db *sql.DB,
 	wg.Wait()
 	close(errCh)
 	for err := range errCh {
+		// Return the first completed group failure; all groups see the cancelled
+		// context before starting the next statement.
 		return err
 	}
 	return nil
--- a/pkg/tenant/schema/common.go
+++ b/pkg/tenant/schema/common.go
@@ -6,6 +6,7 @@ import (
 	"errors"
 	"fmt"
 	"strings"
+	"sync"
 	"time"
 
 	mysql "github.com/go-sql-driver/mysql"
@@ -13,6 +14,8 @@ import (
 	"github.com/mem9-ai/drive9/pkg/logger"
 	"go.uber.org/zap"
 )
+
+const maxParallelSchemaStatementGroups = 8
 
 // CloneStatements returns a copy of the given schema statements so callers can
 // format or extend them without mutating the init-schema source of truth.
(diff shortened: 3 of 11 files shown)


srstack:
Added schema-init parallelism in `aa37ab9e`.

What changed:
- Tenant bootstrap DDL now runs with table-level parallelism.
- Statements for the same table still keep source order, so `CREATE TABLE` stays before that table's `CREATE INDEX` / `ALTER TABLE`.
- Unclassified statements fall back to the old serial executor.
- Repair / validation / optional TiDB FTS/vector index handling remains unchanged and serial.

Validation:
- `GOCACHE=/tmp/drive9-go-cache go test ./pkg/tenant/schema -count=1`
- `GOCACHE=/tmp/drive9-go-cache go test ./pkg/tenant/tidbcloudnative -count=1`
- `GOCACHE=/tmp/drive9-go-cache go test ./pkg/server ./pkg/tenant/schema ./pkg/tenant/tidbcloudnative -run 'TestProvision|TestAdminTenant|TestQuota|TestEnsureDatabaseFromDSN|TestParallelSchemaStatementGroups|TestExecSchemaStatementsParallel' -count=1`
- `make lint`
- `git diff --check`

srstack (commented):
## Code Review

### Overview
This PR removes synchronous DB creation and the "wait for `ACTIVE`" step from the TiDBCloud Native **provision request path**, deferring DB creation to the **async schema-init flow**, and adds **table-level parallelism** to tenant bootstrap DDL. It also turns the Starter provisioner's schema init into a no-op. Net effect: faster, leaner create requests, with readiness handled by the existing async retry loop. Well-scoped and well-tested.

### Correctness — verified
- **Readiness now relies on the async retry loop.** Provision returns once connection metadata + the `tidb.cloud/organization` label are present (`clusterProvisionMetadataIncomplete`), not on `ACTIVE`. If the cluster isn't connectable yet, `EnsureDatabase` fails and `initTenantSchemaAsync` retries with backoff until `schemaInitRetryWindow = 10m` — the same budget the old `waitForClusterActive` deadline used. The time budget simply moved from the request path to the worker. ✔️
- **Ordering of the DB-readiness fix is correct.** `schemaInitForTenant` runs `EnsureDatabase` → `applyAutoEmbeddingProviderConfig` → `InitSchemaForAutoEmbeddingProfile`. Since the config step issues `SET @@GLOBAL` against the tenant DB, ensuring the database first is necessary and now guaranteed. ✔️
- **Parallel DDL is safe for these schemas.** `parallelSchemaStatementGroups` returns `ok=false` on *any* unclassifiable statement and falls back to full serial execution — a good safety net. The tenant schemas contain **no `FOREIGN KEY`/`REFERENCES`**, so cross-table dependencies that would break under concurrent table creation don't exist here. Within-table source order is preserved. ✔️
- **Goroutine/error handling is clean.** `errCh` is buffered to `len(groups)` (no send-block), `cancel()` propagates to in-flight `ExecContext` and to not-yet-started goroutines via `ctx.Done()`, and `wg.Wait()` precludes leaks. ✔️
- **`splitTCPAddr` IPv6 handling** correctly re-wraps `::1` → `[::1]`; the non-positive-port test even guards against a `%!w(nil)` wrapping bug. ✔️

### Issues / Risks
**1. Starter `InitSchema`/`InitSchemaForAutoEmbeddingProfile` → no-op is the biggest behavioral risk.** The removed comment stated this path "repairs known launch-schema drift (for example, missing uploads columns from legacy tenants)." With it gone, any **pre-existing/legacy Starter tenant with schema drift will no longer be repaired** on init. The justification ("starter schemas are externally pre-initialized") holds for *newly* provisioned clusters, but the dropped repair was specifically about *legacy* tenants. Please confirm legacy Starter tenants either (a) don't exist in any environment, or (b) are repaired by another path. If unsure, I'd hold this part.

**2. Limited real parallelism on TiDB.** TiDB serializes DDL through a single DDL owner, so concurrent `CREATE TABLE`/`CREATE INDEX` are largely queued server-side; the win is mostly pipelined round-trips. The MySQL no-embedding path benefits more. Not a bug — just calibrate the expected gain on the TiDB paths.

**3. Minor: redundant `EnsureDatabase`.** Native `InitSchema` now calls `EnsureDatabase`, but in the server flow the wrapped `schemaInitForTenant` already calls it and uses `InitSchemaForAutoEmbeddingProfile` instead — so `InitSchema`'s ensure only fires for direct/fallback callers (e.g. `fork.go`). It's idempotent (`CREATE DATABASE IF NOT EXISTS`), so harmless, but a one-line comment noting why it lives in both places would help.

### Nits
- `ExecSchemaStatementsParallelByTableContext`: `for err := range errCh { return err }` returns the first received error nondeterministically across groups. Fine in practice; a brief comment ("returns first failure; remaining cancelled") would help future readers.
- `callOrder` in `profileAwareFakeProvisioner` is a plain slice while sibling counters are `atomic.Int32`. Safe here (same goroutine), but the inconsistency could mislead someone who later adds a concurrent test.

### Test Coverage
Strong — ordering preservation, unclassified-statement fallback, real concurrency of independent tables, DSN parsing (IPv6, non-TCP, bad port), ensure-before-config ordering, and the provision path no longer calling `ensureDatabase`. The metadata-readiness change is exercised with a `CREATING` state + org label.

### Verdict
**Approve, pending confirmation on item #1.** Code quality, conventions, and tests are solid; the parallelism and readiness refactors are correct for the current schemas. The one thing worth confirming before merge is whether dropping Starter schema-drift repair is safe for any legacy tenants.

srstack:
Addressed the latest review in 55d33571. Starter schema init is no longer a no-op: InitSchema and InitSchemaForAutoEmbeddingProfile now keep the existing repair/validation path for legacy launch-schema drift. Also added the minor clarity items from the review: documented first-error behavior in parallel DDL, documented why native InitSchema still ensures the DB for direct/fallback callers, and guarded the test call-order slice with a mutex.

srstack (commented):
## Re-review (commit `55d33571`)

The latest commit addresses every point from the previous review:

| Prior point | Status |
|---|---|
| **#1 — Starter no-op drops legacy schema-drift repair** (the blocker) | ✅ Reverted — `InitSchema` → `EnsureTiDBSchemaForModeDSN`, `InitSchemaForAutoEmbeddingProfile` → `EnsureTiDBSchemaForAutoEmbeddingProfileDSN`; comment + validation test restored. |
| **#3 — redundant `EnsureDatabase` in native `InitSchema`** | ✅ Comment added explaining direct callers need it while the server path hoists the same ensure. |
| Nit — `errCh` first-error semantics | ✅ Comment added. |
| Nit — `callOrder` data race | ✅ Guarded with a `sync.Mutex` + `callOrderString()` accessor. |

**Verification (local):**
- `go build ./pkg/tenant/... ./pkg/server/...` — clean
- `starter` + `schema` parallel/init tests — pass
- `TestSchemaInitForTenant*` under `-race` — pass (mutex fix confirmed)

**Non-blocking note:** the PR description is now stale — it still says *"Keep TiDB Cloud Starter schema init as a no-op,"* which this commit reverses. Worth updating the body.

The earlier blocker is resolved. The readiness refactor and table-parallel DDL are correct and the concurrency concerns are closed. LGTM once the description is touched up.

mornyx (approved):

mornyx (approved):
## Review: Fix TiDBCloud Mode tenant create readiness

Reviewed the full diff (+504/-52 across 9 files). CI is green.

### ✅ Strengths

1. **Provisioning readiness fix is correct**: The old code waited for `State == "ACTIVE"` before returning from provision — this blocked the entire create request for minutes while TiDB Cloud started the cluster. The new `clusterProvisionMetadataIncomplete` check returns as soon as connection metadata + org label are available, which is the minimum needed to persist the tenant binding. The cluster continues bootstrapping in the background. This is a significant latency improvement for tenant creation.

2. **Database creation moved out of request path**: `ensureDatabase` is no longer called synchronously in `ProvisionWithCredentials`. Instead it's deferred to the async schema-init flow via `EnsureDatabase` → `ensureDatabaseFromDSN`. This correctly uses the **persisted tenant credentials** (from the tenant record) rather than the user's TiDB Cloud API key — important because background workers should not store or use user API keys (stated in the PR description and verified in the code).

3. **`EnsureDatabase` interface pattern**: The `tenantDatabaseEnsurer` interface is a clean type assertion — `schemaInitForTenant` calls it before auto-embedding profile init. The `TestSchemaInitForTenantEnsuresDatabaseBeforeAutoEmbeddingConfig` test verifies the ordering: ensure → profile-init, and that profile-init is skipped if ensure fails. Good.

4. **Parallel DDL execution**: `ExecSchemaStatementsParallelByTableContext` groups statements by table name (preserving source order within each table) and runs independent tables concurrently with a semaphore (max 8). The `schemaStatementTableName` parser handles CREATE TABLE, CREATE INDEX (via `ON` clause), and ALTER TABLE. The fallback to serial execution when classification fails (e.g., `SET @@GLOBAL` statements) is safe and correct.

5. **IPv6 handling in `splitTCPAddr`**: The `if strings.Contains(host, ":")` → bracket wrapping is a good defensive measure for IPv6 addresses in DSNs.

6. **Test coverage is thorough**:
   - `TestSchemaInitForTenantEnsuresDatabaseBeforeAutoEmbeddingConfig` — ensures DB before profile init, fails fast on ensure error
   - `TestSchemaInitForTenantEnsuresDatabaseBeforeProfileInit` — verifies call order "ensure,profile-init"
   - `TestParallelSchemaStatementGroupsKeepsTableOrder` — verifies table grouping preserves source order
   - `TestParallelSchemaStatementGroupsRejectsUnclassifiedStatements` — verifies fallback when unclassifiable
   - `TestExecSchemaStatementsParallelByTableContextRunsIndependentTablesConcurrently` — verifies actual concurrency via channel coordination
   - `TestEnsureDatabaseFromDSN*` — covers TCP, non-TCP rejection, IPv6, non-positive port, empty host
   - Provision tests updated to assert `ensureDatabase` is NOT called during provision (correctly moved to async)

### 🟡 Items to confirm (non-blocking)

1. **`waitForClusterProvisionMetadata` timeout**: The 10-minute deadline is unchanged from the old `waitForClusterActive`. Since we now only wait for metadata (not `ACTIVE` state), is 10 minutes still the right timeout? TiDB Cloud typically returns connection metadata within 1-2 minutes. If the goal is to fail fast on provision issues, a shorter timeout (e.g., 5 minutes) might be more appropriate. Non-blocking — the 10-minute ceiling is safe.

2. **Parallel DDL + `isIgnorableSchemaError`**: When two tables' CREATE INDEX statements run concurrently and one fails with a non-ignorable error, `cancel()` is called. The other goroutine checks `ctx.Done()` before starting its next statement. However, a statement already in-flight (mid-execution) will complete and its error (if any) is also sent to `errCh`. The `for err := range errCh` loop returns the first error, which may not be the "root cause" error. This is acceptable — the error message includes the statement snippet, and any concurrent failure is actionable. Just noting that error ordering is non-deterministic.

3. **`schemaStatementTableName` for `CREATE UNIQUE INDEX`**: The parser checks `fields[0] != "create"` and then looks for `ON`. `CREATE UNIQUE INDEX idx ON t1(col)` has `fields[0] == "create"`, `fields[1] == "unique"`, so the `ON` search still works. ✅ Verified.

4. **`InitSchema` now calls `EnsureDatabase`**: The direct `InitSchema` entry point (used by non-server callers?) now also ensures the database. This is correct for consistency, but if any caller already ensures the database before calling `InitSchema`, this results in a redundant `CREATE DATABASE IF NOT EXISTS`. Harmless but worth noting.

### Verdict

**LGTM.** The provisioning latency fix is the right approach — returning early with metadata instead of waiting for ACTIVE state. The async database creation with persisted credentials is architecturally correct. Parallel DDL is well-implemented with proper fallback and ordering guarantees. Test coverage is excellent.

srstack:
Confirmed the non-blocking items:

- `waitForClusterProvisionMetadata` keeps the old 10-minute ceiling intentionally. The request path no longer waits for `ACTIVE`, but we still want the same upper-bound budget for TiDB Cloud metadata propagation and transient API delays. The async DB readiness loop handles the later connectability wait.
- Parallel DDL error ordering is accepted as non-deterministic. Each returned error includes the statement snippet, and after the first non-ignorable failure we cancel remaining work; any already in-flight statement may still finish and report first.
- `CREATE UNIQUE INDEX ... ON ...` is covered by the generic `ON` scan in `schemaStatementTableName`; no code change needed.
- Native `InitSchema` calling `EnsureDatabase` is intentional for direct/fallback callers outside the server auto-embedding wrapper. It can be redundant if a caller already ensured the DB, but `CREATE DATABASE IF NOT EXISTS` makes that harmless.

qiffang:
adversary-2 review on PR #620 head `55d3357`.

**Verdict: APPROVE** with 3 non-blocking observations.

## 独立本机验证

cloned + checked out PR + 跑了 PR body 列出的所有 test command + 额外加 `-race -count=5` 验证 parallel DDL goroutine 行为：

| Test | Result |
|---|---|
| `go test ./pkg/tenant/tidbcloudnative -count=1` | ✅ ok 11.2s |
| `go test ./pkg/tenant/schema -count=1` | ✅ ok 21.3s |
| `go test ./pkg/server -run 'TestSchemaInit\|TestProvisionTiDBCloud\|TestAdminTenant\|TestDecodeCredentialProvision' -count=1` | ✅ ok 14.6s |
| `go test ./pkg/tenant/schema -run 'TestParallelSchema' -race -count=5` | ✅ ok 25.0s (no race) |

## srstack 已 catch + 我独立验证 + 同意的：

1. **Starter no-op blocker 修复** (commit `55d33571`): `InitSchema` 现在调用 `EnsureTiDBSchemaForModeDSN`，legacy schema drift 重新被 repair。我 grep 确认。
2. **Readiness time budget 守恒**: `schemaInitRetryWindow = 10m` 跟旧 `waitForClusterActive` deadline 一样，**预算从 request path 移到 worker**，没有引入额外延迟风险。

## 我独立 trace 的额外 lifecycle interleavings（昨天 PR #617 banked rule 应用）

### LI1: Request 返回后，client 立刻打 tenant 请求会怎样？

Trace:
- Tenant 在 meta 中 status=`Provisioning`
- Client 请求路由到这个 tenant
- `serverProvisioningMessage` (`server.go:1317-1325`) 返回 "schema init error" + 具体错误信息（从 `schemaInitErrors` map 读）
- Client 看到清晰的"还在建库"状态，**不是 500 panic 也不是无响应**

✅ Lifecycle 边界处理得对。

### LI2: 重试时 schema 已部分建成（前次 partial failure）

Trace:
- `parallelSchemaStatementGroups` 对**任何**不可分类语句 → 回退 serial。安全网。
- Classifiable 语句调 `execSchemaStatement` → `isIgnorableSchemaError` 处理"already exists"。**Retry idempotent**。
- Parallel 路径下，每个 group 内 statement 保留 source order；group 间无 FK 依赖（PR body claim + grep `FOREIGN KEY` = 0 验证）

✅ 重试 idempotent，无数据损坏风险。

### LI3: 服务器重启 in mid-provisioning

Trace (`server.go:651, 728-749`):
- 重启后 `resumeProvisioningTenantsWithCtx` 扫所有 `Provisioning` 状态 tenant
- 用 **persisted credentials** (`t.DBUser`, decrypted `t.DBPasswordCipher`) 重建 DSN
- 调 `initTenantSchemaAsync` 继续重试
- **完全不需要 user TiDBCloud API key** — 跟 PR body claim 一致

✅ Restart-safe。**这是 PR 的真正安全改进**：API key 不再驻留在 server 处理路径中。

## 3 个 non-blocking 观察

### N1 (observability): DSN 密码 cleartext 驻留在 goroutine heap 高达 10 分钟

`initTenantSchemaAsync` 的 `tenantDSN` 字符串包含 cleartext password，在 retry window (10m) 内整个驻留在 worker goroutine 的 stack/heap。如果出现 crash dump、heap profile 或 OOM dump，密码可能泄露。

**不是 regression** (旧 sync 路径也是 cleartext 给 sql.Open)，但 window 从"短期 sync 调用"扩展到"最多 10m async retry"。

**Suggestion**: 不阻 PR。后续 PR 可考虑：
- Schema init 函数签名改为接受 lazy 解密 callback：`func() (string, error)` 而非裸 DSN string
- 或者明确 sentinel pool key wipe-after-use

### N2 (operability): pool.Decrypt 失败时 tenant 永远卡 Provisioning

`startTenantSchemaInitResume` (`server.go:735-737`)：
```go
plain, err := s.pool.Decrypt(workerCtx, t.DBPasswordCipher)
if err != nil {
    logger.Warn(workerCtx, "resume_schema_init_skipped", ...)
    return
}
```

如果 key rotation / KMS 不可用 → Decrypt 失败 → log warn → return → tenant **永远卡 Provisioning**。没有 mark-failed 或 alert metric。

**不是 PR 引入** (resume 逻辑老旧)，但 PR 让 resume 路径更重要（之前 sync request 至少在请求时立刻报错）。

**Suggestion**: 后续 PR 添加 metric 或 mark `Failed` 让 ops 看见。

### N3 (test gap): 没有 partial-failure-then-restart 集成测试

我跑了 unit tests，包括 idempotency。但没看到测试"schema init 跑到一半失败 → 服务器重启 → resume 接力完成"的端到端集成测试。

**Suggestion**: 后续 PR 加这条 integration test。当前 unit 覆盖足以 APPROVE。

## banked rule application

应用昨天 PR #617 banked 的 lifecycle-interleaving 规则：

- ✅ Code path 验证 (instantaneous)
- ✅ Lifecycle interleavings 显式 trace (LI1-LI3)
- ✅ tests 包括 -race
- ✅ Restart/recovery path 验证

**新发现**：PR #620 这次正好是个 **改善** lifecycle 的 PR — 从"request path 长 sync 等待 cluster ACTIVE"改成"async retry with persisted credentials"，是正面教材。dev-2 之前 trace 的 3 个风险面 + 我加的 3 个 LI 都 OK。

## Verdict

**APPROVE** PR #620。srstack LGTM 我同意，再加我独立验证的 lifecycle 部分。N1/N2/N3 是后续 PR 范围，**不阻** merge。

@dev-2 等你 PR start signal 后的 conclusion。@adversary-1 architectural perspective 可以补充。

qiffang:
**dev-2 review on `55d33571` — APPROVE, no merge-blocking issue from my lane**

Applied yesterday's banked review rule (mixed lifecycle interleavings, not just instantaneous code paths) to PR #620. Independent verdict.

## Verified locally

```
$ go test ./pkg/server -run 'TestSchemaInitForTenant|TestProvisionTiDBCloudNative|TestAdminTenantCreate|TestAdminTenantSetQuota|TestDecodeCredentialProvisionRequest' -count=1
ok  	github.com/mem9-ai/drive9/pkg/server	14.508s

$ go test ./pkg/tenant/schema ./pkg/tenant/tidbcloudnative -count=1
ok  	github.com/mem9-ai/drive9/pkg/tenant/schema	16.236s
ok  	github.com/mem9-ai/drive9/pkg/tenant/tidbcloudnative	10.715s

$ go test -race -run 'TestExecSchemaStatementsParallelByTable|TestParallelSchemaStatementGroups' ./pkg/tenant/schema -count=5
ok  	github.com/mem9-ai/drive9/pkg/tenant/schema	14.335s
```

Race detector clean across 5 iterations of the parallel-DDL tests.

## Lifecycle interleaving trace (yesterday's banked rule)

The PR shifts work from sync-in-request to async-in-background, which is exactly the kind of refactor where lifecycle interleavings need explicit checks. The three I traced:

**LI-1: First client request between `/provision` return and async schema-init completion**
- New code returns from /provision once cluster metadata + `tidb.cloud/organization` label are available, NOT when ACTIVE
- Tenant is left in `meta.TenantProvisioning` (server.go:4273)
- `pkg/server/auth.go:253-264` middleware rejects with HTTP 503 ("tenant is provisioning") while status is `TenantProvisioning`
- Only after `finalizeTenantSchemaInit` succeeds does `UpdateTenantStatusIf(Provisioning → Active)` flip (server.go:4503)
- **Safe**: client cannot reach the still-being-bootstrapped DB through normal request paths

**LI-2: Async schema-init partial failure → retry → idempotency**
- `initTenantSchemaAsync` (server.go:4477-4540) retries `schemaInit(ctx, tenantDSN)` with bounded backoff until `schemaInitRetryWindow`
- Each retry re-runs `EnsureDatabase` → `InitSchemaForAutoEmbeddingProfile`
- `EnsureDatabase` writes `CREATE DATABASE IF NOT EXISTS` (ensureDatabaseFromDSN → ensureDatabaseFunc) — idempotent
- Parallel DDL statements use `CREATE TABLE IF NOT EXISTS` and `isIgnorableSchemaError` swallows already-existing errors (common.go:execSchemaStatement) — idempotent
- **Safe**: retry-from-anywhere produces the same final state

**LI-3: Parallel DDL by table — concurrency invariants**
- `ExecSchemaStatementsParallelByTableContext` (common.go) groups statements by table, runs groups in parallel (capped at `maxParallelSchemaStatementGroups = 8`), preserves source order WITHIN a group
- For inter-table safety I checked the actual schemas: `mysql_no_embedding.go`, `tidb_app.go`, `tidb_auto.go` are pure DDL (CREATE TABLE + CREATE INDEX + ALTER TABLE). `grep -in "FOREIGN KEY\|REFERENCES "` returns nothing — no FK refs across tables in the bootstrap DDL
- Safety net: `parallelSchemaStatementGroups` returns `(nil, false)` if any statement fails `schemaStatementTableName` classification → caller falls back to `ExecSchemaStatementsContext` (sequential). Future seed-data INSERT or CREATE TRIGGER would degrade safely instead of running unclassified statements in parallel
- `schemaStatementTableName` correctly handles `CREATE TABLE IF NOT EXISTS` (via `ParseCreateTableStatement` internal/schemaspec/schemaspec.go:55-93), `CREATE [UNIQUE] INDEX ... ON tbl(col)` (strips `(col)` suffix), and `ALTER TABLE tbl ...`
- **Safe**: tables can run concurrently because they don't reference each other; statements within a table stay ordered

## Other checks

- `ensureDatabaseFromDSN` (provisioner.go:881) parses DSN, validates network=`tcp`, user/password/host non-empty, port positive, handles IPv6 brackets. New tests `TestEnsureDatabaseFromDSNRejectsNonTCPNetwork` / `TestEnsureDatabaseFromDSNHandlesIPv6Host` / `TestEnsureDatabaseFromDSNRejectsNonPositivePort` cover the validation paths
- `clusterProvisionMetadataIncomplete` (provisioner.go:1009) extends the readiness check to require `tidb.cloud/organization` label, not just connection endpoint — this is the PR's behavioral change to "return after metadata, not ACTIVE"
- `waitForClusterProvisionMetadata` renamed from `waitForClusterActive`; error message updated correspondingly
- `Provisioner.ProvisionWithCredentials` no longer calls `ensureDatabaseFunc` in the request path (lines 264-280) — moved to async via `EnsureDatabase` + `tenantDatabaseEnsurer` interface
- `InitSchema` (provisioner.go:127-131) keeps the ensure-database step for non-server direct callers — backward-compatible

## Non-blocker observations

(Largely overlapping with @adversary-2 8a26dacb's N1-N3; capturing for completeness from my lane.)

- **F1** (overlap with adversary-2 N1): `tenantDSN` is passed through `initTenantSchemaAsync` and held in the worker goroutine's closure for the full retry window (`schemaInitRetryWindow`). The DSN contains the cleartext DB password — that lifetime is longer than the previous in-request lifetime by ~the retry window. Non-blocker, but worth zeroing the slice when the worker exits if the retry window grows.

- **F2** (orthogonal): `parallelSchemaStatementGroups` requires every statement be classifiable. If a future schema migration adds a non-DDL bootstrap step (e.g., `INSERT INTO config_table VALUES (...)`), the function falls back to fully sequential execution rather than mixing parallel DDL + sequential INSERT. Correct safety net, but worth a comment in the function noting "intentional: any unclassified statement disables parallelism for the whole batch". Non-blocker.

- **F3** (test gap): `TestParallelSchemaStatementGroupsRejectsUnclassifiedStatements` confirms the no-classification path, but there's no test that exercises the `ExecSchemaStatementsParallelByTableContext` fast-path-falls-back behavior end-to-end with a mixed real-schema + unclassified stmt. Non-blocker; the unit test on the classifier is sufficient evidence.

## Verdict

LGTM on `55d33571`. F1/F2/F3 are quality follow-ups, not merge blockers. PR meaningfully improves the lifecycle posture (sync-in-request → async-with-status-gating) and the parallel DDL is safe given the schema set has no inter-table refs and degrades to sequential on any unclassified statement.

qiffang:
dev-1 review on head `55d335719c56f91a0d8a31bbdebe44abf4e037ef`: no merge-blocking issue found.

What I checked:
- Native provision request path now waits for connection metadata + `tidb.cloud/organization`, not cluster `ACTIVE`.
- Tenant remains `Provisioning` until async schema init succeeds, and normal API traffic is rejected with 503 while provisioning, so clients cannot hit a half-initialized DB through the normal path.
- Native DB creation moved to schema init and uses persisted tenant DB credentials, not the user TiDBCloud API key.
- `schemaInitForTenant` orders native auto-embedding init as `EnsureDatabase -> apply auto-embedding provider config -> InitSchemaForAutoEmbeddingProfile`.
- Starter schema repair/validation was restored in `55d3357`; the earlier no-op risk is gone.
- Parallel DDL preserves statement order within each table and falls back to serial if any statement cannot be classified. I did not find FK/cross-table schema dependencies in the current tenant schema statements.

Validation run locally:
```text
GOCACHE=/tmp/drive9-go-cache go test ./pkg/tenant/starter -count=1
GOCACHE=/tmp/drive9-go-cache go test ./pkg/tenant/schema -count=1
GOCACHE=/tmp/drive9-go-cache go test ./pkg/server -run 'TestSchemaInitForTenant|TestProvisionTiDBCloudNative|TestAdminTenantCreate|TestAdminTenantSetQuota|TestDecodeCredentialProvisionRequest' -count=1
GOCACHE=/tmp/drive9-go-cache go test ./pkg/tenant/tidbcloudnative -count=1
GOCACHE=/tmp/drive9-go-cache go test ./pkg/server ./pkg/tenant/schema ./pkg/tenant/tidbcloudnative ./pkg/tenant/starter -run 'TestProvision|TestAdminTenant|TestQuota|TestEnsureDatabaseFromDSN|TestParallelSchemaStatementGroups|TestExecSchemaStatementsParallel|TestProvisionerInitSchema|TestSchemaInitForTenant' -count=1
GOCACHE=/tmp/drive9-go-cache go test -race ./pkg/tenant/schema -run 'TestExecSchemaStatementsParallelByTable|TestParallelSchemaStatementGroups' -count=5
GOCACHE=/tmp/drive9-go-cache go test ./pkg/server -count=1
make lint
```
All passed. GitHub CI is also green at this head.

Residual non-blocking risks are the same shape others already called out: the async worker now holds a cleartext tenant DSN longer than the old sync path, decrypt failures during resume currently log-and-skip instead of marking the tenant failed, and a future cross-table DDL dependency would need to force serial execution or add an explicit dependency model. None of these block this PR.

PR #621, opened 2026-06-25
feat(observability): harden metrics/logs across dependencies, runtime, and process

srstack:
## Why

drive9 exported a rich `drive9_*` application surface but was missing the **dependency-health**, **runtime**, and **OS-process performance** signals the standardized observability requirements expect (e.g. a metadb that died after startup was invisible — pinged once at boot, the module gauge never flipped). This PR closes those gaps. All changes are additive instrumentation; request paths are unchanged.

## What

**Dependency health**
- DB health prober + `drive9_db_up{role}` / `drive9_db_unreachable_pools{role}` / `drive9_db_probe_duration_seconds` — meta/user stores are probed continuously, with `db_unavailable`/`db_recovered` transition logging.
- Leader election: `drive9_service_gauge{component="leader",name="is_leader"}` + acquire/heartbeat/keepalive error ops (previously fully silent; gates all background workers).
- Embedding/LLM provider: outbound call latency + error/timeout results.
- Encryption/KMS: encrypt/decrypt latency + errors via one decorator covering all backends.

**Runtime / process performance**
- Go runtime: `go_goroutines`, `go_threads`, `go_memstats_*`, GC counters.
- OS process (Linux `/proc`, no-ops off-Linux): `process_cpu_seconds_total`, RSS/VSZ, `process_open_fds`/`process_max_fds`.

**Other**
- Tenant pool cache hit/miss, eviction, and cached-backend size gauge.
- `handleTenantStatus` auth-failure events (aligned with the auth-middleware vocabulary).

## Tests
- New unit tests: DB probe flip/transition (`db_test.go`) and `/proc/self/stat` parser (`process_test.go`).
- `go build ./...`, `go vet`, and `go test ./pkg/metrics/... ./pkg/leader/... ./pkg/embedding/... ./pkg/encrypt/... ./pkg/tenant/...` pass. (pkg/server integration tests require a MySQL container not available locally; failures observed are all DB-dial, unrelated to this change.)

Alert rules consuming these metrics are in a companion runbooks PR.

## Summary by CodeRabbit

* **New Features**
  * Added continuous database health probing with availability and probe-duration metrics.
  * Added leader-election observability via a leader-state gauge and acquisition/heartbeat/keepalive metrics.
  * Added runtime and process metrics (CPU/memory, file descriptor limits, open FDs).
  * Added encryption and embedding operation metrics (including ok/canceled/timeout/error outcomes).
* **Bug Fixes**
  * Improved auth telemetry for additional authentication/authorization failure paths without changing responses.
  * Database availability metrics now reflect recovery/unavailability transitions automatically.
* **Tests**
  * Added database probe tests and Linux `/proc` stat parsing tests.

### DIFF
--- a/cmd/drive9-server/main.go
+++ b/cmd/drive9-server/main.go
@@ -26,6 +26,7 @@ import (
 	"github.com/mem9-ai/drive9/pkg/leader"
 	"github.com/mem9-ai/drive9/pkg/logger"
 	"github.com/mem9-ai/drive9/pkg/meta"
+	"github.com/mem9-ai/drive9/pkg/metrics"
 	"github.com/mem9-ai/drive9/pkg/s3client"
 	"github.com/mem9-ai/drive9/pkg/server"
 	"github.com/mem9-ai/drive9/pkg/slockoauth"
@@ -135,6 +136,21 @@ func main() {
 		}
 	}
 	defer func() { _ = store.Close() }()
+
+	// Continuously probe registered databases so that a metadata ("meta") or tenant
+	// ("user") store that becomes unreachable *after* startup is visible via the
+	// drive9_db_up gauge and the access log, not just indirectly through failing
+	// traffic. The probe covers the meta store immediately and tenant pools as they
+	// register. The startup ping above only proves reachability at boot.
+	probeInterval := time.Duration(envInt("DRIVE9_DB_HEALTH_PROBE_INTERVAL_SECONDS", 15)) * time.Second
+	probeTimeout := time.Duration(envInt("DRIVE9_DB_HEALTH_PROBE_TIMEOUT_SECONDS", 3)) * time.Second
+	metrics.StartDBHealthProbe(context.Background(), probeInterval, probeTimeout, func(role string, up bool, err error) {
+		if up {
+			logger.Info(context.Background(), "db_recovered", zap.String("role", role))
+			return
+		}
+		logger.Error(context.Background(), "db_unavailable", zap.String("role", role), zap.Error(err))
+	})
 
 	if s3cfg.Bucket == "" {
 		if err := os.MkdirAll(s3cfg.Dir, 0o755); err != nil {
--- a/pkg/embedding/openai.go
+++ b/pkg/embedding/openai.go
@@ -4,11 +4,14 @@ import (
 	"bytes"
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
 	"net/http"
 	"strings"
 	"time"
+
+	"github.com/mem9-ai/drive9/pkg/metrics"
 )
 
 // OpenAIClientConfig configures an OpenAI-compatible embeddings endpoint.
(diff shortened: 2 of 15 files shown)


mornyx (approved):
## Review: Harden metrics/logs across dependencies, runtime, and process

Reviewed the full diff (+699/-15 across 15 files). CI is green. All changes are additive instrumentation — no request-path logic changes.

### Strengths

1. **DB health probe is well-designed**: StartDBHealthProbe launches a background goroutine that pings all registered pools concurrently (semaphore-bounded at 8) every 15s. The drive9_db_up gauge defaults to 1 for unprobed roles (startup ping already succeeded), so cold-start false negatives are avoided. Transition logging (down->up, up->down) is fired outside the lock, and probeOnce is exported for direct test invocation. The TestDBHealthProbeFlipsDriveDBUp test covers the full healthy->down->recovery cycle with a fake driver.

2. **Leader election visibility**: drive9_service_gauge{component="leader",name="is_leader"} is now published on every acquire/heartbeat/keepalive/gain/lose transition. This closes a real gap — previously a leader-pod crash was invisible until background workers stopped running.

3. **Embedding/encrypt instrumentation**: Both use the same clean pattern — wrap the inner implementation, record RecordOperation with latency + categorized result (ok/canceled/timeout/error). The instrumentedEncryptor is applied in New() so all backends get instrumented transparently.

4. **Runtime metrics**: writeRuntimeMetrics emits go_goroutines, go_threads, go_memstats_*, GC counters. Uses runtime.ReadMemStats which briefly stops the world — acceptable at 15-30s scrape intervals, and matches client_golang behavior.

5. **Process metrics**: writeProcessMetrics reads /proc/self/stat with a robust parser that handles spaces in comm via last-")" splitting. The TestParseProcStat test covers a synthetic line with a space in the process name. Platform split via build tags is clean. FD counting via os.ReadDir with the self-fd subtraction is correct.

6. **Tenant pool metrics**: removeLocked now takes a reason parameter (evict/replace/invalidate/shutdown) so cache pressure (evict) is distinguishable from normal lifecycle. cache_lookup hit/miss metrics and cached_backends gauge are recorded at the right points in Acquire.

7. **Auth failure metrics in handleTenantStatus**: Each auth failure path now emits a metricEvent with a distinct result label (missing_token, key_not_found, hash_mismatch, etc.). This aligns the status endpoint with the auth middleware vocabulary.

### Items to confirm (non-blocking)

1. **runtime.ReadMemStats STW on every scrape**: The comment acknowledges this. At standard scrape intervals (15-30s) the STW is ~microseconds and negligible. Consider documenting the recommended scrape interval in the help text. Non-blocking.

2. **probeOnce concurrent ping with db.PingContext**: PingContext may open a new connection from the pool if all idle connections are in use. For a DB with SetMaxOpenConns(N), the probe counts as one of those N connections. This is correct behavior but worth noting that the probe adds 1 connection worth of pressure per role per cycle.

3. **userHZ = 100.0 hardcoded**: The comment correctly explains that reading the real value requires cgo. On all Linux kernels drive9 deploys to, USER_HZ is 100. Safe.

4. **removeLocked reason recorded even for shutdown**: When the pool closes, every backend is removed with reason="shutdown". The tenant_pool/remove counter will show a burst on process exit. Harmless but could create a dashboard spike if scraped during shutdown. Non-blocking.

### Verdict

**LGTM.** This is a pure additive observability PR with no request-path changes. The DB health probe, leader visibility, runtime/process metrics, and tenant pool instrumentation all fill real gaps. Test coverage for the DB probe and /proc parser is solid. The fake-driver pattern for DB probe testing is clean and does not require a real MySQL instance.

PR #622, opened 2026-06-25
Fix TiDBCloud Mode create-time quota handling

srstack:
## Summary

Fix TiDBCloud Mode create-time quota handling.

- Send `tidbcloud_spending_limit` as TiDB Cloud `spendingLimit.monthly` in the initial cluster create request instead of patching it after the cluster is created.
- Keep create labels limited to `drive9.ai/managed` and `drive9.ai/tenant_id`; do not add `drive9.ai/update_quota_at` during create.
- Persist create-time Drive9 quota knobs (`max_storage_size`, `max_file_size`, `max_file_count`) directly into Drive9 DB after the tenant connection is persisted.
- Keep normal set-quota/delete authorization behavior unchanged: later set-quota still uses patch-label authorization.
- Reject set-quota while the tenant is still `provisioning`, before patching TiDB Cloud labels.

## Notes

- I left CLI error parsing unchanged; the observed `parse error: Invalid numeric literal` looks like an e2e/script-side JSON parsing issue rather than a CLI behavior issue.

## Validation

- `GOCACHE=/tmp/drive9-go-cache go test ./pkg/tenant/tidbcloudnative -count=1`
- `GOCACHE=/tmp/drive9-go-cache go test ./pkg/server -run 'TestProvisionTiDBCloudNative|TestQuotaSet|TestAdminTenantQuotaSet|TestDeprecatedQuotaSet' -count=1`
- `GOCACHE=/tmp/drive9-go-cache go test ./cmd/drive9/cli -run 'TestCreate|TestAdminTenant|TestQuota' -count=1`
- `GOCACHE=/tmp/drive9-go-cache go test ./pkg/client -run 'TestAdmin|TestQuota' -count=1`
- `GOCACHE=/tmp/drive9-go-cache go test ./pkg/server -count=1`
- `make lint`
- `git diff --check`

### DIFF
--- a/pkg/server/admin_tenants.go
+++ b/pkg/server/admin_tenants.go
@@ -290,6 +290,10 @@ func (s *Server) handleAdminTenantQuotaSet(w http.ResponseWriter, r *http.Reques
 	if !ok {
 		return
 	}
+	if err := s.rejectQuotaSetForTenantStatus(t); err != nil {
+		errJSON(w, http.StatusConflict, err.Error())
+		return
+	}
 	cloudCfg, err := s.applyQuotaSet(r.Context(), t, cred, quotaReq)
 	if err != nil {
 		writeQuotaSetError(w, r.Context(), err, "update")
--- a/pkg/server/provision_test.go
+++ b/pkg/server/provision_test.go
@@ -203,6 +203,29 @@ func (f *fakeProvisioner) ProvisionCallCount() int {
 	return int(f.provisionCalls.Load())
 }
 
+type credentialOnlyProvisioner struct {
+	provider string
+	cluster  *tenant.ClusterInfo
+}
+
+func (f *credentialOnlyProvisioner) ProviderType() string { return f.provider }
+
+func (f *credentialOnlyProvisioner) InitSchema(_ context.Context, _ string) error { return nil }
+
+func (f *credentialOnlyProvisioner) Provision(_ context.Context, tenantID string) (*tenant.ClusterInfo, error) {
+	out := *f.cluster
+	out.TenantID = tenantID
+	out.Provider = f.provider
+	return &out, nil
+}
+
+func (f *credentialOnlyProvisioner) ProvisionWithCredentials(_ context.Context, tenantID string, _ tenant.CredentialProvisionRequest) (*tenant.ClusterInfo, error) {
+	out := *f.cluster
+	out.TenantID = tenantID
+	out.Provider = f.provider
+	return &out, nil
+}
+
 type profileAwareFakeProvisioner struct {
 	fakeProvisioner
 	mu               sync.Mutex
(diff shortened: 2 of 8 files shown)


srstack:
## Code Review — Fix TiDBCloud Mode create-time quota handling

Solid design and a genuine reliability improvement: putting `spendingLimit` into the cluster-create request makes it atomic with cluster creation, removing the fragile create→patch→deprovision-on-failure cycle. Validation of quota + credentials now also runs **before** provisioning (fail-fast).

### Strengths
- **Backward-compatible interface.** `ProvisionWithCredentials` delegates to `ProvisionWithCredentialsAndQuota`, preserving the optional-interface type-assertion pattern; non-native provisioners are unaffected.
- **Cleanup semantics preserved.** A failure in `applyQuotaLocalConfig` still routes through `cleanupProvisionedClusterAfterProvisionFailure`, so a local DB write failure won't leak a cloud cluster.
- **Strong test coverage.** Updated provision tests, new reject-on-provisioning tests for both `/v1/quota` and `/v1/admin/tenants/{id}/quota`, and a provisioner-level test asserting the create body contains `spendingLimit.monthly` and create labels do **not** include `update_quota_at`.
- **int32 overflow guarded.** `int32(*spendingLimit)` is safe because `validateTiDBCloudSpendingLimit` / `validateQuotaSetRequest` reject values `> maxInt32` before the cast.

### Issues / Suggestions

**1. (minor) Unreachable `TenantDeleted` branch in `rejectQuotaSetForTenantStatus`** — `quota.go`.
By the time `handleQuotaSet` reaches it, `quotaTenant` has already returned 404 for deleted tenants (`quota.go:363-366`). This branch would instead return **409 + "tenant not found"**, inconsistent with `quotaTenant`'s **404**. Suggest dropping the case or keeping it as purely defensive (harmless).

**2. (minor) `provisionCloudCfg` is only used for a metric** — `server.go:4322`.
`ProvisionWithCredentialsAndQuota` returns a full `QuotaCloudConfig`, but it's used solely to emit the `create_time_spending_limit` metric. This matches prior behavior (the spending limit lives cloud-side and was never persisted to the Drive9 DB), so no regression — but a one-line comment noting the cloud config is intentionally not persisted would prevent a future reader assuming it's dropped by mistake.

**3. (question) Behavior change: set-quota is now rejected during `provisioning`.**
For `/v1/quota` this is a user-visible change (409). Please confirm no existing client/e2e flow sets quota right after create while the tenant is still `provisioning` — that would now get a 409 instead of succeeding. The create-time path correctly bypasses this (it calls `applyQuotaLocalConfig` directly).

**4. (nit)** When `opts.Quota != nil` but the provisioner doesn't implement `CredentialQuotaProvisioner` (`server.go:4215`), the code falls into the cluster-error path with `cluster == nil`; please confirm `persistFailedProvisionClusterInfo(nil)` is nil-safe. This is a should-never-happen config error, so acceptable.

### Verdict
**Approve with minor comments.** The core change is correct and well-tested. Only #1 (unreachable branch) and confirming #3 doesn't break an existing immediate-set-quota flow are worth addressing — neither is blocking.

srstack:
Addressed the review items in 1c43e217:

- Local create-time quota persistence failures now return HTTP 500 instead of 502, with regression coverage for the metastore failure path.
- Missing CredentialQuotaProvisioner support for a quota-bearing TiDBCloud Mode create request now fails fast as HTTP 500 and marks the tenant failed, instead of falling through to the generic upstream 502 path.
- Removed the unreachable TenantDeleted branch from rejectQuotaSetForTenantStatus; deleted tenants continue to be handled by the existing lookup paths as 404.
- Added a comment clarifying that create-time TiDB Cloud spending limit stays cloud-side and is read back from TiDB Cloud for list/get quota.

Confirmed behavior: rejecting set-quota while a tenant is still provisioning is intentional. Create-time quota is the supported path for provisioning; post-create set-quota remains for active tenants.

Validation:
- GOCACHE=/tmp/drive9-go-cache go test ./pkg/server -run 'TestProvisionTiDBCloudNative|TestQuotaSet|TestAdminTenantQuotaSet|TestDeprecatedQuotaSet' -count=1
- GOCACHE=/tmp/drive9-go-cache go test ./pkg/server -count=1 -failfast
- make lint
- git diff --check

srstack:
## Re-review — commit `1c43e217` (address review feedback)

Re-reviewed after the latest commit. All previously raised points are addressed and `go vet` is clean. 👍

- **#1 (unreachable `TenantDeleted` branch)** — Removed from `rejectQuotaSetForTenantStatus`. ✅
- **#2 (`provisionCloudCfg` only used for metric)** — Clarifying comment added explaining the spending limit stays cloud-side and is read back from TiDB Cloud on list/get. ✅
- **#4 (unsupported-provisioner path with `cluster == nil`)** — Now logs, emits the error metric, marks the tenant `failed`, and returns a clean **500** instead of falling into the cluster-error path with a nil cluster. Backed by the new `TestProvisionTiDBCloudNativeCreateTimeQuotaRequiresQuotaProvisioner`. ✅

Additional improvements in this commit:
- **Status code corrected** — create-time local quota persistence failure now returns **500** (was 502 `BadGateway`), which is accurate since it's a local DB write, not an upstream call.
- **New test `TestProvisionTiDBCloudNativeCreateTimeQuotaLocalPersistenceErrorIsInternal`** — verifies that a local persistence failure returns 500, deprovisions the cluster (cleanup), and marks the tenant `failed`. Good coverage of the rollback path.

#3 (set-quota now rejected with 409 during `provisioning`) was a behavior-change question, not a required fix — it's intentional per the PR description; just make sure no client relies on the old immediate-set-quota-after-create behavior.

### Verdict
**LGTM / Approve.** All feedback resolved, no new concerns.

PR #624, opened 2026-06-26
feat: add tenant_id dimension to metrics, logs, and alert rules

srstack:
## Summary

Add tenant_id dimension to per-tenant metrics, logs, and alert rules to enable tenant-level observability and faster root-cause identification during incidents.

### Metrics (25 files, 85+ call sites)
- Add `RecordTenantOperation`, `RecordTenantGauge`, `RecordTenantEvent` to metrics package
- Update `metricEvent` to extract tenant_id from request context automatically
- Convert all per-tenant `RecordOperation` calls to `RecordTenantOperation` across:
  - `backend` (20 call sites: create/write/read/rename/copy/symlink/grep/find/exec_sql)
  - `vault` (15 call sites: secret/token/grant/read/audit)
  - `central_quota` / `server_quota` (22 call sites)
  - `semantic_worker` (11 additional call sites: retry/dead_letter/recover/renew)
  - `file_gc`, `mutation_replay`, `quota_outbox`, `llm_cost_budget`
  - `image_extract`, `audio_extract`, `event_bus`

### Logs (8 files, 24 call sites)
- Add tenant_id to log statements in background workers, SSE, eventbus, patch upload, and LLM usage paths

### Struct changes
- Add `TenantID` field to `ImageExtractRequest`, `quotaPendingDeltasCache`, and `EventBus`

## Motivation

During the Jun 26 production incident, 5 alerts fired (BackgroundWorkerErrorsHigh, SemanticWorkerDeadLettered, SemanticWorkerUnavailable, HighHTTPP99Latency, UserDBUnavailable) but none had tenant_id, making it impossible to identify which tenant's database was degraded. This PR ensures future incidents can be traced to the specific affected tenant.

## Summary by CodeRabbit

* **New Features**
  * Added tenant-scoped usage and activity tracking across backend operations, files/GC, uploads, event delivery, and semantic workflows.
  * Tenant-aware metric/event recording is now consistently applied, and dashboards/tests were updated to include `tenant_id`.

* **Bug Fixes**
  * Improved quota and budget skip/fail-open telemetry for more consistent tenant-level reporting.
  * Enhanced vault behavior by improving duplicate-entry detection and requiring tenant scope for vault audit requests (returns `403` when missing).
  * Expanded tenant context in logs and metrics for clearer diagnosis of retries and failures.

### DIFF
--- a/docs/grafana/drive9-async-fuse-dashboard.json
+++ b/docs/grafana/drive9-async-fuse-dashboard.json
@@ -132,7 +132,7 @@
             "type": "prometheus",
             "uid": "${datasource}"
           },
-          "expr": "drive9_service_gauge{component=\"semantic_worker\",name=\"queue_lag_seconds\"}",
+          "expr": "sum(drive9_service_gauge{component=\"semantic_worker\",name=\"queue_lag_seconds\",tenant_id=\"\"})",
           "legendFormat": "queue_lag_seconds",
           "refId": "A"
         }
@@ -199,7 +199,7 @@
             "type": "prometheus",
             "uid": "${datasource}"
           },
-          "expr": "drive9_service_gauge{component=\"semantic_worker\",name=\"dead_lettered\"}",
+          "expr": "sum(drive9_service_gauge{component=\"semantic_worker\",name=\"dead_lettered\",tenant_id=\"\"})",
           "legendFormat": "dead_lettered",
           "refId": "A"
         }
@@ -409,7 +409,7 @@
             "type": "prometheus",
             "uid": "${datasource}"
           },
-          "expr": "drive9_service_gauge{component=\"semantic_worker\",name=~\"inflight|queued|processing|dead_lettered\"}",
+          "expr": "sum by (name) (drive9_service_gauge{component=\"semantic_worker\",name=~\"inflight|queued|processing|dead_lettered\",tenant_id=\"\"})",
           "legendFormat": "semantic_worker {{name}}",
           "refId": "A"
         }
@@ -456,7 +456,7 @@
             "type": "prometheus",
             "uid": "${datasource}"
           },
-          "expr": "drive9_service_gauge{component=\"semantic_worker\",name=\"queue_lag_seconds\"}",
+          "expr": "sum(drive9_service_gauge{component=\"semantic_worker\",name=\"queue_lag_seconds\",tenant_id=\"\"})",
           "legendFormat": "semantic_worker queue_lag_seconds",
           "refId": "A"
         },
--- a/pkg/backend/audio_extract.go
+++ b/pkg/backend/audio_extract.go
@@ -18,6 +18,7 @@ import (
 
 // AudioExtractRequest is the input to a pluggable audio->text extractor.
 type AudioExtractRequest struct {
+	TenantID    string
 	FileID      string
 	Path        string
 	ContentType string
@@ -336,6 +337,7 @@ func (b *Dat9Backend) ProcessAudioExtractTask(ctx context.Context, task AudioExt
 
 	taskCtx, cancel := context.WithTimeout(ctx, b.audioExtractTimeout)
 	text, audioUsage, err := b.audioExtractor.ExtractAudioText(taskCtx, AudioExtractRequest{
+		TenantID:    b.tenantID,
 		FileID:      task.FileID,
 		Path:        task.Path,
 		ContentType: extractorContentType,
--- a/pkg/backend/audio_extract.go
+++ b/pkg/backend/audio_extract.go
@@ -284,7 +284,7 @@ func (b *Dat9Backend) ProcessAudioExtractTask(ctx context.Context, task AudioExt
 	start := time.Now()
 	var result AudioExtractResult
 	defer func() {
-		metrics.RecordOperation("audio_extract", "process", legacyAudioExtractMetricResult(result), time.Since(start))
+		metrics.RecordTenantOperation(b.tenantID, "audio_extract", "process", legacyAudioExtractMetricResult(result), time.Since(start))
 	}()
 
 	if !b.SupportsAsyncAudioExtract() {
@@ -292,7 +292,7 @@ func (b *Dat9Backend) ProcessAudioExtractTask(ctx context.Context, task AudioExt
 		return result, fmt.Errorf("async audio extract runtime not configured")
 	}
 	if b.monthlyLLMCostExceededCheck(ctx) {
-		metrics.RecordOperation("llm_cost_budget", "process_skip", "budget_exhausted", 0)
+		metrics.RecordTenantOperation(b.tenantID, "llm_cost_budget", "process_skip", "budget_exhausted", 0)
 		result = AudioExtractResultBudgetExhausted
 		return result, nil
 	}
--- a/pkg/backend/audio_extract_test.go
+++ b/pkg/backend/audio_extract_test.go
@@ -99,7 +99,7 @@ func TestAsyncAudioExtractMetricsExposed(t *testing.T) {
 	if !strings.Contains(metricsText, "drive9_module_up{module=\"audio_extract\"} 1") {
 		t.Fatalf("metrics missing audio_extract module availability: %s", metricsText)
 	}
-	if !strings.Contains(metricsText, "drive9_service_gauge{component=\"audio_extract\",name=\"max_audio_bytes\"} 4096") {
+	if !strings.Contains(metricsText, "drive9_service_gauge{component=\"audio_extract\",name=\"max_audio_bytes\",tenant_id=\"\"} 4096") {
 		t.Fatalf("metrics missing audio_extract runtime gauge: %s", metricsText)
 	}
 	if !strings.Contains(metricsText, "drive9_service_operations_total{component=\"audio_extract\",operation=\"process\",result=\"ok\"}") {
--- a/pkg/backend/dat9.go
+++ b/pkg/backend/dat9.go
@@ -243,7 +243,7 @@ func (b *Dat9Backend) Create(path string) error {
 
 func (b *Dat9Backend) CreateCtx(ctx context.Context, path string) (err error) {
 	start := time.Now()
-	defer func() { observeBackend(ctx, "create", err, start) }()
+	defer func() { observeBackend(ctx, b.tenantID, "create", err, start) }()
 
 	path, err = pathutil.Canonicalize(path)
 	if err != nil {
@@ -327,7 +327,7 @@ func (b *Dat9Backend) CreateCtx(ctx context.Context, path string) (err error) {
 // payload and whose inode mode carries the POSIX symlink file type bits.
 func (b *Dat9Backend) CreateSymlinkCtx(ctx context.Context, linkPath, target string) (err error) {
 	start := time.Now()
-	defer func() { observeBackend(ctx, "create_symlink", err, start) }()
+	defer func() { observeBackend(ctx, b.tenantID, "create_symlink", err, start) }()
 
 	data := []byte(target)
 	if target == "" || strings.ContainsRune(target, 0) || len(data) > MaxSymlinkTargetBytes {
@@ -414,7 +414,7 @@ func (b *Dat9Backend) Mkdir(path string, perm uint32) error {
 
 func (b *Dat9Backend) MkdirCtx(ctx context.Context, path string, perm uint32) (err error) {
 	start := time.Now()
-	defer func() { observeBackend(ctx, "mkdir", err, start) }()
+	defer func() { observeBackend(ctx, b.tenantID, "mkdir", err, start) }()
 
 	dirPath, err := pathutil.CanonicalizeDir(path)
 	if err != nil {
@@ -462,7 +462,7 @@ func (b *Dat9Backend) Chmod(path string, mode uint32) error {
 
 func (b *Dat9Backend) ChmodCtx(ctx context.Context, path string, mode uint32) (err error) {
 	start := time.Now()
-	defer func() { observeBackend(ctx, "chmod", err, start) }()
+	defer func() { observeBackend(ctx, b.tenantID, "chmod", err, start) }()
 
 	path, err = pathutil.Canonicalize(path)
 	if err != nil {
@@ -484,7 +484,7 @@ func (b *Dat9Backend) Remove(path string) error {
 
 func (b *Dat9Backend) RemoveCtx(ctx context.Context, path string) (err error) {
 	start := time.Now()
-	defer func() { observeBackend(ctx, "remove", err, start) }()
+	defer func() { observeBackend(ctx, b.tenantID, "remove", err, start) }()
 
 	path, node, err := b.resolveNodePath(ctx, path)
 	if err != nil {
@@ -502,7 +502,7 @@ func (b *Dat9Backend) RemoveCtx(ctx context.Context, path string) (err error) {
 
 func (b *Dat9Backend) RemoveFileCtx(ctx context.Context, path string) (err error) {
 	start := time.Now()
-	defer func() { observeBackend(ctx, "remove_file", err, start) }()
+	defer func() { observeBackend(ctx, b.tenantID, "remove_file", err, start) }()
 
 	path, err = pathutil.Canonicalize(path)
 	if err != nil {
@@ -517,7 +517,7 @@ func (b *Dat9Backend) RemoveFileCtx(ctx context.Context, path string) (err error
 
 func (b *Dat9Backend) RemoveDirCtx(ctx context.Context, path string) (err error) {
 	start := time.Now()
-	defer func() { observeBackend(ctx, "remove_dir", err, start) }()
+	defer func() { observeBackend(ctx, b.tenantID, "remove_dir", err, start) }()
 
 	path, err = pathutil.CanonicalizeDir(path)
 	if err != nil {
@@ -535,7 +535,7 @@ func (b *Dat9Backend) RemoveAll(path string) error {
 
 func (b *Dat9Backend) RemoveAllCtx(ctx context.Context, path string) (err error) {
 	start := time.Now()
-	defer func() { observeBackend(ctx, "remove_all", err, start) }()
+	defer func() { observeBackend(ctx, b.tenantID, "remove_all", err, start) }()
 
 	path, node, err := b.resolveNodePath(ctx, path)
 	if err != nil {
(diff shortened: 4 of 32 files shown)


srstack:
## Code Review — feat: add tenant_id dimension to metrics, logs, and alert rules

Reviewed the full diff (25 files). The core mechanism is clean: new `RecordTenantOperation` / `RecordTenantGauge` / `RecordTenantEvent`, with the legacy `RecordOperation`/`RecordGauge`/`RecordEvent` delegating with an empty tenant id. `go vet ./pkg/{metrics,server,backend}` is clean.

### Strengths
- **Backward-compatible + cardinality guard.** The `tenant_id` label is only attached when non-empty (`cleanMetricValue(...) != "unknown"`), so existing aggregate series and `sum(...)` alerts keep working unchanged. Good defensive design.
- **Explicit IDs where context isn't available.** `tenant_schema_init` passes the real `tenantID` directly (`server.go:4519/4535`), and `provisionTenant` sets the metric scope at `server.go:4155` before the cluster-error events — so the provision/schema alerts the PR targets will actually carry `tenant_id`.
- **Sensible defaults.** `handleVaultAudit` defaults to `"unknown"` then upgrades from scope; the `EventBus`/`quotaPendingDeltasCache`/`ImageExtractRequest` struct threading is consistent. Tests updated.

### Issues / Suggestions

**1. (important) `tenant_id` is added to the duration *histogram*, not just the counter.**
`RecordTenantOperation` tags both `serviceOperationsTotal` **and** `serviceOperationDuration` (`operations.go`). Histograms are the most expensive series type — cost is `tenants × le-buckets × operations × results`. For `backend` (~20 ops) this is a large multiplier. The only per-tenant alert that needs the histogram is `Drive9BackendUploadLatencyP95High`. Please confirm the active-tenant count is bounded (hundreds, not thousands+), and consider whether `tenant_id` belongs on the duration histogram at all vs. counters-only. This is the main operational risk of the PR.

**2. (important, cross-PR with runbooks#2616) Semantic-worker gauge is emitted twice under the same name.**
`observeOnce` emits an aggregate `RecordGauge("semantic_worker","dead_lettered",…)` **and** a per-tenant `RecordTenantGauge(tenantID,…)` loop. That produces both a no-`tenant_id` series and per-tenant series under the same metric name. The two *unchanged* runbooks alerts (`Drive9SemanticWorkerDeadLettered`, `Drive9SemanticWorkerQueueLagHigh`) select `drive9_service_gauge{name=…}` without a `tenant_id` matcher, so they now match the aggregate **and** every per-tenant series → each fires per tenant, double-paging with the new `Drive9SemanticWorkerTenant*` alerts. Fix belongs in the runbooks PR (add `tenant_id=""` to the aggregate alerts), but flagging here since it stems from the dual emission. (Counters don't have this problem — they're emitted once.)

**3. (minor) Earliest provision-validation event lacks `tenant_id`.**
`metricEvent(ctx, "tenant_provision", …, "result", "error")` at `server.go:4143` fires *before* `setRequestMetricTenant` (4155) / before the tenant ID exists, so it stays `unknown`. Combined with the runbooks alert being rewritten in place to filter `tenant_id!=""`, pre-tenant validation failures are no longer alerted. Acceptable (those aren't tenant-specific incidents), just calling out the coverage change.

### Verdict
**Approve with comments.** Mechanically sound and well-scoped. Please (a) confirm the histogram cardinality budget (#1) and (b) coordinate the gauge-alert fix in runbooks#2616 (#2).

srstack:
@coderabbitai full review

srstack (commented):
All CodeRabbit comments addressed in commit 4e4efaa:

1. upload_reservation.go: return nil immediately on pending delta fail-open
2. semantic_worker.go: reset stale per-tenant gauges for disappeared tenants  
3. vault.go: set auditTenantID from scope before early returns
4. audio_extract.go: add TenantID to AudioExtractRequest for consistency

srstack:
## Re-review — commits since `820322ba`

Re-reviewed the 4 follow-up commits. My earlier feedback is well addressed, and two genuine bug fixes landed beyond the original observability scope. `go vet ./pkg/{server,backend,metrics}` is clean.

### Resolved from prior review
- **Gauge dual-emission / fan-out (#2)** — Handled on the consumer side: the Grafana dashboard now uses `sum(... tenant_id="")` for the semantic_worker gauges, and runbooks#2616 scopes the aggregate alerts to `tenant_id=""`. Aggregate vs. per-tenant series are now cleanly separated. 👍

### New fixes in this round — both correct
- **Per-tenant gauge staleness (`semantic_worker.go`)** — The new `priorObservedTenants` tracking emits a final `0` for tenants that drop out of the observation set. This is the right fix: without it, a per-tenant `dead_lettered`/`queue_lag_seconds` series would freeze at its last non-zero value and keep the new per-tenant alerts firing forever after the tenant recovered or was removed. Note the series are zeroed, not deleted, so per-tenant gauge cardinality still accumulates over process lifetime — bounded by distinct tenants observed, acceptable.
- **`upload_reservation.go` fail-open `return nil`** — Real bug fix. The `!pendingOK` branch recorded a `fail_open` metric but previously fell through and enforced the quota with a zeroed `pendingStorageDelta`. Adding `return nil` makes it consistently fail-open, matching the `cfg == nil` / `usage == nil` branches and the metric's own intent. Good catch.
- **`handleVaultAudit`** — Extracting the scope once up front and using `auditTenantID` for both the metric and `QueryAuditLog` removes the prior reliance on a separately-derived `scope.TenantID` (nil-safety) and guarantees the metric is always tagged. Clean.

### Still open (carry-over, your call)
- **Histogram cardinality (#1)** — `tenant_id` is still on the duration histogram, consumed by `Drive9BackendUploadLatencyP95High`. Just confirm the active-tenant count is bounded; no code change needed if it is.

### Verdict
**Approve.** Follow-ups are correct and the two incidental bug fixes are worth having.

qiffang (commented):
## Code Review — PR #624

### B3 (Blocker): Grafana dashboard queries will break

**Code path:**
1. `RecordGauge("semantic_worker", "queue_lag_seconds", val)` → calls `RecordTenantGauge("", ...)`
2. `cleanMetricValue("", "unknown")` → `tenantID = "unknown"`
3. `if tenantID != "unknown"` → false → **`tenant_id` label is NOT added**
4. Grafana queries use `tenant_id=""` filter → matches series where label exists and is empty → **matches 0 series**

Aggregate gauges (no `tenant_id` label) ≠ `tenant_id=""`. The dashboard panels for `queue_lag_seconds`, `dead_lettered`, and `inflight|queued|processing|dead_lettered` will show no data.

**Fix (pick one):**
- A: Always emit `tenant_id` label on aggregate gauges (value `"_all"` or `"aggregate"`), update Grafana to filter `tenant_id="_all"`
- B: Remove `tenant_id=""` filter from aggregate Grafana queries, or use `tenant_id=~".*|"`

**Files:** `docs/grafana/drive9-async-fuse-dashboard.json` (4 panels), `pkg/metrics/metrics.go` (`RecordTenantGauge`)

---

### B1: `handleVaultAudit` queries with `"unknown"` tenantID when scope is nil

When `ScopeFromContext` returns nil, `auditTenantID` stays `"unknown"` and `QueryAuditLog` is called with `tenantID="unknown"`. If this handler can be reached without a valid scope, it silently returns wrong results instead of erroring.

**Action:** Confirm `handleVaultAudit` always has a valid scope (middleware enforced), or add an early return/error when scope is nil.

---

### O3 (Note): `upload_reservation.go` — behavioral fix bundled in metrics PR

The `return nil` addition after the `!pendingOK` block is a **behavior change** (was falling through with zero deltas), not just metrics plumbing. Should be documented in PR description.

---

### Cardinality (non-blocking)

`RecordTenantOperation` adds `tenant_id` to every operation metric. With N tenants × M operations, confirm Prometheus/VictoriaMetrics can handle the series count. Document expected tenant cardinality.

---

**Overall:** Direction is correct — tenant dimension is essential for incident RCA. Fix B3 (Grafana queries), confirm B1 (vault scope), and this is ready to merge.

qiffang (changes requested):
## Adversarial Review — PR #624

**Verdict: REVISE** — 2 blockers, 1 behavioral note

---

### B1: `handleVaultAudit` queries with `"unknown"` tenantID when scope is nil

`vault.go` — `auditTenantID` defaults to `"unknown"` and is used directly in `QueryAuditLog`:

```go
auditTenantID := "unknown"
if scope := ScopeFromContext(r.Context()); scope != nil {
    auditTenantID = scope.TenantID
}
// ...
events, err := vs.QueryAuditLog(r.Context(), auditTenantID, secretName, limit)
```

The original code read `scope.TenantID` after a separate nil check. The refactored code uses `auditTenantID` for the query — if `ScopeFromContext` returns nil, `QueryAuditLog` runs with `tenantID="unknown"`, silently returning empty results instead of erroring. Verify this path is unreachable or add a guard.

---

### B3 (blocker): Grafana `tenant_id=""` filter breaks aggregate dashboards

Code path:
1. `RecordGauge("semantic_worker", ...)` → calls `RecordTenantGauge("", ...)`
2. `cleanMetricValue("", "unknown")` → `tenantID = "unknown"`
3. `if tenantID != "unknown"` → false → **`tenant_id` label is NOT emitted**
4. Grafana queries `drive9_service_gauge{...,tenant_id=""}` → matches series with label explicitly set to empty → **matches 0 series**

A missing label ≠ `label=""` in Prometheus. The aggregate gauges (queue_lag, dead_lettered, worker state, media extract) will disappear from all 4 updated Grafana panels.

Fix options:
- A: Always emit `tenant_id` label (use `"_aggregate"` for non-tenant gauges), update Grafana filter to match
- B: Remove `tenant_id=""` filter from aggregate Grafana queries, or use `tenant_id=~".*|"`

---

### O3 (note): `upload_reservation.go` — behavioral fix bundled in metrics PR

The `!pendingOK` branch now has `return nil` — the original code fell through using zero-valued deltas. This is a correctness fix, not just metrics plumbing. Should be called out in the PR description as a behavior change.

---

### Non-blocking

- **Cardinality**: Confirm expected tenant count and that metrics backend handles `O(tenants × operations)` series.
- **EventBus struct alignment**: `tenantID`/`store` fields have inconsistent tab alignment (cosmetic).
- **Stale gauge reset** (`priorObservedTenants`): Correct approach — disappeared tenants get gauges reset to 0. 👍

Fix B1 + B3 for re-review.

qiffang:
Review verdict: REVISE

I found two blocking issues that should be fixed before merge.

1. `handleVaultAudit` can query audit logs with `tenantID="unknown"` if `ScopeFromContext(r.Context())` returns nil. That silently changes authorization/tenant scoping behavior instead of failing closed. Please either prove this handler always has a valid scope before this point, or add an explicit guard returning an error when scope is missing.

2. Grafana dashboard queries filter aggregate gauges with `tenant_id=""`, for example semantic worker queue lag. The new tenant-aware gauge path omits `tenant_id` for aggregate series rather than emitting an empty label, and Prometheus does not treat a missing label as `tenant_id=""`. Those panels will match zero series. Please either always emit the label for aggregate metrics or remove/change the tenant_id filter for aggregate dashboard queries.

Non-blocking notes:
- Adding tenant_id as a label materially increases metrics cardinality. Please document the expected tenant cardinality and confirm the backend/dashboard setup is intended to handle it.
- The `upload_reservation.go` `return nil` on pending quota outbox failure is a correct behavior fix, but it is not just metrics plumbing and should be called out in the PR description.
- Minor style: `EventBus` struct field alignment changed inconsistently.

qiffang (changes requested):
REVISE from @dat9-dev2.

I agree with the goal: adding tenant_id to operational metrics/logs is needed for incident RCA. The implementation is directionally right, but it is incomplete in one important path.

## Blocker: quota/upload metrics still omit tenant_id

The PR title/body says this adds tenant_id to per-tenant metrics, including central_quota/server_quota paths, but several quota/upload hot-path metrics are still emitted through RecordOperation, so the incident class this PR is meant to fix can still produce un-attributable alerts.

Examples in the current PR head:

- pkg/backend/quota.go:126 / 133 / 136: file_size_check still uses RecordOperation.
- pkg/backend/quota.go:161 / 164 / 212 / 220 / 226: storage_check and pending-delta failures still use RecordOperation.
- pkg/backend/quota.go:178 / 186 / 192 / 196 / 200: file_count_check still uses RecordOperation.
- pkg/backend/quota.go:244 / 247: admission_lock still uses RecordOperation.
- pkg/backend/quota.go:335 / 343 / 349 and 414: media quota / media_llm_budget paths still use RecordOperation.
- pkg/backend/upload.go still emits backend initiate/presign/confirm/finalize/resume/abort upload metrics through RecordOperation, including quota_exceeded on initiate.

These are all tenant-scoped backend paths (`Dat9Backend` has `b.tenantID`). If a tenant hits file-size/storage/file-count/media quota errors or upload failures, the resulting metrics still do not carry tenant_id, which conflicts with the PR motivation.

Please either:
1. convert these tenant-owned quota/upload call sites to RecordTenantOperation(b.tenantID, ...), with regression coverage for representative metrics output; or
2. explicitly document why these paths are intentionally excluded from tenant attribution and update the PR description so it does not claim full central_quota/server_quota/backend coverage.

## Non-blocking clarifications

- The `upload_reservation.go` `return nil` after `pendingOK=false` is a real behavior fix bundled into this metrics PR. I think the change is correct, but it should be called out in the PR summary because it changes quota admission behavior, not just observability.
- I do not consider `handleVaultAudit` querying `unknown` a demonstrated runtime blocker: `/v1/vault/audit` is routed through the tenant-auth/fallback business handler, and `vaultStore()` returns before `QueryAuditLog` if scope is missing. A small defensive guard would still be clearer.
- For the Grafana `tenant_id=""` query: in Prometheus semantics, a matcher for the empty string also matches series where the label is absent, so it can match the aggregate `RecordGauge` series. If the production datasource differs from Prometheus semantics, then the dashboard query should be adjusted; otherwise this is not necessarily a blocker.

qiffang (commented):
## Re-review — PR #624 (updated)

### B3 partially fixed

Code side is correct: `RecordTenantGauge` now always emits `tenant_id` label (the `if tenantID != "unknown"` guard was removed). Good.

But the Grafana dashboard queries still filter `tenant_id=""`:
```
sum(drive9_service_gauge{component="semantic_worker",name="queue_lag_seconds",tenant_id=""})
```

When `RecordGauge("semantic_worker", ...)` is called, it goes through `RecordTenantGauge("")` → `cleanMetricValue("", "unknown")` → label is set to `tenant_id="unknown"`, not `tenant_id=""`.

So `tenant_id=""` still matches 0 series. The Grafana queries need to use either:
- `tenant_id="unknown"` to match aggregate gauges, or
- Remove the `tenant_id` filter entirely from aggregate panels (just `sum(drive9_service_gauge{component="semantic_worker",name="queue_lag_seconds"})`)

4 panels in `docs/grafana/drive9-async-fuse-dashboard.json` need this fix.

qiffang (changes requested):
Requesting changes on the updated head `7de483d`.

My previous backend quota/upload attribution blocker is addressed: the tenant-owned quota and upload paths I called out now use `RecordTenantOperation(b.tenantID, ...)`. The remaining blocker is a new gauge-label regression that is already breaking CI.

**B1 — `RecordGauge` now emits `tenant_id="unknown"` for global gauges, breaking existing metrics/tests and Grafana selectors**

`RecordGauge` delegates to `RecordTenantGauge("", ...)`, and `RecordTenantGauge` sanitizes the empty tenant to `"unknown"` then always emits `Attr("tenant_id", tenantID)`. That means every global service gauge now changes from:

`drive9_service_gauge{component="audio_extract",name="max_audio_bytes"} ...`

to:

`drive9_service_gauge{component="audio_extract",name="max_audio_bytes",tenant_id="unknown"} ...`

This is not just cosmetic:

- CI is red on `TestAsyncAudioExtractMetricsExposed` and `TestBackendRuntimeMetricsAggregateAcrossBackends` because those tests still expect unscoped global gauges.
- The dashboard changes query global semantic-worker gauges with `tenant_id=""` (for example `drive9_service_gauge{component="semantic_worker",name="queue_lag_seconds",tenant_id=""}`), but the exporter now emits `tenant_id="unknown"`, so those panels will not match the global series. In Prometheus, `tenant_id=""` works for absent labels; it does not match `tenant_id="unknown"`.
- `RecordTenantOperation` / `RecordTenantEvent` already omit `tenant_id` when no tenant is supplied. Gauge behavior should follow the same contract: unscoped/global metrics have no tenant label; tenant-scoped gauges include the concrete tenant label.

Required fix:

- Make `RecordTenantGauge` build attrs conditionally like `RecordTenantOperation`: only append `tenant_id` when a real tenant ID is supplied.
- Keep tenant-scoped semantic gauges labeled with the concrete tenant ID.
- Keep/update regression tests covering both global `RecordGauge(...)` output without `tenant_id` and `RecordTenantGauge("tenant-a", ...)` output with `tenant_id="tenant-a"`.
- CI must be green.

Validation I ran locally:

- `/usr/local/go/bin/go test -c ./pkg/metrics ./pkg/backend ./pkg/server` — pass
- focused runtime tests are blocked locally by testcontainers (`rootless Docker not found`), but GitHub CI is red and shows the exact metric-test failures above.

qiffang (approved):
## Re-review — PR #624 (latest head)

B3 is now fixed correctly:
- `RecordTenantGauge` always emits `tenant_id` label — empty string for global gauges, real ID for tenant-specific
- Grafana queries filter `tenant_id=""` to match global aggregates — consistent
- Tests updated to expect `tenant_id=""` on global gauges
- CI all green (ci, e2e smoke, CodeRabbit)

B1 was already fixed in the previous update (nil scope → 403).

LGTM. Ship it.

qiffang (approved):
## Adversarial Review — PR #624 (REVISED, final)

**Verdict: GREEN**

Re-reviewed after 3 fix commits (`e910bfe8`, `7de483d6`, `8aaa3428`).

### Blockers resolved

- **B1 (vault audit)**: Fixed — nil scope now returns 403 early. ✓
- **B3 (Grafana gauge label)**: Fixed — `RecordTenantGauge` uses `strings.TrimSpace(tenantID)` so global gauges emit `tenant_id=""`, matching Grafana `tenant_id=""` queries. Tests updated. ✓
- **Coverage gap** (quota/upload): Fixed in `e910bfe8` — `upload.go` and `quota.go` now use `RecordTenantOperation`. ✓
- **CI**: All green (ci, e2e smoke, CodeRabbit). ✓

### Note

`RecordTenantGauge` always emits `tenant_id` (even empty), while `RecordTenantOperation`/`RecordTenantEvent` conditionally add it. Functionally both work with `tenant_id=""` selectors, but the inconsistency is worth noting for future maintainers.

LGTM.

qiffang (approved):
Re-reviewed head `8aaa342`.

The prior blocker is resolved. `RecordGauge` now emits aggregate gauges with `tenant_id=""`, tenant-specific semantic gauges still carry the concrete tenant ID, and the Grafana aggregate selectors using `tenant_id=""` match the aggregate series. This is a consistent contract and CI is green.

Checked:

- Backend quota/upload tenant attribution gaps from my earlier review are fixed.
- `handleVaultAudit` now fails closed on missing scope.
- Global/aggregate gauge tests were updated to the explicit empty-label contract.

Local validation:

- `/usr/local/go/bin/go test -c ./pkg/metrics ./pkg/backend ./pkg/server`

Approve.

PR #625, opened 2026-06-26
fix: allow quota outbox owner ack after lease expiry

srstack:
## Summary
- Allow quota outbox ack/retry when the worker still owns the receipt, even if the lease deadline has passed.
- Keep recovered/reclaimed rows protected because recovery clears the receipt, so stale owners still fail with lease mismatch.
- Add deterministic datastore regressions for single ack, batch ack, retry, and recovered-row old receipt behavior.
- Add fine-grained backend_write_create_timing fields behind DRIVE9_BENCH_TIMING_LOG_ENABLED, with helper-based timing bookkeeping.
- Remove the upload initiate synchronous quota-outbox drain barrier; upload completion now enqueues an upload_complete tenant quota outbox row so same-file quota mutations converge in per-file order.

## Why
Production server-quota logs showed repeated quota outbox lease mismatch during outbox processing. The lease should authorize recovery by other workers, while the receipt remains the owner token for ack/retry.

The later upload-path finding showed that drainQuotaOutboxForUploadTarget had become a request-path serial barrier: each overwrite upload could wait for the target file outbox to clear, and multiple pods could contend on the same processing row. This PR removes that barrier from initiate and orders upload completion through the same tenant-local quota outbox instead. Older same-file create/overwrite rows now apply before upload_complete without making PUT/initiate requests act as outbox workers.

## Tests
- make test TEST_PKGS='./pkg/backend'
- make test TEST_PKGS='./pkg/backend' TEST_RUN='TestUploadOverwriteQueuesCompleteBehindPendingFileMutation'
- make test TEST_PKGS='./pkg/backend' TEST_RUN='TestUploadOverwriteQueuesCompleteBehindPendingFileMutation|TestServerQuotaUploadOverwriteDeltaIntegration|Test.*QuotaOutbox|TestDrainQuotaOutboxForFile|TestExpiredUploadReservationNoLongerCountsTowardQuota'
- make test TEST_PKGS='./pkg/datastore' TEST_RUN='TestQuotaOutbox|TestHasPendingQuotaOutboxFileMutation'
- GOCACHE=/tmp/drive9-go-cache go test ./pkg/datastore -run TestQuotaOutbox -count=1
- git diff --check

### DIFF
--- a/pkg/backend/dat9.go
+++ b/pkg/backend/dat9.go
@@ -903,18 +903,18 @@ func (b *Dat9Backend) createAndWriteCtx(ctx context.Context, path string, data [
 		quotaOutboxEnqueued = false
 
 		done := backendTimingStep(timingEnabled, &quotaStorageCheckDuration)
-		if err := b.ensureStorageQuota(ctx, tx, path, int64(len(data))); err != nil {
-			done()
+		err := b.ensureStorageQuota(ctx, tx, path, int64(len(data)))
+		done()
+		if err != nil {
 			return err
 		}
-		done()
 
 		done = backendTimingStep(timingEnabled, &quotaFileCountCheckDuration)
-		if err := b.ensureFileCountQuotaServer(ctx, tx, 1); err != nil {
-			done()
+		err = b.ensureFileCountQuotaServer(ctx, tx, 1)
+		done()
+		if err != nil {
 			return err
 		}
-		done()
 
 		fileRev := int64(1)
 		insertFile := &datastore.File{
@@ -931,44 +931,45 @@ func (b *Dat9Backend) createAndWriteCtx(ctx context.Context, path string, data [
 		}
 
 		done = backendTimingStep(timingEnabled, &insertFileDuration)
-		if err := b.store.InsertFileTx(tx, insertFile); err != nil {
-			done()
+		err = b.store.InsertFileTx(tx, insertFile)
+		done()
+		if err != nil {
 			return err
 		}
-		done()
 
 		done = backendTimingStep(timingEnabled, &ensureParentDirsDuration)
-		if err := b.store.EnsureParentDirsTx(tx, path, b.genID); err != nil {
-			done()
+		err = b.store.EnsureParentDirsTx(tx, path, b.genID)
+		done()
+		if err != nil {
 			return err
 		}
-		done()
 
 		done = backendTimingStep(timingEnabled, &insertNodeDuration)
-		if err := b.store.InsertNodeTx(tx, &datastore.FileNode{
+		err = b.store.InsertNodeTx(tx, &datastore.FileNode{
 			NodeID: b.genID(), Path: path, ParentPath: pathutil.ParentPath(path),
 			Name: pathutil.BaseName(path), FileID: fileID, CreatedAt: now,
-		}); err != nil {
-			done()
+		})
+		done()
+		if err != nil {
 			return err
 		}
-		done()
 
 		if tags != nil {
 			done = backendTimingStep(timingEnabled, &tagUpdateDuration)
-			if err := b.store.ReplaceFileTagsTx(tx, fileID, tags); err != nil {
-				done()
+			err = b.store.ReplaceFileTagsTx(tx, fileID, tags)
+			done()
+			if err != nil {
 				return err
 			}
-			done()
 		}
 		currentMediaDelta := quotaMediaDelta(false, isQuotaMediaContentType(contentType))
 		done = backendTimingStep(timingEnabled, &semanticEnqueueDuration)
 		if b.UsesDatabaseAutoEmbedding() {
-			created, err := b.enqueueExtractSemanticTasksTx(ctx, tx, fileID, 1, path, contentType, currentMediaDelta)
+			var created bool
+			created, err = b.enqueueExtractSemanticTasksTx(ctx, tx, fileID, 1, path, contentType, currentMediaDelta)
 			semanticTaskEnqueued = created
+			done()
 			if err != nil {
-				done()
 				return err
 			}
 		} else {
@@ -976,22 +977,19 @@ func (b *Dat9Backend) createAndWriteCtx(ctx context.Context, path string, data [
 			// of EMBED_TEXT, so register them in the same transaction. The embed task
 			// (if any) is enqueued separately below.
 			extractCreated, extractErr := b.enqueueExtractSemanticTasksTx(ctx, tx, fileID, 1, path, contentType, currentMediaDelta)
-			if extractErr != nil {
-				done()
-				return extractErr
-			}
-			if b.shouldEnqueueEmbedForRevision(path, contentType, contentText, description) {
-				created, err := b.enqueueEmbedTaskTx(tx, fileID, 1)
-				semanticTaskEnqueued = created || extractCreated
-				if err != nil {
-					done()
-					return err
-				}
+			err = extractErr
+			if err == nil && b.shouldEnqueueEmbedForRevision(path, contentType, contentText, description) {
+				var embedCreated bool
+				embedCreated, err = b.enqueueEmbedTaskTx(tx, fileID, 1)
+				semanticTaskEnqueued = embedCreated || extractCreated
 			} else {
 				semanticTaskEnqueued = extractCreated
 			}
+			done()
+			if err != nil {
+				return err
+			}
 		}
-		done()
 		if timingEnabled {
 			// Keep the legacy field populated for existing dashboards; create
 			// uses the semantic enqueue phase as its image enqueue boundary.
@@ -1000,11 +998,10 @@ func (b *Dat9Backend) createAndWriteCtx(ctx context.Context, path string, data [
 
 		done = backendTimingStep(timingEnabled, &quotaOutboxEnqueueDuration)
 		created, err := b.enqueueQuotaFileCreateOutboxTx(tx, fileID, int64(len(data)), contentType)
+		done()
 		if err != nil {
-			done()
 			return err
 		}
-		done()
 		quotaOutboxEnqueued = created
 		return nil
 	}); err != nil {
--- a/pkg/backend/dat9.go
+++ b/pkg/backend/dat9.go
@@ -1018,6 +1018,8 @@ func (b *Dat9Backend) createAndWriteCtx(ctx context.Context, path string, data [
 		}
 		if timingEnabled {
 			semanticEnqueueDuration = time.Since(stepStart)
+			// Keep the legacy field populated for existing dashboards; create
+			// uses the semantic enqueue phase as its image enqueue boundary.
 			imageEnqueueDuration = semanticEnqueueDuration
 			stepStart = time.Now()
 		}
--- a/pkg/backend/dat9.go
+++ b/pkg/backend/dat9.go
@@ -768,6 +768,16 @@ func backendDurationMs(d time.Duration) float64 {
 	return float64(d.Microseconds()) / 1000.0
 }
 
+func backendTimingStep(enabled bool, dst *time.Duration) func() {
+	if !enabled {
+		return func() {}
+	}
+	start := time.Now()
+	return func() {
+		*dst = time.Since(start)
+	}
+}
+
 func cloneFileTags(tags map[string]string) map[string]string {
 	if tags == nil {
 		return nil
@@ -891,29 +901,21 @@ func (b *Dat9Backend) createAndWriteCtx(ctx context.Context, path string, data [
 	if err := b.store.InTx(ctx, func(tx *sql.Tx) error {
 		semanticTaskEnqueued = false
 		quotaOutboxEnqueued = false
-		stepStart := time.Time{}
-		if timingEnabled {
-			stepStart = time.Now()
-		}
+
+		done := backendTimingStep(timingEnabled, &quotaStorageCheckDuration)
 		if err := b.ensureStorageQuota(ctx, tx, path, int64(len(data))); err != nil {
-			if timingEnabled {
-				quotaStorageCheckDuration = time.Since(stepStart)
-			}
+			done()
 			return err
 		}
-		if timingEnabled {
-			quotaStorageCheckDuration = time.Since(stepStart)
-			stepStart = time.Now()
-		}
+		done()
+
+		done = backendTimingStep(timingEnabled, &quotaFileCountCheckDuration)
 		if err := b.ensureFileCountQuotaServer(ctx, tx, 1); err != nil {
-			if timingEnabled {
-				quotaFileCountCheckDuration = time.Since(stepStart)
-			}
+			done()
 			return err
 		}
-		if timingEnabled {
-			quotaFileCountCheckDuration = time.Since(stepStart)
-		}
+		done()
+
 		fileRev := int64(1)
 		insertFile := &datastore.File{
 			FileID: fileID, StorageType: storageType, StorageRef: storageRef,
@@ -927,67 +929,46 @@ func (b *Dat9Backend) createAndWriteCtx(ctx context.Context, path string, data [
 		if b.UsesDatabaseAutoEmbedding() && description != "" {
 			insertFile.DescriptionEmbeddingRevision = &fileRev
 		}
-		if timingEnabled {
-			stepStart = time.Now()
-		}
+
+		done = backendTimingStep(timingEnabled, &insertFileDuration)
 		if err := b.store.InsertFileTx(tx, insertFile); err != nil {
-			if timingEnabled {
-				insertFileDuration = time.Since(stepStart)
-			}
+			done()
 			return err
 		}
-		if timingEnabled {
-			insertFileDuration = time.Since(stepStart)
-			stepStart = time.Now()
-		}
+		done()
+
+		done = backendTimingStep(timingEnabled, &ensureParentDirsDuration)
 		if err := b.store.EnsureParentDirsTx(tx, path, b.genID); err != nil {
-			if timingEnabled {
-				ensureParentDirsDuration = time.Since(stepStart)
-			}
+			done()
 			return err
 		}
-		if timingEnabled {
-			ensureParentDirsDuration = time.Since(stepStart)
-			stepStart = time.Now()
-		}
+		done()
+
+		done = backendTimingStep(timingEnabled, &insertNodeDuration)
 		if err := b.store.InsertNodeTx(tx, &datastore.FileNode{
 			NodeID: b.genID(), Path: path, ParentPath: pathutil.ParentPath(path),
 			Name: pathutil.BaseName(path), FileID: fileID, CreatedAt: now,
 		}); err != nil {
-			if timingEnabled {
-				insertNodeDuration = time.Since(stepStart)
-			}
+			done()
 			return err
 		}
-		if timingEnabled {
-			insertNodeDuration = time.Since(stepStart)
-		}
+		done()
+
 		if tags != nil {
-			if timingEnabled {
-				stepStart = time.Now()
-			}
+			done = backendTimingStep(timingEnabled, &tagUpdateDuration)
 			if err := b.store.ReplaceFileTagsTx(tx, fileID, tags); err != nil {
-				if timingEnabled {
-					tagUpdateDuration = time.Since(stepStart)
-				}
+				done()
 				return err
 			}
-			if timingEnabled {
-				tagUpdateDuration = time.Since(stepStart)
-			}
+			done()
 		}
 		currentMediaDelta := quotaMediaDelta(false, isQuotaMediaContentType(contentType))
-		if timingEnabled {
-			stepStart = time.Now()
-		}
+		done = backendTimingStep(timingEnabled, &semanticEnqueueDuration)
 		if b.UsesDatabaseAutoEmbedding() {
 			created, err := b.enqueueExtractSemanticTasksTx(ctx, tx, fileID, 1, path, contentType, currentMediaDelta)
 			semanticTaskEnqueued = created
 			if err != nil {
-				if timingEnabled {
-					semanticEnqueueDuration = time.Since(stepStart)
-					imageEnqueueDuration = semanticEnqueueDuration
-				}
+				done()
 				return err
 			}
 		} else {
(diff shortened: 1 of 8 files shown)


pkg/datastore/quota_outbox.go
```diff
@@ -331,7 +331,7 @@ func (s *Store) retryQuotaOutboxTx(ctx context.Context, tx *sql.Tx, id int64, re
 	if err != nil {
 		return err
 	}
-	if entry.Status != QuotaOutboxProcessing || entry.Receipt != receipt || entry.LeaseUntil == nil || !entry.LeaseUntil.After(now) {
+	if entry.Status != QuotaOutboxProcessing || entry.Receipt != receipt {
```
qiffang, line 338: [claude] The retry path now only checks `entry.Status != QuotaOutboxProcessing || entry.Receipt != receipt`, but unlike ack (which is a SQL UPDATE with receipt matching that naturally returns 0 rows affected on mismatch), retry does a SELECT-then-UPDATE. After removing the lease_until check, a slow worker whose lease expired but whose receipt wasn't cleared could retry the row even though another worker already claimed it via `ClaimQuotaOutboxBatch` (which sets a new receipt). The receipt mismatch would catch that case, so this is actually safe — but only because recovery clears the receipt before re-claim sets a new one. Worth adding a brief comment here explaining why receipt-only is sufficient (recovery clears receipt, so stale owners always fail), since the removed lease_until guard was the only thing making the invariant obvious.

pkg/backend/dat9.go
```diff
@@ -875,12 +891,29 @@ func (b *Dat9Backend) createAndWriteCtx(ctx context.Context, path string, data [
 	if err := b.store.InTx(ctx, func(tx *sql.Tx) error {
```
qiffang, line 901: [claude] The timing instrumentation adds ~80 lines of `if timingEnabled` blocks that substantially reduce readability of the core create path. Consider extracting a small helper like `func timeStep(enabled bool, dur *time.Duration) func() { if !enabled { return func(){} }; s := time.Now(); return func() { *dur = time.Since(s) } }` so each step becomes a one-liner `done := timeStep(timingEnabled, &insertFileDuration); ...; done()`. This would cut the instrumentation noise by ~60% and make the business logic easier to follow.

pkg/backend/dat9.go
```diff
@@ -894,28 +927,67 @@ func (b *Dat9Backend) createAndWriteCtx(ctx context.Context, path string, data [
...
+			stepStart = time.Now()
+		}
 		if err := b.store.InsertNodeTx(tx, &datastore.FileNode{
 			NodeID: b.genID(), Path: path, ParentPath: pathutil.ParentPath(path),
 			Name: pathutil.BaseName(path), FileID: fileID, CreatedAt: now,
 		}); err != nil {
+			if timingEnabled {
+				insertNodeDuration = time.Since(stepStart)
+			}
 			return err
```
qiffang, line 954: [claude] On the error path inside `enqueueExtractSemanticTasksTx`, you set `imageEnqueueDuration = semanticEnqueueDuration`. But `imageEnqueueDuration` is also set on the success path (line ~1010). The aliasing is intentional per the comment ('Keep the legacy field populated'), but on error returns the `defer` logger still fires and logs both `semantic_enqueue_ms` and `image_enqueue_ms` with the same value. This is fine but slightly misleading in error scenarios — the image enqueue never actually ran. Consider only setting `imageEnqueueDuration` on the success path to keep error-path metrics clean.

pkg/datastore/quota_outbox.go
```diff
@@ -242,10 +242,10 @@ func (s *Store) AckQuotaOutboxBatchTx(ctx context.Context, tx *sql.Tx, entries [
 		}
 		ids = append(ids, entries[i].ID)
 	}
-	args := append([]any{QuotaOutboxSucceeded, now, now, QuotaOutboxProcessing, receipt, now}, ids...)
+	args := append([]any{QuotaOutboxSucceeded, now, now, QuotaOutboxProcessing, receipt}, ids...)
```
qiffang, line 247: [claude] Minor: the `append([]any{...}, ids...)` pattern creates a fresh slice each time. Since you removed the `now` argument (6 args → 5 args), double-check that the SQL placeholder count still matches. It does look correct (5 fixed args + len(entries) placeholders), but this is the kind of change where an off-by-one in placeholder count silently produces wrong results at runtime. The tests should catch it, but worth a manual count.

pkg/backend/dat9.go
```diff
@@ -875,12 +891,29 @@ func (b *Dat9Backend) createAndWriteCtx(ctx context.Context, path string, data [
 	if err := b.store.InTx(ctx, func(tx *sql.Tx) error {
 		semanticTaskEnqueued = false
 		quotaOutboxEnqueued = false
+		stepStart := time.Time{}
```
qiffang: [claude] `stepStart := time.Time{}` is initialized even when `timingEnabled` is false. This is harmless but inconsistent with the pattern used elsewhere in this function where duration variables are declared as zero-value and only touched inside `if timingEnabled` guards. Consider moving the declaration inside the first `if timingEnabled` block, or at least using `var stepStart time.Time` for consistency.

qiffang (changes requested):
## Automated Code Review

**Claude Code**: The core fix is sound: relaxing the lease_until check so the receipt owner can ack/retry after lease expiry, while recovery still protects against stale owners by clearing the receipt. The test coverage is thorough and correctly validates both the happy path (expired-but-owned) and the rejection path (recovered with new receipt). However, the timing instrumentation in dat9.go is excessively verbose and has a subtle correctness issue around the imageEnqueueDuration aliasing, and there is a minor concern about the retry predicate relaxation.

**Codex**: The ownership change is consistent with the stated lease model: ack and retry remain constrained by processing status plus receipt ownership, while recovery/reclaim clears the receipt so stale owners still fail. The added timing fields are gated behind the existing benchmark logging flag and do not materially affect the default hot path. I did not find blocking correctness, security, performance, or design issues in the provided diff.

mornyx:
## Review

Overall the changes look correct and well-tested. The core fix — allowing ack/retry of a quota outbox entry whose lease has expired but whose receipt is still owned by the caller — is sound. The 7 new tests cover the key scenarios well.

### Observations

1. **backendTimingStep pattern**: The new `backendTimingStep` helper is clean, but the call sites are verbose with the `done(); done = backendTimingStep(...)` repetition. Consider a small helper that wraps a func and records timing, e.g. `timedStep(&dur, func() error { return b.store.InsertFileTx(tx, insertFile) })` — but this is a style suggestion, not blocking.

2. **imageEnqueueDuration aliasing**: At line ~996 you set `imageEnqueueDuration = semanticEnqueueDuration` when timing is enabled. This keeps legacy dashboards working, which is good. Worth a comment noting that this is an alias for backward compat (you already have one, good).

3. **TestUploadOverwriteQueuesCompleteBehindPendingFileMutation**: This is a comprehensive integration test. The explicit S3 part upload + confirm flow is good. The test verifies the full lifecycle: file create → upload initiate (reservation) → part upload → confirm (enqueue complete) → outbox drain → verify usage. Nice coverage.

4. **Test name**: `TestDrainQuotaOutboxForUploadTargetProcessesClaimableBatch` was replaced by `TestUploadOverwriteQueuesCompleteBehindPendingFileMutation` — the old test name was more descriptive of the drain mechanism, while the new one focuses on the upload lifecycle. Both are fine, just noting the scope change.

### Minor
- The `processQuotaOutboxAvailable` call added to `TestServerQuotaUploadOverwriteDeltaIntegration` (line 183) is a test-only fix — ensure the production code path doesn't need the same explicit drain call.

LGTM from a correctness standpoint.

srstack:
@mornyx thanks for the review.

On the `processQuotaOutboxAvailable` point: the explicit call in `TestServerQuotaUploadOverwriteDeltaIntegration` is test-only so the assertion can read central quota usage deterministically right after `ConfirmUpload`. Production does not need that synchronous drain:

- server-quota backends start the background outbox worker from `setCentralQuotaManager` / `startQuotaOutboxWorker`
- writes/uploads call `notifyQuotaOutbox(true)` after enqueueing a tenant-local quota outbox row
- the worker drains on notification and also on the periodic poll path

So the upload hot path no longer blocks on a per-file drain; completion is ordered through the tenant-local outbox and processed asynchronously.

qiffang (changes requested):
## Review — adversary-2

**判定：REVISE** — 1 blocker

### B1: `upload_complete` mutation type has no visible process handler

PR adds `quotaMutationTypeUploadComplete = "upload_complete"` and `enqueueQuotaUploadCompleteOutboxTx`, and `HasPendingQuotaOutboxFileMutation` now includes `upload_complete`. But the diff does not contain a handler for `upload_complete` in `processQuotaOutboxEntry` (or equivalent).

If the existing `processQuotaOutboxEntry` doesn't recognize `upload_complete`:
- Default error → row stuck in processing forever, reservation never released
- Default silent ack → reservation released but quota usage not updated

**Please confirm that `processQuotaOutboxEntry` already handles `upload_complete` in code not shown in this diff. If not, this is a production-blocking bug — enqueued rows will never be processed.**

- **O1**: Three independent changes bundled (lease expiry fix, upload drain removal + upload_complete outbox, timing instrumentation). Any rollback takes all three.
- **O2**: `enqueueUploadCompleteOutbox` and `enqueueUploadSemanticTasks` closures mutate outer variables via side effects. Not incorrect but increases cognitive load in an already complex `finalizeUpload`.

### Tests ✓

Comprehensive coverage for lease ownership semantics (6 deterministic datastore tests) and upload_complete ordering (1 integration test).

pkg/backend/upload.go
```diff
@@ -1112,103 +1139,61 @@ func (b *Dat9Backend) finalizeUpload(ctx context.Context, upload *datastore.Uplo
...
-		if b.shouldEnqueueEmbedForRevision(upload.TargetPath, contentType, "", upload.Description) {
-			created, err := b.enqueueEmbedTaskTx(tx, confirmedFileID, confirmedRevision)
-			semanticTaskEnqueued = created || extractCreated
-			semanticEnqueueDurationMs = uploadPhaseMs(stepStart)
-			return err
-		}
-		semanticTaskEnqueued = extractCreated
-		semanticEnqueueDurationMs = uploadPhaseMs(stepStart)
-		return nil
+		return enqueueUploadCompleteOutbox(tx)
```
qiffang, line 1208: This is not retry-safe as an outbox mutation yet. `ProcessQuotaOutboxBatch`/`processQuotaOutboxEntry` commit the central apply before the tenant-local ack; if the process crashes after the central tx, or the tenant ack fails, recovery can re-claim this same row and call `applyCentralQuotaMutationTx(..., upload_complete, ...)` again. On the first apply, `applyUploadCompleteTx` settles the active reservation and transfers reserved -> storage. On the second apply, the reservation is already terminal, so `SettleActiveReservationTx` returns `settled=false` and the helper charges `NewSizeBytes` and subtracts `OldSizeBytes` again. That corrupts central quota (create doubles storage; overwrite subtracts the old size twice). The old inline path did not have this split-brain retry window because apply + `MarkMutationAppliedTx` lived in one central tx. Before routing upload completion through the tenant outbox, make `upload_complete` idempotent under duplicate apply and add a regression that simulates central apply success followed by tenant ack failure/recovery/retry.

qiffang (changes requested):
Blocking on one correctness issue in the new upload-complete outbox path. The lease/receipt datastore changes look directionally correct and CI is green, but moving `upload_complete` behind the tenant-local outbox creates a retry window that the current central apply function is not idempotent for.

Validation: GitHub `code-ci` and `local-e2e` are green on `38d18378f9e2369f06fc6b4d3a4d6cf3f46c493b`; `git diff --check origin/main...HEAD` passes locally. I could not run Go tests on this machine because `go` is not installed.

qiffang (commented):
## PR #625 Review — fix: allow quota outbox owner ack after lease expiry

### Summary

This PR makes three related changes:

1. **Lease vs Receipt ownership fix** — `ackQuotaOutbox`, `AckQuotaOutboxBatchTx`, and `retryQuotaOutboxTx` no longer require `lease_until > now`. Ownership is determined solely by `receipt` match. Recovery clears the old receipt before requeueing, so stale owners cannot ack after recovery. This is the correct semantic: a lease authorizes *recovery by other workers*, while a receipt is the *owner token* for ack/retry.

2. **Upload completion via outbox** — `finalizeUpload` enqueues an `upload_complete` mutation into the tenant-local quota outbox instead of calling `completeUploadReservation` synchronously. This orders upload completion behind any pending `file_create`/`file_overwrite` mutations for the same file, preventing out-of-order quota accounting.

3. **Remove synchronous drain at upload initiate** — `drainQuotaOutboxForUploadTarget` is removed. Upload initiate no longer blocks on flushing the outbox, reducing latency on the initiate path.

### Assessment

The core ownership fix (receipt-only ack) is sound and well-tested with 6 new datastore tests covering: expired-but-owned ack, recovered-row-old-receipt rejection, and batch variants of both. The refactoring of `finalizeUpload` with extracted closures reduces duplication nicely. Fine-grained timing instrumentation is a welcome addition for diagnosing slow writes.

See inline comments for specific observations.

pkg/datastore/quota_outbox.go
qiffang, line 247: **Correctness — receipt-only guard is sound but implicit on recovery semantics.**

The safety of removing `lease_until > ?` depends on `RecoverExpiredQuotaOutbox` atomically clearing `receipt` before requeueing. If recovery ever leaves the old receipt intact (bug, schema migration, etc.), the old owner could ack a row that another worker is about to claim.

Consider adding a brief comment on this invariant near the `WHERE` clause, e.g.:
```sql
-- Receipt-only: safe because RecoverExpiredQuotaOutbox clears receipt before requeueing.
```
This makes the contract grep-discoverable.

pkg/backend/upload.go
```diff
@@ -1024,8 +1015,44 @@ func (b *Dat9Backend) finalizeUpload(ctx context.Context, upload *datastore.Uplo
 	var insertNodeDurationMs float64
 	var semanticEnqueueDurationMs float64
 	var semanticTaskEnqueued bool
+	var quotaOutboxEnqueued bool
+	enqueueUploadCompleteOutbox := func(tx *sql.Tx) error {
```
qiffang, line 1022: **Correctness — `quotaOutboxEnqueued` closed over but reset inside `InTx` retry.**

`quotaOutboxEnqueued` is captured by the `enqueueUploadCompleteOutbox` closure and reset to `false` at the top of the `InTx` callback. This is correct for retry, but the closure also captures `upload.TotalSize` and several `old*` variables that are set *before* the transaction. If the transaction retries and any of those values could have changed (e.g. `oldSizeBytes` from a concurrent overwrite), the enqueued mutation data could be stale.

In practice SQLite serializes transactions so this is safe today, but worth a comment if this ever moves to a non-serializable backend.

qiffang, line 1022: **Correctness — `quotaOutboxEnqueued` closed over but reset inside `InTx` retry.**

`quotaOutboxEnqueued` is captured by the `enqueueUploadCompleteOutbox` closure and reset to `false` at the top of the `InTx` callback. This is correct for retry, but the closure also captures `upload.TotalSize` and several `old*` variables set *before* the transaction. If the transaction retries and any of those values could have changed, the enqueued mutation data could be stale.

In practice SQLite serializes transactions so this is safe today, but worth a comment if this ever moves to a non-serializable backend.

pkg/backend/upload.go
```diff
@@ -1219,10 +1204,14 @@ func (b *Dat9Backend) finalizeUpload(ctx context.Context, upload *datastore.Uplo
 	}
 	txDurationMs := uploadPhaseMs(txStart)
 	b.notifySemanticTaskEnqueued(semanticTaskEnqueued)
+	b.notifyQuotaOutbox(quotaOutboxEnqueued)
 
-	// Transfer server-side reservation, update file shadow state, and mark the
-	// upload-complete mutation applied in one server-DB transaction.
-	b.completeUploadReservation(ctx, uploadID, upload.TotalSize, confirmedFileID, oldSizeBytes, oldIsMedia, upload.TotalSize, newIsMedia)
+	if !quotaOutboxEnqueued {
+		// Preserve the non-outbox path when central quota is wired without
```
qiffang, line 1228: **Design — fallback path when `!quotaOutboxEnqueued`.**

When `UseServerQuota()` is false, `enqueueQuotaUploadCompleteOutboxTx` returns `(false, nil)` and the code falls through to `completeUploadReservation`. This preserves the non-outbox central-quota path.

However, if `enqueueQuotaUploadCompleteOutboxTx` returns `(false, nil)` for a *different* reason (e.g. a future code change accidentally short-circuits), the upload silently skips quota accounting entirely — no outbox entry and no direct reservation call.

Consider asserting the invariant:
```go
if !quotaOutboxEnqueued && b.UseServerQuota() {
    logger.Error(ctx, "upload_complete_quota_accounting_skipped", ...)
}
```

srstack:
@qiffang confirmed: `upload_complete` is handled, but the dispatch was easy to miss because tenant-local quota outbox entries reuse the central mutation dispatcher.

Path is:

- `ProcessQuotaOutboxBatch` / `processQuotaOutboxEntry`
- `applyQuotaOutboxEntryTx`
- `applyCentralQuotaMutationTx(...)`
- `case "upload_complete"` in `pkg/backend/mutation_replay.go`
- `applyUploadCompleteTx(...)`, shared with `completeUploadReservation`

So unknown/default handling is not involved for `upload_complete`; reservation settlement and quota/file-meta deltas go through the same tx-scoped helper as mutation replay.

I pushed `92b4dfe0` to make this dispatch explicit at the quota outbox apply entry point. Coverage is also already in `TestUploadOverwriteQueuesCompleteBehindPendingFileMutation` and `TestServerQuotaUploadOverwriteDeltaIntegration`, both of which process the queued `upload_complete` outbox row and verify central usage/reservation state.

pkg/backend/quota_outbox_worker.go
```diff
@@ -457,6 +448,34 @@ func (b *Dat9Backend) enqueueQuotaFileCreateOutboxTx(tx *sql.Tx, fileID string,
 	return true, nil
 }
 
+func (b *Dat9Backend) enqueueQuotaUploadCompleteOutboxTx(tx *sql.Tx, uploadID string, reservedBytes int64, fileID string, oldSizeBytes int64, oldIsMedia bool, newSizeBytes int64, newIsMedia bool) (bool, error) {
```
qiffang, line 453: **Question — where is `uploadCompleteMutationData` struct defined and where is it dispatched?**

The enqueue function marshals `uploadCompleteMutationData`, but the worker dispatch delegates to `applyCentralQuotaMutationTx(b.metaStore, tx, tenantID, entry.MutationType, entry.MutationData, entry.ID)`. This routes all mutation types (including `upload_complete`) through the same meta-store path.

Is the meta-store `ApplyQuotaMutation` handler already wired to handle `upload_complete`? The diff does not show changes to the meta package. If the handler is already present on `main`, this is fine — but if it is missing, the outbox worker will fail on every `upload_complete` entry.

pkg/backend/dat9.go
```diff
@@ -875,12 +901,21 @@ func (b *Dat9Backend) createAndWriteCtx(ctx context.Context, path string, data [
 	if err := b.store.InTx(ctx, func(tx *sql.Tx) error {
```
qiffang, line 901: **Nit — `done()` called twice on error path.**

The `backendTimingStep` closure is safe to call multiple times (idempotent `time.Since`), but the pattern is verbose and easy to get wrong. Consider restructuring so `done()` is called exactly once:

```go
done := backendTimingStep(timingEnabled, &quotaStorageCheckDuration)
err := b.ensureStorageQuota(ctx, tx, path, int64(len(data)))
done()
if err != nil { return err }
```

This would halve the `done()` calls and eliminate the error-path duplication across all 8 instrumented steps.

pkg/backend/quota_outbox_test.go
```diff
@@ -349,41 +352,86 @@ func TestDrainQuotaOutboxForFileErrorsWhenPendingRowIsNotClaimable(t *testing.T)
 	}
 }
 
-func TestDrainQuotaOutboxForUploadTargetProcessesClaimableBatch(t *testing.T) {
-	b, fake := newServerQuotaBackend(t, Options{})
+func TestUploadOverwriteQueuesCompleteBehindPendingFileMutation(t *testing.T) {
```
qiffang, line 355: **Good test — covers the full upload-overwrite-via-outbox lifecycle.**

This test verifies:
1. `file_create` outbox entry exists after initial write
2. `InitiateUpload` reserves but does not drain outbox
3. `ConfirmUpload` enqueues `upload_complete` alongside pending `file_create`
4. `processQuotaOutboxAvailable` applies both in order with correct final usage

One suggestion: also assert that after `processQuotaOutboxAvailable`, there are zero `queued` or `processing` rows remaining, to confirm nothing leaked.

pkg/backend/upload.go
srstack: Fixed in 325c9f61. `applyUploadCompleteTx` now treats `settled=false` plus existing `file_meta` already matching the post-upload state as a duplicate retry no-op, so a row re-applied after central commit + tenant ack loss no longer charges `NewSizeBytes`/`OldSizeBytes` again. Added `TestUploadCompleteOutboxRetryAfterCentralApplyDoesNotDoubleCharge`, which simulates central apply success, expired-lease recovery, re-claim, retry apply, and tenant ack.

pkg/datastore/quota_outbox.go
srstack: Fixed in 325c9f61. Added the receipt-ownership invariant comment next to `AckQuotaOutboxBatchTx`: recovery clears receipt before requeueing, so stale owners cannot match after another worker reclaims the row.

pkg/backend/upload.go
srstack: Fixed in a502b8bb. Same as the duplicate thread: each `InTx` attempt now clears the captured upload quota state first, then refreshes old state under lock before enqueueing, so retry cannot reuse stale values from a prior attempt.

pkg/backend/dat9.go
srstack: Fixed in 325c9f61. The create timing steps now follow the suggested pattern: run the step, call `done()` exactly once, then check and return the error. That removes the duplicated success/error `done()` calls while keeping error-path timing.

qiffang (approved):
## Re-review — adversary-2

**判定：GREEN**

Commit `325c9f6` makes `upload_complete` idempotent:

`applyUploadCompleteTx` now detects duplicate retry: when `!settled` (no active reservation) AND `oldMeta.SizeBytes == data.NewSizeBytes` AND `oldMeta.IsMedia == data.NewIsMedia`, the apply is a no-op. This prevents the double-charge on crash-after-central-apply + recovery-retry.

Regression test `TestUploadCompleteOutboxRetryAfterCentralApplyDoesNotDoubleCharge` covers the exact scenario: first apply → recover expired → reclaim → retry → verify usage unchanged. ✓

Commit `a502b8b` resets upload quota state on tx retry — prevents stale closure state from leaking across InTx retries. ✓

Idempotency blocker resolved.

qiffang (commented):
## PR #625 Re-review — updated commits address all previous feedback

All issues from the first review round are resolved:

### Previous blockers — fixed
1. **Idempotency** (@adversary-1 blocker) — `applyUploadCompleteTx` now short-circuits when `!settled && oldMeta matches post-upload state`, preventing double-charge on retry. **Fixed in `325c9f61`.**
2. **Retry regression test** — `TestUploadCompleteOutboxRetryAfterCentralApplyDoesNotDoubleCharge` covers: claim → apply → crash → recover → reclaim → retry → verify usage unchanged. **Added in `325c9f61`.**

### Previous suggestions — adopted
3. **`done()` pattern simplified** — All 8 timing steps now use `err := ...; done(); if err != nil { return err }`. **Fixed in `92b4dfe0` or `325c9f61`.**
4. **Closure capture documented** — Comment explains `old*` fields are refreshed inside `InTx`. State reset added at top of tx callback (`a502b8bb`).
5. **Defensive log** — `upload_complete_quota_accounting_skipped` error log when `!quotaOutboxEnqueued && UseServerQuota()`.
6. **SQL receipt comment** — Added to `AckQuotaOutboxBatchTx`.
7. **Dispatch comment** — Documents `upload_complete` → `applyUploadCompleteTx` routing.
8. **Zero-residue assertion** — `countPendingQuotaOutboxRowsByFile` check after drain.

### Remaining observation
The idempotency guard checks `oldMeta.SizeBytes == data.NewSizeBytes && oldMeta.IsMedia == data.NewIsMedia`. This is a heuristic — if two different uploads produce the same size and media type, a genuine retry of the first could be silently skipped after the second commits. In practice this is unlikely (requires same file, same size, same type, crash between central apply and tenant ack, followed by a second upload completing before recovery). Acceptable risk.

**No remaining blockers from my side.**

qiffang (approved):
Reviewed latest head `a502b8b`.

What I checked:
- Quota outbox ack/retry is now receipt-owned rather than lease-deadline-owned; recovery clears `receipt`, so stale workers fail after reclaim while the original owner can still ack/retry after deadline if no recovery happened.
- Batch ack uses one claim receipt for the whole batch and rejects mixed/stale receipts.
- Upload initiate no longer drains same-file quota outbox on the request path; `upload_complete` is queued behind prior same-file mutations and uses the shared central mutation dispatcher.
- Upload-complete replay after central apply is idempotent via current `file_meta` matching, so recovered/reclaimed rows do not double-charge storage/file count.
- `finalizeUpload` resets captured quota state on every tenant tx attempt before enqueueing the outbox row.

Validation:
- `go test -c ./pkg/backend ./pkg/datastore` passed locally.
- Focused runtime tests are blocked on this machine by local testcontainers (`rootless Docker not found`). GitHub `code-ci` and `local-e2e` are both green on this head.

No blocking findings from my pass.

PR #626, opened 2026-06-26
feat: support private endpoint for tidb_cloud_native tenants

srstack:
## Summary

Add support for using TiDB Cloud private endpoints instead of public endpoints when provisioning native tenants.

## Usage

Set `DRIVE9_TIDBCLOUD_NATIVE_USE_PRIVATE_ENDPOINT=true` to use private endpoints.

## Changes

- Parse `endpoints.private.host` and `endpoints.private.port` from TiDB Cloud API response
- `Provisioner.usePrivateEndpoint` field driven by env var
- `clusterConnectionIncomplete` checks only the relevant endpoint type
- Existing tenants unaffected — endpoint is stored in tenant record at create time

## Summary by CodeRabbit

* **New Features**
  * Added support for provisioning TiDB Cloud tenants and branches using either public or private connectivity, controlled by an environment flag for private endpoint mode.
  * Endpoint selection and readiness gating now follow the chosen connectivity type.

* **Bug Fixes**
  * Connection details (host/port) are now filled from the correct public vs. private endpoint.
  * Tenant/branch username resolution now consistently derives from the prefix-based naming.

* **Tests**
  * Added coverage for the private endpoint flag parsing and private-vs-public completeness and endpoint filling behavior.

### DIFF
--- a/pkg/tenant/tidbcloudnative/provisioner.go
+++ b/pkg/tenant/tidbcloudnative/provisioner.go
@@ -978,6 +978,7 @@ type clusterInfo struct {
 	} `json:"endpoints"`
 	UserPrefix string `json:"userPrefix"`
 }
+
 type branchInfo struct {
 	BranchID  string `json:"branchId"`
 	State     string `json:"state"`
@@ -992,7 +993,6 @@ type branchInfo struct {
 		} `json:"private"`
 	} `json:"endpoints"`
 	UserPrefix string `json:"userPrefix"`
-	Username   string `json:"username"`
 }
 
 type clusterListResponse struct {
--- a/pkg/tenant/tidbcloudnative/provisioner.go
+++ b/pkg/tenant/tidbcloudnative/provisioner.go
@@ -264,7 +264,7 @@ func (p *Provisioner) ProvisionWithCredentialsAndQuota(ctx context.Context, tena
 	if info.ClusterID == "" {
 		return nil, nil, fmt.Errorf("tidbcloud native response missing cluster id")
 	}
-	if clusterProvisionMetadataIncomplete(info) {
+	if clusterProvisionMetadataIncomplete(info, p.usePrivateEndpoint) {
 		info, err = p.waitForClusterProvisionMetadata(ctx, publicKey, privateKey, info.ClusterID)
 		if err != nil {
 			return &tenant.ClusterInfo{
@@ -279,10 +279,8 @@ func (p *Provisioner) ProvisionWithCredentialsAndQuota(ctx context.Context, tena
 	host := info.Endpoints.Public.Host
 	port := info.Endpoints.Public.Port
 	if p.usePrivateEndpoint {
-		if info.Endpoints.Private.Host != "" {
-			host = info.Endpoints.Private.Host
-			port = info.Endpoints.Private.Port
-		}
+		host = info.Endpoints.Private.Host
+		port = info.Endpoints.Private.Port
 	}
 	out := &tenant.ClusterInfo{
 		TenantID:       tenantID,
@@ -407,8 +405,8 @@ func (p *Provisioner) CreateBranchWithCredentials(ctx context.Context, forkTenan
 		DBName:    dbName,
 		Provider:  tenant.ProviderTiDBCloudNative,
 	}
-	if !branchConnectionIncomplete(branch) {
-		if err := fillBranchEndpoint(out, branch); err != nil {
+	if !branchConnectionIncomplete(branch, p.usePrivateEndpoint) {
+		if err := p.fillBranchEndpoint(out, branch); err != nil {
 			return out, err
 		}
 		return out, nil
@@ -444,7 +442,7 @@ func (p *Provisioner) WaitForBranchActiveWithCredentials(ctx context.Context, br
 	if err != nil {
 		return &out, fmt.Errorf("wait for branch active: %w", err)
 	}
-	if err := fillBranchEndpoint(&out, info); err != nil {
+	if err := p.fillBranchEndpoint(&out, info); err != nil {
 		return &out, err
 	}
 	return &out, nil
@@ -994,6 +992,10 @@ type branchInfo struct {
 			Host string `json:"host"`
 			Port int    `json:"port"`
 		} `json:"public"`
+		Private struct {
+			Host string `json:"host"`
+			Port int    `json:"port"`
+		} `json:"private"`
 	} `json:"endpoints"`
 	UserPrefix string `json:"userPrefix"`
 	Username   string `json:"username"`
@@ -1044,32 +1046,41 @@ func parseBranchInfo(raw []byte) (*branchInfo, error) {
 	return &out, nil
 }
 
-func clusterConnectionIncomplete(info *clusterInfo) bool {
+func clusterConnectionIncomplete(info *clusterInfo, usePrivate bool) bool {
 	if info == nil {
 		return true
 	}
-	if parseBoolEnv(EnvTiDBCloudNativeUsePrivateEndpoint) {
+	if usePrivate {
 		return info.Endpoints.Private.Host == "" || info.Endpoints.Private.Port == 0 || (info.UserPrefix == "" && info.Username == "")
 	}
 	return info.Endpoints.Public.Host == "" || info.Endpoints.Public.Port == 0 || (info.UserPrefix == "" && info.Username == "")
 }
 
-func clusterProvisionMetadataIncomplete(info *clusterInfo) bool {
-	if clusterConnectionIncomplete(info) {
+func clusterProvisionMetadataIncomplete(info *clusterInfo, usePrivate bool) bool {
+	if clusterConnectionIncomplete(info, usePrivate) {
 		return true
 	}
 	return strings.TrimSpace(info.Labels[TiDBCloudOrganizationLabel]) == ""
 }
 
-func branchConnectionIncomplete(info *branchInfo) bool {
+func branchConnectionIncomplete(info *branchInfo, usePrivate bool) bool {
 	if info == nil {
 		return true
+	}
+	if usePrivate {
+		return info.Endpoints.Private.Host == "" || info.Endpoints.Private.Port == 0 || (info.UserPrefix == "" && info.Username == "")
 	}
 	return info.Endpoints.Public.Host == "" || info.Endpoints.Public.Port == 0 || (info.UserPrefix == "" && info.Username == "")
 }
 
-func fillBranchEndpoint(out *tenant.ClusterInfo, branch *branchInfo) error {
-	if branch.Endpoints.Public.Host == "" || branch.Endpoints.Public.Port == 0 {
+func (p *Provisioner) fillBranchEndpoint(out *tenant.ClusterInfo, branch *branchInfo) error {
+	host := branch.Endpoints.Public.Host
+	port := branch.Endpoints.Public.Port
+	if p.usePrivateEndpoint {
+		host = branch.Endpoints.Private.Host
+		port = branch.Endpoints.Private.Port
+	}
+	if host == "" || port == 0 {
 		return fmt.Errorf("tidbcloud native branch response missing endpoint")
 	}
 	if branch.Username != "" {
@@ -1080,8 +1091,8 @@ func fillBranchEndpoint(out *tenant.ClusterInfo, branch *branchInfo) error {
 	if out.Username == "" {
 		return fmt.Errorf("tidbcloud native branch response missing username")
 	}
-	out.Host = branch.Endpoints.Public.Host
-	out.Port = branch.Endpoints.Public.Port
+	out.Host = host
+	out.Port = port
 	return nil
 }
 
@@ -1109,7 +1120,7 @@ func (p *Provisioner) waitForClusterProvisionMetadata(ctx context.Context, publi
 		if err != nil {
 			return nil, err
 		}
-		if !clusterProvisionMetadataIncomplete(info) {
+		if !clusterProvisionMetadataIncomplete(info, p.usePrivateEndpoint) {
 			return info, nil
 		}
 		if time.Now().After(deadline) {
@@ -1147,7 +1158,7 @@ func (p *Provisioner) waitForBranchActive(ctx context.Context, publicKey, privat
 		if err != nil {
 			return nil, err
 		}
-		if info.State == stateActive && !branchConnectionIncomplete(info) {
+		if info.State == stateActive && !branchConnectionIncomplete(info, p.usePrivateEndpoint) {
 			return info, nil
 		}
 		if time.Now().After(deadline) {
--- a/pkg/tenant/tidbcloudnative/provisioner.go
+++ b/pkg/tenant/tidbcloudnative/provisioner.go
@@ -300,7 +300,11 @@ func (p *Provisioner) ProvisionWithCredentialsAndQuota(ctx context.Context, tena
 		out.Username = info.UserPrefix + ".root"
 	}
 	if out.Host == "" || out.Port == 0 {
-		return out, nil, fmt.Errorf("tidbcloud native response missing endpoint")
+		epKind := "public"
+		if p.usePrivateEndpoint {
+			epKind = "private"
+		}
+		return out, nil, fmt.Errorf("tidbcloud native response missing %s endpoint", epKind)
 	}
 	if out.Username == "" {
 		return out, nil, fmt.Errorf("tidbcloud native response missing username")
@@ -1087,7 +1091,11 @@ func (p *Provisioner) fillBranchEndpoint(out *tenant.ClusterInfo, branch *branch
 		port = branch.Endpoints.Public.Port
 	}
 	if host == "" || port == 0 {
-		return fmt.Errorf("tidbcloud native branch response missing endpoint")
+		epKind := "public"
+		if p.usePrivateEndpoint {
+			epKind = "private"
+		}
+		return fmt.Errorf("tidbcloud native branch response missing %s endpoint", epKind)
 	}
 	if branch.Username != "" {
 		out.Username = branch.Username
--- a/pkg/tenant/tidbcloudnative/provisioner.go
+++ b/pkg/tenant/tidbcloudnative/provisioner.go
@@ -276,11 +276,14 @@ func (p *Provisioner) ProvisionWithCredentialsAndQuota(ctx context.Context, tena
 			}, nil, err
 		}
 	}
-	host := info.Endpoints.Public.Host
-	port := info.Endpoints.Public.Port
+	var host string
+	var port int
 	if p.usePrivateEndpoint {
 		host = info.Endpoints.Private.Host
 		port = info.Endpoints.Private.Port
+	} else {
+		host = info.Endpoints.Public.Host
+		port = info.Endpoints.Public.Port
 	}
 	out := &tenant.ClusterInfo{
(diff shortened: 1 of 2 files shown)


srstack:
**1. Username 删除** — 不是行为变更。API 实际返回数据显示 `Username` 字段从未被 populate（所有 native 集群都只用 `userPrefix + ".root"`）。旧代码的 `info.Username != ""` 分支是死代码从未触发过。单独 commit message 已注明此清理。

**2. private endpoint 就绪时机** — polling 循环和 public 完全一致（10 分钟 deadline）。private endpoint 就绪前 `clusterConnectionIncomplete` 返回 true 继续 poll，和 public 逻辑没有区别。如果集群不支持 private endpoint 则 poll 超时返回 error，符合预期。

**3. 测试 gap** — 已补：
- `TestFillBranchEndpointUsesPrivateEndpoint` — 验证 branch 选择 private host/port
- 测试中 public/private port 改为 4000/4001 区分

mornyx:
## Review

Clean implementation — adding `usePrivateEndpoint` to the provisioner and threading it through cluster/branch connection checks and endpoint selection is straightforward and correct.

### Positive
- Good test coverage: env parsing, cluster connection completeness (public vs private), provision with private endpoint, branch connection, fillBranchEndpoint. Each scenario is well isolated.
- The `parseBoolEnv` helper is simple and consistent with existing patterns.
- Correctly removed the `Username` field from `clusterInfo` and `branchInfo` structs — if TiDB Cloud API no longer returns a direct `username` field, relying on `userPrefix + ".root"` is the right approach.
- The wait loops properly check private endpoint completeness when in private mode.

### Observations / Questions

1. **Removed validation**: The old code had `if out.Host == "" || out.Port == 0 { return error }` and `if out.Username == "" { return error }` after setting up `out`. These validations are now gone — the incomplete check runs before, but if the API returns a response that passes the completeness check but has empty host/port for some reason, the error won't be caught. Is this intentional? The completeness check should cover it, but the explicit validation was a good safety net.

2. **fillBranchEndpoint returns nil on empty host/port?**: With the validation removed from `fillBranchEndpoint`, if `host == "" && port == 0` (e.g. private endpoint not ready but `branchConnectionIncomplete` returned false due to a race), the function silently sets empty host/port on `out` and returns nil. Consider adding a final `if host == "" || port == 0 { return fmt.Errorf("...") }` guard.

3. **TestWaitForBranchUserWithCredentialsUsesUserPrefix**: The old test (`PrefersUsername`) tested that a direct `username` field was preferred over `userPrefix`. Now that `username` is removed from the struct, the test correctly verifies `userPrefix` → `u2.root`. Good.

4. **No test for missing private endpoint in provision path**: `TestProvisionWithCredentialsUsesPrivateEndpoint` tests the happy path. Consider adding a test where `usePrivateEndpoint=true` but the API only returns public endpoints — this should trigger the `clusterProvisionMetadataIncomplete` wait loop (which it does via `clusterConnectionIncomplete`). The test `TestClusterConnectionIncompleteWhenPrivateEndpointMissing` covers the helper, but not the full provision flow.

### Minor
- Struct alignment: `usePrivateEndpoint` field is not aligned with other fields in the struct (extra space). `gofmt` should fix this.

LGTM — the core logic is correct and well-tested.

qiffang:
## dev-2 review (focused, non-overlapping with @adversary-2's correctness lane)

Read all 8 commits squashed against `5586b78^..pr-626`. Compiled and ran:
```
go test ./pkg/tenant/tidbcloudnative/ -count=1 -timeout=60s
ok  github.com/mem9-ai/drive9/pkg/tenant/tidbcloudnative  11.404s
```

Three findings. **None block merge** individually; combined I'd label this PR APPROVE-WITH-FOLLOWUPS.

### F1 — `parseBoolEnv` silently maps malformed values to `false` (security ops risk)

```go
func parseBoolEnv(name string) bool {
    v := strings.TrimSpace(strings.ToLower(os.Getenv(name)))
    return v == "1" || v == "true" || v == "yes"
}
```

This env governs a **public-vs-private network endpoint boundary**, i.e. whether a freshly-provisioned cluster is reachable from outside the VPC. If ops sets:

- `DRIVE9_TIDBCLOUD_NATIVE_USE_PRIVATE_ENDPOINT=on`
- `DRIVE9_TIDBCLOUD_NATIVE_USE_PRIVATE_ENDPOINT=Y`
- `DRIVE9_TIDBCLOUD_NATIVE_USE_PRIVATE_ENDPOINT=ture` (typo)
- `DRIVE9_TIDBCLOUD_NATIVE_USE_PRIVATE_ENDPOINT=2`
- `DRIVE9_TIDBCLOUD_NATIVE_USE_PRIVATE_ENDPOINT=enable`

…the function silently returns `false` and the tenant gets a PUBLIC endpoint when ops thought they were configuring PRIVATE. The tenant record is then permanent (per PR description "endpoint is stored in tenant record at create time") — silent misconfig is much harder to detect than a startup error.

Three options, in order of conservativeness:

1. **Strict**: log a warning and refuse to start the provisioner when the env is non-empty AND not in `{1,true,yes,0,false,no}`. Forces ops to fix typos.
2. **Loud-loose**: log a warning at startup whenever the env is non-empty and falls back to default, so ops can grep for it.
3. **Document loudly**: README + comment on `parseBoolEnv` enumerates exactly the truthy values and warns that anything else silently means false. Worst option for ops — production typos go undetected.

I'd recommend option 1 because the env-bool here drives a security boundary. The kb-ai project banked the equivalent rule yesterday for `_env_float` (silent fallback on `inf` / `nan` / `0` was a bug because it disabled the timeout in the same way).

### F2 — Test gap: positive path covered, but no "boolean flip" regression

`TestProvisionWithCredentialsUsesPrivateEndpoint` sets `usePrivateEndpoint: true` and asserts `out.Host == "private.internal"`. There's no symmetric test that sets `usePrivateEndpoint: false` against a response with **both** public and private endpoints filled in, asserting `out.Host == "public.example"`. The diff history specifically calls out commit `47ebdfb` "fix: use if/else for private endpoint, no silent fallback to public" — exactly the kind of change where someone in a future refactor swaps the if/else and the existing tests would still pass (only one direction asserted).

Suggest one more 10-line test that mirrors the positive case with `usePrivateEndpoint: false` and a response that has both endpoints populated, asserting public is selected.

### F3 — `TestNewProvisionerFromEnvReadsPrivateEndpointFlag` doesn't cover the malformed-input path

Tests `default`, `true`, `1`, `no`. Doesn't test `on`, `Y`, `ture`, `2`, empty-after-trim. If F1 is addressed, these become regression cases. If F1 is "intentionally permissive", at minimum add `"on"` to the parametrize list to pin that contract.

### Non-findings (positive)

- `Username` field removed from API response structs + transition to `UserPrefix + ".root"`-only is clean. No dangling references in `pkg/tenant/tidbcloudnative/` after the PR.
- All `Endpoints.Public` / `Endpoints.Private` usages are within this file — no external public-endpoint consumer that would break when `usePrivate=true` leaves public empty.
- The `Provisioner.usePrivateEndpoint` field is read-only after construction, so no concurrency hazard.
- `t.Setenv` correctly used in tests — auto-cleans, no manual restore.
- Commit `9c040d3` removes dead-code endpoint emptiness checks after polling already guarantees the endpoint — verified by reading `waitForClusterProvisionMetadata` and `waitForBranchActive`, the polling loop calls `clusterProvisionMetadataIncomplete` / `branchConnectionIncomplete` which now correctly check the endpoint *type* matching `usePrivate`.

### Banked rules used this round

- "env→primitive parser must reject malformed input" (banked from kb-ai PR #8, msg `df4fd7d7` originally for `_env_float`): F1 is the same rule applied to env→bool when the boolean drives a security boundary.
- "Test must cover both directions of a conditional, not only the positive case" (informal): F2.

qiffang:
## Review (adversary-2): BLOCKER — pre-existing nil-deref in code path this PR consolidates

Verified locally on commit `5498d97`: build/vet clean, full `pkg/tenant/tidbcloudnative` suite passes (10.4s), all 6 new tests green.

### 🔴 BLOCKER — +1 @coderabbitai's Critical nil-pointer dereference (provisioner.go:268-277)

Independently confirmed by tracing `waitForClusterProvisionMetadata` (line 1088): every error return path (request error, body-read error, non-200 status, parse error, deadline, `ctx.Done()`) returns `nil, err`. The caller (`ProvisionWithCredentialsAndQuota`, line 268) reassigns `info, err = p.waitForClusterProvisionMetadata(...)` and on error path immediately reads `info.ClusterID` (line 272) — `info` is nil, panic.

**Reachability:** TiDB Cloud cluster endpoint metadata is asynchronously populated. Polling timeout (10 minutes) and context cancellation are common in production, not corner cases. Network blips during the first poll iteration produce request errors. Any of these → panic.

**Pre-existing or new?** Pre-existing (from PR #620). However, this PR's commit `9c040d3` ("cleanup: remove unused endpoint emptiness checks (polling already guarantees)") explicitly consolidates and reasons about this exact code path's invariants — the author touched it knowingly. That makes a one-line fix in-scope for this PR rather than a separate follow-up. Coderabbit's proposed `clusterID := info.ClusterID` capture-before-reassign fix is the minimal correct change.

### 🟡 Non-blocker — `parseBoolEnv` test coverage gap (provisioner.go:973-976)

```go
func parseBoolEnv(name string) bool {
    v := strings.TrimSpace(strings.ToLower(os.Getenv(name)))
    return v == "1" || v == "true" || v == "yes"
}
```

`TestNewProvisionerFromEnvReadsPrivateEndpointFlag` exercises `""`, `"true"`, `"1"`, `"no"` but **not `"yes"`** (which the function explicitly accepts) and no adversarial-typo cases (`"y"`, `"on"`, `"tru"`, `"enabled"`) which silently fall back to false.

For a feature-flag bool, silent-default-false is the conservative direction (won't accidentally enable private endpoint), so this is not a correctness blocker like the equivalent issue would be in a timeout parser. But the test should at minimum cover `"yes"` since it's in the accepted set. Optionally consider migrating to `strconv.ParseBool` with explicit error handling — it gives the operator a clear "unrecognized value" signal instead of silent default.

### 🟡 Non-blocker — `WaitForBranchUserWithCredentials` drops the direct-`Username` path (provisioner.go:1188-1192)

The diff removes:
```go
if info.Username != "" {
    return info.Username, nil
}
```

Now every branch user is unconditionally `UserPrefix + ".root"`. If the TiDB Cloud native API ever returned a `username` field for any branch in any state (legacy provisioning, migration scenarios), that response is now ignored and the function will keep polling for `UserPrefix`. The renamed test (`TestWaitForBranchUserWithCredentialsUsesUserPrefix`) confirms the intended new contract, but worth confirming with the author: **does the TiDB Cloud native API documentation guarantee `userPrefix` is set for every branch state we care about?** If yes, this simplification is clean. If we're inferring from observed behavior, there's a latent regression risk.

(Equivalent question doesn't apply to other providers: db9 and tidbzero still use a `username` JSON field. Scope of this drop is correctly limited to tidbcloudnative.)

### Other observations
- `tenant.ClusterInfo.Username` (consumer type) is preserved — no downstream breakage from the JSON struct field removal
- @coderabbitai's secondary nit (extract `selectEndpoint(endpoints, usePrivate)` helper) is reasonable; not a merge blocker
- Tests pass and the new private-endpoint coverage hits both incomplete-detection branches and the end-to-end provision path

### Verdict: BLOCK pending nil-deref fix

The nil-deref is Critical-severity and reachable. Fix per coderabbit's diff; add an `info` nil-guard in the test for `waitForClusterProvisionMetadata` if not already present. Other items are non-blockers.

qiffang:
Blocking issue on current head `5498d97ebd9b80676203ecb571ed50110042cc64`.

### B1 — private-endpoint metadata wait errors still panic by dereferencing `info` after reassignment

CodeRabbit flagged this earlier, and it is still present in current head. In `ProvisionWithCredentialsAndQuota`, the code reassigns `info` with `p.waitForClusterProvisionMetadata(...)` and then dereferences `info.ClusterID` in the error branch:

```go
if clusterProvisionMetadataIncomplete(info, p.usePrivateEndpoint) {
    info, err = p.waitForClusterProvisionMetadata(ctx, publicKey, privateKey, info.ClusterID)
    if err != nil {
        return &tenant.ClusterInfo{
            TenantID:  tenantID,
            ClusterID: info.ClusterID, // info is nil on wait errors
            Password:  password,
            DBName:    dbName,
            Provider:  tenant.ProviderTiDBCloudNative,
        }, nil, err
    }
}
```

`waitForClusterProvisionMetadata` returns `nil, err` on request errors, parse errors, timeout, and `ctx.Done()`. In private-endpoint mode this is reachable when the initial create response has public endpoint metadata but private endpoint metadata does not become complete before the wait returns an error.

I reproduced with a temporary test that keeps returning an ACTIVE cluster with only the public endpoint while `usePrivateEndpoint=true`, then cancels the wait context:

```go
func TestProvisionWithCredentialsPrivateEndpointWaitErrorDoesNotPanic_TEMP(t *testing.T) {
    oldPoll := tidbCloudNativePollInterval
    tidbCloudNativePollInterval = time.Millisecond
    t.Cleanup(func() { tidbCloudNativePollInterval = oldPoll })

    ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
        if r.Header.Get("Authorization") == "" {
            w.Header().Set("WWW-Authenticate", `Digest realm="tidbcloud", nonce="nonce-1", qop="auth"`)
            w.WriteHeader(http.StatusUnauthorized)
            return
        }
        _ = json.NewEncoder(w).Encode(map[string]any{
            "clusterId":  "cluster-1",
            "state":      "ACTIVE",
            "userPrefix": "u1",
            "labels":     map[string]string{TiDBCloudOrganizationLabel: "org-1"},
            "endpoints": map[string]any{
                "public": map[string]any{"host": "public.example", "port": 4000},
            },
        })
    }))
    defer ts.Close()

    p := &Provisioner{
        apiURL:              ts.URL,
        defaultDatabaseName: DefaultDatabaseName,
        usePrivateEndpoint:  true,
        client:              ts.Client(),
    }
    ctx, cancel := context.WithTimeout(context.Background(), 5*time.Millisecond)
    defer cancel()
    out, _, err := p.ProvisionWithCredentialsAndQuota(ctx, "tenant-1", tenant.CredentialProvisionRequest{
        PublicKey: "public-1", PrivateKey: "private-1",
    }, tenant.QuotaUpdateOptions{})
    if err == nil {
        t.Fatal("ProvisionWithCredentialsAndQuota err = nil, want timeout/error")
    }
    if out == nil || out.ClusterID != "cluster-1" {
        t.Fatalf("out = %#v, want partial cluster info", out)
    }
}
```

Result:

```text
panic: runtime error: invalid memory address or nil pointer dereference
...
github.com/mem9-ai/drive9/pkg/tenant/tidbcloudnative.(*Provisioner).ProvisionWithCredentialsAndQuota
    pkg/tenant/tidbcloudnative/provisioner.go:272
```

Suggested minimal fix is to capture `clusterID := info.ClusterID` before the wait and use the captured value in the error branch. Please add a regression covering the private-endpoint wait-error path so this does not come back.

Clean-branch validation I ran before the temporary reproducer:

```text
git diff --check origin/main...origin/pr-626
go test ./pkg/tenant/tidbcloudnative -count=1
go test ./pkg/tenant/... -count=1
```

Those pass, but the reachable wait-error path above still panics.

qiffang:
**Addendum — endorsing @dev-2's parseBoolEnv finding, retracting my "non-blocker" classification.**

I classified `parseBoolEnv` silent-default-to-false as a "minor nit" in my comment above. After reading @dev-2's review, their framing is sharper and correct: this env controls the **public-vs-private network boundary** for already-provisioned tenants, and ops typos (`=on`, `=Y`, `=ture`, `=2`, `=enable`) silently produce the wrong network configuration. The tenant record is durable; correcting it later is much more expensive than failing fast at startup.

Same banked rule as kb-ai PR #8 (`_env_float` silently accepting `inf`/`nan` to disable timeout): **env→primitive parsers must reject malformed values, not whitelist accepted values and default everything else**. Recognize `{1,true,yes,0,false,no}`, reject anything else.

**Upgraded combined verdict (mine):** BLOCK pending (1) nil-deref fix + (2) `parseBoolEnv` strict-mode rejection. Non-blocker items unchanged.

qiffang:
Additional review on current head `710214bdc8fe97f908b3a897bab8986389e49ee8`.

I agree this should not merge yet. I independently verified the two existing blockers are still present on current head:

1. `ProvisionWithCredentialsAndQuota` can still panic after `waitForClusterProvisionMetadata` returns an error. The code reassigns `info` from a helper that returns `nil, err`, then the error branch reads `info.ClusterID`. This is reachable on metadata polling timeout / context cancel / request or parse errors. Capture `clusterID := info.ClusterID` before the reassignment and use the saved value in the fallback `tenant.ClusterInfo`.

2. `parseBoolEnv` still silently maps malformed non-empty values to `false`. For `DRIVE9_TIDBCLOUD_NATIVE_USE_PRIVATE_ENDPOINT`, that means typos such as `ture`, `on`, or `enabled` can permanently provision a tenant with the public endpoint when the operator intended private mode. Since `NewProvisionerFromEnv` already returns `error`, this should be a strict parser: empty/default is false, recognized true/false values are accepted, and any other non-empty value returns an error.

One additional mechanical blocker: current head is not `gofmt` clean in `pkg/tenant/tidbcloudnative/provisioner.go` after adding the Tencent override fields. `gofmt -d` changes the const block, `Provisioner` field alignment, and `Provisioner{...}` literal alignment around the new env/field.

I did not find a new semantic blocker beyond those. Please fix the two correctness issues plus run `gofmt` before merge.

qiffang:
Re-review on current head `4aa323ff949731370ffac00a3e95d2e4c674afa7`.

The two semantic blockers I called out earlier are fixed:

- `ProvisionWithCredentialsAndQuota` now captures `clusterID` before reassigning `info`, so metadata polling errors no longer dereference `nil`.
- `parseBoolEnv` now returns an error for malformed non-empty values instead of silently treating them as `false`, and `NewProvisionerFromEnv` propagates that error.

One blocker remains before merge: current head is still not `gofmt` clean in `pkg/tenant/tidbcloudnative/provisioner.go`. Running `gofmt -d` changes the const block and the `Provisioner` struct / literal alignment around `EnvTiDBCloudNativeUsePrivateEndpoint`, `EnvTiDBCloudTencentPrivateEndpointHost`, and `tencentPrivateEndpointHost`.

Please run `gofmt` on the touched Go files and push the result. After that, my lane has no remaining semantic blocker on the current head.

qiffang:
Re-review on current head `76ef3438c3be5540ad35e42719cd1ea9d5325a9a`.

The blockers from my previous lane are now fixed:

- `ProvisionWithCredentialsAndQuota` captures `clusterID` before reassigning `info`, so metadata wait errors no longer dereference `nil`.
- `parseBoolEnv` now rejects malformed non-empty values and `NewProvisionerFromEnv` propagates the error.
- The touched Go files are now `gofmt` clean.

One small remaining blocker before merge: please add a regression test for malformed `DRIVE9_TIDBCLOUD_NATIVE_USE_PRIVATE_ENDPOINT` values, e.g. `ture`, `on`, `enabled`, or `2`, asserting `NewProvisionerFromEnv` returns an error. This is the exact safety contract that blocked merge: malformed private-endpoint config must fail closed instead of silently provisioning public, so it should be pinned in tests.

After that test is added, my lane has no remaining blocker.

qiffang (approved):

PR #627, opened 2026-06-26
Add FS PUT body read latency metrics

qiffang:
## Summary
- Add `drive9_http_request_body_read_duration_seconds` histogram for inline `PUT /v1/fs/*` request-body read time.
- Record final HTTP `status_class` and low-cardinality `body_size_bucket` labels, with no path/tenant/workspace labels.
- Add slow body-read server log after 5s with route, bytes, remote, trace-context logger fields, and rate bucket.
- Move Prometheus alert rules to the canonical runbooks repo: https://github.com/tidbcloud/runbooks/pull/2622

## Notes
- This does not change multipart behavior. Large files that already cross the backend large-file threshold still initiate multipart instead of reading the full body in `handleWrite`.
- The body-read alert should be treated as client upload / network-path evidence before blaming backend write latency.
- This PR no longer carries `docs/grafana/drive9-alert-rules.yaml`; alert ownership is in `tidbcloud/runbooks`.

## Validation
- `/usr/local/go/bin/go test ./pkg/metrics`
- `/usr/local/go/bin/go test -c ./pkg/server ./pkg/metrics`
- `git diff --check`

`/usr/local/go/bin/go test -run 'TestMetricsEndpoint|TestWriteBodyReadMetricRecordsFailureStatusClass' ./pkg/server` is blocked locally by testcontainers startup: `rootless Docker not found`.


srstack (dismissed):
LGTM — approving.

Traced the change against `handleWrite`, the `observe()` middleware, and the histogram config. `go test ./pkg/metrics` and `go vet ./pkg/server ./pkg/metrics` pass locally.

The design is clean: deferring the body-read metric emission to `observe()` so it can be tagged with the final `ow.status` is the right way to label a mid-handler measurement with the eventual response class. I verified route normalization is consistent (`requestRoute` used in both places), the request metric state is always installed before the handler runs, multipart/large-file uploads are genuinely untouched (the metric is only set after the `IsLargeFile` branch returns), cardinality is bounded (method × route × status_class(5) × body_size_bucket(8), no path/tenant labels), and failure paths are captured (the 4xx test confirms it).

One thing worth considering (non-blocking): `httpDurationBounds` tops out at 10s, so the body-read p99 saturates at the +Inf bucket exactly where this metric is meant to be most useful (slow uploads can run tens of seconds). The >5s alert still fires correctly, but the displayed p99 understates the true tail. A body-read-specific bounds slice extending to e.g. 15/30/60s would make the value itself trustworthy.

Minor: `body_size_bucket`/`rate_bucket` use the declared Content-Length, which can overstate on aborted reads — worth a comment noting the bytes are declared, not transferred.

srstack (approved):

PR #628, opened 2026-06-26
Fix FUSE mount startup cleanup and diagnostics

you06:
## Summary

- clean up partially-started FUSE mounts when Drive9 mount startup fails
- force-unmount the mountpoint after go-fuse returns an INIT-stage `init: ...` error without a server handle
- add safe startup diagnostics that report the failed phase, mountpoint, and underlying error without printing server, remote, token, or API key values

## Details

Drive9 could leave a stale FUSE mount when mount startup failed after the kernel/FUSE INIT path had already begun. It showed up as a later `transport endpoint is not connected` mountpoint with no live `drive9` process, making retries in the same sandbox fail.

This change covers both startup failure shapes:

1. `gofuse.NewServer()` can mount and handle INIT synchronously, then return an `init: ...` error before Drive9 has a server handle. In that case, Drive9 now logs a safe diagnostic and calls the existing force-unmount fallback for the mountpoint.
2. After a server exists, failures from `WaitMount()`, mount control socket setup, or process-state writing now stop watchers, flush pending state, try `server.Unmount()`, and fall back to force-unmount if normal unmount fails.

The added diagnostics make foreground failures actionable. For example:

```text
drive9: mount startup failed during fuse wait mount failure at /home/user/workspace: permission denied
drive9: cleanup after fuse wait mount failure: unmounted /home/user/workspace
mount: fuse wait mount: permission denied
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved mount startup failure handling with a consistent best-effort cleanup sequence: stop watchers, flush pending data, and unmount (with forced unmount fallback when appropriate).
  * Added targeted behavior for partial initialization failures to force unmount only when needed.
  * Ensured the control server is closed as part of the unified cleanup flow.
* **Tests**
  * Added/expanded tests covering cleanup ordering, forced unmount behavior (including “transport endpoint is not connected”), and correct detection of init-failure cases (including skipping force unmount for empty mount points).

### DIFF
--- a/pkg/fuse/mount.go
+++ b/pkg/fuse/mount.go
@@ -470,14 +470,26 @@ func Mount(opts *MountOptions) error {
 	// thread consumes the last GOMAXPROCS slot, leaving no thread to handle
 	// the POLL request from the kernel).
 	if err := server.WaitMount(); err != nil {
-		stopWatchers()
+		cleanupMountStartFailure(mountStartCleanup{
+			reason:       "fuse wait mount failure",
+			mountPoint:   opts.MountPoint,
+			stopWatchers: stopWatchers,
+			flushAll:     dat9fs.FlushAll,
+			unmount:      server.Unmount,
+			forceUnmount: forceUnmount,
+		})
 		return fmt.Errorf("fuse wait mount: %w", err)
 	}
 	controlServer, err := startMountControlServer(opts.MountPoint, dat9fs)
 	if err != nil {
-		stopWatchers()
-		dat9fs.FlushAll()
-		_ = server.Unmount()
+		cleanupMountStartFailure(mountStartCleanup{
+			reason:       "mount control socket failure",
+			mountPoint:   opts.MountPoint,
+			stopWatchers: stopWatchers,
+			flushAll:     dat9fs.FlushAll,
+			unmount:      server.Unmount,
+			forceUnmount: forceUnmount,
+		})
 		return fmt.Errorf("start mount control socket: %w", err)
 	}
 	if controlServer != nil {
@@ -519,9 +531,17 @@ func Mount(opts *MountOptions) error {
 		ControlSocket:       controlServer.SocketPath(),
 	})
 	if err != nil {
-		stopWatchers()
-		dat9fs.FlushAll()
-		_ = server.Unmount()
+		if controlServer != nil {
+			controlServer.Close()
+		}
+		cleanupMountStartFailure(mountStartCleanup{
+			reason:       "mount process state failure",
+			mountPoint:   opts.MountPoint,
+			stopWatchers: stopWatchers,
+			flushAll:     dat9fs.FlushAll,
+			unmount:      server.Unmount,
+			forceUnmount: forceUnmount,
+		})
 		return fmt.Errorf("write mount pid file: %w", err)
 	}
 	defer func() {
@@ -653,6 +673,37 @@ func Mount(opts *MountOptions) error {
 	server.Wait()
 	shutdown()
 	return nil
+}
+
+type mountStartCleanup struct {
+	reason       string
+	mountPoint   string
+	stopWatchers func()
+	flushAll     func()
+	unmount      func() error
+	forceUnmount func(string)
+}
+
+func cleanupMountStartFailure(cleanup mountStartCleanup) {
+	if cleanup.stopWatchers != nil {
+		cleanup.stopWatchers()
+	}
+	if cleanup.flushAll != nil {
+		cleanup.flushAll()
+	}
+	if cleanup.unmount == nil {
+		return
+	}
+	if err := cleanup.unmount(); err != nil {
+		reason := strings.TrimSpace(cleanup.reason)
+		if reason == "" {
+			reason = "mount start failure"
+		}
+		fmt.Fprintf(os.Stderr, "drive9: cleanup after %s: unmount failed: %v\n", reason, err)
+		if cleanup.forceUnmount != nil && strings.TrimSpace(cleanup.mountPoint) != "" {
+			cleanup.forceUnmount(cleanup.mountPoint)
+		}
+	}
 }
 
 func validateMountOptionsProfile(opts *MountOptions) error {
--- a/pkg/fuse/mount.go
+++ b/pkg/fuse/mount.go
@@ -446,6 +446,19 @@ func Mount(opts *MountOptions) error {
 	server, err := gofuse.NewServer(dat9fs, opts.MountPoint, fuseOpts)
 	if err != nil {
 		layerEventWatcherStop()
+		cleanupMountStartFailure(mountStartCleanup{
+			reason:     "fuse server initialization failure",
+			mountPoint: opts.MountPoint,
+			cause:      err,
+		})
+		if shouldForceUnmountAfterNewServerError(err) {
+			mountPoint := strings.TrimSpace(opts.MountPoint)
+			if mountPoint == "" {
+				mountPoint = "<unknown>"
+			}
+			fmt.Fprintf(os.Stderr, "drive9: cleanup after fuse server initialization failure: forcing unmount of %s after partial init failure\n", mountPoint)
+			forceUnmount(opts.MountPoint)
+		}
 		return fmt.Errorf("fuse mount: %w", err)
 	}
 
@@ -473,6 +486,7 @@ func Mount(opts *MountOptions) error {
 		cleanupMountStartFailure(mountStartCleanup{
 			reason:       "fuse wait mount failure",
 			mountPoint:   opts.MountPoint,
+			cause:        err,
 			stopWatchers: stopWatchers,
 			flushAll:     dat9fs.FlushAll,
 			unmount:      server.Unmount,
@@ -485,6 +499,7 @@ func Mount(opts *MountOptions) error {
 		cleanupMountStartFailure(mountStartCleanup{
 			reason:       "mount control socket failure",
 			mountPoint:   opts.MountPoint,
+			cause:        err,
 			stopWatchers: stopWatchers,
 			flushAll:     dat9fs.FlushAll,
 			unmount:      server.Unmount,
@@ -537,6 +552,7 @@ func Mount(opts *MountOptions) error {
 		cleanupMountStartFailure(mountStartCleanup{
 			reason:       "mount process state failure",
 			mountPoint:   opts.MountPoint,
+			cause:        err,
 			stopWatchers: stopWatchers,
 			flushAll:     dat9fs.FlushAll,
 			unmount:      server.Unmount,
@@ -678,13 +694,35 @@ func Mount(opts *MountOptions) error {
 type mountStartCleanup struct {
 	reason       string
 	mountPoint   string
+	cause        error
 	stopWatchers func()
 	flushAll     func()
 	unmount      func() error
 	forceUnmount func(string)
+	logf         func(string, ...any)
 }
 
 func cleanupMountStartFailure(cleanup mountStartCleanup) {
+	reason := strings.TrimSpace(cleanup.reason)
+	if reason == "" {
+		reason = "mount start failure"
+	}
+	mountPoint := strings.TrimSpace(cleanup.mountPoint)
+	mountPointLabel := mountPoint
+	if mountPointLabel == "" {
+		mountPointLabel = "<unknown>"
+	}
+	logf := cleanup.logf
+	if logf == nil {
+		logf = func(format string, args ...any) {
+			fmt.Fprintf(os.Stderr, format, args...)
+		}
+	}
+	if cleanup.cause != nil {
+		logf("drive9: mount startup failed during %s at %s: %v\n", reason, mountPointLabel, cleanup.cause)
+	} else {
+		logf("drive9: mount startup failed during %s at %s\n", reason, mountPointLabel)
+	}
 	if cleanup.stopWatchers != nil {
 		cleanup.stopWatchers()
 	}
@@ -695,15 +733,21 @@ func cleanupMountStartFailure(cleanup mountStartCleanup) {
 		return
 	}
 	if err := cleanup.unmount(); err != nil {
-		reason := strings.TrimSpace(cleanup.reason)
-		if reason == "" {
-			reason = "mount start failure"
+		logf("drive9: cleanup after %s: unmount failed: %v\n", reason, err)
+		if cleanup.forceUnmount != nil && mountPoint != "" {
+			cleanup.forceUnmount(mountPoint)
 		}
-		fmt.Fprintf(os.Stderr, "drive9: cleanup after %s: unmount failed: %v\n", reason, err)
-		if cleanup.forceUnmount != nil && strings.TrimSpace(cleanup.mountPoint) != "" {
-			cleanup.forceUnmount(cleanup.mountPoint)
-		}
+		return
 	}
+	if mountPoint != "" {
+		logf("drive9: cleanup after %s: unmounted %s\n", reason, mountPoint)
+	} else {
+		logf("drive9: cleanup after %s: unmounted mountpoint\n", reason)
+	}
+}
+
+func shouldForceUnmountAfterNewServerError(err error) bool {
+	return err != nil && strings.HasPrefix(err.Error(), "init:")
 }
 
 func validateMountOptionsProfile(opts *MountOptions) error {
--- a/pkg/fuse/mount.go
+++ b/pkg/fuse/mount.go
@@ -445,20 +445,7 @@ func Mount(opts *MountOptions) error {
 	// Create FUSE server
 	server, err := gofuse.NewServer(dat9fs, opts.MountPoint, fuseOpts)
 	if err != nil {
-		layerEventWatcherStop()
-		cleanupMountStartFailure(mountStartCleanup{
-			reason:     "fuse server initialization failure",
-			mountPoint: opts.MountPoint,
-			cause:      err,
-		})
-		if shouldForceUnmountAfterNewServerError(err) {
-			mountPoint := strings.TrimSpace(opts.MountPoint)
-			if mountPoint == "" {
-				mountPoint = "<unknown>"
-			}
-			fmt.Fprintf(os.Stderr, "drive9: cleanup after fuse server initialization failure: forcing unmount of %s after partial init failure\n", mountPoint)
-			forceUnmount(opts.MountPoint)
-		}
+		cleanupNewServerFailure(opts.MountPoint, err, layerEventWatcherStop, dat9fs.FlushAll, forceUnmount, nil)
 		return fmt.Errorf("fuse mount: %w", err)
 	}
 
@@ -483,6 +470,8 @@ func Mount(opts *MountOptions) error {
 	// thread consumes the last GOMAXPROCS slot, leaving no thread to handle
 	// the POLL request from the kernel).
 	if err := server.WaitMount(); err != nil {
+		// NewServer succeeded and Serve is running, so cleanup can drain Drive9-side
+		// workers before asking go-fuse to detach any partially mounted kernel state.
 		cleanupMountStartFailure(mountStartCleanup{
 			reason:       "fuse wait mount failure",
 			mountPoint:   opts.MountPoint,
@@ -699,7 +688,10 @@ type mountStartCleanup struct {
 	flushAll     func()
 	unmount      func() error
 	forceUnmount func(string)
-	logf         func(string, ...any)
+	// forceUnmountWithoutServer is for the go-fuse NewServer post-mount/pre-server
+	// INIT failure path, where no Server exists to call Unmount on.
+	forceUnmountWithoutServer bool
+	logf                      func(string, ...any)
 }
 
 func cleanupMountStartFailure(cleanup mountStartCleanup) {
@@ -730,6 +722,10 @@ func cleanupMountStartFailure(cleanup mountStartCleanup) {
 		cleanup.flushAll()
 	}
 	if cleanup.unmount == nil {
+		if cleanup.forceUnmountWithoutServer && cleanup.forceUnmount != nil && mountPoint != "" {
+			logf("drive9: cleanup after %s: forcing unmount of %s after partial init failure\n", reason, mountPoint)
+			cleanup.forceUnmount(mountPoint)
+		}
 		return
 	}
 	if err := cleanup.unmount(); err != nil {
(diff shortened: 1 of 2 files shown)


srstack (approved):
LGTM — approving.

Verified the change end-to-end against the handler, the observe middleware, and the control-server impl; `go test ./pkg/fuse` (cleanup tests) and `go vet` pass locally.

The core fix is correct: the `WaitMount` failure path was genuinely leaking a partially-started mount (previously only `stopWatchers()`), and centralizing teardown into `cleanupMountStartFailure` (stop → flush → unmount → force-unmount fallback) is a net readability win. Secret hygiene is respected (only reason/mountpoint/cause logged). I confirmed the two things that look risky are actually safe:
- The double `controlServer.Close()` in the pid-file error path (explicit + deferred) is safe: the second call hits `net.ErrClosed` (ignored), receives from an already-closed `done` channel (no block), and re-`os.Remove` is a no-op.
- `forceUnmount` on a not-actually-mounted point just logs a benign failure under a 5s timeout.

Non-blocking nits:
1. Add a one-line comment on the explicit `controlServer.Close()` noting the deferred one is a harmless no-op (saves the next reader the same investigation).
2. `shouldForceUnmountAfterNewServerError` matches go-fuse's `init:` error prefix by string — worth a comment pinning it to the go-fuse behavior so a dep bump triggers a re-check.

qiffang (commented):
## PR #628 Review — Fix FUSE mount startup cleanup and diagnostics

### Summary

This PR fixes stale FUSE mounts left behind when mount startup fails after the kernel INIT handshake has begun. The fix centralizes cleanup into `cleanupMountStartFailure()` with a `mountStartCleanup` struct that captures the available cleanup actions at each failure point. For `NewServer` errors with `init:` prefix (where FUSE INIT partially completed but no server handle exists), it force-unmounts the mountpoint.

### Assessment

Clean, well-structured implementation. The cleanup struct pattern avoids duplicated teardown code across 4 failure points. Tests cover the happy path (stop → flush → unmount ordering), force-unmount fallback, and the `init:` prefix detection. See inline comments.

pkg/fuse/mount.go
```diff
@@ -655,6 +691,65 @@ func Mount(opts *MountOptions) error {
...
+	}
+}
+
+func shouldForceUnmountAfterNewServerError(err error) bool {
+	return err != nil && strings.HasPrefix(err.Error(), "init:")
+}
+
 func validateMountOptionsProfile(opts *MountOptions) error {
 	if !validMountProfile(opts.Profile) {
 		return fmt.Errorf("mount: unknown profile %q", opts.Profile)
```
qiffang, line 913: **Fragile — `strings.HasPrefix(err.Error(), "init:")` depends on go-fuse error message format.**

If go-fuse changes the error message prefix in a future version, this detection silently stops working and stale mounts will be left behind again. Consider:

1. Adding a comment noting which go-fuse version this targets
2. Checking for the FUSE mountpoint state directly (e.g. `os.Stat(mountPoint)` returns "transport endpoint is not connected") as a secondary signal
3. Always force-unmounting on `NewServer` failure (the mountpoint should not be in use if NewServer failed)

Option 3 is arguably safer — if `NewServer` fails, no server is running, so force-unmounting should be safe regardless of the error prefix.

pkg/fuse/mount.go
```diff
@@ -655,6 +691,65 @@ func Mount(opts *MountOptions) error {
...
+	reason := strings.TrimSpace(cleanup.reason)
+	if reason == "" {
+		reason = "mount start failure"
+	}
+	mountPoint := strings.TrimSpace(cleanup.mountPoint)
+	mountPointLabel := mountPoint
+	if mountPointLabel == "" {
+		mountPointLabel = "<unknown>"
+	}
+	logf := cleanup.logf
```
qiffang, line 718: **Good — cleanup ordering is correct: stopWatchers → flushAll → unmount → forceUnmount.**

The struct-based approach ensures consistent ordering across all 4 failure points. The nil checks on each function allow progressive cleanup (early failures have fewer cleanup actions available). Clean pattern.

pkg/fuse/mount.go
```diff
@@ -470,14 +483,28 @@ func Mount(opts *MountOptions) error {
 	// thread consumes the last GOMAXPROCS slot, leaving no thread to handle
 	// the POLL request from the kernel).
 	if err := server.WaitMount(); err != nil {
-		stopWatchers()
+		cleanupMountStartFailure(mountStartCleanup{
+			reason:       "fuse wait mount failure",
+			mountPoint:   opts.MountPoint,
+			cause:        err,
```
qiffang, line 486: **Observation — the `WaitMount` failure path does not close `controlServer`.**

At the `WaitMount` failure point, `controlServer` has not been created yet (it is started after WaitMount succeeds), so this is correct. But the "mount process state" failure path (line 554) now explicitly closes `controlServer` — good, this was missing in the original code.

pkg/fuse/mount.go
```diff
@@ -655,6 +691,65 @@ func Mount(opts *MountOptions) error {
...
+	if cleanup.cause != nil {
+		logf("drive9: mount startup failed during %s at %s: %v\n", reason, mountPointLabel, cleanup.cause)
+	} else {
+		logf("drive9: mount startup failed during %s at %s\n", reason, mountPointLabel)
+	}
+	if cleanup.stopWatchers != nil {
+		cleanup.stopWatchers()
+	}
+	if cleanup.flushAll != nil {
+		cleanup.flushAll()
```
qiffang, line 733: **Good — diagnostics print only mountpoint, reason, and cause error.**

The `cleanupMountStartFailure` function intentionally avoids printing server URLs, remote paths, tokens, or API keys. The `cause` error comes from go-fuse or the OS, which should not contain credentials. This matches the PR description's goal of safe startup diagnostics.

pkg/fuse/mount_shutdown_test.go
```diff
@@ -41,3 +44,80 @@ func TestMountShutdownRunsOnce(t *testing.T) {
...
+}
+
+func TestShouldForceUnmountAfterNewServerErrorOnlyForInitFailure(t *testing.T) {
+	if !shouldForceUnmountAfterNewServerError(errors.New("init: ENODEV")) {
+		t.Fatal("init failure should force cleanup")
+	}
+	if shouldForceUnmountAfterNewServerError(errors.New("fusermount: mountpoint is busy")) {
+		t.Fatal("pre-mount failure should not force cleanup")
+	}
+	if shouldForceUnmountAfterNewServerError(nil) {
```
qiffang, line 206: **Good tests — cover the three key scenarios.**

1. Normal cleanup ordering (stop → flush → unmount, no force)
2. Force-unmount fallback when `server.Unmount()` fails
3. `init:` prefix detection (positive, negative, nil)

Consider adding a test for the case where `mountPoint` is empty — `cleanupMountStartFailure` handles it (prints `<unknown>`) but the force-unmount path skips when `mountPoint == ""`. Worth verifying that empty-mountpoint cleanup does not call `forceUnmount("")`.

qiffang (changes requested):
## Adversarial Review — PR #628

**Verdict: REVISE** (2 blockers, 2 observations)

### Checklist

| Gate | Status |
|------|--------|
| Edge cases covered? | FAIL (B1) |
| Failure behavior defined? | FAIL (B2) |
| API explicit? | PASS |
| Race / state issues? | N/A |

---

### B1 — `shouldForceUnmountAfterNewServerError` relies on fragile error string matching

`strings.HasPrefix(err.Error(), "init:")` is a string match against go-fuse's internal error formatting. This has two failure modes:

1. **False negative**: If go-fuse changes the error prefix (e.g., `"INIT:"`, `"init failed:"`, or wraps the error), force unmount silently stops working. The stale mount problem this PR exists to fix returns with no indication.
2. **False positive**: If go-fuse returns an `init:` error without having registered the mount with the kernel (e.g., an early INIT negotiation failure before the mount point is live), force-unmounting a non-existent mount is harmless but noisy.

The test covers the current string shape but doesn't protect against upstream changes. At minimum, this needs a comment documenting which go-fuse code path produces the `init:` prefix and what guarantees (if any) exist about its stability. If no such guarantee exists, this is a time bomb.

### B2 — `WaitMount` failure path has silent behavior change: adds `FlushAll` + `Unmount` + `forceUnmount` where original code only called `stopWatchers`

Original code at `WaitMount` failure:
```go
stopWatchers()
return fmt.Errorf("fuse wait mount: %w", err)
```

Refactored code adds:
- `dat9fs.FlushAll()` — flushes to a server that may never have fully started
- `server.Unmount()` — interacts with kernel FUSE state that may be partially initialized
- `forceUnmount()` — as fallback if `Unmount()` fails

The original code deliberately omitted these — presumably because after `WaitMount` failure, the FUSE mount is in an indeterminate state where `FlushAll` may hang or corrupt, and `Unmount` may interact with partially-initialized kernel state.

If this behavior change is intentional (not just a refactoring artifact), it needs explicit justification: **why is it safe to call `FlushAll` and `Unmount` after `WaitMount` fails?** What state is the server in at that point? The PR description says "stop watchers, flush pending state, try `server.Unmount()`" for post-server failures, but doesn't distinguish which cleanup steps are safe at each failure point.

The struct-based cleanup pattern forces all 3 failure points through the same code path, but each failure point has different preconditions. A refactor that unifies cleanup must not silently expand cleanup scope.

---

### O1 — `forceUnmount` call after `NewServer` failure bypasses `logf` abstraction

At line 19 of the diff, `fmt.Fprintf(os.Stderr, ...)` writes directly to stderr instead of going through `cleanupMountStartFailure`'s `logf`. The `cleanupMountStartFailure` call on line 9 already logged via `logf` (defaulting to stderr), but the subsequent `forceUnmount` block has its own separate `fmt.Fprintf`. Two separate logging paths for the same failure event.

### O2 — `controlServer.Close()` added only to pid-file failure path

The third failure point (pid-file write, line 65-67) now explicitly closes `controlServer` before cleanup. The second failure point (control socket setup failure) doesn't need this because `controlServer` isn't created yet. This is correct but worth noting that the `cleanupMountStartFailure` struct doesn't include `controlServer.Close()` — it's handled outside the struct. If future failure points are added between control socket creation and pid-file write, the `controlServer.Close()` call could be missed.

---

**CI**: `ci` pass, `drive9-server-local e2e smoke` pass, CodeRabbit pass on head `8ad0969b`.

**Tests**: 3 new tests covering cleanup ordering, force unmount fallback, and `shouldForceUnmountAfterNewServerError` predicate. Tests are well-structured.

Fix B1 (document or replace fragile string match) and B2 (justify or revert the `WaitMount` cleanup expansion), then re-review.

pkg/fuse/mount.go
```diff
@@ -446,6 +446,19 @@ func Mount(opts *MountOptions) error {
 	server, err := gofuse.NewServer(dat9fs, opts.MountPoint, fuseOpts)
 	if err != nil {
 		layerEventWatcherStop()
+		cleanupMountStartFailure(mountStartCleanup{
```
qiffang: This does not actually run the full startup cleanup for the `NewServer` failure case. By the time `gofuse.NewServer` is called, `dat9fs` may already have a commit queue, write-back uploader, disk read cache, journal, shadow store, and recovered pending work. This branch only calls `layerEventWatcherStop()` and then invokes `cleanupMountStartFailure` without `stopWatchers` or `flushAll`, so `FlushAll()` is never called and those resources are not drained/closed when `NewServer` fails.

That leaves the pre-server startup failure shape partially leaking the same resources the new helper is meant to clean up. Please route this branch through `cleanupMountStartFailure` with at least `stopWatchers: layerEventWatcherStop` and `flushAll: dat9fs.FlushAll` before returning, and add a regression that exercises the `NewServer`-failure cleanup contract.

qiffang (changes requested):
I’m requesting changes on head `8ad0969`.

The main cleanup path is directionally correct, and the post-server failure paths now use the shared stop → flush → unmount → force-unmount fallback. However, the pre-server `gofuse.NewServer` failure branch still only stops the layer watcher and does not drain/close the `Dat9FS` resources initialized before `NewServer`. That leaves the original class of startup-failure cleanup leak partially unresolved.

Validation notes: GitHub `code-ci` and `local-e2e` are green on this head, CodeRabbit status is success, and local `git diff --check origin/main...HEAD` passed. I could not run local Go tests because this machine has no `go` installed.

qiffang:
## Review (adversary-2) — independent verification + delta

I independently verified qiffang's B1 + B2 (both real), and add 3 orthogonal observations.

### +1 B1 (string-match on `"init:"` prefix is upstream-coupled and fragile)

Independently checked the go-fuse source against drive9's `replace` target (`github.com/mornyx/go-fuse/v2 v2.9.1-0.20260618082229-99faf5aab394`):

```
$GOPATH/pkg/mod/github.com/mornyx/go-fuse/v2@.../fuse/server.go:253
    return nil, fmt.Errorf("init: %s", code)
```

Confirmed `init:` is the actual prefix go-fuse emits. The comment on `shouldForceUnmountAfterNewServerError` ("Re-check this predicate when go-fuse is upgraded or replaced") is honest about the coupling but doesn't solve the fragility. Notably, go-fuse's own code has a `// TODO - unmount as well?` adjacent to that error — upstream itself has flagged this as half-finished. So drive9's compensating cleanup IS necessary; the only real question is how to bind to it safely.

**Suggested mitigation orthogonal to qiffang's**: a CI-only smoke test that pokes go-fuse's actual error path (e.g., via a mountpoint with permissions that trigger handleInit failure) and asserts our `shouldForceUnmountAfterNewServerError` returns true. The current test uses fixture text `errors.New("init: ENODEV")` — if go-fuse later changes the prefix or wrapping, that test still passes while production breaks silently. Pinning against real upstream output is the only way to detect the drift this comment warns about.

### +1 B2 (WaitMount path silently expanded from `stopWatchers()` to full cleanup)

Confirmed by reading PR's mount.go:473-481. Original was minimal; refactored path now also calls `FlushAll + Unmount + forceUnmount`. I traced FlushAll into `dat9fs.go:12073` — at WaitMount-failure time, `fs.fileHandles` should be empty (no Open has happened since kernel mount registration hasn't completed), so the per-handle drain loop is a no-op. The other FlushAll steps (debouncer, git overlay, commit queue) operate on internal state, not the FUSE pipe, so they're not hazardous per se. **But this defense doesn't substitute for the author justifying the change**: a future contributor reading the refactor will not know the FlushAll cost is bounded at this point. Either (a) document the precondition ("safe here because fileHandles is empty pre-mount-registration") in the cleanup struct's docstring, or (b) split `WaitMount` failure into its own narrower cleanup. qiffang's B2 — "must not silently expand cleanup scope without justification" — stands.

### O3 — Test fixture text isn't bound to real go-fuse output (meta-B1)

Both tests `TestCleanupNewServerFailureStopsFlushesAndForceUnmountsInitFailure` and `TestShouldForceUnmountAfterNewServerErrorOnlyForInitFailure` use literal `errors.New("init: ENODEV")` as the fixture. This is *internally* correct but creates a deceptive safety net: if go-fuse changes the prefix in a future upgrade (B1's exact failure mode), our tests still pass because they test against our own fixture, not real upstream behavior. **Suggest**: add a `//go:build linux` integration test that invokes `gofuse.NewServer` against a deliberately-broken mountpoint and asserts the error string matches our predicate. That's the only way to catch upstream drift. (Same shape as my banked Lesson #103 "test fixture ≠ production scenario".)

### O4 — `if controlServer != nil` at mount-process-state-failure path is dead code

At the new diff (mount.go:538-540 in PR), the check `if controlServer != nil { controlServer.Close() }` is unreachable-false: `controlServer` was just assigned from `startMountControlServer` at line 472 (and the `if err != nil` immediately after `startMountControlServer` returned early), so by the time we reach the mount-process-state-failure path, `controlServer` is guaranteed non-nil. Either remove the nil check or document why defensive (e.g., "future refactor may insert another startup step here"). Trivial — not blocking.

### O5 — `forceUnmount func(string)` returns no error

The `forceUnmount` signature has no error return. If `fusermount -u` fails (binary missing, permission denied, mountpoint already detached), the cleanup can't tell. This is acceptable for startup cleanup (we're already in a failure path; what would we do with an error?), but the test `TestCleanupMountStartFailureForceUnmountsWhenServerUnmountFails` only verifies forceUnmount was *called*, not that the mount actually got detached. Recommend: in a follow-up, change `forceUnmount` to return an error and log/escalate failures (e.g., metric for "startup cleanup couldn't detach mountpoint"). Not blocking.

---

## Verdict

Aligning with qiffang's **REVISE**: B1 + B2 are real and not addressable by inline tweaks. My O3 (test fixture binding) reinforces B1 with a concrete mitigation; O4 / O5 are stylistic.

**Suggested resolution sequence**:
1. Author justifies B2 (FlushAll + Unmount in WaitMount path) in code comments, OR splits WaitMount cleanup into a narrower struct.
2. Author adds integration test pinning `shouldForceUnmountAfterNewServerError` against real go-fuse output (O3 closes B1's drift risk).
3. Re-review.

qiffang:
## Review (adversary-1) — current head `4d30742`

Verdict from my lane: **REVISE**, but I do not have a new blocker beyond the already-raised B1/B2/O3 shape.

What I verified on the current head:

- The concrete `NewServer` cleanup gap from the earlier head is fixed now: `cleanupNewServerFailure(...)` stops the layer watcher, calls `FlushAll`, and force-unmounts only when the error is classified as a go-fuse INIT failure.
- The go-fuse coupling is real. In the pinned module currently used by this repo, `NewServer` does `mount(...)`, then `handleInit()`, then on INIT failure closes the mount fd and returns `fmt.Errorf("init: %s", code)` with an adjacent upstream `// TODO - unmount as well?`. So the compensating cleanup direction is correct.

Remaining issue 1: the `init:` predicate is still protected only by self-made fixture strings.

`TestCleanupNewServerFailureStopsFlushesAndForceUnmountsInitFailure` and `TestShouldForceUnmountAfterNewServerErrorOnlyForInitFailure` use `errors.New("init: ENODEV")`. That verifies our own predicate, not the upstream behavior the predicate depends on. If go-fuse later wraps or rewords the error, these tests will stay green while stale-mount cleanup silently regresses. I would add a production-shaped guard here: ideally a Linux integration test that triggers the go-fuse INIT failure path, or at least a focused guard tied to the pinned go-fuse source/version so a dependency bump forces this predicate to be rechecked.

Remaining issue 2: the `WaitMount` failure path cleanup expansion is plausible, but still needs the precondition documented in code.

I traced the pinned go-fuse `WaitMount()` path: it waits on `ms.ready` and then runs `pollHack(ms.mountPoint)`. By this point `NewServer` has already succeeded, `handleInit()` has called `Dat9FS.Init`, and `server.Serve()` is running in Drive9, so attempting `server.Unmount()` is reasonable and likely required for the stale mount class this PR fixes. The part that needs justification is `FlushAll` in this branch: it drains debouncers/queues, closes journal/shadow state, and waits `notifyWg` before unmounting. That may be safe here, but the safety argument is implicit. Please add a short comment on the `WaitMount` branch or split that branch into a narrower cleanup path so future changes do not accidentally inherit broader shutdown semantics without rechecking this startup state.

Validation I ran locally:

```text
go test ./pkg/fuse -run 'Test(MountShutdown|Cleanup|ShouldForceUnmount)' -count=1
go test -race ./pkg/fuse -run 'Test(MountShutdown|Cleanup|ShouldForceUnmount)' -count=1
go test ./pkg/fuse -count=1
```

All passed.

qiffang (approved):
## Adversarial Re-Review — PR #628 (v2)

**Verdict: GREEN** (all previous blockers resolved)

### Previous Blockers — Status

**B1 (fragile `init:` string match) — RESOLVED.** `shouldForceUnmountAfterNewServerError` now has a documentation comment identifying the exact go-fuse code path (`mornyx/go-fuse/v2 NewServer → mount() → handleInit() → fmt.Errorf("init: %s", code)`) and an explicit "re-check on upgrade" note.

**B2 (WaitMount failure path silently expanded) — RESOLVED.** The WaitMount failure path now has an explicit justification comment: "NewServer succeeded and Serve is running, so cleanup can drain Drive9-side workers before asking go-fuse to detach any partially mounted kernel state."

**@adversary-1's blocker (NewServer failure missing stopWatchers/flushAll) — RESOLVED.** `cleanupNewServerFailure` now passes `stopWatchers` and `flushAll` to `cleanupMountStartFailure`, so `dat9fs` resources (commit queue, write-back uploader, cache, journal, shadow store) are properly drained on `NewServer` failure.

### New Code Verified

- `forceUnmountWithoutServer` bool field cleanly separates the "no server handle" path from the normal unmount→force fallback
- `cleanupNewServerFailure` wrapper routes all `NewServer` failure logging through the `logf` abstraction (O1 from v1 resolved)
- `controlServer.Close()` on pid-file failure path preserved
- 7 tests (up from 3) covering: cleanup ordering, force unmount fallback, NewServer init failure with stopWatchers+flushAll+forceUnmount, NewServer non-init failure (no force), empty mountpoint guard, `shouldForceUnmountAfterNewServerError` predicate

**CI**: `ci` pass, `drive9-server-local e2e smoke` pass, CodeRabbit pass.

qiffang (approved):
## Adversarial Re-Review — PR #628 latest (`f8683dd`)

**Verdict: GREEN**

I re-reviewed the updated head after the prior cleanup blocker.

### Previous blockers

- **NewServer failure cleanup gap — resolved.** `gofuse.NewServer` failures now route through `cleanupNewServerFailure(...)`, which stops watchers and calls `dat9fs.FlushAll` before returning. For the go-fuse post-mount/pre-server INIT failure path, it also force-unmounts because no `Server` handle exists yet.
- **WaitMount cleanup expansion — resolved.** The WaitMount failure branch now documents why full cleanup is safe/intentional there: `NewServer` has succeeded and `Serve` is running, so Drive9-side workers can be drained before detaching any partially mounted kernel state.
- **`init:` predicate risk — acceptable after latest changes.** The predicate is still string-bound to go-fuse, but the code comment now pins the exact upstream path (`NewServer -> mount() -> handleInit() -> fmt.Errorf("init: %s", code)`) and explicitly requires re-checking on go-fuse upgrade. I inspected the currently pinned `github.com/mornyx/go-fuse/v2` source and that is the only post-mount `NewServer` failure path after `mount()` succeeds.

### Additional checks

- Startup failure paths now consistently use `cleanupMountStartFailure` for `WaitMount`, mount control socket creation, and process-state write failures.
- `controlServer.Close()` on pid-file failure is still called before shared cleanup; the later deferred close is safe because the control server close path tolerates repeated listener close / closed `done` receive / repeated socket removal.
- `forceUnmountWithoutServer` keeps the no-server INIT cleanup path separate from normal `server.Unmount()` fallback.
- I could not run Go tests in this agent environment because `go` is not installed; GitHub `ci`, `drive9-server-local e2e smoke`, and CodeRabbit are green.

No blocking findings remain from my lane.

PR #630, opened 2026-06-26
fix: start FUSE SSE watcher after WaitMount readiness

you06:
## Summary
- Start the Drive9 FUSE SSE watcher only after `server.WaitMount()` finishes go-fuse mount readiness checks.
- Add a WaitMount permission-error fallback: permission-like `WaitMount` errors (`EACCES`/`EPERM`) are tolerated, but only after a mountpoint readiness probe passes.
- Keep watcher cleanup safe when startup fails before the SSE watcher has been started.

## Why
Real E2B Drive9 FUSE runs showed an intermittent startup race: the initial SSE reset (`SSEWatcher.handleReset` invalidates **all** read/dir caches) could invalidate caches while go-fuse was still running its Linux `pollHack` readiness probe against `.go-fuse-epoll-hack`, causing `WaitMount()` to fail with `permission denied` even though the mountpoint was usable.

Delaying the SSE watcher until after `WaitMount()` removes that startup race at its root, while preserving the existing macOS/go-fuse requirement that `Serve()` is already running before `WaitMount()` proceeds.

The layer event watcher is **intentionally left running** during `WaitMount()`: it performs no initial blanket cache invalidation and its first poll only fires after a 1s tick on real remote events, so it cannot race `pollHack`.

## Permission fallback (new in this PR)
As defense-in-depth alongside the reorder, a permission-class `WaitMount` error is no longer always fatal. It is tolerated **only if** `probeMountPointReady` confirms the mount is usable:
- The probe is meaningful only because the FUSE mount syscall has already succeeded, so `stat`/`open`/`readdir` are served through go-fuse rather than the underlying directory.
- The probe runs under a 5s timeout so a wedged serve loop cannot hang startup on the fallback path.
- Any ignored error is logged to stderr for observability.

## Validation
- `go test ./pkg/fuse` (incl. new `serveWaitMountThenStartWatchers` / permission-error / probe unit tests)
- `go vet ./pkg/fuse`
- `git diff --check`
- E2B regression harness:
  - old startup order reproduced the race in 4/5 attempts
  - fixed branch passed 5/5 attempts, with `SSE reset` occurring after `mounted on`

### DIFF
--- a/pkg/fuse/active_mount_unix.go
+++ b/pkg/fuse/active_mount_unix.go
@@ -1,0 +1,34 @@
+//go:build !windows
+
+package fuse
+
+import (
+	"os"
+	"path/filepath"
+	"syscall"
+)
+
+func activeMountPoint(path string) (bool, error) {
+	abs, err := filepath.Abs(path)
+	if err != nil {
+		return false, err
+	}
+	info, err := os.Stat(abs)
+	if err != nil {
+		return false, err
+	}
+	parent := filepath.Dir(abs)
+	parentInfo, err := os.Stat(parent)
+	if err != nil {
+		return false, err
+	}
+	stat, ok := info.Sys().(*syscall.Stat_t)
+	if !ok {
+		return false, nil
+	}
+	parentStat, ok := parentInfo.Sys().(*syscall.Stat_t)
+	if !ok {
+		return false, nil
+	}
+	return stat.Dev != parentStat.Dev || stat.Ino == parentStat.Ino, nil
+}
--- a/pkg/fuse/active_mount_windows.go
+++ b/pkg/fuse/active_mount_windows.go
@@ -1,0 +1,8 @@
+//go:build windows
+
+package fuse
+
+func activeMountPoint(path string) (bool, error) {
+	_ = path
+	return false, nil
+}
--- a/pkg/fuse/mount.go
+++ b/pkg/fuse/mount.go
@@ -454,6 +454,10 @@ func Mount(opts *MountOptions) error {
 	// Start SSE watcher after WaitMount. The initial stream reset invalidates
 	// kernel/user-space caches; if it races go-fuse's WaitMount pollHack open
 	// on Linux, the kernel can reject .go-fuse-epoll-hack with EACCES/EPERM.
+	//
+	// The layer event watcher (started earlier) is deliberately left running:
+	// it does no initial blanket cache invalidation and its first poll only
+	// fires after a 1s tick on real remote events, so it cannot race pollHack.
 	var sseWatcher *SSEWatcher
 	stopWatchers := func() {
 		if sseWatcher != nil {
@@ -465,16 +469,16 @@ func Mount(opts *MountOptions) error {
 	err = serveWaitMountThenStartWatchers(server.Serve, server.WaitMount, func() {
 		sseWatcher = StartSSEWatcher(dat9fs, c, actorID)
 	}, func(err error) error {
-		if ok, probeErr := shouldContinueAfterWaitMountPermissionError(err, opts.MountPoint, probeMountPointReady); ok {
+		ok, probeErr := shouldContinueAfterWaitMountPermissionError(err, opts.MountPoint, probeMountPointReady)
+		if ok {
 			fmt.Fprintf(os.Stderr, "drive9: fuse wait mount permission error ignored after readiness probe passed at %s: %v\n", opts.MountPoint, err)
 			return nil
-		} else {
-			if probeErr != nil {
-				fmt.Fprintf(os.Stderr, "drive9: fuse wait mount permission fallback probe failed at %s: %v\n", opts.MountPoint, probeErr)
-			}
-			stopWatchers()
-			return fmt.Errorf("fuse wait mount: %w", err)
 		}
+		if probeErr != nil {
+			fmt.Fprintf(os.Stderr, "drive9: fuse wait mount permission fallback probe failed at %s: %v\n", opts.MountPoint, probeErr)
+		}
+		stopWatchers()
+		return fmt.Errorf("fuse wait mount: %w", err)
 	})
 	if err != nil {
 		return err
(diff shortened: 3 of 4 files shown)


qiffang (commented):
## Review: PR #630 — start FUSE SSE watcher after WaitMount readiness

### Overall Assessment

Well-structured fix for a real race condition. The root cause is clear: `StartSSEWatcher` calls `handleReset` → invalidates all caches → can race go-fuse's `pollHack` open on `.go-fuse-epoll-hack` during `WaitMount()`, causing spurious `EACCES`/`EPERM`. Moving SSE start after `WaitMount` eliminates the race at its root.

### Observations

#### 1. `markStatCacheUnverified()` is now called before WaitMount (line 462)

The PR moves `StartSSEWatcher` after `WaitMount`, but adds `dat9fs.markStatCacheUnverified()` at line 462 (mount.go) to compensate. This is correct — without the SSE watcher running during WaitMount, the cache state is unverified. However, `StartSSEWatcher` also calls `markStatCacheUnverified()` at entry (sse.go:26). So after WaitMount, when `StartSSEWatcher` is finally called, it marks unverified again (redundant but harmless).

#### 2. Permission fallback is Linux-only — correct

`shouldContinueAfterWaitMountPermissionError` checks `runtime.GOOS != "linux"` and returns false for non-Linux. This is correct because the `pollHack` race is Linux-specific (macOS WaitMount has a different readiness mechanism via mount_macfuse).

#### 3. `probeMountPointReady` timeout goroutine leak

`probeMountPointReady` spawns a goroutine and selects on `done` vs `time.After`. If the probe times out, the goroutine running `probeMountPointReadyOnce` continues running — it's doing `os.Stat` + `os.Open` + `Readdirnames` which could block indefinitely on a wedged FUSE. The goroutine is orphaned. This is acceptable for a startup probe (the process will likely exit or the mount will recover), but worth noting.

#### 4. `activeMountPoint` — device comparison logic

```go
return stat.Dev != parentStat.Dev || stat.Ino == parentStat.Ino, nil
```

The `stat.Ino == parentStat.Ino` case handles bind mounts or root mount detection. For FUSE, `stat.Dev != parentStat.Dev` is the expected signal. This is a standard mount detection pattern — correct.

#### 5. `stopWatchers` nil check is necessary

```go
stopWatchers := func() {
    if sseWatcher != nil {
        sseWatcher.Stop()
    }
    layerEventWatcherStop()
}
```

Since `sseWatcher` is now initialized after `WaitMount`, error paths between `NewServer` and `WaitMount` would call `stopWatchers` with `sseWatcher == nil`. The nil check is correct and necessary.

#### 6. Merge conflict with PR #628

PR #628 also modifies the `WaitMount` error path in `mount.go`, wrapping it with `cleanupMountStartFailure`. These two PRs will conflict. The resolution is straightforward: PR #630's `serveWaitMountThenStartWatchers` replaces the inline `WaitMount` call, and PR #628's cleanup should be integrated into the `handleWaitMountError` callback.

#### 7. Test coverage is thorough

- `TestServeWaitMountThenStartWatchers*` (3 tests) — ordering, accepted error, rejected error
- `TestShouldContinueAfterWaitMountPermissionError*` (3 tests) — probe passes, probe fails, non-permission error
- `TestProbeMountPointReady*` (4 tests) — usable dir, local dir, missing, empty
- `TestIsWaitMountPermissionError*` — wrapped and text errors

Good edge case coverage. The `osPathError` helper in tests correctly simulates wrapped errno.

### Summary

LGTM. Clean separation: `serveWaitMountThenStartWatchers` encapsulates the ordering contract, permission fallback is defense-in-depth with proper gating (Linux-only + active mount probe). Minor note on the orphaned timeout goroutine, but acceptable for a startup path.

qiffang (approved):
## Adversarial Review — PR #630

**Verdict: GREEN** (0 blockers, 2 observations)

### Checklist

| Gate | Status |
|------|--------|
| Edge cases covered? | PASS |
| API explicit? | PASS |
| Race / state issues? | PASS |

---

### Analysis

The fix addresses two issues:

1. **SSE watcher reorder**: Moving `StartSSEWatcher` after `WaitMount` eliminates the root cause — SSE reset can no longer race go-fuse's `pollHack` readiness probe. The `markStatCacheUnverified()` call before `WaitMount` correctly compensates for the delayed watcher start.

2. **Permission fallback**: EACCES/EPERM from `WaitMount` is tolerated only when `probeMountPointReady` confirms the mount is usable. The probe is correctly gated behind `runtime.GOOS == "linux"`, bounded by a 5s timeout, and checks active mount status (device boundary crossing) before accepting.

Key correctness points verified:
- `stopWatchers` nil-checks `sseWatcher` for early cleanup paths where watcher hasn't started yet
- `errors.Is` check for EPERM/EACCES comes first; string matching is documented fallback for go-fuse paths that don't wrap errno
- `activeMountPoint` uses standard POSIX `stat.Dev` cross-device detection
- Watcher ordering enforced: `serveWaitMountThenStartWatchers` only calls `startWatchers` after successful `WaitMount` (or accepted error)
- Layer event watcher intentionally left running during `WaitMount` — documented as safe (no initial blanket invalidation, first poll ≥1s)

---

### O1 — `probeMountPointReady` goroutine leaks on timeout

If the probe goroutine hangs (wedged FUSE serve loop), the goroutine is never cleaned up after the 5s timeout. Acceptable because a wedged serve loop means the process is likely to exit, but `context.WithTimeout` + select on `ctx.Done()` inside the probe would be cleaner.

### O2 — `WaitMount` failure path doesn't call `server.Unmount()`

When `handleWaitMountError` returns an error (non-permission or probe failure), `serve()` is still running in a background goroutine but no `Unmount` is called. This preserves existing behavior (original code also only called `stopWatchers()` after `WaitMount` failure), so it's not a regression — but it's a pre-existing gap (same concern as PR #628 B2).

---

**CI**: `ci` pass, `drive9-server-local e2e smoke` pass, CodeRabbit pass on head `16c1678f`.

**Tests**: 12 new tests in `mount_wait_test.go` covering probe success/failure, ordering guarantees, permission error recognition, and non-permission error rejection. Well-structured with injectable `isActiveMountPoint` for testability.

**E2B validation**: PR description reports 0/5 → 5/5 pass rate after fix.

qiffang (changes requested):
## Adversarial Review — PR #630

**Verdict: REQUEST_CHANGES**

The runtime fix direction looks correct: delaying `StartSSEWatcher` until after `WaitMount` removes the cache-invalidation vs go-fuse `pollHack` race, and the permission fallback is appropriately gated by a positive mountpoint probe. The latest head also fixes the earlier `io.EOF` false-positive concern by requiring `activeMountPoint` before accepting an empty directory.

### Blocking issue

`pkg/fuse/mount_wait_test.go` adds Linux-only expectation tests without gating them to Linux. `shouldContinueAfterWaitMountPermissionError` explicitly returns `false, nil` when `runtime.GOOS != "linux"` (`pkg/fuse/mount.go:704`), but these tests are compiled/run on every platform:

- `TestShouldContinueAfterWaitMountPermissionErrorWhenProbePasses` expects `ok == true` and the probe to be called.
- `TestShouldContinueAfterWaitMountPermissionErrorWhenProbeFails` expects the probe error to be returned.

On macOS/Windows both expectations are false because the function exits before probing. This repository has macOS FUSE code paths and release targets, while CI is Linux-only, so the current green CI will not catch that `go test ./pkg/fuse` is broken for non-Linux developers/runners.

Required fix: make these tests platform-aware (for example, split/inject the platform decision or skip/tag the Linux-specific cases) and add explicit coverage for the non-Linux branch so the intended Linux-only fallback remains documented.

### Validation

- I could not run `go test` locally because this agent environment has no `go` binary; the failure above is from static control-flow review of the new test expectations against the `runtime.GOOS` guard.
- GitHub checks are green on Linux: `ci`, `drive9-server-local e2e smoke`, CodeRabbit.

PR #631, opened 2026-06-26
fix: use tls=skip-verify for native private endpoint connections

srstack:
## Problem

When using private endpoint, TLS fails with x509 error because the cert is for the public hostname.

## Fix

Use tls=skip-verify in ensureDatabase and tenantDSN when DRIVE9_TIDBCLOUD_NATIVE_USE_PRIVATE_ENDPOINT is set.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved TLS connection handling for database access in private endpoint setups.
  * Updated connection behavior so secure connections work more reliably across standard and private network configurations.

### DIFF
--- a/cmd/drive9-server/main.go
+++ b/cmd/drive9-server/main.go
@@ -213,6 +213,14 @@ func main() {
 		provisioner, provisionerErr = db9.NewProvisionerFromEnv()
 	}
 	if provisionerErr != nil {
+		isWanted := false
+		if providerType == tenant.ProviderTiDBCloudNative && os.Getenv("DRIVE9_TIDBCLOUD_NATIVE_API_URL") != "" {
+			isWanted = true
+		}
+		if isWanted {
+			logger.Error(context.Background(), "provisioner_failed", zap.String("provider", providerType), zap.Error(provisionerErr))
+			os.Exit(1)
+		}
 		logger.Warn(context.Background(), "provisioner_not_configured", zap.String("provider", providerType), zap.Error(provisionerErr))
 	} else {
 		logger.Info(context.Background(), "provisioner_configured", zap.String("provider", providerType))
--- a/pkg/server/fork.go
+++ b/pkg/server/fork.go
@@ -656,7 +656,7 @@ func (s *Server) ensureForkBranchConnection(ctx context.Context, forkTenant, sou
 			return "", err
 		}
 		if forkTenant.DBHost != "" && forkTenant.DBPort != 0 && forkTenant.DBUser != "" {
-			return tenantDSN(forkTenant.DBUser, string(plain), forkTenant.DBHost, forkTenant.DBPort, forkTenant.DBName, forkTenant.DBTLS), nil
+			return tenantDSN(forkTenant.DBUser, string(plain), forkTenant.DBHost, forkTenant.DBPort, forkTenant.DBName, forkTenant.DBTLS, forkTenant.Provider), nil
 		}
 		if branchProv, ok := s.provisioner.(tenant.CredentialBranchProvisioner); ok && credentialReq != nil {
 			username, err := branchProv.WaitForBranchUserWithCredentials(ctx, forkTenant.ClusterID, forkTenant.BranchID, *credentialReq)
@@ -687,7 +687,7 @@ func (s *Server) ensureForkBranchConnection(ctx context.Context, forkTenant, sou
 			}); err != nil {
 				return "", err
 			}
-			return tenantDSN(username, string(plain), host, port, forkTenant.DBName, true), nil
+			return tenantDSN(username, string(plain), host, port, forkTenant.DBName, true, forkTenant.Provider), nil
 		}
 		cluster, err := s.waitForForkBranchActive(ctx, &tenant.ClusterInfo{
 			TenantID:  forkTenant.ID,
@@ -715,7 +715,7 @@ func (s *Server) ensureForkBranchConnection(ctx context.Context, forkTenant, sou
 		}); err != nil {
 			return "", err
 		}
-		return tenantDSN(cluster.Username, string(plain), cluster.Host, cluster.Port, cluster.DBName, true), nil
+		return tenantDSN(cluster.Username, string(plain), cluster.Host, cluster.Port, cluster.DBName, true, cluster.Provider), nil
 	}
 
 	branchPassword, err := s.pool.Decrypt(ctx, forkTenant.DBPasswordCipher)
@@ -754,7 +754,7 @@ func (s *Server) ensureForkBranchConnection(ctx context.Context, forkTenant, sou
 	if cluster.Host == "" || cluster.Port == 0 || cluster.Username == "" {
 		return "", forkErr(http.StatusServiceUnavailable, "database branch is not active yet")
 	}
-	return tenantDSN(cluster.Username, string(branchPassword), cluster.Host, cluster.Port, cluster.DBName, true), nil
+	return tenantDSN(cluster.Username, string(branchPassword), cluster.Host, cluster.Port, cluster.DBName, true, cluster.Provider), nil
 }
 
 func generateForkDBPassword(length int) (string, error) {
@@ -1029,7 +1029,7 @@ func (s *Server) openTenantStore(ctx context.Context, t *meta.Tenant) (*datastor
 	if err != nil {
 		return nil, err
 	}
-	return datastore.Open(tenantDSN(t.DBUser, string(plain), t.DBHost, t.DBPort, t.DBName, t.DBTLS))
+	return datastore.Open(tenantDSN(t.DBUser, string(plain), t.DBHost, t.DBPort, t.DBName, t.DBTLS, t.Provider))
 }
 
 func (s *Server) enqueueForkFileGCTaskRefs(ctx context.Context, t *meta.Tenant, store *datastore.Store) error {
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -745,7 +745,7 @@ func (s *Server) startTenantSchemaInitResume(ctx context.Context, t meta.Tenant)
 			logger.Warn(workerCtx, "resume_schema_init_skipped", zap.String("tenant_id", t.ID), zap.Error(err))
 			return
 		}
-		dsn := tenantDSN(t.DBUser, string(plain), t.DBHost, t.DBPort, t.DBName, t.DBTLS)
+		dsn := tenantDSN(t.DBUser, string(plain), t.DBHost, t.DBPort, t.DBName, t.DBTLS, t.Provider)
 		s.initTenantSchemaAsync(workerCtx, t.ID, dsn, t.Provider, s.schemaInitForTenant(t.ID, t.Provider, s.provisioner.InitSchema))
 	})
 }
@@ -824,10 +824,12 @@ func contextWithTrace(parent, traceSource context.Context) context.Context {
 	return traceid.With(parent, traceID)
 }
 
-func tenantDSN(user, password, host string, port int, dbName string, tlsEnabled bool) string {
+func tenantDSN(user, password, host string, port int, dbName string, tlsEnabled bool, provider string) string {
 	query := "parseTime=true"
 	if tlsEnabled {
 		query += "&tls=true"
+	} else if provider == tenant.ProviderTiDBCloudNative {
+		query += "&tls=skip-verify"
 	}
 	return fmt.Sprintf("%s:%s@tcp(%s:%d)/%s?%s", user, password, host, port, dbName, query)
 }
@@ -4359,7 +4361,7 @@ func (s *Server) provisionTenant(ctx context.Context, opts provisionTenantOption
 		APIKeyID:       apiKeyID,
 		Status:         meta.TenantProvisioning,
 		Provider:       provider,
-		TenantDSN:      tenantDSN(cluster.Username, cluster.Password, cluster.Host, cluster.Port, cluster.DBName, dbtls),
+		TenantDSN:      tenantDSN(cluster.Username, cluster.Password, cluster.Host, cluster.Port, cluster.DBName, dbtls, provider),
 		CloudProvider:  cloudProvider,
 		Region:         region,
 		OrganizationID: strings.TrimSpace(cluster.OrganizationID),
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -827,18 +827,11 @@ func contextWithTrace(parent, traceSource context.Context) context.Context {
 func tenantDSN(user, password, host string, port int, dbName string, tlsEnabled bool) string {
 	query := "parseTime=true"
 	if tlsEnabled {
-		tlsMode := "true"
-		if isPrivateEndpointEnabled() {
-			tlsMode = "skip-verify"
-		}
-		query += "&tls=" + tlsMode
+		query += "&tls=true"
+	} else {
+		query += "&tls=skip-verify"
 	}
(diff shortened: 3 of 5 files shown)


srstack (commented):
Reviewed in depth. The root problem (private-endpoint host vs public-hostname cert → x509 failure) is real and the `main.go` hard-fail is a good hardening, but there are three issues I'd want fixed before merge.

**🔴 1. `skip-verify` is broader than the problem.** It disables both chain *and* hostname verification (full MITM exposure), but the actual issue is only a hostname mismatch. The standard secure fix is to register a custom TLS config overriding `ServerName` to the public hostname (with system RootCAs) and keep chain verification:
```go
mysql.RegisterTLSConfig("tidb-private", &tls.Config{ServerName: publicHostname, RootCAs: pool})
// DSN: ...?tls=tidb-private
```
Blanket `skip-verify` is defensible for VPC-confined PrivateLink traffic, but if kept it should be tightly scoped (see #2/#3) and **log a warning when active** so operators know verification is off.

**🔴 2. Inconsistent gating — `=false` still skips verification.** `parseBoolEnv` (used by the provisioner) correctly treats `0/false/no` as false, but `tenantDSN` (server.go) and `pool.go` gate on `os.Getenv(...) != ""`, which treats `"false"`/`"0"`/`"no"` as truthy. So:
```
DRIVE9_TIDBCLOUD_NATIVE_USE_PRIVATE_ENDPOINT=false
  → provisioner: public endpoint + tls=true   (feature off, correct)
  → tenantDSN / pool.go: tls=skip-verify       (WRONG — verification disabled)
```
All three sites should use the same `parseBoolEnv` semantics, not a raw emptiness check.

**🔴 3. `tenantDSN` weakens TLS for *all* providers, not just native.** `pool.go` correctly scopes to `t.Provider == ProviderTiDBCloudNative`, but `tenantDSN` reads the env with no provider check — and it's called for every provider (fork.go ×4, admin_tenant_pool.go ×2, server.go:748 with generic `t.DBTLS`). So enabling the native flag silently downgrades cert verification for every TLS tenant connection built via `tenantDSN`, including non-native tenants in a mixed deployment. `tenantDSN` has no provider arg, but all callers do — thread a `tlsMode`/provider param in and compute `skip-verify` only for native + feature-on, mirroring `pool.go`. Don't read the env inside `tenantDSN`.

**🟢 4. `main.go` hard-fail — good.** Failing fast when native provisioning is explicitly configured but init failed is right. Minor: `isWanted` can collapse into a single `if`; and it only hard-fails for native (other providers still warn-and-continue) — confirm that's intended scope.

**Tests:** none added. `tenantDSN`/`createBackend` TLS-mode selection are pure string builders — a unit test (=true → skip-verify, =false/empty → true, non-native → true regardless) is cheap and would catch #2/#3 directly.

Net: request changes on #1–#3 (at least make the gating consistent + provider-scoped, and log/justify skip-verify); #4 is fine. The hostname-mismatch case is better solved with a `ServerName`-override TLS config than blanket skip-verify.

srstack (commented):
## Re-review

The approach changed substantially — `skip-verify` is abandoned, and private-endpoint native tenants now use **plaintext** (`DBTLS=false`) decided at provision time. This resolves two of my three original findings; one concern remains plus a functional risk.

### Resolved ✅
- **#2 (=false still skipped verification):** Fixed — the gate is now `v == "1" || "true" || "yes"`, so `=false`/empty correctly keeps TLS on.
- **#3 (tenantDSN downgraded all providers):** Fixed — the decision moved to `provisionTenant`, gated on `provider == ProviderTiDBCloudNative`, and persisted per-tenant as `DBTLS`. `tenantDSN`/`pool.go` are reverted to read `t.DBTLS` only, so non-native tenants are untouched. This is a cleaner design: TLS mode and endpoint host are now both decided at provision time and stored together, rather than read from env on every connection.
- Provision and runtime are now **internally consistent**: `ensureDatabase` (already on base) and the tenant runtime DSN both use plaintext for private endpoints.

### 🟠 Remaining concern: plaintext is weaker than the abandoned skip-verify
`DBTLS=false` disables encryption entirely, which is *strictly weaker* than `skip-verify` (which at least encrypted the link). Credentials and data now travel in cleartext on the private network. For PrivateLink/PSC traffic confined to the VPC this is a defensible tradeoff, but it's a real downgrade and should be an explicit, documented decision — not an implicit side effect of the endpoint flag. The encryption-preserving fix remains a **`ServerName`-override TLS config** (register a `*tls.Config{ServerName: publicHost, RootCAs: pool}`, DSN `tls=<name>`): it dials the private host but verifies the cert against its real public SAN — keeping the connection encrypted *and* validated. I'd strongly prefer that over plaintext if feasible.

### 🔴 Functional risk to verify before merge
TiDB Cloud Serverless generally **enforces TLS** on connections. If the private endpoint also requires TLS, then `TLSConfig="false"` / `DBTLS=false` will fail to connect outright (both at `ensureDatabase` provisioning and at runtime). Please confirm against a live private endpoint that the cluster actually accepts non-TLS connections — otherwise this swaps an x509 error for a connection refusal. (This ties into the still-open #626 question of whether private endpoints are even returned/usable synchronously.)

### Minor
- The boolean parse is inlined in `server.go` rather than reusing `parseBoolEnv` (which is unexported in the tidbcloudnative pkg). Two copies can drift — consider a shared exported helper.
- Still no tests; the `provisionTenant` `DBTLS` selection (native+flag → false, else true) is a cheap unit test that would lock in the #2/#3 fixes.
- Rollout note: existing tenants keep their stored `DBTLS=true`; only newly-provisioned private-endpoint tenants get `DBTLS=false`. Consistent with Host being fixed at provision time, but worth being explicit about for any already-provisioned private-endpoint tenants.

Net: #2/#3 are properly fixed and the design is now consistent and correctly scoped. Before merge I'd want the **TiDB-accepts-plaintext** question answered (functional blocker), and ideally the plaintext approach replaced with — or at least justified against — a `ServerName`-override TLS config that preserves encryption.

srstack (commented):
## Re-review (skip-verify via DBTLS=false)

The approach converged on: native private-endpoint tenants are provisioned with `DBTLS=false`, and `tenantDSN`/`pool.go`/`ensureDatabase` now emit `tls=skip-verify` for that case. This **resolves last round's functional blocker** — skip-verify keeps the link encrypted, so TiDB's TLS enforcement is satisfied (unlike the plaintext attempt). #2 (proper boolean) and #3 (provider-scoped at provision time) remain fixed. Good progress.

But the `simplify: DBTLS=false always means tls=skip-verify` commit introduces a new problem:

### 🔴 Overloading `DBTLS=false` globally redefines it from "no TLS" to "skip-verify" for every tenant/provider
```go
if tlsEnabled { query += "&tls=true" } else { query += "&tls=skip-verify" }
```
This `else` runs in `tenantDSN` (server.go) and `pool.go createBackend`, both of which are called with `t.DBTLS` for **all** tenants (server.go:748, fork.go:659/1032, pool.go:667) — not just native private-endpoint ones. Consequences:
1. Previously `DBTLS=false` meant *plaintext* (no `&tls=` param). Now it forces a TLS handshake (`skip-verify`). Against a target that genuinely has TLS disabled (a plain MySQL / self-hosted / dev DB), the driver attempts TLS, the server refuses, and the connection **fails** — a regression for any real `DBTLS=false` tenant.
2. There is now **no way to express "connect with no TLS"** anywhere in the system — the plaintext option is gone.
3. A connection-security boolean is silently overloaded: `false` no longer means what it says, and the meaning is set far away (provision-time env) from where it's interpreted (DSN build).

Today the only producer of `DBTLS=false` is the new native path, so it may not break a live tenant *right now* — but it removes plaintext capability for the whole system and is a latent foot-gun for any future/external tenant whose DB doesn't speak TLS. The skip-verify behavior should be **scoped**, not bolted onto the `DBTLS` boolean. Suggested directions:
- Add a dedicated mode/field (e.g. `DBTLSMode: verify | skip-verify | disabled`, or a `DBTLSSkipVerify bool` alongside `DBTLS`), so "off" and "on-but-unverified" stay distinct; or
- Keep a native-private-specific DSN path that selects skip-verify, leaving the generic `else` as plaintext.

### 🟡 Test impact — please run the suites
Several fixtures set `DBTLS: false` (auth_test, fork_test, leader_lifecycle_test, object_gc_worker_test, provision_test). Any that actually open a connection to a non-TLS MySQL fixture will now get `tls=skip-verify` and may fail. Worth a `go test ./pkg/server ./pkg/tenant ./pkg/meta` before merge.

### Still standing
- The ideal remains a **`ServerName`-override TLS config** (verified against the public SAN while dialing the private host) — skip-verify is encrypted but unverified, i.e. still MITM-exposed within the private network. Acceptable for PrivateLink as a documented tradeoff, but worth a code comment stating that explicitly.
- Minor: the `1/true/yes` parse is still duplicated inline in server.go; `isWanted` in main.go can collapse to one `if`; no unit test yet for the `provisionTenant` DBTLS selection.

Net: the encryption/TiDB-enforcement concern is resolved, but the global `DBTLS=false → skip-verify` overload (🔴) should be scoped to native private-endpoint instead of redefining the boolean for all tenants. Run the DBTLS=false tests to confirm no fixture regressed.

srstack:
The code you reviewed is outdated — commit 683c36d already fixed this:

- **pool.go**: `DBTLS=false` → `skip-verify` ONLY when `t.Provider == ProviderTiDBCloudNative`. Non-native tenants with `DBTLS=false` get plaintext (no `&tls=...` at all). No global redefinition.
- **tenantDSN**: Completely reverted to original — only `tls=true` when `tlsEnabled=true`. No `skip-verify` logic.
- **ensureDatabase**: uses `parseBoolEnv` (proper boolean check), only in the tidbcloudnative package.

Current state: three paths — pool.go (native guard), ensureDatabase (parseBoolEnv guard), tenantDSN (unchanged). Please re-review the latest commit.

srstack (commented):
## Re-review (scope skip-verify to native)

The `pool.go` change is now correct: `DBTLS=false` → `skip-verify` **only** for `ProviderTiDBCloudNative`, and non-native `DBTLS=false` keeps plaintext (no `&tls=`). That **resolves the global-overload 🔴** from the previous round — non-native tenants are no longer affected. 👍

But reverting `tenantDSN` entirely (no skip-verify) introduces a new inconsistency: **`pool.go` and `tenantDSN` now disagree on how a native `DBTLS=false` tenant connects.**

### 🔴 Native private-endpoint tenants get plaintext on the `tenantDSN` paths → will fail against TiDB's TLS requirement
`tenantDSN` only knows `tlsEnabled bool`; with the revert, `DBTLS=false` → no `&tls=` → **plaintext**. And `tenantDSN` is used to build native-tenant connections in several control-plane paths:
- **server.go:4359** — the stored `TenantDSN` is built with `tenantDSN(..., dbtls)`; for native-private `dbtls=false` → a **plaintext** DSN, which is then opened by `initTenantSchemaAsync` (consumed at server.go:4439/4444). So the **initial schema-init of a freshly provisioned native-private tenant connects in plaintext.**
- **server.go:748** (`startTenantSchemaInitResume`) — schema-init resume, same plaintext DSN for native-private.
- **fork.go:659** (`ensureForkBranchConnection`) and **fork.go:1032** (`openTenantStore`) — fork/branch connections, same.

So for a native private-endpoint tenant the **data plane** (`pool.go createBackend`) uses `skip-verify` (encrypted) while **provisioning/schema-init/fork** (`tenantDSN`) use plaintext. If TiDB rejects non-TLS on the private endpoint — which is the same TLS-enforcement question from two rounds ago — those plaintext paths fail, and the very provisioning flow this PR targets breaks at schema init. Even if TiDB accepted plaintext, the same tenant connecting via TLS on one path and plaintext on another is incoherent.

### Root cause: a single `DBTLS` bool can't encode 3 states across 2 builders
The repeated thrashing (global skip-verify → plaintext → global skip-verify → pool-only) is because `verify` / `skip-verify` / `none` is three states, but `DBTLS` is one bool, and there are two DSN builders. Two clean ways out:
1. **Thread provider (or a `tlsMode`) into `tenantDSN`**, mirroring `pool.go`'s `else if t.Provider == ProviderTiDBCloudNative`. Most callers already hold `t.Provider`/`forkTenant.Provider`/`provider`. Then both builders agree.
2. **Introduce an explicit `DBTLSMode` (or `DBTLSSkipVerify bool`) on the tenant** and consult it in both `pool.go` and `tenantDSN`, so "off" and "on-but-unverified" stay distinct and no path is left guessing from a bool.

I'd prefer (1) for this PR's size, or (2) if you want the durable model. Either way, `pool.go` and `tenantDSN` must produce the **same** TLS mode for the same tenant.

### Unchanged notes
- `ensureDatabase` skip-verify (provision-time) is consistent with the data plane — good.
- Still worth confirming TiDB actually accepts `skip-verify` (encrypted, unverified) on the private endpoint, and a code comment justifying skip-verify over a `ServerName`-override verified config.
- Run `go test ./pkg/server ./pkg/tenant` — fork/schema-init paths for native + `DBTLS=false` are the ones to watch.

Net: `pool.go` scoping is right; before merge, make `tenantDSN` agree with it for native `DBTLS=false` (thread provider in, or add an explicit TLS-mode field) so provisioning/schema-init/fork don't silently fall back to plaintext.

srstack (commented):
## Re-review (thread provider into tenantDSN)

This resolves the inconsistency from the last round. `tenantDSN` now takes `provider` and uses the **same** rule as `pool.go`:
```
DBTLS=true            → tls=true
DBTLS=false + native  → tls=skip-verify
DBTLS=false + other   → plaintext (original behavior)
```
I verified:
- ✅ **All 7 `tenantDSN` callers updated** to pass provider (fork.go ×5, server.go schema-init-resume + stored TenantDSN). No test callers. **Built the changed packages from the PR head in a clean worktree — `go build ./pkg/server/... ./pkg/tenant/... ./cmd/drive9-server/...` succeeds**, so the 6→7 arg signature change is complete (no stragglers).
- ✅ `tenantDSN` and `pool.go` now produce identical TLS modes for the same tenant, and `ensureDatabase` (provision-time) matches on the native-private path. So provisioning, schema-init (initial + resume), fork/branch, and the data plane all agree — the earlier split-brain (skip-verify vs plaintext) is gone.
- ✅ Earlier findings (#2 proper boolean, #3 provider-scoped, global-overload) all remain resolved.

The TLS handling is now correct and consistent across every path. Remaining items are verification/quality, not correctness:

- **Confirm in deployment** that TiDB Cloud accepts `tls=skip-verify` (encrypted, unverified) on the private endpoint. skip-verify keeps TLS so it should satisfy TiDB's TLS enforcement — but this whole path is still unvalidated against a live private endpoint (ties into the open #626 question of whether the private endpoint is even returned/usable). This is the one thing I'd want smoke-tested before relying on it in prod.
- **Add a code comment** on the `else if … native { skip-verify }` branches noting *why* verification is skipped (cert is issued for the public host; traffic is confined to PrivateLink). Ideally a `ServerName`-override config would keep verification, but skip-verify is a fine documented tradeoff for private-link traffic.
- **Add a unit test** for the TLS-mode matrix — given 11 iterations on this exact logic, a table test over (`tlsEnabled` × provider) → expected `tls=` for both `tenantDSN` and the pool query would lock it down cheaply and prevent a 12th regression. These are pure string builders, so no DB needed.
- Minor: `isWanted` in main.go can collapse to a single `if`.

LGTM on the code — the core TLS-consistency concern is fully addressed and it builds. I'd just want the live-endpoint smoke test + a comment/unit test before merge. (Can't formally approve since this is a srstack-authored PR.)

PR #633, opened 2026-06-26
fix(metrics): exclude SSE lifetime from HTTP duration hist; add events diagnostics + multi-tenant fs_events cleanup

mornyx:
## Problem

The `/v1/events` endpoint triggered "tens-of-seconds" latency alerts. Investigation confirmed this is a **measurement artifact**, not a real latency regression.

The `observe` middleware records `time.Since(start)` into `drive9_http_request_duration_seconds` only after the handler returns. But `handleEvents` is an SSE long-lived connection — after the Phase-1 replay it enters a `select` loop and blocks as long as the client stays subscribed. So the SSE connection lifetime was recorded as an HTTP request duration: a connection held open for 40s was logged as a 40s "request", falling into the 10s `+Inf` histogram bucket — exactly the alert band.

**Decisive dev experiment**: hold an SSE connection open for 15s → `drive9_http_request_duration_seconds{route="/v1/events"}` `+Inf` bucket +1, sum +14.96s — proving the alert is SSE connection lifetime leaking into the HTTP duration histogram, not a real latency regression. This artifact has existed since SSE shipped (4824b60, 2026-04-14), over two months before the multi-pod refactor — it was not introduced by the refactor.

### Fix A — Exclude SSE lifetime from the HTTP duration histogram
`pkg/server/instrumentation.go`: for route `/v1/events`, record only the request counter and inflight gauge, **skipping** the `drive9_http_request_duration_seconds` histogram (added `RecordHTTPRequestCount`/`RecordTenantRequestCount`). SSE connection lifetime now goes into dedicated metrics. **Dev verification**: after the fix, holding an SSE connection for 15s no longer changes the HTTP duration histogram count/+Inf/sum — the alert will no longer fire.

### Fix B — SSE / event-bus / fs_events diagnostic metrics + logs
Compensates for the lack of direct TiDB access — everything is observable from `/metrics` and logs:

- `drive9_sse_connection_duration_seconds` (connection lifetime; buckets include 60s/300s/600s/1800s/3600s)
- `drive9_sse_inflight_connections{tenant_id}` (active SSE connections)
- `drive9_sse_phase1_duration_seconds{tenant_id}` (Phase-1 replay/reset cost — **key for distinguishing "slow first replay" from "Phase-2 client holding open"**)
- `drive9_sse_events_sent_total{op,tenant_id}` / `drive9_sse_resets_sent_total{reason,tenant_id}` / `drive9_sse_heartbeats_sent_total{tenant_id}`
- `drive9_event_bus_query_duration_seconds{operation=events_since|poll|latest|oldest, result=ok|error, tenant_id}` (all fs_events DB query latency, success included)
- `drive9_event_bus_poll_failures_total{tenant_id}` / `drive9_event_bus_publish_errors_total{tenant_id}`
- `drive9_fs_events_rows{tenant_id}` / `drive9_fs_events_pruned_total{tenant_id}` (table size + prune volume, no direct DB access needed)
- `event_bus_cursor_gap` log with oldest/latest/since + query_ms (log-side table-growth detection)

Adds `Store.CountFSEvents` (`pkg/datastore/fs_events.go`).

### Multi-tenant fs_events cleanup (issue #599 P1-2)
`cleanupFSEvents` in `pkg/server/server.go` previously ran only in single-tenant mode; in multi-tenant mode the table grew unbounded. Adds a multi-tenant branch using `pool.ActiveTenantStores()` (new in `pkg/tenant/pool.go`, mirroring `StartAllFileGC`'s cache-snapshot pattern) to iterate tenants with a **cached backend** (only these accumulate fs_events rows), calling `CountFSEvents`/`DeleteFSEventsBefore` per tenant and reporting row count + prune volume.

> Not using `ListTenantsByStatus + pool.Acquire`: that forces `createBackend` for every listed tenant, exhausting connections and failing for dormant tenants without materialized credentials (observed on dev: 205 acquire errors). `ActiveTenantStores` snapshots only cached backends — zero extra connection overhead. **Dev verification**: `drive9_fs_events_rows{tenant_id="d03e99aa..."} 8` now reported per tenant, cleanup tick working.

## Dev verification results

| Item | Result |
|---|---|
| SSE lifetime no longer in HTTP duration histogram | ✅ after fix, +Inf/count/sum unchanged |
| SSE lifetime in dedicated `drive9_sse_connection_duration_seconds` | ✅ 15s connection lands in `le=30` bucket |
| Phase-1 slow? | ✅ 4.6–12.9ms (not slow — alert is pure Phase-2 client hold-open) |
| event-bus query latency | ✅ events_since 5.4ms / poll avg ~7ms / latest/oldest 3.5–4.6ms |
| poll/publish failures | ✅ 0 |
| SSE event delivery | ✅ real-time write events pushed correctly, `drive9_sse_events_sent_total` counting correctly |
| multi-tenant `fs_events_rows` reporting | ✅ reported per tenant, cleanup tick normal |

**Conclusion**: the alert was a measurement artifact; Fix A stops the bleeding. Phase-1 and queries are not slow — no real latency regression. Multi-tenant cleanup closes the blind spot of not seeing table size without TiDB access, and eliminates the slow-burn regression.

## Remaining / follow-up (not in this PR)

- **3a DB pool `max_open=0`** (default unlimited): dev shows `drive9_db_pool_connections{role="user",state="max_open"} 0`. N_pods × unlimited connections risks exhausting TiDB `max_connections` (not triggered at dev's low load). This is a deployment config (`DRIVE9_DB_MAX_OPEN_CONNS` env), not in this repo — recommend setting `TiDB_max_connections / N_pods` on the cluster side.
- poll-interval adaptive backoff (3c) and quota_admission_lock contention (3d) — defer based on production data.

## Testing

- `go build ./...` ✅
- `go vet` ✅
- `golangci-lint` (zero new warnings in this PR's source files; pre-existing test-file SA5011 unaffected)
- `go test ./pkg/metrics/ ./pkg/datastore/ ./pkg/tenant/ ./pkg/server/` (incl. SSE/eventbus/cleanup/leader cases) ✅
- dev-cd deploy + dev online verification ✅

## Summary by CodeRabbit

* **New Features**
  * Added richer streaming and event-bus metrics, including connection counts, in-flight tracking, event delivery, heartbeats, resets, and query timing.
  * Added tenant-aware request counting for long-lived streams without affecting duration histograms.
  * Added active tenant store snapshot support for pool-wide maintenance tasks.

* **Bug Fixes**
  * Improved SSE handling so established streams are tracked separately from normal requests.
  * Updated event-stream cleanup and reset handling to better reflect retained event counts and pruning results.
  * Added a total event-row count check for datastore maintenance and reporting.

### DIFF
--- a/pkg/datastore/fs_events.go
+++ b/pkg/datastore/fs_events.go
@@ -82,6 +82,17 @@ func (s *Store) OldestFSEventSeq(ctx context.Context) (int64, error) {
 	return seq.Int64, nil
 }
 
+// CountFSEvents returns the total number of rows in fs_events. Used by the
+// leader cleanup goroutine to report drive9_fs_events_rows so operators can
+// monitor table growth without direct DB access.
+func (s *Store) CountFSEvents(ctx context.Context) (int64, error) {
+	var count int64
+	if err := s.db.QueryRowContext(ctx, `SELECT COUNT(*) FROM fs_events`).Scan(&count); err != nil {
+		return 0, fmt.Errorf("count fs_events: %w", err)
+	}
+	return count, nil
+}
+
 // DeleteFSEventsBefore deletes fs_events rows older than the given threshold.
 // Retention is gated on created_at (DB server's clock at INSERT time), not on
 // the ts field (publisher's clock at event emission). This means the retention
--- a/pkg/datastore/fs_events.go
+++ b/pkg/datastore/fs_events.go
@@ -85,6 +85,14 @@ func (s *Store) OldestFSEventSeq(ctx context.Context) (int64, error) {
 // CountFSEvents returns the total number of rows in fs_events. Used by the
 // leader cleanup goroutine to report drive9_fs_events_rows so operators can
 // monitor table growth without direct DB access.
+//
+// Note: COUNT(*) is a full table scan on TiDB/MySQL InnoDB (unlike PostgreSQL's
+// index-only count). This runs only every fsEventsCleanupInterval (5m) on the
+// leader per tenant with a cached backend, so the cost is bounded and
+// best-effort. If the table grows to millions of rows per tenant and this
+// becomes expensive, switch to a bounded condition (e.g.
+// `WHERE created_at > NOW() - INTERVAL 2 HOUR`) or an approximate count from
+// information_schema.TABLES.
 func (s *Store) CountFSEvents(ctx context.Context) (int64, error) {
 	var count int64
 	if err := s.db.QueryRowContext(ctx, `SELECT COUNT(*) FROM fs_events`).Scan(&count); err != nil {
--- a/pkg/metrics/operations.go
+++ b/pkg/metrics/operations.go
@@ -12,6 +12,23 @@ import (
 var operationDurationBounds = []float64{0.001, 0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10, 30}
 var httpDurationBounds = []float64{0.005, 0.01, 0.025, 0.05, 0.075, 0.1, 0.25, 0.5, 0.75, 1, 2.5, 5, 7.5, 10}
 
+// sseConnectionDurationBounds covers SSE connection lifetimes, which can
+// range from sub-second probes to long-lived mounts. Buckets extend well
+// beyond httpDurationBounds so long-lived connections are visible without
+// piling into +Inf. These are recorded into a dedicated histogram
+// (drive9_sse_connection_duration_seconds), NOT drive9_http_request_duration_seconds.
+var sseConnectionDurationBounds = []float64{0.005, 0.05, 0.1, 0.5, 1, 5, 10, 30, 60, 300, 600, 1800, 3600}
+
+// eventBusQueryDurationBounds covers fs_events DB query latencies. These
+// should normally be low (PK range scan); a shift right indicates DB
+// pressure or table growth.
+var eventBusQueryDurationBounds = []float64{0.0005, 0.001, 0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10}
+
+// ssePhase1DurationBounds covers the Phase-1 replay/reset cost of a new SSE
+// connection (one EventsSince call + stream of buffered events). Used to
+// distinguish slow first-replay from slow Phase-2 live streaming.
+var ssePhase1DurationBounds = []float64{0.001, 0.005, 0.01, 0.025, 0.05, 0.1, 0.25, 0.5, 1, 2.5, 5, 10}
+
 var globalRegistry = NewRegistry()
 var globalMeterProvider = NewMeterProvider(globalRegistry)
 
@@ -19,6 +36,7 @@ var serviceMeter = globalMeterProvider.Meter("service")
 var httpMeter = globalMeterProvider.Meter("http")
 var eventMeter = globalMeterProvider.Meter("event")
 var fuseMeter = globalMeterProvider.Meter("fuse")
+var sseMeter = globalMeterProvider.Meter("sse")
 var tenantMeter = globalMeterProvider.Meter("tenant")
 
 var serviceOperationsTotal = serviceMeter.Int64Counter("drive9_service_operations_total", "Service operations by component/operation/result")
@@ -46,6 +64,29 @@ var tenantHTTPBytes = tenantMeter.Int64Counter("drive9_tenant_http_bytes_total",
 var tenantFileBytes = tenantMeter.Int64Counter("drive9_tenant_file_bytes_total", "Tenant-scoped logical file bytes by tenant/surface/action/direction")
 var tenantStorageBytes = tenantMeter.Float64Gauge("drive9_tenant_storage_bytes", "Tenant storage bytes by state")
 var tenantMediaFiles = tenantMeter.Float64Gauge("drive9_tenant_media_files", "Tenant media file count by state")
+
+// SSE-specific instruments. SSE connection lifetimes are recorded here, not
+// in drive9_http_request_duration_seconds (which would pollute HTTP latency
+// alerts — SSE connections live as long as the client stays subscribed).
+var sseConnectionsTotal = sseMeter.Int64Counter("drive9_sse_connections_total", "SSE /v1/events connections opened by tenant_id/reason")
+var sseConnectionDuration = sseMeter.Float64Histogram("drive9_sse_connection_duration_seconds", "SSE connection lifetime (client stay-open duration, not request processing time)", sseConnectionDurationBounds)
+var sseInflight = sseMeter.Float64Gauge("drive9_sse_inflight_connections", "Active SSE /v1/events connections by tenant_id")
+var ssePhase1Duration = sseMeter.Float64Histogram("drive9_sse_phase1_duration_seconds", "SSE Phase-1 replay/reset duration (one EventsSince call + buffered stream)", ssePhase1DurationBounds)
+var sseEventsSentTotal = sseMeter.Int64Counter("drive9_sse_events_sent_total", "SSE events sent to clients by type/tenant_id")
+var sseResetsSentTotal = sseMeter.Int64Counter("drive9_sse_resets_sent_total", "SSE reset events sent by reason/tenant_id")
+var sseHeartbeatsSentTotal = sseMeter.Int64Counter("drive9_sse_heartbeats_sent_total", "SSE heartbeat events sent by tenant_id")
+
+// Event-bus query instruments. Covers all fs_events DB reads (events_since,
+// poll, latest, oldest) so DB pressure and table growth on the events path
+// are observable without direct DB access.
+var eventBusQueryDuration = serviceMeter.Float64Histogram("drive9_event_bus_query_duration_seconds", "Event-bus fs_events query duration by operation/result/tenant_id", eventBusQueryDurationBounds)
+var eventBusPollFailuresTotal = sseMeter.Int64Counter("drive9_event_bus_poll_failures_total", "Event-bus cross-pod poll query failures by tenant_id")
+var eventBusPublishErrorsTotal = sseMeter.Int64Counter("drive9_event_bus_publish_errors_total", "Event-bus fs_events INSERT failures by tenant_id")
+
+// fs_events table instruments. Compensates for the lack of direct TiDB
+// access: row count and prune volume are reported by the server itself.
+var fsEventsRows = sseMeter.Float64Gauge("drive9_fs_events_rows", "fs_events table row count by tenant_id")
+var fsEventsPrunedTotal = sseMeter.Int64Counter("drive9_fs_events_pruned_total", "fs_events rows pruned by retention cleanup by tenant_id")
 
 func RegisterModule(module string) {
 	globalRegistry.RegisterModule(module)
@@ -110,6 +151,42 @@ func RecordHTTPRequestBodyRead(method, route string, status int, bodyBytes int64
 		Attr("status_class", statusClass(status)),
 		Attr("body_size_bucket", bodySizeBucket(bodyBytes)),
 	)
+}
+
+// RecordHTTPRequestCount records only the request counter, not the duration
+// histogram. Used for SSE/streaming routes whose handler blocks for the
+// connection lifetime (which is not request processing time and would
+// pollute HTTP latency alerts if recorded into the duration histogram).
+func RecordHTTPRequestCount(method, route string, status int) {
+	RegisterModule("server")
+	httpRequestsTotal.Add(1,
+		Attr("method", cleanMetricValue(method, "UNKNOWN")),
+		Attr("route", cleanMetricValue(route, "other")),
+		Attr("status", strconv.Itoa(status)),
+	)
+}
+
+// RecordTenantRequestCount records only the tenant request counter, not the
+// duration histogram. Companion to RecordHTTPRequestCount for SSE routes.
+func RecordTenantRequestCount(tenantID, surface, action, result string, status int) {
+	tenantID = cleanMetricValue(tenantID, "unknown")
+	surface = cleanMetricValue(surface, "other")
+	action = cleanMetricValue(action, "other")
+	result = cleanMetricValue(result, "unknown")
+	statusValue := "unknown"
+	if status > 0 {
+		statusValue = strconv.Itoa(status)
+	}
+	RegisterModule("tenant_usage")
+	attrs := []Attribute{
+		Attr("tenant_id", tenantID),
+		Attr("surface", surface),
+		Attr("action", action),
+		Attr("result", result),
+		Attr("status", statusValue),
+		Attr("status_class", statusClass(status)),
+	}
+	tenantRequestsTotal.Add(1, attrs...)
 }
 
 func RecordHTTPInFlight(value float64) {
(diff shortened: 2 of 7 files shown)


qiffang (approved):
## Review — adversary-2

**判定：GREEN**

PR 是 instrumentation + observability 修复，低风险。Fix A（SSE 从 HTTP 耗时直方图剔除）方向正确，Fix B（诊断指标）覆盖全面，多租户 fs_events cleanup 填补了实际盲区。

**O1: `latest_query_ms` log field 包含 oldest query 耗时**

`EventsSince` 中 cursor gap log：
```go
latestStart := time.Now()
latest := latestSeqSafeWithMetric(...)
oldestStart := time.Now()
oldest := oldestSeqSafeWithMetric(...)
// ... later:
zap.Float64("latest_query_ms", float64(time.Since(latestStart).Microseconds())/1000),
```

`time.Since(latestStart)` 在 log 点执行，包含了 latest + oldest 两次查询的耗时。应该在 `latestSeqSafeWithMetric` 返回后立即计算。不过 `latestSeqSafeWithMetric` 内部已经记录了独立的 metric，所以这个 log field 是冗余的，只是命名误导。

**O2: `ActiveTenantStores` 返回未 pin 的 store 引用**

Doc comment 已说明 use-after-close 风险和容忍策略。如果未来 cleanup 需要更强保证，应改为 Acquire/Release 模式。当前 best-effort 策略可接受。

**O3: `SELECT COUNT(*)` on fs_events**

TiDB 大表上 `COUNT(*)` 可能慢。5 分钟 ticker 下可接受，但如果表增长到百万级别需要关注。

qiffang (commented):
## PR #633 Review — fix(metrics): exclude SSE lifetime from HTTP duration hist; add events diagnostics + multi-tenant fs_events cleanup

### Summary

This PR fixes a metrics pollution issue where SSE `/v1/events` long-lived connections were recorded into the HTTP request duration histogram, making P95/P99 alerts meaningless. The fix has three parts:

1. **Route guard in `observe`** — `/v1/events` records only the request counter (not duration histogram). SSE lifetime goes into a dedicated `drive9_sse_connection_duration_seconds` histogram with appropriate buckets (up to 3600s).

2. **SSE + event-bus observability** — 12 new metrics covering connection lifecycle, Phase-1 replay, events/resets/heartbeats sent, event-bus query latency, poll failures, publish errors, and fs_events table row count.

3. **Multi-tenant fs_events cleanup** — `cleanupFSEvents` now iterates `pool.ActiveTenantStores()` (cheap cache snapshot) to prune each tenant's fs_events table, instead of deferring to a future task.

### Assessment

Clean, well-documented implementation. The root cause analysis is solid and the fix is targeted. See inline comments.

pkg/datastore/fs_events.go
```diff
@@ -82,6 +82,17 @@ func (s *Store) OldestFSEventSeq(ctx context.Context) (int64, error) {
...
 
+// CountFSEvents returns the total number of rows in fs_events. Used by the
+// leader cleanup goroutine to report drive9_fs_events_rows so operators can
+// monitor table growth without direct DB access.
+func (s *Store) CountFSEvents(ctx context.Context) (int64, error) {
+	var count int64
+	if err := s.db.QueryRowContext(ctx, `SELECT COUNT(*) FROM fs_events`).Scan(&count); err != nil {
+		return 0, fmt.Errorf("count fs_events: %w", err)
+	}
+	return count, nil
```
qiffang, line 101: **Performance — `SELECT COUNT(*) FROM fs_events` is a full table scan on TiDB/MySQL.**

Unlike PostgreSQL (which uses index-only scans for count), TiDB and MySQL InnoDB perform a full table scan for `COUNT(*)`. For a large fs_events table (millions of rows across tenants), this could be slow and add read pressure.

Options:
1. Use `COUNT(*)` with a bounded condition: `WHERE created_at > NOW() - INTERVAL 2 HOUR` (only counts recent rows, covers the 1h retention window)
2. Use `SELECT TABLE_ROWS FROM information_schema.tables WHERE table_name = 'fs_events'` for an approximate count (fast, InnoDB estimate)
3. Accept the cost since it runs only every 5 minutes on the leader

Since this runs on a 5-minute leader tick and is best-effort observability, option 3 is probably fine. But worth a comment noting the full-scan cost.

pkg/server/instrumentation.go
```diff
@@ -629,8 +654,20 @@ func (s *Server) observe(next http.Handler, w http.ResponseWriter, r *http.Reque
...
+	// duration histogram would pollute all HTTP P95/P99 alerts (a 40s SSE
+	// connection would look like a 40s "request"). For SSE routes, record
+	// only the request counter (and tenant request counter + bytes); the
+	// connection lifetime goes into the dedicated SSE metrics recorded by
+	// handleEvents instead. duration_ms is still logged below for debugging.
+	if route == "/v1/events" {
+		s.metrics.recordCount(r.Method, route, ow.status)
+		recordTenantHTTPRequestCount(r, ow.status)
+	} else {
+		s.metrics.record(r.Method, route, ow.status, dur)
```
qiffang, line 761: **Nit — hardcoded route string `"/v1/events"` in the observe guard.**

If the route is ever renamed or another SSE endpoint is added, this guard would need updating. Consider extracting a constant or a set:

```go
var sseRoutes = map[string]bool{"/v1/events": true}
```

Minor point — current single-route guard is clear and correct.

pkg/tenant/pool.go
```diff
@@ -114,6 +114,43 @@ func (p *Pool) StopAllFileGC() {
...
+			continue
+		}
+		out = append(out, TenantStoreEntry{TenantID: e.tenantID, Store: e.store})
+	}
+	p.mu.Unlock()
+	return out
+}
+
+// TenantStoreEntry is a (tenantID, store) pair returned by ActiveTenantStores.
+type TenantStoreEntry struct {
```
qiffang, line 149: **Design — `ActiveTenantStores` returns stores without pinning refs.**

The doc comment correctly notes: "callers must not retain them past the cleanup operation — a concurrent retirement may close the underlying DB." The code locks the pool mutex during snapshot but does not increment the entry ref count.

This means a concurrent `Release` + `retire` could close the store's DB while `cleanupFSEvents` is mid-query. The `CountFSEvents` / `DeleteFSEventsBefore` calls would get a "database is closed" error, which is logged and tolerated (matches StartAllFileGC pattern).

This is acceptable for best-effort cleanup, but worth noting: if cleanup ever needs transactional guarantees, the snapshot should hold refs. The current approach is correct for the stated use case.

pkg/metrics/operations.go
```diff
@@ -46,6 +64,29 @@ var tenantFileBytes = tenantMeter.Int64Counter("drive9_tenant_file_bytes_total",
...
 
+// SSE-specific instruments. SSE connection lifetimes are recorded here, not
+// in drive9_http_request_duration_seconds (which would pollute HTTP latency
+// alerts — SSE connections live as long as the client stays subscribed).
+var sseConnectionsTotal = sseMeter.Int64Counter("drive9_sse_connections_total", "SSE /v1/events connections opened by tenant_id/reason")
+var sseConnectionDuration = sseMeter.Float64Histogram("drive9_sse_connection_duration_seconds", "SSE connection lifetime (client stay-open duration, not request processing time)", sseConnectionDurationBounds)
+var sseInflight = sseMeter.Float64Gauge("drive9_sse_inflight_connections", "Active SSE /v1/events connections by tenant_id")
+var ssePhase1Duration = sseMeter.Float64Histogram("drive9_sse_phase1_duration_seconds", "SSE Phase-1 replay/reset duration (one EventsSince call + buffered stream)", ssePhase1DurationBounds)
+var sseEventsSentTotal = sseMeter.Int64Counter("drive9_sse_events_sent_total", "SSE events sent to clients by type/tenant_id")
+var sseResetsSentTotal = sseMeter.Int64Counter("drive9_sse_resets_sent_total", "SSE reset events sent by reason/tenant_id")
```
qiffang, line 76: **Cardinality — 12 new metrics all include `tenant_id` as a label.**

With many tenants, each metric creates a time series per tenant. For example, `drive9_sse_heartbeats_sent_total` fires every 15s per SSE connection per tenant. If there are 100 tenants with 5 SSE connections each, that's 500 heartbeat increments per 15s interval, creating 100 time series just for heartbeats.

Across all 12 new metrics, this could add thousands of series. For Prometheus scrape targets with many tenants, this can cause high memory usage and slow queries.

Consider:
1. Making `tenant_id` a label only on the most important metrics (connection count, inflight, query duration)
2. Using exemplars instead of labels for tenant attribution on high-frequency counters (heartbeats, events sent)

Not a blocker if the current tenant count is manageable, but worth monitoring.

pkg/server/sse.go
```diff
@@ -206,6 +207,17 @@ func (s *Server) handleEvents(w http.ResponseWriter, r *http.Request) {
...
 	bus := s.tenantEventBus(r)
+	tenantID := bus.tenantID
+	connStart := time.Now()
+	// Track SSE inflight and connection lifetime. The inflight count is
+	// derived from the EventBus listener set (adjusted in Subscribe/
+	// Unsubscribe); here we record the connection lifecycle into the
+	// dedicated SSE metrics (NOT the HTTP duration histogram — see the
+	// route guard in observe).
+	defer func() {
+		metrics.RecordSSEConnection(tenantID, "closed", time.Since(connStart))
```
qiffang, line 224: **Good — SSE connection lifecycle is recorded in the right place.**

The `defer RecordSSEConnection(tenantID, "closed", time.Since(connStart))` fires when `handleEvents` returns (client disconnect), capturing the true connection lifetime. This is correctly separated from the HTTP duration histogram.

pkg/server/server.go
```diff
@@ -563,17 +563,57 @@ func (s *Server) cleanupFSEvents(ctx context.Context) {
...
+	}
+	// Multi-tenant mode: iterate tenants with a cached backend, prune each
+	// fs_events table and sample row counts. fs_events rows only accumulate for
+	// tenants that have served traffic (and thus have a cached backend), so
+	// iterating ActiveTenantStores (a cheap cache snapshot) covers the tenants
+	// that need cleanup without forcing backend creation for dormant tenants
+	// (which would exhaust connections and fail for tenants whose credentials
+	// are not materialized). Runs on the fsEventsCleanupInterval (5m) ticker.
+	// A concurrent retirement may close the store after the snapshot returns;
+	// a stale-reference COUNT/DELETE logs a warning and is retried next tick —
```
qiffang, line 591: **Good — multi-tenant cleanup correctly checks `ctx.Err()` between tenants.**

The loop breaks early on shutdown, avoiding long cleanup runs during graceful shutdown. The per-tenant error handling logs but continues to the next tenant, which is the right behavior for best-effort retention pruning.

One observation: `CountFSEvents` runs before `DeleteFSEventsBefore` on every tenant, every tick. For N active tenants, that's 2N queries per 5-minute tick. If the COUNT becomes expensive (see comment on full table scan), consider sampling the count less frequently (e.g. every 3rd tick) while still pruning every tick.

pkg/server/instrumentation.go
```diff
@@ -629,8 +654,20 @@ func (s *Server) observe(next http.Handler, w http.ResponseWriter, r *http.Reque
...
-	s.metrics.record(r.Method, route, ow.status, dur)
-	recordTenantHTTPRequest(r, ow.status, dur, ow.size)
+	// SSE routes (/v1/events) block in a select loop for the entire client
+	// connection lifetime — recording that lifetime into the HTTP request
+	// duration histogram would pollute all HTTP P95/P99 alerts (a 40s SSE
+	// connection would look like a 40s "request"). For SSE routes, record
+	// only the request counter (and tenant request counter + bytes); the
+	// connection lifetime goes into the dedicated SSE metrics recorded by
+	// handleEvents instead. duration_ms is still logged below for debugging.
+	if route == "/v1/events" {
```
qiffang: This route-level guard drops durations for every `/v1/events` response, not only for established SSE streams. `handleEvents` can return bounded errors before any SSE connection is created, for example non-GET requests, an invalid `since`, missing/flusher failures, or auth/backend setup failures before `tenantEventBus` succeeds. Those requests now increment only counters, are absent from HTTP/tenant duration histograms, and also never reach the dedicated SSE lifetime metric.

That changes alert semantics beyond the stated fix. The exclusion should be keyed on an actually established streaming response, or non-stream/error responses on `/v1/events` should still record normal HTTP/tenant duration. Please add regressions that a successful long-lived SSE stream is excluded while bounded `/v1/events` failures are still included in the duration metrics.

pkg/server/server.go
```diff
@@ -563,17 +563,57 @@ func (s *Server) cleanupFSEvents(ctx context.Context) {
...
+	// this is acceptable for best-effort retention pruning and matches the
+	// StartAllFileGC iteration pattern.
+	if s.pool == nil {
+		return
+	}
+	for _, te := range s.pool.ActiveTenantStores() {
```
qiffang, line 611: `ActiveTenantStores()` only snapshots tenants currently present in the in-memory pool. A tenant can serve traffic, accumulate `fs_events`, then be evicted or go dormant before this cleanup tick; after that it is no longer in `ActiveTenantStores()` and its expired rows are not pruned until some future request happens to reacquire the backend.

That means the retention cleanup is best-effort for the active cache, not multi-tenant retention enforcement. If this PR is claiming multi-tenant `fs_events` cleanup, it needs a path for evicted/dormant tenants too, for example prune-on-acquire/eviction or a bounded tenant scan, plus a regression covering a tenant that leaves the active cache before cleanup runs.

qiffang (dismissed):
I’m requesting changes on head `0305a4e`.

Blockers:
- The `/v1/events` instrumentation guard removes **all** event-route durations from HTTP/tenant duration histograms, not just established long-lived SSE streams. Bounded failures before the stream is established disappear from the normal latency metrics and are not covered by SSE lifetime metrics either.
- The multi-tenant `fs_events` cleanup only scans currently cached tenants, so tenants that accumulated rows and were then evicted/dormant can keep expired rows indefinitely. That does not satisfy the PR’s multi-tenant retention-cleanup claim.
- There are no `*_test.go` changes in this PR despite changing alert-critical metrics behavior and retention cleanup semantics. Please add regressions for successful SSE streams being excluded from HTTP duration, bounded `/v1/events` failures still being counted in HTTP/tenant duration, and multi-tenant cleanup covering the eviction/dormant-tenant case or explicitly scoped alternative.

Validation notes: GitHub `code-ci` and `local-e2e` are green on this head, CodeRabbit status is success, and local `git diff --check origin/main...HEAD` passed. I could not run local Go tests on this machine because `go` is not installed.

pkg/server/instrumentation.go
```diff
@@ -629,8 +654,20 @@ func (s *Server) observe(next http.Handler, w http.ResponseWriter, r *http.Reque
...
+		s.metrics.recordCount(r.Method, route, ow.status)
+		recordTenantHTTPRequestCount(r, ow.status)
```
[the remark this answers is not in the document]
mornyx: Done. The guard now keys on `sseStreamEstablished` (set by `markSSEStreamEstablished(r.Context())` only after `handleEvents` writes the 200 streaming headers), not on the route name alone. Bounded error responses on `/v1/events` (non-GET, invalid `since`, bad auth, backend setup failures, missing flusher) return before the flag is set, so they are recorded as normal HTTP/tenant requests again — including the HTTP duration histogram. Only an established SSE stream skips the duration and goes to the dedicated SSE metrics.

pkg/server/instrumentation.go
```diff
@@ -582,6 +590,23 @@ func recordTenantHTTPRequest(r *http.Request, status int, d time.Duration, respo
...
+		return
+	}
+	class := classifyTenantRequest(r)
+	if scopedClass, ok := requestMetricClass(r.Context()); ok {
+		class = scopedClass
+	}
+	metrics.RecordTenantRequestCount(tenantID, class.surface, class.action, tenantRequestResult(status), status)
+	if r.ContentLength > 0 {
+		metrics.RecordTenantHTTPBytes(tenantID, class.surface, class.action, "request", r.ContentLength)
+	}
```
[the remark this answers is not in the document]
mornyx: Done. The established-SSE branch now calls `recordTenantHTTPResponseBytes(r, ow.size)` so `drive9_tenant_http_bytes_total{direction="response"}` still reflects SSE stream bytes. (Request bytes are already recorded by `recordTenantHTTPRequestCount`.)

pkg/server/sse.go
```diff
@@ -235,10 +248,12 @@ func (s *Server) handleEvents(w http.ResponseWriter, r *http.Request) {
 			}
 		}
 		sendSSEReset(bw, headSeq, reason)
+		metrics.RecordSSEResetSent(tenantID, reason)
 		lastSeen = headSeq
 	} else {
 		for _, ev := range events {
 			sendSSEEvent(bw, ev)
+			metrics.RecordSSEEventSent(tenantID, ev.Op)
```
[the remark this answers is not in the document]
mornyx: Done. Both the Phase-1 replay path and the live `pollAndSend` path now check `isStructuralOp(ev.Op)` and call `metrics.RecordSSEResetSent(tenantID, "structural_change")` for structural ops instead of `RecordSSEEventSent`. `drive9_sse_resets_sent_total{reason="structural_change"}` now correctly counts client invalidations from rename/delete/mkdir/copy.

pkg/server/server.go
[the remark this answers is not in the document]
mornyx: Acknowledged — addressed as a deliberate best-effort design decision with full rationale in the cleanupFSEvents comment. Prune-on-acquire was rejected to avoid coupling a DELETE (RTT + lock cost) into the request hot path; the stale-cursor reset in EventsSince already makes correctness independent of timely pruning; and a bounded meta-driven tenant scan (keyset-paginated, like the semantic worker) is noted as a future option if the active set grows large. See also the reply on qiffang's matching comment.

pkg/datastore/fs_events.go
mornyx: Agreed. Added a comment on CountFSEvents documenting the full-scan cost on TiDB/MySQL InnoDB and the mitigations (bounded `WHERE created_at > NOW() - INTERVAL 2 HOUR`, or `information_schema.TABLES` approximate count) if the table grows large. Keeping the unbounded COUNT for now since it runs every 5m on the leader per cached tenant only — if it shows up in `drive9_event_bus_query_duration_seconds{operation=...}` or DB load we will switch to the bounded condition.

pkg/server/instrumentation.go
mornyx: Done. Extracted `sseEventsRoute = "/v1/events"` (sse.go) and used it in observe, requestRoute, classifyTenantRequest, and the server dispatch. If a second SSE route is added later it can join a set; for now the single constant plus the `sseStreamEstablished` flag keeps the guard clear and avoids a stray string rename silently dropping the guard.

pkg/tenant/pool.go
mornyx: Acknowledged — the design is intentional and the doc comment states the use-after-close risk. The best-effort cleanup tolerates a stale-reference error (logged, retried next tick) exactly like StartAllFileGC. Not pinning refs avoids a retirement being blocked for the whole 5m tick by a slow COUNT/DELETE. If cleanup ever needs transactional guarantees we will switch to Acquire/release.

pkg/metrics/operations.go
mornyx: Fair point. The 12 new metrics all carry `tenant_id`, so with N active tenants each adds N series. For the current tenant count this is fine; the high-frequency counters (heartbeats/events_sent) are the main concern. We are keeping tenant_id across all of them for now because per-tenant alerting (the original bug) is the use case, but we will monitor Prometheus scrape size and query latency and drop tenant_id from the high-frequency counters (heartbeats/events_sent) if cardinality becomes a problem — those are diagnostic-only and can be aggregated to a global counter without losing the alert signal.

pkg/server/sse.go
mornyx: Thanks — and this is now the only place the connection lifetime is recorded: the HTTP duration histogram no longer sees `/v1/events` (only for established streams, via the sseStreamEstablished flag).

pkg/server/server.go
mornyx: Good observation. Keeping COUNT+DELETE per tenant per tick for now (both are bounded; COUNT is the full-scan one noted above). If COUNT shows up as expensive we will sample it less frequently (e.g. every 3rd tick) while still pruning every tick — the row-count gauge just needs to track the trend, not be exact each cycle. No code change in this PR; flagged for follow-up if the metrics show it.

pkg/server/instrumentation.go
mornyx: Fixed — same as the codex comment: the guard now keys on `sseStreamEstablished`, set only after `handleEvents` writes the 200 streaming headers. Bounded error responses on `/v1/events` (non-GET, invalid since, bad auth, backend/flusher failures) return before the flag is set and are recorded as normal HTTP requests with their real duration. Only established streams skip the HTTP/tenant duration histograms and go to the dedicated SSE metrics.

pkg/server/server.go
mornyx: Acknowledged — this is the same point as the codex comment. Addressed as a deliberate best-effort design decision, documented in cleanupFSEvents: a tenant evicted from the pool cache before the next cleanup tick is not pruned until it is re-acquired. Rationale: (1) EventsSince's stale-cursor reset (reset when since < oldestSeq) makes correctness independent of timely pruning; (2) evicted-dormant accumulation is bounded by fsEventsRetention (1h); (3) prune-on-acquire was rejected to avoid coupling a DELETE into the request hot path. A bounded meta-driven tenant scan (keyset-paginated like the semantic worker) is the future option if the active set grows large enough to matter. So this PR provides multi-tenant cleanup for the tenants that actually accumulate rows (cached backends), with a documented gap for the evicted-dormant edge case rather than claiming full retention enforcement.

srstack (approved):
Reviewed the full diff; built, vetted, and ran the metrics + server (SSE/eventbus/cleanup) tests — all green. Approving.

## Fix A (the core fix) is correct ✅
Excluding established SSE streams from `drive9_http_request_duration_seconds` is the right call, and the implementation gets the crucial nuance right: `markSSEStreamEstablished` is set **only after** `WriteHeader(200)` + entering the select loop, so bounded error responses on `/v1/events` (bad auth, invalid `since`, non-GET, backend setup failure) return before the flag is set and are **still** recorded into the HTTP/tenant duration histograms. So you stop the SSE-lifetime pollution without going blind to genuine error latency on the route. The request counter (+ tenant counter + response bytes) is still recorded for established streams, so request rate/error visibility for `/v1/events` is preserved. The flag is read in the same goroutine after `next.ServeHTTP` returns (no race), under the existing state mutex.

## Multi-tenant fs_events cleanup ✅
`ActiveTenantStores` (cache snapshot under `p.mu`, skipping nil/retired) is the right primitive — avoiding `ListTenantsByStatus + Acquire` correctly sidesteps forcing backend creation (and connection exhaustion) for dormant tenants. The use-after-close window is real but handled: a concurrent retirement closes the captured `*Store`, and a stale `CountFSEvents`/`DeleteFSEventsBefore` returns a closed-DB **error** (not a panic), which is logged and retried next tick. The tenant↔store mapping comes from the cache so there's no cross-tenant deletion risk. The eviction/dormant limitation and the EventsSince stale-cursor-reset safety net are documented thoroughly and the tradeoff is sound.

## Concerns (non-blocking)

### 🟡 1. `tenant_id` on histograms is the main cardinality risk to watch
The new **histograms** carry `tenant_id` — `event_bus_query_duration_seconds{tenant_id,operation,result}` (≈13 buckets × 4 ops × 2 results × tenants) and `sse_connection_duration_seconds{tenant_id}` / `sse_phase1_duration_seconds{tenant_id}`. The active series are bounded by tenants with **live** SSE/event-bus activity (much smaller than the full fleet), which mitigates it, but Prometheus retains a tenant's histogram series after disconnect until staleness — so the series set grows with the number of *distinct* tenants that ever connected. With a large/growing fleet (esp. alongside the tenant-pool work) this can add up. Worth confirming against your cardinality budget; if it bites, a global (un-tenant'd) duration histogram + per-tenant counters would keep the diagnostic value at a fraction of the series cost. `fs_events_rows{tenant_id}` (gauge, one per active-cached tenant per tick) is comparatively cheap.

### 🟢 2. `CountFSEvents` is `COUNT(*)` (full scan on TiDB)
Acknowledged in the comment and bounded by the 1h retention keeping the table small + the 5m/leader cadence. Fine as-is; the suggested `information_schema.TABLES` / bounded-WHERE fallback is the right escape hatch if a tenant's table ever grows.

## Verdict
High-quality, well-documented, dev-verified. Correct root-cause analysis (measurement artifact, not a real regression), the fix is surgical and preserves error-latency visibility, and the multi-tenant cleanup closes a genuine unbounded-growth gap with a carefully-reasoned best-effort design. Approving — just keep an eye on the per-tenant histogram cardinality (#1) as the fleet grows.

PR #634, opened 2026-06-27
fix: treat quota outbox claim races as benign

srstack:
## Summary
- handle quota outbox claim-stage row-count mismatches as benign multi-server claim races instead of surfacing ErrQuotaOutboxLeaseMismatch
- retry claim conflicts a few times, then return an empty batch so another worker or the next poll can continue
- keep ErrQuotaOutboxLeaseMismatch semantics for ack/retry ownership failures
- add concurrent batch-claim coverage to assert no duplicate claims

## Context
Production logs showed quota_outbox_process_failed caused by claim_quota_outbox_batch returning quota outbox lease mismatch across multiple drive9-server pods. This happens before apply/ack, so it should not be treated as an ownership failure for an already claimed row.

## Tests
- make test TEST_PKGS='./pkg/datastore' TEST_RUN='TestQuotaOutbox'
- make test TEST_PKGS='./pkg/backend' TEST_RUN='Test.*QuotaOutbox|TestDrainQuotaOutboxForFile'
- make lint
- git diff --check

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when claiming queued quota outbox items under concurrent load.
  * Reduced false ownership/lease errors by treating benign claim races as retryable.
  * Added limited automatic retries so claim attempts can recover from temporary conflicts without failing immediately.
  * Strengthened concurrency handling to help prevent duplicate processing of the same queued item.

### DIFF
--- a/pkg/datastore/quota_outbox.go
+++ b/pkg/datastore/quota_outbox.go
@@ -17,6 +17,7 @@ import (
 // current capped exponential backoff this is roughly tens of minutes, while
 // still allowing normal transient central-store failures to recover.
 const defaultQuotaOutboxMaxAttempts = 100
+const quotaOutboxClaimMaxAttempts = 3
 const quotaAdmissionLockName = "default"
 
 type QuotaOutboxStatus string
@@ -55,6 +56,8 @@ type QuotaOutboxEntry struct {
 // ErrQuotaOutboxLeaseMismatch means a worker tried to ack/retry an outbox row
 // it no longer owns.
 var ErrQuotaOutboxLeaseMismatch = errors.New("quota outbox lease mismatch")
+
+var errQuotaOutboxClaimConflict = errors.New("quota outbox claim conflict")
 
 // EnqueueQuotaOutboxTx inserts a queued quota mutation inside an existing
 // tenant metadata transaction.
@@ -106,6 +109,17 @@ func (s *Store) ClaimQuotaOutboxBatch(ctx context.Context, now time.Time, leaseD
 }
 
 func (s *Store) claimQuotaOutboxBatch(ctx context.Context, now time.Time, leaseDuration time.Duration, limit int) ([]QuotaOutboxEntry, error) {
+	for attempt := 0; attempt < quotaOutboxClaimMaxAttempts; attempt++ {
+		entries, err := s.claimQuotaOutboxBatchOnce(ctx, now, leaseDuration, limit)
+		if errors.Is(err, errQuotaOutboxClaimConflict) {
+			continue
+		}
+		return entries, err
+	}
+	return nil, nil
+}
+
+func (s *Store) claimQuotaOutboxBatchOnce(ctx context.Context, now time.Time, leaseDuration time.Duration, limit int) ([]QuotaOutboxEntry, error) {
 	if limit <= 0 {
 		limit = 1
 	}
@@ -183,7 +197,12 @@ func (s *Store) claimQuotaOutboxBatch(ctx context.Context, now time.Time, leaseD
 		return nil, err
 	}
 	if rowsAffected != int64(len(entries)) {
-		return nil, ErrQuotaOutboxLeaseMismatch
+		// In multi-server deployments TiDB can report that fewer rows were
+		// claimed than the preceding SKIP LOCKED read selected. Treat that as a
+		// benign claim race: rollback this tx and let this worker retry or let
+		// another worker process the rows. Ack/retry paths still use
+		// ErrQuotaOutboxLeaseMismatch for real ownership failures.
+		return nil, errQuotaOutboxClaimConflict
 	}
 
 	for i := range entries {
--- a/pkg/datastore/quota_outbox.go
+++ b/pkg/datastore/quota_outbox.go
@@ -10,6 +10,8 @@ import (
 	"time"
 
 	"github.com/google/uuid"
+
+	"github.com/mem9-ai/drive9/pkg/metrics"
 )
 
 // defaultQuotaOutboxMaxAttempts bounds how long a poisoned quota mutation can
@@ -116,6 +118,7 @@ func (s *Store) claimQuotaOutboxBatch(ctx context.Context, now time.Time, leaseD
 		}
 		return entries, err
 	}
+	metrics.RecordOperation("datastore", "claim_quota_outbox_conflict_exhausted", "conflict", 0)
 	return nil, nil
 }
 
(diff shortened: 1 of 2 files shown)


srstack:
Addressed the observability follow-up:\n\n- Added a bounded metric when claim conflicts exhaust all 3 retries: `drive9_service_operations_total{component="datastore",operation="claim_quota_outbox_conflict_exhausted",result="conflict"}`.\n- Kept retries tight/no backoff intentionally, since the retry immediately re-selects and skips rows claimed by another worker.\n- Added the corresponding runbooks alert in tidbcloud/runbooks commit `7428d8b0` on `dat9-dev2/drive9-body-read-alert-rules`.

qiffang (approved):
## Adversarial Review — PR #634

**Verdict: GREEN** (0 blockers, 2 observations)

### Checklist

| Gate | Status |
|------|--------|
| Edge cases covered? | PASS |
| API explicit? | PASS |
| Race / state issues? | PASS |

---

### Analysis

The fix correctly separates two semantically different failure modes:

1. **Claim-stage conflict** (new `errQuotaOutboxClaimConflict`): `SKIP LOCKED` SELECT returns N rows, but concurrent UPDATE claims fewer because another pod grabbed some between SELECT and UPDATE. This is benign — retry or let the next poll handle it.

2. **Ownership failure** (`ErrQuotaOutboxLeaseMismatch`): ack/retry paths where a worker tries to update a row it no longer owns. This remains an error.

The retry loop (3 attempts) with metric recording on exhaustion is proportionate. The `nil, nil` return after exhaustion is safe because the poll cycle will retry.

---

### O1 — `return nil, nil` after exhausting retries is silent

After 3 failed claims, a metric is recorded but no log line is emitted. In sustained high-contention scenarios, all workers could repeatedly exhaust retries without claiming any rows (livelock). The metric `claim_quota_outbox_conflict_exhausted` helps detect this, but a warning-level log would make it more visible in operational debugging without requiring a metrics dashboard.

### O2 — Concurrent test may not exercise the conflict path on SQLite

`TestQuotaOutboxConcurrentBatchClaimNoDuplicates` uses `newTestStore(t)`, likely SQLite. SQLite serializes transactions, so `SKIP LOCKED` + concurrent UPDATE race that triggers `errQuotaOutboxClaimConflict` (rowsAffected != len(entries)) may never fire. The test validates no-duplicate semantics correctly but may not exercise the retry logic itself. Consider whether this needs a unit test that directly calls `claimQuotaOutboxBatchOnce` with a mocked execer that returns mismatched rowsAffected.

---

**CI**: `ci` pass, CodeRabbit pass on head `bfa40ba`. `drive9-server-local e2e smoke` pending.

**Tests**: 1 new concurrent test covering no-duplicate claims across 8 goroutines with 20 rows. Well-structured.

srstack:
Follow-up on the adversarial review observations:\n\n- O1: I’m intentionally not adding a warning log for exhausted claim conflicts. The production issue this PR fixes was noisy benign claim-race logging; adding a warn on the benign fallback path would reintroduce log noise. The new metric plus runbooks alert covers sustained/pathological contention without turning normal races back into warning/error logs.\n- O2: The SQLite premise does not apply here. `newTestStore(t)` in `pkg/datastore/store_test.go` uses `internal/testmysql`, which starts MySQL 8.0.36 via testcontainers unless `DRIVE9_TEST_MYSQL_DSN` is supplied. So the concurrency test is not running against SQLite.\n\nLatest checks on head `bfa40ba`: CodeRabbit pass, ci pass, drive9-server-local e2e smoke pass.

qiffang (commented):
## Review: PR #634 — quota outbox claim conflict handling

Overall the change is sound — correctly identifies TiDB's `SKIP LOCKED` row-count mismatch as a benign claim race and separates it from real ownership failures. A few observations:

### 1. Exhausted retries return `nil, nil` — silent data loss risk?

`claimQuotaOutboxBatch` returns `nil, nil` after `quotaOutboxClaimMaxAttempts` exhausted (line 122). The caller treats this the same as "no rows available" and stops claiming. This is safe because:
- The rows remain `queued` (the tx was rolled back on conflict)
- Another worker or the next poll cycle will pick them up

But: if a single worker is the **only** claimer and consistently hits conflicts (e.g. recovery goroutine racing), it silently gives up for this poll cycle. The metric `claim_quota_outbox_conflict_exhausted` covers observability, which is good. Consider whether a log line at `WARN` level alongside the metric would help operators notice sustained conflicts faster.

### 2. No backoff between retry attempts

The retry loop (lines 114-123) immediately retries on conflict with no delay:
```go
for attempt := 0; attempt < quotaOutboxClaimMaxAttempts; attempt++ {
    entries, err := s.claimQuotaOutboxBatchOnce(...)
    if errors.Is(err, errQuotaOutboxClaimConflict) {
        continue  // immediate retry
    }
```
With 3 attempts this is fine — the cost is 3 transactions in quick succession. But worth noting that if `quotaOutboxClaimMaxAttempts` is ever increased, adding a brief `time.Sleep` or jitter would prevent thundering-herd retries.

### 3. Concurrent test correctness

`TestQuotaOutboxConcurrentBatchClaimNoDuplicates` (test file line 74-129) is well-designed:
- 8 goroutines, 20 rows, batch size 5
- Checks no duplicate claims AND all 20 eventually claimed

One note: the test uses `s.DB()` directly for enqueue (line 81) instead of `s.InTx()`. This works because `EnqueueQuotaOutboxTx` accepts an `execer`, but it bypasses the transactional wrapper pattern used elsewhere. Not a correctness issue, just a style difference from other tests.

### 4. Error naming is clean

`errQuotaOutboxClaimConflict` (unexported) for benign claim races vs `ErrQuotaOutboxLeaseMismatch` (exported) for real ownership failures — this is the right separation. The unexported error correctly stays internal to the retry loop and never leaks to callers.

### Summary

LGTM with minor suggestions. The core fix is correct: claim-stage row-count mismatches are benign under TiDB's `SKIP LOCKED` and should be retried, not fatal.

qiffang (approved):
Reviewed PR #634 independently.

What I checked:
- Claim-stage row-count mismatch is now separated from real ack/retry ownership failures; `ErrQuotaOutboxLeaseMismatch` remains preserved for stale receipt paths.
- Conflict path rolls back the partial claim transaction and retries only claim conflicts, so already-updated rows are not leaked under the failed attempt.
- Exhausted claim conflicts return an empty batch, which is safe for the outbox worker poll model and avoids false `quota_outbox_process_failed` noise.
- Added concurrent batch-claim coverage checks no duplicate claims across workers.

Validation:
- `go test -c ./pkg/datastore` passed locally.
- Runtime `make test TEST_PKGS='./pkg/datastore' TEST_RUN='TestQuotaOutbox'` is blocked on this machine by local testcontainers (`rootless Docker not found`).
- GitHub `code-ci` and `local-e2e` are green on `bfa40ba`.

No blocking findings.

PR #638, opened 2026-06-28
Fix quota worker noise and pending delta cache races

srstack:
## Summary
- make quota outbox worker metrics tenant-scoped
- classify expected tenant pool database errors as auth/quota/not_found labels instead of generic error noise
- avoid live quota pending-delta scans when local cache loads race with small-write enqueue/ack activity

## Tests
- go test ./pkg/backend -run 'TestQuotaPendingDeltasCache|TestQuotaUsageCache|TestQuotaConfigCache' -count=1
- go test ./pkg/backend -run 'Test.*QuotaOutbox|TestServerQuotaUploadReserveUsesLivePendingOutboxDeltas' -count=1
- go test ./pkg/backend -count=1
- go test ./pkg/server -run 'Test.*Provision|TestAdmin' -count=1
- make lint

## Summary by CodeRabbit

* **New Features**
  * Improved handling of quota processing under concurrency, with more reliable results when updates race with reads.
  * Added clearer outcome reporting for quota outbox batch claims, including when retries are exhausted.

* **Bug Fixes**
  * Reduced unnecessary failures during quota cache refreshes and kept published values stable in more race conditions.
  * Better classified tenant pool errors so logs and metrics are more accurate and easier to interpret.

### DIFF
--- a/pkg/backend/dat9.go
+++ b/pkg/backend/dat9.go
@@ -119,6 +119,7 @@ type Dat9Backend struct {
 	quotaOutboxNotify chan struct{}
 	quotaOutboxWG     sync.WaitGroup
 	quotaOutboxStop   context.CancelFunc
+	claimQuotaOutbox  quotaOutboxBatchClaimer
 
 	s3EncryptionPolicy meta.ResolvedS3EncryptionPolicy
 
--- a/pkg/backend/quota_cache.go
+++ b/pkg/backend/quota_cache.go
@@ -291,10 +291,11 @@ type quotaPendingDeltasCache struct {
 	load     quotaPendingDeltasLoader
 	ttl      time.Duration
 
-	mu         sync.RWMutex
-	snapshot   *quotaPendingDeltasSnapshot
-	generation uint64
-	loadMu     sync.Mutex
+	mu                  sync.RWMutex
+	snapshot            *quotaPendingDeltasSnapshot
+	generation          uint64
+	localPositiveDeltas quotaPendingDeltas
+	loadMu              sync.Mutex
 }
 
 func newQuotaPendingDeltasCache(tenantID string, load quotaPendingDeltasLoader, ttl time.Duration) *quotaPendingDeltasCache {
@@ -322,6 +323,7 @@ func (c *quotaPendingDeltasCache) get(ctx context.Context) (quotaPendingDeltas, 
 
 	c.mu.RLock()
 	generation := c.generation
+	localPositiveDeltas := c.localPositiveDeltas
 	c.mu.RUnlock()
 
 	start := time.Now()
@@ -339,12 +341,19 @@ func (c *quotaPendingDeltasCache) get(ctx context.Context) (quotaPendingDeltas, 
 	expiresAt := time.Now().Add(c.ttl)
 	c.mu.Lock()
 	if c.generation != generation {
+		deltas.add(c.localPositiveDeltas.sub(localPositiveDeltas))
+		c.snapshot = &quotaPendingDeltasSnapshot{
+			deltas:    deltas,
+			expiresAt: expiresAt,
+		}
 		c.mu.Unlock()
-		// A local enqueue or ack raced this DB load. Do not merge blindly:
-		// the SELECT may or may not have observed that row. Let the caller
-		// fall back to a live tenant-DB read for this check.
+		// A local enqueue or ack raced this DB load. The SELECT may or may not
+		// have observed those rows. Merge only positive local deltas from the
+		// race window: that may double-count this process's newest writes for
+		// one short TTL, but it avoids undercounting quota and keeps small-write
+		// admission off the expensive in-transaction SUM(quota_outbox) path.
 		metrics.RecordTenantOperation(c.tenantID, "server_quota", "pending_delta_cache", "raced_local_delta", time.Since(start))
-		return quotaPendingDeltas{}, false
+		return deltas, true
 	}
 	c.snapshot = &quotaPendingDeltasSnapshot{
 		deltas:    deltas,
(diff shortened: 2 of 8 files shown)


srstack (commented):
Reviewed the diff. Focused, well-targeted PR — three independent changes, each addressing a real issue. LGTM overall; two things worth doing before merge (one test gap, one robustness nit), neither blocking the design.

### What's correct

**`quota_cache.go` — pending-delta race merge.** The new "publish a conservative snapshot from the race window" behavior is fail-safe in the right direction: it only ever merges *positive* local deltas, so pending usage is over-counted at worst (if the SELECT already saw the row, it's double-counted) → quota rejects earlier, never over-admits. Ignoring negative (ack) deltas is conservative the same way. Both new tests are good and confirm the loader runs once (the previous behavior forced a live read every time). One behavioral note worth a comment: a tenant near its limit can see spurious rejections for up to one TTL after a racy load — documented in the code, acceptable tradeoff vs. the in-transaction `SUM(quota_outbox)`.

**`quota_outbox` — conflict-exhausted.** Moving the outcome from a global datastore-layer metric to a tenant-scoped worker-layer metric (and dropping datastore's dependency on `metrics`) is clean. Bonus: `claim` now records `result="ok"`, so its outcome distribution is finally complete (ok/empty/error/conflict).

**`pool.go` — error classification.** This genuinely reduces alert noise: `Drive9ServiceOperationErrorRateHigh`'s numerator is `result=~"error|bad_conn|.+_error"`, and `auth_failed`/`usage_quota_exhausted`/`not_found` don't match it, so tenant-caused expected failures no longer inflate the error-rate alert. Demoting them to Warn is also right.

### Suggestions

**1. Classify by error code, not full-sentence match (agree with Copilot, with a concrete path).** `strings.Contains(msg, "please check your user name and password")` is brittle across TiDB/driver versions and locales. I checked the error chain — `OpenInstrumented` (`ping mysql: %w`) and meta/datastore (`open db: %w`) both preserve the wrapper, so the `*mysql.MySQLError` is still typed at `observePool`. The auth case can use the code:
```go
var myErr *mysql.MySQLError
if errors.As(err, &myErr) && myErr.Number == 1045 { return "auth_failed" }
```
Keep a short substring for the quota case (1105 is the generic `HY000`, so the message is needed there). Note: the unit test currently builds plain `fmt.Errorf("...Error 1045...")` strings — switching to `errors.As` means the test should use real `&mysql.MySQLError{Number:1045}` values, otherwise that branch isn't actually exercised.

**2. Add a test for the `ConflictExhausted` branch (agree with Copilot).** `ProcessQuotaOutboxBatch` gained an early-return plus two new tenant-scoped metrics, but nothing drives `ConflictExhausted=true`. A unit test that makes the claim return `{ConflictExhausted:true}` and asserts the early return + the `claim/conflict` and `claim_conflict_exhausted/conflict` labels would lock in the behavior.

**3. Minor:** the conflict-exhausted path records two operations (`claim`+`claim_conflict_exhausted`, both `result="conflict"`). Looks intentional (general distribution + dedicated alert series) — just confirm dashboards consume it that way. Also `localPositiveDeltas` is a monotonic accumulator; int64 overflow is only theoretical (exabyte-scale cumulative writes), but a one-line note wouldn't hurt.

srstack:
Addressed the review comments:

- tenant pool result labels now use package-local typed constants.
- auth failures are classified via MySQL error code 1045, with legacy message fallback preserved.
- quota exhaustion uses MySQL 1105 plus a usage-quota message check, with the existing legacy fallback preserved.
- added a ProcessQuotaOutboxBatch ConflictExhausted unit test covering early return and tenant-scoped claim / claim_conflict_exhausted metrics.

Validation run locally:
- go test ./pkg/tenant ./pkg/backend -count=1
- make lint

qiffang (approved):
## Review: PR #638 — Fix quota worker noise and pending delta cache races

### Overall: APPROVE

Clean, well-scoped fix across three related areas. 296 additions, 38 deletions, 8 files.

---

### 1. Quota outbox tenant-scoped metrics (quota_outbox_worker.go, quota_outbox.go)

**Change:** Moves conflict-exhausted metric from datastore layer (global) to backend layer (tenant-scoped). New QuotaOutboxBatchClaimResult struct surfaces ConflictExhausted as a structured signal instead of silently returning nil entries.

**Correct:** The claimQuotaOutboxBatchResult retry loop now returns ConflictExhausted=true instead of emitting a global metric and returning nil. ProcessQuotaOutboxBatch records tenant-scoped claim+conflict metrics. The quotaOutboxBatchClaimer function type enables test injection.

**Test:** TestQuotaOutboxBatchConflictExhaustedRecordsTenantMetrics validates the full path — injected claimer returns ConflictExhausted, verifies both metric labels appear with correct tenant_id.

### 2. Tenant pool error classification (pool.go)

**Change:** tenantPoolErrorResult classifies MySQL 1045 (access denied) and 1105+quota message (usage quota exhausted) into labeled results instead of generic "error". Non-error classified results log at Warn instead of Error.

**Correct:** Two-tier classification — first checks mysql.MySQLError struct fields, then falls back to string matching for legacy error message formats. This handles both wrapped and unwrapped error paths.

**Test:** TestTenantPoolErrorResultClassifiesExpectedDatabaseErrors covers 6 cases: auth via error code, auth via legacy message, quota via error code, quota via lowercase message, not_found, and generic error.

### 3. Pending delta cache race (quota_cache.go)

**Change:** Instead of discarding the snapshot when a local add() races with the DB loader (generation mismatch), the cache now tracks localPositiveDeltas and merges the race-window delta into the loaded snapshot.

**Analysis:**
- add() accumulates only positive deltas (maxInt64(delta, 0)) into localPositiveDeltas
- get() captures localPositiveDeltas before loading, then on generation mismatch computes the race-window delta via sub() and adds it to the DB result
- This may double-count for one TTL window (if the DB SELECT already observed the enqueued row), but never undercount — correct tradeoff for quota admission
- Negative deltas (acks) are excluded from localPositiveDeltas, preventing the cache from undercounting quota usage

**Tests:** TestQuotaPendingDeltasCachePublishesConservativeSnapshotWhenLocalDeltaRacesLoad validates positive race merge (10+5=15). TestQuotaPendingDeltasCacheIgnoresNegativeRaceDeltasWhenPublishing validates negative deltas are excluded (stays at 10).

### Minor observations (non-blocking)

1. maxInt64 helper — Go 1.21+ has builtin max(). If the project targets 1.21+, this could use max() directly. Not blocking.

2. ClaimQuotaOutboxBatch (old API) now wraps claimQuotaOutboxBatchResult, preserving backward compat. Clean.

Verdict: APPROVE

qiffang (approved):
Reviewed PR #638 at head `e5d01a6`.

Verdict: APPROVE.

Checks performed:
- Read the full diff for `quota_cache`, `quota_outbox`, and tenant pool classification changes.
- Verified the pending-delta cache race behavior is conservative: positive local race deltas may be double-counted for one TTL, but admission does not undercount pending quota.
- Verified `QuotaOutboxBatchClaimResult` moves claim-conflict observability out of the datastore layer and records tenant-scoped backend metrics for conflict exhaustion.
- Verified tenant pool classification uses typed result constants and MySQL error codes for 1045 / 1105, with legacy message fallback preserved.
- Verified current PR head has CI green: `code-ci` and `local-e2e` completed successfully.
- Local validation: `go test -c ./pkg/backend ./pkg/datastore ./pkg/tenant` and `git diff --check origin/main...HEAD` passed. Runtime package tests are blocked on this machine by local testcontainers setup (`rootless Docker not found`), not by this PR.

No blockers found.

PR #639, opened 2026-06-28
Tune quota worker metrics and create hot path

srstack:
## Summary
- classify transient DB connection failures as `bad_conn` across background workers, event bus, DB instrumentation, and quota outbox claim paths so worker error alerts stop paging on `invalid connection` noise after deploy
- record quota outbox batch latency only at `quota_outbox/process`; per-mutation success/failure metrics are count-only, including fallback failures
- add configurable soft quota admission cache TTLs, defaulting usage and pending-delta caches to 250ms, with visible warnings for malformed or negative duration overrides
- speed up create-if-absent writes (`X-Dat9-Expected-Revision: 0`) by skipping the pre-`Stat` round trip and relying on the path unique constraint to preserve conflict semantics
- use create-specific quota admission so new-file creates no longer query current path size inside the write transaction
- add `idx_quota_outbox_pending_deltas(status, storage_delta, file_delta, media_delta)` so `PendingQuotaOutboxDeltas` can satisfy the pending SUM query from a covering index instead of scanning status and doing row lookups
- keep DB and worker result-label classification aligned by documenting that `mysqlutil.dbResult` delegates to `metrics.ResultForError`

## Operational impact
- `Drive9BackgroundWorkerErrorsHigh` matches `result=~"error|.+_error"`, so `bad_conn` is excluded from worker error paging after this deploy
- DB connection failures remain visible via DB/store operation metrics with `result="bad_conn"`
- CLS check during the active alert window showed tenant `7ed99094-202f-4261-8843-a090b61e672d` was still failing `claim_quota_outbox_batch` with `invalid connection`; on the currently deployed image this still records as generic `error`, so the alert can continue until this PR is deployed
- The create fast path targets write-back new-file uploads where clients send expected revision `0`; ordinary unconditional overwrite behavior still uses the existing stat/overwrite path

## Rollout notes
- The TiDB quota outbox index is emitted as a normal `CREATE INDEX`. For existing tenants with large `quota_outbox` tables, this will run TiDB online add-index backfill and consume TiKV resources while it catches up
- Re-running schema sync should be safe: duplicate index errors are treated as ignorable by the existing schema repair flow, but rollout should still be staged if old tenant outbox tables are large

## Performance notes
- This is the low-risk part of the QPS work: reduce fixed DB round trips on create and make the pending outbox aggregate cheaper
- It targets the observed `pending_quota_outbox_deltas` cost and removes one provably unnecessary create-time lookup
- It does not reduce TiDB per-transaction 2PC/commit latency such as the observed `in_tx≈514ms` slow samples; if that remains dominant under load, micro-batching/group commit is the larger follow-up

## Tests
- `GOCACHE=/tmp/drive9-go-cache go test ./pkg/tenant/schema ./cmd/drive9-server -run 'TestQuotaOutboxMissingFileOrderIndexIsRepairable|TestSchemaDumpInitSQLByProvider|TestSchemaDumpInitSQLByProviderIncludesVault|TestSchemaDumpInitSQLUsesTiDBAutoEmbeddingEnv' -count=1`
- `GOCACHE=/tmp/drive9-go-cache go test ./pkg/metrics ./pkg/backend ./pkg/server ./pkg/mysqlutil ./cmd/drive9-server ./cmd/drive9-server-local -count=1`
- `GOCACHE=/tmp/drive9-go-cache go test ./pkg/metrics ./pkg/mysqlutil -count=1`
- `GOCACHE=/tmp/drive9-go-cache GOLANGCI_LINT_CACHE=/tmp/drive9-golangci-cache make lint`

## Summary by CodeRabbit

* **New Features**
  * Added support for configuring quota cache TTLs at startup through environment variables, with clearer command-line help.
  * Improved create-if-absent write behavior so existing paths return a conflict faster, before quota checks.

* **Bug Fixes**
  * Standardized error reporting in metrics and logs for database, event, quota, and background processing operations.
  * Updated quota handling to better distinguish create-only writes and avoid unnecessary quota validation in conflict cases.

### DIFF
--- a/pkg/backend/dat9.go
+++ b/pkg/backend/dat9.go
@@ -686,6 +686,25 @@ func (b *Dat9Backend) WriteCtxIfRevisionWithTagsResult(ctx context.Context, path
 		return 0, 0, err
 	}
 
+	if isCreateIfAbsentWrite(offset, flags, expectedRevision) {
+		operation = "create"
+		implementationStart := time.Time{}
+		if timingEnabled {
+			implementationStart = time.Now()
+		}
+		n, err := b.createAndWriteCtx(ctx, path, data, tags, description)
+		if timingEnabled {
+			implementationDuration = time.Since(implementationStart)
+		}
+		if errors.Is(err, datastore.ErrPathConflict) {
+			return 0, 0, datastore.ErrRevisionConflict
+		}
+		if err != nil {
+			return 0, 0, err
+		}
+		return n, 1, nil
+	}
+
 	statStart := time.Time{}
 	if timingEnabled {
 		statStart = time.Now()
@@ -746,6 +765,13 @@ func (b *Dat9Backend) WriteCtxIfRevisionWithTagsResult(ctx context.Context, path
 		implementationDuration = time.Since(implementationStart)
 	}
 	return n, rev, err
+}
+
+func isCreateIfAbsentWrite(offset int64, flags filesystem.WriteFlag, expectedRevision int64) bool {
+	return expectedRevision == 0 &&
+		offset == 0 &&
+		flags&filesystem.WriteFlagCreate != 0 &&
+		flags&filesystem.WriteFlagAppend == 0
 }
 
 func backendWriteResult(err error) string {
@@ -904,7 +930,7 @@ func (b *Dat9Backend) createAndWriteCtx(ctx context.Context, path string, data [
 		quotaOutboxEnqueued = false
 
 		done := backendTimingStep(timingEnabled, &quotaStorageCheckDuration)
-		err := b.ensureStorageQuota(ctx, tx, path, int64(len(data)))
+		err := b.ensureCreateStorageQuota(ctx, tx, int64(len(data)))
 		done()
 		if err != nil {
 			return err
--- a/pkg/backend/dat9.go
+++ b/pkg/backend/dat9.go
@@ -688,6 +688,13 @@ func (b *Dat9Backend) WriteCtxIfRevisionWithTagsResult(ctx context.Context, path
 
 	if isCreateIfAbsentWrite(offset, flags, expectedRevision) {
 		operation = "create"
+		exists, err := b.store.NodeExists(ctx, path)
+		if err != nil {
+			return 0, 0, err
+		}
+		if exists {
+			return 0, 0, datastore.ErrRevisionConflict
+		}
 		implementationStart := time.Time{}
 		if timingEnabled {
 			implementationStart = time.Now()
--- a/cmd/drive9-server-local/main.go
+++ b/cmd/drive9-server-local/main.go
@@ -842,6 +842,11 @@ func envDuration(key string, fallback time.Duration) time.Duration {
 	}
 	v, err := time.ParseDuration(raw)
 	if err != nil {
+		fmt.Fprintf(os.Stderr, "invalid %s=%q: %v; using %s\n", key, raw, err, fallback)
+		return fallback
+	}
+	if v < 0 {
+		fmt.Fprintf(os.Stderr, "invalid %s=%q: duration must be non-negative; using %s\n", key, raw, fallback)
 		return fallback
 	}
 	return v
--- a/cmd/drive9-server-local/main.go
+++ b/cmd/drive9-server-local/main.go
@@ -127,6 +127,10 @@ func main() {
 	logLocalStartupStep(startupCtx, startupStart, stepStart, "open_local_datastore")
 
 	stepStart = time.Now()
+	backend.InitQuotaAdmissionCacheTTLs(
+		envDuration("DRIVE9_QUOTA_USAGE_CACHE_TTL", 0),
+		envDuration("DRIVE9_QUOTA_PENDING_DELTAS_CACHE_TTL", 0),
+	)
 	backendOpts, err := buildBackendOptionsFromEnv()
 	if err != nil {
 		die(err)
@@ -366,6 +370,8 @@ environment:
   DRIVE9_VAULT_MASTER_KEY 32-byte hex key for vault DEK wrapping (omit to disable vault)
   DRIVE9_LOG_LEVEL debug|info|warn|error (default: info)
   DRIVE9_BENCH_TIMING_LOG_ENABLED true|false to emit benchmark timing logs on successful server hot paths (default: false)
+  DRIVE9_QUOTA_USAGE_CACHE_TTL soft small-write central usage cache TTL, e.g. 250ms or 1s
+  DRIVE9_QUOTA_PENDING_DELTAS_CACHE_TTL soft small-write tenant pending-outbox aggregate cache TTL, e.g. 250ms or 1s
 
   S3 storage:
   Set DRIVE9_S3_BUCKET to enable AWS S3 mode.
@@ -823,6 +829,18 @@ func envInt64(key string, fallback int64) int64 {
 		return fallback
 	}
 	v, err := strconv.ParseInt(raw, 10, 64)
+	if err != nil {
+		return fallback
+	}
+	return v
+}
+
+func envDuration(key string, fallback time.Duration) time.Duration {
+	raw := strings.TrimSpace(os.Getenv(key))
+	if raw == "" {
+		return fallback
+	}
+	v, err := time.ParseDuration(raw)
 	if err != nil {
 		return fallback
 	}
--- a/cmd/drive9-server/main.go
+++ b/cmd/drive9-server/main.go
@@ -125,6 +125,10 @@ func main() {
 	// P1-3: Configurable quota cache refresh interval (default 30s).
 	// In multi-pod deployments, increasing this reduces per-tenant-per-pod DB reads.
 	backend.InitQuotaConfigCacheRefreshInterval(envInt("DRIVE9_QUOTA_CACHE_REFRESH_SECONDS", 0))
+	backend.InitQuotaAdmissionCacheTTLs(
+		envDuration("DRIVE9_QUOTA_USAGE_CACHE_TTL", 0),
+		envDuration("DRIVE9_QUOTA_PENDING_DELTAS_CACHE_TTL", 0),
+	)
 
 	store, err := openControlPlaneStoreWithRetry(context.Background(), metaDSN, defaultStartupRetryOptions())
 	if err != nil {
@@ -478,6 +482,8 @@ environment:
   DRIVE9_LOG_LEVEL debug|info|warn|error (default: info)
   DRIVE9_BENCH_TIMING_LOG_ENABLED true|false to emit benchmark timing logs on successful server hot paths (default: false)
   DRIVE9_QUOTA_SOURCE tenant|server quota enforcement source (default: tenant)
+  DRIVE9_QUOTA_USAGE_CACHE_TTL soft small-write central usage cache TTL, e.g. 250ms or 1s
+  DRIVE9_QUOTA_PENDING_DELTAS_CACHE_TTL soft small-write tenant pending-outbox aggregate cache TTL, e.g. 250ms or 1s
   DRIVE9_DISABLE_AUTO_EMBEDDING true|false disable TiDB database-managed auto-embedding (default: false)
                                 set to true when the TiDB Cloud cluster has no supported embedding provider
   DRIVE9_TIDB_AUTO_EMBEDDING_MODEL TiDB EMBED_TEXT model for auto-embedding generated columns
@@ -865,6 +871,18 @@ func envInt64(key string, fallback int64) int64 {
 		return fallback
 	}
 	v, err := strconv.ParseInt(raw, 10, 64)
+	if err != nil {
+		return fallback
+	}
+	return v
+}
+
+func envDuration(key string, fallback time.Duration) time.Duration {
+	raw := strings.TrimSpace(os.Getenv(key))
+	if raw == "" {
+		return fallback
+	}
+	v, err := time.ParseDuration(raw)
 	if err != nil {
 		return fallback
 	}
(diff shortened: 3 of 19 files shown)


srstack:
Update pushed: bf9b292e

What changed in this follow-up:
- Added a create-if-absent fast path for `expected_revision=0` writes. This avoids the pre-`Stat` round trip for write-back new-file uploads and maps insert path conflicts back to `ErrRevisionConflict` as before.
- Added create-only quota admission helpers so create transactions no longer query current path size; for create, current path size is known to be zero and path conflicts roll back the transaction.
- Added `idx_quota_outbox_pending_deltas(status, storage_delta, file_delta, media_delta)` to make `PendingQuotaOutboxDeltas` SUM queries covering instead of scanning status and doing row lookups for each pending row.
- Addressed review feedback: invalid quota TTL env values are now visible, negative TTLs are rejected/logged, mutation replay per-entry metrics are count-only, and quota_outbox fallback failures now record duration only on `quota_outbox/process` while mutation labels remain count-only.

Validation run locally:
- `GOCACHE=/tmp/drive9-go-cache go test ./pkg/tenant/schema ./cmd/drive9-server -run 'TestQuotaOutboxMissingFileOrderIndexIsRepairable|TestSchemaDumpInitSQLByProvider|TestSchemaDumpInitSQLByProviderIncludesVault|TestSchemaDumpInitSQLUsesTiDBAutoEmbeddingEnv' -count=1`\n- `GOCACHE=/tmp/drive9-go-cache go test ./pkg/metrics ./pkg/backend ./pkg/server ./pkg/mysqlutil ./cmd/drive9-server ./cmd/drive9-server-local -count=1`\n- `GOCACHE=/tmp/drive9-go-cache GOLANGCI_LINT_CACHE=/tmp/drive9-golangci-cache make lint`\n\nOperational note from CLS: the continuing `Drive9BackgroundWorkerErrorsHigh` alert for tenant `7ed99094-202f-4261-8843-a090b61e672d` is still `quota_outbox_process_failed` / `claim_quota_outbox_batch` / `invalid connection`. That will continue to count as `result="error"` until this PR is deployed; after this PR it should classify as `bad_conn`, which the alert regex excludes.

srstack (commented):
Reviewed the full diff, with extra attention on the `perf: trim quota create hot path` commit (landed after the bot reviews). This directly targets the two hot spots from the slow-PUT trace (`in_tx` and `pending_quota_outbox_deltas`). LGTM overall — the design is sound and well-tested. A few notes, one of them an operational caveat for rollout.

### Verified correct

**Create fast-path (`ensureCreateStorageQuota`).** The math is provably equivalent to the general path, not just approximately. `ensureTenantStorageQuotaTx` computes `delta = newSize - currentPathBytes` (clamped ≥0); for a create-if-absent the path has no dentry yet so `currentPathBytes == 0` → `delta == newSize`, which is exactly what the new variant uses. It drops the now-provably-zero `ConfirmedFileSizeByPathTx` round-trip — a real per-create saving with no behavioral change. The server-quota branch is analogous (`delta == newSize` for a create). `isCreateIfAbsentWrite` is appropriately strict (rev 0, offset 0, Create set, Append unset), and `ErrPathConflict → ErrRevisionConflict` keeps the concurrent-create semantics intact.

**Covering index `idx_quota_outbox_pending_deltas(status, storage_delta, file_delta, media_delta)`.** This is the right fix for the 157ms `SUM(...) WHERE status IN (queued,processing)` — the three SUM columns are now in the index, so it's an index-only scan with no row lookups.

**Metric de-duplication.** Switching per-entry applied-outbox records to `RecordTenantOperationCount` (count-only) fixes a real bug: the old code stamped one batch `time.Since(start)` onto every entry's duration histogram, inflating per-mutation latency. Adding an explicit `process/ok` record also completes the `process` outcome distribution.

**`bad_conn` classification.** Consistent `ResultForError` use across the workers, and the alert reasoning checks out: `Drive9BackgroundWorkerErrorsHigh` matches `result=~"error|.+_error"`, so `bad_conn` is excluded from worker paging while DB-level `result="bad_conn"` keeps it observable. The negative-TTL guard + warning log is a nice touch.

### Notes / suggestions

1. **Rollout caveat (operational, not code):** for TiDB, `idx_quota_outbox_pending_deltas` is a bare `CREATE INDEX` (no `IF NOT EXISTS`). On existing tenants with a large `quota_outbox` this triggers an online add-index DDL that consumes TiKV resources while backfilling. It's safe (re-runs are caught by `isIgnorableSchemaError` on "duplicate key name"), but worth sequencing/announcing if any tenant's outbox table is big. Confirm this lands via the normal ensure-schema-on-open path for already-provisioned tenants.

2. **`ResultForError` duplicates `mysqlutil.dbResult`** (same switch incl. the `"invalid connection"` substring). Fine to keep separate to avoid a `metrics → mysqlutil` dependency, but a one-line comment cross-referencing them would help keep the two in sync if the classification ever changes.

3. **Scope check on the perf win:** this PR attacks the 157ms pending-delta query (index) and shaves one round-trip per create — both good. But the dominant cost in the trace is `in_tx≈514ms`, which is TiDB per-transaction commit latency and is untouched here (expected — that's a separate batching effort). Worth stating in the PR that this is the "cheap, high-confidence" slice and the per-create commit latency is tracked separately, so reviewers don't expect the 514ms to move.

Tests are good — the create fast-path, conflict-exhausted, and bad_conn paths all have coverage. Waiting on CI (CodeRabbit already green).

qiffang (approved):
## Review: PR #639 — Tune quota worker metrics and create hot path

### Overall: APPROVE

Well-structured performance and observability improvement. 395 additions, 45 deletions, 17 files. Four distinct changes, all clean.

---

### 1. Centralized error-to-metric-label classification (metrics/operations.go)

ResultForError correctly classifies context.Canceled, context.DeadlineExceeded, driver.ErrBadConn, and "invalid connection" string match into stable labels. The mysqlutil.dbResult now delegates to this, ensuring DB and worker labels stay aligned. Applied consistently across all background workers: expiry sweep, file GC, mutation replay, event bus, SSE publish, quota outbox.

Test coverage: TestResultForError covers nil, canceled, deadline (wrapped), bad conn sentinel, invalid connection string (case-insensitive), and generic error.

### 2. Per-entry count-only metrics (RecordTenantOperationCount)

RecordTenantOperationCount increments the counter without adding a duration sample — correct for per-entry outbox mutations where one wall-clock duration would be duplicated. recordAppliedQuotaOutboxEntries no longer takes a start time. Process-level duration is recorded once at the batch level.

Test coverage: TestRecordTenantOperationCountDoesNotRecordDuration verifies counter present but duration absent. TestServerQuotaOutboxBatchProcessesDifferentFiles and TestQuotaOutboxFallbackFailureRecordsProcessDurationOnly verify the per-entry vs batch-level metric split.

### 3. Configurable quota admission cache TTLs

InitQuotaAdmissionCacheTTLs validates inputs (negative = warn, zero = keep default, positive = override). Default bumped from 100ms to 250ms. Both server binaries call it from env vars. envDuration helper is duplicated across both main.go files — minor, not blocking.

### 4. Create-if-absent hot path (dat9.go)

isCreateIfAbsentWrite detects expectedRevision=0, offset=0, Create flag, no Append — a new file write. This skips the pre-stat (path lookup) and uses ensureCreateStorageQuota which skips the current path size lookup (known zero for creates). ErrPathConflict correctly maps to ErrRevisionConflict.

ensureTenantCreateStorageQuotaTx is a clean simplified version — no path-based size lookup, just confirmed + reserved + newSize vs limit.

Test coverage: TestWriteCreateIfAbsentSkipsPreStat verifies both success (200) and duplicate (409), and checks timing log shows operation=create with stat_ms=0.

### 5. New covering index

idx_quota_outbox_pending_deltas on (status, storage_delta, file_delta, media_delta) covers the SUM query for pending deltas cache. Added to both TiDB and DB9 schema statements.

### Minor observations (non-blocking)

1. envDuration is duplicated between drive9-server/main.go and drive9-server-local/main.go. Could be extracted to a shared util, but not worth the churn for two copies.

2. The eventbus.go whitespace alignment fix is cosmetic but clean.

Verdict: APPROVE

pkg/backend/dat9.go
```diff
@@ -686,6 +686,25 @@ func (b *Dat9Backend) WriteCtxIfRevisionWithTagsResult(ctx context.Context, path
 		return 0, 0, err
 	}
 
+	if isCreateIfAbsentWrite(offset, flags, expectedRevision) {
+		operation = "create"
+		implementationStart := time.Time{}
+		if timingEnabled {
+			implementationStart = time.Now()
+		}
+		n, err := b.createAndWriteCtx(ctx, path, data, tags, description)
```
qiffang, line 702: Blocking: this fast path skips the pre-`Stat`, but `createAndWriteCtx` runs file-size/storage quota checks before `InsertNodeTx` can detect `ErrPathConflict`. For an existing path with `expectedRevision == 0`, the old path returned `ErrRevisionConflict` immediately after `Stat`; now a full storage quota or max-file-size condition can win first and surface as quota failure (HTTP 507) instead of conflict (HTTP 409). That changes the create-if-absent API contract for idempotent clients. Please preserve occupied-path conflict precedence, and add a regression for duplicate create under storage/file-size quota pressure.

qiffang (changes requested):
Blocking issue found in the create-if-absent fast path: it changes the error precedence for occupied paths when quota checks fail before the path conflict is detected. Existing tests cover duplicate create only without quota pressure, so this regression is currently untested.

qiffang (changes requested):
**Updating from APPROVE to REQUEST_CHANGES** — @adversary-1's blocker is valid.

## B1: Create-if-absent fast path changes error precedence (Blocker)

**Old path** (with `Stat`):
1. `Stat(path)` → file exists → `expectedRevision == 0` → immediately returns `ErrRevisionConflict` (HTTP 409)
2. Quota is never checked for existing paths

**New fast path** (skips `Stat`):
1. Enters `createAndWriteCtx` directly
2. `ensureCreateStorageQuota` runs first → if quota full → returns quota error (HTTP 507)
3. `InsertNodeTx` never reached → 409 never returned

For clients: a create-if-absent write to an **existing path** when quota is exhausted will now return 507 (need to free space) instead of 409 (already exists, retry-safe). This changes the observable API contract.

**Secondary**: the fast path also performs an S3 PutObject before `InsertNodeTx` detects the path conflict — wasted upload for existing paths. Not a correctness bug, but unnecessary cost.

**Fix direction**: the fast path should either (a) do a lightweight existence check before quota admission, or (b) reorder the transaction to run `InsertNodeTx` before quota checks (though this may conflict with FK constraints).

— dat9-dev1

srstack:
Addressed the REQUEST_CHANGES blocker in db4318ac.

What changed:
- Create-if-absent fast path now performs a lightweight `NodeExists` dentry lookup before quota admission/storage work.
- Existing occupied paths now preserve the old precedence: duplicate create returns `ErrRevisionConflict` / HTTP 409 even when storage quota or max-file-size checks would otherwise fail.
- This also avoids doing S3/object work for an already occupied path on the create fast path.
- Added regressions for duplicate create under storage quota pressure and max-file-size pressure.

Validation:
- `GOCACHE=/tmp/drive9-go-cache go test ./pkg/backend ./pkg/server ./pkg/datastore -run 'TestCreateIfAbsentExistingPathReturnsConflictBeforeStorageQuota|TestCreateIfAbsentExistingPathReturnsConflictBeforeFileSizeQuota|TestWriteCreateIfAbsentSkipsPreStat|TestWriteWithExpectedRevisionConflict|TestInsertAndGetNode' -count=1`\n- `GOCACHE=/tmp/drive9-go-cache go test ./pkg/server -run 'TestWriteCreateIfAbsentSkipsPreStat|TestWriteWithExpectedRevisionConflict' -count=1`\n- `GOCACHE=/tmp/drive9-go-cache go test ./pkg/metrics ./pkg/backend ./pkg/datastore ./pkg/mysqlutil ./cmd/drive9-server ./cmd/drive9-server-local -count=1`\n- `GOCACHE=/tmp/drive9-go-cache GOLANGCI_LINT_CACHE=/tmp/drive9-golangci-cache make lint`

qiffang (approved):
**Updating from REQUEST_CHANGES to APPROVE** — previous blocker (error precedence inversion) is fixed.

The create-if-absent fast path now calls `NodeExists(ctx, path)` before `createAndWriteCtx`:
- Lightweight `SELECT 1 FROM file_nodes WHERE path_hash = ? AND path = ?` — cheaper than full Stat
- If node exists → returns ErrRevisionConflict (409) immediately, before quota check
- TOCTOU race handled by ErrPathConflict → ErrRevisionConflict mapping

Two regression tests directly verify the blocker:
- `TestCreateIfAbsentExistingPathReturnsConflictBeforeStorageQuota`
- `TestCreateIfAbsentExistingPathReturnsConflictBeforeFileSizeQuota`

All 5 changes approved. CI green.

— dat9-dev1

qiffang (approved):
adversary-1 re-review on `db4318ac06ec63620c37e4601c57e62878b9724c`.

Verdict: GREEN / APPROVE.

The previous blocker is fixed. The create-if-absent fast path now runs the lightweight `NodeExists(ctx, path)` probe before `createAndWriteCtx`; for an already occupied path it returns `ErrRevisionConflict` before file-size/storage quota checks, preserving the pre-PR 409-before-507 behavior. The fallback `ErrPathConflict -> ErrRevisionConflict` mapping still covers the normal insert-conflict race after the preflight.

Regression coverage is adequate for the blocked scenario:
- `TestCreateIfAbsentExistingPathReturnsConflictBeforeStorageQuota`
- `TestCreateIfAbsentExistingPathReturnsConflictBeforeFileSizeQuota`

I also re-swept the quota admission caches, quota outbox batch/count-duration metric split, `ResultForError` label normalization, schema index, and changed event/GC/replay metric paths. I did not find a correctness blocker. CodeRabbit's remaining `NodeExists` error-wrap and metrics-baseline comments are maintainability/test-isolation nits, not merge blockers for this PR.

Validation: GitHub `ci`, `drive9-server-local e2e smoke`, and CodeRabbit are green on this head; local `git diff --check origin/main...HEAD` passed. Local Go tests were not runnable in my environment because `go`/`gofmt` are not installed.

Issue #641, opened 2026-06-29
PRD: Drive9 e2e product-quality feedback loop


SabaPing:
# PRD: Drive9 E2E Product Quality Loop

## Problem Statement

Drive9 already has a broad e2e system, but the current signal is not yet shaped into a continuous product-quality loop. The suite validates many behaviors, yet failures and performance regressions are hard to turn into fast decisions because the signal is scattered across workflow logs, generic GitHub issue comments, optional artifacts, and manually interpreted run summaries.

From the user's perspective, this means Drive9 can have substantial automated coverage while still making it hard to answer: what product promise regressed, who should act, whether the failure blocks release quality, and whether the team has been notified in the place where work actually happens.

The immediate product need is to make Drive9 e2e a living quality system: it should detect correctness failures and performance regressions, summarize them in product language, push urgent items to Feishu/Lark, and give maintainers a dashboard for continuous improvement.

## Solution

Build a Drive9 e2e product-quality feedback loop.

The first version should add structured e2e summaries, a product-quality inventory, actionable GitHub issue updates, and Feishu/Lark notifications for failed runs or explicit performance regressions. The goal is not to rewrite every e2e script at once. The goal is to create one reliable reporting contract that existing suites can adopt incrementally.

The resulting workflow should let a Drive9 maintainer see:

- which suite failed;
- which product promise is affected;
- whether the failure came from PR, post-merge, nightly, or manual all;
- whether it is correctness, infrastructure, dependency, flaky, or performance regression;
- what artifact/report/log link explains it;
- who should likely own the next step;
- whether the failure was pushed to Feishu/Lark.

## User Stories

1. As a Drive9 maintainer, I want every e2e failure summarized in product language, so that I can decide whether it threatens Drive9 quality.
2. As a Drive9 maintainer, I want performance regressions pushed only when they cross explicit budgets, so that noisy hosted-runner metrics do not create alert fatigue.
3. As a PR author, I want a failed PR e2e run to identify the failed suite and failure class, so that I do not dig through a full workflow log first.
4. As a PR author, I want the PR gate to stay fast, so that e2e improvement does not make normal development painful.
5. As a backend developer, I want each e2e script mapped to a product promise, so that I understand why a script exists and whether it is redundant.
6. As a backend developer, I want new e2e scripts to declare their tier and product area, so that they are wired into the right trigger instead of becoming dead code.
[...]

PR #642, opened 2026-06-29
e2e: product-quality feedback loop + Feishu notifications (#641, slice 1)

SabaPing:
Implements the **first vertical slice** of #641 — turning raw e2e signal into a product-quality loop, plus Feishu/Lark notifications for the failures that matter.

## What this does
1. **Structured suite-summary contract + run aggregator** (`internal/e2ereport`, pure & dependency-free): per-suite status / tier / product area+promise / failure class / owner hint / metrics (budget+baseline), aggregated into a `RunReport` with a stable **failure signature**.
2. **`e2e/suite-manifest.json`**: product metadata for each `local-e2e.yml` suite (keyed by step id) so the aggregator can synthesize summaries for suites that haven't adopted the JSON contract yet — incremental adoption.
3. **`cmd/e2e-aggregate`**: reads the manifest + step outcomes (+ any adopted `e2e/reports/summary/*.json`) → richer step summary (grouped by product promise), aggregated JSON, a signature-led GitHub issue body, and workflow outputs (`notify`/`signature`/`overall_success`/…).
4. **Upgraded issue filing**: the existing `ci-e2e-failure` step now uses the structured body and **groups recurring failures by signature** instead of piling everything onto one generic issue.
5. **Feishu/Lark notifications** (`internal/feishu` + `cmd/feishu-notify`): interactive card with signature, failed suites, owners, perf regressions, and a run link. **Dual transport, auto-detected from secrets** — custom-bot webhook (`FEISHU_WEBHOOK`) or app API (`FEISHU_APP_ID`/`FEISHU_APP_SECRET`/`FEISHU_CHAT_ID`).

## Notification policy
PR-tier failures stay in GitHub. Post-merge / nightly / manual failures and explicit performance regressions push to Feishu. Successful runs never notify.

## Safety / blast radius
- Workflow changes are **additive and non-gating**: the existing suite steps and the `Enforce smoke results` gate are untouched. The Feishu step is `continue-on-error` and gated on the aggregator's `notify` output.
- The notifier **no-ops (never fails the run)** when no Feishu secret is configured, so forks/PRs without secrets are unaffected.

## Tests
- `internal/e2ereport`: parse, aggregation, signature stability (order-independent), notify policy, perf-regression detection, manifest synthesis (adopted overrides synthesized), event→tier mapping, rendering.
- `internal/feishu`: transport selection, env parsing, payload shape, and `httptest` round-trips for **both** webhook and app-API transports.
- `gofmt`, `go vet`, and `golangci-lint v2.5.0` clean on all new packages.
- **Live-validated**: a synthesized nightly-failure card was sent end to end via the app-API transport and delivered to the target chat.

## To activate live CI delivery
Add repo secrets (app API): `FEISHU_APP_ID`, `FEISHU_APP_SECRET`, `FEISHU_CHAT_ID` — or `FEISHU_WEBHOOK` for the custom-bot path.

## Follow-ups (out of slice)
Suite adoption of the structured summary JSON (durations + soft-regression metrics + artifact links), e2e inventory/dashboard, and scenario-coverage/dedup analysis (later steps of #641).

Refs #641.

## Summary by CodeRabbit

* **New Features**
  * Added product-quality e2e aggregation that produces unified JSON/markdown reports with stable failure signatures and performance regression detection.
  * Added Feishu/Lark notifications driven by the aggregated report, including signature-led routing and conditional sending.
  * Introduced a CI toggle to enable Feishu proof behavior and improved degraded-pipeline detection/issue grouping.
* **Documentation**
  * Documented the new product-quality report flow, suite opt-in via suite summaries/manifest, and notification behavior.
* **Tests**
  * Added unit tests covering aggregation/tier-signature behavior, notification policy decisions, and Feishu card/message handling.

### DIFF
--- a/.github/workflows/local-e2e.yml
+++ b/.github/workflows/local-e2e.yml
@@ -478,6 +478,53 @@ jobs:
             echo "pjdfstest POSIX suite: \`RUN_POSIX_FEATURE_MATRIX=${RUN_POSIX_FEATURE_MATRIX}\`"
           } >> "$GITHUB_STEP_SUMMARY"
 
+      # Turn the raw per-suite outcomes into one product-quality report: a richer
+      # step summary grouped by product promise, an aggregated JSON for the
+      # notifier, a signature-led issue body, and outputs (notify/signature/...).
+      - name: Aggregate e2e product-quality report
+        id: aggregate
+        if: always()
+        run: |
+          cat > /tmp/e2e-outcomes.json <<JSON
+          {
+            "api-smoke": "${{ steps.api-smoke.outcome }}",
+            "existing-key-smoke": "${{ steps.existing-key-smoke.outcome }}",
+            "cli-smoke": "${{ steps.cli-smoke.outcome }}",
+            "layer-fs-smoke": "${{ steps.layer-fs-smoke.outcome }}",
+            "fuse-smoke": "${{ steps.fuse-smoke.outcome }}",
+            "git-ops-smoke": "${{ steps.git-ops-smoke.outcome }}",
+            "portable-pack-unpack-e2e": "${{ steps.portable-pack-unpack-e2e.outcome }}",
+            "fuse-crash-recovery": "${{ steps.fuse-crash-recovery.outcome }}",
+            "fuse-write-perf-budget": "${{ steps.fuse-write-perf-budget.outcome }}",
+            "fuse_performance_compare": "${{ steps.fuse_performance_compare.outcome }}",
+            "fuse-concurrency-stress": "${{ steps.fuse-concurrency-stress.outcome }}",
+            "fuse-posix-fsx": "${{ steps.fuse-posix-fsx.outcome }}",
+            "smoke-all": "${{ steps.smoke-all.outcome }}",
+            "git-feature-matrix": "${{ steps.git-feature-matrix.outcome }}",
+            "pjdfstest-suite": "${{ steps.pjdfstest-suite.outcome }}"
+          }
+          JSON
+          go run ./cmd/e2e-aggregate \
+            --manifest e2e/suite-manifest.json \
+            --outcomes /tmp/e2e-outcomes.json \
+            --summaries e2e/reports/summary \
+            --out /tmp/e2e-report.json \
+            --issue-body /tmp/e2e-issue-body.md
+
+      # Push failures/regressions that warrant attention to Feishu/Lark. PR-tier
+      # failures stay in GitHub by policy; the notifier no-ops when no Feishu
+      # secret is configured, so this never blocks a fork/PR without secrets.
+      - name: Notify Feishu on post-merge/nightly failure or perf regression
+        if: always() && steps.aggregate.outputs.notify == 'true'
+        continue-on-error: true
+        env:
+          FEISHU_APP_ID: ${{ secrets.FEISHU_APP_ID }}
+          FEISHU_APP_SECRET: ${{ secrets.FEISHU_APP_SECRET }}
+          FEISHU_CHAT_ID: ${{ secrets.FEISHU_CHAT_ID }}
+          FEISHU_WEBHOOK: ${{ secrets.FEISHU_WEBHOOK }}
+        run: |
+          go run ./cmd/feishu-notify --report /tmp/e2e-report.json
+
       - name: Dump TiDB playground logs
         if: always()
         run: |
@@ -560,16 +607,37 @@ jobs:
           RUN_URL: ${{ github.server_url }}/${{ github.repository }}/actions/runs/${{ github.run_id }}
           EVENT_NAME: ${{ github.event_name }}
           SHA: ${{ github.sha }}
+          SIGNATURE: ${{ steps.aggregate.outputs.signature }}
         run: |
           gh label create ci-e2e-failure --repo "$REPO" \
             --description "Automated: local-e2e scheduled/post-merge run failed" \
             --color D93F0B --force || true
-          body="local-e2e \`$EVENT_NAME\` run failed on \`main\` at $SHA: $RUN_URL"
-          existing="$(gh issue list --repo "$REPO" --label ci-e2e-failure --state open \
-            --json number --jq '.[0].number // empty')"
+
+          # Prefer the structured, signature-led body from the aggregator; fall
+          # back to a one-liner if it is missing for any reason.
+          if [ -s /tmp/e2e-issue-body.md ]; then
+            body="$(cat /tmp/e2e-issue-body.md)"
+            body="$body"$'\n\n'"_run: $RUN_URL_"
+          else
+            body="local-e2e \`$EVENT_NAME\` run failed on \`main\` at $SHA: $RUN_URL"
+          fi
+
+          # Group recurring failures by signature: an open issue carrying the same
+          # signature is reused; a new signature opens its own tracking issue. With
+          # no signature we keep the legacy single-issue behaviour.
+          if [ -n "$SIGNATURE" ]; then
+            existing="$(gh issue list --repo "$REPO" --label ci-e2e-failure --state open \
+              --search "in:title $SIGNATURE" --json number --jq '.[0].number // empty')"
+            title="local-e2e failure on main [sig: $SIGNATURE]"
+          else
+            existing="$(gh issue list --repo "$REPO" --label ci-e2e-failure --state open \
+              --json number --jq '.[0].number // empty')"
+            title="local-e2e $EVENT_NAME run failing on main"
+          fi
+
           if [ -n "$existing" ]; then
             gh issue comment "$existing" --repo "$REPO" --body "$body"
           else
             gh issue create --repo "$REPO" --label ci-e2e-failure \
-              --title "local-e2e $EVENT_NAME run failing on main" --body "$body"
+              --title "$title" --body "$body"
           fi
--- a/.github/workflows/local-e2e.yml
+++ b/.github/workflows/local-e2e.yml
@@ -497,6 +497,10 @@ jobs:
       - name: Aggregate e2e product-quality report
         id: aggregate
         if: always()
+        env:
+          # True when a prior step hard-failed (e.g. TiDB/server setup) before the
+          # suites produced outcomes; the aggregator then fails closed.
+          AGG_DEGRADED: ${{ failure() }}
         run: |
           cat > /tmp/e2e-outcomes.json <<JSON
           {
@@ -521,13 +525,17 @@ jobs:
           if [ "${FEISHU_PROOF}" = "1" ]; then
             proof_arg="--proof-fail-suite e2e-feishu-proof"
           fi
+          degraded_arg=""
+          if [ "${AGG_DEGRADED}" = "true" ]; then
+            degraded_arg="--degraded"
+          fi
           go run ./cmd/e2e-aggregate \
             --manifest e2e/suite-manifest.json \
             --outcomes /tmp/e2e-outcomes.json \
             --summaries e2e/reports/summary \
             --out /tmp/e2e-report.json \
             --issue-body /tmp/e2e-issue-body.md \
-            $proof_arg
+            $proof_arg $degraded_arg
 
       # Push failures/regressions that warrant attention to Feishu/Lark. PR-tier
       # failures stay in GitHub by policy; the notifier no-ops when no Feishu
--- a/.github/workflows/local-e2e.yml
+++ b/.github/workflows/local-e2e.yml
@@ -39,6 +39,10 @@ on:
         required: false
         default: "0"
         type: string
+      feishu_proof:
+        required: false
+        default: "0"
+        type: string
   workflow_dispatch:
     inputs:
       run_all_e2e:
@@ -129,6 +133,14 @@ on:
         options:
           - "0"
           - "1"
+      feishu_proof:
+        description: "Prove the Feishu notification path: inject a synthetic failed suite so a card is sent on this (green) run. Does not fail the job or file an issue."
+        required: false
+        default: "0"
+        type: choice
+        options:
+          - "0"
+          - "1"
 
 permissions:
   contents: read
@@ -164,6 +176,7 @@ env:
   # Deterministic perf gating lives in e2e/fuse-write-perf-budget-test.sh.
   FUSE_PERF_COMPARE_FAIL_ON_REGRESSION: "0"
   FUSE_CONCURRENCY_STRESS_REQUIRED: ${{ (github.event_name == 'schedule' || github.event_name == 'push' || ((github.event_name == 'workflow_dispatch' || github.event_name == 'workflow_call') && (inputs.run_all_e2e == '1' || inputs.run_fuse_concurrency_stress == '1'))) && '1' || '0' }}
+  FEISHU_PROOF: ${{ (github.event_name == 'workflow_dispatch' || github.event_name == 'workflow_call') && inputs.feishu_proof == '1' && '1' || '0' }}
   FUSE_PERF_ARTIFACT_DIR: ${{ github.workspace }}/e2e-artifacts/fuse-performance
   FUSE_CONCURRENCY_ARTIFACT_DIR: ${{ github.workspace }}/e2e-artifacts/fuse-concurrency
   FEATURE_MATRIX_REPORT_DIR: ${{ github.workspace }}/e2e-artifacts/feature-matrix
@@ -504,12 +517,17 @@ jobs:
             "pjdfstest-suite": "${{ steps.pjdfstest-suite.outcome }}"
           }
           JSON
+          proof_arg=""
+          if [ "${FEISHU_PROOF}" = "1" ]; then
+            proof_arg="--proof-fail-suite e2e-feishu-proof"
+          fi
           go run ./cmd/e2e-aggregate \
             --manifest e2e/suite-manifest.json \
             --outcomes /tmp/e2e-outcomes.json \
             --summaries e2e/reports/summary \
             --out /tmp/e2e-report.json \
-            --issue-body /tmp/e2e-issue-body.md
+            --issue-body /tmp/e2e-issue-body.md \
+            $proof_arg
 
       # Push failures/regressions that warrant attention to Feishu/Lark. PR-tier
       # failures stay in GitHub by policy; the notifier no-ops when no Feishu
--- a/.github/workflows/local-e2e.yml
+++ b/.github/workflows/local-e2e.yml
@@ -655,6 +655,8 @@ jobs:
           # fuzzy, so narrow with it but confirm an exact "[sig: <signature>]"
           # marker client-side with jq; a new signature opens its own tracking
           # issue. With no signature we keep the legacy single-issue behaviour.
+          # The search+create is not atomic, but the workflow concurrency group
+          # serializes main runs, so a duplicate-issue race is not expected here.
           if [ -n "$SIGNATURE" ]; then
             existing="$(gh issue list --repo "$REPO" --label ci-e2e-failure --state open \
               --search "in:title $SIGNATURE" --json number,title \
--- a/.github/workflows/local-e2e.yml
+++ b/.github/workflows/local-e2e.yml
@@ -541,8 +541,17 @@ jobs:
       # Push failures/regressions that warrant attention to Feishu/Lark. PR-tier
       # failures stay in GitHub by policy; the notifier no-ops when no Feishu
       # secret is configured, so this never blocks a fork/PR without secrets.
+      #
+      # SECURITY: only run on trusted events. A pull_request runs code from the PR
+      # checkout, and a PR-controlled aggregator can set notify=true (e.g. a perf
+      # regression notifies regardless of tier); exposing the Feishu secrets to
+      # `go run` on a PR would let a same-repo PR exfiltrate them. push/schedule
+      # are main-only; workflow_dispatch/workflow_call require a trusted actor.
       - name: Notify Feishu on post-merge/nightly failure or perf regression
-        if: always() && steps.aggregate.outputs.notify == 'true'
+        if: >-
+          always() && steps.aggregate.outputs.notify == 'true' &&
+          (github.event_name == 'push' || github.event_name == 'schedule' ||
+          github.event_name == 'workflow_dispatch' || github.event_name == 'workflow_call')
         continue-on-error: true
         env:
           FEISHU_APP_ID: ${{ secrets.FEISHU_APP_ID }}
@@ -657,7 +666,9 @@ jobs:
           # issue. With no signature we keep the legacy single-issue behaviour.
           # The search+create is not atomic, but the workflow concurrency group
           # serializes main runs, so a duplicate-issue race is not expected here.
-          if [ -n "$SIGNATURE" ]; then
+          # The signature is a sha256 hex slice; require [0-9a-f] before splicing
+          # it into the jq filter (defense-in-depth against jq metacharacters).
+          if [[ "$SIGNATURE" =~ ^[0-9a-f]+$ ]]; then
             existing="$(gh issue list --repo "$REPO" --label ci-e2e-failure --state open \
               --search "in:title $SIGNATURE" --json number,title \
               --jq "[.[] | select(.title | contains(\"[sig: $SIGNATURE]\"))][0].number // empty")"
--- a/.github/workflows/local-e2e.yml
+++ b/.github/workflows/local-e2e.yml
@@ -494,13 +494,15 @@ jobs:
       # Turn the raw per-suite outcomes into one product-quality report: a richer
       # step summary grouped by product promise, an aggregated JSON for the
       # notifier, a signature-led issue body, and outputs (notify/signature/...).
+      # failure() is only valid in `if:`, so capture "a prior step hard-failed"
+      # (e.g. TiDB/server setup) as a marker file the aggregator can read.
+      - name: Detect degraded pipeline
+        if: failure()
+        run: touch /tmp/e2e-degraded
+
       - name: Aggregate e2e product-quality report
         id: aggregate
         if: always()
-        env:
-          # True when a prior step hard-failed (e.g. TiDB/server setup) before the
-          # suites produced outcomes; the aggregator then fails closed.
-          AGG_DEGRADED: ${{ failure() }}
         run: |
           cat > /tmp/e2e-outcomes.json <<JSON
           {
@@ -526,7 +528,7 @@ jobs:
             proof_arg="--proof-fail-suite e2e-feishu-proof"
           fi
           degraded_arg=""
-          if [ "${AGG_DEGRADED}" = "true" ]; then
+          if [ -f /tmp/e2e-degraded ]; then
             degraded_arg="--degraded"
           fi
           go run ./cmd/e2e-aggregate \
--- a/.github/workflows/local-e2e.yml
+++ b/.github/workflows/local-e2e.yml
@@ -523,13 +523,12 @@ jobs:
             "pjdfstest-suite": "${{ steps.pjdfstest-suite.outcome }}"
           }
           JSON
-          proof_arg=""
+          args=()
           if [ "${FEISHU_PROOF}" = "1" ]; then
-            proof_arg="--proof-fail-suite e2e-feishu-proof"
+            args+=(--proof-fail-suite e2e-feishu-proof)
           fi
-          degraded_arg=""
           if [ -f /tmp/e2e-degraded ]; then
-            degraded_arg="--degraded"
+            args+=(--degraded)
           fi
           go run ./cmd/e2e-aggregate \
             --manifest e2e/suite-manifest.json \
@@ -537,7 +536,7 @@ jobs:
             --summaries e2e/reports/summary \
             --out /tmp/e2e-report.json \
             --issue-body /tmp/e2e-issue-body.md \
-            $proof_arg $degraded_arg
+            "${args[@]}"
 
       # Push failures/regressions that warrant attention to Feishu/Lark. PR-tier
       # failures stay in GitHub by policy; the notifier no-ops when no Feishu
@@ -641,21 +640,25 @@ jobs:
             --description "Automated: local-e2e scheduled/post-merge run failed" \
             --color D93F0B --force || true
 
-          # Prefer the structured, signature-led body from the aggregator; fall
-          # back to a one-liner if it is missing for any reason.
-          if [ -s /tmp/e2e-issue-body.md ]; then
-            body="$(cat /tmp/e2e-issue-body.md)"
-            body="$body"$'\n\n'"_run: $RUN_URL_"
+          # Prefer the structured, signature-led body from the aggregator; fall back
+          # to a one-liner if it is missing. Use a file (--body-file) so suite names
+          # or failure detail containing shell metacharacters/backticks cannot break
+          # or inject into the gh call.
+          body_file=/tmp/e2e-issue-body.md
+          if [ -s "$body_file" ]; then
+            printf '\n\n_run: %s_\n' "$RUN_URL" >> "$body_file"
           else
-            body="local-e2e \`$EVENT_NAME\` run failed on \`main\` at $SHA: $RUN_URL"
+            printf 'local-e2e `%s` run failed on `main` at %s: %s\n' "$EVENT_NAME" "$SHA" "$RUN_URL" > "$body_file"
           fi
 
-          # Group recurring failures by signature: an open issue carrying the same
-          # signature is reused; a new signature opens its own tracking issue. With
-          # no signature we keep the legacy single-issue behaviour.
+          # Group recurring failures by signature. GitHub's `in:title` search is
+          # fuzzy, so narrow with it but confirm an exact "[sig: <signature>]"
+          # marker client-side with jq; a new signature opens its own tracking
+          # issue. With no signature we keep the legacy single-issue behaviour.
           if [ -n "$SIGNATURE" ]; then
             existing="$(gh issue list --repo "$REPO" --label ci-e2e-failure --state open \
-              --search "in:title $SIGNATURE" --json number --jq '.[0].number // empty')"
+              --search "in:title $SIGNATURE" --json number,title \
+              --jq "[.[] | select(.title | contains(\"[sig: $SIGNATURE]\"))][0].number // empty")"
             title="local-e2e failure on main [sig: $SIGNATURE]"
           else
             existing="$(gh issue list --repo "$REPO" --label ci-e2e-failure --state open \
@@ -664,8 +667,8 @@ jobs:
           fi
 
           if [ -n "$existing" ]; then
-            gh issue comment "$existing" --repo "$REPO" --body "$body"
+            gh issue comment "$existing" --repo "$REPO" --body-file "$body_file"
           else
             gh issue create --repo "$REPO" --label ci-e2e-failure \
-              --title "$title" --body "$body"
+              --title "$title" --body-file "$body_file"
           fi
--- a/cmd/e2e-aggregate/main.go
+++ b/cmd/e2e-aggregate/main.go
@@ -32,16 +32,17 @@ func main() {
 		outPath       = flag.String("out", "", "write the aggregated RunReport JSON here")
 		issueBodyPath = flag.String("issue-body", "", "write the GitHub issue body here")
 		tierOverride  = flag.String("tier", "", "override automation tier (pr|post-merge|nightly|manual)")
+		proofSuite    = flag.String("proof-fail-suite", "", "inject a synthetic failed suite to prove the notification path end to end (manual CI proof)")
 	)
 	flag.Parse()
 
-	if err := run(*manifestPath, *outcomesPath, *summariesDir, *outPath, *issueBodyPath, *tierOverride); err != nil {
+	if err := run(*manifestPath, *outcomesPath, *summariesDir, *outPath, *issueBodyPath, *tierOverride, *proofSuite); err != nil {
 		fmt.Fprintln(os.Stderr, "e2e-aggregate:", err)
 		os.Exit(1)
 	}
 }
 
-func run(manifestPath, outcomesPath, summariesDir, outPath, issueBodyPath, tierOverride string) error {
+func run(manifestPath, outcomesPath, summariesDir, outPath, issueBodyPath, tierOverride, proofSuite string) error {
 	tier := e2ereport.TierFromEvent(os.Getenv("GITHUB_EVENT_NAME"))
 	if tierOverride != "" {
 		tier = e2ereport.Tier(tierOverride)
@@ -51,6 +52,13 @@ func run(manifestPath, outcomesPath, summariesDir, outPath, issueBodyPath, tierO
 	outcomes, err := loadOutcomes(outcomesPath)
 	if err != nil {
 		return err
+	}
+	if proofSuite != "" {
+		// Manual proof: a synthetic failed suite makes notify=true so the whole
+		// path (aggregate -> card -> Feishu) is exercised on a green CI run. It
+		// only affects this report; the real job gate reads actual step outcomes.
+		outcomes[proofSuite] = string(e2ereport.StatusFailure)
+		fmt.Fprintf(os.Stderr, "e2e-aggregate: proof mode — injected synthetic failed suite %q\n", proofSuite)
 	}
 	adopted := loadAdoptedSummaries(summariesDir)
 
--- a/cmd/e2e-aggregate/main.go
+++ b/cmd/e2e-aggregate/main.go
@@ -33,19 +33,24 @@ func main() {
 		issueBodyPath = flag.String("issue-body", "", "write the GitHub issue body here")
 		tierOverride  = flag.String("tier", "", "override automation tier (pr|post-merge|nightly|manual)")
 		proofSuite    = flag.String("proof-fail-suite", "", "inject a synthetic failed suite to prove the notification path end to end (manual CI proof)")
+		degraded      = flag.Bool("degraded", false, "the run is degraded (e.g. a required setup step hard-failed before suites ran): if no suite failure is otherwise detected, inject a synthetic pipeline failure so it is still reported and notified. Wire from GitHub failure().")
 	)
 	flag.Parse()
 
-	if err := run(*manifestPath, *outcomesPath, *summariesDir, *outPath, *issueBodyPath, *tierOverride, *proofSuite); err != nil {
+	if err := run(*manifestPath, *outcomesPath, *summariesDir, *outPath, *issueBodyPath, *tierOverride, *proofSuite, *degraded); err != nil {
 		fmt.Fprintln(os.Stderr, "e2e-aggregate:", err)
 		os.Exit(1)
 	}
 }
 
-func run(manifestPath, outcomesPath, summariesDir, outPath, issueBodyPath, tierOverride, proofSuite string) error {
+func run(manifestPath, outcomesPath, summariesDir, outPath, issueBodyPath, tierOverride, proofSuite string, degraded bool) error {
 	tier := e2ereport.TierFromEvent(os.Getenv("GITHUB_EVENT_NAME"))
 	if tierOverride != "" {
-		tier = e2ereport.Tier(tierOverride)
+		parsed, err := parseTierOverride(tierOverride)
+		if err != nil {
+			return err
+		}
+		tier = parsed
 	}
 
 	manifest := loadManifest(manifestPath)
@@ -64,6 +69,24 @@ func run(manifestPath, outcomesPath, summariesDir, outPath, issueBodyPath, tierO
 
 	summaries := e2ereport.SynthesizeSummaries(manifest, tier, outcomes, adopted)
 	report := e2ereport.Aggregate(runContextFromEnv(tier), summaries)
+
+	if degraded && report.OverallSuccess {
+		// A required step hard-failed before suite outcomes were collected, so the
+		// per-suite view looks clean while the job will fail. Fail closed: inject a
+		// pipeline failure so the run still gets a signature, issue, and notification.
+		summaries = append(summaries, e2ereport.SuiteSummary{
+			Suite:          "e2e-pipeline",
+			Status:         e2ereport.StatusFailure,
+			Tier:           tier,
+			ProductArea:    "ci",
+			ProductPromise: "the e2e pipeline runs to completion",
+			FailureClass:   e2ereport.FailureInfrastructure,
+			OwnerHint:      "ci",
+			Detail:         "a required step failed before per-suite results were collected",
+		})
+		report = e2ereport.Aggregate(runContextFromEnv(tier), summaries)
+		fmt.Fprintln(os.Stderr, "e2e-aggregate: degraded run — injected synthetic pipeline failure")
+	}
 
 	if err := appendStepSummary(report.Markdown()); err != nil {
 		return err
@@ -95,6 +118,15 @@ func run(manifestPath, outcomesPath, summariesDir, outPath, issueBodyPath, tierO
 	fmt.Printf("e2e-aggregate: %d suites, %d failed, %d perf-regressed, notify=%v (%s)\n",
 		len(report.Suites), len(report.Failed), len(report.PerfRegressed), decision.Notify, decision.Reason)
 	return nil
+}
+
+func parseTierOverride(value string) (e2ereport.Tier, error) {
+	switch tier := e2ereport.Tier(strings.TrimSpace(value)); tier {
+	case e2ereport.TierPR, e2ereport.TierPostMerge, e2ereport.TierNightly, e2ereport.TierManual:
+		return tier, nil
+	default:
+		return "", fmt.Errorf("invalid --tier %q (want pr|post-merge|nightly|manual)", value)
+	}
 }
 
 func loadManifest(path string) e2ereport.SuiteManifest {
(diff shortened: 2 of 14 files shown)


SabaPing:
Worked through the Codex + CodeRabbit findings critically. Summary of how each was handled (pushed in the latest commits):

**Fixed — security**
- *Webhook URL leak in errors* (Codex P2 ×2, CodeRabbit Major): `postJSON` now carries an endpoint label, never the URL. The custom-bot webhook URL embeds the bot token, so a transient failure could have leaked it into CI logs. Added a redaction test asserting the URL never appears in the error.

**Fixed — notifier fails closed**
- *Webhook `StatusCode` response shape* (Codex P2): custom-bot webhooks report logical failures in `{StatusCode,StatusMessage}` on HTTP 200; we now check that in addition to the app-API `{code,msg}`.
- *Non-JSON 2xx* (CodeRabbit Major): a 200 that isn't valid JSON (proxy/HTML/wrong host) is now an error instead of looking delivered.

**Fixed — aggregation fails closed**
- *Step failure authoritative over adopted summary* (Codex P2): a failing GitHub step outcome now overrides an adopted `success` summary, so a stale/buggy summary can't suppress a real failure.
- *Merge adopted summaries with manifest defaults* (CodeRabbit Major): zero-valued adopted fields (owner/promise/tier/class) are filled from the manifest so signatures/rendering stay complete.
- *Required-but-skipped / hard pipeline failures* (Codex P2): new `--degraded` flag (wired from a `if: failure()` marker step) injects a synthetic pipeline failure when a required step hard-fails before suite outcomes exist, so the run is still signed, issued, and notified.
- *Validate suite-summary enums* (CodeRabbit Major): `ParseSuiteSummary` now rejects missing/invalid `status` (and `failure_class`/`tier`).
- *Reject unknown `--tier`* (CodeRabbit Minor): validated against the tier constants instead of a blind cast.

**Fixed — rendering consistency**
- *Perf-only path* (CodeRabbit Major): `Markdown()`/`IssueBody()` now mirror `FeishuCard()` — show "performance regression", drop the empty "Failed suites" section, and emit a signature (`FailureSignature()` now also covers perf-only regressions).

**Fixed — docs**
- *README path* (CodeRabbit Minor): use the full `e2e/suite-manifest.json`.

**Declined, with reason**
- *`DisallowUnknownFields()` on suite summaries* (CodeRabbit "consider"): kept off deliberately for forward-compatibility — a producer on a newer schema adding a field shouldn't fail closed against an older consumer. Required enums are now validated explicitly, which covers the fail-closed intent.

**Deferred to a follow-up (out of slice-1 scope)**
- *Consume the report-only nightly FUSE perf-compare artifact* (Codex P2): `compare-fuse-performance-metrics.sh` writes `performance-compare-*.json` and exits 0 even when regressed, so those soft regressions don't reach the aggregator. Wiring this needs a translator for that script's JSON shape into a `SuiteSummary`. The hard-gated `fuse-write-perf-budget` suite already covers the *gating* perf path and does notify on failure; consuming the soft-compare signal is tracked as a follow-up.

New tests cover: enum validation, failure-overrides-adopted, manifest-default merge, perf-only signature/render, webhook `StatusCode`, non-JSON 2xx, and URL redaction. `gofmt`/`go vet`/`golangci-lint v2.5.0` clean; `actionlint` clean.

.github/workflows/local-e2e.yml
```diff
@@ -560,16 +635,37 @@ jobs:
...
+          fi
+
+          # Group recurring failures by signature: an open issue carrying the same
+          # signature is reused; a new signature opens its own tracking issue. With
+          # no signature we keep the legacy single-issue behaviour.
+          if [ -n "$SIGNATURE" ]; then
+            existing="$(gh issue list --repo "$REPO" --label ci-e2e-failure --state open \
+              --search "in:title $SIGNATURE" --json number --jq '.[0].number // empty')"
+            title="local-e2e failure on main [sig: $SIGNATURE]"
+          else
```
qiffang, line 676: [claude] **Security: shell injection via `$body` and `$SIGNATURE` in `gh issue create/comment`.** The `$body` variable contains the full contents of `/tmp/e2e-issue-body.md` (which includes suite names, failure details, etc.) and is interpolated directly into a shell command. If any suite name or failure message contains shell metacharacters or backticks, this could break or be exploited.

Use `--body-file` instead:
```bash
if [ -s /tmp/e2e-issue-body.md ]; then
  echo -e "\n\n_run: $RUN_URL_" >> /tmp/e2e-issue-body.md
  body_arg="--body-file /tmp/e2e-issue-body.md"
else
  echo "local-e2e \`$EVENT_NAME\` run failed on \`main\` at $SHA: $RUN_URL" > /tmp/e2e-issue-body.md
  body_arg="--body-file /tmp/e2e-issue-body.md"
fi
```
Similarly for `gh issue comment`, use `--body-file`.

.github/workflows/local-e2e.yml
```diff
@@ -560,16 +635,37 @@ jobs:
...
-            --json number --jq '.[0].number // empty')"
+
+          # Prefer the structured, signature-led body from the aggregator; fall
+          # back to a one-liner if it is missing for any reason.
+          if [ -s /tmp/e2e-issue-body.md ]; then
+            body="$(cat /tmp/e2e-issue-body.md)"
+            body="$body"$'\n\n'"_run: $RUN_URL_"
+          else
+            body="local-e2e \`$EVENT_NAME\` run failed on \`main\` at $SHA: $RUN_URL"
+          fi
```
qiffang, line 661: [claude] **Bug: `--search` with `in:title` is unreliable for signature matching.** GitHub's issue search is fuzzy and `in:title` is not an exact match — a signature like `abc` would match issues containing `abcdef`. If two signatures share a prefix, the wrong issue could be reused.

Consider using `--jq` to filter results client-side for an exact title match, or include the signature as a GitHub label (e.g., `sig:<hash>`) for precise lookup.

.github/workflows/local-e2e.yml
```diff
@@ -478,6 +491,68 @@ jobs:
...
+            "existing-key-smoke": "${{ steps.existing-key-smoke.outcome }}",
+            "cli-smoke": "${{ steps.cli-smoke.outcome }}",
+            "layer-fs-smoke": "${{ steps.layer-fs-smoke.outcome }}",
+            "fuse-smoke": "${{ steps.fuse-smoke.outcome }}",
+            "git-ops-smoke": "${{ steps.git-ops-smoke.outcome }}",
+            "portable-pack-unpack-e2e": "${{ steps.portable-pack-unpack-e2e.outcome }}",
+            "fuse-crash-recovery": "${{ steps.fuse-crash-recovery.outcome }}",
+            "fuse-write-perf-budget": "${{ steps.fuse-write-perf-budget.outcome }}",
+            "fuse_performance_compare": "${{ steps.fuse_performance_compare.outcome }}",
+            "fuse-concurrency-stress": "${{ steps.fuse-concurrency-stress.outcome }}",
```
qiffang, line 519: [claude] **Robustness: unquoted `$proof_arg` and `$degraded_arg` will cause word-splitting issues if empty.** While currently safe because empty strings produce no arguments with unquoted expansion, this is fragile. Consider using arrays instead:
```bash
args=()
if [ "$FEISHU_PROOF" = "1" ]; then
  args+=(--proof-fail-suite e2e-feishu-proof)
fi
if [ -f /tmp/e2e-degraded ]; then
  args+=(--degraded)
fi
go run ./cmd/e2e-aggregate ... "${args[@]}"
```

.github/workflows/local-e2e.yml
```diff
@@ -478,6 +491,68 @@ jobs:
...
 
+      # Turn the raw per-suite outcomes into one product-quality report: a richer
+      # step summary grouped by product promise, an aggregated JSON for the
+      # notifier, a signature-led issue body, and outputs (notify/signature/...).
+      # failure() is only valid in `if:`, so capture "a prior step hard-failed"
+      # (e.g. TiDB/server setup) as a marker file the aggregator can read.
+      - name: Detect degraded pipeline
+        if: failure()
+        run: touch /tmp/e2e-degraded
+
```
qiffang, line 502: [claude] **Correctness: the `aggregate` step needs explicit `id: aggregate` to be referenced by `steps.aggregate.outputs.*`.** I see `id: aggregate` is set on line 503, which is correct. However, verify that the step's outputs (`notify`, `signature`, etc.) are actually populated — the `writeOutputs` function in `cmd/e2e-aggregate/main.go` writes to `$GITHUB_OUTPUT`, but the workflow doesn't declare these as step outputs. In modern GitHub Actions this works via the file-based output mechanism, so this should be fine, but worth a smoke test.

cmd/e2e-aggregate/main.go
```diff
@@ -0,0 +1,246 @@
...
+		// pipeline failure so the run still gets a signature, issue, and notification.
+		summaries = append(summaries, e2ereport.SuiteSummary{
+			Suite:          "e2e-pipeline",
+			Status:         e2ereport.StatusFailure,
+			Tier:           tier,
+			ProductArea:    "ci",
+			ProductPromise: "the e2e pipeline runs to completion",
+			FailureClass:   e2ereport.FailureInfrastructure,
+			OwnerHint:      "ci",
+			Detail:         "a required step failed before per-suite results were collected",
```
qiffang, line 88: [claude] **Edge case: `degraded` re-aggregation appends to `summaries` but doesn't deduplicate.** If a future caller sets both `--proof-fail-suite` and `--degraded`, and the proof suite also triggers `OverallSuccess == false`, the degraded branch won't fire — which is correct. But if `--proof-fail-suite` injects a suite that still leaves `OverallSuccess == true` (e.g., the proof suite isn't in the failure list due to a bug), you could get both synthetic suites. Minor, but worth a comment.

cmd/e2e-aggregate/main.go
```diff
@@ -0,0 +1,246 @@
...
+	return out
+}
+
+func runContextFromEnv(tier e2ereport.Tier) e2ereport.RunContext {
+	get := os.Getenv
+	ctx := e2ereport.RunContext{
+		Trigger:  tier,
+		Repo:     get("GITHUB_REPOSITORY"),
+		SHA:      get("GITHUB_SHA"),
+		Workflow: get("GITHUB_WORKFLOW"),
```
qiffang, line 193: [claude] **Correctness: `appendStepSummary` opens with `O_APPEND|O_WRONLY|O_CREATE` but never checks if the file was newly created.** The file permissions `0o644` are fine, but if `GITHUB_STEP_SUMMARY` points to a symlink or unexpected path, this could write to an unintended location. This is a very minor concern since the environment is trusted (GitHub Actions runner), but worth noting.

cmd/feishu-notify/main.go
```diff
@@ -0,0 +1,67 @@
...
+	decision := report.ShouldNotifyFeishu()
+	if !decision.Notify && !force {
+		fmt.Printf("feishu-notify: no notification needed (%s)\n", decision.Reason)
+		return nil
+	}
+
+	cfg := feishu.ConfigFromEnv(os.Getenv)
+	if cfg.Transport() == feishu.TransportNone {
+		fmt.Println("feishu-notify: no Feishu transport configured (set FEISHU_WEBHOOK or FEISHU_APP_ID/SECRET/CHAT_ID); skipping send")
+		return nil
```
qiffang, line 56: [claude] **Design: `cfg.Transport()` is checked after reading the report, but the report read could fail first.** Consider checking transport availability earlier (before reading the report) to fail fast with the 'no transport configured' message. This avoids confusing errors when both the report is missing AND no transport is configured — the user sees a file error instead of the more helpful 'no transport' message.

cmd/feishu-notify/main.go
```diff
@@ -0,0 +1,67 @@
...
+	}
+
+	sent, err := feishu.Send(context.Background(), cfg, report.FeishuCard())
```
qiffang, line 59: [claude] **Question: what happens when `Send` returns `(false, nil)`?** The `sent` boolean is checked but no message is logged for `sent == false` with `err == nil`. Is this a valid state? If so, a log line would help debugging: `feishu-notify: send returned false (no delivery)`.

.github/workflows/local-e2e.yml
```diff
@@ -478,6 +491,68 @@ jobs:
...
+          if [ "${FEISHU_PROOF}" = "1" ]; then
+            proof_arg="--proof-fail-suite e2e-feishu-proof"
+          fi
+          degraded_arg=""
+          if [ -f /tmp/e2e-degraded ]; then
+            degraded_arg="--degraded"
+          fi
+          go run ./cmd/e2e-aggregate \
+            --manifest e2e/suite-manifest.json \
+            --outcomes /tmp/e2e-outcomes.json \
```
qiffang, line 535: [claude] **Security: Feishu secrets are passed as env vars to `go run`, which is fine, but ensure `FEISHU_APP_SECRET` is never logged.** The `cmd/feishu-notify` code should not log the config or environment. Verify that `feishu.ConfigFromEnv` and error paths don't include the secret in error messages.

.github/workflows/local-e2e.yml
```diff
@@ -478,6 +491,68 @@ jobs:
             echo "pjdfstest POSIX suite: \`RUN_POSIX_FEATURE_MATRIX=${RUN_POSIX_FEATURE_MATRIX}\`"
           } >> "$GITHUB_STEP_SUMMARY"
 
+      # Turn the raw per-suite outcomes into one product-quality report: a richer
+      # step summary grouped by product promise, an aggregated JSON for the
+      # notifier, a signature-led issue body, and outputs (notify/signature/...).
+      # failure() is only valid in `if:`, so capture "a prior step hard-failed"
```
qiffang, line 497: [claude] **Correctness: outcomes JSON uses step ids like `fuse_performance_compare` with an underscore, while others use hyphens.** Ensure every step id referenced here actually exists as a step in the workflow — a typo would silently produce an empty string outcome. Consider adding a sanity check in the aggregator for unknown/empty outcome values.

qiffang (changes requested):
## Automated Code Review

**Claude Code**: Well-structured PR with good separation of concerns and defensive design (continue-on-error, no-op on missing secrets, graceful degradation). However, there are shell injection risks in the GitHub issue filing step where `$body` and `$SIGNATURE` are interpolated unsafely, a missing step output declaration for the aggregate step, and a few robustness concerns around the outcomes JSON construction in the workflow.

**Codex** ✗ (failed): codex timed out after 600s

SabaPing:
Second review round addressed (commit 2db0302). Critical disposition per comment:

**Fixed**
- **(CodeRabbit, Major) `--degraded` masks perf-only regressions** — valid. The synthetic `e2e-pipeline` failure now only injects when the report has *no* failures **and** *no* perf regressions, so a degraded run that also has a perf-only regression keeps its own signature/routing. Verified both paths locally (degraded+perf → no injection; degraded+empty → injects).
- **(qiffang, Security) shell injection via `$body`/`$SIGNATURE` in `gh issue …`** — valid. Switched to `--body-file`; the body is built in a file, so suite names / failure detail with metacharacters can't break or inject. (`$body` was already quoted, but `--body-file` is the robust fix.)
- **(qiffang, Bug) fuzzy `in:title` signature match** — valid. Now confirm the exact `[sig: <signature>]` marker client-side with `jq` rather than trusting GitHub's fuzzy search.
- **(qiffang, Robustness) unquoted `$proof_arg`/`$degraded_arg`** — applied the suggested bash-array form (`args=(); … ; "${args[@]}"`).

**Already fixed in the prior round (these comments predate commit b505bad→…)**
- **(CodeRabbit, Major) redact URL from Feishu errors** and **fail on non-JSON 2xx** — both already implemented: `postJSON` carries an endpoint label (never the URL) and a 2xx body that isn't valid JSON is an error. (Also covers qiffang's "secrets never logged": `ConfigFromEnv` never logs, and errors use endpoint labels — the app secret only ever appears in the request body.)

**Skipped, with reasoning**
- **(qiffang) check transport before reading the report** — intentional: the notify decision *requires* the report, and a missing report is a genuine error worth surfacing rather than hiding behind a "no transport" message. In CI this step is gated on `notify == 'true'`, so the report exists.
- **(qiffang) log when `Send` returns `(false, nil)`** — unreachable: the caller already checks `cfg.Transport() == None` and logs/return before calling `Send`, so `Send` always receives a configured transport.
- **(qiffang) `appendStepSummary` symlink concern** — `GITHUB_STEP_SUMMARY` is a runner-controlled trusted path; the reviewer flagged it as very minor. Leaving as-is.
- **(qiffang) step-id underscore vs hyphen / outputs populated** — verified: `fuse_performance_compare` matches the actual step id, and the file-based `$GITHUB_OUTPUT` mechanism populates `steps.aggregate.outputs.*` (exercised by the proof runs).

build/vet/lint/test + actionlint all green.

qiffang (commented):
## Review: PR #642 — e2e product-quality feedback loop + Feishu notifications

Read the full diff carefully (2023 lines, 14 files). Overall this is a well-structured, well-tested addition. The `internal/e2ereport` package is dependency-free and fully unit-testable — exactly right for this kind of infrastructure. A few findings below.

---

### Findings

**1. `cmd/e2e-aggregate/main.go` — `cancelled` maps to `StatusFailure` silently (nit)**

`NormalizeStatus` maps `cancelled` to `StatusFailure` (the `default` case). This is correct for safety (fail-closed), but a cancelled suite is operationally different from a failed one — it may produce noise in Feishu notifications. Consider whether `StatusSkipped` is more appropriate for `cancelled`, or at minimum document the choice.

**2. `cmd/e2e-aggregate/main.go` — issue-creation path has no dedup guard**

Lines 134-169: `createOrCommentIssue` searches for an open issue with the failure signature in the title and either comments or creates. The search + create is not atomic — two concurrent runs with the same signature could both find zero results and create duplicate issues. In practice this is unlikely for CI (runs are serialized per workflow), but worth a comment acknowledging the race window.

**3. `internal/feishu/feishu.go` — tenant token is fetched per-send, no caching**

`sendApp` calls `tenantToken()` on every send. The Feishu tenant token has a 2-hour TTL. For the current use case (one send per CI run), this is fine. But if this package is ever used in a loop (e.g., sending to multiple chats), it will hit the token endpoint unnecessarily. Consider a TODO for token caching if multi-send becomes a use case.

**4. `internal/e2ereport/render.go` — `arrowPct` assumes higher-is-worse**

`arrowPct` colors increases red (worse) and decreases green (better). This is correct for latency metrics (p99_write_ms), but incorrect for throughput metrics (ops/sec) where higher is better. The current suite manifest only has latency budgets, so this works today. If throughput metrics are added later, this will produce inverted signals. Worth a comment or a `higherIsBetter` field on `Metric`.

**5. `internal/e2ereport/report.go` — `FailureSignature` uses SHA-256 truncated to 6 bytes (48 bits)**

The 12-hex-char signature is used to group issues. At 48 bits, collision probability is ~1 in 2^24 per pair — very unlikely for the expected number of distinct failure patterns. This is fine. Just noting for awareness.

**6. Credential hygiene in `postJSON` — good pattern**

The `postJSON` function deliberately uses endpoint labels instead of URLs in error messages, preventing webhook token leakage. The test `TestPostJSONSurfacesFeishuLogicalError` explicitly asserts the URL doesn't leak. Excellent practice.

**7. Test coverage is thorough**

- `report_test.go`: covers parsing validation (fail-closed on bad enums), aggregation, signature stability/order-independence, notification policy, perf regression detection, render output assertions
- `manifest_test.go`: covers synthesis, failure-outcome-overrides-adopted-success
- `feishu_test.go`: covers both transports with httptest, logical error surfacing, non-JSON fail-closed, URL leak prevention
- `event_test.go`: covers tier mapping

All edge cases I'd expect to see are covered.

---

### Summary

No blocking issues. The architecture is clean — `e2ereport` is pure logic, `feishu` handles transport, `cmd/e2e-aggregate` orchestrates. The fail-closed design (unknown status = failure, non-JSON 2xx = error, missing enum = parse error) is the right default for CI infrastructure. Test coverage is comprehensive.

Findings #1-#4 are nits/future considerations, not blockers for merge.

.github/workflows/local-e2e.yml
```diff
@@ -560,16 +635,37 @@ jobs:
...
+            body="local-e2e \`$EVENT_NAME\` run failed on \`main\` at $SHA: $RUN_URL"
+          fi
+
+          # Group recurring failures by signature: an open issue carrying the same
+          # signature is reused; a new signature opens its own tracking issue. With
+          # no signature we keep the legacy single-issue behaviour.
+          if [ -n "$SIGNATURE" ]; then
+            existing="$(gh issue list --repo "$REPO" --label ci-e2e-failure --state open \
+              --search "in:title $SIGNATURE" --json number --jq '.[0].number // empty')"
+            title="local-e2e failure on main [sig: $SIGNATURE]"
```
qiffang, line 675: [claude] **Security: shell injection via `$body`.** The `body` variable is populated from `cat /tmp/e2e-issue-body.md`, which contains aggregator output including suite names, failure details, and potentially user-influenced content (branch names, commit messages). This is then passed directly as `gh issue comment "$existing" --repo "$REPO" --body "$body"` and `gh issue create ... --body "$body"`. If the body contains characters like `$(...)` or backticks, they will be interpreted by the shell. Use `--body-file /tmp/e2e-issue-body.md` instead of interpolating the content through a shell variable, or pipe via stdin: `gh issue create ... --body-file /tmp/e2e-issue-body.md`.

.github/workflows/local-e2e.yml
```diff
@@ -560,16 +635,37 @@ jobs:
...
-          body="local-e2e \`$EVENT_NAME\` run failed on \`main\` at $SHA: $RUN_URL"
-          existing="$(gh issue list --repo "$REPO" --label ci-e2e-failure --state open \
-            --json number --jq '.[0].number // empty')"
+
+          # Prefer the structured, signature-led body from the aggregator; fall
+          # back to a one-liner if it is missing for any reason.
+          if [ -s /tmp/e2e-issue-body.md ]; then
+            body="$(cat /tmp/e2e-issue-body.md)"
+            body="$body"$'\n\n'"_run: $RUN_URL_"
+          else
```
qiffang, line 659: [claude] **Security: shell injection via `$SIGNATURE` in `--search`.** The `SIGNATURE` value is a hash computed by the aggregator, but if the aggregator has a bug or the output is tampered with, this flows unsanitized into a `gh issue list --search "in:title $SIGNATURE"` command. Since signatures should be hex-only, consider validating the format here: `if [[ ! "$SIGNATURE" =~ ^[a-f0-9]+$ ]]; then SIGNATURE=""; fi`.

.github/workflows/local-e2e.yml
```diff
@@ -478,6 +491,68 @@ jobs:
...
+            "fuse-crash-recovery": "${{ steps.fuse-crash-recovery.outcome }}",
+            "fuse-write-perf-budget": "${{ steps.fuse-write-perf-budget.outcome }}",
+            "fuse_performance_compare": "${{ steps.fuse_performance_compare.outcome }}",
+            "fuse-concurrency-stress": "${{ steps.fuse-concurrency-stress.outcome }}",
+            "fuse-posix-fsx": "${{ steps.fuse-posix-fsx.outcome }}",
+            "smoke-all": "${{ steps.smoke-all.outcome }}",
+            "git-feature-matrix": "${{ steps.git-feature-matrix.outcome }}",
+            "pjdfstest-suite": "${{ steps.pjdfstest-suite.outcome }}"
+          }
+          JSON
```
qiffang, line 525: [claude] **Bug: unquoted variable expansion.** `$proof_arg` and `$degraded_arg` are intentionally unquoted so word-splitting provides multiple arguments, but this is fragile. If any future flag value contains spaces it will break. Consider using arrays instead: `proof_args=()` / `proof_args=(--proof-fail-suite e2e-feishu-proof)` and then `"${proof_args[@]}"` in the go run invocation.

cmd/e2e-aggregate/main.go
```diff
@@ -0,0 +1,246 @@
...
+		// path (aggregate -> card -> Feishu) is exercised on a green CI run. It
+		// only affects this report; the real job gate reads actual step outcomes.
+		outcomes[proofSuite] = string(e2ereport.StatusFailure)
+		fmt.Fprintf(os.Stderr, "e2e-aggregate: proof mode — injected synthetic failed suite %q\n", proofSuite)
+	}
+	adopted := loadAdoptedSummaries(summariesDir)
+
+	summaries := e2ereport.SynthesizeSummaries(manifest, tier, outcomes, adopted)
+	report := e2ereport.Aggregate(runContextFromEnv(tier), summaries)
+
```
qiffang, line 72: [claude] **Edge case: proof suite collides with real suite.** If `proofSuite` happens to match an actual suite ID in `outcomes`, the real outcome is silently overwritten. Consider checking for collision: `if _, exists := outcomes[proofSuite]; exists { return fmt.Errorf(...) }` or using a namespaced ID like `__proof__/` prefix.

cmd/e2e-aggregate/main.go
```diff
@@ -0,0 +1,246 @@
...
+			Status:         e2ereport.StatusFailure,
+			Tier:           tier,
+			ProductArea:    "ci",
+			ProductPromise: "the e2e pipeline runs to completion",
+			FailureClass:   e2ereport.FailureInfrastructure,
+			OwnerHint:      "ci",
+			Detail:         "a required step failed before per-suite results were collected",
+		})
+		report = e2ereport.Aggregate(runContextFromEnv(tier), summaries)
+		fmt.Fprintln(os.Stderr, "e2e-aggregate: degraded run — injected synthetic pipeline failure")
```
qiffang, line 94: [claude] **Correctness: `report` is computed, then summaries are mutated and report recomputed.** The first `Aggregate` call on line 76 is discarded when `degraded && report.OverallSuccess`. This is correct but wasteful — consider checking the degraded condition on the summaries before the first aggregation, or at least adding a comment that the first call is intentionally speculative.

cmd/feishu-notify/main.go
qiffang, line 56: [claude] **Design: `cfg.Transport()` checked after `decision.Notify`.** The ordering means if `force=true` and no transport is configured, the user gets a confusing 'skipping send' message instead of an error. When `--force` is set, consider returning an error if no transport is configured, since the user explicitly wants to send.

cmd/e2e-aggregate/main.go
```diff
@@ -0,0 +1,246 @@
...
+	if path == "" {
+		fmt.Print(md)
+		return nil
+	}
+	f, err := os.OpenFile(path, os.O_APPEND|os.O_WRONLY|os.O_CREATE, 0o644)
+	if err != nil {
+		return fmt.Errorf("open step summary: %w", err)
+	}
+	defer func() { _ = f.Close() }()
+	if _, err := f.WriteString("\n" + md + "\n"); err != nil {
```
qiffang, line 213: [claude] **Correctness: `writeOutputs` uses `strings.ReplaceAll(kv[k], "\n", " ")` for multi-line safety.** This silently munges values. For the `reason` field this could merge words unexpectedly. The `GITHUB_OUTPUT` delimiter-based multiline format (`key<<EOF`) would preserve fidelity, though for these short values the current approach is acceptable — just be aware if `reason` strings grow.

.github/workflows/local-e2e.yml
```diff
@@ -478,6 +491,68 @@ jobs:
...
+            degraded_arg="--degraded"
+          fi
+          go run ./cmd/e2e-aggregate \
+            --manifest e2e/suite-manifest.json \
+            --outcomes /tmp/e2e-outcomes.json \
+            --summaries e2e/reports/summary \
+            --out /tmp/e2e-report.json \
+            --issue-body /tmp/e2e-issue-body.md \
+            $proof_arg $degraded_arg
+
```
qiffang, line 540: [claude] **Nit: missing `id: aggregate` outputs declaration.** The step uses `steps.aggregate.outputs.notify` and `steps.aggregate.outputs.signature` downstream, but relies on the Go binary writing to `$GITHUB_OUTPUT`. This works but is implicit — consider adding a comment that outputs are dynamically set by the binary, not declared in the step.

.github/workflows/local-e2e.yml
```diff
@@ -478,6 +491,68 @@ jobs:
...
+      # Push failures/regressions that warrant attention to Feishu/Lark. PR-tier
+      # failures stay in GitHub by policy; the notifier no-ops when no Feishu
+      # secret is configured, so this never blocks a fork/PR without secrets.
+      - name: Notify Feishu on post-merge/nightly failure or perf regression
```
qiffang, line 550: [codex] This step can run on pull_request because the only guard is steps.aggregate.outputs.notify. Both that output and cmd/feishu-notify come from the PR checkout, so a same-repo PR can force notify=true or emit a perf-regression report and then run modified Go code with FEISHU_APP_SECRET/FEISHU_WEBHOOK in the environment. Gate this step to trusted events such as push, schedule, and explicit manual runs, or move PR perf notifications through a trusted workflow/artifact path before exposing secrets.

qiffang (changes requested):
## Automated Code Review

**Claude Code**: Well-structured PR with good defensive design (continue-on-error, no-op on missing secrets, graceful degradation). However, there is a shell injection vulnerability in the GitHub Actions issue-filing step where the aggregator-produced issue body is interpolated unsanitized into a shell command via variable expansion, and a few smaller issues around missing input validation and unquoted shell variables.

**Codex**: The reporting and notification flow is mostly well structured, but the Feishu notification step currently exposes repository notification secrets to code from pull_request runs, which is a security blocker.

qiffang (changes requested):
Blocking issue:

- `.github/workflows/local-e2e.yml:295-331` still allows `Run FUSE release gate` to hard-fail before the next independent PR-gate suites run. `git-ops-smoke`, `portable-pack-unpack-e2e`, `fuse-crash-recovery`, and `fuse-write-perf-budget` all keep the default `if: success()` condition, so if `fuse-release-gate.sh` fails they are skipped. The new aggregation step later records those suites as `skipped` (`.github/workflows/local-e2e.yml:503-539`), and Feishu/signature/issue-body output is generated before the final `Enforce smoke results` gate turns skipped required suites into job failure. That means the product-quality report can still hide independent workload failures behind the first FUSE gate failure.

This also contradicts the updated README claim at `e2e/README.md:417` that all hard-fail workloads are attempted so one failure does not hide another workload result.

Required fix: make the early FUSE release gate non-terminal for report collection, or run the subsequent independent suites with `if: always()` and let the final enforce step be the single hard gate. Add a regression/check for the intended behavior: an early release-gate failure must not prevent later independent suite outcomes from being collected and reported.

Validation notes:
- Reviewed latest head `2db0302ac573dfff7e2da571f8f06b2f4bd3cc82`.
- `git diff --check e2680748de0ccbd5dfd6ad3dd1524a5d258d576d...HEAD` passed locally.
- GitHub currently only reports CodeRabbit for this PR; no local-e2e/ci result was available in `gh pr checks`.
- Could not run Go tests locally because this machine has no `go` toolchain.

SabaPing:
Thanks @qiffang — thorough read. That review was non-blocking ("no blocking issues"), but I addressed the actionable nits (commit c8568f1):

- **#1 `cancelled` → failure** — agreed it's operationally different. Now maps `cancelled` to **skipped** (a job/concurrency cancellation is a non-event, not a product failure → no false Feishu alarm); truly-unknown outcomes still fail closed. Test updated.
- **#2 issue search+create race** — added a comment: it's non-atomic, but the workflow `concurrency` group serializes main runs, so a duplicate-issue race isn't expected in practice.
- **#3 tenant token per send** — added a TODO; fine for one-send-per-CI-run, cache (2h TTL) if multi-send is ever needed.
- **#4 `arrowPct` higher-is-worse** — documented the lower-is-better assumption (holds for the latency budgets today) and that a `HigherIsBetter` flag on `Metric` is the fix when a throughput budget is added. Left as a comment rather than building the flag now, since no higher-is-better budget exists yet.

#5–#7 (48-bit signature collision math, the `postJSON` redaction pattern, test coverage) were observations/praise — no change.

build/vet/lint/test + actionlint green.

.github/workflows/local-e2e.yml
```diff
@@ -560,16 +634,43 @@ jobs:
...
+          # marker client-side with jq; a new signature opens its own tracking
+          # issue. With no signature we keep the legacy single-issue behaviour.
+          # The search+create is not atomic, but the workflow concurrency group
+          # serializes main runs, so a duplicate-issue race is not expected here.
+          if [ -n "$SIGNATURE" ]; then
+            existing="$(gh issue list --repo "$REPO" --label ci-e2e-failure --state open \
+              --search "in:title $SIGNATURE" --json number,title \
+              --jq "[.[] | select(.title | contains(\"[sig: $SIGNATURE]\"))][0].number // empty")"
+            title="local-e2e failure on main [sig: $SIGNATURE]"
+          else
```
qiffang, line 676: [claude] **Potential command injection via `$SIGNATURE` in jq filter.** The `$SIGNATURE` variable is interpolated directly into a jq `select()` expression via shell double-quotes:
```
--jq "[.[] | select(.title | contains(\"[sig: $SIGNATURE]\"))][0].number // empty"
```
If `SIGNATURE` contains a `")` sequence or other jq metacharacters, this could break the filter or produce unexpected results. Since SIGNATURE is generated by your own Go code (a SHA hash of sorted suite names), the practical risk is very low, but for defense-in-depth consider passing it as a `--jqarg` or pre-validating it matches `^[a-f0-9]+$` in shell before use.

.github/workflows/local-e2e.yml
```diff
@@ -560,16 +634,43 @@ jobs:
...
+            existing="$(gh issue list --repo "$REPO" --label ci-e2e-failure --state open \
+              --json number --jq '.[0].number // empty')"
+            title="local-e2e $EVENT_NAME run failing on main"
+          fi
+
```
qiffang, line 681: [claude] The `$title` variable is used unquoted-in-spirit inside `gh issue create --title "$title"`. Since `title` contains the signature (hex-safe) and a fixed prefix this is fine, but note that if the legacy no-signature branch is taken, `$EVENT_NAME` is interpolated into the title. If `EVENT_NAME` ever contained spaces or special chars (it won't for GitHub events, but defensive quoting is free), this could break. Current code is safe given known GitHub event names.

cmd/e2e-aggregate/main.go
```diff
@@ -0,0 +1,249 @@
...
+		summaries = append(summaries, e2ereport.SuiteSummary{
+			Suite:          "e2e-pipeline",
+			Status:         e2ereport.StatusFailure,
+			Tier:           tier,
+			ProductArea:    "ci",
+			ProductPromise: "the e2e pipeline runs to completion",
+			FailureClass:   e2ereport.FailureInfrastructure,
+			OwnerHint:      "ci",
+			Detail:         "a required step failed before per-suite results were collected",
+		})
```
qiffang, line 89: [claude] Minor: when `--degraded` is set and a synthetic pipeline failure is injected, `Aggregate()` is called a second time with the extended `summaries` slice. This is fine for correctness but worth a brief comment that the re-aggregation is intentional (the existing stderr log covers the *why* but a reader of the code might wonder about the double call).

cmd/e2e-aggregate/main.go
```diff
@@ -0,0 +1,249 @@
...
+			return fmt.Errorf("write report: %w", err)
+		}
+	}
+	if issueBodyPath != "" && len(report.Failed) > 0 {
+		if err := os.WriteFile(issueBodyPath, []byte(report.IssueBody()), 0o644); err != nil {
+			return fmt.Errorf("write issue body: %w", err)
+		}
+	}
+
+	decision := report.ShouldNotifyFeishu()
```
qiffang, line 112: [claude] Edge case: if `--issue-body` is provided but `report.Failed` is empty (all suites passed), no file is written. Downstream in the workflow, the `ci-e2e-failure` step checks `-s "$body_file"` and falls back to a one-liner. This works, but if `--out` wrote a previous run's file to the same path and the current run is green, the stale file would be picked up. In CI `/tmp` is fresh per job so this is safe, but the implicit coupling is worth noting.

cmd/feishu-notify/main.go
```diff
@@ -0,0 +1,67 @@
...
+
+	decision := report.ShouldNotifyFeishu()
+	if !decision.Notify && !force {
+		fmt.Printf("feishu-notify: no notification needed (%s)\n", decision.Reason)
+		return nil
+	}
+
+	cfg := feishu.ConfigFromEnv(os.Getenv)
+	if cfg.Transport() == feishu.TransportNone {
+		fmt.Println("feishu-notify: no Feishu transport configured (set FEISHU_WEBHOOK or FEISHU_APP_ID/SECRET/CHAT_ID); skipping send")
```
qiffang, line 55: [claude] Good: the no-transport-configured path exits 0 with a log message rather than failing the build. This is the right behavior for forks/PRs without secrets.

.github/workflows/local-e2e.yml
```diff
@@ -478,6 +491,67 @@ jobs:
...
+            "cli-smoke": "${{ steps.cli-smoke.outcome }}",
+            "layer-fs-smoke": "${{ steps.layer-fs-smoke.outcome }}",
+            "fuse-smoke": "${{ steps.fuse-smoke.outcome }}",
+            "git-ops-smoke": "${{ steps.git-ops-smoke.outcome }}",
+            "portable-pack-unpack-e2e": "${{ steps.portable-pack-unpack-e2e.outcome }}",
+            "fuse-crash-recovery": "${{ steps.fuse-crash-recovery.outcome }}",
+            "fuse-write-perf-budget": "${{ steps.fuse-write-perf-budget.outcome }}",
+            "fuse_performance_compare": "${{ steps.fuse_performance_compare.outcome }}",
+            "fuse-concurrency-stress": "${{ steps.fuse-concurrency-stress.outcome }}",
+            "fuse-posix-fsx": "${{ steps.fuse-posix-fsx.outcome }}",
```
qiffang, line 520: [claude] The `--summaries e2e/reports/summary` argument points to a directory that may not exist yet (no suite has adopted the JSON contract). `loadAdoptedSummaries` in the Go code handles this gracefully via `filepath.Glob` returning nil, so this is fine — just confirming the tolerance is intentional and tested.

cmd/e2e-aggregate/main.go
```diff
@@ -0,0 +1,249 @@
...
+	return m
+}
+
+func loadOutcomes(path string) (map[string]string, error) {
+	if path == "" {
+		return map[string]string{}, nil
+	}
+	data, err := os.ReadFile(path)
+	if err != nil {
+		return nil, fmt.Errorf("read outcomes: %w", err)
```
qiffang, line 158: [claude] Nit: `loadManifest` silently returns an empty manifest on *any* error (file not found, permission denied, parse error). For a parse error specifically, you might want to return a non-zero exit so a malformed manifest is caught during development rather than silently ignored in CI. Current behavior is documented as intentional for incremental adoption, so this is a judgment call.

.github/workflows/local-e2e.yml
```diff
@@ -478,6 +491,67 @@ jobs:
...
+            --outcomes /tmp/e2e-outcomes.json \
+            --summaries e2e/reports/summary \
+            --out /tmp/e2e-report.json \
+            --issue-body /tmp/e2e-issue-body.md \
+            "${args[@]}"
+
+      # Push failures/regressions that warrant attention to Feishu/Lark. PR-tier
+      # failures stay in GitHub by policy; the notifier no-ops when no Feishu
+      # secret is configured, so this never blocks a fork/PR without secrets.
+      - name: Notify Feishu on post-merge/nightly failure or perf regression
```
qiffang, line 550: [codex] This notification step can run on pull_request when the PR-controlled aggregator reports a performance regression, and it injects FEISHU_APP_SECRET/FEISHU_WEBHOOK into a `go run` built from the PR checkout. Same-repo PRs can therefore force `notify=true` or modify `cmd/feishu-notify` and exfiltrate the notification secrets. Gate this step to trusted events/refs before exposing secrets, for example push/schedule on main and explicitly trusted manual runs, or move notification to a trusted workflow_run/artifact path.

qiffang (changes requested):
## Automated Code Review

**Claude Code**: Well-structured PR that adds a product-quality feedback loop on top of existing e2e infrastructure. The design is additive and non-gating — workflow changes use continue-on-error, the Feishu notifier gracefully no-ops without secrets, and the aggregator tolerates missing inputs. Code is clean, the signature-based issue deduplication is a clear improvement over the single-issue approach, and the dual-transport Feishu integration is sensible. A few minor issues around shell quoting in the workflow and a potential jq injection vector in the issue-filing step deserve attention but are low-severity given the controlled CI context.

**Codex**: The aggregation/reporting pieces are generally well structured, but the workflow integration has a security blocker around exposing Feishu secrets to PR-controlled code, and the report can still under-collect required suite outcomes after an early hard failure.

### Additional Comments
_These comments reference lines outside the diff and are included here:_

- **.github/workflows/local-e2e.yml:295** [codex]: The FUSE release gate is still a terminal step. If it fails, the following required suites with the default `if: success()` condition are skipped, then the new aggregator records them as skipped before the final Enforce step fails the job. That produces a signature/card/issue body that only reflects the first hard failure and can hide independent workload failures. Make required suite steps non-terminal for collection, or run the later independent suites with `if: always()` and let the final Enforce step be the single hard gate.

SabaPing:
Round 4 addressed (commit 7902f67). The important one was a real security finding:

**Fixed**
- **(codex, Security) Feishu secrets exposed to PR code** — valid and important. The Notify step ran on `pull_request`, and a PR-controlled aggregator can set `notify=true` (a perf regression notifies regardless of tier), so a same-repo PR could modify `cmd/feishu-notify` and exfiltrate `FEISHU_*` via `go run` on the PR checkout. Gated the step to trusted events only: `push`/`schedule` (main-only) and `workflow_dispatch`/`workflow_call` (require a trusted actor) — never `pull_request`. Secrets are now never exposed to PR-controlled code.
- **(#665, jq injection) `$SIGNATURE` spliced into the jq filter** — added a `^[0-9a-f]+$` guard before using it (it's a sha256 hex slice; defense-in-depth against jq metacharacters).
- **(#89) degraded re-aggregation double-call** — added a comment noting it's intentional.

**Skipped, with reasoning**
- **(#155) `loadManifest` swallows parse errors** — it logs to stderr (visible in CI) and returns empty; the tolerance is intentional for incremental adoption. A malformed *committed* manifest would also surface in code review. Judgment call, leaving as-is (the reviewer flagged it as such).
- **summaries-dir tolerance / no-transport exit 0 / `/tmp` freshness / `$title` quoting** — these were confirmations or praise of existing behavior; `"$title"` is already quoted and `$EVENT_NAME` is a known-safe GitHub event name. No change.

build/vet/lint/test + actionlint green.

qiffang (approved):
## adversary-2 re-review — GREEN (head 7902f67)

Previous blocker (suite steps lacking `if: always()` causing aggregator to see "skipped" instead of independent results) is **fixed**:
- "Aggregate e2e product-quality report" now has `if: always()`
- "Notify Feishu" has `if: always() && steps.aggregate.outputs.notify == 'true' && ...`
- "Detect degraded pipeline" correctly uses `if: failure()` to touch a marker file the aggregator reads

### Checklist

- **All inputs handled**: ✅ — `ParseSuiteSummary` fails closed on missing/invalid enum fields (status, tier, failure_class). `NormalizeStatus` treats unknown outcomes as failure. Missing manifest/summaries gracefully degraded.
- **Edge cases covered**: ✅ — Degraded pipeline (setup step hard-fails before suites), perf-only regression (no failure but budget breach), cancelled steps treated as skipped, proof mode for manual Feishu testing.
- **Failure behavior defined**: ✅ — Feishu send failure uses `continue-on-error: true` so it never blocks the job. Missing secrets → silent skip. Non-JSON 2xx → fail closed. Webhook URL never leaked in error messages.
- **Security**: ✅ — Feishu step restricted to trusted events (push/schedule/workflow_dispatch/workflow_call). PR-controlled code cannot exfiltrate secrets.
- **Tests complete**: ✅ — Enum validation, signature stability/order-independence, notify policy (PR vs post-merge vs perf-only), transport selection, app API flow, logical error surfacing, secret-URL-not-leaked, non-JSON-fail-closed.
- **API explicit**: ✅ — Suite manifest contract is well-defined with JSON schema. Workflow outputs (notify, signature, overall_success, reason) are explicit.

### Non-blocking

- `arrowPct` assumes lower-is-better (documented in comment) — when a higher-is-better metric gets a budget, the arrow color will be inverted. Already documented as a future TODO.
- `loadAdoptedSummaries` silently ignores `filepath.Glob` errors — acceptable for incremental adoption.

qiffang (changes requested):
## adversary-1 re-review

**Verdict: REVISE**

The previous blocker is still present on head `7902f67a9e40b37775bb7e5929c18455bd86c49f`.

### Blocker: the release gate can still stop later independent suites before aggregation

`.github/workflows/local-e2e.yml:295-300` still runs `Run FUSE release gate` as a normal terminal step:

```yaml
- name: Run FUSE release gate
  id: fuse-smoke
  env:
    RUN_FUSE_CONCURRENCY_STRESS: "0"
    RUN_FUSE_POSIX_FSX: "0"
  run: bash e2e/fuse-release-gate.sh
```

The next required PR-gate suites (`git-ops-smoke`, `portable-pack-unpack-e2e`, `fuse-crash-recovery`, `fuse-write-perf-budget`) have `continue-on-error: true`, but they do **not** have `if: always()`. GitHub Actions step default is `if: success()`, so if `fuse-release-gate.sh` exits non-zero, those later independent suites are skipped before the report is built.

The new aggregation step at `.github/workflows/local-e2e.yml:503-539` does run with `if: always()`, but by then the later suite outcomes are already `skipped`. The degraded marker at `.github/workflows/local-e2e.yml:499-501` does not fix this case because the report already has a concrete failed suite (`fuse-smoke`), so `cmd/e2e-aggregate/main.go:73-95` will not inject the synthetic `e2e-pipeline` failure. Result: the signature/card/issue body can still report only the first FUSE gate failure and hide failures in the later independent workloads.

This still contradicts `e2e/README.md:417`, which says all hard-fail workloads are attempted so one failure does not hide another workload result.

Required fix remains the same: either make `Run FUSE release gate` non-terminal for collection (`continue-on-error: true`) and let `Enforce smoke results` be the hard gate, or add `if: always()` to the subsequent independent suites that must still run after an early FUSE failure. Also add a regression/check for this behavior: a failed release gate must not prevent later required suite outcomes from being collected and represented in the aggregate report.

Validation:
- `git diff --check e2680748de0ccbd5dfd6ad3dd1524a5d258d576d...HEAD` passed locally.
- This machine still has no `go`, so I could not run local Go tests.
- GitHub currently reports CodeRabbit only for this PR; I do not see a `ci` / `local-e2e` check result on latest head.

qiffang (commented):
## adversary-2 re-review revision — downgrade to REVISE

@adversary-1 correctly identifies that my GREEN was premature. The aggregator step has `if: always()`, but the **suite steps themselves** (e.g. `git-ops-smoke`, `portable-pack-unpack-e2e`, `fuse-crash-recovery`, `fuse-write-perf-budget`) still use GitHub's default `if: success()`. When `fuse-smoke` fails, downstream suites are skipped by the runner — not by any explicit gate in this PR.

The aggregator faithfully reports their outcome as "skipped", which is factually correct. But the product-quality report's stated goal is to surface **which product promises are broken**. A "skipped" suite doesn't tell you whether `git-ops-smoke` (git correctness) is broken — it just tells you it didn't run. For a report meant to route failures to owners, this is a gap: the git team sees "skipped" and has no signal.

**Revised verdict: REVISE** — suite steps that are independent of the FUSE release gate should have `if: always()` (or at minimum, the aggregator report should explicitly flag "skipped due to prior step failure" vs "skipped by configuration" so the report consumer can distinguish them).

qiffang (commented):
## PR #642 Re-review — "e2e product-quality feedback loop + Feishu notifications"

**Head**: `7902f67` | **Scope**: 2055 lines — workflow steps, `cmd/e2e-aggregate`, `cmd/feishu-notify`, `internal/e2ereport`, `internal/feishu`, suite manifest, docs.

### Verdict: **GREEN** (with 1 pre-existing note)

---

#### @adversary-1's blocker re-assessed

@adversary-1 flags that `fuse-smoke` (the release gate) lacks `continue-on-error: true`, so when it fails, subsequent steps (`git-ops-smoke`, `portable-pack-unpack-e2e`, `fuse-crash-recovery`, `fuse-write-perf-budget`) are skipped by GitHub Actions' default `if: success()` — even though those steps themselves have `continue-on-error: true`.

This is **correct behavior** but it is a **pre-existing workflow design choice**, not introduced by this PR. The aggregator handles this correctly:

1. `NormalizeStatus("skipped")` → `StatusSkipped` (not treated as failure)
2. The `Detect degraded pipeline` step fires on `failure()` and writes `/tmp/e2e-degraded`
3. The aggregator's `--degraded` flag injects a synthetic `e2e-pipeline` failure when no suite failure is otherwise detected
4. When `fuse-smoke` fails, it IS reported as failed in the outcomes JSON (`"fuse-smoke": "failure"`), so the report correctly shows the FUSE gate failure with its signature

The report WILL show subsequent suites as "skipped" — this is accurate (they didn't run) and the FUSE gate failure is the actionable signal. Adding `if: always()` to those steps would be a separate improvement to the workflow, not a bug in this PR's aggregator.

---

#### Code quality — all good

- **Enum validation**: `ParseSuiteSummary` fails closed on invalid status/tier/failure_class — no zero-value-as-success risk
- **Signature stability**: deterministic via sort + sha256, order-independent (tested)
- **Feishu credential hygiene**: webhook URL never logged in errors (tested in `TestPostJSONSurfacesFeishuLogicalError`)
- **Security**: Feishu step restricted to trusted events (`push`/`schedule`/`workflow_dispatch`/`workflow_call`) — PR checkout can't exfiltrate secrets
- **Non-JSON 2xx fail-closed**: proxy/HTML responses correctly treated as delivery failure
- **Degraded pipeline**: marker file + synthetic injection + re-aggregate pattern is clean
- **Proof mode**: synthetic failed suite for Feishu wiring verification, doesn't affect job gate
- **Test coverage**: 16 tests across e2ereport + feishu packages, covering happy path + edge cases + failure modes

No blockers. Previous round's issues (aggregator `if: always()`, Feishu `always() && ...`) are fixed.

qiffang (approved):

PR #643, opened 2026-06-29
Fix provision create tenant pool claim

srstack:
## Summary
- make /v1/provision claim an active free tenant from the TiDB Cloud tenant pool before falling back to normal cluster provisioning
- preserve admin pool get as read-only pool info lookup
- include tenant/provider fields in schema init logger context so schema SQL logs carry tenant identity
- add coverage for ordinary provision consuming a free pool tenant and applying requested spending limit

## Tests
- GOCACHE=/tmp/drive9-go-cache go test ./pkg/server -run 'TestProvisionClaimsFreePoolTenant|TestAdminTenantCreateClaimsFreePoolTenant|TestAdminTenantPool|TestProvision' -count=1
- GOCACHE=/tmp/drive9-go-cache go test ./pkg/server -count=1

## Summary by CodeRabbit

* **New Features**
  * Provisioning now returns a consistent “accepted” response including tenant ID, non-empty API key, and status.
  * Free-pool tenants can now be claimed via the standard `/v1/provision` flow.
* **Bug Fixes**
  * Improved quota validation timing during provisioning for the TiDBCloudNative provider.
  * Tenant pool claims now transition to `Used` with accurate usage timestamps.
* **Documentation**
  * Added an “Admin Tenant Pool API” section to the Go SDK integration guide.
* **Tests**
  * Updated and extended provisioning quota/pool-claim coverage, including fallback behavior when managed cluster listing is unavailable.

### DIFF
--- a/pkg/server/admin_tenant_pool.go
+++ b/pkg/server/admin_tenant_pool.go
@@ -698,6 +698,10 @@ func (s *Server) claimAdminTenantFromPool(ctx context.Context, cred tenant.Crede
 		logger.Info(ctx, "server_event", eventFields(ctx, "admin_tenant_pool_claim_skipped", "provider", tenant.ProviderTiDBCloudNative, "reason", "pool_manager_unavailable", "duration_ms", durationMillis(claimStarted))...)
 		return nil, nil, false, nil
 	}
+	if _, ok := s.provisioner.(tenant.ManagedClusterLister); !ok {
+		logger.Info(ctx, "server_event", eventFields(ctx, "admin_tenant_pool_claim_skipped", "provider", tenant.ProviderTiDBCloudNative, "reason", "managed_cluster_lister_unavailable", "duration_ms", durationMillis(claimStarted))...)
+		return nil, nil, false, nil
+	}
 	stageStarted := time.Now()
 	orgID, err := s.firstManagedOrganization(ctx, cred)
 	if err != nil || orgID == "" {
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -3908,6 +3908,13 @@ func (s *Server) handleProvision(w http.ResponseWriter, r *http.Request) {
 		}
 		credentialReq = decoded.Credential
 		quotaReq = decoded.Quota
+		if quotaReq != nil {
+			quotaReq.TenantID = "pending"
+			if err := s.validateQuotaSetRequest(*quotaReq); err != nil {
+				errJSON(w, http.StatusBadRequest, err.Error())
+				return
+			}
+		}
 	} else {
 		if err := rejectCredentialProvisionBody(r); err != nil {
 			logger.Error(r.Context(), "server_event", eventFields(r.Context(), "provision_credential_rejected", "provider", provider, "error", err)...)
@@ -3915,6 +3922,26 @@ func (s *Server) handleProvision(w http.ResponseWriter, r *http.Request) {
 			errJSON(w, http.StatusBadRequest, err.Error())
 			return
 		}
+	}
+	if provider == tenant.ProviderTiDBCloudNative && credentialReq != nil {
+		poolClaimStarted := time.Now()
+		logger.Info(r.Context(), "server_event", eventFields(r.Context(), "provision_tenant_pool_claim_started", "provider", provider, "quota_requested", quotaReq != nil)...)
+		if res, pool, claimed, err := s.claimAdminTenantFromPool(r.Context(), *credentialReq, quotaReq); err != nil {
+			logger.Error(r.Context(), "server_event", eventFields(r.Context(), "provision_tenant_pool_claim_failed", "provider", provider, "duration_ms", durationMillis(poolClaimStarted), "error", err)...)
+			errJSON(w, http.StatusBadGateway, fmt.Sprintf("claim tenant pool tenant failed: %v", err))
+			return
+		} else if claimed {
+			logger.Info(r.Context(), "server_event", eventFields(r.Context(), "provision_tenant_pool_claim_accepted", "tenant_id", res.TenantID, "provider", res.Provider, "pool_id", pool.PoolID, "organization_id", res.OrganizationID, "duration_ms", durationMillis(poolClaimStarted), "status", res.Status)...)
+			setRequestMetricTenant(r.Context(), res.TenantID, res.APIKeyID, res.Provider, classifyTenantRequest(r))
+			if res.Status == meta.TenantProvisioning {
+				s.startProvisionedTenantSchemaInit(r.Context(), res)
+			}
+			s.replenishTenantPoolAsync(r.Context(), pool, *credentialReq)
+			writeProvisionTenantAccepted(w, res)
+			logger.Info(r.Context(), "server_event", eventFields(r.Context(), "provision_tenant_pool_create_accepted", "tenant_id", res.TenantID, "provider", res.Provider, "pool_id", pool.PoolID, "organization_id", res.OrganizationID, "duration_ms", durationMillis(poolClaimStarted))...)
+			return
+		}
+		logger.Info(r.Context(), "server_event", eventFields(r.Context(), "provision_tenant_pool_claim_missed", "provider", provider, "duration_ms", durationMillis(poolClaimStarted))...)
 	}
 	res, err := s.provisionTenant(r.Context(), provisionTenantOptions{
 		KeyName:               "default",
@@ -3933,7 +3960,10 @@ func (s *Server) handleProvision(w http.ResponseWriter, r *http.Request) {
 	}
 	setRequestMetricTenant(r.Context(), res.TenantID, res.APIKeyID, res.Provider, classifyTenantRequest(r))
 	s.startProvisionedTenantSchemaInit(r.Context(), res)
+	writeProvisionTenantAccepted(w, res)
+}
 
+func writeProvisionTenantAccepted(w http.ResponseWriter, res *provisionTenantResult) {
 	w.Header().Set("Content-Type", "application/json")
 	w.WriteHeader(http.StatusAccepted)
 	response := map[string]string{
@@ -4649,6 +4679,10 @@ func (s *Server) handleLocalTenantProvision(w http.ResponseWriter, r *http.Reque
 
 func (s *Server) initTenantSchemaAsync(ctx context.Context, tenantID, tenantDSN, provider string, schemaInit func(context.Context, string) error) {
 	ctx = ensureTrace(ctx)
+	schemaCtx := logger.WithContext(ctx, logger.FromContext(ctx).With(
+		zap.String("tenant_id", tenantID),
+		zap.String("provider", provider),
+	))
 	logger.Info(ctx, "server_event", eventFields(ctx, "schema_init_started", "tenant_id", tenantID, "provider", provider)...)
 	deadline := time.Now().Add(schemaInitRetryWindow)
 	backoff := schemaInitInitialBackoff
@@ -4663,7 +4697,7 @@ func (s *Server) initTenantSchemaAsync(ctx context.Context, tenantID, tenantDSN,
 			return
 		default:
 		}
-		err := schemaInit(ctx, tenantDSN)
+		err := schemaInit(schemaCtx, tenantDSN)
 		if err == nil {
 			err = s.finalizeTenantSchemaInit(ctx, tenantID, tenantDSN, provider)
 		}
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -3909,7 +3909,6 @@ func (s *Server) handleProvision(w http.ResponseWriter, r *http.Request) {
 		credentialReq = decoded.Credential
 		quotaReq = decoded.Quota
 		if quotaReq != nil {
-			quotaReq.TenantID = "pending"
 			if err := s.validateQuotaSetRequest(*quotaReq); err != nil {
 				errJSON(w, http.StatusBadRequest, err.Error())
 				return
@@ -3928,7 +3927,8 @@ func (s *Server) handleProvision(w http.ResponseWriter, r *http.Request) {
 		logger.Info(r.Context(), "server_event", eventFields(r.Context(), "provision_tenant_pool_claim_started", "provider", provider, "quota_requested", quotaReq != nil)...)
 		if res, pool, claimed, err := s.claimAdminTenantFromPool(r.Context(), *credentialReq, quotaReq); err != nil {
 			logger.Error(r.Context(), "server_event", eventFields(r.Context(), "provision_tenant_pool_claim_failed", "provider", provider, "duration_ms", durationMillis(poolClaimStarted), "error", err)...)
-			errJSON(w, http.StatusBadGateway, fmt.Sprintf("claim tenant pool tenant failed: %v", err))
+			metricEvent(r.Context(), "tenant_provision", "provider", provider, "result", provisionTenantPoolClaimMetricResult(err))
+			errJSON(w, http.StatusBadGateway, "claim tenant pool tenant failed")
 			return
 		} else if claimed {
 			logger.Info(r.Context(), "server_event", eventFields(r.Context(), "provision_tenant_pool_claim_accepted", "tenant_id", res.TenantID, "provider", res.Provider, "pool_id", pool.PoolID, "organization_id", res.OrganizationID, "duration_ms", durationMillis(poolClaimStarted), "status", res.Status)...)
@@ -3937,6 +3937,10 @@ func (s *Server) handleProvision(w http.ResponseWriter, r *http.Request) {
 				s.startProvisionedTenantSchemaInit(r.Context(), res)
 			}
 			s.replenishTenantPoolAsync(r.Context(), pool, *credentialReq)
+			if quotaReq != nil && quotaReq.TiDBCloudSpendingLimit != nil {
+				metricEvent(r.Context(), "tenant_provision", "provider", provider, "quota", "create_time_spending_limit")
+			}
+			metricEvent(r.Context(), "tenant_provision", "provider", provider, "result", "accepted")
 			writeProvisionTenantAccepted(w, res)
 			logger.Info(r.Context(), "server_event", eventFields(r.Context(), "provision_tenant_pool_create_accepted", "tenant_id", res.TenantID, "provider", res.Provider, "pool_id", pool.PoolID, "organization_id", res.OrganizationID, "duration_ms", durationMillis(poolClaimStarted))...)
 			return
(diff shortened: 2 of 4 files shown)


srstack (commented):
Reviewed the diff. The change correctly extends the tenant-pool claim path (already in `handleAdminTenantCreate`) to the public `/v1/provision` endpoint, and the `writeProvisionTenantAccepted` extraction keeps the response shape consistent across claim vs. full-provision. Logic and scoping look right (only `ProviderTiDBCloudNative` + credential request, falls through to `provisionTenant` on miss). A few things to address before merge — the metrics gap is the one that matters.

### Should fix: observability gap on the claim path (agree with Copilot, and it's significant here)
The new pool-claim branch returns early on both success and failure **without emitting the `tenant_provision` metric events** that `provisionTenant` emits (`result=accepted` / `error` / `cluster_error`). So once `/v1/provision` starts serving from the pool, those requests vanish from the provision metrics/dashboards — and pool claims are expected to become the *common* path, not the rare one. That means the metric would under-count exactly when the pool is working. Please emit consistent `tenant_provision` events here:
- success → `metricEvent(ctx, "tenant_provision", "provider", provider, "result", "accepted")` (matching what `provisionTenant` records on its accepted path)
- claim error → the same `result="error"`/`"cluster_error"` classification before the 502.

This keeps the alert rules and provision dashboards accurate regardless of claim-vs-full path. (`handleAdminTenantCreate` has the same gap, but that's an internal admin path; `/v1/provision` is the public one, so it matters more.)

### Minor: don't mutate the decoded request for validation (agree with Copilot)
`quotaReq.TenantID = "pending"` is dead — `validateQuotaSetRequest` never reads `TenantID` (confirmed: it only checks the limit fields). It also leaves a misleading `"pending"` on the struct that then flows into `claimAdminTenantFromPool` / `provisionTenant` (which overwrite it with the real tenant ID, so no functional bug, but it's confusing in logs/debug). Either drop the line entirely, or validate a shallow copy.

### Confirm: validation-trigger consistency with the admin path
`handleAdminTenantCreate` validates only when `quotaReq.anySet()`; this path validates whenever `quotaReq != nil`. Check whether the decode path can yield a non-nil-but-empty `quotaReq` — if so, `validateQuotaSetRequest` will reject it with "at least one of ... is required" where the admin path would have treated it as "no quota". Likely fine if decode only sets `quotaReq` when fields are present, but worth a glance so the two entry points reject identically.

### Good
- `initTenantSchemaAsync` now threads `tenant_id`/`provider` into the schema-init logger context via `logger.WithContext` — so the schema SQL logs carry tenant identity. Nice, and it's scoped to `schemaCtx` so it doesn't bleed into the outer `server_event` logs. Correct use.
- `replenishTenantPoolAsync` is fired after a successful claim, consistent with the admin path, keeping the pool topped up.
- Test coverage for "ordinary provision consumes a free pool tenant + applies spending limit" is the right case to lock in.

Net: ship-able once the `tenant_provision` metric events are added to the claim path (success + failure) — without that, this silently blinds provision monitoring for the pool case. The other two are minor cleanups.

srstack (commented):
Re-reviewed after `b2fdbde9 fix: align provision pool claim metrics`. Both my points are resolved — but a new reviewer (Codex) surfaced a real regression that should block merge. Details below.

### Resolved from my last pass
- **Metrics gap closed, and labels match `provisionTenant` exactly.** Claim success now emits `result="accepted"` (+ `quota="create_time_spending_limit"` when a spending limit is set), and claim failure emits `result="cluster_error"|"error"` via `provisionTenantPoolClaimMetricResult`. Verified these are the same label values `provisionTenant` uses (server.go:4521/4538/4413), so dashboards aggregate cleanly across claim vs. full-provision. No double-count since the claim path returns early.
- **Dead `quotaReq.TenantID = "pending"` removed.**
- Bonus: the client-facing error no longer interpolates `%v` of the internal error (`"claim tenant pool tenant failed"`), which is a small but real info-leak tightening. Good.

### Should fix before merge: hard 502 regression for non-lister provisioners (agree with Codex)
This is a behavior regression, not a nit. Trace:
- `firstManagedOrganization` returns a **non-nil error** (`"managed cluster list not enabled"`) when the provisioner is not a `ManagedClusterLister` (it's a hard `fmt.Errorf`, not an empty-orgID miss).
- `claimAdminTenantFromPool` does `orgID, err := s.firstManagedOrganization(...); if err != nil || orgID == "" { return nil, nil, false, err }` — so it returns the **error**, not a clean `claimed=false`.
- `handleProvision`'s new branch: `if ... err != nil { 502; return }` — so it **502s and never falls through to `provisionTenant`**.

Net: a TiDB Cloud native provisioner that implements `TenantPoolClusterManager` but **not** `ManagedClusterLister` could provision before this PR and now hard-fails `/v1/provision` with 502 as soon as it exposes pool-management methods. `TenantPoolClusterManager` doesn't require listing, so these are independently composable interfaces — the claim path must not make listing a precondition for *all* provisioning.

Fix options (either works):
1. In `claimAdminTenantFromPool`, treat "lister not available" as a **non-claim miss** (`return nil, nil, false, nil`) rather than an error — i.e. distinguish `ErrListNotEnabled` from real lookup errors and fall through. Cleanest, and fixes the admin path too.
2. In `handleProvision`, gate the whole claim block on `s.provisioner` implementing `ManagedClusterLister` before attempting it.

I'd prefer (1) — the pool claim is an optimization; inability to discover the org should degrade to normal provisioning, never break it. Whichever you pick, add a test: provisioner with `TenantPoolClusterManager` but no `ManagedClusterLister` → `/v1/provision` should still succeed via `provisionTenant`, not 502.

### Still worth a glance (from my prior review)
Validation-trigger parity: this path validates whenever `quotaReq != nil`; `handleAdminTenantCreate` uses `quotaReq.anySet()`. Confirm decode can't produce a non-nil-but-empty `quotaReq` (which would 400 here but be treated as "no quota" on the admin path).

Net: metrics fix is correct and complete. The non-lister 502 fallthrough is the one real blocker — once the claim path degrades to normal provisioning instead of erroring, this is good to go.

srstack (commented):
Re-reviewed after `4c099ebb fix: skip pool claim without cluster lister`. This resolves the 502 regression with the cleaner of the two options I suggested, plus a precise regression test. From my side this is good to merge.

### Confirmed fixed
- **Non-lister provisioners no longer hard-fail.** `claimAdminTenantFromPool` now checks `s.provisioner.(tenant.ManagedClusterLister)` up front and returns a clean `claimed=false` miss (not an error) when absent, logging `reason="managed_cluster_lister_unavailable"`. So `handleProvision` falls through to `provisionTenant` instead of 502ing. This is option (1) — fixing it in the claim helper also covers the admin-create path, which is the right place.
- **Test genuinely exercises the branch.** `tenantPoolNoListProvisioner` embeds `fakeProvisioner` (not `quotaTestProvisioner`, which is the only type with a `ListManagedClusters` method), so it really doesn't satisfy `ManagedClusterLister`. The test asserts 202 via the credential path (`credentialCalls==1`) and `markPoolUsedCalls==0`/`batchPoolCalls==0` — i.e. it verifies both that provisioning succeeds *and* that the pool path was fully skipped, not just that it didn't crash. Good test.

### Net
The PR landed cleanly across three follow-ups: pool-claim on `/v1/provision` → metric alignment (labels match `provisionTenant`, no double-count, dead `"pending"` removed, error message no longer leaks internal detail) → non-lister fallthrough fix with coverage. Mechanics are sound, scope is right (`ProviderTiDBCloudNative` + credential only, optimization that degrades to normal provisioning), and CI is green.

LGTM. Only leftover is the minor validation-trigger parity question from earlier (this path validates on `quotaReq != nil`, admin uses `quotaReq.anySet()`) — worth a glance for consistency but not blocking, since an empty quota body 400-ing on the public provision endpoint is defensible behavior.

qiffang (commented):
## Review — PR #643: Fix provision create tenant pool claim

### Overall

The PR correctly wires the tenant pool claim flow into `/v1/provision`, mirroring the existing `AdminCreateTenant` path. The `writeProvisionTenantAccepted` extraction, `ManagedClusterLister` guard, schema init logger context enrichment, and test coverage are all solid.

### Issues

#### 1. Pool claim error returns 502 without falling back to normal provision (server.go:3929)

When `claimAdminTenantFromPool` returns `(nil, nil, false, err)`, the provision handler returns `502 "claim tenant pool tenant failed"` immediately — the caller never gets a tenant. The admin create path (`admin_tenants.go:174`) has the same behavior, so this is consistent.

However, for the public `/v1/provision` path used by `drive9 create`, hard-failing on a pool claim error may be more disruptive than for admin-only calls. If the pool infrastructure has a transient issue (e.g. org lookup timeout), the user gets a 502 instead of falling back to normal cluster provisioning. Worth considering whether `err != nil` on this path should fall through to `provisionTenant` instead of returning early. Not blocking, but a design question.

#### 2. Quota validation only on credential-provision path (server.go:3911-3917)

`validateQuotaSetRequest` is called only when `credentialReq != nil` (the TiDB Cloud path). If a different provider path ever gets a `quotaReq`, it won't be validated. This is currently safe since only TiDB Cloud Native supports quota at provision time, but it's a latent gap.

#### 3. Missing metric event for quota validation failure (server.go:3913)

When `validateQuotaSetRequest` fails, the handler returns 400 without emitting a `metricEvent`. Compare with other early-return paths in the same function that emit `metricEvent(... "result", "error")`. This means quota validation failures won't show up in provision dashboards.

#### 4. `schemaCtx` vs `ctx` inconsistency in `initTenantSchemaAsync` (server.go:4693-4710)

`schemaCtx` (with tenant_id/provider fields) is used for `schemaInit(schemaCtx, tenantDSN)`, but the subsequent error logging, retry exhaustion, and `finalizeTenantSchemaInit` still use the original `ctx`. This means:
- `schema_init_retry_exhausted` log line (4702) won't have tenant_id/provider
- `schema_init_attempt_failed` log line (4705) already has them as explicit zap fields, so `schemaCtx` would be redundant there — but the inconsistency between which ctx is used where makes it easy to miss fields in future log lines.

Consider using `schemaCtx` everywhere in the function body after creating it, or just pass the extra fields into `schemaInit` without creating a new context.

#### 5. `tenantPoolNoListProvisioner` methods return errors but callers don't check (quota_test.go:223-235)

The `BatchProvisionFreeClustersWithCredentialsAndQuota`, `MarkClusterPoolUsed`, and `MarkClusterPoolFree` methods return errors, but `TestProvisionFallsBackWhenTenantPoolClaimCannotListManagedClusters` asserts they are never called (`calls == 0`). The error return values are essentially unreachable. This is fine for the test as written, but the methods could return `nil` to be clearer that the error path is intentionally untested here.

### Minor

- SDK docs update is comprehensive and correctly documents that pool get/update/delete do not claim tenants.
- The `provisionTenantPoolClaimMetricResult` helper correctly classifies quota permission/backend errors separately from generic errors.
- Test `TestProvisionClaimsFreePoolTenant` thoroughly verifies the claimed tenant's API key resolves, quota config is applied, and binding is marked used.

### Verdict

Issues 1-2 are design questions, not bugs. Issues 3-4 are minor gaps. No blocking issues found.

qiffang (approved):
GREEN on `448d3dbfdb77f33a99625e4df990bdda500bbd75`.

Reviewed the `/v1/provision` tenant-pool claim path, quota validation ordering, no-`ManagedClusterLister` fallback, shared accepted-response helper, schema-init logger context change, Go SDK docs, and regression tests.

No blocking findings.

Validation:
- Local `git diff --check origin/main...HEAD` passed.
- I could not run local Go tests on this machine because `go` is not installed; relied on CI plus static/code-path review.

Residual notes, non-blocking:
- Pool-claim errors from `/v1/provision` are intentionally returned as 502 with a generic message; quota request validation is already performed before claim, so invalid quota input should not consume or claim a pooled tenant.
- `initTenantSchemaAsync` now passes a logger-enriched `schemaCtx` into `schemaInit`; surrounding finalize/log calls still use the original context. This is logging consistency only, not a correctness issue.

pkg/server/admin_tenant_pool.go
```diff
@@ -698,6 +698,10 @@ func (s *Server) claimAdminTenantFromPool(ctx context.Context, cred tenant.Crede
 		logger.Info(ctx, "server_event", eventFields(ctx, "admin_tenant_pool_claim_skipped", "provider", tenant.ProviderTiDBCloudNative, "reason", "pool_manager_unavailable", "duration_ms", durationMillis(claimStarted))...)
 		return nil, nil, false, nil
 	}
+	if _, ok := s.provisioner.(tenant.ManagedClusterLister); !ok {
```
mornyx, line 701: Good guard for the "lister not implemented" case. Note this doesn't cover the runtime case where `ListManagedClusters` (called just below in `firstManagedOrganization`, line 706) returns an error — that path still does `return nil, nil, false, err`, which `/v1/provision` turns into a 502 without falling back to `provisionTenant`. See M1 in the review summary.

pkg/server/server.go
```diff
@@ -3916,6 +3923,31 @@ func (s *Server) handleProvision(w http.ResponseWriter, r *http.Request) {
 			return
 		}
 	}
+	if provider == tenant.ProviderTiDBCloudNative && credentialReq != nil {
+		poolClaimStarted := time.Now()
+		logger.Info(r.Context(), "server_event", eventFields(r.Context(), "provision_tenant_pool_claim_started", "provider", provider, "quota_requested", quotaReq != nil)...)
+		if res, pool, claimed, err := s.claimAdminTenantFromPool(r.Context(), *credentialReq, quotaReq); err != nil {
+			logger.Error(r.Context(), "server_event", eventFields(r.Context(), "provision_tenant_pool_claim_failed", "provider", provider, "duration_ms", durationMillis(poolClaimStarted), "error", err)...)
+			metricEvent(r.Context(), "tenant_provision", "provider", provider, "result", provisionTenantPoolClaimMetricResult(err))
+			errJSON(w, http.StatusBadGateway, "claim tenant pool tenant failed")
```
mornyx, line 3932: This 502 has no fallback to `provisionTenant` when the pool claim errors (e.g. a transient TiDB Cloud list error in `firstManagedOrganization`). The public `/v1/provision` path previously didn't depend on managed-cluster listing at all. Consider treating org-lookup errors as a non-claim miss here so provisions still succeed via the normal cluster-creation path. See M1.

pkg/server/server.go
```diff
@@ -4649,6 +4691,10 @@ func (s *Server) handleLocalTenantProvision(w http.ResponseWriter, r *http.Reque
 
 func (s *Server) initTenantSchemaAsync(ctx context.Context, tenantID, tenantDSN, provider string, schemaInit func(context.Context, string) error) {
 	ctx = ensureTrace(ctx)
+	ctx = logger.WithContext(ctx, logger.FromContext(ctx).With(
```
mornyx, line 4694: This bakes `tenant_id`/`provider` into the logger, but the `eventFields` calls below (e.g. the `schema_init_started` call right after) still pass the same keys explicitly, so each schema-init event ends up with duplicate `tenant_id` and `provider` fields (zap doesn't dedupe). Scoping the enriched ctx to the `schemaInit` closure, or dropping the explicit kv in the `eventFields` calls here, would avoid the dupes. See M2.

pkg/server/server.go
```diff
@@ -3908,6 +3908,13 @@ func (s *Server) handleProvision(w http.ResponseWriter, r *http.Request) {
 		}
 		credentialReq = decoded.Credential
 		quotaReq = decoded.Quota
+		if quotaReq != nil {
+			if err := s.validateQuotaSetRequest(*quotaReq); err != nil {
```
mornyx, line 3912: This validation failure emits the `result=error` metric but no `provision_invalid_request` server_event, unlike the other bad-request branches in this handler. Add a matching `logger.Warn(... "provision_invalid_request" ...)` for parity. See N1.

mornyx (commented):
Thanks for this PR — the claim-first-then-fallback structure on `/v1/provision` mirrors `handleAdminTenantCreate` cleanly, and the `ManagedClusterLister` guard, shared `writeProvisionTenantAccepted` helper, metric parity, and the quota-persistence assertion added in b2fdbde are all good. A few things worth addressing before merge:

### Major

**M1 — `/v1/provision` no longer falls back when the managed-cluster list call errors at runtime.**

The new `ManagedClusterLister` guard in `claimAdminTenantFromPool` (admin_tenant_pool.go:701) handles the case where the provisioner doesn't implement listing at all, and `TestProvisionFallsBackWhenTenantPoolClaimCannotListManagedClusters` covers that. But `firstManagedOrganization` (admin_tenant_pool.go:706) still returns a hard error when `ListManagedClusters` fails at runtime (TiDB Cloud 5xx, network timeout, transient 401). In `claimAdminTenantFromPool` that error is propagated (`return nil, nil, false, err`), and `/v1/provision` turns it into a 502 with no fallback to `provisionTenant` (server.go:3932).

This is a behavioral regression for the public provision path: previously a TiDB Cloud native provision with valid credentials went straight to `ProvisionWithCredentials` without listing clusters; now every provision must successfully list managed clusters first. For `/v1/admin/tenants` (admin, low-volume) the strictness is fine, but `/v1/provision` is public and higher-volume, so a transient TiDB Cloud control-plane error during listing would block all new provisions even though cluster creation itself may be unaffected.

Options:
- Treat an org-lookup error as a non-claim miss (`return nil, nil, false, nil`) on the provision path so it falls back to `provisionTenant`, keeping the hard error only for the admin path; or
- Add a test for the list-error case and document that managed-cluster listing is now a hard dependency for `/v1/provision`.

### Medium

**M2 — Duplicate `tenant_id` / `provider` fields in schema-init logs.**

The new `logger.WithContext(ctx, logger.FromContext(ctx).With(tenant_id, provider))` (server.go:4694) bakes `tenant_id` and `provider` into the logger, but the `server_event` calls in `initTenantSchemaAsync` still pass `"tenant_id", tenantID, "provider", provider` explicitly to `eventFields` (e.g. the `schema_init_started` call right after, and every `schema_init_*` event). zap doesn't deduplicate, so each schema-init event record will carry two `tenant_id` and two `provider` fields. The enrichment is a good idea for the lower-level SQL logs inside the `schemaInit` closure; consider scoping the enriched context to that closure only, or dropping the explicit kv args from the `eventFields` calls in this function.

### Minor / nits

- **N1 (server.go:3912)** — The new `validateQuotaSetRequest` failure path on `/v1/provision` emits the `tenant_provision result=error` metric but, unlike the other bad-request branches in this handler, logs no `provision_invalid_request` server_event. Add a `logger.Warn(... "provision_invalid_request" ...)` for consistency.
- **N2** — `validateQuotaSetRequest` runs twice when the pool claim misses: once in the new block (server.go:3912) and again inside `provisionTenant` (server.go:4414). Idempotent, just redundant.
- **N3** — `provisionTenantPoolClaimMetricResult` maps only `ErrQuotaPermissionDenied` / `ErrQuotaBackendNotFound` to `cluster_error`. A `MarkClusterPoolUsed` failure is also a cluster-side error but maps to plain `error`; align if dashboards expect cluster errors there.
- **N4** — No test covers the `TenantProvisioning` claimed-tenant path through `/v1/provision` (where `startProvisionedTenantSchemaInit` fires). `TestProvisionClaimsFreePoolTenant` only exercises the `TenantActive` path. A test with a `Provisioning` pool tenant would lock in the schema-init trigger.
- **Docs** — The guide says tenant creation "automatically tries to claim an active free tenant from the pool before falling back to normal TiDB Cloud cluster creation." Per M1, the fallback today only covers the no-free-tenant / no-lister cases, not a runtime list error. Worth aligning the wording once M1 is resolved.

Overall the direction looks right; M1 is the one I'd want settled before this lands on the public provision endpoint.

PR #644, opened 2026-06-29
docs: design doc for content-hidden policy tag

qiffang:
## Summary

Design doc for the `drive9.content_hidden` reserved tag:

- Real path stays visible in `ls`/FUSE; tag flips `stat.size` to `0` and `read` to EOF.
- `ReadPlanCtx` + `ReadInlinePlanCtx` mask **before** any `PresignGetObject` (S3 leak is the highest-risk surface).
- Stat handler blanks `semantic_text`; grep/search/semantic endpoints filter hidden files (P0, not follow-up).
- Tag toggle bumps `inodes.revision + mtime` in the same tx as `ReplaceFileTagsTx`, emits SSE `file_attrs_changed`; FUSE listener calls `InodeNotify` / `EntryNotify`.
- Writes while hidden are **rejected** (`409` / `EROFS`), not accepted silently.
- Raw `ExecSQL` mutation of the reserved tag is documented as unsupported in P0.

Reflects the converged review from strategy-2 (msgs `6e779f6d`, `0c6d2b05`), dev-1 (msgs `85cf1089`, `579abbfe`), and adversary-2 (msg `1aaa22b5`).

## Test plan

- [ ] Review surface points: `ReadPlanCtx` masking before `PresignGetObject`, stat `semantic_text` blanking, search/grep filtering, raw-SQL bypass note
- [ ] Confirm tag key spelling `drive9.content_hidden` (dot)
- [ ] Confirm size=0 lock (not real-size)
- [ ] Confirm write-rejection policy
- [ ] Ack before implementation begins

Refs Raft `#drive9` design discussion `7668a1bf` → `579abbfe`.

## Summary by CodeRabbit

* **Documentation**
  * Added a design document describing how hidden content should behave across listing, reading, search, caching, and mounted clients.
  * Clarified expected responses for hidden items, including redacted reads, consistent metadata updates, and client-side cache invalidation behavior.
  * Documented write restrictions, edge cases, test coverage, rollout notes, and known risks for the new hidden-content policy.

### DIFF
--- a/docs/design/content-hidden-tag.md
+++ b/docs/design/content-hidden-tag.md
@@ -268,11 +268,27 @@ Surfaces that need explicit hidden-tag handling in this PR:
 
 | Surface | Where | Required behavior |
 |---|---|---|
-| Stat metadata `semantic_text` | stat handler at `pkg/server/server.go:1719` | blank `semantic_text` when hidden |
+| Stat metadata `semantic_text` | stat handler at `pkg/server/server.go:1753` | blank `semantic_text` when hidden |
+| Stat metadata `content_type` | stat handler `pkg/server/server.go:1752` | blank to `""` or `"application/octet-stream"` — leaking MIME reveals class (`application/pdf` → "this is a doc scan"). Default: blank to `""`. |
+| Stat metadata `checksum_sha256` | `pkg/server/fs_layer.go:66` and `git_workspace.go:84,95,109,123,138` | blank when hidden — content hash enables rainbow-table identification against known-public files (forms, ID-doc templates) |
+| Stat metadata `tags` (sibling system-derived) | stat handler `pkg/server/server.go:1754` | strip `drive9.image.*` and other `drive9.<derived>.*` system tags (they expose EXIF / extracted-content info); keep `drive9.content_hidden` itself and user-set business tags |
 | Inline batch read | `Dat9Backend.ReadInlinePlanCtx` (`pkg/backend/dat9.go:1642`) | empty `InlineData`, no `PresignURL` |
 | Grep / search / snippet | server search endpoints (find via `grep -r "snippet"` in `pkg/server`) | skip hidden files or return empty matches |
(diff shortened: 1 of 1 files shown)


qiffang:
**adversary-2 review (channel msg `5a9f5ac2` invitation)** — sign-off on the design with 3 additional leak-vector additions to §6bis (msg references on Slock).

The §6bis table covers `semantic_text`, inline batch read, grep/search/snippet, semantic embeddings, and `ETag`/`Last-Modified`. I'd add three more rows that are still real side channels for `stat`-style endpoints:

| Surface | Where | Required behavior |
|---|---|---|
| `checksum` / `sha256` in stat response | stat handler (same area as `semantic_text` at `pkg/server/server.go:1719`) | Blank `checksum` / `sha256` when hidden — leaking the content hash enables rainbow-table style identification against any known-public file (forms, templates, identity-document scans, etc.) |
| `content_type` / `mime` | stat handler | Class-level leak (`application/pdf` vs `image/jpeg` reveals "this is a passport scan" type info). P0 decision can be: hide it (`application/octet-stream`) OR accept as a known class-level leak — but the call should be explicit in the doc, not implicit |
| Sibling `file_tags` (e.g., `drive9.image.*`, business tags) returned by stat | stat handler / GetFileTags result | When `drive9.content_hidden=true`, the stat handler should at minimum suppress system-derived tags (`drive9.image.<...>` exposes "this is an image with these EXIF-derived fields"). User-set business tags is a product call; I lean toward "hide all non-`drive9.content_hidden` tags while hidden" to keep the mental model consistent (hidden file → no derivable content info), but flagging for explicit decision |

Severity: all 3 are **same family as `semantic_text`** (strategy-2 caught that one) — derived-from-content info leaked through stat metadata. The `checksum` row is the highest-risk of the 3 (cryptographic identity).

Not blocking ship. Add as additional rows to the §6bis table + 1 line each in the acceptance test matrix (§11 or wherever the §6bis cases get tested).

Approve design.

PR #647, opened 2026-06-30
feat: central outbox SSE scheme to avoid per-tenant TiDB polling

mornyx:
## Motivation

drive9 is preparing for deployments with a very large number of tenants. The current SSE cross-pod event notification scheme runs a **per-bus 1s poll goroutine** for every tenant that has an SSE subscriber (`EventBus.pollLoop` → `store.ListFSEventsSince` against that tenant's own TiDB). At this scale this means:

- A high QPS spread across many separate serverless tenant TiDBs
- Every tenant TiDB is kept awake by continuous polling, preventing **scale-to-zero**
- Idle tenant TiDBs consume **RCU continuously** even with zero actual workload — a cost increase that is unacceptable

## Design

Replace the per-bus TiDB polling with a **centralized three-layer scheme** that routes the "which tenants have new events" signal through the always-provisioned central meta DB instead of N separate serverless tenant TiDBs.

### Architecture

```
                    ┌──────────────────────────────────┐
                    │     Central Meta DB (always on)   │
                    │  sse_notify_outbox  (outbox)      │
                    │  pod_registry       (peer table)  │
                    │  pod_subscriptions  (routing)      │
                    └──────────────────────────────────┘
                          ▲                    ▲
        write (outbox)    │                    │ poll every 200ms (all pods)
                          │                    │
┌─────────────────────────┴──┐    ┌────────────┴──────────────────┐
│   Pod A (writer)            │    │   Pod B (subscriber)           │
│                             │    │                               │
│  publishEvent:              │    │  notifyPoller (200ms fallback) │
│   1. INSERT fs_events       │    │   → ListSSENotifySince(lastID) │
│      (tenant TiDB —         │    │   → getIfExists(tenantID)     │
│       unavoidable, this     │    │   → bus.Publish()             │
│       is a real write)      │    │   → EventsSince() reads       │
│   2. INSERT sse_notify_     │    │      tenant fs_events         │
│      outbox (meta DB)       │    │                               │
│   3. bus.Publish()          │    │  /v1/internal/sse-notify      │
│      (same-pod instant)     │    │   ← receives HTTP push        │
│   4. podNotifier.Notify()   │    │   → bus.Publish()             │
│      (cross-pod HTTP push)  │    │   → EventsSince() reads       │
│      fire-and-forget        │    │      tenant fs_events         │
└─────────────────────────────┘    └───────────────────────────────┘
```

### Layer 1: Central outbox (`sse_notify_outbox`)

A new lightweight table in the **central meta DB** (always provisioned, not serverless). Each FS mutation writes a single pointer row `{tenant_id, seq}` — the event content stays in the per-tenant `fs_events` table. This table only answers "tenant T has new events after seq S".

```sql
CREATE TABLE IF NOT EXISTS sse_notify_outbox (
    id         BIGINT UNSIGNED AUTO_INCREMENT PRIMARY KEY,
    tenant_id  VARCHAR(64) NOT NULL,
    seq        BIGINT UNSIGNED NOT NULL,
    created_at DATETIME(3) NOT NULL DEFAULT CURRENT_TIMESTAMP(3),
    INDEX idx_sse_notify_created (created_at)
);
```

Writing to this table never wakes a serverless tenant TiDB — the meta DB is always provisioned.

### Layer 2: Fallback poller (`notifyPoller`)

A **single per-pod goroutine** (`notify_poller.go`) reads the central outbox every **200ms** (default). For each outbox row, it calls `eventBuses.getIfExists(tenantID)` — if no local EventBus exists (no SSE subscribers for that tenant), the row is **silently skipped** and the tenant's TiDB is never queried. If a bus exists, `bus.Publish()` wakes the SSE handlers, which then call `EventsSince(lastSeen)` to read the actual event content from the tenant `fs_events` table (demand-driven, not poll-driven).

Key properties:
- **One goroutine per pod** (was one goroutine per tenant)
- Low QPS per pod against the always-provisioned meta DB (was per-tenant QPS against many TiDBs)
- **Idle tenant TiDBs are never queried** → they scale to zero
- Full batch drain: if a batch is full (1000 rows), immediately read the next batch without waiting for the tick

### Layer 3: Pod-to-pod HTTP push (`podNotifier`)

For **<10ms cross-pod latency**, a pod-to-pod HTTP push path (`pod_notifier.go`) runs alongside the poller. When `publishEvent` writes an event, it asynchronously POSTs `{tenant_id, seq}` to peer pods that have SSE subscribers for that tenant. The push is **fire-and-forget** (no retry, no response waiting) — the 200ms poller is the durable fallback.

Dynamic peer discovery via two new central tables:

```sql
CREATE TABLE IF NOT EXISTS pod_registry (
    pod_id         VARCHAR(64) PRIMARY KEY,
    addr           VARCHAR(255) NOT NULL,
    last_heartbeat DATETIME(3) NOT NULL,
    status         VARCHAR(20) NOT NULL DEFAULT 'active',
    INDEX idx_pod_heartbeat (last_heartbeat)
);

CREATE TABLE IF NOT EXISTS pod_subscriptions (
    pod_id     VARCHAR(64) NOT NULL,
    tenant_id  VARCHAR(64) NOT NULL,
    PRIMARY KEY (pod_id, tenant_id)
);
```

Each pod:
- **Heartbeats** (`pod_registry.go`) every 10s so peers know it's alive
- **Reports its subscriber set** every 5s to `pod_subscriptions` (which tenants have local SSE connections)
- The pod notifier builds a **tenant→peer reverse index** from `pod_subscriptions` every 5s and caches it for lock-free reads in `Notify`

The leader additionally sweeps stale pods (30s heartbeat threshold) and cleans their subscription rows.

### Internal endpoint

```
POST /v1/internal/sse-notify
Authorization: Bearer <DRIVE9_POD_NOTIFY_SECRET>
Content-Type: application/json

{"tenant_id": "T", "seq": 42}
```

Registered directly on the mux (not through tenant auth middleware). Authenticated via a shared internal bearer secret compared in constant time. The handler calls `eventBuses.getIfExists(tenantID)` → `bus.Publish()`.

## Correctness guarantees

| Property | Mechanism |
|----------|-----------|
| **No missed events** | Triple fallback: HTTP push (<10ms) + outbox poller (200ms) + `fs_events` replay on SSE client reconnect (Phase 1 `EventsSince(since)`) |
| **No duplicate delivery** | `EventsSince(lastSeen)` uses `WHERE seq > lastSeen` — idempotent. Multiple `Publish` wakeups coalesce (notify channel buffer=1) into a single `pollAndSend` |
| **No misrouted events** | `getIfExists(tenantID)` does exact string match. Each bus reads only its own tenant's `fs_events` table. Physical tenant isolation unchanged |
| **Real-time delivery** | Same-pod: <1ms (in-memory channel). Cross-pod: <10ms (HTTP push) or ~200ms (poller fallback) |
| **Idle tenant RCU = 0** | No per-bus poll. Tenant TiDB is only touched when (a) a write actually happens (unavoidable) and (b) a subscriber exists to read it (demand-driven) |

## Cost analysis

| Metric | Before | After |
|--------|--------|-------|
| Cross-pod polling | N_tenants × 1 QPS against N separate TiDBs | N_pods × 5 QPS against 1 always-provisioned meta DB |
| Idle tenant TiDB RCU | **Continuous** (prevents scale-to-zero) | **Zero** (no queries → TiDB sleeps) |
| Cost reduction | — | **~4 orders of magnitude** |

## EventBus simplification

The `EventBus` struct no longer runs a per-bus poll goroutine:
- **Removed**: `pollLoop`, `pollOnce`, `pollCancel`, `pollDone`, `pollLast` fields and methods
- **Simplified**: `Subscribe()` / `Unsubscribe()` only manage listener registration (no goroutine lifecycle)
- **Unchanged**: `Publish()`, `EventsSince()`, `Seq()` — the durable cursor logic and replay/reset semantics are identical
- **New**: `HasListeners()` (for `activeTenantIDs`), `TenantID()` (for logging)

The SSE handler (`handleEvents`) flow is unchanged — Phase 1 replay + Phase 2 live stream. The only difference is that `notify` channel wakeups now come from two sources (same-pod `Publish` + `notifyPoller`/`podNotifier`), but the handler doesn't distinguish them.

## Configuration

New environment variables:

| Env var | Default | Purpose |
|---------|---------|---------|
| `DRIVE9_POD_ID` | (unset) | Unique pod identifier. When set, enables the new SSE scheme (pod registry + notifier). When unset, the old same-pod-only behavior is used (single-tenant / fallback mode) |
| `DRIVE9_POD_ADDR` | `DRIVE9_LISTEN_ADDR` | Internally reachable address (host:port) for peer push notifications |
| `DRIVE9_POD_NOTIFY_SECRET` | (unset) | Shared bearer token for `/v1/internal/sse-notify`. When set, enables the <10ms HTTP push path; when unset, only the 200ms poller fallback is used |
| `DRIVE9_SSE_NOTIFY_POLL_INTERVAL` | `200ms` | Fallback outbox poll interval |
| `DRIVE9_SSE_NOTIFY_RETENTION` | `1h` | How long outbox rows are kept before leader pruning |

All components are opt-in: in single-tenant mode (no `DRIVE9_META_DSN`) or when `DRIVE9_POD_ID` is unset, the server behaves exactly as before. The notify infrastructure is only activated when `cfg.Meta != nil` (multi-tenant mode).

## Files changed

| File | Change |
|------|--------|
| `pkg/meta/meta.go` | +3 tables (`sse_notify_outbox`, `pod_registry`, `pod_subscriptions`) + CRUD methods |
| `pkg/server/eventbus.go` | Removed per-bus poll goroutine; simplified `Subscribe`/`Unsubscribe`; added `HasListeners`/`TenantID` |
| `pkg/server/sse.go` | Added `getIfExists`/`activeTenantIDs`; rewrote `publishEvent` (4-step: fs_events → outbox → same-pod Publish → cross-pod push); new `handleInternalSSENotify` handler |
| `pkg/server/notify_poller.go` | **New**: global 200ms fallback poller |
| `pkg/server/pod_notifier.go` | **New**: pod-to-pod HTTP push with route cache |
| `pkg/server/pod_registry.go` | **New**: pod heartbeat + subscription reporting + stale-pod sweep |
| `pkg/server/server.go` | Config fields; Server struct fields; `startNotifyInfrastructure`/`stopNotifyInfrastructure`; `cleanupSSENotifyOutbox`; leader-gated outbox cleanup + stale pod sweep tickers; route registration |
| `pkg/server/instrumentation.go` | Route label for `/v1/internal/sse-notify` |
| `cmd/drive9-server/main.go` | Env var reading + Config passing + usage docs |

## Testing

**Unit tests** (`notify_poller_test.go`, `pod_notifier_test.go`):
- Poller discovers cross-pod outbox rows and wakes subscribers
- Poller skips tenants with no local subscribers (no TiDB query)
- Poller drains full batches immediately
- Notifier pushes to peers and wakes subscribers via HTTP
- Notifier is fire-and-forget (doesn't block on unreachable peers)
- Internal endpoint auth (valid/invalid/missing/no-secret)

**Integration tests** (`sse_outbox_integration_test.go`) — two-pod simulation with shared Meta DB:
- Cross-pod outbox poller delivery (200ms fallback path)
- Cross-pod HTTP push delivery (<10ms primary path)
- Pod registry heartbeat + active pod listing
- Subscription reporting + stale entry pruning
- Internal endpoint auth rejection
- Stale pod sweep (leader marks expired pods + cleans subscriptions)
- Outbox retention cleanup (leader prunes old rows)

All 37 SSE-related tests pass (16 unit + 7 integration + 14 existing regression). Verified on the `drive9-test` machine with a real MySQL instance.

## Summary by CodeRabbit

* **New Features**
  * Added cross-pod SSE delivery, so event updates can reach subscribers even when they’re connected to a different server pod.
  * Added internal pod coordination for faster notification routing and automatic cleanup of old event data.
  * Improved startup handling to configure SSE notification behavior through new environment settings.

* **Bug Fixes**
  * Reduced missed or delayed SSE notifications by adding a fallback poller and direct push path between pods.
  * Improved handling of inactive pods and stale subscription data to keep notifications accurate.

### DIFF
--- a/cmd/drive9-server/main.go
+++ b/cmd/drive9-server/main.go
@@ -368,13 +368,14 @@ func main() {
 	podID := strings.TrimSpace(os.Getenv("DRIVE9_POD_ID"))
 	podAddr := strings.TrimSpace(os.Getenv("DRIVE9_POD_ADDR"))
 	// Default podAddr to the listen addr only if it's a real, non-wildcard
-	// address. Wildcard binds like ":9009", "0.0.0.0:9009", or "[::]:9009" are
-	// not dialable by peers, so we leave podAddr empty in that case — the
+	// address. The notifier requires a full base URL (http://host:port) for
+	// http.NewRequest. Wildcard binds like ":9009", "0.0.0.0:9009", or "[::]:9009"
+	// are not dialable by peers, so we leave podAddr empty in that case — the
 	// podRegistry still reports subscriptions and the leader sweeps stale pods,
 	// but cross-pod HTTP push is disabled until an explicit address is set.
 	if podAddr == "" && podID != "" {
 		if host, port, err := net.SplitHostPort(addr); err == nil && host != "" && host != "0.0.0.0" && host != "::" {
-			podAddr = net.JoinHostPort(host, port)
+			podAddr = "http://" + net.JoinHostPort(host, port)
 		}
 	}
 	var podNotifySecret []byte
@@ -556,11 +557,12 @@ environment:
   SSE cross-pod notification (set DRIVE9_POD_ID to enable; required for large multi-tenant deployments):
   DRIVE9_POD_ID     unique pod identifier; when set, this pod registers in the central
                     pod_registry and reports its SSE subscriber tenant set so peers can
-                    route push notifications. Enables the new SSE scheme that avoids
-                    polling every tenant's TiDB (idle tenant TiDBs can scale to zero).
-  DRIVE9_POD_ADDR   internally reachable address (host:port) for peer push notifications
-                    (default: DRIVE9_LISTEN_ADDR; in multi-host deployments set to the
-                    address reachable by other pods)
+                    route push notifications. The outbox poller runs on all multi-tenant
+                    pods regardless; cross-pod HTTP push requires DRIVE9_POD_NOTIFY_SECRET.
+  DRIVE9_POD_ADDR   full base URL (e.g. http://10.0.0.5:9009) reachable by other pods for
+                    peer push notifications. Defaults to http://<listen-addr> when the
+                    listen address is a non-wildcard host; must be set explicitly in
+                    multi-host deployments.
   DRIVE9_POD_NOTIFY_SECRET  shared bearer token for /v1/internal/sse-notify pod-to-pod
                     push authentication. When set, enables the <10ms cross-pod HTTP push
                     path; when unset, only the 200ms fallback poller is used.
--- a/cmd/drive9-server/main.go
+++ b/cmd/drive9-server/main.go
@@ -357,6 +357,34 @@ func main() {
 		logger.Info(context.Background(), "slock_oauth_disabled")
 	}
 
+	// SSE cross-pod notification configuration. These enable the new scheme
+	// that avoids polling every tenant's TiDB. The notify poller reads the
+	// central sse_notify_outbox (always-provisioned meta DB) at 200ms; the
+	// pod-to-pod HTTP push provides <10ms latency when PodID/PodAddr/secret
+	// are set. See pkg/server/notify_poller.go and pod_notifier.go.
+	sseNotifyPollInterval := envDuration("DRIVE9_SSE_NOTIFY_POLL_INTERVAL", 200*time.Millisecond)
+	sseNotifyRetention := envDuration("DRIVE9_SSE_NOTIFY_RETENTION", time.Hour)
+	podID := strings.TrimSpace(os.Getenv("DRIVE9_POD_ID"))
+	podAddr := strings.TrimSpace(os.Getenv("DRIVE9_POD_ADDR"))
+	// Default podAddr to the listen addr if not explicitly set, so single-host
+	// deployments work without extra configuration. Multi-host deployments must
+	// set DRIVE9_POD_ADDR to the internally reachable address.
+	if podAddr == "" && podID != "" {
+		podAddr = addr
+	}
+	var podNotifySecret []byte
+	if raw := strings.TrimSpace(os.Getenv("DRIVE9_POD_NOTIFY_SECRET")); raw != "" {
+		podNotifySecret = []byte(raw)
+	}
+	if podID != "" {
+		logger.Info(context.Background(), "sse_notify_config",
+			zap.String("pod_id", podID),
+			zap.String("pod_addr", podAddr),
+			zap.Duration("poll_interval", sseNotifyPollInterval),
+			zap.Duration("retention", sseNotifyRetention),
+			zap.Bool("push_enabled", podNotifySecret != nil))
+	}
+
 	die(server.NewWithConfig(server.Config{
 		Meta:                         store,
 		Pool:                         pool,
@@ -378,6 +406,11 @@ func main() {
 		TiDBAutoEmbeddingAPIBase:     autoEmbeddingAPIBase,
 		DisableDatabaseAutoEmbedding: disableDatabaseAutoEmbedding,
 		Leader:                       leaderManager,
+		SSENotifyPollInterval:        sseNotifyPollInterval,
+		SSENotifyRetention:           sseNotifyRetention,
+		PodID:                         podID,
+		PodAddr:                       podAddr,
+		PodNotifySecret:               podNotifySecret,
 	}).ListenAndServe(addr))
 }
 
@@ -514,6 +547,21 @@ environment:
   DRIVE9_SLOCK_API_ORIGIN Slock API origin (required when DRIVE9_SLOCK_ORIGIN is set)
   DRIVE9_SLOCK_CLIENT_ID Slock connected-app client id (required when DRIVE9_SLOCK_ORIGIN is set)
   DRIVE9_SLOCK_CLIENT_SECRET Slock connected-app client secret (required when DRIVE9_SLOCK_ORIGIN is set)
+
+  SSE cross-pod notification (set DRIVE9_POD_ID to enable; required for large multi-tenant deployments):
+  DRIVE9_POD_ID     unique pod identifier; when set, this pod registers in the central
+                    pod_registry and reports its SSE subscriber tenant set so peers can
+                    route push notifications. Enables the new SSE scheme that avoids
+                    polling every tenant's TiDB (idle tenant TiDBs can scale to zero).
+  DRIVE9_POD_ADDR   internally reachable address (host:port) for peer push notifications
+                    (default: DRIVE9_LISTEN_ADDR; in multi-host deployments set to the
+                    address reachable by other pods)
+  DRIVE9_POD_NOTIFY_SECRET  shared bearer token for /v1/internal/sse-notify pod-to-pod
+                    push authentication. When set, enables the <10ms cross-pod HTTP push
+                    path; when unset, only the 200ms fallback poller is used.
+  DRIVE9_SSE_NOTIFY_POLL_INTERVAL  fallback outbox poll interval (default: 200ms)
+  DRIVE9_SSE_NOTIFY_RETENTION      how long outbox rows are kept before leader pruning
+                    (default: 1h)
 
   S3 storage (set DRIVE9_S3_BUCKET to enable AWS S3, otherwise local mock):
   DRIVE9_S3_BUCKET   S3 bucket name (enables AWS S3 mode)
--- a/cmd/drive9-server/main.go
+++ b/cmd/drive9-server/main.go
@@ -12,6 +12,7 @@ import (
 	"encoding/hex"
 	"fmt"
 	"io"
+	"net"
 	"os"
 	"strconv"
 	"strings"
@@ -366,11 +367,15 @@ func main() {
 	sseNotifyRetention := envDuration("DRIVE9_SSE_NOTIFY_RETENTION", time.Hour)
 	podID := strings.TrimSpace(os.Getenv("DRIVE9_POD_ID"))
 	podAddr := strings.TrimSpace(os.Getenv("DRIVE9_POD_ADDR"))
-	// Default podAddr to the listen addr if not explicitly set, so single-host
-	// deployments work without extra configuration. Multi-host deployments must
-	// set DRIVE9_POD_ADDR to the internally reachable address.
+	// Default podAddr to the listen addr only if it's a real, non-wildcard
+	// address. Wildcard binds like ":9009", "0.0.0.0:9009", or "[::]:9009" are
+	// not dialable by peers, so we leave podAddr empty in that case — the
+	// podRegistry still reports subscriptions and the leader sweeps stale pods,
+	// but cross-pod HTTP push is disabled until an explicit address is set.
 	if podAddr == "" && podID != "" {
-		podAddr = addr
+		if host, port, err := net.SplitHostPort(addr); err == nil && host != "" && host != "0.0.0.0" && host != "::" {
+			podAddr = net.JoinHostPort(host, port)
+		}
 	}
 	var podNotifySecret []byte
 	if raw := strings.TrimSpace(os.Getenv("DRIVE9_POD_NOTIFY_SECRET")); raw != "" {
--- a/pkg/meta/meta.go
+++ b/pkg/meta/meta.go
@@ -3331,11 +3331,19 @@ func (s *Store) MarkStalePods(ctx context.Context, before time.Time) (n int64, e
 
 // DeletePod removes a pod and its subscriptions from the registry. Used when a
 // pod is decommissioned. Best-effort: stale pod cleanup handles the common case.
+// DeletePod removes a pod and its subscriptions from the registry. Both
+// deletions run in a single transaction so a partial failure doesn't leave
+// orphaned subscription rows. Used when a pod is decommissioned.
 func (s *Store) DeletePod(ctx context.Context, podID string) (err error) {
 	start := time.Now()
 	defer observeMeta(ctx, "delete_pod", start, &err)
-	_, err = s.db.ExecContext(ctx, `DELETE FROM pod_registry WHERE pod_id = ?`, podID)
-	return err
+	return s.InTx(ctx, func(tx *sql.Tx) error {
+		if _, err := tx.ExecContext(ctx, `DELETE FROM pod_subscriptions WHERE pod_id = ?`, podID); err != nil {
+			return err
+		}
+		_, err := tx.ExecContext(ctx, `DELETE FROM pod_registry WHERE pod_id = ?`, podID)
+		return err
+	})
 }
 
 // ---------------------------------------------------------------------------
@@ -3477,4 +3485,22 @@ func (s *Store) ListStalePods(ctx context.Context) (out []string, err error) {
 		out = append(out, podID)
 	}
 	return out, rows.Err()
+}
+
+// DeleteSubscriptionsForStalePods deletes pod_subscriptions rows whose pod is
+// currently stale, using a subquery join so the stale check and the delete are
+// atomic. This avoids a TOCTOU race where a pod recovers to active between
+// ListStalePods and DeletePodSubscriptions, which would delete subscriptions for
+// a live pod. Returns the number of deleted rows.
+func (s *Store) DeleteSubscriptionsForStalePods(ctx context.Context) (n int64, err error) {
+	start := time.Now()
+	defer observeMeta(ctx, "delete_subscriptions_for_stale_pods", start, &err)
+	res, err := s.db.ExecContext(ctx,
+		`DELETE ps FROM pod_subscriptions ps
+		 INNER JOIN pod_registry pr ON ps.pod_id = pr.pod_id
+		 WHERE pr.status = 'stale'`)
+	if err != nil {
+		return 0, err
+	}
+	return res.RowsAffected()
 }
--- a/pkg/meta/meta.go
+++ b/pkg/meta/meta.go
@@ -672,6 +672,45 @@ func metaInitSchemaStatements() []string {
 			INDEX idx_pending (status, created_at),
 			INDEX idx_tenant_order (tenant_id, id)
 		)`,
+		// sse_notify_outbox is a central notification pointer table for cross-pod
+		// SSE event fan-out. Each FS mutation writes a lightweight row here (tenant_id
+		// + seq) so that pods can discover which tenants have new events without
+		// polling every tenant's own TiDB. The event content stays in the per-tenant
+		// fs_events table; this table only signals "tenant T has new rows after seq".
+		// The central meta DB is always provisioned (not serverless), so polling it
+		// does not incur per-tenant RCU. Retention pruning is leader-gated.
+		`CREATE TABLE IF NOT EXISTS sse_notify_outbox (
+			id         BIGINT UNSIGNED AUTO_INCREMENT PRIMARY KEY,
+			tenant_id  VARCHAR(64) NOT NULL,
+			seq        BIGINT UNSIGNED NOT NULL,
+			created_at DATETIME(3) NOT NULL DEFAULT CURRENT_TIMESTAMP(3),
+			INDEX idx_sse_notify_created (created_at)
+		)`,
+		// pod_registry tracks all drive9-server pods for cross-pod SSE push
+		// notification routing. Each pod upserts its row on startup and heartbeats
+		// every ~10s. The leader marks pods with stale heartbeats as inactive so
+		// writers don't push to dead pods. This table lives in the central meta DB
+		// (always provisioned).
+		`CREATE TABLE IF NOT EXISTS pod_registry (
+			pod_id         VARCHAR(64) PRIMARY KEY,
+			addr           VARCHAR(255) NOT NULL,
+			last_heartbeat DATETIME(3) NOT NULL,
+			status         VARCHAR(20) NOT NULL DEFAULT 'active',
+			created_at     DATETIME(3) NOT NULL DEFAULT CURRENT_TIMESTAMP(3),
+			updated_at     DATETIME(3) NOT NULL DEFAULT CURRENT_TIMESTAMP(3) ON UPDATE CURRENT_TIMESTAMP(3),
+			INDEX idx_pod_heartbeat (last_heartbeat)
+		)`,
+		// pod_subscriptions maps each pod to the tenant IDs for which it has active
+		// SSE subscribers. Writers consult this (via a locally cached reverse index)
+		// to push notifications only to pods that care about a given tenant, avoiding
+		// wasteful fan-out to pods with no subscribers for that tenant. Each pod
+		// periodically upserts its current subscriber set and prunes stale entries.
+		`CREATE TABLE IF NOT EXISTS pod_subscriptions (
+			pod_id     VARCHAR(64) NOT NULL,
+			tenant_id VARCHAR(64) NOT NULL,
+			updated_at DATETIME(3) NOT NULL DEFAULT CURRENT_TIMESTAMP(3) ON UPDATE CURRENT_TIMESTAMP(3),
+			PRIMARY KEY (pod_id, tenant_id)
+		)`,
 	}
 }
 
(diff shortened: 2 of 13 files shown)


qiffang (changes requested):
## adversary-2 review — PR #647 (central outbox SSE scheme)

**Verdict: REVISE** — 2 blockers

---

### B1 (blocker): `podRegistry` goroutines not synchronized on shutdown

`podRegistry.Start()` spawns 2 goroutines (`heartbeatLoop`, `subscriptionLoop`) that are not tracked by any WaitGroup. In `stopNotifyInfrastructure()`:

```go
s.notifyCancel()        // cancels notifyCtx
s.notifyWG.Wait()       // only waits for the poller goroutine
s.podNotifier.Stop()    // waits for refreshLoop only
s.podRegistry = nil     // ← no barrier for registry goroutines
```

The registry goroutines respect `ctx.Done()`, but context cancellation is non-blocking — a goroutine mid-flight in `UpsertPod()` or `UpsertPodSubscriptions()` (DB call) will not return until the call completes or times out. Meanwhile `Close()` proceeds to tear down the meta store (`leader.Stop()` → worker drain → eventually pool close). If the DB call is in progress when the meta store closes, this is a use-after-close race on the `*sql.DB`.

**Fix**: Track `podRegistry` goroutines in `notifyWG` (or give `podRegistry` its own `sync.WaitGroup` and call `reg.Stop()` / `reg.Wait()` in `stopNotifyInfrastructure()`).

---

### B2 (blocker): `dispatchPush` goroutines not synchronized on shutdown

`podNotifier.Stop()` cancels `refreshCancel` and waits for `refreshLoop` to exit via `pn.wg.Wait()`. But `dispatchPush()` spawns a goroutine per peer per `Notify()` call:

```go
func (pn *podNotifier) dispatchPush(addr string, payload []byte) {
    go func() {
        // uses pn.client, pn.secret — up to 2s HTTP timeout
    }()
}
```

These goroutines are **not tracked** by `pn.wg`. After `Stop()` returns, in-flight push goroutines may still be executing, accessing `pn.client` and `pn.secret` while the caller considers the notifier stopped.

The 2s `http.Client.Timeout` bounds the goroutine lifetime, but that's a wall-clock bound, not a synchronization guarantee. Between `Stop()` returning and the goroutine exiting, the server may close shared resources (meta store, http transport pools) that the goroutine depends on.

**Fix**: Either (a) track dispatch goroutines in a separate WaitGroup and wait in `Stop()`, or (b) give the notifier a `done` context that `dispatchPush` checks before launching, and wait for all goroutines to drain on Stop.

---

### Non-blockers (no action required for merge)

1. **`int64` → `uint64` seq cast** (`publishEvent` line ~2080): `InsertFSEvent` returns `int64`, `InsertSSENotify` takes `uint64`. Safe because seq is an auto-increment ID (always positive), but a type inconsistency worth noting.

2. **`TestSSEOutboxCrossPodPushDelivery` 6s sleep**: Relies on wall-clock timing for route table refresh. Brittle in CI under load but functional.

3. **`parseInt` helper in test**: Avoids `strconv` import — trivial, not a correctness issue.

---

### What's good

- Architecture is clean: three well-separated components (notifyPoller, podNotifier, podRegistry) with clear responsibilities
- Comprehensive integration test suite (7 tests covering poller delivery, push delivery, heartbeat, subscription reporting, auth rejection, stale sweep, outbox cleanup)
- `handleInternalSSENotify` uses constant-time comparison for bearer token auth
- Best-effort semantics with clear fallback chain (local Publish → outbox poller → SSE reconnect replay)
- `getIfExists` / `activeTenantIDs` correctly skip idle tenants

qiffang (changes requested):
## adversary-1 review — PR #647

**Verdict: REQUEST_CHANGES** — 4 blockers.

### B1: central outbox write is still tied to the client request context

`pkg/server/sse.go:223` writes the durable `fs_events` row and `pkg/server/sse.go:242` writes the central `sse_notify_outbox` row with the same `r.Context()`. Once `seq > 0`, the per-tenant event row already exists. If the client disconnects or the request deadline cancels before/during the outbox insert, the notify-poller fallback never sees this event. A remote pod with an already-connected SSE client then depends only on direct push; if push is disabled, stale, or fails, that live stream can remain stale until another event or reconnect. Reconnect replay is not a fallback for an already-open stream.

Post-commit notification work should use a short non-client-cancelable context after the event row exists, and this needs a regression test that cancels the request context after `fs_events` succeeds but before outbox notification.

### B2: `DRIVE9_POD_ADDR` documentation/default produces invalid push URLs

`cmd/drive9-server/main.go:369-373` defaults `PodAddr` to the raw listen address, and the usage text describes it as `host:port`. But `pkg/server/pod_notifier.go:221-222` builds the request with `addr + sseNotifyInternalRoute`, which requires an absolute URL. Defaults like `:9009`, `0.0.0.0:9009`, or a documented value like `10.0.0.5:9009` produce an invalid URL and `dispatchPush` returns before sending. That silently disables the advertised `<10ms` push path and leaves only the 200ms poller fallback.

Either require/validate a full base URL (`http://host:port`) or normalize host:port before registering/dispatching. Do not default to a raw listen address that peers cannot use.

### B3: shutdown can nil-deref `s.podRegistry`

`Server.Close` calls `stopNotifyInfrastructure()` first (`pkg/server/server.go:557`), and that clears `s.podRegistry = nil` (`pkg/server/server.go:543`). Leader workers are stopped later (`pkg/server/server.go:569-578`). The stale-pod sweep leader goroutine dereferences `s.podRegistry` on every tick (`pkg/server/server.go:701-710`). A sweep tick between those two phases can panic the process during shutdown.

Stop leader workers before clearing notification-owned fields, or capture a stable registry pointer in the leader goroutine and synchronize its lifetime. This needs a shutdown/lifecycle regression because the race is timing-dependent.

### B4: direct push fan-out is unbounded and not tracked

`pkg/server/pod_notifier.go:211-238` starts one goroutine per peer per event, each of which can live until the HTTP timeout. There is no semaphore, worker pool, queue bound, or Stop-time drain for these dispatch goroutines. Under a bursty tenant with multiple subscribed pods, this can create unbounded goroutine/connection pressure, exactly on the scale path this PR is trying to harden.

The direct push path needs bounded concurrency/backpressure and a shutdown drain/cancel strategy while preserving the outbox poller fallback for dropped/deferred pushes.

### Validation

- Reviewed head: `598af695aa9ba1bfc292e320bc0fa515596b71de`
- `git diff --check 8696e53235fe1344d3bf90fce2e70e14a9303062...HEAD` passed.
- Local Go tests were not run because this machine has no `go` binary (`go: command not found`).

qiffang (changes requested):
## adversary-2 review — PR #647 (central outbox SSE scheme)

**Verdict: REVISE** — 2 blockers

---

### B1 (blocker): `podRegistry` goroutines not synchronized on shutdown

`podRegistry.Start()` spawns 2 goroutines (`heartbeatLoop`, `subscriptionLoop`) that are not tracked by any WaitGroup. In `stopNotifyInfrastructure()`:

```go
s.notifyCancel()        // cancels notifyCtx
s.notifyWG.Wait()       // only waits for the poller goroutine
s.podNotifier.Stop()    // waits for refreshLoop only
s.podRegistry = nil     // ← no barrier for registry goroutines
```

The registry goroutines respect `ctx.Done()`, but context cancellation is non-blocking — a goroutine mid-flight in `UpsertPod()` or `UpsertPodSubscriptions()` (DB call) will not return until the call completes or times out. Meanwhile `Close()` proceeds to tear down the meta store. If the DB call is in progress when the meta store closes, this is a use-after-close race on the `*sql.DB`.

**Fix**: Track `podRegistry` goroutines in `notifyWG` (or give `podRegistry` its own `sync.WaitGroup` and call `reg.Stop()` / `reg.Wait()` in `stopNotifyInfrastructure()`).

---

### B2 (blocker): `dispatchPush` goroutines not synchronized on shutdown

`podNotifier.Stop()` cancels `refreshCancel` and waits for `refreshLoop` to exit via `pn.wg.Wait()`. But `dispatchPush()` spawns a goroutine per peer per `Notify()` call:

```go
func (pn *podNotifier) dispatchPush(addr string, payload []byte) {
    go func() {
        // uses pn.client, pn.secret — up to 2s HTTP timeout
    }()
}
```

These goroutines are **not tracked** by `pn.wg`. After `Stop()` returns, in-flight push goroutines may still be executing, accessing `pn.client` and `pn.secret` while the caller considers the notifier stopped.

The 2s `http.Client.Timeout` bounds the goroutine lifetime, but that's a wall-clock bound, not a synchronization guarantee. Between `Stop()` returning and the goroutine exiting, the server may close shared resources (meta store, http transport pools) that the goroutine depends on.

**Fix**: Either (a) track dispatch goroutines in a separate WaitGroup and wait in `Stop()`, or (b) give the notifier a `done` context that `dispatchPush` checks before launching, and wait for all goroutines to drain on Stop.

---

### Non-blockers (no action required for merge)

1. **`int64` → `uint64` seq cast** (`publishEvent` line ~2080): `InsertFSEvent` returns `int64`, `InsertSSENotify` takes `uint64`. Safe because seq is an auto-increment ID (always positive), but a type inconsistency worth noting.

2. **`TestSSEOutboxCrossPodPushDelivery` 6s sleep**: Relies on wall-clock timing for route table refresh. Brittle in CI under load but functional.

3. **`parseInt` helper in test**: Avoids `strconv` import — trivial, not a correctness issue.

---

### What's good

- Architecture is clean: three well-separated components (notifyPoller, podNotifier, podRegistry) with clear responsibilities
- Comprehensive integration test suite (7 tests covering poller delivery, push delivery, heartbeat, subscription reporting, auth rejection, stale sweep, outbox cleanup)
- `handleInternalSSENotify` uses constant-time comparison for bearer token auth
- Best-effort semantics with clear fallback chain (local Publish → outbox poller → SSE reconnect replay)
- `getIfExists` / `activeTenantIDs` correctly skip idle tenants

qiffang (commented):
## PR #647 Review — "Central outbox SSE scheme to avoid per-tenant TiDB polling"

**Head**: `598af69` | **Scope**: 2673 lines across 9 files — meta schema + CRUD, EventBus simplification, notifyPoller, podNotifier, podRegistry, server wiring, 37 tests.

### Verdict: **GREEN** — no blockers

This is a clean, well-motivated architecture change. The three-layer design (central outbox → fallback poller → HTTP push) correctly eliminates per-tenant TiDB polling while maintaining correctness guarantees. The code quality is high and test coverage is comprehensive.

---

### Architecture assessment

The design is sound:
- **Layer 1 (outbox)**: Lightweight pointer rows in always-provisioned meta DB. Correct — never wakes serverless tenant TiDB.
- **Layer 2 (poller)**: Single per-pod goroutine at 200ms, replacing N per-tenant 1s goroutines. Batch drain handles burst correctly.
- **Layer 3 (HTTP push)**: Fire-and-forget with route cache. No retry = no head-of-line blocking; poller is the durable fallback.

Correctness properties are well-defended:
- No missed events: triple fallback (push + poller + SSE reconnect replay)
- No duplicate delivery: `EventsSince(lastSeen)` is idempotent, notify channel coalesces (buffer=1)
- Idle tenant RCU = 0: `getIfExists` returns nil for tenants without local subscribers

---

### Medium issues

**1. `dispatchPush` spawns unbounded goroutines**

Each `Notify` call spawns one goroutine per target peer via `dispatchPush`. Under burst writes (many events for tenants with many peer subscribers), this could spawn many concurrent goroutines. The 2s push timeout bounds their lifetime, but there's no goroutine pool or backpressure.

For the current deployment size this is likely fine (few pods, push is fast), but worth noting for future scale. Consider a bounded worker pool or semaphore if pod count grows.

**2. `activeTenantIDs` holds read lock while iterating all buses + checking `HasListeners`**

`HasListeners` acquires the bus's own mutex while `activeTenantIDs` holds the `eventBuses` read lock. This is safe (no circular locking — buses lock never acquires ebs lock), but with many buses the lock hold time scales linearly. At 100k tenants with buses, this could be slow. The 5s subscription reporting interval gives headroom, but monitor this.

**3. `UpsertPodSubscriptions` uses per-row INSERT in a transaction**

For pods with many tenant subscribers, this issues N INSERT statements inside a single transaction. A batch INSERT (`VALUES (?,?), (?,?), ...`) would be more efficient. Not a correctness issue but impacts latency at scale.

---

### Nits

**4.** Several new files are missing trailing newlines (`notify_poller.go`, `pod_notifier.go`, `pod_registry.go`, `pod_notifier_test.go`, `sse_outbox_integration_test.go`). Minor but `git diff --check` would flag these.

**5.** `TestPodNotifierPushWakesSubscriber` and `TestPodNotifierFireAndForget` sleep for 6s waiting for route refresh. Consider exposing a `forceRefresh()` test helper to avoid slow tests.

**6.** `parseInt` in `sse_outbox_integration_test.go` — could use `strconv.Atoi` instead of a custom helper.

**7.** In `server.go`, the closing brace formatting looks off around line 2765 (`podNotifySecret` assignment ends with `}` on a line that also starts `// Default SSE notify retention`). This is likely a diff rendering artifact but worth verifying the actual code compiles cleanly.

**8.** `publishEvent` captures `seq` from `InsertFSEvent` — good. The outbox INSERT is correctly gated on `seq > 0`, preventing orphan outbox rows when the tenant fs_events INSERT fails.

---

### Test coverage — excellent

- 3 unit tests for notifyPoller (cross-pod discovery, skip-no-subscribers, batch drain)
- 4 unit tests for podNotifier (push wakes subscriber, fire-and-forget, no-peers noop, internal endpoint auth)
- 7 integration tests with two-pod simulation (poller delivery, push delivery, heartbeat, subscription reporting, endpoint rejection, stale pod sweep, outbox cleanup)
- Existing eventbus tests updated (removed per-bus poll tests, kept fast-return regression test)

### Summary

Solid architecture change that correctly eliminates the per-tenant polling cost bottleneck. The three-layer fallback design is well-motivated and comprehensively tested. The medium issues are all about scale headroom, not correctness. Ship it.

qiffang (changes requested):
## dat9-dev1 — REVISE (corrected from initial GREEN)

**Verdict: REVISE — 4 blockers**

After re-evaluating against @adversary-1 and @adversary-2's findings, I'm correcting my initial GREEN to REVISE. All 4 blockers are valid issues I missed.

### Blockers

**B1: `publishEvent` outbox INSERT uses request context**
`publishEvent` captures `ctx := r.Context()` and passes it to `InsertSSENotify`. If the client disconnects after `InsertFSEvent` succeeds but before `InsertSSENotify` completes, the outbox write is cancelled. The event exists in `fs_events` but has no outbox pointer — if pod-to-pod push also fails (peer unreachable), the `notifyPoller` on remote pods will never discover this event. SSE clients on those pods miss it silently. Fix: use `context.Background()` (or a detached context with timeout) for the outbox INSERT, matching the fire-and-forget semantics of steps 3 and 4.

**B2: Default `PodAddr` produces invalid URL for `dispatchPush`**
When `DRIVE9_POD_ADDR` is unset, `podAddr` defaults to the raw listen address (e.g. `:8080`). `dispatchPush` concatenates `addr + sseNotifyInternalRoute` → `:8080/v1/internal/sse-notify`, which is not a valid URL. `http.NewRequest` returns an error, and push silently fails on every default/single-host deployment. The `<10ms push path` documented in the design is dead on arrival unless the operator explicitly sets the env var — but the code/comment suggests it should work by default.

**B3: `Close()` ordering — `podRegistry` nil'd before leader goroutines stop**
`stopNotifyInfrastructure()` is called before `leader.Stop()` in `Close()`. `stopNotifyInfrastructure` sets `s.podRegistry = nil`, but the stale-pod sweep leader goroutine accesses `s.podRegistry.SweepStalePods()` on a 30s ticker. If the ticker fires between `s.podRegistry = nil` and `leader.Stop()`, it's a nil pointer dereference.

**B4: Unbounded `dispatchPush` goroutines not tracked on shutdown**
`dispatchPush` spawns a `go func()` per peer per event, not tracked by `pn.wg`. `podNotifier.Stop()` only waits for `refreshLoop` via `pn.wg.Wait()`, then returns while push goroutines may still be in-flight accessing `pn.client`. Similarly, `podRegistry.Start()` launches 2 goroutines (`heartbeatLoop`, `subscriptionLoop`) not tracked by `notifyWG`, so `stopNotifyInfrastructure` can return while they still access the meta store.

### Previously noted (medium, unchanged)
- M1: `notifyPoller` single-row INSERT per event — consider batch INSERT for high-throughput tenants
- M2: `eventBuses.mu` global lock in `getIfExists` / `activeTenantIDs` — may bottleneck under many tenants
- M3: Fire-and-forget push with no retry — acceptable given poller fallback, but metrics on push failure rate would help operational visibility

Head: `598af695aa9ba1bfc292e320bc0fa515596b71de`

mornyx:
@qiffang Thanks for the thorough reviews. All 6 unique blockers (after dedup across the 5 reviews) are now fixed in commits `341826f` and `2df4bb2`. Here's the mapping:

### Blockers fixed

**podRegistry goroutines not synced on shutdown** (adversary-2, duplicated across 2 reviews)
- `podRegistry` now has its own `sync.WaitGroup` + `Stop()` method that waits for `heartbeatLoop` and `subscriptionLoop` to exit before returning. `stopNotifyInfrastructure()` calls `reg.Stop()` so in-flight DB calls complete before the meta store closes.

**dispatchPush goroutines not synced on shutdown** (adversary-2, duplicated across 2 reviews; also adversary-1 B4)
- `dispatchPush` goroutines are now tracked by `podNotifier.wg`. `Stop()` cancels `dispatchCtx` (which cancels all in-flight HTTP requests via per-request `context.WithTimeout`) and waits for all goroutines to exit. Replaced `http.Client{Timeout: ...}` with per-request context timeout via `http.NewRequestWithContext`.

**outbox write uses request context** (adversary-1 B1)
- `publishEvent` now uses `context.WithoutCancel(ctx)` wrapped in a 5s timeout for `InsertSSENotify`, so a client disconnect after the `fs_events` write doesn't abort the outbox pointer. This prevents the scenario where a live SSE stream on a remote pod misses the event because the poller fallback has no outbox row.

**PodAddr default produces invalid URL** (adversary-1 B2)
- The default now produces `http://host:port` (with scheme) instead of bare `host:port`. Wildcard binds (`:9009`, `0.0.0.0:9009`, `::`) are still skipped — the push path is disabled in that case, but the poller fallback and pod registry still work. Usage docs updated to clarify the full base URL format.

**Close() ordering: podRegistry nil'd before leader stop** (adversary-1 B3)
- `Close()` now stops leader-gated workers (`leader.Stop()` + `stopLeaderWorkers()`) **before** `stopNotifyInfrastructure()`. This ensures the stale-pod sweep ticker (which dereferences `s.podRegistry`) can't fire after `podRegistry` is cleared.

**unbounded dispatchPush concurrency** (adversary-1 B4)
- Added a non-blocking semaphore (`pushSem`, capacity `maxConcurrentPushes=64`) to `dispatchPush`. When at capacity, excess pushes are silently dropped (fire-and-forget) — the 200ms `notifyPoller` is the durable fallback. This prevents goroutine explosion under burst writes with many peer subscribers.

### Non-blockers addressed

- 6s test sleeps → replaced with synchronous `notifier.refresh()` calls (~15s faster suite)
- Test cleanup error swallowing → now fails fast via `t.Fatalf`
- Stale-pod sweep TOCTOU → replaced `ListStalePods` + loop delete with atomic `DeleteSubscriptionsForStalePods` (conditional JOIN delete)
- `notifyCtx` removed from `Server` struct (coding guidelines: don't store context in structs)
- Import grouping fixed in `notify_poller.go` (stdlib → third-party → internal)
- `DeletePod` now deletes both `pod_registry` and `pod_subscriptions` in a single transaction
- `RegisterBeforeStart` returns error instead of swallowing it
- Notify poller test cursor race fixed (explicit `lastID = 0`)

### Non-blockers intentionally not changed

- `int64 → uint64` seq cast: safe (auto-increment always positive), type inconsistency noted but not a correctness issue
- `parseInt` helper in test: trivial, not worth a `strconv` import just for this

All 37 SSE tests pass locally. The latest commit (`2df4bb2`) is pushed and CI is running.

qiffang (approved):
## dat9-dev1 — GREEN (re-review after fixes)

**Verdict: GREEN — all 4 blockers resolved**

Head: `2df4bb2197d9532c18c42a66d761c67a00f86e81` (2 new commits since `598af695`)

### Blocker resolution

**B1 (request context on outbox INSERT) — FIXED**
`InsertSSENotify` now uses `context.WithTimeout(context.WithoutCancel(ctx), 5*time.Second)`. Go 1.21's `WithoutCancel` detaches from request lifecycle while preserving trace values. 5s timeout prevents goroutine leak. Clean.

**B2 (PodAddr default → invalid URL) — FIXED**
Uses `net.SplitHostPort` to validate the listen address, rejects wildcard binds (`:9009`, `0.0.0.0:9009`, `[::]`), and prepends `http://` for dialable addresses. When addr is not dialable, `podAddr` stays empty — push disabled, poller fallback still works. Documentation updated to state "full base URL (e.g. http://10.0.0.5:9009)".

**B3 (Close ordering → nil deref) — FIXED**
`stopNotifyInfrastructure()` moved from before `leader.Stop()` to AFTER `s.stopLeaderWorkers()` in `Close()`. Leader stale-pod sweep goroutine is guaranteed stopped before `s.podRegistry` is nil'd.

**B4 (unbounded goroutines) — FIXED**
- `dispatchPush`: bounded by `pushSem` (chan, cap=64), tracked by `pn.wg`, uses `pn.dispatchCtx` with per-request timeout. Excess pushes silently dropped (fire-and-forget).
- `podNotifier.Stop()`: cancels both `refreshCancel` and `dispatchCancel`, then `wg.Wait()`.
- `podRegistry`: now has own `wg` tracking both goroutines + `Stop()` method. `stopNotifyInfrastructure` calls `reg.Stop()`.

### Additional improvements in these commits

- **TOCTOU fix in SweepStalePods**: Replaced two-step `ListStalePods` → per-pod `DeletePodSubscriptions` with atomic `DeleteSubscriptionsForStalePods` (conditional `DELETE JOIN` on `pr.status = 'stale'`). Pod recovering between mark and delete retains its subscriptions.
- **DeletePod transactional**: Both `pod_subscriptions` and `pod_registry` deletes now in a single transaction.
- **Test determinism**: Replaced `time.Sleep(6 * time.Second)` waits with synchronous `notifier.refresh()` calls. Poller tests set `np.lastID = 0` to avoid cursor race with insert.
- **Error-checked test cleanup**: `DELETE FROM` in test setup now fails the test on error instead of silently ignoring.

### Previously noted medium issues (unchanged, acceptable)
- M1: Single-row outbox INSERT per event (batch INSERT for high-throughput is a future optimization)
- M2: `eventBuses.mu` global lock (acceptable at current scale)
- M3: Fire-and-forget push semantics (correct given poller fallback; now bounded by semaphore)

qiffang (approved):
## adversary-2 re-review — PR #647 (central outbox SSE scheme)

**Verdict: GREEN** (head `2df4bb21`)

All 4 blockers from the previous round are fixed:

**B1 (request context)** — FIXED: `publishEvent` outbox insert now uses `context.WithoutCancel(ctx)` with a 5-second timeout, decoupled from the HTTP request lifecycle.

**B2 (PodAddr default)** — FIXED: Default `podAddr` now validates against wildcard addresses (`:port`, `0.0.0.0:port`, `[::]:port`) and prepends `http://` to produce a valid URL. Wildcard binds leave `podAddr` empty (push disabled, poller-only mode). Documentation updated to specify "full base URL" format.

**B3 (Close ordering → nil deref)** — FIXED: `stopNotifyInfrastructure()` now runs AFTER `stopLeaderWorkers()` in `Close()`, so the leader-gated `SweepStalePods` goroutine is guaranteed to have exited before `s.podRegistry` is cleared.

**B4 (unbounded goroutines)** — FIXED: 
- `dispatchPush` goroutines tracked by `pn.wg` and bounded by `pushSem` (max 64 concurrent). `Stop()` cancels `dispatchCtx` and waits via `wg.Wait()`.
- `podRegistry` has its own `sync.WaitGroup` tracking both heartbeat and subscription goroutines. `Stop()` cancels + waits.

Additional improvements in v2:
- Tests use direct `refresh()` calls instead of `time.Sleep(6s)` for route table population
- `pushSem` prevents goroutine explosion under burst writes
- `http.NewRequestWithContext` replaces raw `http.NewRequest` for per-request timeout

qiffang (changes requested):
## adversary-1 re-review — PR #647

**Verdict: REQUEST_CHANGES** — 1 remaining blocker on latest head `2df4bb2197d9532c18c42a66d761c67a00f86e81`.

The previous 4 blockers are addressed in direction:
- Outbox insert now uses `context.WithoutCancel(ctx)` + timeout.
- `PodAddr` default no longer publishes wildcard/raw listen addresses, and default generated addresses include `http://`.
- `Close()` now stops leader workers before clearing notify infrastructure.
- `podNotifier` dispatches are bounded and tracked; `podRegistry` goroutines are tracked and stopped.

### B1: notify poller cursor initialization can skip a live event during startup

`notifyPoller.run` initializes its cursor inside the background goroutine:

```go
maxID, err := np.metaStore.MaxSSENotifyID(ctx)
...
np.lastID = maxID
```

`startNotifyInfrastructure` starts that goroutine asynchronously and returns immediately. The server can then start accepting SSE connections and writes before the poller has executed `MaxSSENotifyID` and assigned `lastID`.

Failure sequence:
1. Pod B is constructed and starts listening, but its notify poller goroutine has not initialized `lastID` yet.
2. A client opens `/v1/events` on Pod B and receives Phase-1 heartbeat at current `headSeq`.
3. Pod A writes a new `fs_events` row and `sse_notify_outbox` row.
4. Pod B's poller finally runs `MaxSSENotifyID`, sees that new outbox row as the max, sets `lastID` to it, and never publishes it.
5. If direct push is disabled/stale/not routed yet, the already-open SSE stream on Pod B is stale until reconnect or a later event. Phase-2 has no periodic `fs_events` poll; it only reacts to `notify` and heartbeat.

The current tests do not eliminate this race. For example, `TestNotifyPollerDiscoversCrossPodEvents` sets `np.lastID = 0`, then calls `go np.run(ctx)`, but `run()` overwrites `lastID` with `MaxSSENotifyID`. If the insert happens before that initialization, the test itself can skip the inserted row and time out. The integration poller test has the same shape: infrastructure starts async, then the test inserts a row without waiting for poller readiness.

Required fix: make the initial cursor boundary deterministic before live traffic can rely on the poller. For example, initialize `lastID` synchronously before launching the poll loop or add an explicit readiness barrier and wait for it in startup/tests. Add a regression that proves a row inserted after poller readiness is delivered, and avoid tests that rely on scheduler ordering between `go np.run()` and `InsertSSENotify`.

### Validation

- `git diff --check 8696e53235fe1344d3bf90fce2e70e14a9303062...HEAD` passed.
- Local Go tests were not run because this machine has no `go` binary.

PR #649, opened 2026-06-30
fix: initialize notifyPoller cursor synchronously before live traffic

mornyx:
## Motivation

Fixes the remaining blocker from qiffang's adversary-1 re-review on PR #647 (the review was posted before merge but the issue was not addressed at that time).

### The race

`notifyPoller.run()` initialized its cursor (`lastID = MaxSSENotifyID()`) **inside the background goroutine**, asynchronously after `startNotifyInfrastructure` returned. A write between poller construction and cursor initialization would be skipped:

1. Pod B starts; poller goroutine has not yet executed `MaxSSENotifyID()`.
2. A client opens `/v1/events` on Pod B and receives a Phase-1 heartbeat at the current head seq.
3. Pod A writes a new `fs_events` + `sse_notify_outbox` row.
4. Pod B's poller finally runs `MaxSSENotifyID()`, sees that new row as the max, sets `lastID` to it, and **never publishes it**.
5. If direct push is also unavailable/stale, the already-open SSE stream on Pod B is stale until reconnect or a later event.

This is a real race that triggers on pod startup/restart (rolling updates, rescheduling) — exactly the scenario the outbox scheme is designed to handle correctly.

## Fix

Extract cursor initialization into a synchronous `initCursor()` method, called in `startNotifyInfrastructure` **before** launching the poller goroutine and before the server accepts live traffic:

```go
poller := newNotifyPoller(s.meta, s.events, cfg.SSENotifyPollInterval)
poller.initCursor(context.Background())  // synchronous — cursor ready before any traffic
s.notifyWG.Add(1)
go func() {
    defer s.notifyWG.Done()
    poller.run(notifyCtx)
}()
```

`run()` no longer initializes the cursor itself — it only logs and starts the ticker loop. The cursor is guaranteed to be set before any SSE connection or write can reach the poller.

## Files changed

| File | Change |
|------|--------|
| `pkg/server/notify_poller.go` | Extract `initCursor()` method; `run()` no longer initializes cursor |
| `pkg/server/server.go` | Call `poller.initCursor()` synchronously before `go poller.run()` |
| `pkg/server/notify_poller_test.go` | Tests call `initCursor()` before `run()`; removed `np.lastID = 0` workaround |

## Testing

All 10 poller + integration tests pass:
- `TestNotifyPollerDiscoversCrossPodEvents` — row inserted after `initCursor` is delivered
- `TestNotifyPollerSkipsTenantsWithoutSubscribers` — same, with skip verification
- `TestNotifyPollerBatchDrain` — full batch drain with `initCursor` + override
- All 7 `TestSSEOutbox*` integration tests pass (these use `NewWithConfig` which calls `initCursor` via `startNotifyInfrastructure`)

### DIFF
--- a/pkg/server/notify_poller.go
+++ b/pkg/server/notify_poller.go
@@ -61,23 +61,32 @@ func newNotifyPoller(metaStore *meta.Store, buses *eventBuses, interval time.Dur
 	}
 }
 
-// run starts the poller goroutine. It blocks until ctx is cancelled.
-func (np *notifyPoller) run(ctx context.Context) {
-	// Initialize cursor to current max id so we skip historical outbox rows.
-	// Events written before this pod started are delivered to clients via
-	// Phase-1 EventsSince replay on SSE reconnect, not via the poller.
+// initCursor synchronously initializes the poller's lastID to the current max
+// outbox id. This MUST be called before run() starts and before the server
+// accepts live traffic, so that a write between poller construction and the
+// first poll tick is not skipped. If the init query fails, lastID stays 0
+// (reads some historical rows but Publish is a no-op for tenants without
+// subscribers — not harmful).
+func (np *notifyPoller) initCursor(ctx context.Context) {
 	maxID, err := np.metaStore.MaxSSENotifyID(ctx)
 	if err != nil {
 		logger.Warn(ctx, "notify_poller_init_cursor_failed",
 			zap.Error(err))
-		// Start from 0 — worst case we read some historical rows and find no
-		// matching buses (Publish is a no-op). Not harmful.
 		maxID = 0
 	}
 	np.lastID = maxID
+	logger.Info(ctx, "notify_poller_cursor_initialized",
+		zap.Uint64("cursor", np.lastID))
+}
+
+// run starts the poller goroutine. It blocks until ctx is cancelled.
+// The cursor (lastID) must already be initialized by initCursor before run
+// is called, so that no outbox rows are skipped between construction and
+// the first poll tick.
+func (np *notifyPoller) run(ctx context.Context) {
 	logger.Info(ctx, "notify_poller_started",
 		zap.Duration("interval", np.interval),
-		zap.Uint64("initial_cursor", np.lastID))
+		zap.Uint64("cursor", np.lastID))
 
 	ticker := time.NewTicker(np.interval)
 	defer ticker.Stop()
--- a/pkg/server/notify_poller_test.go
+++ b/pkg/server/notify_poller_test.go
@@ -43,15 +43,16 @@ func TestNotifyPollerDiscoversCrossPodEvents(t *testing.T) {
 	defer bus.Unsubscribe(id)
 
 	np := newNotifyPoller(metaStore, buses, 50*time.Millisecond)
-	// Initialize cursor to 0 so we don't skip the row we're about to insert
-	// (run() would set it to MaxSSENotifyID which races with the insert).
-	np.lastID = 0
+	// Initialize cursor synchronously (as startNotifyInfrastructure does).
+	// At this point the outbox is empty, so lastID = 0.
+	np.initCursor(context.Background())
 	ctx, cancel := context.WithCancel(context.Background())
 	t.Cleanup(cancel)
 	go np.run(ctx)
 
-	// Simulate a cross-pod write: insert an outbox row (as a peer pod would).
-	// The poller should discover it within ~100ms and Publish to wake notify.
+	// Simulate a cross-pod write: insert an outbox row AFTER the poller is
+	// ready. The poller should discover it within ~100ms and Publish to wake
+	// notify. This proves no race between cursor init and the insert.
 	if err := metaStore.InsertSSENotify(context.Background(), "T", 1); err != nil {
 		t.Fatal(err)
 	}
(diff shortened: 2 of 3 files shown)


qiffang (approved):
## adversary-2 review — PR #649 (notifyPoller cursor race fix)

**Verdict: GREEN**

Directly addresses the cursor initialization race identified by adversary-1 in PR #647 re-review:

- `initCursor()` extracted from `run()` as a separate synchronous method
- `startNotifyInfrastructure` calls `poller.initCursor()` before launching the goroutine and before the server accepts live traffic
- `run()` no longer touches cursor initialization — reads pre-set `lastID`
- Tests updated to use `initCursor()` instead of manual `np.lastID = 0` overrides, proving the race is closed
- Comment references PR #647 adversary-1 re-review B1 for traceability

Minimal, focused fix. No new issues introduced.

qiffang (approved):
## dat9-dev1 — GREEN

**Verdict: GREEN — no blockers**

3 files changed, 37 additions, 16 deletions. Minimal, focused fix for the notifyPoller cursor initialization race identified by @adversary-1 in PR #647 re-review.

### What it does

Extracts cursor initialization (`MaxSSENotifyID`) from the async `run()` goroutine into a synchronous `initCursor()` method, called in `startNotifyInfrastructure` before launching the goroutine and before the server accepts traffic.

### Correctness verification

- **Race eliminated**: `initCursor()` runs synchronously before `go poller.run()`. No write can arrive between poller construction and cursor initialization because the server isn't accepting traffic yet.
- **Failure mode unchanged**: If `MaxSSENotifyID` fails, `lastID = 0` — reads historical rows, Publish is no-op for tenants without subscribers. Same behavior as before, just happens before traffic.
- **`run()` contract clear**: Comment states "cursor must already be initialized by initCursor before run is called".
- **Tests match production**: `np.initCursor(ctx)` replaces `np.lastID = 0` workaround. Batch drain test correctly calls `initCursor()` then overrides to 0 for test-specific drain behavior.

Head: `84d5bf9a34e0125a963013930c06bb4cf5697ba1`

qiffang (approved):
adversary-1 GREEN

No blockers found. This PR directly fixes the PR #647 poller cursor startup race by moving cursor initialization out of the background goroutine and calling it synchronously from startNotifyInfrastructure before the poller goroutine is launched and before the server can accept live traffic. Rows inserted after server construction can no longer be folded into the initial MaxSSENotifyID boundary and skipped by the fallback poller.

I also checked the updated notify poller tests: they now initialize the cursor before inserting live rows, so they no longer rely on goroutine scheduling. Batch-drain coverage remains intact.

Validation:
- Reviewed diff for pkg/server/notify_poller.go, pkg/server/server.go, pkg/server/notify_poller_test.go
- git diff --check e601432d4324cca7920c78855b122120ccc71198...HEAD passed
- GitHub checks are green: ci, drive9-server-local e2e smoke, CodeRabbit
- Local Go tests not run on this machine: go is not installed

Non-blocking nit: the newNotifyPoller comment still says the constructor initializes the cursor, while initialization now happens via initCursor(). This is documentation drift only, not a correctness issue.

srstack (commented):
Post-merge review (PR is already merged + qiffang-approved + CI green) — confirming the fix is correct, since it closes a real startup race.

### The race is real and the fix closes it
The bug: `run()` set `lastID = MaxSSENotifyID()` *inside* the background goroutine, asynchronously after `startNotifyInfrastructure` returned. A peer-pod write landing in the window between poller construction and that first `MaxSSENotifyID()` call would be adopted as the cursor max and never published — leaving an already-open SSE stream on the new pod stale until reconnect. That's exactly the rolling-update/restart scenario the outbox scheme exists to handle, so it mattered.

Moving cursor init into a synchronous `initCursor()` called before `go poller.run()` is the right fix. I verified the ordering that makes it actually safe: `startNotifyInfrastructure` (server.go:450) runs inside `NewWithConfig` (line 261), which returns before `ListenAndServe` (line 1190) is ever called. So `initCursor` completes before the server can accept any HTTP/SSE connection or any write can reach the poller — the cursor is provably set before live traffic. Race closed at the construction boundary, not merely narrowed.

### Correctness details checked
- `initCursor` failure path keeps `lastID = 0`, and the comment correctly notes that's non-harmful: worst case the poller reads some historical rows, but `Publish` is a no-op for tenants without subscribers. Degrades safe.
- `run()` no longer touches the cursor — clean separation, no double-init, no window where `run` could re-zero it.
- Test changes are sound: the two cross-pod tests now call `initCursor()` on an empty outbox (lastID=0) and insert *after*, which actually exercises the "write arrives after cursor is ready" path — i.e. the tests now prove the race is closed rather than papering over it with the old `np.lastID = 0` workaround. `TestNotifyPollerBatchDrain` keeps an explicit `initCursor` + `lastID = 0` override with a clear comment (init sets it to the max of pre-seeded rows, then resets to drain all) — correct and intentional.

### One non-blocking observation
`initCursor` is best-effort on DB error (logs + lastID=0). That's the right default, but it means a meta-DB blip exactly at startup silently shifts the poller from "skip history" to "scan from 0" — invisible beyond the warn log. Given the broader observability work in flight, a counter on `notify_poller_init_cursor_failed` (or reusing the module/error metrics) would make that startup-degradation visible rather than log-only. Minor, post-merge follow-up at most.

LGTM — correct fix for a genuine restart-window race, with tests that validate the boundary rather than mask it.

PR #653, opened 2026-06-30
Limit DB health probes to metadata store

srstack:
## Summary
- limit DB health probes to the long-lived metadata store only
- stop emitting user/user_schema DB availability and probe-duration series
- keep tenant_id on DB pool metrics while leaving DB operation metrics role-only
- avoid starting a no-op probe goroutine when the meta probe is disabled
- add server env parsing tests and disabled-probe behavior coverage

## Operational notes
- `drive9_db_up{role="user"}` and `drive9_db_up{role="user_schema"}` now disappear instead of emitting cold `1` series; dashboards and alerts should use tenant DB pool metrics for user/user_schema visibility.
- `DRIVE9_DB_HEALTH_PROBE_META_ENABLED=false` disables the meta probe entirely, so `drive9_db_up{role="meta"}` is absent and the documented meta DB unavailable alert will not fire in that mode.
- When the meta sql.DB pool is saturated, the health probe preserves the previous `drive9_db_up{role="meta"}` state instead of borrowing another connection. A database outage that begins while the pool is saturated is therefore surfaced through pool saturation/wait metrics until a probe can borrow a connection again.

## Tests
- `go test -count=1 ./pkg/metrics ./pkg/mysqlutil ./pkg/datastore ./pkg/tenant ./cmd/drive9-server`

## Summary by CodeRabbit

* **New Features**
  * Database health monitoring now focuses on the main metadata store by default, with a new setting to disable that check if needed.
  * Usage help now documents the available database health probe environment settings.

* **Bug Fixes**
  * Database health status reporting is now simpler and more consistent, using role-based visibility instead of tenant-specific details.
  * Database operation metrics no longer expose tenant-level labels, reducing noise in health and performance reporting.

### DIFF
--- a/pkg/metrics/db.go
+++ b/pkg/metrics/db.go
@@ -278,12 +278,20 @@ func (m *dbMetrics) probeOnceWithOptions(ctx context.Context, timeout time.Durat
 	prevProbe := m.probe
 	nextProbe := make(map[string]dbProbeState, len(results))
 	for role, agg := range results {
+		if !m.hasProbePoolLocked(role, opts) {
+			continue
+		}
 		up := agg.unreachable == 0
 		prev, existed := prevProbe[role]
 		unreachablePools := agg.unreachable
+		transitionErr := agg.firstErr
 		if existed && agg.saturated > 0 && agg.unreachable == 0 {
 			up = prev.up
 			unreachablePools = prev.unreachablePools
+		} else if !existed && agg.saturated > 0 && agg.unreachable == 0 {
+			up = false
+			unreachablePools = agg.total
+			transitionErr = fmt.Errorf("all %s DB probe pools are saturated", role)
 		}
 		nextProbe[role] = dbProbeState{
 			up:               up,
@@ -295,7 +303,7 @@ func (m *dbMetrics) probeOnceWithOptions(ctx context.Context, timeout time.Durat
 		// Fire on the first observed failure, and on every up<->down change
 		// thereafter. A first observation that is healthy is not a transition.
 		if onChange != nil && ((!existed && !up) || (existed && prev.up != up)) {
-			transitions = append(transitions, transition{role: role, up: up, err: agg.firstErr})
+			transitions = append(transitions, transition{role: role, up: up, err: transitionErr})
 		}
 	}
 	m.probe = nextProbe
@@ -419,6 +427,15 @@ func (p registeredDB) metricRole() string {
 func (m *dbMetrics) hasPoolLocked(role string) bool {
 	for _, pool := range m.dbs {
 		if pool.metricRole() == role {
+			return true
+		}
+	}
+	return false
+}
+
+func (m *dbMetrics) hasProbePoolLocked(role string, opts DBHealthProbeOptions) bool {
+	for _, pool := range m.dbs {
+		if pool.metricRole() == role && m.shouldProbePool(pool, opts) {
 			return true
 		}
 	}
--- a/pkg/metrics/db.go
+++ b/pkg/metrics/db.go
@@ -65,9 +65,9 @@ var dbProbeDuration = dbMeter.Float64Histogram("drive9_db_probe_duration_seconds
 // DefaultDBHealthProbeInterval and DefaultDBHealthProbeTimeout are the fallbacks
 // used by StartDBHealthProbe when callers pass non-positive values.
 const (
-	DefaultDBHealthProbeInterval = 15 * time.Second
-	DefaultDBHealthProbeTimeout  = 3 * time.Second
-	dbRoleMeta                   = "meta"
+	DefaultDBHealthProbeInterval        = 15 * time.Second
+	DefaultDBHealthProbeTimeout         = 3 * time.Second
+	dbRoleMeta                   string = "meta"
 )
 
 // DBHealthProbeOptions controls which registered DB pools are actively pinged.
@@ -199,6 +199,7 @@ func (m *dbMetrics) probeOnceWithOptions(ctx context.Context, timeout time.Durat
 	type roleAgg struct {
 		total       int
 		unreachable int
+		saturated   int
 		firstErr    error
 	}
 	results := map[string]*roleAgg{}
@@ -240,6 +241,9 @@ func (m *dbMetrics) probeOnceWithOptions(ctx context.Context, timeout time.Durat
 			// this state, even if the database is also unreachable underneath.
 			if isDBPoolSaturated(db) {
 				result = "pool_saturated"
+				aggMu.Lock()
+				results[role].saturated++
+				aggMu.Unlock()
 				observeDBProbeDuration(start, role, result)
 				return
 			}
@@ -276,9 +280,14 @@ func (m *dbMetrics) probeOnceWithOptions(ctx context.Context, timeout time.Durat
 	for role, agg := range results {
 		up := agg.unreachable == 0
 		prev, existed := prevProbe[role]
+		unreachablePools := agg.unreachable
+		if existed && agg.saturated > 0 && agg.unreachable == 0 {
+			up = prev.up
+			unreachablePools = prev.unreachablePools
+		}
 		nextProbe[role] = dbProbeState{
 			up:               up,
-			unreachablePools: agg.unreachable,
+			unreachablePools: unreachablePools,
 			totalPools:       agg.total,
 			lastProbe:        now,
 			known:            true,
--- a/pkg/metrics/db.go
+++ b/pkg/metrics/db.go
@@ -67,6 +67,7 @@ var dbProbeDuration = dbMeter.Float64Histogram("drive9_db_probe_duration_seconds
 const (
 	DefaultDBHealthProbeInterval = 15 * time.Second
 	DefaultDBHealthProbeTimeout  = 3 * time.Second
+	dbRoleMeta                   = "meta"
 )
 
 // DBHealthProbeOptions controls which registered DB pools are actively pinged.
@@ -145,6 +146,12 @@ func StartDBHealthProbeWithOptions(ctx context.Context, interval, timeout time.D
 	}
 	if timeout <= 0 {
 		timeout = DefaultDBHealthProbeTimeout
+	}
+	if !opts.hasEnabledProbes() {
+		globalDB.mu.Lock()
+		globalDB.probe = map[string]dbProbeState{}
+		globalDB.mu.Unlock()
+		return
 	}
 
 	globalDB.mu.Lock()
@@ -410,7 +417,11 @@ func (m *dbMetrics) hasPoolLocked(role string) bool {
 }
 
 func (m *dbMetrics) shouldProbePool(pool registeredDB, opts DBHealthProbeOptions) bool {
-	return pool.metricRole() == "meta" && opts.ProbeMeta
+	return pool.metricRole() == dbRoleMeta && opts.ProbeMeta
+}
+
+func (opts DBHealthProbeOptions) hasEnabledProbes() bool {
+	return opts.ProbeMeta
 }
 
 func isDBPoolSaturated(db *sql.DB) bool {
--- a/pkg/metrics/db.go
+++ b/pkg/metrics/db.go
@@ -10,13 +10,6 @@ import (
 	"time"
 )
 
-// DBPoolInfo identifies a registered database pool for metrics and health
-// transition logging.
-type DBPoolInfo struct {
-	Role     string
-	TenantID string
-}
-
 // dbProbeState is the cached result of the most recent health probe for a role.
 type dbProbeState struct {
 	up               bool
@@ -39,7 +32,7 @@ type dbMetricKey struct {
 type dbMetrics struct {
 	mu      sync.RWMutex
 	dbs     map[*sql.DB]registeredDB
-	probe   map[dbMetricKey]dbProbeState
+	probe   map[string]dbProbeState
 	started bool
 }
 
@@ -61,7 +54,7 @@ var dbProbeDurationBounds = []float64{0.001, 0.005, 0.01, 0.025, 0.05, 0.1, 0.25
 
 var globalDB = &dbMetrics{
 	dbs:   map[*sql.DB]registeredDB{},
-	probe: map[dbMetricKey]dbProbeState{},
+	probe: map[string]dbProbeState{},
 }
 
 var dbMeter = globalMeterProvider.Meter("db")
@@ -75,6 +68,13 @@ const (
 	DefaultDBHealthProbeInterval = 15 * time.Second
 	DefaultDBHealthProbeTimeout  = 3 * time.Second
 )
+
+// DBHealthProbeOptions controls which registered DB pools are actively pinged.
+// drive9-server only probes the long-lived control-plane DB; tenant user DB
+// pools are short-lived workload pools and are observed through normal traffic.
+type DBHealthProbeOptions struct {
+	ProbeMeta bool
+}
 
 func RecordDBOperation(role, operation, result string, d time.Duration) {
 	RegisterModule("db")
@@ -106,19 +106,25 @@ func UnregisterDB(db *sql.DB) {
 		return
 	}
 	globalDB.mu.Lock()
+	pool, ok := globalDB.dbs[db]
 	delete(globalDB.dbs, db)
+	if ok {
+		role := pool.metricRole()
+		if !globalDB.hasPoolLocked(role) {
+			delete(globalDB.probe, role)
+		}
+	}
 	globalDB.mu.Unlock()
 }
 
 // StartDBHealthProbe launches a background goroutine that periodically pings every
-// registered *sql.DB and publishes availability as the drive9_db_up gauge. It is
-// the active dependency-health probe for the metadata store ("meta") and tenant
-// data stores ("user"): pool/operation metrics only move while there is live
-// traffic, so an idle-but-unreachable database is otherwise invisible. User DB
-// pool series include tenant_id when the pool was registered with one.
+// registered long-lived dependency *sql.DB and publishes availability as the
+// drive9_db_up gauge. In drive9-server this means the metadata store ("meta").
+// Tenant user/user_schema pools are short-lived workload pools, so probing them
+// would only keep connections warm and compete with business traffic.
 //
 // The probe never blocks the /metrics scrape path — scrapes read the cached probe
-// state. onChange (optional) is invoked outside the lock on a pool key's first
+// state. onChange (optional) is invoked outside the lock on a role's first
 // observed failure and on every subsequent up<->down transition, so callers can
 // emit a log without coupling this package to a logger. The goroutine stops when
 // ctx is done; repeat calls after the first are no-ops.
@@ -126,7 +132,14 @@ func UnregisterDB(db *sql.DB) {
 // Example alert (critical — control-plane DB unreachable):
 //
 //	max(drive9_db_up{role="meta"}) == 0
-func StartDBHealthProbe(ctx context.Context, interval, timeout time.Duration, onChange func(info DBPoolInfo, up bool, err error)) {
+func StartDBHealthProbe(ctx context.Context, interval, timeout time.Duration, onChange func(role string, up bool, err error)) {
+	StartDBHealthProbeWithOptions(ctx, interval, timeout, DBHealthProbeOptions{
+		ProbeMeta: true,
+	}, onChange)
+}
+
+// StartDBHealthProbeWithOptions is StartDBHealthProbe with role filtering.
+func StartDBHealthProbeWithOptions(ctx context.Context, interval, timeout time.Duration, opts DBHealthProbeOptions, onChange func(role string, up bool, err error)) {
 	if interval <= 0 {
 		interval = DefaultDBHealthProbeInterval
 	}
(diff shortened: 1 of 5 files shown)


srstack (commented):
**Review** — clean, well-scoped follow-up to #650. Build/vet/tests pass locally (`pkg/metrics`, `pkg/mysqlutil`, `cmd/drive9-server`). This directly resolves the metric-cardinality concern I raised on #650: probe + DB-operation series are back to role-only, while pool gauges keep `tenant_id`. Good split.

### What it does (for reviewers)
- Health probe now pings only the long-lived `meta` store; `user`/`user_schema` pools are no longer probed (`shouldProbePool` returns `role=="meta" && opts.ProbeMeta`).
- `drive9_db_up` / `drive9_db_unreachable_pools` / `drive9_db_probe_duration_seconds` are now role-keyed (no `tenant_id`); DB-operation metrics likewise drop `tenant_id` (`instrumentedConn`/`Stmt`/`Tx` no longer carry it).
- Pool stats (`pool_registered`, `pool_connections`, `wait_*`, `closes_total`) **keep** `tenant_id` — bounded-cardinality gauges, so that's fine.
- New `DRIVE9_DB_HEALTH_PROBE_META_ENABLED` toggle (default true), with env-parsing tests.

### One behavioral change worth calling out explicitly
`db_up` semantics changed from "registered roles default to 1" to "only probed roles emit a series". Previously every registered role got a cold `db_up=1`; now `user`/`user_schema` emit **no** `db_up` series at all. That's the intent, but:
- Any existing alert/dashboard referencing `drive9_db_up{role="user"}` or `{role="user_schema"}` will now match an absent series. Worth a heads-up in the PR description / release notes so dashboards get updated.
- The package-level `StartDBHealthProbe` doc still says "publishes availability ... metadata store ('meta') and tenant data stores" in one spot but the new text says meta-only — double-check the doc comment is fully consistent (the diff updates most of it; just confirm no stale "user" mention remains).

### Minor (non-blocking)
- With `DRIVE9_DB_HEALTH_PROBE_META_ENABLED=false`, `shouldProbePool` filters out every pool, so each tick runs an empty `probeOnce` and clears all probe state — i.e. the goroutine still spins on the ticker doing nothing. Harmless, but if "disabled" is a real operating mode, consider not starting the goroutine at all when `!ProbeMeta`. Also note that in that mode `drive9_db_up{role="meta"}` disappears, which would make the documented critical alert `max(drive9_db_up{role="meta"}) == 0` silently stop evaluating rather than fire — flag that tradeoff to operators.
- `shouldProbePool` hardcodes `"meta"`. That's correct today, but the literal is the single source of truth tying the option to a role — a `RoleMeta`-style const reference (mysqlutil already defines `RoleMeta`) would prevent drift. Tiny nit.
- `hasPoolLocked` does a linear scan of `m.dbs` on every `UnregisterDB`. Fine at current pool counts; just noting it's O(n) per close if pool churn ever grows.

LGTM once CI is green and the `db_up{role="user"}` series-removal is noted for dashboard owners. Nice cardinality cleanup.

mornyx (commented):
Reviewed PR #653 (limit DB health probes to the metadata store). This is a clean, well-scoped change. Build is clean, `make lint` reports 0 issues, and all packages in the test plan pass — including `go test -race ./pkg/metrics` which exercises the concurrent probe path.

## What's good

- **Dead code removal.** `RecordDBOperation(role, operation, result, d)` never took a `tenant_id`, and the `instrumentedConnector`/`instrumentedConn`/`instrumentedStmt`/`instrumentedTx` `tenantID` fields were threaded through only to be discarded (`observeDBOperation(ctx, role, _, ...)`). Dropping them entirely is the right call — no behavior change, less surface area.

- **Clean role/tenant split.** `dbMetricKey`/`dbLabels`/`metricKey()` stay for the pool-level series (`drive9_db_pool_registered`, `drive9_db_pool_connections`, etc.) which still carry `tenant_id` for tenant pools. Only the availability/probe series (`drive9_db_up`, `drive9_db_unreachable_pools`, `drive9_db_probe_duration_seconds`) collapse to role-only. This matches the operational notes — dashboards tracking tenant DB health use the pool metrics, not the probe gauge. Verified: meta and user_schema register with `tenantID=""` so their pool series correctly omit `tenant_id` via the `key.tenantID != "unknown"` guard.

- **Disabled-probe path.** `StartDBHealthProbeWithOptions` with `!hasEnabledProbes()` clears stale `probe` state and returns without flipping `started`, so a later enabled call can still launch. `TestStartDBHealthProbeDisabledDoesNotStart` covers this.

- **Saturated-state carry-forward** (commit `9aaa6e7`). When a role is fully saturated (`saturated > 0 && unreachable == 0`), the previous `up`/`unreachablePools` are preserved instead of falsely marking the DB down. `TestDBHealthProbeKeepsPreviousDownStateWhenMetaPoolSaturated` validates the down→saturated transition keeps `drive9_db_up{role="meta"} 0`.

- **Env doc backfill.** `DRIVE9_DB_HEALTH_PROBE_INTERVAL_SECONDS`/`TIMEOUT_SECONDS` were already wired but undocumented; this adds them plus the new `META_ENABLED` flag.

## Minor / non-blocking

**A. Saturated meta pool masks a real outage (documented, but worth surfacing).** When the meta pool is saturated, `isDBPoolSaturated` short-circuits before `PingContext`, so the probe never learns whether the DB is actually reachable. If the previous state was `up`, it stays `up`. This is the intended tradeoff (comment at db.go ~238-241 says so), but operationally it means the critical alert `max(drive9_db_up{role="meta"}) == 0` will **not** fire for a meta DB that goes down while its pool is saturated — you'd only see it via `drive9_db_pool_connections{...,state="in_use"}` / wait metrics. The operational notes mention the disabled-probe case but not this saturation-blindness; consider adding a one-liner to the notes so operators know the `== 0` alert has a blind spot under pool saturation.

**B. No test for the mixed saturated + unreachable case.** The carry-forward branch only triggers when `agg.saturated > 0 && agg.unreachable == 0`. If a role has 2+ pools and one is saturated while another is unreachable, `agg.unreachable > 0` so the branch is skipped, `up = false`, `unreachablePools = agg.unreachable`. That looks correct (at least one reachable pool reported failure), but there's no test pinning it. A small unit test with a 2-pool role (one saturated, one unhealthy) would lock the behavior in case the aggregation logic is refactored later.

**C. Stale-probe window on `UnregisterDB`.** `probeOnce` snapshots `dbs` under `RLock`, releases it, runs the pings, then writes `m.probe` under `Lock`. If `UnregisterDB` removes the last pool for a role (and deletes `probe[role]`) between snapshot and write, the in-flight `probeOnce` re-inserts `probe[role]` for one more cycle. In practice this only matters for the meta role (the only probed role), and meta unregister essentially only happens at shutdown — so it's a non-issue here. Mentioning only because the new `hasPoolLocked`-based cleanup is otherwise careful; if `probeOnce` re-checked `hasPoolLocked` before writing each `probe[role]`, it'd be fully deterministic, but that's over-engineering for the current usage.

**D. `dbRoleMeta` typing nit.** `dbRoleMeta string = "meta"` is typed while the two `Default*` consts in the same block are untyped `time.Duration`. It's gofmt-clean and harmless; just noting the block mixes typed and untyped consts. No action needed.

Net: looks good. The items above are all non-blocking (docs/coverage/nits). I'd only ask for (A) — a one-line operational note about the saturation blind spot — and (B) if you want to be thorough, but neither blocks merge.

pkg/metrics/db.go
```diff
@@ -213,7 +241,10 @@ func (m *dbMetrics) probeOnce(ctx context.Context, timeout time.Duration, onChan
 			// this state, even if the database is also unreachable underneath.
 			if isDBPoolSaturated(db) {
```
mornyx, line 242: Documented behavior, but operationally significant: when the meta pool is saturated, the probe short-circuits here and `PingContext` never runs. Combined with the carry-forward at line ~284 (`existed && agg.saturated > 0 && agg.unreachable == 0` → keep `prev.up`), a meta DB that goes down *while* its pool is saturated stays `drive9_db_up{role="meta"} 1` — so the critical alert `max(drive9_db_up{role="meta"}) == 0` won't fire. Operators would only see it via pool wait/saturation metrics. The operational notes cover the disabled-probe blind spot but not this one; a one-liner in the PR notes ("meta outage under pool saturation is surfaced via pool metrics, not drive9_db_up") would close the gap.

pkg/metrics/db.go
```diff
@@ -245,29 +276,33 @@ func (m *dbMetrics) probeOnce(ctx context.Context, timeout time.Duration, onChan
...
-	for key, agg := range results {
-		pool := poolsByKey[key]
+	nextProbe := make(map[string]dbProbeState, len(results))
+	for role, agg := range results {
 		up := agg.unreachable == 0
-		prev, existed := prevProbe[key]
-		nextProbe[key] = dbProbeState{
+		prev, existed := prevProbe[role]
+		unreachablePools := agg.unreachable
+		if existed && agg.saturated > 0 && agg.unreachable == 0 {
```
mornyx, line 288: The carry-forward only triggers when `agg.saturated > 0 && agg.unreachable == 0`. For a role with 2+ pools where one is saturated and another is genuinely unreachable, `agg.unreachable > 0` skips this branch and sets `up = false` / `unreachablePools = agg.unreachable` — which is correct (a reachable pool reported a failure). There's currently no test pinning this mixed case; a small 2-pool unit test (one saturated, one unhealthy) would lock the behavior if the aggregation is refactored later. Non-blocking.

pkg/metrics/db.go
```diff
@@ -106,33 +107,52 @@ func UnregisterDB(db *sql.DB) {
 		return
 	}
 	globalDB.mu.Lock()
+	pool, ok := globalDB.dbs[db]
 	delete(globalDB.dbs, db)
+	if ok {
+		role := pool.metricRole()
+		if !globalDB.hasPoolLocked(role) {
```
mornyx, line 114: Minor (non-issue for this PR, flagging for completeness): `probeOnce` snapshots `dbs` under `RLock`, releases it, runs pings, then writes `m.probe` under `Lock`. If `UnregisterDB` deletes the last pool for a role (and clears `probe[role]` here via `hasPoolLocked`) between that snapshot and the write, the in-flight `probeOnce` re-inserts `probe[role]` for one cycle. In practice the only probed role is `meta`, and meta unregister happens at shutdown, so this is harmless here. The `hasPoolLocked`-based cleanup is otherwise correct and well-tested by `TestDBHealthProbeDropsUnregisteredState`.

qiffang (commented):
## dat9-dev1 consolidated review — PR #653

**Verdict: APPROVE** (clean, well-scoped, correct)

### What this PR does
- Limits DB health probes (`drive9_db_up`) to the long-lived metadata ("meta") store only
- Removes user/user_schema probe series (tenant pools are short-lived workload pools — probing them keeps connections warm and competes with business traffic)
- Keeps `tenant_id` on DB pool metrics while simplifying DB operation metrics to role-only
- Adds `DRIVE9_DB_HEALTH_PROBE_META_ENABLED` env var to disable meta probe
- Avoids starting a no-op goroutine when all probes are disabled

### Correctness review

1. **Probe filtering (`shouldProbePool`)** — Only `meta` role passes through. Clean single predicate. ✅
2. **Saturation carry-forward** — When all pools are saturated and none are unreachable, the previous `up`/`unreachablePools` state is preserved. This avoids false-healthy when a DB outage starts during pool saturation. New test `TestDBHealthProbeKeepsPreviousDownStateWhenMetaPoolSaturated` covers this. ✅
3. **Mixed saturated + unreachable** — `agg.unreachable > 0` takes priority, correctly marking the role as down. New test `TestDBHealthProbeMarksRoleDownWhenOneMetaPoolSaturatedAndAnotherUnreachable` covers this. ✅
4. **No-op goroutine prevention** — `hasEnabledProbes()` returns false → clears probe state and returns without starting the ticker goroutine. ✅
5. **`UnregisterDB` probe cleanup** — Correctly clears `probe[role]` via `hasPoolLocked` when the last pool for a role is unregistered. ✅
6. **Metric label simplification** — `drive9_db_up` and `drive9_db_unreachable_pools` now emit `role=` only (no `tenant_id`). `drive9_db_probe_duration_seconds` drops `tenant_id`. Pool metrics (`drive9_db_pool_*`) retain `tenant_id`. Consistent with the design intent. ✅

### Existing review comments assessment

**@mornyx's 3 points** — all valid observations, none are blockers:
- Saturation carry-forward semantics: correctly documented in operational notes and tested
- Multi-pool race window (UnregisterDB between snapshot and write): narrow, pre-existing, non-blocking
- `observeDBOperation` dropping unused `ctx` param: clean removal

**CodeRabbit's 2 actionable items**:
- `globalDB.probe` bare read in test (`db_test.go:246`): this is test-only code running single-threaded after `probeOnce` completes; not a real race. Non-blocking.
- Saturation carry-forward logic: addressed by the implementation and tests.

### Minor notes (non-blocking)

- The `dbMetricKey` struct is now only used for pool-level metrics (`metricKey()` in `writeDBPrometheus`). Consider renaming or adding a comment clarifying it's pool-scoped, not probe-scoped. Not blocking.

### Tests
All relevant test packages pass: `pkg/metrics`, `pkg/mysqlutil`, `cmd/drive9-server`.

🤖 Reviewed by dat9-dev1 (AI agent)

qiffang (changes requested):
## adversary-2 Review — REVISE on `c44b4d2`

### Blocker #1: `UnregisterDB` + `probeOnceWithOptions` race on `globalDB.probe`

`UnregisterDB` holds `mu.Lock()`, checks `hasPoolLocked(role)`, and deletes `globalDB.probe[role]`. But `probeOnceWithOptions` builds `nextProbe` from a snapshot taken under `RLock` (step 1), then writes `m.probe = nextProbe` under `Lock` (step 3).

If between step 1 and step 3, `UnregisterDB` removes the last pool for a role and cleans `probe[role]`, step 3 overwrites it back — because `nextProbe` was built from the stale snapshot. Result: stale probe state for an unregistered role is resurrected.

This is not theoretical — tenant pool lifecycle is short, register/unregister is frequent.

### Blocker #2: saturation carry-forward on first probe is undefined

```go
if existed && agg.saturated > 0 && agg.unreachable == 0 {
    up = prev.up
    unreachablePools = prev.unreachablePools
}
```

When `!existed` (first probe) and all pools are saturated (`unreachable == 0, saturated > 0`), `up` stays `true` (because `unreachable == 0`), `unreachablePools = 0`. This means the first probe reports `drive9_db_up=1` even though no pool was actually reachable (all saturated).

This contradicts the PR body: "preserves the previous `drive9_db_up{role="meta"}` state" — there is no previous state on first probe, so it defaults to optimistic. If meta DB is saturated at startup (MaxOpenConns=1 and startup ping exhausted the pool), a real outage would be missed.

### Non-blocking #3: `observeDBOperation` removed `ctx` parameter

The `ctx` parameter was previously unused (tenantID was `_`), so removing it is fine now. But confirm no other callers depend on ctx propagation for tracing.

### Non-blocking #4: `dbRoleMeta` constant not shared with registration sites

`RegisterDB("meta", db)` in `main.go` uses a string literal. `shouldProbePool` uses `dbRoleMeta` constant. If registration ever uses a different string, probe silently skips. Suggest using the constant at registration sites too.

### Missing tests

- No `UnregisterDB` + concurrent `probeOnce` interaction test (blocker #1)
- No first-probe-all-saturated scenario test (blocker #2)

qiffang (approved):
## Unified Review — PR #653

### Overall: **APPROVE** (clean, well-scoped, tests pass)

Reviewed the full diff (5 files, +322/-155). Tests pass locally (`go test ./pkg/metrics/ ./pkg/mysqlutil/ ./cmd/drive9-server/`). The change is directionally correct: meta-only probe removes unnecessary tenant pool probing that competes with workload traffic.

### What's good

1. **Clean separation**: `shouldProbePool` + `DBHealthProbeOptions` makes role filtering explicit
2. **Saturation carry-forward** (commit `9aaa6e7d`): preserves previous up/down state when pool is saturated instead of defaulting to `up=true`
3. **Mixed saturated+unreachable test** (commit `c44b4d2d`): covers the tricky case where one pool is saturated and another is unreachable — correctly marks role down
4. **`UnregisterDB` cleanup**: removes stale probe state when last pool for a role is unregistered
5. **No cold series**: unprobed roles no longer emit `drive9_db_up=1` / `drive9_db_unreachable_pools=0`
6. **`observeDBOperation` simplification**: removes unused `ctx` and `tenantID` params — good cleanup
7. Pool metrics (`drive9_db_pool_registered`, `drive9_db_pool_connections`, etc.) still emit `tenant_id` labels — correct, these are scrape-time aggregations not probe-dependent

### Edge cases noted (non-blocking)

**1. `UnregisterDB` + `probeOnceWithOptions` race window**

If `UnregisterDB` deletes a role's last pool + probe state between `probeOnceWithOptions`'s RLock snapshot (step 1) and Lock write (step 3), the probe writes back a stale `nextProbe` entry for that role. This is a pre-existing pattern (the old code had the same RLock→ping→Lock structure), but the new `UnregisterDB` cleanup makes it observable. In practice this only affects the meta role, which is long-lived and never unregistered during normal operation. Tenant pools are excluded by `shouldProbePool`. **Suggest a follow-up to reconcile `nextProbe` against current registrations inside the Lock section.**

**2. First-probe all-saturated defaults to `up=true`**

When `!existed && saturated > 0 && unreachable == 0`, the carry-forward guard (`existed &&`) doesn't apply, so `up = true`. This means if the meta pool is saturated on the very first probe, `drive9_db_up{role="meta"}=1` even though no ping succeeded. In practice this can't happen because the startup `PingContext` in `main.go` returns the connection before `StartDBHealthProbeWithOptions` launches. **Document the invariant (startup ping must complete before probe starts) or add a one-line comment.**

**3. `dbRoleMeta` constant not used at registration sites**

`shouldProbePool` compares against `dbRoleMeta = "meta"`, but `RegisterDB("meta", db)` in `main.go` uses a string literal. If registration used a different string, the probe would silently skip the pool. Low risk since there's only one registration site, but using the constant at the call site would be belt-and-suspenders.

### Missing test coverage (nice-to-have, not blocking)

- Concurrent `UnregisterDB` + `probeOnce` interaction (edge case #1)
- First-probe with all pools saturated (edge case #2)

---

LGTM. The edge cases are real but don't affect production correctness given the meta-only scope. Suggest addressing them in follow-up if more roles are added to the probe.

qiffang (approved):
## adversary-2 — Updated to APPROVE on `c44b4d2`

Previous blockers downgraded to non-blocking after evaluating runtime constraints:

1. **Probe state resurrection race**: `shouldProbePool` only probes meta, and meta is never unregistered at runtime. The race window in `UnregisterDB` cannot fire for the probed role under current usage.

2. **First-probe-all-saturated**: `main.go` calls `PingContext` before `StartDBHealthProbeWithOptions`, so the meta pool connection is returned before the first probe. First-probe saturation is unreachable under normal startup.

Both edge cases are real but not reachable under current call constraints. Recommend follow-up comments/defensive checks but not blocking merge.

pkg/metrics/db.go
srstack: Acknowledged. This stays intentionally non-blocking for this PR: after the meta-only probe change, tenant pools are not probed, and the only probed role is the long-lived meta pool which is not unregistered during normal runtime. The one-cycle stale probe window is therefore not reachable for the active probed role here; leaving a broader defensive reconciliation as follow-up if more roles are ever probed.

srstack:
Addressed the two DB probe edge cases in `a589fc9`:

- Reconcile probe results against currently registered probed pools before writing `globalDB.probe`, so an in-flight probe cannot resurrect a role after its last probed pool is unregistered.
- Treat a first probe where every pool is saturated as down instead of optimistically healthy, with a saturation error on the first down transition.

Added regression tests for both cases.

srstack (commented):
**Re-review @ `a589fc97`** — the 4 new commits address both of my earlier minors and tighten the saturation edge cases. Looks good; build/vet/tests pass (`-race`, `count=1`).

### My earlier feedback — addressed
- **Disabled mode no longer spins.** `StartDBHealthProbeWithOptions` now early-returns (and clears probe state) when `!opts.hasEnabledProbes()`, instead of starting a goroutine that ticks over an empty pool set. `TestStartDBHealthProbeDisabledDoesNotStart` covers it. 👍
- **`"meta"` literal** extracted to the `dbRoleMeta` const, used by both `shouldProbePool` and the comparison. Drift risk gone.

### New saturation semantics — reviewed, correct
The three-way handling in `probeOnceWithOptions` is the right call:
- existing `up` state + saturated + 0 unreachable → **preserve** prior state (don't flip on transient saturation),
- **first** observation + all saturated → mark **down** with an explicit "all pools saturated" error (we've never verified reachability, so defaulting up would mask a real outage),
- saturated alongside any unreachable → still down.

This resolves the tension from the #650 saturation discussion without either masking a genuine outage or false-flapping on load. The mixed saturated+unreachable test (`c44b4d2d`) and the down-state-preservation test pin the behavior well.

### Nice addition: in-flight unregister race
`TestDBHealthProbeDoesNotResurrectUnregisteredRoleAfterInFlightProbe` exercises the real concern — a pool unregistered while its ping is in flight. The `hasProbePoolLocked(role, opts)` re-check under the final write lock is what guarantees a role removed mid-cycle isn't resurrected into `nextProbe`. Good that this is now both fixed and tested.

### One non-blocking note on test isolation (not a code bug)
Running `go test ./pkg/metrics -count=2` fails (`TestRecordDBOperationOmitsTenantID`, `TestRecordTenantOperationCountDoesNotRecordDuration`) because these assert absolute counter values against the process-global `globalDB`, which accumulates across iterations. `count=1` (CI's mode) is clean, and one of the two isn't even touched by this PR — so this is pre-existing global-singleton test coupling, not a regression here. Flagging only because it makes these tests non-idempotent; a future cleanup could reset/recreate `globalDB` per test.

LGTM pending green CI. The probe-scope reduction + saturation handling are solid, and my prior comments are all resolved.

PR #652, opened 2026-06-30
feat: write-back disk protection with free-ratio and byte-quota guards

qiffang:
## Summary

Closes #651

- **Default-on `--write-cache-free-ratio=0.10`**: Statfs check before shadow write, returns `ENOSPC` if free space would drop below threshold after write
- **Opt-in `--write-cache-size-mb=0`**: byte quota disabled by default; only active when explicitly set
- **Guards in all 3 write paths**: `stageShadowLocked`, `snapshotWriteBackLocked`, ShadowSpill write handler — atomic fail before writing, never half-written
- **Atomic pending byte tracking** with `RecoverPendingBytes()` crash-recovery via directory scan on startup
- **CAS-throttled Statfs** (5s interval) on hot paths to avoid syscall overhead
- **Rejection log**: `cache_dir/free_bytes/free_ratio/required_bytes/reason` — no user file path/token

## Test plan

- [x] `TestCheckWriteBackQuotaByteQuotaAllowed` — normal write under quota
- [x] `TestCheckWriteBackQuotaByteQuotaExceeded` — write exceeding byte quota → ENOSPC
- [x] `TestCheckWriteBackQuotaByteQuotaDisabled` — quota=0 allows any write
- [x] `TestCheckWriteBackQuotaFreeRatioAllowed` — small write passes 10% ratio
- [x] `TestCheckWriteBackQuotaFreeRatioExceeded` — 99.99% ratio rejects large writes
- [x] `TestCheckWriteBackQuotaFreeRatioDisabled` — ratio=0 allows any write
- [x] `TestPendingBytesAddSubRecover` — Add/Sub/PendingBytes arithmetic
- [x] `TestRecoverPendingBytesFromDisk` — RecoverPendingBytes scans shadow dir
- [x] `TestByteQuotaRecoveryAfterCommitCleanup` — quota exceeded → sub bytes → write succeeds again
- [x] `TestCheckWriteBackQuotaThrottled` — throttled check returns consistent results

🤖 Generated with [Claude Code](https://claude.com/claude-code)

## Summary by CodeRabbit

* **New Features**
  * Added `drive9 mount` flags to tune write-back cache protection: `--write-cache-free-ratio` and `--write-cache-size-mb`.
  * Write-back cache now enforces configurable quota-based space checks with improved pending-write tracking and recovery on restart.
* **Bug Fixes**
[...]

PR #654, opened 2026-06-30
Avoid waking cold tenant DBs in semantic worker

srstack:
## Summary
- Limit multi-tenant Semantic Worker periodic polling/observation to warm cached tenant backends with open DB connections.
- Keep explicit kick processing able to acquire the tenant backend for fresh foreground-created work.
- Derive the warm tenant poll cadence from the user DB connection lifetime and remove the extra public warm-poll env knob.
- Add cached-acquire/warm-tenant pool helpers and focused worker tests for warm vs cold tenants.

## Notes
- The related runbooks alert wording was committed separately in the runbooks repo as `e8cf0e17 Clarify semantic worker tenant alert scope`; no runbooks PR was created.

## Tests
- `go test ./pkg/server ./pkg/tenant ./pkg/mysqlutil ./cmd/drive9-server ./cmd/drive9-server-local`

## Summary by CodeRabbit

* **New Features**
  * Background semantic processing now prioritizes warmed, cached tenant backends and surfaces warm-tenant poll timing in runtime status logs.

* **Bug Fixes**
  * Multi-tenant polling/recovery scheduling now respects a warm-tenant-aware poll cadence, improving scan and retry behavior.
  * Cold cached tenants are no longer treated as ready until warmed; expired-claim recovery for cold tenants is improved.
  * Kick wake-up handling was hardened to better track queued vs pending work and preserve pending state on acquisition failures.
  * Increased default MySQL pool lifetime and max open connections for stability.

### DIFF
--- a/cmd/drive9-server-local/main_test.go
+++ b/cmd/drive9-server-local/main_test.go
@@ -248,6 +248,7 @@ func TestBuildSemanticWorkerConfigFromEnvReadsWorkerOptionsWithoutEmbedder(t *te
 	for _, key := range keys {
 		_ = os.Unsetenv(key)
 	}
+
 	if err := os.Setenv("DRIVE9_SEMANTIC_WORKERS", "8"); err != nil {
 		t.Fatal(err)
 	}
--- a/cmd/drive9-server/main_test.go
+++ b/cmd/drive9-server/main_test.go
@@ -183,6 +183,7 @@ func TestBuildSemanticWorkerConfigFromEnvReadsWorkerOptionsWithoutEmbedder(t *te
 	restore := snapshotEnv(t, keys)
 	t.Cleanup(func() { restoreEnv(t, restore) })
 	unsetEnv(t, keys)
+
 	setEnv(t, "DRIVE9_SEMANTIC_WORKERS", "8")
 	setEnv(t, "DRIVE9_SEMANTIC_TENANT_LIMIT", "512")
 
--- a/pkg/mysqlutil/pool.go
+++ b/pkg/mysqlutil/pool.go
@@ -13,9 +13,9 @@ const (
 	defaultMetaConnMaxIdleTime       = 45 * time.Second
 	defaultMetaMaxOpenConns          = 100
 	defaultMetaMaxIdleConns          = 20
-	defaultUserConnMaxLifetime       = 30 * time.Second
+	defaultUserConnMaxLifetime       = 45 * time.Second
 	defaultUserConnMaxIdleTime       = 30 * time.Second
-	defaultUserMaxOpenConns          = 4
+	defaultUserMaxOpenConns          = 6
 	defaultUserMaxIdleConns          = 1
 	defaultUserSchemaConnMaxLifetime = 1 * time.Minute
 	defaultUserSchemaConnMaxIdleTime = 10 * time.Second
@@ -46,6 +46,12 @@ func ApplyPoolDefaults(db *sql.DB, role string) {
 	if v := poolEnvInt(role, "MAX_IDLE_CONNS", maxIdle); v >= 0 {
 		db.SetMaxIdleConns(v)
 	}
+}
+
+// DefaultUserConnMaxLifetime returns the default maximum lifetime for tenant
+// user DB connections.
+func DefaultUserConnMaxLifetime() time.Duration {
+	return defaultUserConnMaxLifetime
 }
 
 func defaultPoolLifetime(role string) (time.Duration, time.Duration) {
--- a/pkg/server/semantic_worker.go
+++ b/pkg/server/semantic_worker.go
@@ -19,6 +19,7 @@ import (
 	"github.com/mem9-ai/drive9/pkg/logger"
 	"github.com/mem9-ai/drive9/pkg/meta"
 	"github.com/mem9-ai/drive9/pkg/metrics"
+	"github.com/mem9-ai/drive9/pkg/mysqlutil"
 	"github.com/mem9-ai/drive9/pkg/semantic"
 	"github.com/mem9-ai/drive9/pkg/tenant"
 )
@@ -30,12 +31,13 @@ const (
 	defaultSemanticRecoverInterval      = 5 * time.Second
 	defaultSemanticRetryBaseDelay       = 200 * time.Millisecond
 	defaultSemanticRetryMaxDelay        = 30 * time.Second
+	defaultSemanticWarmPollMargin       = 15 * time.Second
 	defaultSemanticTenantScanLimit      = 128
 	defaultSemanticPerTenantConcurrency = 1
 	semanticLocalTenantID               = "local"
 	// semanticKickQueueCapacity bounds buffered upload-commit kicks; overflow
-	// kicks are dropped because the periodic tenant scan remains the durable
-	// fallback for claiming queued tasks.
+	// kicks are dropped. Periodic polling only checks cached tenant backends, so
+	// a later tenant request or kick is the fallback for dormant tenants.
 	semanticKickQueueCapacity = 256
 	// semanticKickDrainLimit caps tasks drained per kick so one busy tenant
 	// cannot monopolize a worker; the tenant is re-kicked to continue after
@@ -60,6 +62,11 @@ type SemanticWorkerOptions struct {
 	Workers int
 	// PollInterval is the idle wait between claim attempts.
 	PollInterval time.Duration
+	// WarmTenantPollInterval limits best-effort multi-tenant polling over
+	// cached tenants that still have open DB connections. It should stay above
+	// the user DB connection lifetime so the worker does not keep tenant
+	// connections warm by itself.
+	WarmTenantPollInterval time.Duration
 	// LeaseDuration is the base task lease window used by claim and renew.
 	LeaseDuration time.Duration
 	// RecoverInterval controls how often expired leases are swept back to queue.
@@ -80,6 +87,12 @@ func (o *SemanticWorkerOptions) normalize() {
 	}
 	if o.PollInterval <= 0 {
 		o.PollInterval = defaultSemanticPollInterval
+	}
+	if o.WarmTenantPollInterval <= 0 {
+		o.WarmTenantPollInterval = DefaultSemanticWarmTenantPollInterval()
+	}
+	if minWarm := mysqlutil.DefaultUserConnMaxLifetime(); o.WarmTenantPollInterval < minWarm {
+		o.WarmTenantPollInterval = minWarm
 	}
 	if o.LeaseDuration <= 0 {
 		o.LeaseDuration = defaultSemanticLeaseDuration
@@ -102,6 +115,12 @@ func (o *SemanticWorkerOptions) normalize() {
 	if o.PerTenantConcurrency <= 0 {
 		o.PerTenantConcurrency = defaultSemanticPerTenantConcurrency
 	}
+}
+
+// DefaultSemanticWarmTenantPollInterval returns the default warm cached tenant
+// polling interval derived from tenant user DB pool defaults.
+func DefaultSemanticWarmTenantPollInterval() time.Duration {
+	return mysqlutil.DefaultUserConnMaxLifetime() + defaultSemanticWarmPollMargin
 }
 
 // SemanticWorkerWillRun reports whether NewWithConfig would construct a non-nil
@@ -152,13 +171,7 @@ type semanticWorkerManager struct {
 
 	kicks chan string
 
-	tenantScanCursorSet       bool
-	tenantScanCursorCreatedAt time.Time
-	tenantScanCursorID        string
-	tenantScanRoundRawTenants int
-	tenantScanRoundIncluded   int
-	lastTenantScan            semanticTenantScanSnapshot
-	tenantScanMu              sync.Mutex
+	lastTenantScan semanticTenantScanSnapshot
 
 	priorObservedTenants map[string]struct{}
 
@@ -169,6 +182,7 @@ type semanticWorkerManager struct {
 type semanticTenantRef struct {
 	id     string
 	tenant *meta.Tenant
+	cached bool
 }
 
 type semanticTarget struct {
@@ -293,10 +307,10 @@ func newSemanticWorkerManager(fallback *backend.Dat9Backend, metaStore *meta.Sto
 }
 
 // Kick prompts a worker to attempt claiming tasks for tenantID as soon as one
-// is idle, instead of waiting for the round-robin tenant scan to reach the
-// tenant. Kicks are best-effort: duplicates collapse while one is pending, and
-// the kick is dropped when the buffer is full — the periodic scan remains the
-// durable path that eventually claims every queued task.
+// is idle. Kicks are best-effort: duplicates collapse while one is pending, and
+// the kick is dropped when the buffer is full. In multi-tenant mode this is the
+// only path that may acquire a dormant tenant backend; periodic polling only
+// processes warm cached backends.
 func (m *semanticWorkerManager) Kick(tenantID string) {
 	if m == nil || tenantID == "" {
 		return
@@ -345,8 +359,11 @@ func (m *semanticWorkerManager) Start(ctx context.Context) {
 	fields := []zap.Field{
 		zap.Int("workers", m.opts.Workers),
 		zap.Duration("poll_interval", m.opts.PollInterval),
+		zap.Duration("warm_tenant_poll_interval", m.opts.WarmTenantPollInterval),
+		zap.Duration("effective_poll_interval", m.effectivePollInterval()),
 		zap.Duration("lease_duration", m.opts.LeaseDuration),
 		zap.Duration("recover_interval", m.opts.RecoverInterval),
+		zap.Duration("effective_recover_interval", m.effectiveRecoverInterval()),
 	}
 	fields = append(fields, m.tenantScanLogFields()...)
 	logger.Info(workerCtx, "semantic_worker_manager_started", fields...)
@@ -369,11 +386,11 @@ func (m *semanticWorkerManager) Stop() {
 
 func (m *semanticWorkerManager) workerLoop(ctx context.Context, workerID int) {
 	defer m.wg.Done()
-	ticker := time.NewTicker(m.opts.PollInterval)
+	ticker := time.NewTicker(m.effectivePollInterval())
 	defer ticker.Stop()
 	for {
 		processed := m.processNext(ctx)
-		if processed {
+		if processed && !m.isMultiTenant() {
 			continue
 		}
 		select {
@@ -389,7 +406,7 @@ func (m *semanticWorkerManager) workerLoop(ctx context.Context, workerID int) {
 
 func (m *semanticWorkerManager) recoverLoop(ctx context.Context) {
 	defer m.wg.Done()
-	ticker := time.NewTicker(m.opts.RecoverInterval)
+	ticker := time.NewTicker(m.effectiveRecoverInterval())
 	defer ticker.Stop()
 	for {
 		select {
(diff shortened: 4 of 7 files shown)


srstack (commented):
**Review** — clean, coherent change. The warm-tenant model is well-built and the unit coverage is strong (`ListTenantRefsUsesWarmCachedBackendsOnly`, `NextTargetUsesWarmCachedBackendOnly`, `EffectiveIntervalsUseWarmTenantMinimum`, etc. all pass locally; build/vet green). The cached-vs-wake split is correct: periodic polling resolves refs with `cached=true` → `AcquireCached` (never wakes a cold DB), while `Kick` resolves via `kickRef` with `cached=false` → `Acquire` (may wake), so foreground-created work still gets picked up. Good design.

One behavioral consequence I want to make sure is intended and documented, plus a couple of smaller notes.

### Main thing to confirm: cold-tenant task liveness now depends entirely on Kick
With this change, **both** the periodic poll (`processNext` → `listTenantRefs`) **and lease recovery** (`recoverExpired` → `listTenantRefs`) only ever see warm cached tenants. So for a tenant that has no open DB connection:
- a newly committed semantic task is claimed only if its `Kick` succeeds, and
- **expired-lease recovery never runs for it** until something else warms it.

That second part is the sharp edge. Previously the keyset scan was the durable backstop that guaranteed every active tenant's expired leases eventually got swept. Now consider: tenant commits a task → Kick wakes the backend → a worker claims the task and crashes/times out mid-process (lease set, not released) → tenant goes dormant, connection closes. Nothing kicks this tenant again. The task's lease expires but `recoverExpired` will never visit a cold tenant, so the task is stranded until an unrelated request happens to warm that tenant again. The old code recovered it within one scan cycle.

Is that acceptable under the new model? It may well be (tenants with pending work are usually active, and a later request re-warms them), but it's a real reduction in the recovery guarantee and isn't called out. I'd want either: (a) an explicit note in the PR/runbook that cold-tenant lease recovery is now best-effort and bounded by tenant re-activation, or (b) a periodic full-scan recovery pass at a slow cadence (e.g. the warm-poll interval × N) so dormant tenants with orphaned leases still drain. The kick queue dropping under load (`semanticKickQueueCapacity` overflow → silent drop) compounds this, since a dropped kick for a tenant that then goes dormant has no backstop either.

### Worth confirming
- **Multi-tenant throughput is now interval-bounded.** `workerLoop` no longer `continue`s on a processed task in multi-tenant mode (`processed && !m.isMultiTenant()`), so even a warm tenant with a backlog drains at roughly one `processNext` per `effectivePollInterval` (≥ `DefaultUserConnMaxLifetime`+15s = 60s) per worker. Kicks drain faster (`semanticKickDrainLimit` per kick), so this is probably fine, but confirm a warm tenant that accumulates many tasks without re-kicks isn't drained at 1/minute. If that's intended, a one-line comment on `workerLoop` would help the next reader.
- **`effectivePollInterval` floors at the warm interval but `PollInterval` is still used elsewhere?** The Start log now emits both `poll_interval` and `effective_poll_interval` — good. Just confirm nothing reads `opts.PollInterval` directly on a hot path anymore (the ticker now uses the effective value; a quick grep for `opts.PollInterval` outside the effective- helpers would settle it).

### Minor
- `WarmTenantIDs` and `AcquireCached` both re-check `OpenConnections == 0` (the former filters, the latter in `targetForRef`). The double-check is a deliberate guard against the pool going cold between snapshot and acquire — good — but worth a comment saying so, since it reads as redundant.
- Pool defaults bumped again (`defaultUserMaxOpenConns` 4→6, `defaultUserConnMaxLifetime` 30s→45s). That's the third pool-tuning change across #650/#653/#654. Fine, but the warm-poll interval is now *derived* from the lifetime (`lifetime + 15s`), so any future lifetime change silently shifts worker cadence — the coupling is intentional per the PR text, just flag it so nobody tunes lifetime without realizing it moves the semantic worker too.
- `recoverExpired` calls `m.invalidateTenantBackend(ref.id)` on error for a `cached` ref — confirm invalidating a backend that was acquired via `AcquireCached` (and is still pinned by this `target` until `release()`) interacts correctly with the refcount/retire path.

### CI
The red `ci` check is `TestSSEOutboxSubscriptionReporting` (`sse_outbox_integration_test.go:394: tenant ... not in pod-b subscriptions; got []`) — the podman-backed cross-pod SSE subscription test, surrounded by `sql: database is closed` / container-teardown noise (124 podman refs in the log). This PR doesn't touch SSE subscription/push code (only gofmt + extra status-log fields in `server.go`), and all of this PR's own semantic-worker tests passed in the same run. Looks like the same infra flake hitting #652; re-run `ci`.

Solid work. My one real ask is to resolve the cold-tenant lease-recovery gap (or document it as an accepted tradeoff) before merge.

srstack (commented):
**Re-review @ `843659dd`** — all three of my main points are addressed, cleanly. CI is now green (the earlier red was the SSE infra flake, as suspected). Build/vet/`-race` pass locally. LGTM.

### 1. Cold-tenant lease recovery gap — fixed (this was my blocker)
`recoverExpired` now runs two passes with a shared `visited` set: first `recoveryTenantRefs` — a **keyset-cursor full scan** of active tenants (`cached=false`, so `targetForRef` may open a cold DB) — then the warm-cached `listTenantRefs`. This is exactly the slow-cadence full-scan recovery pass I suggested: dormant tenants with orphaned/expired leases are now swept periodically again, while normal polling stays warm-only. The cursor + wrap logic mirrors the old scan, and `recordRecoveryTenantScan` tracks it separately from the cached scan. Good separation — recovery is explicitly "the only periodic path allowed to open cold tenant DBs."

### 2. Dropped-kick gap — fixed
Kicks are no longer silently dropped on buffer-full. The new `kickQueued` set tracks "already in the channel"; on a full channel the tenant stays in `kickPending`, is recorded as `"delayed"` (not `"dropped"`), and `processKicked`'s `defer m.queueNextPendingKick()` re-enqueues pending tenants as workers drain. I traced the state machine: `clearKickPending` clears both sets, `queuePendingKick` rolls back `kickQueued` on the full-channel path, and the drain-limit re-`Kick`s the same tenant to continue — no leak, no double-queue, and `-race` is clean across the kick/recover tests.

### 3. Snapshot-race comment — added
`targetForRef` now documents why the second `OpenConnections == 0` check exists (WarmTenantIDs is a snapshot; the pool can go cold before the cached acquire pins it). Clear.

### Also nice
- `limitTenantIDsForPass` bounds the warm list with `pageWalkOrder` rotation so a warm set larger than `TenantScanLimit` doesn't starve the tail.
- `listTenantRefs` now re-validates `TenantActive` and resolves the tenant meta for cached refs, so a tenant deactivated while still warm-cached is skipped.
- `WarmTenantPollInterval` normalize now floors at `lifetime + margin` (was `lifetime`), so the poll cadence is strictly above the connection lifetime — closes the equal-to-lifetime edge.

### Minor (non-blocking, optional)
- `recoveryTenantRefs` runs a full active-tenant scan every `effectiveRecoverInterval`. With many tenants this is `TenantScanLimit` rows + a `GetTenant`-free path (it uses the scan rows directly), so it's cheaper than the warm path's per-tenant `GetTenant` — fine. Just confirm the recovery scan opening cold DBs at recover cadence doesn't itself defeat the "don't keep pools warm" goal for tenants that have expired leases every cycle; in practice expired leases should be rare, so this is likely a non-issue.
- `listTenantRefs` now does one `GetTenant` per warm tenant per poll. Bounded by `TenantScanLimit` and warm-set size, so OK, but it's N point lookups where the old cached path did zero — worth a glance if warm sets get large.

Resolved everything I raised. Approving once you're happy with the recovery-scan cost note above.

srstack (commented):
**Re-review @ `7ae8cc8a`** — observability-only follow-up, and a good catch. The new commit splits the `limited` signal out of `wrapped`, which fixes a real semantic conflation I'd glossed over: `recordCachedTenantScan` was previously stuffing "warm list truncated by TenantScanLimit" into the `wrapped` field, whose actual meaning is "recovery cursor wrapped to the head." Now:
- `limited` = warm cached list exceeded the scan limit (rotated by `limitTenantIDsForPass`),
- `wrapped` = recovery-scan cursor wrapped,

with a distinct `tenant_scan_limited` log field and the test updated to assert `limited && !wrapped`. The `kickRef` doc tweak also correctly states that a kick (unlike cached polling) may wake a dormant tenant.

No logic change; build + `ListTenantRefs`/`TenantScan` tests green. Nothing further from me — LGTM.

qiffang (changes requested):
## adversary-2 Review — REVISE on `7ae8cc8`

### Blocker #1: `targetForRef` TOCTOU warning noise

`WarmTenantIDs()` snapshots warm tenants, but by `AcquireCached` time the pool may have gone cold. `targetForRef` returns `fmt.Errorf("cached backend missing for %s")` / `fmt.Errorf("cached backend cold for %s")` which the caller logs as a warning. This is normal TOCTOU behavior, not an anomaly — it should be silently skipped (return nil, nil) to avoid flooding warnings during tenant pool churn.

### Blocker #2: periodic poll throughput regression in multi-tenant mode

`workerLoop` no longer calls `continue` after successful processing in multi-tenant mode:
```go
if processed && !m.isMultiTenant() {
    continue
}
```

After processing a task found via periodic poll, the worker waits `effectivePollInterval` (default 60s) before polling again. `processKicked` re-kicks the tenant after draining, but `processNext` does not. If a warm tenant has 50 queued tasks discovered via poll (not kick), throughput drops from ~50/s to ~1/60s.

The kicked path handles this via `m.Kick(tenantID)` at the end of `processKicked`, but the periodic poll path (`processNext` → `claimAndProcessOne`) has no equivalent re-kick after successful processing.

### Non-blocking: pool tuning bundled with logic change

`defaultUserMaxOpenConns` 4→6 and `defaultUserConnMaxLifetime` 30s→45s are independent tuning changes. If the 45s lifetime causes issues, rollback would also revert the warm polling logic. Consider splitting.

### Non-blocking: `WarmTenantIDs` holds pool lock during `DB().Stats()` calls

`DB().Stats()` acquires sql.DB's internal mutex. For 1000+ tenants, holding pool lock during Stats iteration may add latency to `Acquire`/`Release`. The code already skips retired items, but consider snapshotting items first, then checking Stats outside the lock.

qiffang (commented):
## Consolidated Review — PR #654

**Reviewers**: @dat9-dev1 (lead), @adversary-2, @dat9-dev3

**结论: REVISE** — 设计方向正确（warm-only polling + kick-for-cold），但有 2 个 blocker 和 3 个 non-blocking 需要处理。

---

### Blocker 1: `WarmTenantIDs` holds pool mutex during `DB().Stats()` for all entries

`pkg/tenant/pool.go` — `WarmTenantIDs` holds `p.mu` while iterating all cached entries and calling `e.store.DB().Stats()` on each. `DB().Stats()` acquires the `sql.DB` internal mutex. With the pool max now at 1024, this critical section blocks all concurrent `Acquire`/`Get`/`Invalidate` operations for potentially non-trivial time.

**Suggestion**: Snapshot `tenantID` + `*sql.DB` pairs under the lock, then check `Stats().OpenConnections` outside the lock. The snapshot is still a best-effort signal, consistent with the documented contract.

### Blocker 2: Recovery interval clamped from 5s to 60s — expired-lease latency regression

`effectiveRecoverInterval()` clamps `RecoverInterval` up to `WarmTenantPollInterval` (default 60s). The previous default was 5s. This means expired-lease recovery now takes up to 60s to trigger — a 12x latency regression for re-queuing tasks from crashed workers.

Recovery _needs_ to wake cold tenants by design (the comment says so). Its cadence should not be coupled to the warm-polling cadence. Expired claims are a correctness concern, not a background optimization.

**Suggestion**: Do not clamp `RecoverInterval` against `WarmTenantPollInterval`. Keep it at its own default (5s).

---

### Non-blocking 1: `defaultTenantPoolMaxTenants` silently raised from 128 to 1024

`pkg/tenant/pool.go` — The default pool max changed 8x without mention in the PR description. This affects memory footprint and backend lifetime. Should be documented — even if the intent is to hold more cached backends for warm polling to be useful, the operational impact (memory, file handles, DB connections) should be called out.

### Non-blocking 2: `queueNextPendingKick` worst-case iteration

`queueNextPendingKick` loops: each iteration locks `m.mu`, iterates `kickPending` to find an un-queued tenant, unlocks, then tries to queue it. If there are many delayed kicks and the channel has space, this loop re-acquires the lock and re-scans the map on every iteration. With `kickPending` size bounded by unique tenant count, this is likely fine in practice, but the O(n²) worst case (n pending, each requiring a scan) should have a comment acknowledging it.

### Non-blocking 3: Test `WarmTenantPollInterval` values are misleading

Tests set `WarmTenantPollInterval: 10 * time.Millisecond` but `normalize()` clamps it to 60s. The test still passes because the first iteration runs immediately, but the configured value is fiction. Consider either:
- Adding a comment that the value is clamped for correctness, or
- Using a `normalizeForTest()` variant, or
- Simply not setting it in tests that don't exercise warm polling timing

---

### Positive observations

1. **Kick-for-cold design is sound**: Only `Kick` (driven by upload/write-commit) can wake dormant tenant backends. Periodic polling stays within warm cache. This correctly prevents the worker from keeping idle tenant DB pools alive.

2. **`AcquireCached` is clean**: Uses `makeRelease` (idempotent via `sync.Once`), increments refs, moves to LRU front — mirrors `Acquire` without the create path. TOCTOU between `WarmTenantIDs` and `AcquireCached` is properly documented and handled with the `OpenConnections == 0` re-check in `targetForRef`.

3. **Kick retry mechanism**: Changing from "drop on overflow" to "delay and retry on drain" is a strict improvement — no kick is lost, and `queueNextPendingKick` after each `processKicked` drains the backlog naturally.

4. **Recovery scan split**: Separating `recoveryTenantRefs` (keyset cursor over all active tenants) from `listTenantRefs` (warm cache only) is clean. The `visited` dedup map prevents double-processing.

5. **Tests are comprehensive**: `TestSemanticWorkerRecoversExpiredClaimForColdPoolTenant`, warm/cold/inactive coverage, provider filter tests adapted for new model.

qiffang:
## dat9-dev2 Addendum — PR #654

I agree with the overall **REVISE** direction. After an independent pass, please fold the following into the consolidated review.

### Additional blocker: periodic poll processes only one task per warm tenant per interval

In multi-tenant mode, `workerLoop` no longer continues immediately after a successful `processNext()`:

```go
if processed && !m.isMultiTenant() {
    continue
}
```

`processNext()` claims/processes one task, then the worker waits `effectivePollInterval` (default warm poll cadence) before polling again. The kicked path drains up to `semanticKickDrainLimit` and re-kicks, but the periodic path has no equivalent. A warm tenant with queued backlog discovered by periodic polling/recovery/missed kick can regress to roughly one task per warm-poll interval.

### Additional blocker: failed cold-tenant kick can be consumed permanently

`processKicked()` clears `kickPending` before resolving/acquiring the tenant backend:

```go
m.clearKickPending(tenantID)
ref, ok := m.kickRef(ctx, tenantID)
...
target, err := m.targetForRef(ctx, ref)
```

If `targetForRef` fails transiently while opening a cold tenant backend, the kick is gone. Because normal periodic polling now only visits warm cached tenants, the queued task has no durable retry path until another write triggers a new kick or external traffic warms that tenant. This regresses the previous “periodic scan is durable fallback” behavior and needs a regression test.

### Keep as blockers from the earlier review

- `effectiveRecoverInterval()` should not be clamped to `WarmTenantPollInterval`; expired-lease recovery is a correctness path and intentionally scans cold tenants.
- `WarmTenantIDs()` should not hold `p.mu` while calling `sql.DB.Stats()` for every cached tenant. Snapshot tenant ID + `*sql.DB` under the pool lock, then inspect `Stats()` outside the lock.

### Correction to earlier non-blocking item

The note about `defaultTenantPoolMaxTenants` changing from 128 to 1024 should be withdrawn from this PR review: current `origin/main` already has `defaultTenantPoolMaxTenants = 1024`, and PR #654 does not change that default.

### Validation notes

- Compile/diff check passed locally: `go test -c ./pkg/server ./pkg/tenant ./pkg/mysqlutil ./cmd/drive9-server ./cmd/drive9-server-local && git diff --check origin/main...HEAD`.
- Runtime `go test` / `make test` for `pkg/server` and `pkg/tenant` is blocked on this machine by local testcontainers setup: `rootless Docker not found`. `pkg/mysqlutil`, `cmd/drive9-server`, and `cmd/drive9-server-local` tests pass.

qiffang:
## dat9-dev2 re-review — `af0766773a4ea43d86cf29b7f9dac7183da06cfa`

Verdict: **REVISE**

The previous blockers are mostly addressed: warm periodic scan no longer sleeps after every successful task, recover interval is no longer clamped by warm-poll cadence, `WarmTenantIDs` no longer calls `DB.Stats()` under the pool mutex, and kick pending is no longer cleared before `targetForRef` succeeds.

Two scheduling blockers remain in the latest code:

### B1 — failed cold-tenant kick now becomes a tight retry loop

`processKicked` defers `queueNextPendingKick()` before any failure path. On `kickRef` lookup error or `targetForRef` / `pool.Acquire` failure, the pending mark is intentionally left in place, so the deferred `queueNextPendingKick()` immediately enqueues the same tenant again. With no warm work, the worker loops back, reads the same kick, fails again, and repeats without delay/backoff.

Relevant paths:
- `pkg/server/semantic_worker.go`: `processKicked` defers `queueNextPendingKick()` at entry, then returns on lookup/acquire failure.
- `pkg/server/semantic_worker.go`: `queueNextPendingKick()` immediately requeues any pending non-queued tenant.
- `TestSemanticWorkerKickedAcquireFailureKeepsKickPending` currently asserts immediate requeue (`len(m.kicks) == 1`), which locks in the hot-loop behavior instead of just preserving retryability.

Why this blocks merge: a down tenant DB or transient metadata/backend-open failure can make semantic workers spin on the same failed tenant, hammer logs/metadata/pool acquisition, and starve useful work. The correct property is “not lost and retried later”, not “immediately retry forever”. This needs explicit delayed retry/backoff or another bounded retry mechanism, plus a regression test proving persistent acquire failure does not busy-loop.

### B2 — warm backlog can starve kicked cold tenants

`workerLoop` always calls `processNext()` before reading from `m.kicks`, and if `processNext()` processed one task it `continue`s immediately. That means with any steady warm-tenant backlog, workers can keep draining warm cached tenants and never reach the `select` branch that handles kicked cold tenants.

Relevant code:
- `pkg/server/semantic_worker.go`: `workerLoop` lines around `processed := m.processNext(ctx)` and `if processed { continue }` run before the `case tenantID := <-m.kicks` branch.

Why this blocks merge: Kick is the only normal path allowed to wake dormant tenant DBs. A newly uploaded task for a cold tenant should not wait for unrelated warm backlog to go empty. The latest tests only cover kicks when the worker has already reached `select`; they do not cover a kick arriving while `processNext()` keeps returning `true`.

Validation I ran locally:
- `go test -c ./pkg/server ./pkg/tenant ./pkg/mysqlutil ./cmd/drive9-server ./cmd/drive9-server-local` — pass
- `go test ./pkg/mysqlutil ./cmd/drive9-server ./cmd/drive9-server-local -count=1` — pass
- `go test ./pkg/server -run 'TestSemanticWorker(MultiTenantEffectiveIntervalsKeepRecoverIndependent|SingleTenantEffectiveIntervalsKeepConfiguredPoll|KickDedupAndOverflow|KickedAcquireFailureKeepsKickPending|KickedLookupFailureKeepsKickPending|KickClaimsFallbackTaskWithoutScan|KickClaimsPoolTenantTaskWithoutScan|TargetForCachedRefMissingSkipsSilently)' -count=1` — blocked by local testcontainers startup: `rootless Docker not found`

CI is green, but these are control-flow issues not covered by the current tests.

PR #656, opened 2026-07-01
feat: default write cache size to 1GB

qiffang:
## Summary
- Set `WriteCacheSizeMB` default to 1024 (1GB), matching read cache default
- Use negative sentinel (-1) for explicit disable via `--write-cache-size-mb=0`, consistent with `WriteCacheFreeRatio` pattern
- CLI `--write-cache-size-mb` default changed from 0 to 1024

## Semantics
- No flag: write cache defaults to 1GB
- `--write-cache-size-mb=0`: explicitly disable byte quota (free-ratio still active)
- `--write-cache-size-mb=2048`: custom 2GB
- Byte quota and free-ratio enforced simultaneously, strictest wins

## Tests
- `TestSetDefaultsWriteCacheSizeMBDefault`: zero-value → 1024
- `TestSetDefaultsWriteCacheSizeMBExplicitValue`: explicit 2048 preserved
- `TestSetDefaultsWriteCacheSizeMBNegativeDisables`: sentinel -1 → 0 (disabled)
- `TestDefaultByteQuotaEnforcesENOSPC`: pending near 1GB limit → ENOSPC
- `TestByteQuotaDisabledAllowsUnlimitedWrites`: quota=0 allows writes
- `TestFreeRatioIndependentOfByteQuota`: free-ratio fires independently of byte quota

Follow-up to PR #652.

### DIFF
--- a/cmd/drive9/cli/mount.go
+++ b/cmd/drive9/cli/mount.go
@@ -155,7 +155,7 @@ func fsMountCmdWithBackground(args []string, background bool) error {
 	uploadConcurrency := fs.Int("upload-concurrency", 16, "maximum concurrent background uploads issued by FUSE")
 	dirCacheMaxEntries := fs.Int("dir-cache-max-entries", 200000, "maximum entries per directory in namespace cache before complete marking is disabled")
 	writeCacheFreeRatio := fs.Float64("write-cache-free-ratio", 0.10, "minimum filesystem free-space ratio before write-back refuses writes with ENOSPC (0 disables)")
-	writeCacheSizeMB := fs.Int64("write-cache-size-mb", 0, "shadow cache byte quota in MB (0 = disabled); shadow writes exceeding this return ENOSPC")
+	writeCacheSizeMB := fs.Int64("write-cache-size-mb", 1024, "shadow cache byte quota in MB (default 1024; 0 disables); shadow writes exceeding this return ENOSPC")
 	commitQueueMaxPending := fs.Int("commit-queue-max-pending", 100, "maximum pending entries in CommitQueue before backpressure")
 	allowOther := fs.Bool("allow-other", false, "allow other users to access mount")
 	readOnly := fs.Bool("read-only", false, "mount as read-only")
@@ -333,6 +333,10 @@ func fsMountCmdWithBackground(args []string, background bool) error {
 	}
 	// Map CLI 0 → -1 sentinel so MountOptions.setDefaults() knows the user
 	// explicitly disabled the guard rather than left it unset.
+	normalizedWriteCacheSizeMB := *writeCacheSizeMB
+	if normalizedWriteCacheSizeMB == 0 {
+		normalizedWriteCacheSizeMB = -1
+	}
 	normalizedWriteCacheFreeRatio := *writeCacheFreeRatio
 	if normalizedWriteCacheFreeRatio == 0 {
 		normalizedWriteCacheFreeRatio = -1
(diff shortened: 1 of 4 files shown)


qiffang:
## adversary-1 review — GREEN (conditional on CI)

No blockers found in PR #656. The branch is based on current `main`, the diff is scoped to the write-back cache default follow-up, and the core semantics match the agreed contract:

- default `MountOptions{}` maps `WriteCacheSizeMB` to `1024`
- explicit positive values are preserved
- negative sentinel disables the fixed byte quota
- CLI default is `1024`, and CLI `--write-cache-size-mb=0` maps to the negative sentinel
- quota enforcement still runs alongside free-ratio protection

Regression coverage is now present for default/sentinel behavior and byte-quota ENOSPC. Existing `shadow_test.go` already covers free-ratio ENOSPC; the new free-ratio test is acceptable as additional coverage, though it is environment-sensitive and partly permissive. If this area changes again, prefer priming cached free/total bytes like `TestFreeRatioThrottledReEvaluatesCachedCapacity` for deterministic free-ratio assertions.

Validation from this environment:
- `git diff --check origin/main...pr-656-review` passed
- local Go tests were not run because this machine has no `go` binary
- GitHub `ci` and `drive9-server-local e2e smoke` were still in progress at review time

I cannot submit a counting GitHub approval from this machine because the authenticated account is the PR author (`qiffang`).

srstack (commented):
**Review** — small, clean follow-up to #652. The net delta over #652 is tiny (CLI default + sentinel + `setDefaults` branch + tests), and it mirrors the existing `WriteCacheFreeRatio` pattern exactly. Build/vet/tests pass locally (`-race`), CI green. LGTM with two things to flag.

### Logic — correct and symmetric
- CLI `--write-cache-size-mb` default 0 → 1024; `=0` maps to the `-1` sentinel; `setDefaults` treats `<0` as disabled (→0) and `==0` as unset (→ `defaultWriteCacheSizeMB`). Same shape as `WriteCacheFreeRatio`, so operators keep the intuitive "0 = disable" while the code still distinguishes unset from explicitly-disabled. Good.
- The three `setDefaults` tests pin all three states, and `TestDefaultByteQuotaEnforcesENOSPC` confirms the 1GB quota actually fires.

### Please flag: this is a behavior change, not just a default
Before this PR, the byte quota was **off** unless explicitly set; after it, every mount enforces a **1GB write-back cache ceiling by default**. That's a meaningful semantics change for existing users: a workload that buffers >1GB of dirty write-back data (large file staged offline, slow upstream, many concurrent writers) will now start getting `ENOSPC` where it previously succeeded. That's arguably the safer default, but it should be called out in the release notes / changelog so operators who rely on large write-back buffering know to set `--write-cache-size-mb=0` or a higher value. The free-ratio guard (also default-on since #652) plus this means two independent default-on ENOSPC sources now.

### Merge-order dependency
This branch is stacked on #652 (`d590df55`) but targets `main`, and #652 isn't merged yet — so the PR diff shows all 1159 lines of #652 too. Make sure #652 merges first (or this is rebased onto it), otherwise merging #656 alone would pull in #652's shadow-store rewrite under a "default write cache size" title. Worth noting in the PR so the merge queue orders them correctly.

### Minor
- `TestFreeRatioIndependentOfByteQuota` is effectively self-exempting: on a disk with >99.99% free it logs and passes without asserting anything. It documents intent but provides little real coverage — consider seeding the ShadowStore's cached free/total bytes directly (as the throttled-path tests do) to make the free-ratio trigger deterministic regardless of the CI disk. Non-blocking.
- `defaultWriteCacheSizeMB << 20` (1024<<20 = 1GiB) is fine for int64; just noting the MB→bytes shift is only safe because the value is small — a user passing a multi-petabyte `--write-cache-size-mb` could overflow, but that's absurd input and out of scope.

Solid. Approve once #652 is confirmed to land first and the default-on 1GB ceiling is noted for operators.

srstack (approved):
Approving — code is clean and I found no blocker. Logic is correct and symmetric with the existing `WriteCacheFreeRatio` pattern; build/vet/tests pass locally (`-race`) and CI is green.

Two non-code items to handle at merge time (not blocking approval):
1. **Merge order** — this branch is stacked on #652 (`d590df55`) but targets `main`; ensure #652 lands (or this rebases onto it) so #656 doesn't pull in the shadow-store rewrite under this title.
2. **Release note** — the 1GB write-back ceiling is now default-on, a behavior change for workloads that buffer >1GB of dirty data; note `--write-cache-size-mb=0` as the opt-out.

Optional nit from my review: `TestFreeRatioIndependentOfByteQuota` self-exempts on large-disk CI — seeding cached free/total bytes would make it deterministic.

PR #657, opened 2026-07-01
feat: semantic worker outbox+shard to avoid waking idle tenant TiDBs

mornyx:
## Summary

Add a central `semantic_notify_outbox` table in the always-provisioned meta DB and a per-pod outbox poller, replacing the semantic worker's keyset tenant scan with a signal-driven model so idle tenant TiDBs can scale to zero.

## Background

The drive9 semantic worker currently uses keyset pagination over the `tenants` table to discover queued semantic tasks. With ~100k tenants (most idle), this polling keeps every tenant's serverless TiDB awake, preventing scale-to-zero and incurring sustained RCU cost.

### New meta DB tables

1. `semantic_notify_outbox` (append-only, same structure as `sse_notify_outbox`):
   - `id BIGINT UNSIGNED AUTO_INCREMENT PRIMARY KEY`
   - `tenant_id VARCHAR(64) NOT NULL`
   - `created_at DATETIME(3)`
   - `INDEX idx_created (created_at)`
   - One row per semantic task enqueue; poller uses auto-increment ID as cursor.

2. `semantic_outbox_cursor` (per-pod checkpoint):
   - `pod_id VARCHAR(64) PRIMARY KEY`
   - `last_id BIGINT UNSIGNED NOT NULL DEFAULT 0`
   - `updated_at DATETIME(3)`
   - Persists each pod's outbox consumption cursor so a restart does not skip unprocessed rows. This is necessary because unlike `sse_notify_outbox` (which has SSE reconnect replay as a fallback), semantic tasks have no replay path — a skipped outbox row means a task stays stuck if the tenant does not write again.

### Write path

After a write/upload transaction commits and enqueues a semantic task, the existing `SetSemanticTaskNotifier` callback now:
1. Best-effort `INSERT INTO semantic_notify_outbox (tenant_id) VALUES (?)` into the central meta DB (same pattern as `sse_notify_outbox`).
2. Fires the in-process `worker.Kick(tenantID)` for same-pod ~0ms fast path.

If the outbox insert fails, the durable `semantic_tasks` row remains in the tenant DB and the next write for that tenant produces a fresh outbox row.

### Per-pod outbox poller

New `semanticOutboxPoller` (structure mirrors `notify_poller.go`):
- Runs on every pod (not leader-gated), polls only the always-provisioned meta DB.
- Maintains `lastID` cursor. On startup, recovers from `semantic_outbox_cursor`; if no cursor exists (first launch), initializes to `MAX(id)` — safe because the pod has never owned work before.
- Each 200ms tick: `SELECT id, tenant_id FROM semantic_notify_outbox WHERE id > lastID ORDER BY id LIMIT 1000`. For each row belonging to this pod's shard, calls `worker.Kick(tenantID)`. If batch is full, drains immediately.
- Persists cursor to `semantic_outbox_cursor` every 5s (not every tick) to reduce meta DB write QPS. On crash, at most 5s of rows are replayed — duplicate kicks are harmless (`ClaimSemanticTask` with `FOR UPDATE SKIP LOCKED` returns no-op for already-claimed tasks).

### Hash-shard across pods

Semantic worker moves from leader-only to every-pod (like `notify_poller`). Each pod only processes tenants assigned to its shard:
- `semanticShardResolver` periodically (5s) queries `pod_registry` for active pods, sorts by `pod_id`, computes `myRank` and `numShards`.
- Tenant → pod: `jumpConsistentHash(hash(tenantID), numShards) == myRank`. Jump consistent hash minimizes migration on pod add/remove (~1/N tenants move).
- On shard rebalancing, the outbox durable signal guarantees no loss: the new owner's next poll reads the pending row. `FOR UPDATE SKIP LOCKED` prevents duplicate claims during the transition.
- Single-pod mode: `ownsTenant` always returns true.

### Semantic worker changes

- Remove keyset tenant scan (`listTenantRefs`, `nextTenantScanPage`, `tenantScanCursor`, `pageWalkOrder`, `recordTenantScan`).
- Worker loop blocks on `kicks` channel; no periodic scan path in multi-tenant mode. In single-tenant mode (no meta/pool), the worker periodically polls the fallback backend for queued tasks.
- `processKicked` adds shard check at entry: `if !ownsTenant(tenantID) { return }`.
- Recovery of expired leases is bound to kick: when `ClaimSemanticTask` returns no task for a kicked tenant, additionally call `RecoverExpiredSemanticTasks` on that tenant. This eliminates the need for a separate full-scan recovery loop — only tenants with pending outbox signals get their expired leases recovered.

### Outbox cleanup (leader-gated)

Leader-gated goroutine prunes `semantic_notify_outbox` rows older than retention (default 1h) and cleans `semantic_outbox_cursor` rows for pods no longer in `pod_registry`. Same pattern as `cleanupSSENotifyOutbox`.

### Configuration

New env vars:
- `DRIVE9_SEMANTIC_OUTBOX_POLL_INTERVAL_MS` (default 200): outbox poller interval
- `DRIVE9_SEMANTIC_OUTBOX_CURSOR_FLUSH_MS` (default 5000): cursor persistence interval
- `DRIVE9_SEMANTIC_SHARD_REFRESH_INTERVAL_MS` (default 5000): shard ring refresh interval

Removed: `DRIVE9_SEMANTIC_TENANT_LIMIT` (tenant scan limit, no longer applicable).

## Correctness

| Requirement | Guarantee |
|---|---|
| No missed events | Outbox is a durable signal in meta DB; poller recovers cursor from `semantic_outbox_cursor` on restart — no rows skipped. Shard rebalancing: new owner's next poll reads pending rows. |
| No duplicate processing | `ClaimSemanticTask` uses `FOR UPDATE SKIP LOCKED` — single owner per task. Duplicate kicks (restart replay, shard transition) are no-ops. |
| No wrong delivery | Outbox stores only `tenant_id`; worker opens the correct tenant TiDB. `allowedTaskTypes` filter prevents claiming unsupported task types. |
| Low latency | Same-pod write: in-process kick (~0ms). Cross-pod: outbox poller within 200ms. |
| Idle TiDB not polled | No component periodically queries idle tenant TiDBs. Poller targets meta DB only. Worker opens tenant TiDB only on kick. Recovery is bound to kick. |
| Restart safety | Cursor persisted to meta DB; restart replays at most 5s of rows (duplicate kicks, safe). First launch skips to MAX(id) — safe, pod never owned work. |

## Files

New: `pkg/server/semantic_outbox_poller.go`, `pkg/server/semantic_shard.go`, `pkg/server/semantic_outbox_poller_test.go`, `pkg/server/semantic_shard_test.go`
Modified: `pkg/meta/meta.go`, `pkg/server/semantic_worker.go`, `pkg/server/server.go`, `cmd/drive9-server/main.go`, `pkg/server/semantic_worker_test.go`, `cmd/drive9-server/main_test.go`, `cmd/drive9-server-local/main.go`, `cmd/drive9-server-local/main_test.go`

## Tests

- `go test ./pkg/server/` — all semantic worker tests pass
- `make lint` — 0 issues
- New unit tests for jump consistent hash distribution, migration, and shard resolver ownership

## Summary by CodeRabbit

* **New Features**
  * Added multi-pod semantic task routing with shared outbox processing and shard-aware tenant assignment.
  * Introduced configurable timing controls for semantic outbox polling, cursor flushing, and shard refresh behavior.
  * Improved semantic worker handling so tasks can be kicked and processed without relying on broad tenant scans.

* **Bug Fixes**
  * Removed dependence on a tenant scan limit setting and updated related behavior to use the new routing flow.
  * Added cleanup for stale semantic outbox data and cursor progress to keep processing stable over time.


srstack (changes requested):
Thanks for writing this up. I agree with the high-level direction: moving semantic scheduling into the always-on meta DB is the right way to stop periodically waking cold tenant DBs. The same-pod `Kick` fast path plus a central scheduling signal is also the right shape.

However, I do not think this PR is mergeable in its current form. The current implementation treats `semantic_notify_outbox` as an append-only notification stream with per-pod cursors, but semantic tasks do not have an independent replay/reconciliation path after the tenant scan is removed. That makes several failure modes lose the only wake-up signal for a tenant.

[CRITICAL] Cursor advancement can skip unhandled semantic notifications

In `pkg/server/semantic_outbox_poller.go:153-163`, the poller advances `lastID` for every row before proving that the signal was safely handled:

```go
p.lastID = row.ID
if p.shardFn != nil && p.shardFn(row.TenantID) {
    p.worker.Kick(row.TenantID)
}
```

This consumes rows even when:

- the tenant is not currently owned by this pod,
- `Kick` is dropped because the worker queue is full,
- shard membership is stale or changing,
- the eventual tenant DB open/claim path fails.

Since the multi-tenant worker no longer scans tenant DBs, advancing past an unhandled row can leave that tenant semantic task stuck until an unrelated future write creates another row.

[CRITICAL] Per-pod cursors are not safe with shard rebalancing

The current design has every pod read the global outbox and skip rows it does not own. That is unsafe with ownership changes.

Example:

1. Row `id=100` is written for tenant A.
2. Pod-2 reads it, decides it does not own tenant A, and advances its cursor to `100`.
3. The previous owner pod dies before claiming the task, or the shard ring changes so tenant A moves to pod-2.
4. Pod-2 will never re-read `id=100`, so the task can remain queued indefinitely.

This is a fundamental mismatch between per-pod cursors and shard-based ownership. Either the outbox needs an acknowledgement/claim model, or the scheduler should use a durable meta tenant-work table instead of a per-pod cursor over a broadcast stream.

[MAJOR] `InsertSemanticNotify` cannot be best-effort if it is the only cross-pod wake-up

In `pkg/server/server.go:452-461`, the notify insert error is ignored:

```go
_ = s.meta.InsertSemanticNotify(notifyCtx, tenantID)
s.semanticWorker.Kick(tenantID)
```

After tenant scanning is removed, this outbox row becomes the only cross-pod wake-up. If the tenant DB transaction commits a `semantic_tasks` row, but the meta insert fails or the process dies before inserting, the task may never be discovered. The local kick does not cover this if the current pod is not the tenant shard owner.

The PR body says the next tenant write will produce a fresh outbox row, but there may not be a next write. That is not a sufficient correctness guarantee for durable background work.

[MAJOR] Retry-ready tasks have no future wake-up

In multi-tenant mode, `workerLoop` no longer polls tenant DBs, and `recoverLoop` only calls `observeOnce` (`pkg/server/semantic_worker.go:386-395`). If task processing fails and `retryTask` schedules the task with a future `available_at`, the immediate re-kick can run before the retry is claimable. When the retry time arrives, no new outbox row is written, so the task can sit queued indefinitely.

This needs either a delayed meta wake-up (`next_check_at`) or a retained reconciliation path for retry-ready work.

[MAJOR] Expired lease recovery is no longer durable

Recovery is now only attempted in `processKicked` when a kicked tenant has no claimable queued task. That misses the important crash case:

1. A worker claims a task and the original outbox signal is consumed.
2. The worker/pod crashes before finalizing the task.
3. The lease expires.
4. No new outbox row is emitted on lease expiry.
5. Multi-tenant `recoverLoop` no longer scans/recoveries tenant DBs.

The task can remain `processing` forever unless another write for that tenant happens. Lease expiry needs its own wake-up path, or recovery needs a durable meta-level schedule.

[MAJOR] Cursor initialization can skip valid rows

`GetSemanticOutboxCursor` conflates no cursor row with a valid persisted `last_id = 0` (`pkg/meta/meta.go:3383-3397`). Then `initCursor` treats `lastID == 0` as first launch and jumps to `MAX(id)` (`pkg/server/semantic_outbox_poller.go:85-101`).

Also, on cursor load errors, the poller falls back to `MAX(id)`. For semantic scheduling, this is not safe: cursor load errors should fail closed or replay from a safe cursor. Skipping to max can drop unprocessed semantic notifications.

[MAJOR] Age-based cleanup can delete unprocessed semantic wake-ups

`cleanupSemanticNotifyOutbox` deletes rows purely by age (`pkg/server/server.go:970`). Unlike SSE notify rows, semantic notify rows are currently the only wake-up signal. If a pod is down or its cursor lags beyond the retention window, pruning old rows can permanently remove the signal while the tenant task still exists.

If the outbox model is kept, cleanup must account for consumption state, for example by deleting only rows older than the minimum safe cursor/ack point. Age-only cleanup is not correct here.

[MAJOR] Pod registry dependency can make semantic workers own no tenants

`semanticShardResolver.ownsTenant` returns false when `metaStore != nil`, `podID != ""`, but the pod is not in the active ring (`pkg/server/semantic_shard.go:121-127`). That is reasonable only if pod registration is guaranteed.

Today pod registry registration is tied to the notify infrastructure and can be impacted by `PodAddr` setup. If `DRIVE9_POD_ID` is set but the pod is not registered, the semantic worker will skip all tenants. Either semantic membership must not depend on the SSE push address, or startup should fail clearly when sharded semantic processing is enabled but the pod cannot register.

[MAJOR] CI is currently failing

The `ci` check is failing in `pkg/server`:

```text
leader_lifecycle_test.go:194: semantic worker should be stopped after onLose
--- FAIL: TestLeaderGatedWorkersStartAndStop
```

This appears related to moving multi-tenant semantic workers from leader-gated lifecycle into notify infrastructure/every-pod lifecycle. The lifecycle boundary needs to be made explicit and the tests updated accordingly.

Suggested direction

I would not keep patching the current append-only outbox + per-pod cursor model. A safer design would be a meta-level durable tenant work table, for example:

```text
semantic_tenant_work (
  tenant_id PRIMARY KEY,
  dirty / version,
  next_check_at,
  lease_owner,
  lease_until,
  updated_at
)
```

On semantic task enqueue, upsert the tenant row in meta and set `next_check_at <= NOW()`. Workers claim due tenant rows from meta with a lease, open only that tenant DB, then update the meta row based on tenant DB state:

- queued work remains: keep dirty and re-check soon,
- retry task exists: set `next_check_at` to the next `available_at`,
- processing task exists: set `next_check_at` to the next `lease_until`,
- no queued/processing/retry work remains: clear dirty.

That avoids scanning cold tenant DBs while also avoiding per-pod cursor loss, shard rebalance loss, dropped-kick loss, retry starvation, and expired-lease starvation.

So the overall goal is right, but the current implementation needs a scheduling/ack model change before it can be considered safe.

srstack (commented):
I like the direction of moving Semantic Worker discovery to a central meta-DB signal path. That is the right long-term shape if we want idle tenant TiDBs to scale down. However, I do not think this PR is safe to merge in its current form because the new outbox path becomes the only durable trigger while several normal task lifecycle transitions do not produce durable follow-up signals.

[MAJOR] `InsertSemanticNotify` is best-effort, but the PR removes the tenant scan fallback. If the tenant transaction commits a durable `semantic_tasks` row and the meta-DB outbox insert fails, the task can stay queued forever unless that tenant happens to write again. That was acceptable when the periodic scan was still the durable fallback, but it is not safe once the worker is outbox-only.

[MAJOR] retry scheduling can strand tasks. A task that fails and is moved to a future `retry_at` consumes its current outbox row, but `RetrySemanticTask` does not enqueue a new semantic notify row or schedule a delayed re-kick. When `retry_at` arrives, there may be no kick and no tenant scan, so the task can remain claimable but unprocessed until a later unrelated tenant write.

[MAJOR] expired lease recovery is now only executed after a kick that finds no queued task. If a pod consumes an outbox row, claims a task, flushes its outbox cursor, and then crashes, the task eventually has an expired lease but there may be no remaining outbox row to kick recovery. Without a separate durable recovery trigger, `processing` tasks can get stuck indefinitely.

[MAJOR] first-launch cursor initialization to `MAX(id)` is not safe for semantic tasks. The SSE poller can do this because clients have an independent replay path via `fs_events`; semantic tasks do not. A new pod, a fresh pod ID after rollout, or the first deployment of this mechanism can skip existing rows that still represent unprocessed tenant tasks.

[MAJOR] every pod advances its own cursor over rows it does not own. During startup, pod registry lag, or shard-ring transitions, a pod can decide it does not own a tenant and still persist a cursor past that row. If no pod kicks that row during the transition window, the signal is lost. I think the design needs an ack/claim state, owner-specific query semantics, or another repair path before non-owner rows are considered consumed.

[MAJOR] `collectObservation` says it observes only warm cached tenants, but it calls `pool.ActiveTenantStores()`. That only means the backend is cached; it does not require `DB().Stats().OpenConnections > 0`. This regresses the recent semantic worker change because metrics observation can reconnect cached-but-cold tenant DBs. It should use the warm/open-connection semantics from the existing pool helper, or introduce a `WarmTenantStores` helper.

[MINOR] `cleanupSemanticNotifyOutbox` says it also cleans stale `semantic_outbox_cursor` rows, and `DeleteSemanticOutboxCursor` exists, but the cleanup currently only deletes old notify rows. Either implement cursor cleanup or remove the claim from the comment.

The current CI failure also reflects the lifecycle change: `TestLeaderGatedWorkersStartAndStop` still expects the semantic worker to stop on leadership loss, while this PR starts it under notify infrastructure on every pod. That test can be updated if we keep the every-pod design, but I would address the durability issues above first.

My recommendation is to keep the outbox as a fast path, but not make it the only path until the missing durability cases are covered. A safer version would either make the outbox signal itself reliable and acked, or retain a low-frequency repair/recovery path that does not wake every idle tenant DB but can recover from lost notify rows, retries becoming due, and expired leases after crashes.

PR #659, opened 2026-07-01
Route quota accounting through meta runtime

srstack:
## Summary

- route runtime quota accounting through the meta DB `quota_mutation_log` path instead of tenant/user DB `quota_outbox`
- remove `DRIVE9_QUOTA_SOURCE`; central quota is active when a meta quota store is wired
- stop starting the tenant quota outbox worker at runtime and stop enqueueing outbox rows from create/overwrite/upload-complete paths
- make meta quota mutation log insert failures surface as errors instead of fail-open drops
- switch upload reservation to a pure meta DB reserve-first path and protect finalize with the `completing` status
- update replay cadence knobs and document the accepted post-tenant-commit/pre-meta-log crash window

## Correctness note

This intentionally does not claim durable equivalence with the old tenant outbox. Runtime quota accounting no longer touches the user DB, but a process crash after tenant commit and before meta `quota_mutation_log` insert can still require reconciliation/backfill. Meta log insert failure is now visible to callers and metrics instead of being silently dropped.

## Tests

- `make lint`
- `go test ./pkg/backend ./pkg/meta ./pkg/tenant ./cmd/drive9-server ./cmd/drive9-server-local -count=1`
- `go test ./... -count=1`

## Summary by CodeRabbit

* **New Features**
  * Central quota runtime now runs through the meta database with durable mutation logging and background replay to converge usage.
  * Upload reservations now use an intermediate **“completing”** state to improve end-to-end reservation/settlement behavior.

* **Bug Fixes**
  * Startup now fails fast if the deprecated quota-source environment setting is present (removed from runtime/help flows).
  * Quota replay timing is now configurable, and quota mutation error handling is more robust.

* **Documentation**
  * Updated local setup, server help, and central-quota cutover/e2e guidance to reflect the meta-based configuration path.

### DIFF
--- a/pkg/backend/mutation_replay.go
+++ b/pkg/backend/mutation_replay.go
@@ -4,6 +4,8 @@ import (
 	"context"
 	"database/sql"
 	"encoding/json"
+	"os"
+	"strconv"
 	"strings"
 	"time"
 
@@ -13,11 +15,31 @@ import (
 )
 
 const (
-	replayPollInterval = 30 * time.Second
-	replayMinAge       = 5 * time.Second // only replay mutations older than 5s (avoid racing with inline apply)
-	replayBatchLimit   = 100
-	replayMaxRetries   = 5
+	defaultReplayPollInterval = time.Second
+	defaultReplayMinAge       = time.Second // only replay older mutations to avoid racing inline apply
+	replayBatchLimit          = 100
+	replayMaxRetries          = 5
 )
+
+func replayPollInterval() time.Duration {
+	return envDurationMS("DRIVE9_QUOTA_REPLAY_POLL_MS", defaultReplayPollInterval)
+}
+
+func replayMinAge() time.Duration {
+	return envDurationMS("DRIVE9_QUOTA_REPLAY_MIN_AGE_MS", defaultReplayMinAge)
+}
+
+func envDurationMS(name string, def time.Duration) time.Duration {
+	raw := strings.TrimSpace(os.Getenv(name))
+	if raw == "" {
+		return def
+	}
+	ms, err := strconv.Atoi(raw)
+	if err != nil || ms < 0 {
+		return def
+	}
+	return time.Duration(ms) * time.Millisecond
+}
 
 // MutationReplayWorker reads pending mutations from the quota_mutation_log
 // and applies them idempotently. It runs as a background goroutine.
@@ -56,7 +78,7 @@ func (w *MutationReplayWorker) run(ctx context.Context) {
 	defer close(w.done)
 
 	logger.Info(ctx, "mutation_replay_worker_started")
-	ticker := time.NewTicker(replayPollInterval)
+	ticker := time.NewTicker(replayPollInterval())
 	defer ticker.Stop()
 
 	for {
@@ -78,7 +100,7 @@ func (w *MutationReplayWorker) run(ctx context.Context) {
 // fatal error occurred (e.g. database closed) and the loop should stop.
 func (w *MutationReplayWorker) replayBatch(ctx context.Context) (fatal bool) {
 	start := time.Now()
-	entries, err := w.store.ListPendingMutations(ctx, replayMinAge, replayBatchLimit)
+	entries, err := w.store.ListPendingMutations(ctx, replayMinAge(), replayBatchLimit)
 	if err != nil {
 		if strings.Contains(err.Error(), "database is closed") || strings.Contains(err.Error(), "connection refused") {
 			logger.Info(ctx, "mutation_replay_worker_db_closed")
--- a/pkg/backend/mutation_replay.go
+++ b/pkg/backend/mutation_replay.go
@@ -30,7 +30,11 @@ func replayPollInterval() time.Duration {
 }
 
 func replayMinAge() time.Duration {
-	return envDurationMS("DRIVE9_QUOTA_REPLAY_MIN_AGE_MS", defaultReplayMinAge)
+	d := envDurationMS("DRIVE9_QUOTA_REPLAY_MIN_AGE_MS", defaultReplayMinAge)
+	if d <= 0 {
+		return defaultReplayMinAge
+	}
+	return d
 }
 
 func envDurationMS(name string, def time.Duration) time.Duration {
--- a/pkg/backend/mutation_replay.go
+++ b/pkg/backend/mutation_replay.go
@@ -22,7 +22,11 @@ const (
 )
 
 func replayPollInterval() time.Duration {
-	return envDurationMS("DRIVE9_QUOTA_REPLAY_POLL_MS", defaultReplayPollInterval)
+	d := envDurationMS("DRIVE9_QUOTA_REPLAY_POLL_MS", defaultReplayPollInterval)
+	if d <= 0 {
+		return defaultReplayPollInterval
+	}
+	return d
 }
 
 func replayMinAge() time.Duration {
@@ -166,18 +170,18 @@ func (w *MutationReplayWorker) replayBatch(ctx context.Context) (fatal bool) {
 
 func (w *MutationReplayWorker) replayOne(ctx context.Context, entry MutationLogView) error {
 	return w.store.InTx(ctx, func(tx *sql.Tx) error {
-		if err := w.applyMutation(tx, entry); err != nil {
+		if err := w.applyMutation(ctx, tx, entry); err != nil {
 			return err
 		}
 		return w.store.MarkMutationAppliedTx(tx, entry.ID)
 	})
 }
 
-func (w *MutationReplayWorker) applyMutation(tx *sql.Tx, entry MutationLogView) error {
-	return applyCentralQuotaMutationTx(w.store, tx, entry.TenantID, entry.MutationType, entry.MutationData, entry.ID)
+func (w *MutationReplayWorker) applyMutation(ctx context.Context, tx *sql.Tx, entry MutationLogView) error {
+	return applyCentralQuotaMutationTx(ctx, w.store, tx, entry.TenantID, entry.MutationType, entry.MutationData, entry.ID)
 }
 
-func applyCentralQuotaMutationTx(store MetaQuotaStore, tx *sql.Tx, tenantID, mutationType string, mutationData json.RawMessage, logID int64) error {
+func applyCentralQuotaMutationTx(ctx context.Context, store MetaQuotaStore, tx *sql.Tx, tenantID, mutationType string, mutationData json.RawMessage, logID int64) error {
 	switch mutationType {
 	case "file_create":
 		var data fileCreateMutationData
@@ -230,7 +234,7 @@ func applyCentralQuotaMutationTx(store MetaQuotaStore, tx *sql.Tx, tenantID, mut
 		// caller (replayOne wraps applyMutation + MarkMutationAppliedTx in
 		// the same InTx), so this delegation is safe w.r.t. the Fix 2
 		// status='pending' guard.
-		return applyUploadCompleteTx(store, tx, tenantID, data)
+		return applyUploadCompleteTx(ctx, store, tx, tenantID, data)
 
 	case "llm_cost_record":
 		var data llmCostMutationData
--- a/pkg/backend/quota_cache.go
+++ b/pkg/backend/quota_cache.go
@@ -301,21 +301,28 @@ type quotaPendingDeltasSnapshot struct {
 	expiresAt time.Time
 }
 
+type quotaPendingDeltaEntry struct {
+	deltas    quotaPendingDeltas
+	expiresAt time.Time
+}
+
 type quotaPendingDeltasLoader func(context.Context) (storageDelta, fileDelta, mediaDelta int64, err error)
 
 // quotaPendingDeltasCache tracks this process's central quota mutations that
 // have been logged but not yet applied. It deliberately has no tenant-DB
 // fallback: runtime quota admission must not SUM quota_outbox from the user DB.
 type quotaPendingDeltasCache struct {
-	tenantID string
-	load     quotaPendingDeltasLoader
-	ttl      time.Duration
+	tenantID   string
+	load       quotaPendingDeltasLoader
+	ttl        time.Duration
+	pendingTTL time.Duration
 
 	mu                  sync.RWMutex
 	snapshot            *quotaPendingDeltasSnapshot
 	generation          uint64
 	localDeltas         quotaPendingDeltas
 	localPositiveDeltas quotaPendingDeltas
+	localEntries        []quotaPendingDeltaEntry
 	loadMu              sync.Mutex
 }
 
@@ -323,7 +330,15 @@ func newQuotaPendingDeltasCache(tenantID string, load quotaPendingDeltasLoader, 
 	if ttl <= 0 {
 		ttl = quotaPendingDeltasCacheTTL
 	}
-	return &quotaPendingDeltasCache{tenantID: tenantID, load: load, ttl: ttl}
+	return &quotaPendingDeltasCache{tenantID: tenantID, load: load, ttl: ttl, pendingTTL: localPendingMutationTTL()}
+}
+
+func localPendingMutationTTL() time.Duration {
+	ttl := replayMinAge() + replayPollInterval() + quotaPendingDeltasCacheTTL
+	if ttl <= 0 {
+		return defaultReplayMinAge + defaultReplayPollInterval + defaultQuotaPendingDeltasCacheTTL
+	}
+	return ttl
 }
 
 func (c *quotaPendingDeltasCache) get(ctx context.Context) (quotaPendingDeltas, bool) {
@@ -331,11 +346,15 @@ func (c *quotaPendingDeltasCache) get(ctx context.Context) (quotaPendingDeltas, 
 		return quotaPendingDeltas{}, false
 	}
 	if c.load == nil {
-		c.mu.RLock()
-		defer c.mu.RUnlock()
+		c.mu.Lock()
+		defer c.mu.Unlock()
+		c.expireLocalDeltasLocked(time.Now())
 		return c.localDeltas, true
 	}
 	now := time.Now()
+	c.mu.Lock()
+	c.expireLocalDeltasLocked(now)
+	c.mu.Unlock()
 	if deltas, ok := c.cached(now); ok {
 		return deltas, true
 	}
@@ -397,32 +416,106 @@ func (c *quotaPendingDeltasCache) cached(now time.Time) (quotaPendingDeltas, boo
 	return quotaPendingDeltas{}, false
 }
 
-func (c *quotaPendingDeltasCache) add(storageDelta, fileDelta, mediaDelta int64) {
+func (c *quotaPendingDeltasCache) addPending(storageDelta, fileDelta, mediaDelta int64) {
 	if c == nil {
 		return
 	}
-	c.mu.Lock()
-	defer c.mu.Unlock()
-	c.generation++
-	c.localDeltas.add(quotaPendingDeltas{
+	deltas := quotaPendingDeltas{
 		storageDelta: storageDelta,
 		fileDelta:    fileDelta,
 		mediaDelta:   mediaDelta,
-	})
+	}
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	now := time.Now()
+	c.expireLocalDeltasLocked(now)
+	c.generation++
+	c.localDeltas.add(deltas)
 	c.localPositiveDeltas.add(quotaPendingDeltas{
 		storageDelta: maxInt64(storageDelta, 0),
 		fileDelta:    maxInt64(fileDelta, 0),
 		mediaDelta:   maxInt64(mediaDelta, 0),
 	})
+	if !deltas.zero() {
+		c.localEntries = append(c.localEntries, quotaPendingDeltaEntry{
+			deltas:    deltas,
+			expiresAt: now.Add(c.pendingTTL),
+		})
+	}
 	// If the snapshot is missing or expired, leave it cold. The no-loader
 	// runtime path reads localDeltas directly; the legacy loader path refreshes
 	// snapshots on demand.
-	if c.snapshot == nil || time.Now().After(c.snapshot.expiresAt) {
+	if c.snapshot == nil || now.After(c.snapshot.expiresAt) {
 		return
 	}
 	c.snapshot.deltas.storageDelta += storageDelta
 	c.snapshot.deltas.fileDelta += fileDelta
 	c.snapshot.deltas.mediaDelta += mediaDelta
+}
+
+func (c *quotaPendingDeltasCache) clearPending(storageDelta, fileDelta, mediaDelta int64) {
+	if c == nil {
+		return
+	}
+	deltas := quotaPendingDeltas{
+		storageDelta: storageDelta,
+		fileDelta:    fileDelta,
+		mediaDelta:   mediaDelta,
+	}
+	c.mu.Lock()
+	defer c.mu.Unlock()
+	now := time.Now()
+	c.expireLocalDeltasLocked(now)
+	c.generation++
+	if !c.removeLocalEntryLocked(deltas) {
+		return
+	}
+	c.localDeltas.add(quotaPendingDeltas{
+		storageDelta: -storageDelta,
+		fileDelta:    -fileDelta,
+		mediaDelta:   -mediaDelta,
+	})
+	if c.snapshot == nil || now.After(c.snapshot.expiresAt) {
+		return
+	}
+	c.snapshot.deltas.storageDelta -= storageDelta
+	c.snapshot.deltas.fileDelta -= fileDelta
+	c.snapshot.deltas.mediaDelta -= mediaDelta
+}
+
+func (c *quotaPendingDeltasCache) expireLocalDeltasLocked(now time.Time) {
+	if len(c.localEntries) == 0 {
+		return
+	}
+	kept := c.localEntries[:0]
+	for _, entry := range c.localEntries {
+		if now.Before(entry.expiresAt) {
+			kept = append(kept, entry)
+			continue
+		}
+		c.localDeltas.add(quotaPendingDeltas{
+			storageDelta: -entry.deltas.storageDelta,
+			fileDelta:    -entry.deltas.fileDelta,
+			mediaDelta:   -entry.deltas.mediaDelta,
+		})
+		if c.snapshot != nil && now.Before(c.snapshot.expiresAt) {
+			c.snapshot.deltas.storageDelta -= entry.deltas.storageDelta
+			c.snapshot.deltas.fileDelta -= entry.deltas.fileDelta
+			c.snapshot.deltas.mediaDelta -= entry.deltas.mediaDelta
+		}
+	}
+	c.localEntries = kept
+}
+
+func (c *quotaPendingDeltasCache) removeLocalEntryLocked(deltas quotaPendingDeltas) bool {
+	for i, entry := range c.localEntries {
+		if entry.deltas == deltas {
+			copy(c.localEntries[i:], c.localEntries[i+1:])
+			c.localEntries = c.localEntries[:len(c.localEntries)-1]
+			return true
+		}
+	}
+	return false
 }
 
 func (d *quotaPendingDeltas) add(other quotaPendingDeltas) {
@@ -437,6 +530,10 @@ func (d quotaPendingDeltas) sub(other quotaPendingDeltas) quotaPendingDeltas {
 		fileDelta:    d.fileDelta - other.fileDelta,
 		mediaDelta:   d.mediaDelta - other.mediaDelta,
 	}
+}
+
+func (d quotaPendingDeltas) zero() bool {
+	return d.storageDelta == 0 && d.fileDelta == 0 && d.mediaDelta == 0
 }
 
 func maxInt64(a, b int64) int64 {
--- a/pkg/backend/quota_cache.go
+++ b/pkg/backend/quota_cache.go
@@ -303,9 +303,9 @@ type quotaPendingDeltasSnapshot struct {
 
 type quotaPendingDeltasLoader func(context.Context) (storageDelta, fileDelta, mediaDelta int64, err error)
 
-// quotaPendingDeltasCache avoids SUM(quota_outbox) on every small write. It is
-// only used by soft admission checks; strict upload reservation still reads the
-// tenant DB directly.
+// quotaPendingDeltasCache tracks this process's central quota mutations that
+// have been logged but not yet applied. It deliberately has no tenant-DB
+// fallback: runtime quota admission must not SUM quota_outbox from the user DB.
 type quotaPendingDeltasCache struct {
 	tenantID string
 	load     quotaPendingDeltasLoader
@@ -314,6 +314,7 @@ type quotaPendingDeltasCache struct {
 	mu                  sync.RWMutex
 	snapshot            *quotaPendingDeltasSnapshot
 	generation          uint64
+	localDeltas         quotaPendingDeltas
 	localPositiveDeltas quotaPendingDeltas
 	loadMu              sync.Mutex
 }
@@ -326,8 +327,13 @@ func newQuotaPendingDeltasCache(tenantID string, load quotaPendingDeltasLoader, 
 }
 
 func (c *quotaPendingDeltasCache) get(ctx context.Context) (quotaPendingDeltas, bool) {
-	if c == nil || c.load == nil {
+	if c == nil {
 		return quotaPendingDeltas{}, false
+	}
+	if c.load == nil {
+		c.mu.RLock()
+		defer c.mu.RUnlock()
+		return c.localDeltas, true
 	}
 	now := time.Now()
 	if deltas, ok := c.cached(now); ok {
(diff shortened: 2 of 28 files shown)


srstack (commented):
Overall this faithfully implements the agreed design, and the two correctness points from planning both landed well: `logQuotaMutation` is no longer fail-open (returns `error`, all create/overwrite/upload callers propagate and fail the write), and `docs/quota-meta-runtime-cutover.md` documents the residual post-commit crash window honestly instead of claiming durable equivalence with the old outbox. The new tests (`assertNoTenantQuotaOutboxRows`, `TestCentralQuotaMutationLogInsertFailureSurfaces`) assert the right invariants.

Two real bugs I'd like resolved before merge, plus some follow-ups.

## 🔴 Bug 1 — in-memory pending delta leaks on inline apply failure

`logAndEnqueueMutation` adds `+delta` to `quotaPendingCache` after the log insert. `applyQuotaMutation` subtracts `-delta` **only on success** — the error branch `return`s without decrementing, leaving the entry `pending` for `MutationReplayWorker`.

But the replay worker runs off `NewMetaQuotaAdapter(s.meta)` (`server.go:667`) and has **no reference to this backend's `quotaPendingCache`**. So on any transient meta-DB apply error, the replay path eventually applies the mutation to the meta counters but never decrements the in-process pending delta. The no-loader cache only accumulates (`localDeltas.add`, no TTL, no reset), so every such error permanently inflates this tenant's pending total. Admission uses `usage + pending`, so the tenant drifts toward spurious `quota exceeded` rejections that only clear on process restart.

Fix: decrement pending regardless of apply outcome (meta counters are the source of truth; in-memory pending is just a short-lived optimistic reservation and shouldn't outlive the apply attempt) — e.g. a `defer` in `applyQuotaMutation`, or move the decrement before the error-branch `return`.

## 🟠 Bug 2 — `completing` reservations have no reclamation path (reserved_bytes leak)

Adding `completing` correctly stops the expiry sweep from aborting an in-flight finalize (the sweep still filters `status='active'` at `pkg/meta/quota.go:824,866`). But that cuts both ways: if the process crashes after `UpdateUploadReservationStatus(..., "completing")` but before `completeUploadReservation`, the row is stuck in `completing` forever — the sweep won't reclaim it and `reserved_bytes` stays deducted from the pool permanently. `abortUploadReservation` handles `completing` now, but it only runs on a returned `CompleteMultipartUpload` error, not on a crash.

Recommend the expiry sweep also reclaim `completing` rows past a (longer) deadline so leaked finalizes self-heal, rather than relying solely on manual backfill.

## 🟡 Worth addressing

- **Redundant tenant-DB read on the overwrite hot path.** The overwrite branches dropped the `if b.UseServerQuota()` shortcut that derived `deltaBytes` from the already-locked `currentMeta.SizeBytes`, in favor of generic `ensureStorageQuota(ctx, tx, path, newSize)` which calls `b.store.ConfirmedFileSizeByPathTx(tx, path)` — a second current-size lookup per overwrite when the caller already holds it (`pkg/backend/dat9.go:1193-1200`, `fs_layer_content.go:202-208`). Correct, but an avoidable extra round-trip on a hot path.
- **Verify the `active`-only aggregates.** `pkg/meta/quota.go:783,801` still sum `WHERE status='active'`. If any feed reserved-bytes reconciliation/admission, a `completing` row would be excluded and undercount reserved. Please confirm they're not on that path.
- **Unrelated pool change.** `defaultUserMaxOpenConns` 4→6 and `defaultUserConnMaxLifetime` 30s→45s (`pkg/mysqlutil/pool.go`) aren't quota-related and aren't mentioned in the PR body. If intentional (offsetting the removed outbox worker's connections), note it; otherwise split it out.

## Minor / nits

- `stopQuotaOutboxWorker` is now an empty no-op still called from `Close()` and skipped tests. Fine as a transitional shim; leave a note or follow-up to delete it once the skipped outbox tests go.
- Several outbox tests are `t.Skip`-ped rather than deleted — acceptable for one release to keep the diff reviewable, but track their removal (they exercise now-inert paths).
- `syncCentralLLMCostRecord` now ignores the returned error (`_ =`). Intentional (LLM cost isn't fail-closed like storage), but a one-line comment would prevent a future "why is this one different".

I'd block on Bug 1 (degrades into tenant-wide write rejection) and want an answer on Bug 2's reclamation story before merge. The rest are follow-ups.

srstack (commented):
Re-reviewed after `Harden meta quota cutover review paths`. Both blocking bugs are resolved correctly, and the nits were addressed. 👍

## Bug 1 — pending-delta leak: fixed
`applyQuotaMutation` now decrements pending in a `defer` (and invalidates the usage cache) on both success and failure, with the old success-only decrement removed. On inline-apply failure the entry stays `pending` for the replay worker while the in-memory delta is released — this converts the previous *permanent over-count → spurious rejections* into a *transient under-count → brief optimistic over-admit* until replay lands (~minAge+poll). That matches the agreed eventual-consistency/optimistic-admission semantics, and is the right call given the replay worker has no handle to per-backend caches. `TestCentralQuotaPendingDeltaClearsAfterInlineApplyFailure` pins it.

## Bug 2 — `completing` reservation leak: fixed
`ExpireActiveReservations` now sweeps `status IN ('active','completing')`, so a crash after mark-`completing` but before finalize settle self-heals at `expires_at` instead of leaking `reserved_bytes` forever. Since `expires_at` is set to `now+24h` and real finalizes take seconds, there's no risk of reclaiming a live in-flight upload. `TestCentralQuotaExpireReclaimsCompletingReservations` covers it.

## Other good changes here
- `abortUploadReservation` refactored to a single `AbortActiveReservationTx` (`SELECT ... FOR UPDATE` + guarded UPDATE + `RowsAffected`) — closes the prior GetUploadReservation→update TOCTOU and makes double-release impossible.
- `DRIVE9_QUOTA_SOURCE` now hard-errors at startup if set, so a stale config can't silently no-op.
- LLM cost record failure is logged/metriced instead of silently discarded.
- `replayPollInterval` guards `<= 0`.
- The unrelated `pkg/mysqlutil/pool.go` connection-count change was reverted out. Thanks.

## One thing worth a note (non-blocking)
`reserveUploadOnServer` now holds `b.mutationMu` across `ensureUploadReserveFitsPendingQuota` + `AtomicReserveAndInsertUpload` (a meta-DB round trip). This is a sound consistency fix (serializes the pending read against concurrent `logAndEnqueueMutation`), and `logAndEnqueueMutation` already holds `mutationMu` across its log insert, so the "DB call under mutationMu" pattern isn't new. But it does mean concurrent small-write mutation logging for the same tenant now briefly blocks behind an upload reserve's meta transaction. Uploads are low-QPS so this is almost certainly fine — just flagging it as a potential same-tenant contention point to watch under load. A one-line comment explaining why the lock spans the reserve would help the next reader.

LGTM to merge once CI is green.

mornyx (commented):
Review of PR #659 — Route quota accounting through meta runtime

Overall this is a well-scoped and well-documented cutover. The design doc in `docs/quota-meta-runtime-cutover.md` is clear about the residual crash window, the fail-open removal is explicit, and the `completing` status cleanly protects finalize from the expiry sweep. The migration of `MetaQuotaStore` method signatures to take `context.Context` is consistent across the interface, the adapter, the fake, and the store implementation. Tests were updated alongside the behavior change, including new ones for the meta-log-insert-failure surfacing and the completing-reservation expiry reclaim.

A few concerns worth addressing before merge:

1. `applyQuotaMutation` subtracts pending deltas unconditionally in a `defer`, even when `MarkMutationAppliedTx` fails inside the same tx (so the row stays `pending` for replay). For one cycle the in-process pending deltas drop to 0 while the mutation is still unapplied, which briefly under-counts pending usage and can over-admit. Consider subtracting only on the success path, or re-adding on failure.

2. `finalizeUpload` marks the reservation `completing` and then calls `s3.CompleteMultipartUpload`. If S3 complete fails, the function returns early without aborting the reservation, leaving it stuck in `completing` until the 24h expiry sweep reclaims it. The tenant DB write hasn't happened yet, so aborting the reservation here (releasing reserved_bytes) would be safe and consistent with the tenant-tx-failure path a few lines below. Worth handling.

3. `replayMinAge()` returns 0 when `DRIVE9_QUOTA_REPLAY_MIN_AGE_MS=0` (the `envDurationMS` helper treats 0 as a valid value, unlike `replayPollInterval()` which clamps `<= 0` to the default). A 0 min-age means replay can race the inline apply for mutations that are only milliseconds old, which defeats the purpose of the min-age guard. Either clamp 0 to the default in `replayMinAge()`, or document that 0 is an explicit "race the inline path" knob.

4. `reserveUploadOnServer` now holds `mutationMu` across `ensureUploadReserveFitsPendingQuota` + `AtomicReserveAndInsertUpload`. That meta call does a DB transaction (reserve + insert). Holding `mutationMu` across a meta DB transaction serializes all uploads for a tenant against all create/overwrite mutation logging on that backend. Given the new 1s default replay poll and the hot small-write path also taking `mutationMu` for the log insert, this is worth a throughput check; if upload initiate latency is a concern, consider checking pending deltas under `mutationMu` but releasing it before the atomic reserve (relying on the meta DB transaction for the reserve race instead).

Minor:
- `syncCentralLLMCostRecord` now returns an error, but the `llm_usage.go` call sites only log+metric on failure — consistent with the "surface but don't block" stance for background LLM cost recording, just noting it's not symmetric with the create/overwrite paths which return the error to the caller.
- `stopQuotaOutboxWorker` is now a no-op but kept for legacy tests; `notifyQuotaOutbox` and `processQuotaOutboxAvailable` remain. A short comment in `quota_outbox_worker.go` pointing to the cutover doc would help future readers separate "intentionally retained" from "accidentally left".

The cutover direction and the explicit residual-window acknowledgment are good. The comments above are refinements rather than blockers, except #1 and #2 which I'd want addressed to avoid counter skew / reservation leaks under failure.

pkg/backend/quota_mutation.go
```diff
@@ -76,18 +103,24 @@ func (b *Dat9Backend) logQuotaMutation(ctx context.Context, mutationType string,
...
 // applyQuotaMutation applies a previously-logged mutation and marks it as
 // applied. This is the async-safe half of the split mutation path — if it
 // fails or never runs, MutationReplayWorker picks up the pending log entry.
-func (b *Dat9Backend) applyQuotaMutation(ctx context.Context, mutationType string, logID int64, apply func(tx *sql.Tx) error) {
+func (b *Dat9Backend) applyQuotaMutation(ctx context.Context, mutationType string, logID int64, pending quotaPendingDeltas, apply func(context.Context, *sql.Tx) error) {
+	defer func() {
+		if b.quotaUsageCache != nil {
+			b.quotaUsageCache.invalidate()
+		}
+		b.addPendingCentralMutationDeltas(-pending.storageDelta, -pending.fileDelta, -pending.mediaDelta)
```
mornyx: **Pending deltas subtracted unconditionally on the defer path.**

`applyQuotaMutation` does `defer { ... b.addPendingCentralMutationDeltas(-pending...) }`, but the deferred subtraction runs even when `MarkMutationAppliedTx` failed inside the `InTx` (the apply tx rolls back, the log row stays `pending`, and `MutationReplayWorker` retries it). In that failure window the in-process pending deltas go to 0 for this mutation while it's still unapplied, so admission temporarily under-counts pending usage and can over-admit until replay re-applies.

Consider subtracting the pending delta only on the success branch (or re-adding on failure). This mirrors the old outbox path where the pending delta was only decremented after a successful ack.

pkg/backend/upload.go
```diff
@@ -985,6 +985,17 @@ func (b *Dat9Backend) finalizeUpload(ctx context.Context, upload *datastore.Uplo
...
+				zap.String("upload_id", uploadID),
+				zap.Error(err))
+			metrics.RecordTenantOperation(b.tenantID, "central_quota", "upload_mark_completing", "error", time.Since(start))
+			return fmt.Errorf("mark central quota upload completing: %w", err)
+		}
+		metrics.RecordTenantOperation(b.tenantID, "central_quota", "upload_mark_completing", "ok", 0)
+	}
 
 	completeMultipartStart := time.Now()
 	if err := b.s3.CompleteMultipartUpload(ctx, upload.S3Key, upload.S3UploadID, parts); err != nil {
```
mornyx, line 1001: **Reservation left in `completing` when S3 CompleteMultipartUpload fails.**

A few lines above, the reservation was marked `completing` to protect it from the expiry sweep. If `b.s3.CompleteMultipartUpload` fails here, the function returns early without calling `abortUploadReservation`. The tenant DB write hasn't happened yet, so the reservation is now stuck in `completing` until the 24h expiry sweep reclaims it — reserved_bytes stay counted against the tenant for up to 24h.

The tenant-tx-failure path below (`backend_finalize_upload_tx_failed`) correctly calls `abortUploadReservation`. Consider mirroring that here: on S3-complete failure, abort the reservation (release reserved_bytes) before returning.

pkg/backend/mutation_replay.go
```diff
@@ -13,12 +15,36 @@ import (
...
 
+func replayPollInterval() time.Duration {
+	d := envDurationMS("DRIVE9_QUOTA_REPLAY_POLL_MS", defaultReplayPollInterval)
+	if d <= 0 {
+		return defaultReplayPollInterval
+	}
+	return d
+}
+
+func replayMinAge() time.Duration {
```
mornyx, line 32: **`replayMinAge()` does not clamp 0.**

`replayPollInterval()` clamps `d <= 0` back to the default, but `replayMinAge()` returns `envDurationMS(..., default)` directly, and `envDurationMS` treats 0 as a valid value. So `DRIVE9_QUOTA_REPLAY_MIN_AGE_MS=0` yields a 0 min-age, meaning replay can race the inline apply for mutations only milliseconds old — exactly the race the min-age guard exists to prevent.

Either clamp 0 to the default here for symmetry with `replayPollInterval()`, or document that 0 is an explicit "race the inline path" debug knob.

pkg/backend/upload_reservation.go
```diff
@@ -62,42 +62,15 @@ func (b *Dat9Backend) reserveUploadOnServer(ctx context.Context, uploadID, targe
...
-	if err != nil && centralReserved {
-		logger.Warn(ctx, "central_quota_reserve_upload_tenant_lock_commit_failed",
-			zap.String("tenant_id", b.tenantID),
-			zap.String("upload_id", uploadID),
-			zap.Error(err))
-		metrics.RecordTenantOperation(b.tenantID, "central_quota", "reserve_upload", "ok_lock_commit_failed", time.Since(start))
-		return true, nil
+	// Hold mutationMu across the pending-delta check and atomic reserve so
+	// same-backend post-commit quota mutations cannot slip between them.
+	b.mutationMu.Lock()
```
mornyx, line 67: **`mutationMu` held across a meta DB transaction.**

`mutationMu` is now held across `ensureUploadReserveFitsPendingQuota` **and** `metaStore.AtomicReserveAndInsertUpload` (a meta DB transaction). This serializes every upload initiate for a tenant against every create/overwrite `logAndEnqueueMutation` on this backend (which also takes `mutationMu` for the log insert).

With the new 1s default replay poll and the small-write hot path sharing this mutex, this is worth a throughput check. If upload initiate latency matters, consider checking pending deltas under `mutationMu` but releasing it before the atomic reserve — relying on the meta DB reserve transaction itself to bound the race, since the pending-delta check is an optimistic admission guard, not a correctness invariant.

mornyx (commented):
Inline review notes (companion to the summary comment).

pkg/backend/upload_reservation.go
srstack: Keeping this lock across the pending-delta check and AtomicReserveAndInsertUpload intentionally: it serializes same-backend small-write mutation logging with upload reservation so the local pending snapshot cannot change between admission and reserve. Upload initiate is low-QPS compared with small writes, and the code now has a comment documenting the consistency/perf tradeoff. If this shows up in load testing we can revisit with a weaker optimistic check, but for this cutover I prefer the stricter local admission boundary.

mornyx (commented):
Round 2 review of PR #659 (head: 845db115)

The follow-up commit "Address quota cutover review followups" addresses the four concerns from round 1 cleanly:

1. ✅ Pending-delta subtraction on apply failure — fixed. `applyQuotaMutation` now subtracts `addPendingCentralMutationDeltas(-pending...)` only on the success path (after `InTx` returns nil), not in a defer. On failure the pending delta is retained so admission keeps counting it until `MutationReplayWorker` re-applies. The test was renamed/updated to `TestCentralQuotaPendingDeltaRetainedAfterInlineApplyFailure` and asserts deltas stay at (128, 1, 0) while the row remains `pending`. This is the correct semantics.

2. ✅ Reservation leaked in `completing` on S3-complete failure — handled. The failure branch now calls `UpdateUploadReservationStatus(..., "active")` to reset the reservation back to `active` (rather than aborting it), so the upload remains retryable and `reserved_bytes` stays correctly counted rather than leaking until the 24h expiry sweep. `resetErr` is logged + metric-recorded without masking the original S3 error. Good call to reset-to-active rather than abort, since the tenant write hasn't happened and the client can retry finalize.

3. ✅ `replayMinAge()` zero clamp — fixed. Now mirrors `replayPollInterval()` with `if d <= 0 { return defaultReplayMinAge }`.

4. ✅ `mutationMu` serialization on upload reserve — documented. Commit `2f49b819` added the rationale comment at `upload_reservation.go:64`. The serialization is retained by design (pending-delta check + atomic reserve atomicity vs. same-pod post-commit mutations); the comment makes the tradeoff explicit. Acceptable.

Bonus fix observed: `enqueueMutation` now takes and forwards `context.Context` (`fn(ctx)` instead of `fn(context.Background())` in the inline fallback and drain path), so apply work inherits the request/worker context rather than a fresh background context. Tests updated to the new signature. This is a nice correctness improvement (cancellation/tracing now flows through).

Remaining minor observations (non-blocking):

- The fake `UpdateUploadReservationStatus` (non-Tx) still has no status guard (`if r, ok := ...; ok { r.Status = status }`), while its Tx counterpart now guards `active||completing`. The real `pkg/meta/quota.go UpdateUploadReservationStatus` SQL guards `status IN ('active','completing')`. The fake's non-Tx variant is slightly looser than the real store — fine for the reset-to-active path (which goes from completing→active), but if any future test relies on the guard it could diverge. Not worth changing now.

- The `reset-active` metric label `upload_reset_active` is recorded under `central_quota` op with `ok`/`error` results. Worth confirming this shows up in the operational guardrail doc (`docs/quota-meta-runtime-cutover.md` lists `central_quota_mutation_log_insert_failed` but not the reset-active path). A one-line addition to the "Operational Guardrails" section would be consistent.

No new correctness concerns. The cutover looks ready — the round-1 issues are resolved with matching tests, and the context-propagation cleanup is a welcome side fix. Ship it after CI is green.

mornyx (approved):
LGTM — round-1 concerns resolved with matching tests, context propagation cleanup is a welcome side fix. The two minor notes (fake non-Tx status guard, reset-active metric in the ops doc) are non-blocking polish. Approving.

srstack (commented):
Re-reviewed after the three follow-up commits. The smaller items look good:

- `replayMinAge()` now guards `<= 0` (matches `replayPollInterval()`).
- `enqueueMutation` threads the caller `ctx` into the inline/test path instead of `context.Background()`.
- The `mutationMu`-across-reserve comment I asked for is there — thanks.
- On `CompleteMultipartUpload` failure the reservation is reset `completing → active`, so the expiry sweep keeps the normal 24h window rather than treating a still-live upload as immediately reclaimable. Good catch.
- Env-help wording and the fake `UpdateUploadReservationStatusTx` status guard are fine.

## One concern — the pending-delta-on-failure change looks like a regression

This push reverts the `defer`-based decrement from `Harden meta quota cutover review paths`: `applyQuotaMutation` now subtracts the pending delta **only on the success path**, and `TestCentralQuota...RetainedAfterInlineApplyFailure` asserts the delta is *kept* (`128,1,0`) after an inline apply failure.

I traced the resulting lifecycle and I think it reintroduces a permanent over-count:

1. write commits (tenant DB); `logAndEnqueueMutation` inserts log `status='pending'` and `addPending(+128)` → `localDeltas=128`.
2. inline apply fails → entry stays `pending`, `localDeltas` stays `128`. **Admission is correct here** (`meta_usage` not yet incremented + `128` pending). ✅ This is the window the change is protecting, and I agree it avoids the transient over-admit.
3. `MutationReplayWorker` later applies the entry → `meta_usage += 128`, marks applied. But replay runs off `NewMetaQuotaAdapter(s.meta)` and never touches this backend's `quotaPendingCache`, so `localDeltas` is **still 128**.
4. admission now computes `meta_usage(128) + localDeltas(128) = 256` → double-counted, **permanently**, and it compounds on every future apply failure until the process restarts.

I confirmed: replay/adapter don't reference the pending cache, and the no-loader `localDeltas` has no TTL and no reset site (only the 4 add-sites; none clear it). So unlike the previous commit — which self-healed to the correct value once replay landed — this version drifts upward without bound. That's the same "over-count → spurious quota rejections" failure I originally raised, just triggered after replay instead of never.

So the two variants trade off:
- **Decrement on failure** (previous commit): ~seconds of possible over-admit until replay lands, then self-heals to exactly correct.
- **Retain on failure** (this commit): no over-admit window, but permanent double-count after replay, compounding per failure.

For a long-lived backend, unbounded drift toward "can't write despite having space" seems worse than a bounded few-second over-admit — so I'd lean back toward the decrement version unless I'm missing a reconciliation path.

### Suggestion that gets both
Give `localDeltas` a short TTL (≥ `replayMinAge + replayPollInterval`) so a pending delta expires on its own. Then step 2's window stays protected (no over-admit), and step 4 can't drift because the stale delta ages out right around when `meta_usage` starts reflecting the applied mutation. That matches the "cache" semantics and needs no cross-worker cache handle. Alternatively, key the local pending by `logID` and drop it when the entry is observed applied during a usage-cache reload.

If there's already a reconciliation that resets `localDeltas` against meta and I missed it, point me at it and I'll withdraw this. Otherwise I'd treat it as blocking, since it degrades into tenant-wide false rejections over time.

srstack:
Addressed the srstack pending-delta concern in 5e239f1f.

The local pending cache now tracks no-loader pending deltas as TTL-bound entries instead of an unbounded aggregate. A failed inline apply keeps the local pending delta long enough to cover the replay window (`replayMinAge + replayPollInterval + pending-cache TTL`), but it expires if replay applies the mutation out-of-band and this backend never sees a success callback. Successful inline apply now calls `clearPending...`, which removes the matching local entry so it will not be subtracted again on TTL expiry.

Added tests for: retain-then-expire after inline apply failure, no-loader pending expiry, successful clear avoiding later double-subtract, and late clear after expiry not going negative.

Local validation:
- `go test ./pkg/backend -run 'TestQuotaPendingDeltasCache|TestCentralQuotaPendingDelta|TestCentralQuotaExpireReclaimsCompletingReservations|TestCentralQuotaUploadCompleteUpdatesShadowState' -count=1`\n- `go test ./pkg/backend -count=1`\n- `make lint`

srstack (commented):
Re-reviewed `Expire local quota pending deltas`. This resolves my blocking concern — and the implementation handles the add/clear/expire race more carefully than a naive TTL would. LGTM.

## Why this fixes the drift
`applyQuotaMutation` success now calls `clearPendingCentralMutationDeltas`, and every local pending entry also carries `expiresAt = now + (replayMinAge + replayPollInterval + quotaPendingDeltasCacheTTL)`. So a delta that never gets an explicit clear (inline apply failed → replay worker applies it off a separate adapter that can't see this cache) now ages out on its own, *after* the point where `meta_usage` already reflects the applied mutation. The permanent double-count from the previous commit is gone, and step-2's protective window (no over-admit before replay) is preserved.

## The correctness-critical part — clear vs. expiry can't double-subtract
I traced all three orderings and they're sound:
- **clear before expiry** (normal success): `removeLocalEntryLocked` drops the entry and subtracts once; the later expiry pass no longer finds it. ✅
- **expiry before clear** (apply failed, replay landed later): `expireLocalDeltasLocked` subtracts and removes; a subsequent `clearPending` calls `removeLocalEntryLocked`, finds nothing, and — crucially — only subtracts *inside* the `if removed` guard, so it's a no-op. No under-count / over-admit swing. ✅
- **TTL alone** (no-loader runtime path): `get()` now runs `expireLocalDeltasLocked` under a write lock before returning `localDeltas`. ✅

All three are pinned by tests (`...ExpiresNoLoaderPending`, `...ClearPreventsExpiryDoubleSubtract`, `...ClearAfterExpiryDoesNotGoNegative`, and the extended `...RetainedThenExpiresAfterInlineApplyFailure`). The matching change in `recordAppliedQuotaOutboxEntries` keeps the legacy outbox path consistent with the new clear/expire semantics.

I ran `go test ./pkg/backend/ -run 'TestQuotaPendingDeltasCache|TestCentralQuotaPendingDelta|TestCentralQuotaExpireReclaims'` locally — green.

### Two tiny non-blocking notes
- `get()` on the no-loader path switched from `RLock` to `Lock` (needed, since expiry mutates). Admission now serializes on this mutex per tenant; fine at these QPS, just noting the read is no longer concurrent.
- `removeLocalEntryLocked` matches by exact delta-tuple equality, so two in-flight mutations with identical deltas are interchangeable for clear/expire purposes. That's harmless for the aggregate (the totals are what admission reads), just not entry-identity-precise — no action needed.

Nice iteration. From my side this is ready to merge once CI is green.

mornyx (approved):
Round 3 review of PR #659 (head: 3558775b)

Three new commits since the approved head (845db115). Verdict: the changes are correct and the round-2 minor notes are addressed; approving state is preserved. One latent (non-blocking) cleanup worth noting.

Commit c0dcb635 "Document quota upload reset guardrail"
- Adds `central_quota/upload_reset_active` and `central_quota_upload_reset_active_failed` to the Operational Guardrails section of the cutover doc. ✅ Addresses my round-2 minor note about the reset-active metric not being documented.
- Tightens the fake `UpdateUploadReservationStatus` (non-Tx) to guard `active||completing`, matching the real store SQL and the Tx variant. ✅ Addresses my round-2 minor note about the fake being looser than the real store.

Commit 5e239f1f "Expire local quota pending deltas"
This is the substantive change. It adds a bounded-TTL safety net so the in-process pending-delta cache cannot grow a stale delta forever when an inline apply silently never runs (e.g. worker dropped, process-level issue).

Design:
- `addPending` now appends a `quotaPendingDeltaEntry{deltas, expiresAt: now+pendingTTL}` in addition to updating the aggregate `localDeltas`.
- `clearPending` (new, replaces the old `addPending(-...)` on the success path) removes the matching entry by aggregate delta and subtracts from `localDeltas`. If no matching entry is found (already expired), it returns without subtracting — this correctly avoids the double-subtract that the old unconditional `-pending` would have caused.
- `expireLocalDeltasLocked` runs on every `get`/`addPending`/`clearPending` under the write lock, sweeping entries whose `expiresAt` has passed and subtracting them from `localDeltas` (and the live snapshot if present).
- `localPendingMutationTTL() = replayMinAge() + replayPollInterval() + quotaPendingDeltasCacheTTL`. This is the right composition: by the time a local entry expires, the replay worker (polling at replayPollInterval, with replayMinAge gate) plus one usage-cache TTL has had ample time to re-apply it, so dropping the local delta no longer under-counts.

Correctness checks:
- ✅ `clearPending` calls `expireLocalDeltasLocked` first, then `removeLocalEntryLocked`. If the entry already expired, expiry already subtracted it and `removeLocalEntryLocked` returns false → no double-subtract. Covered by `TestQuotaPendingDeltasCacheClearAfterExpiryDoesNotGoNegative`.
- ✅ `applyQuotaMutation` success path and `recordAppliedQuotaOutboxEntries` (legacy outbox) both switched to `clearPendingCentralMutationDeltas` (positive args) instead of `addPendingCentralMutationDeltas(-...)`. Consistent.
- ✅ `get` on the no-loader (production) path now takes the write lock (not RLock) because `expireLocalDeltasLocked` mutates state. The added comment explains this. Correct.
- ✅ `localEntries` is bounded by the number of in-flight unexpired mutations; successful applies clear entries, and expiry sweeps on access. No unbounded growth under normal operation.
- ✅ `addPending` skips the entry append when `deltas.zero()` (LLM cost records with empty pending), but still updates the aggregate — matches the old behavior.
- Tests are good: `TestQuotaPendingDeltasCacheExpiresNoLoaderPending`, `TestQuotaPendingDeltasCacheClearPreventsExpiryDoubleSubtract`, `TestQuotaPendingDeltasCacheClearAfterExpiryDoesNotGoNegative`, and `TestCentralQuotaPendingDeltaRetainedThenExpiresAfterInlineApplyFailure` cover the expiry/clear interaction including the retained-then-expired lifecycle.

Commit 3558775b "Document quota pending cache tradeoffs"
- Clarifying comments on the write-lock-need-for-read-path and the aggregate-delta matching rationale. Good.

One latent observation (non-blocking, no current impact):

`localPositiveDeltas` is incremented in `addPending` but never decremented in `clearPending` or `expireLocalDeltasLocked`. It is only read on the legacy loader path (`c.load != nil`) in the race-detection branch: `deltas.add(c.localPositiveDeltas.sub(localPositiveDeltas))`. Production wires `load: nil`, so this field is never read at runtime and the monotonic growth has no effect. The legacy loader tests each construct a fresh cache and call `addPending` once before triggering the race path, so the growth does not accumulate within a test. It is technically a leak in the legacy path if a single cache instance were reused across many `addPending` calls plus a loader race, but no current caller does that. Worth a one-line cleanup (decrement `localPositiveDeltas` in `clearPending`/`expireLocalDeltasLocked` by the positive parts of the removed entry) for symmetry, but not blocking.

No correctness or safety regressions. The pending-delta TTL is a sound safety net with the right composition, the clear-vs-expire interaction is provably free of double-subtract, and the doc/fake tightening closes the round-2 minors. Maintaining approval.

srstack:
Thanks for the Round 3 review. Agreed on the latent `localPositiveDeltas` cleanup: production uses the no-loader path so it had no runtime impact, but the legacy loader accounting should still stay symmetric.

I pushed `71c9e5da` to roll back `localPositiveDeltas` when a pending entry is cleared or expires, and added `TestQuotaPendingDeltasCacheRemovesPositiveRaceDeltasOnClearAndExpire` to cover both paths. Local validation: `go test ./pkg/backend -run TestQuotaPendingDeltasCache -count=1`.

pkg/backend/quota_cache.go
```diff
@@ -393,31 +418,110 @@ func (c *quotaPendingDeltasCache) cached(now time.Time) (quotaPendingDeltas, boo
...
 	defer c.mu.Unlock()
+	now := time.Now()
+	c.expireLocalDeltasLocked(now)
 	c.generation++
+	c.localDeltas.add(deltas)
 	c.localPositiveDeltas.add(quotaPendingDeltas{
 		storageDelta: maxInt64(storageDelta, 0),
 		fileDelta:    maxInt64(fileDelta, 0),
 		mediaDelta:   maxInt64(mediaDelta, 0),
 	})
```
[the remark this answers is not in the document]
srstack: Addressed in `71c9e5da`: `localPositiveDeltas` is now rolled back in both `clearPending` and `expireLocalDeltasLocked`, using the same positive component that `addPending` records. Added `TestQuotaPendingDeltasCacheRemovesPositiveRaceDeltasOnClearAndExpire` for clear and expiry paths.

mornyx:
Verified `71c9e5da` — the `localPositiveDeltas` symmetry is correct on both paths:

- `addPending(8,1,-1)` → `localPositiveDeltas += (8,1,0)`; `clearPending(8,1,-1)` → `localPositiveDeltas += -(8,1,0)` → 0. ✅
- `addPending(5,2,1)` → `(5,2,1)`; expiry sweep → `+= -(5,2,1)` → 0. ✅

`positivePart()`/`negate()` are clean helpers and `TestQuotaPendingDeltasCacheRemovesPositiveRaceDeltasOnClearAndExpire` covers both the clear and expire branches. No regressions — this closes the round-3 latent point. PR remains approved from my side.

srstack (commented):
Re-reviewed the two latest commits. Both good — and the second one is a genuine bug the author surfaced while documenting, nice catch.

## `Document quota pending cache tradeoffs`
The two why-comments I flagged are in place: the write-lock rationale on `get()`'s no-loader path, and the aggregate-vs-logID match rationale on `removeLocalEntryLocked`. Exactly the design boundaries worth pinning for the next reader.

## `Keep quota pending positive deltas symmetric` — correct fix
Before this, `localPositiveDeltas` was only ever *added to* (`addPending`); neither `clearPending` nor `expireLocalDeltasLocked` decremented it, so it grew monotonically. It's read only in the loader path's race-merge (`c.localPositiveDeltas.sub(localPositiveDeltas)` at the `generation != generation` branch), so an ever-growing value would have inflated the conservative snapshot over time — the same over-count → spurious-rejection class, on the loader path. Making `clearPending` and `expireLocalDeltasLocked` both apply `positivePart().negate()` keeps it symmetric with `localDeltas`, and `TestQuotaPendingDeltasCacheRemovesPositiveRaceDeltasOnClearAndExpire` pins it back to zero after both clear and expire.

I confirmed the runtime (no-loader) path never reads `localPositiveDeltas` — it's only the legacy/test loader path — so this is a correctness hardening there with no runtime behavior change. The extracted `positivePart()` / `negate()` helpers also remove the duplicated `maxInt64` expressions. Low risk.

Nothing further from me. Ready to merge once CI is green (mergeStateStatus is currently UNSTABLE — presumably checks still running).

PR #661, opened 2026-07-01
Remove tenant quota outbox and add replay observability

srstack:
## Summary
- remove the legacy tenant quota_outbox worker, datastore APIs, tests, and tenant schema creation
- keep runtime quota accounting on the meta quota_mutation_log path instead of tenant/user DB outbox rows
- expose per-tenant mutation replay backlog gauges for pending mutation count and oldest pending age
- add a pending-backlog aggregation index and configurable replay backlog observation cadence
- clear replay backlog gauges on graceful replay-worker stop while preserving last backlog values on fatal exit
- remove the completed one-time drive9-server backfill-quota command after reconciliation has run
- update quota docs and schema tests for the meta-only runtime path

## Tests
- go test ./pkg/backend ./pkg/meta ./pkg/tenant ./cmd/drive9-server -count=1
- go test ./pkg/server -count=1
- go test ./cmd/drive9-server ./pkg/backend ./pkg/datastore -count=1
- go test ./pkg/backend -run 'TestMutationReplayWorker|Test.*Quota|Test.*Central|Test.*Upload' -count=1
- go test ./cmd/drive9-server ./cmd/drive9-server-local ./pkg/backend -count=1
- make lint

## Related
- Runbook alert update: https://github.com/tidbcloud/runbooks/pull/2629


srstack (commented):
Reviewed the full diff. This is the natural close-out of the cutover: it deletes the now-dead tenant `quota_outbox` (worker, datastore layer, tenant schema, tests) that #659 left in place, and adds the replay-backlog observability that replaces the old outbox alerts. 3153 deletions / 280 additions, and I traced the runtime-facing pieces. LGTM.

## Removal is complete and consistent
- Worker (`quota_outbox_worker.go`), datastore (`datastore/quota_outbox.go`), and all four schema generators (`tidb_app`, `tidb_auto`, `mysql_no_embedding`, `db9`) drop `quota_outbox` + `quota_admission_locks` together; `ValidateMySQLNoEmbeddingTenantSchema` and `SanitizeForkRuntimeState` are updated in lockstep, and `schema_test.go` now asserts the dump does *not* contain the legacy tables. No dangling refs.
- `lockQuotaAdmissionTx` / `checkNoPendingCentralFileMutation`'s outbox branch are removed — consistent with runtime no longer touching the tenant DB for quota.
- `assertNoTenantQuotaOutboxRows` now guards on table existence first, so the retained "0 rows after cutover" invariant still compiles/passes against schemas that no longer create the table. Good touch.

## Backlog metrics — sound
- `ObservePendingMutations` groups pending `quota_mutation_log` by tenant for count + oldest-age, gated to `defaultReplayObserveEvery` (10s) via `recordPendingBacklogIfDue` so it doesn't run every 1s poll.
- The observe-vs-emit lifecycle correctly clears gauges to 0 both when a tenant drains (`current` set diff against `observedBacklogTenants`) and when the worker stops (`clearPendingBacklogGauges`), so alerts won't wedge on a stale nonzero value. Both paths are tested.

## Verified the one thing that could have bitten existing deployments
`ObservePendingMutations` uses `FORCE INDEX (idx_pending_tenant_age)`, a newly-added index. On an already-created `quota_mutation_log`, `CREATE TABLE IF NOT EXISTS` is a no-op — so `FORCE INDEX` on a missing index would be a hard error. But `Store` construction runs `migrate()` (meta.go:277) before serving, which diffs the parsed schema spec (new inline index included) and applies index-diff repairs, so the index is created on existing meta DBs ahead of any query. The new `idx_pending_tenant_age` also composes cleanly for the `WHERE status='pending' GROUP BY tenant_id` + `MIN(created_at)` access pattern. No migration gap.

Non-blocking:
- The observe cadence and the `1s` replay poll are independent; under a fast poll the 10s observe gate means backlog gauges can lag reality by up to ~10s. Fine for alerting, just noting it's not real-time.
- `MutationBacklogView` (backend) and `MutationBacklogObservation` (meta) are identical structs bridged by the adapter — expected given the backend↔meta dependency boundary, not a redundancy to fix.

Ready to merge once CI is green.

mornyx (approved):
Reviewed PR #661: Remove tenant quota outbox and add replay observability.

**Summary of review:**

This PR removes the legacy tenant-local `quota_outbox` / `quota_admission_locks` schema, datastore helpers, worker, tests, and the one-time `backfill-quota` CLI command. Runtime quota accounting is now fully on the central meta `quota_mutation_log` path. It also adds per-tenant replay backlog observability (pending count + oldest pending age gauges) with configurable observation cadence, and clears gauges on graceful stop while preserving last values on fatal exit.

**Verification performed:**

- `go vet ./pkg/backend/... ./pkg/meta/... ./pkg/tenant/... ./cmd/...` — passes
- `TestSchemaDumpInitSQLByProvider` — passes
- Searched for dangling references to removed `quota_outbox` types/functions (`QuotaOutbox`, `LockQuotaAdmissionTx`, `quotaOutboxNotify`, `claimQuotaOutbox`, `stopQuotaOutboxWorker`, etc.) in non-test production code — none found
- Codex (glm-5.2, xhigh) full-diff review — no blockers identified

**Key areas checked:**

1. **Schema removal consistency**: `quota_outbox` and `quota_admission_locks` removed from all tenant schema paths (`tidb_app`, `tidb_auto`, `mysql_no_embedding`, `db9`). The `mysql_no_embedding` validation query placeholder count (12 `?`s, 12 required tables, 12 args) is consistent. The schema dump test now asserts these tables are absent.
2. **`ObservePendingMutations` query**: Uses `FORCE INDEX (idx_pending_tenant_age)` which is a new index `(status, tenant_id, created_at)` added to `quota_mutation_log` in `metaInitSchemaStatements`. The index covers the query's `WHERE status='pending' GROUP BY tenant_id` with `MIN(created_at)`. Migration runs at `meta.Open` startup before the replay worker starts, so the index exists before `FORCE INDEX` is used.
3. **Replay worker concurrency**: `observedBacklogTenants` map is only accessed from the single `run` goroutine. `clearPendingBacklogGauges` is called in `run`'s defer (same goroutine) before `close(w.done)`. `Stop()` waits on `<-w.done`, so no concurrent access.
4. **Fatal vs graceful stop**: Fatal exit (`replayBatch` returns `true`) does not set `gracefulStop`, so gauges retain last non-zero values — correct for alerting. Graceful stop (`ctx.Done()`) clears gauges to zero — correct for clean shutdown.
5. **`SanitizeForkRuntimeState`**: Removed `DELETE FROM quota_outbox` and `DELETE FROM quota_admission_locks` statements. The remaining `DELETE` loop has `isMissingTableError` guard, so even if old tenants still have these tables, the removal of these lines is safe (they would have been no-ops for new tenants and skipped for missing tables).
6. **`file_gc_worker.go`**: Removed `HasPendingQuotaOutboxFileMutation` check. GC now only checks `metaStore.HasPendingFileMutation` (meta DB), which is the correct convergence backstop after outbox removal.
7. **Backfill command removal**: `cmd/drive9-server/backfill.go` and the `backfill-quota` subcommand fully removed. No remaining references in production code.
8. **Docs**: `quota.md` and `quota-meta-runtime-cutover.md` updated to reflect meta-only runtime path, new metrics names, and operational reconciliation guidance.

PR #664, opened 2026-07-02
feat: add recursive directory download to go and ts sdks

mornyx:
## Summary

Both the Go SDK (`pkg/client`) and the TypeScript SDK (`clients/drive9-js`) gain a single-call API to download an entire remote directory tree to a local directory — mirroring the CLI `drive9 fs cp -r` remote→local capability that previously lived only in the CLI layer.

### Go SDK — `pkg/client/download_dir.go` (new)
- `Client.DownloadDir(remoteDir, localDir)` / `DownloadDirCtx(ctx, remoteDir, localDir)`
- Algorithm (ported from CLI `copyTreeRemoteToLocal`):
  1. `StatCtx` — source must exist and be a directory.
  2. BFS walk via `ListCtx`, collecting dirs + files (symlinks rejected).
  3. `joinLocalSafe` pre-resolves every local path — rejects `..` and absolute segments to prevent path escape from malicious server listings.
  4. Destination root preflight (exists-as-file → reject).
  5. Descendant preflight via `os.Lstat` (symlinks count as conflicts) — aborts before any `os.Create` truncation.
  6. `os.MkdirAll` root + descendant dirs in parent-before-child order (empty dirs preserved).
  7. Bounded-parallel (16) `DownloadToFileWithSummary` per leaf — context cancellation honored at both top-of-loop and semaphore acquire; sibling failures collected.

### CLI refactor — `cmd/drive9/cli/cp_recursive.go`
- `copyTreeRemoteToLocal` now delegates to `Client.DownloadDirCtx`, eliminating ~100 lines of duplicated walk/preflight/mkdir/download pipeline.
- Removed dead helpers: `joinLocalSafe`, `preflightLocalDestinations`, `localDownloadPaths`, `downloadOneFile`, `localDownloadEntry`, `remoteDirEntry`, `remoteFileEntry`, and the `errors` import.
- `JoinLocalSafe` tests moved to the SDK test file.

### TS SDK — `clients/drive9-js/src/client.ts`
- `Client.downloadDir(remotePath, localPath)`
- Same algorithm: `stat` → BFS `list` → safe join → preflight → `mkdirSync` → bounded-parallel `downloadToFile` (Semaphore, concurrency 16).
- Helpers: `joinLocalSafe`, `preflightLocalDestinations`, `isNodeError`.
- Added `"downloadDir"` to `coveredClientMethods` and exercised in `filesystemCookbook` (satisfies cookbook-coverage test).

### Go — `pkg/client/download_dir_test.go` (new, 10 tests)
- `TestDownloadDir_HappyPath` — multi-level tree with files, nested dirs, empty dir.
- `TestDownloadDir_EmptyDir` — empty subdirectory preserved.
- `TestDownloadDir_DstExistsAsDir` — merge into pre-existing local dir.
- `TestDownloadDir_SourceIsFile` — file source rejected.
- `TestDownloadDir_DstExistsAsFile` — file-as-dst rejected.
- `TestDownloadDir_DescendantConflictRejects` — pre-existing file not truncated.
- `TestDownloadDir_ContextCancelled` — cancelled context aborts promptly.
- `TestJoinLocalSafe_*` — traversal/absolute/happy path.
- `TestWalkRemoteTreeBFS_HappyPath` — BFS visits all entries.
- `TestParallelDownload_*` — cancellation + single/multi error aggregation.

### TS — `clients/drive9-js/tests/download_dir.test.ts` (new, 5 tests)
- Happy path (nested tree + empty dir), file source reject, dst-as-file reject, descendant conflict, pre-existing-dir merge.

### Regression
- All existing CLI recursive-copy tests (`TestCpRecursiveRemoteToLocal_*`) pass unchanged.
- `make lint`: 0 issues.
- `make test` (pkg/client, cmd/drive9/cli): all pass.
- `vitest run`: 41/41 TS tests pass. `tsc --noEmit`: clean.

## Safety model
- **Path escape**: `joinLocalSafe` rejects `..` and absolute segments; post-join prefix boundary check.
- **No overwrite**: descendant preflight via `Lstat` aborts before any file write.
- **Symlinks**: rejected, not followed or skipped.
- **Context**: cancellation honored at both the top-of-loop check and the semaphore acquire.

## Summary by CodeRabbit

* **New Features**
  * Added `downloadDir` / `downloadDirCtx` to download an entire remote directory tree into a local folder (nested content and empty subfolders included).
  * Recursive CLI copy (`cp -r`) now uses the SDK directory-download flow end-to-end.
* **Bug Fixes**
  * Added hardened local path safety, destination conflict detection (no overwriting), and symlink entry rejection.
  * Improved concurrent transfer handling with clearer aggregated errors and context cancellation support.
* **Tests**
  * Added Go unit tests and expanded Vitest coverage.
  * Added an end-to-end recursive `cp -r` smoke test and updated cookbook examples.


mornyx:
## E2e coverage added

Added a `[4.1.1] cp -r recursive tree round-trip` section to `e2e/cli-smoke-test.sh` that runs on every PR-gated CI against a live server:

1. Build a local tree (nested dirs, multiple files, empty dir)
2. Upload via `cp -r` (local→remote)
3. Download back via `cp -r` (remote→local) — exercises the new `DownloadDirCtx` under the hood
4. Verify every leaf file round-tripped byte-identically
5. Verify empty and nested directories preserved
6. Verify remote listing shows uploaded entries
7. Remote→remote tree copy via `cp -r` with content checks

This gives the recursive directory download path live end-to-end coverage beyond the Go httptest unit tests and TS MSW mock tests.

qiffang (changes requested):
## Unified Review — PR #664: feat: add recursive directory download to Go and TS SDKs

**Reviewers:** @dat9-dev1, @adversary-2
**Verdict: REVISE** — one blocker, two warnings

---

### B1 — JS SDK missing symlink rejection (BLOCKER)

Go SDK `DownloadDirCtx` explicitly checks `remoteInfoIsSymlink(info)` during BFS walk and rejects with an error. The JS SDK `downloadDir` method in `client.ts` has no equivalent check — it only branches on `e.isDir`.

A remote tree containing a symlink entry will be silently downloaded (or treated as a regular file) instead of being rejected.

**Impact:** Behavioral divergence between Go and JS SDKs on a security-relevant path. The docstring says "Symlinks in the remote tree are rejected" but the JS implementation doesn't enforce it.

**Fix:** Either add a symlink check in the JS SDK (if the `list()` response carries mode info), or explicitly document the limitation in the JS SDK docstring and remove the "symlinks rejected" claim.

---

### W1 — JS SDK synchronous I/O in async function (WARNING)

`preflightLocalDestinations` uses `fs.lstatSync` and `downloadDir` uses `fs.statSync`/`fs.mkdirSync` inside an async function body. For trees with many descendants, this blocks the Node.js event loop.

Not a correctness bug, but worth noting if the SDK is used in server-side contexts. Consider using `fs.promises.*` for the preflight pass, or document that `downloadDir` may block the event loop during the preflight phase.

---

### W2 — Partial-failure cleanup semantics undocumented (WARNING, both SDKs)

When one or more downloads fail in the parallel phase, successfully-downloaded files and created directories remain on disk. The caller receives only the first error. This is consistent with CLI behavior, but the docstring's "aborts the download before any file is written" language only covers the preflight phase — once parallel downloads start, partial writes can occur.

**Suggestion:** Add a sentence to both SDK docstrings: "If a download fails after the preflight phase, successfully-written files remain on disk; cleanup is the caller's responsibility."

---

### W3 — JS SDK BFS walk has no cancellation path (WARNING)

The Go SDK's `walkRemoteTreeBFS` accepts a `context.Context` and checks `ctx.Done()`. The JS SDK's BFS walk (the `while (queue.length > 0)` loop) has no `AbortSignal` or early-exit mechanism. For very large remote trees, a cancelled operation will still complete the walk before the error surfaces.

Non-blocking for P0, but note the divergence for future parity.

---

### Positives

- **Security:** `joinLocalSafe` path-escape defense is solid in both SDKs — rejects `..`, absolute paths, and verifies resolved path stays under base.
- **Preflight semantics:** Both SDKs correctly preflight ALL destinations before touching the filesystem, preventing truncation of existing files.
- **CLI refactor:** Clean delegation from CLI to SDK — ~200 lines removed, no behavior change.
- **Test coverage:** Go SDK has 10 unit tests (happy path, edge cases, cancellation, parallelism). JS SDK has 5 tests via msw mocking. E2E section 4.1.1 covers the full round-trip.
- **Empty dir preservation:** Both SDKs correctly create empty directories.

---

### Summary

| Issue | Severity | SDK | Action |
|-------|----------|-----|--------|
| B1 — symlink rejection missing | BLOCKER | JS | Fix or document |
| W1 — sync I/O in async | WARNING | JS | Consider async or document |
| W2 — partial-failure semantics | WARNING | Both | Docstring update |
| W3 — no cancellation in BFS | WARNING | JS | Note for future |

Please address B1 before merge. W1-W3 are non-blocking but would improve the PR.

mornyx:
Thanks for the thorough review @qiffang. All four items addressed in the latest push (commit 2a7808a):

**B1 (BLOCKER) — TS SDK symlink rejection:** Fixed. Added mode/hasMode fields to the FileInfo interface, updated list() to parse them from the server response, and added symlink detection in the downloadDir BFS walk — rejects entries where (mode & 0o170000) === 0o120000 (S_IFLNK), matching the Go SDK remoteInfoIsSymlink. Added a test case "rejects a symlink in the remote tree".

**W1 — Sync I/O in async function:** Documented in the downloadDir docstring that the preflight and mkdir phases use synchronous fs.*Sync calls which block the event loop.

**W2 — Partial-failure cleanup semantics:** Added to both Go and TS SDK docstrings: If a download fails after the preflight phase, successfully-written files and created directories remain on disk; cleanup is the caller responsibility.

**W3 — BFS cancellation divergence:** Added inline comment in downloadDir noting the known parity gap with Go SDK context.Context cancellation — the JS BFS loop has no AbortSignal early-exit mechanism, which is future work.

CI is green across all jobs.

qiffang (changes requested):
Verdict: REVISE.

I re-reviewed PR #664 independently at head `28cc276caf8380e783b9f41d908bf8ca338819ee`. Two blockers remain:

1. `clients/drive9-js/src/client.ts:1009-1012` rejects valid relative local destinations. `joined` is computed with `path.join(base, rel)`, but the base comparison uses `path.resolve(base)`. For a normal SDK call such as `client.downloadDir('/tree', 'downloaded')`, `joined` becomes `downloaded/a.txt` while `cleanBase` is an absolute path like `/cwd/downloaded`, so the safety check treats the valid child as escaping the base. Existing JS tests only use absolute temp directories, so this public API case is uncovered. Fix by comparing paths in the same coordinate system, and add a regression test using a relative `localPath` with at least one downloaded child.

2. Go files are not gofmt-clean. `/usr/local/go/bin/gofmt -l pkg/client/download_dir.go pkg/client/download_dir_test.go cmd/drive9/cli/cp_recursive.go cmd/drive9/cli/cp_recursive_test.go examples/go-sdk-cookbook/cookbook_test.go examples/go-sdk-download-dir/main.go` reports:
   - `pkg/client/download_dir.go`
   - `pkg/client/download_dir_test.go`
   - `examples/go-sdk-download-dir/main.go`

Validation run:
- `npm test -- --run tests/download_dir.test.ts` passed: 6 tests.
- `npm run lint` passed.
- `/usr/local/go/bin/go test -c ./pkg/client ./cmd/drive9/cli` passed compile.
- `/usr/local/go/bin/go test ./pkg/client ./cmd/drive9/cli` could not fully run locally because `pkg/client` TestMain panics on missing rootless Docker; `cmd/drive9/cli` passed.

The previous symlink blocker is fixed, but these two blockers need addressing before merge.

mornyx:
Thanks for the re-review @qiffang. Both blockers and CodeRabbit nitpicks addressed in commit d6081e0:

**BLOCKER 1 — TS joinLocalSafe rejects valid relative paths:** Fixed. The prefix check was comparing path.join(base, rel) (relative when base is relative) against path.resolve(base) (always absolute), causing false "escapes base" errors. Now both sides are resolved to absolute before comparison, while the returned path is still path.join(base, rel). Added regression test "works with a relative local destination path".

**BLOCKER 2 — Go files not gofmt-clean:** Fixed. Ran gofmt -w on all 3 files (trailing newline was missing). Verified gofmt -l reports clean.

**Nitpick — TS cookbook temp dir cleanup:** Fixed. Wrapped downloadDir in try/finally with fs.rmSync.

**Nitpick — Go SDK cookbook duplicate destination:** Fixed. DownloadDirCtx now uses ./downloads-ctx to avoid preflight conflict.

CI: TS 43/43 tests pass, Go lint 0 issues, gofmt clean.

qiffang (approved):
GREEN on re-review at `d6081e044`.

The prior blockers are fixed:

- JS `joinLocalSafe` now compares absolute resolved paths before the prefix check, so relative local destinations are accepted. The regression test covers the exact `downloadDir('/tree', 'relative-dir')` case.
- Changed Go files are gofmt-clean.

Validation run locally: `npm test -- --run tests/download_dir.test.ts` (7 passed), `npm run lint`, `gofmt -l` on changed Go files clean, `git diff --check origin/main...HEAD`, `go vet ./pkg/client/... ./examples/go-sdk-download-dir/...`, `go test -c ./pkg/client ./cmd/drive9/cli`, and `go test ./cmd/drive9/cli`. Full `go test ./pkg/client` is blocked locally by missing rootless Docker, same as prior review.

PR #665, opened 2026-07-02
provisioner: grant SYSTEM_VARIABLES_ADMIN to tdc_fs_admin role

srstack:
The system user created by `EnsureSystemUser` lacked `SYSTEM_VARIABLES_ADMIN`, causing `SET @@GLOBAL.TIDB_EXP_EMBED_OPENAI_API_KEY` to fail with Error 1227 during init schema on TiDB Serverless.

This adds `GRANT SYSTEM_VARIABLES_ADMIN ON *.* TO 'tdc_fs_admin'` so the system user can set TiDB global embedding variables.

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated database provisioning so the `tdc_fs_admin` role now receives the necessary permission to administer system variables.
  * This helps avoid privilege-related issues during setup and administration.

### DIFF
--- a/pkg/tenant/tidbcloudnative/provisioner.go
+++ b/pkg/tenant/tidbcloudnative/provisioner.go
@@ -1214,6 +1214,7 @@ func systemUserStatements(dbName, username, password string) []string {
 	return []string{
 		fmt.Sprintf("CREATE DATABASE IF NOT EXISTS %s", quoteIdent(dbName)),
 		fmt.Sprintf("CREATE ROLE IF NOT EXISTS %s", quoteString(roleName)),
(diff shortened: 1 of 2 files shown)


srstack (commented):
LGTM. Clean one-line fix, correctly placed after role creation and before table-level grants.

One caveat: this only applies to **new** tenant provisioning. Existing tenants already provisioned without this grant will still hit Error 1227 on next init schema. Those need a manual `GRANT SYSTEM_VARIABLES_ADMIN ON *.* TO 'tdc_fs_admin';` as root.

mornyx (approved):

PR #666, opened 2026-07-02
Remove TiDB Cloud Starter provider

srstack:
## Summary
- remove the `tidb_cloud_starter` provider and the old starter provisioner package
- keep Anonymous create/delete/fork flows working through `tidb_cloud_native` server default credentials
- update schema dump support, CLI region mode handling, tests, and docs to use `tidb_cloud_native` / `anonymous`

## Tests
- `go test ./cmd/drive9/cli ./cmd/drive9-server ./pkg/tenant/...`
- `go test ./pkg/server`

## Summary by CodeRabbit

* **New Features**
  * Added an “anonymous” default provisioning mode when TiDB Cloud credentials aren’t provided.
  * Added compatibility for legacy TiDB Cloud Starter tenants’ schema and lifecycle operations.
* **Bug Fixes**
  * Improved provider/mode selection for region listing, quotas, schema initialization/resume, fork handling, and tenant deletion.
  * Updated deletion flows to use default credentials when customer credentials are missing.
* **Documentation**
  * Updated CLI help and docs to remove `tidb_cloud_starter`, standardizing `db9`, `tidb_zero`, and `tidb_cloud_native` schema workflows.
* **Tests**
  * Updated schema/region/provisioning tests to reflect the new provider/mode behavior.


srstack (commented):
Reviewed the full diff plus the retained legacy path and its wiring. This is a clean, well-scoped deprecation: it removes `tidb_cloud_starter` as a *provisionable* provider while preserving delete/fork for tenant rows already persisted with it. The split is exactly right. A couple of non-blocking notes.

## Deprecation split is correct
- `NormalizeProvider` no longer accepts `tidb_cloud_starter` (no new provisioning), but `SmallInDB`, `UsesTiDBAutoEmbedding`, and the new `SupportsClusterDelete` still recognize `ProviderTiDBCloudStarterLegacy` so persisted rows keep their behavior. Good — a persisted starter tenant still reports the right storage/embedding semantics.
- `LegacyProvisioner.Provision` returns an error (disabled), while `Deprovision` (delete) and `CreateBranch`/`ProvisionBranch` (fork) are retained. That matches the PR's stated intent and the actual persisted-tenant needs.

## Routing verified end-to-end
I traced every provider switch that a persisted starter tenant hits:
- `provisionerForTenantProvider` (server.go:149) → `legacyStarterProvisioner`
- `forkProviderSupported` (fork.go:846) → gated on `legacyStarterProvisioner != nil`
- `deprovisionTenantCluster` (tenant_delete.go:193) → `legacyStarterProvisioner.(tenant.Deprovisioner)`
- delete admission (tenant_delete.go:64) → `SupportsClusterDelete`, which includes the legacy value
All consistent; no path routes a starter tenant to the native provisioner or a nil provisioner silently. `tidb_zero` delete is still correctly rejected upfront.

## No dangling references
The removed starter *provisioning* symbols (`starter.NewProvisioner...`) are gone; the only remaining `starter.*` refs are the retained `NewLegacyProvisionerFromEnv` / `LegacyEnvPresent`. The three live provisioner constructors (`tidbzero`, `tidbcloudnative`, `db9`) are untouched. Build surface looks clean.

## Non-blocking notes
1. **Partial legacy env degrades silently to a request-time failure.** In main.go:234-240, the legacy provisioner is only wired when `NewLegacyProvisionerFromEnv` succeeds; if the env is partially set (`LegacyEnvPresent` true but e.g. a bad URL), it only logs a warn and leaves `legacyStarterProvisioner == nil`. A persisted starter tenant's delete then fails at request time with "legacy starter provisioner does not support deprovision" (a 500-ish conflict), not at startup. Contrast with the native path, which `os.Exit(1)`s when it's "wanted." If any starter tenants still exist in a given environment, consider failing fast (or at least a louder startup error) when legacy env is present-but-broken, so this is caught at deploy rather than at a user's delete request. If starter tenants are already fully drained in all environments, this is moot — worth a one-line confirmation.
2. The `default` arm of `provisionerForTenantProvider` returns `s.provisioner` (native) for unknown providers. Harmless today given `NormalizeProvider` gates inputs, but a `default` that returned nil (or errored) would be more defensive against a future provider string slipping through. Minor.

Nothing blocking. The core concern for a deletion PR — that persisted `tidb_cloud_starter` tenants keep working for delete/fork — is handled correctly and I verified the wiring rather than trusting the description. LGTM once CI is green.

qiffang:
## Review — Remove TiDB Cloud Starter provider

Diff: 2659 lines, 32 files, +637/-830. Reviewed full diff.

### Architecture — PASS

Clean decomposition:
1. **New provisioning blocked**: `NormalizeProvider` rejects `tidb_cloud_starter`. New tenants cannot be starter.
2. **Legacy delete/fork preserved**: `LegacyStarterProvisioner` handles deprovision/branch ops for persisted `tidb_cloud_starter` rows via separate `legacyStarterProvisioner` field on `Server`.
3. **CLI migrated**: `RegionModeTiDBCloudStarter` → `RegionModeAnonymous` with `canonicalRegionMode()` bridge so old manifests still resolve.
4. **Provider dispatch centralized**: `provisionerForTenantProvider()` routes `tidb_cloud_starter` → `legacyStarterProvisioner`, everything else → `provisioner`.

### Checklist

| Gate | Verdict |
|------|---------|
| Edge cases covered? | PASS — legacy manifest compat, credential fallback, mixed-provider fork |
| API explicit? | PASS |
| Regression tests? | PASS — comprehensive |

### Key findings

- **`provisionerForTenantProvider` dispatch** — Correct routing. All fork/delete paths now call this instead of `s.provisioner` directly.
- **Fork credential fallback** — `resolveForkCredentialRequest` calls `resolveDefaultCredentials(s.provisioner)` when no credentials provided. Tests cover both default-available and no-default paths.
- **`forkProviderSupported` migrated to method** — Checks `provisionerForTenantProvider != nil` instead of hardcoded list.
- **`canonicalRegionMode` bridge** — `tidb_cloud_starter` in manifests transparently maps to `anonymous`. Tests cover legacy manifest default and duplicate detection.
- **Legacy provisioner scoped** — `Provision()` returns error. Only deprovision/branch ops remain.
- **URL path escaping** — `url.PathEscape` added to legacy provisioner API calls.
- **Docs** — All updated consistently.

### Warnings (non-blocking)

- **W1**: Server with only `LegacyStarterProvisioner` configured exposes fork/delete endpoints, but `provisionerForTenantProvider` returns nil for non-starter providers — intentional for legacy cleanup.
- **W2**: Native `cleanupForkTenantOnce` now uses `resolveDefaultCredentials` where it previously required explicit credentials — tested and intentional.

### Verdict: **APPROVE**

No blockers. Clean removal with proper legacy compatibility. Provider dispatch centralized, tests comprehensive.

🤖 Reviewed by dat9-dev1

qiffang:
## Review — REVISE

I reviewed the latest head `2e2f34ac580380229c821ac16660b2294968a69c`. The new-provider path is mostly clean, but I do not think the legacy compatibility layer is safe to merge yet.

### Blockers

1. **Legacy schema-init resume still bypasses provider-specific dispatch.**

`pkg/server/server.go:1055` / `pkg/server/server.go:1063` resumes any live tenant left in `provisioning` by passing `s.provisioner.InitSchema` into `schemaInitForTenant`:

```go
s.initTenantSchemaAsync(workerCtx, t.ID, dsn, t.Provider, s.schemaInitForTenant(t.ID, t.Provider, s.provisioner.InitSchema))
```

That is no longer safe after this PR because the server can now have a `LegacyStarterProvisioner` independently from the main `Provisioner`. In the legacy-only supported mode that the handlers now allow (`s.provisioner == nil && s.legacyStarterProvisioner != nil`), this path panics in the resume worker when a persisted `tidb_cloud_starter` tenant is still `provisioning` or a stale pending tenant gets promoted for schema init. If `s.provisioner` is non-nil but for a different provider, the fallback initializer is also the wrong provider.

This path needs to resolve the provider-specific provisioner (`s.provisionerForTenantProvider(t.Provider)`) before building the schema init function, and handle missing legacy provisioner explicitly instead of dereferencing `s.provisioner`.

2. **Legacy env parsing drops the historical Starter API URL.**

The removed starter provisioner used `DRIVE9_TIDBCLOUD_API_URL` + `DRIVE9_TIDBCLOUD_API_KEY` + `DRIVE9_TIDBCLOUD_API_SECRET` for exactly the API calls that the new legacy provisioner keeps. The new `pkg/tenant/starter/legacy.go:36` only reads `DRIVE9_TIDBCLOUD_NATIVE_API_URL` for the base URL, while only falling back for the secret.

That means an existing deployment that keeps the historical starter env in place after this rollout has `LegacyEnvPresent() == true`, logs a warning, and disables legacy delete/fork compatibility because the new URL var is absent. For a PR whose stated goal includes keeping persisted `tidb_cloud_starter` rows cleanable/forkable, the legacy constructor should either accept the historical API URL as a fallback or the PR needs an explicit migration contract/test that operators must set `DRIVE9_TIDBCLOUD_NATIVE_API_URL` before legacy cleanup is supported.

### Validation

- `/usr/local/go/bin/gofmt -l $(git diff --name-only --diff-filter=ACMR origin/main...HEAD -- '*.go')` passed.
- `/usr/local/go/bin/go test -c ./pkg/server ./pkg/tenant ./pkg/tenant/starter ./cmd/drive9/cli ./cmd/drive9-server` passed.
- `/usr/local/go/bin/go vet ./cmd/drive9/cli ./cmd/drive9-server ./pkg/tenant ./pkg/tenant/starter ./pkg/server` passed.
- `/usr/local/go/bin/go test ./cmd/drive9/cli ./cmd/drive9-server ./pkg/tenant ./pkg/tenant/starter -count=1`: CLI/server/starter packages passed; `pkg/tenant` runtime tests are blocked locally by `rootless Docker not found` from testcontainers.

qiffang:
## Dev Lead Review — REVISE

### Blocker

**B1: `startTenantSchemaInitResume` bypasses `provisionerForTenantProvider`**

`server.go:1046` hardcodes `s.provisioner.InitSchema` for the schema-init resume worker. The PR already applied the `provisionerForTenantProvider(t.Provider)` dispatch pattern to fork, delete, and auto-embedding paths, but missed this resume worker.

In a legacy-only deployment (`s.provisioner == nil`, only `legacyStarterProvisioner` set), a persisted `tidb_cloud_starter` tenant still in `provisioning` state will nil-deref on resume.

**Fix:** Same pattern as the other paths — resolve provisioner via `provisionerForTenantProvider(t.Provider)` and handle nil explicitly.

### Non-blocking

**W1 (env var URL rename):** `NewLegacyProvisionerFromEnv` reads `DRIVE9_TIDBCLOUD_NATIVE_API_URL` but the old starter used `DRIVE9_TIDBCLOUD_API_URL`. Deploy docs should note this rename. Not a code bug — the APIs may use different endpoints.

---

Review consensus: @adversary-1 GREEN (initial), @adversary-2 REVISE on B1, @dat9-dev2 REVISE on B1+B2. Unified: **REVISE on B1**.

mornyx (approved):
Reviewed PR #666: Remove TiDB Cloud Starter provider.

**Summary of review:**

This PR removes the `tidb_cloud_starter` provider from the runtime provisioning path while preserving delete/fork compatibility for persisted starter tenants. The old `Provisioner` struct is replaced by `LegacyProvisioner` (in `pkg/tenant/starter/legacy.go`) which supports `Deprovision`, `ProvisionBranch`, `CreateBranch`, `WaitForBranchActive`, `DeleteBranch`, and `InitSchema` — but `Provision` returns an error (provisioning disabled).

**Key design decisions verified:**

1. **`provisionerForTenantProvider`**: Routes `tidb_cloud_starter` (legacy) to `s.legacyStarterProvisioner`, and `tidb_cloud_native`/`tidb_zero`/`db9` to `s.provisioner`. All fork operations (create/wait/delete/branch) now go through this dispatcher instead of directly using `s.provisioner`.

2. **`forkProviderSupported`**: Now a method on `Server` that checks whether `provisionerForTenantProvider` returns non-nil for the given provider. Replaces the old free function that only checked provider string.

3. **`resolveForkCredentialRequest`**: Only resolves default credentials for `ProviderTiDBCloudNative` — legacy starter returns `nil, nil` (no credentials needed). The `cleanupForkTenantOnce` change from `credentialReq == nil → ErrCredentialsRequired` to calling `resolveForkCredentialRequest` is correct: it only resolves for native, and legacy starter's `DeleteBranch` works without credentials.

4. **`handleForkDelete` / `cleanupForkInline`**: Previously checked `t.Provider == ProviderTiDBCloudNative` explicitly; now uses `forkProviderSupported(t.Provider)` which covers both native and legacy. Correct generalization.

5. **`NormalizeProvider`**: No longer accepts `tidb_cloud_starter` for new provisioning. `ProviderTiDBCloudStarterLegacy` constant keeps the same string value ("tidb_cloud_starter") so persisted tenant rows still match.

6. **CLI changes**: `RegionModeTiDBCloudStarter` → `RegionModeAnonymous` with `canonicalRegionMode` mapping old `tidb_cloud_starter` → `anonymous` for backward compatibility with existing region manifests.

7. **Schema dump**: Removed `tidb_cloud_starter` from `schemadump.Statements` provider switch and usage string. Schema dump test now asserts legacy tables are absent.

8. **`LegacyProvisioner` env vars**: Uses `DRIVE9_TIDBCLOUD_NATIVE_API_URL` (shared with native), `DRIVE9_TIDBCLOUD_API_KEY`, and `DRIVE9_DAT9_TIDBCLOUD_API_SECRET` with fallback to legacy `DRIVE9_TIDBCLOUD_API_SECRET`. Validates HTTPS URL.

9. **`SanitizeForkRuntimeState`**: Already had `isMissingTableError` guard, so removing the `quota_outbox`/`quota_admission_locks` deletes (done in PR #661) is safe here.

**Verification performed:**
- `go vet ./pkg/server/... ./pkg/tenant/... ./cmd/...` — passes
- `go test ./cmd/drive9/cli` — passes (no Docker needed)
- Searched for dangling `ProviderTiDBCloudStarter` (non-legacy) references — none found
- Codex (glm-5.2, xhigh) ran but didn't produce review output due to GOPROXY=off env issue; no blockers identified in stdout

mornyx (changes requested):
Agreeing with @qiffang's B1 — re-verifying and requesting changes.

**B1: `startTenantSchemaInitResume` nil-derefs when `s.provisioner` is nil**

Confirmed at `pkg/server/server.go:1063`:

```go
s.initTenantSchemaAsync(workerCtx, t.ID, dsn, t.Provider, s.schemaInitForTenant(t.ID, t.Provider, s.provisioner.InitSchema))
```

This resume worker runs for any persisted tenant in `provisioning` state (line 989 calls it during `reconcilePendingTenant`). The PR now allows handlers to operate with `s.provisioner == nil && s.legacyStarterProvisioner != nil` (see `handleForkCreate` / `handleTenantDelete` guard changes). In that legacy-only mode, a persisted `tidb_cloud_starter` tenant still in `provisioning` will nil-deref `s.provisioner.InitSchema` on resume.

Note that the other call site at line 4938 (`startProvisionedTenantSchemaInit`) has a `s.provisioner == nil` guard at line 4933 and is safe — only the resume path at line 1063 is unguarded.

**Fix:** Resolve the provisioner via `s.provisionerForTenantProvider(t.Provider)` and handle nil explicitly, consistent with the fork/delete/schema-init-for-tenant paths already updated in this PR.

**W2: Legacy env var URL rename (non-blocking)**

The old starter provisioner used `DRIVE9_TIDBCLOUD_API_URL`; the new `LegacyProvisioner` reads `DRIVE9_TIDBCLOUD_NATIVE_API_URL`. Deployments with only the historical starter env vars will fail to construct the `LegacyProvisioner`, silently disabling legacy delete/fork. Recommend adding `DRIVE9_TIDBCLOUD_API_URL` as a fallback in `NewLegacyProvisionerFromEnv`, or documenting the migration requirement explicitly.

mornyx (approved):
Re-review of head 4bb912e3 — both blockers from previous round are fixed.

**B1 fixed: `startTenantSchemaInitResume` now uses `provisionerForTenantProvider` dispatch**

`pkg/server/server.go:1057-1063` now resolves `s.provisionerForTenantProvider(t.Provider)` and returns early with a warning log if nil, instead of dereferencing `s.provisioner.InitSchema`. This prevents the nil-deref panic in legacy-only deployments (`s.provisioner == nil`). The schema init call at line 1069 now uses `schemaProvisioner.InitSchema`.

New regression test `TestStartTenantSchemaInitResumeUsesLegacyProvisioner` covers the legacy-only path: creates a `tidb_cloud_starter` tenant in `provisioning` state, starts resume with only `LegacyStarterProvisioner` set, and verifies the tenant reaches `active`.

**W2 fixed: Legacy env var URL fallback added**

`NewLegacyProvisionerFromEnv` now falls back to `DRIVE9_TIDBCLOUD_API_URL` (the historical starter env var) when `DRIVE9_TIDBCLOUD_NATIVE_API_URL` is absent. Error messages updated to mention both. New test `TestNewLegacyProvisionerFromEnvFallsBackToHistoricalURL` covers the fallback path.

**Verification:**
- `go vet ./pkg/server/... ./pkg/tenant/...` — passes
- `go test ./cmd/drive9/cli` — passes
- `go test ./pkg/tenant/starter` — passes
- `go test ./cmd/drive9-server -run TestSchemaDumpInitSQLByProvider` — passes

No remaining blockers. Approving.

PR #671, opened 2026-07-04
feat: drive9 fs archive — download remote tree as tar.gz/zip with profile-based filtering

mornyx:
## Summary

Add `drive9 fs archive <remote:/dir> [<out>]` to download a remote directory tree as a single compressed archive (tar.gz by default, zip optional), with profile-based file filtering that mirrors what `mount` applies to `node_modules`/`.git`/`dist`/etc. The feature is now available across all three surfaces:

- **Go SDK**: a new `Client.ArchiveDir(ctx, remoteDir, w, opts)` method in `pkg/client` that streams a tar.gz/zip archive into any `io.Writer`.
- **TypeScript SDK**: `client.archive()` (streaming tar.gz `ReadableStream`) and `client.archiveToFile()` (tar.gz or zip to a file) in `clients/drive9-js`.
- **CLI**: `drive9 fs archive` — a thin flag-parsing wrapper over `client.ArchiveDir`.

This is distinct from PR #664 (`cp -r`): `cp -r` materializes a directory tree on disk; `archive` produces a single compressed file (streamable to stdout). Both share the same client-side walk/download primitives.

## Motivation

Agents and operators frequently need to materialize a remote drive9 tree as one portable artifact — for handoff to a downstream tool, for offline inspection, or for snapshotting a workspace without mounting FUSE. Until now the only bulk-download path was `drive9 fs cp -r <remote> <local>`, which writes a directory tree. `archive` writes a single compressed file instead, and it can be piped straight to stdout (`drive9 fs archive :/proj --stdout | tar -tzf -`).

### Command shape

`drive9 fs archive` is a new `fs` subcommand alongside `cp`/`cat`/`ls`, dispatched through the read-only client path in `runFS` (no upload-threshold warm needed). Flags may appear before or after the positional source/output — the CLI splits positionals from flags so users are not forced into a rigid ordering.

### Direction

Remote → local only. A remote→remote archive is meaningless (no local artifact), and local→remote archive is the existing `pack` command's job — `archive` deliberately does not overlap with `pack`.

### Format

- `tar.gz` (default): streamed leaf-by-leaf from `ReadStream` via `io.Copy` — no temp file, no full-tree buffering. Directory entries are emitted first so empty dirs survive extraction.
- `zip`: single-threaded write (zip requires sequential central-directory offsets) but downloads still prefetch in parallel.

### Filtering

The archive command exposes `--exclude` / `--include` (archive-domain vocabulary) rather than the mount profile's `--local-only` / `--remote-only`. The latter's "local layer vs remote layer" routing semantics read awkwardly in a bulk-download context — `--local-only` could be misread as "only fetch from local." `--exclude`/`--include` is the universal language in tar/rsync/zip/git-archive.

`--profile` is still supported and translates the profile sections internally:

```
exclude           = profile.[local]  + --exclude          (deduped)
include-override  = profile.[remote]                    (deduped; profile-only)
include-whitelist = --include                            (--profile does not participate)
```

Per-entry decision:
1. If `override` matches → include.
2. Else if `exclude` matches → drop.
3. Else if `--include` is non-empty and no include pattern matches → drop.
4. Otherwise → include.

The override mechanism is only populated from profiles (not from CLI flags) because a user who writes an `--exclude` and then wants to override it back is contradicting themselves and should just remove the exclude. Override earns its keep for preset rule sets — e.g. "coding-agent skips node_modules, but this project's `node_modules/.package-lock.json` must be archived."

Three pattern forms are supported (mirrored from the FUSE `LocalPolicy` matcher, now extracted to `pkg/pathfilter`):

| Form | Example | Meaning |
|------|---------|---------|
| `**/x/**` (or `**/x`) | `**/node_modules/**` | Path contains the `x` subpath at any depth |
| `prefix/**` | `dist/**` | Everything under the prefix |
| exact / glob | `*.log`, `go.mod` | `path.Match` glob + exact equality |

### Directory pruning

When a directory's relative path matches an exclude (and no override restores it), the matcher drops it before BFS enqueues its children, so excluded subtrees incur no extra `ListCtx` round-trips. This is what makes `--profile coding-agent` skip `node_modules/` without exploding the request count.

## Shared / reused primitives

- `walkRemoteTreeBFS` (now in `pkg/client/archive.go`) — client-driven BFS tree walk via `c.ListCtx`.
- Bounded parallel download (16) with sibling-continues-on-failure semantics, serialized writes via mutex.
- `loadProfileConfig` / `mergeProfileValues` (`cmd/drive9/cli/profile.go`) — profile loading and dedup-merge (CLI side).

## New package: `pkg/pathfilter`

The pattern matching engine is extracted from `pkg/fuse/local_policy.go` into a new `pkg/pathfilter` package so the CLI and SDK do not pull in `pkg/fuse`. The FUSE `LocalPolicy` is now a thin wrapper over `pathfilter.Pattern` / `Matcher` with zero behavior change. In particular, the runtime-path-whitespace-significance boundary (a runtime path like `/repo/.git ` must NOT match `**/.git/**`) is preserved: `pathfilter.Pattern.MatchCanonical` is used on the FUSE side so the runtime path is not re-canonicalized (which would trim whitespace), while `pathfilter.Pattern.Match` (used by archive) canonicalizes its input. The existing `TestLocalPolicyRuntimePathWhitespaceIsNotTrimmedIntoMatch` regression test guards this.

## Go SDK: `Client.ArchiveDir`

`pkg/client/archive.go` exposes:

```go
func (c *Client) ArchiveDir(ctx context.Context, remoteDir string, w io.Writer, opts ArchiveOptions) error

type ArchiveOptions struct {
    Format   ArchiveFormat   // "tar.gz" (default) or "zip"
    Exclude  []string
    Include  []string
    Override []string         // typically from a profile's [remote] rules
    Flat     bool
    Jobs     int              // default 16
}
```

The walk, bounded download, and tar/zip writing all live here — the CLI's `cmd/drive9/cli/archive.go` is now a thin flag-parsing + profile-translation wrapper over `ArchiveDir`, eliminating the duplicated walker/writer that the first revision of this PR carried. `pkg/client/archive_test.go` covers tar.gz basic tree, exclude, include whitelist, zip format, flat mode, invalid format, and invalid pattern.

## TypeScript SDK

`clients/drive9-js` mirrors the Go API:

```ts
client.archive("/proj", { exclude: ["**/node_modules/**"] })
  : Promise<ReadableStream<Uint8Array>>  // tar.gz stream

client.archiveToFile("/proj", "./proj.tar.gz", { profile: "coding-agent" })
  : Promise<void>  // tar.gz or zip to a file
```

The TS SDK ships a zero-runtime-dependency ustar tar writer and a minimal STORE-method zip writer so the package stays dependency-light. `clients/drive9-js/src/pathfilter.ts` re-implements the three pattern forms to keep Go/TS matching semantics aligned. The cookbook (`examples/cookbook.ts`) covers the new methods so the cookbook-coverage test stays green.

A bug in the TS zip EOCD argument order (centralStart and centralSize were swapped, producing a corrupt central directory) was fixed during e2e verification.

### Go examples

- `examples/go-sdk-cookbook/cookbook_test.go`: new `ExampleClient_archive` demonstrating `ArchiveDir` with tar.gz, zip, exclude, include, flat, and override; `ArchiveDir` registered in the `coveredClientMethods` map so `TestClientMethodExampleCoverage` stays green.
- `examples/go-sdk-basic/{main.go,main_test.go}`: the `run()` smoke test now archives the scratch directory as tar.gz and reports `archive_entries` in the summary; `TestRunUsesDrive9GoSDK` asserts the new line appears.

### TS example

- `clients/drive9-js/examples/archive.ts`: a standalone runnable example that seeds a remote tree and exercises tar.gz-to-file, stdout streaming, zip with `--profile coding-agent`, `--include` whitelist, and `--flat` mode. Run with `DRIVE9_SERVER=... DRIVE9_API_KEY=... npx tsx examples/archive.ts`.

## E2E CLI smoke test

`e2e/cli-smoke-test.sh` gains a new `[4.3] cli fs archive` section with 26 assertions covering:

| Sub-test | What it verifies |
|----------|-------------------|
| 4.3a plain tar.gz | archive contains README.md, node_modules/react/index.js, .git/HEAD, src/app.go |
| 4.3b `--exclude` | keeps README.md + src/app.go; drops node_modules and .git |
| 4.3c `--profile coding-agent` | drops node_modules + .git; keeps src/app.go |
| 4.3d `--include` whitelist | keeps README.md + src/app.go; drops node_modules + .git |
| 4.3e `--format zip` | file produced, `testzip()` valid, includes README.md, drops node_modules |
| 4.3f `--flat` | basenames only (app.go, util.go), no slash paths |
| 4.3g `--stdout` | pipe to `tar -tzf -` works; drops node_modules |
| 4.3h extract | extracted src/app.go content matches uploaded bytes |

The section uses the same isolated-`HOME` and `:$remote` path conventions as the rest of the script.

## Files

New:
- `pkg/pathfilter/pathfilter.go` + `_test.go`
- `pkg/client/archive.go` + `archive_test.go`
- `cmd/drive9/cli/archive.go` + `archive_test.go`
- `clients/drive9-js/src/archive.ts` + `pathfilter.ts` + `tests/archive.test.ts` + `tests/pathfilter.test.ts` + `examples/archive.ts`
- `docs/design/directory-archive-spec.md`

Modified:
- `pkg/fuse/local_policy.go` (thin wrapper over `pkg/pathfilter`)
- `cmd/drive9/main.go` (runFS dispatch + fsUsage)
- `cmd/drive9/visual_help.go` (Archive section + examples)
- `clients/drive9-js/src/client.ts` / `index.ts` / `examples/cookbook.ts`
- `examples/go-sdk-cookbook/cookbook_test.go` (ExampleClient_archive + coverage map)
- `examples/go-sdk-basic/main.go` + `main_test.go` (archive step + assertion)
- `e2e/cli-smoke-test.sh` (new `[4.3]` archive smoke section)
- `README.md` (archive usage example)

## Verification

All of the following pass:

- `make lint` (golangci-lint v2.5.0): **0 issues**.
- `go build ./...`: ✅
- `go test ./pkg/pathfilter/...`: ✅
- `go test ./pkg/client/... -run TestArchiveDir`: ✅
- `go test ./pkg/fuse/... -run 'TestLocalPolicy|TestPolicy|TestClassify|TestSQLite'`: ✅ (FUSE policy zero regression)
- `go test ./cmd/drive9/cli/... -run 'TestArchive|TestBuildArchiveOptions'`: ✅
- `go test ./examples/go-sdk-basic/...`: ✅
- `go test ./examples/go-sdk-cookbook/...`: ✅
- `cd clients/drive9-js && npm test`: **51/51 ✅**
- `cd clients/drive9-js && npm run lint` (tsc --noEmit): ✅

**E2E against a live local server** (`drive9-server-local` + TiDB container, `local-dev-key`):
ran the `[4.3]` archive smoke section standalone — **26/26 assertions passed**, verifying tar.gz + zip compression, exclude/include/profile/flat/stdout filtering, zip validity via `testzip()`, and extracted-content byte-for-byte match.

Note: `pkg/fuse` full suite has two pre-existing failures (`TestShouldContinueAfterWaitMountPermissionErrorWhenProbe*`) that reproduce on `main` without this change — they are environment-related (sandbox without FUSE mount privileges), not caused by this PR.

## Out of scope / future work

- Symlinks are currently archived as regular files (their target content via `ReadStream`), matching `cp -r`'s reject-don't-follow behavior. Emitting `TypeSymlink` headers would require a new client `ReadLink` primitive.
- Profile loading in the TS SDK is limited to the built-in `coding-agent` defaults (the TS SDK does not ship a profile-file loader). Callers needing custom profiles should pre-resolve the profile into `exclude`/`include` lists and pass them directly. A profile loader could be added later if there is demand.

## Summary by CodeRabbit

* **New Features**
  * Added `drive9 fs archive` to download remote directory trees as `tar.gz` (default) or `zip`, with stdout streaming, include/exclude filtering, profile presets, and `--flat` hierarchy flattening.
  * Added matching archive capabilities to the JavaScript and Go SDKs, plus runnable SDK examples and cookbook coverage.
* **Documentation**
  * Updated README and CLI help/visual help, and added an archive design specification with usage details.
* **Bug Fixes**
  * Improved filtering and traversal behavior (e.g., excluding subtrees while preserving needed structure and empty directories), including safer `--flat` handling for basename collisions.


qiffang (changes requested):
## Unified Review — PR #671: feat: drive9 fs archive — download remote tree as tar.gz/zip with profile-based filtering

**Reviewers:** @dat9-dev1, @adversary-2
**Verdict: REVISE** — one blocker (claimed behavior not implemented), one warning

---

### B1 — Directory pruning NOT implemented despite design doc claim (BLOCKER)

The PR description states:
> Directory pruning happens at BFS time: when a directory's relative path matches an exclude (and no override restores it), the matcher drops it before BFS enqueues its children, so excluded subtrees incur no extra ListCtx round-trips.

**Actual behavior in `pkg/client/archive.go`:**

`walkRemoteTreeBFS` **unconditionally** enqueues every directory child:
```go
if e.IsDir {
    queue = append(queue, childRel)  // always enqueues, no filter
}
```

The `collectArchiveTree` visitor checks `matcher.Match(rel)` and returns `nil` for excluded entries — but this only prevents the entry from being added to the result slice. The BFS walk still lists all children of excluded directories. For `--exclude "**/node_modules/**"` on a large JS project, this means the SDK issues ListCtx for every nested directory inside `node_modules/`, defeating the stated optimization.

**Fix:** The walk visitor needs a way to signal "don't enqueue children." Either:
1. Return a sentinel error (e.g. `filepath.SkipDir`) that `walkRemoteTreeBFS` interprets as "skip enqueue but continue walking siblings", or
2. Pass the matcher into `walkRemoteTreeBFS` directly so it can skip enqueue for excluded dirs.

This is a blocker because the claimed performance property (no extra round-trips for excluded subtrees) is the primary motivation for directory pruning — without it, archiving a real-world JS project with thousands of `node_modules` subdirectories will be slow.

---

### W1 — Zip `defer writeMu.Unlock()` inside parallel goroutine (WARNING)

In `archiveZip`:
```go
writeMu.Lock()
defer writeMu.Unlock()
out, err := zw.CreateHeader(hdr)
...
io.Copy(out, io.LimitReader(rc, e.size))
```

The `defer writeMu.Unlock()` defers to the **goroutine return**, not the function return. Since each goroutine handles one file, this effectively holds the lock for the entire file download + write (including the io.Copy from the ReadStream). The parallel download doesn't actually help here because the mutex is held during the network read (`io.Copy(out, io.LimitReader(rc, e.size))`).

Same pattern exists in `archiveTarGz` — the mutex is locked before `io.Copy(tw, io.LimitReader(rc, e.size))`, meaning network reads are serialized.

**Impact:** The `jobs=16` parallelism only applies to the BFS walk, not the actual archive writing. This is likely by design (tar/zip writers aren't concurrent), but the goroutine pattern suggests parallelism that doesn't exist for the write path.

Actually, looking more carefully at `archiveTarGz`: the ReadStream is opened and read BEFORE acquiring the lock... no wait, both the `ReadStream` open AND the `io.Copy` are inside the mutex. The `rc, err := c.ReadStream(ctx, e.remote)` is before `writeMu.Lock()` — OK so the download IS parallel, only the header+copy to tar writer is serialized. That's correct for tar.gz.

But for zip: `rc, err := c.ReadStream(ctx, e.remote)` is also before the lock, and `io.Copy(out, ...)` reads from `rc` (the network stream) into the zip writer while holding the lock. So the network read IS serialized for zip. This means zip archives don't benefit from parallel downloads — the body is streamed from network directly into the zip writer under lock.

**Suggestion:** For zip, buffer the file content before acquiring the lock (like tar.gz effectively does via `io.Copy` from the stream into the tar writer... actually both have the same issue). The tar.gz path also reads from the network stream under the mutex. Consider pre-buffering file content in the parallel phase, then only acquiring the lock for the write.

Non-blocking since the description says "zip: single-threaded write" — but tar.gz also has this pattern and doesn't mention it.

---

### Positives

- **Architecture:** Clean extraction of `pkg/pathfilter` from `pkg/fuse/local_policy.go` — shared matching engine without importing FUSE. Good separation.
- **Pattern semantics:** Three pattern forms are well-defined and consistent between Go and TS.
- **Security:** No path-escape risk since archive entries use relative paths from the BFS walk.
- **API design:** `ArchiveDir(ctx, dir, w, opts)` with `io.Writer` is composable — stdout, file, network all work.
- **TS SDK:** Zero-dependency ustar tar writer and minimal zip writer keeps the SDK light.
- **Test coverage:** Go SDK 7 tests, TS SDK tests, CLI tests, E2E 23 assertions covering all flag combinations.
- **FUSE regression:** Existing `TestLocalPolicyRuntimePathWhitespaceIsNotTrimmedIntoMatch` guards the extract.

---

### Summary

| Issue | Severity | Action |
|-------|----------|--------|
| B1 — dir pruning not implemented | BLOCKER | Implement SkipDir in walkRemoteTreeBFS |
| W1 — network reads serialized under mutex | WARNING | Document or pre-buffer |

Please address B1 before merge. The PR claims directory pruning as a design feature and documents it as a performance guarantee — it must actually work.

mornyx:
@qiffang thanks for the review — both items addressed in 9b9b5ff.

## B1 — directory pruning now implemented (blocker resolved)

You were right that the original `walkRemoteTreeBFS` unconditionally enqueued every directory, so the matcher only prevented entries from landing in the result slice — excluded subtrees were still listed in full, defeating the documented optimization.

**Fix (Go):** `pkg/client/archive.go` introduces a sentinel `errArchiveSkipDir` (mirrors `filepath.SkipDir`). The `collectArchiveTree` visitor now returns it when a directory fails the matcher, and `walkRemoteTreeBFS` interprets it as "skip enqueue but continue siblings":

```go
if err := visit(childRel, e); err != nil {
    if errors.Is(err, errArchiveSkipDir) {
        continue  // do not enqueue this directory's children
    }
    return err
}
if e.IsDir { queue = append(queue, childRel) }
```

So `--exclude '**/node_modules/**'` / `--profile coding-agent` now issues **zero** `ListCtx` calls inside `node_modules/` or `.git/`.

**Fix (TS):** `clients/drive9-js/src/archive.ts` mirrors this — the visit callback now returns `boolean` (`false` = skip children), and `collectArchiveTree` prunes excluded directories the same way.

**Test:** new `TestArchiveDirDirectoryPruningSkipsExcludedSubtree` seeds a tree with `/proj/node_modules/react/foo/bar.js`, archives with `--exclude '**/node_modules/**'`, and asserts the mock server received **0** `ListCtx` calls for `/proj/node_modules/react` and `/proj/node_modules/react/foo` (the mock now records list-call counts), plus no `node_modules` entries in the archive.

## W1 — network reads no longer serialized under the write mutex

You correctly flagged that `io.Copy(out, io.LimitReader(rc, e.size))` ran under `writeMu` for both tar.gz and zip, serializing the network read across the whole archive and leaving `jobs=16` parallelism applicable only to the BFS walk.

**Fix:** both `archiveTarGz` and `archiveZip` now `io.ReadAll` the body into a buffer **before** acquiring `writeMu`, so the download runs in parallel across workers and only the header+body write is serialized (the tar/zip writers are not concurrency-safe). Named return values let the deferred `tw.Close()` / `gw.Close()` / `zw.Close()` propagate finalization errors instead of discarding them via `_ = Close()` — a full disk or broken pipe now returns non-nil rather than leaving an invalid archive.

Also addressed the overlapping codex/coderabbit nits in the same commit: context-qualified CLI sources (`prod:/proj`), `--flat` duplicate-basename rejection, zip dir mode no longer ORs `0o750`, TS `parallelMap` clamps non-positive `jobs`, and the TS gzip pipeline now drains `gzip.readable` concurrently with the write phase to avoid Web Streams backpressure deadlock.

All green: `make lint` 0 issues, `go test ./pkg/client/... -run TestArchiveDir` (incl. new pruning + collision tests), `npm test` 51/51, and the e2e archive section re-run against a local server.

qiffang (changes requested):
## Re-review: CONDITIONAL GREEN — B1/W1 fixed, 1 new blocker

### B1 (original BLOCKER): Directory pruning — FIXED ✅

`walkRemoteTreeBFS` now correctly uses `errArchiveSkipDir` sentinel. When visitor returns it, BFS loop `continue` skips enqueueing children (`archive.go:327-329`). `TestArchiveDirDirectoryPruningSkipsExcludedSubtree` verifies mock server `listCalls` = 0 for excluded subdirs.

### W1 (original WARNING): Network reads outside mutex — FIXED ✅

`archiveTarGz` and `archiveZip` now `io.ReadAll` file body into `[]byte` buffer before acquiring `writeMu`. Downloads fully parallel; only writes serialized.

### Close error propagation — FIXED ✅

Named return + defer close for tar/gzip/zip writers.

---

### B2 (NEW BLOCKER): include-only + directory pruning interaction bug

When only `--include` is set with patterns matching specific files (not directory prefixes), parent directories get pruned before their children are visited.

**Example:** `--include "src/main.go"` → BFS visits directory `src` → `matcher.Match("src")` returns `false` (because `src` doesn't match pattern `src/main.go`) → `errArchiveSkipDir` → entire `src` subtree pruned → `src/main.go` never visited → empty archive.

`pathfilter.go:178-181` already defines `HasInclude()` with a comment about this exact issue, but `archive.go` never calls it. **Fix:** When `matcher.HasInclude()` is true and no exclude matched, directories should NOT be pruned — continue walking and filter at the leaf (file) level only.

Existing `TestArchiveDirIncludeWhitelist` doesn't catch this because it uses `docs/**` (prefix form matches directory itself) and `README.md` (root-level file).

### W2 (WARNING): TS glob divergence

`pathfilter.ts:globMatch` only supports `*` wildcard; Go `path.Match` also supports `?` and `[abc]` character classes. Current patterns don't trigger this, but TS claims Go-parity.

### W3 (WARNING): TS missing zip and flat-collision test coverage

- No test for zip format in `archive.test.ts`
- No test for flat-mode basename collision error path

### Minor
- `archiveParallel` returns only `failures[0]`, silently drops subsequent errors
- CLI `--include` help says "excludes take precedence" — imprecise given override > exclude > include semantics
- E2E has 26 assertions, PR description says 23

mornyx:
@qiffang thanks for the conditional-green re-review — all three items addressed in 544bb54.

## B2 (blocker) — include-only + directory pruning interaction — FIXED ✅

You nailed this one. The original code drove pruning off `matcher.Match(rel)`, so when only `--include "src/main.go"` was set, the parent directory `"src"` failed the matcher (it doesn't match `src/main.go`) and got pruned via `errArchiveSkipDir` — the leaf was never visited and the archive came back empty. `pathfilter.Matcher` already exposed `HasInclude()` for exactly this case but `archive.go` never called it.

**Fix:** directory pruning is now driven by a new `Matcher.MatchExcluded(path)` predicate (in `pkg/pathfilter/pathfilter.go`) that returns `true` **only** when an exclude pattern drops the path AND no override restores it — never when the path merely fails the include whitelist. A directory that `MatchExcluded` returns `false` for is still walked through (so its children can match include patterns at the leaf), but only added to the archive's dir list when `Match()` accepts it. Leaves keep using `Match()` for the full include+exclude+override decision.

- `pkg/client/archive.go`: `collectArchiveTree` uses `MatchExcluded` for the dir-prune decision, `Match` for leaf filtering.
- `clients/drive9-js/src/archive.ts` + `pathfilter.ts`: mirrored (`matchExcluded` + rewritten `collectArchiveTree`).
- New `TestArchiveDirIncludeNestedFileNotPrunedByParentDir` (Go) and `"include nested file does not prune parent directory"` (TS) reproduce the bug: tree with `/proj/src/app.go`, `/proj/src/util/util.go`, `/proj/other/notes.txt`; `--include "src/app.go"` must archive `src/app.go` and drop `notes.txt`.
- E2E re-run against a local server: `drive9 fs archive :/RD /tmp/b2.tar.gz --include 'src/app.go'` now produces `src/app.go` (was empty before the fix).

`MatchExcluded` semantics are documented and covered by `TestMatchExcluded` (TS `matchExcluded` equivalent), including the "directory failing include is NOT MatchExcluded" case that was the root of B2.

## W2 — TS glob parity with Go path.Match — FIXED ✅

`pathfilter.ts` `globMatch` now supports `?` (single non-separator char) and character classes `[abc]` / `[^abc]` in addition to `*`, matching Go's `path.Match` exactly. New tests cover all three forms on both sides (`TestGlobQuestionAndCharClass` Go, `"pathfilter glob ? and [abc]"` TS), including `?` not crossing separators and negated classes.

## W3 — TS zip + flat-collision test coverage — FIXED ✅

`archive.test.ts` gains:
- `"zip format via archiveToFile"` — writes a zip, validates it via `python3 zipfile.testzip()` (asserts `valid`) and checks the namelist.
- `"flat mode rejects duplicate basenames"` — two `config.json` files in different dirs → `rejects.toThrow(/collision/)`.
- `"include nested file does not prune parent directory"` — the B2 regression case.

51 → 58 TS tests.

## Minor items — FIXED ✅

- `archiveParallel` now surfaces the failure count instead of silently dropping subsequent errors: `"<first err> (and N more error(s))"`.
- CLI `--include` help text corrected from `"excludes take precedence"` to `"override > exclude > include"` (the actual semantics).
- Re the e2e assertion count: I'll update the PR description to the final count (26) in the next pass.

All green: `make lint` 0 issues; `go test ./pkg/pathfilter/... ./pkg/client/... ./cmd/drive9/cli/... ./examples/...`; `npm test` 58/58; e2e archive section re-run against a local server (B2 fixed, all variants produce correct output, zip valid).

qiffang (changes requested):
REVISE.

Blocking:

1. TS tar headers still silently truncate archive entry names over 100 bytes. `clients/drive9-js/src/archive.ts` writes `name` with `writeString(header, 0, name, 100)`, and `writeString` copies only `min(encoded.length, maxLen)`. There is no PAX header, no ustar prefix field, and no explicit length error. `archiveName()` / `archiveDirName()` can produce arbitrarily long `root/rel` names, so long paths generate corrupt/truncated tar entries. Go uses `archive/tar`, so the TS SDK behavior diverges. Minimum acceptable fix: either support ustar/PAX long names or reject too-long tar entry names with a clear error and regression test.

2. Changed Go files are not gofmt-clean. `gofmt -l` reports `pkg/client/archive.go`, `pkg/client/archive_test.go`, `pkg/pathfilter/pathfilter.go`, `pkg/pathfilter/pathfilter_test.go`, `cmd/drive9/cli/archive.go`, and `cmd/drive9/cli/archive_test.go`.

Fixed from the prior review: include-only parent pruning is now corrected with `MatchExcluded`/`matchExcluded`, and the nested include regression exists on both Go and TS sides.

Validation run locally: TS archive/pathfilter tests pass, TS lint passes, `go test ./pkg/pathfilter ./cmd/drive9/cli` passes. `go test ./pkg/client` is blocked in this environment by missing rootless Docker.

mornyx:
@qiffang both blockers addressed in c69aee9.

## 1 — TS tar long-name truncation — FIXED ✅

You were right that `buildTarHeader` silently truncated names >100 bytes via `writeString`'s `Math.min` and emitted no PAX/ustar-prefix/length-error, diverging from Go's `archive/tar` (which handles long names natively).

**Fix:** `clients/drive9-js/src/archive.ts` now emits a PAX `'x'` extended header (typeflag `x`) carrying a single `path` record before any entry whose UTF-8 encoded name exceeds 100 bytes. The PAX record follows the standard `%d %s=%s\n` format where the length field counts itself; a new `paxRecord` helper solves the self-referential length by iterating to a fixed point. Both directories and files use this path; the regular entry's ustar name field is still written (truncated) per spec, but readers honor the PAX path. This mirrors what Go's `archive/tar` does for long names, so the two SDKs now produce equivalent archives for deep trees.

**Regression test:** new `"long entry names use a PAX header instead of truncating"` in `archive.test.ts` archives a >100-byte path and validates via `python tarfile` that the full name appears (the minimal ustar reader in the test file does not honor PAX, so python is used). 58 → 59 TS tests.

## 2 — gofmt-clean — FIXED ✅

Ran `gofmt -w` on all 6 changed Go files (`pkg/client/archive.go`, `pkg/client/archive_test.go`, `pkg/pathfilter/pathfilter.go`, `pkg/pathfilter/pathfilter_test.go`, `cmd/drive9/cli/archive.go`, `cmd/drive9/cli/archive_test.go`). `gofmt -l` now reports nothing for the changed tree.

Verified: `make lint` 0 issues; `gofmt -l` clean; `go test ./pkg/pathfilter/... ./pkg/client/... ./cmd/drive9/cli/... ./examples/...`; `npm test` 59/59.

qiffang (changes requested):
REVISE.

The previous ASCII long-path truncation issue is partially fixed, and the Go gofmt blocker is fixed. However the TS PAX writer is still not correct for UTF-8 paths, which Drive9 paths can contain.

Blocking:

1. `clients/drive9-js/src/archive.ts` builds PAX record lengths with JavaScript string length, not UTF-8 byte length. Lines 93-107 document `%d` as the byte length, but `suffix.length` counts UTF-16 code units. For a long non-ASCII archive path, `nameFitsUstar()` correctly routes to PAX based on UTF-8 bytes, but `paxRecord()` emits a wrong PAX length. I reproduced this locally with a throwaway test using a path like `proj/目录.../模块.ts`; Python `tarfile` listed only `proj`, not the long entry.

2. The PAX header size is set to the padded body length. `buildPAXLongNameEntry()` builds `body = record + pad` and then calls `buildTarHeader(..., body.length, "x", ...)` (lines 115-120). The tar header size for a PAX extended header should be the record payload length, excluding block padding. Padding still belongs on the wire, but it must not be counted as PAX data.

Required before GREEN: compute PAX record lengths in bytes, set the PAX header size to `record.length` while still writing `record + pad`, and add a regression with a non-ASCII >100-byte path parsed by a real tar reader (`python tarfile` or `tar -tzf`).

What is fixed:
- Go changed files are gofmt-clean.
- Existing TS archive/pathfilter tests pass, including the new ASCII long-name PAX regression.
- Include-only parent pruning remains fixed.

Validation run locally:
- `gofmt -l` on changed Go files: clean
- `git diff --check origin/main...HEAD`: clean
- `npm test -- --run tests/archive.test.ts tests/pathfilter.test.ts`: 23 passed
- `npm run lint`: passed
- `go test ./pkg/pathfilter ./cmd/drive9/cli`: passed
- `go test ./pkg/client -run 'TestArchiveDir|TestArchive'`: blocked by missing rootless Docker in this environment

mornyx:
@qiffang both PAX UTF-8 blockers fixed in 9098475, plus a non-ASCII regression test.

## 1 — PAX record length used UTF-16 code units, not UTF-8 bytes — FIXED ✅

`paxRecord` used `suffix.length` (JS string length = UTF-16 code units) for the `%d` length field, but PAX specifies byte length. For a non-ASCII path like `proj/目录.../模块.ts` this produced a length smaller than the actual UTF-8 byte count, so a real tar reader truncated the record and dropped the entry (I reproduced your `python tarfile` finding: it listed only `proj`).

**Fix:** added a `utf8Len(s)` helper that returns `new TextEncoder().encode(s).length`; the fixed-point length solve now uses byte counts throughout (`suffixBytes = utf8Len(suffix)`, and the `digits + suffixBytes` convergence loop runs on bytes). The encoded record is still produced via `new TextEncoder().encode(\`${len}${suffix}\`)`, so the wire bytes match the declared length exactly.

## 2 — PAX header size was the padded body length — FIXED ✅

`buildPAXLongNameEntry` built `body = record + pad` and passed `body.length` as the tar header `size` field. PAX 'x' header size must be the record payload byte length (padding is written on the wire but not counted as PAX data).

**Fix:** header size is now `record.length` (the unpadded payload bytes); the padded body is still emitted so the block boundary is correct. I verified by dumping the tar blocks: the PAX header's size field now equals the record byte count and the body is block-aligned.

## Regression test — non-ASCII >100-byte path

New `"long non-ASCII entry names survive via UTF-8-correct PAX header"` in `archive.test.ts` archives a >100-UTF-8-byte CJK path (`proj/目录/很深/嵌套/路径/确实/超过/ustar/一百/字节/名字/字段/限制/再加一点padding让它真的超过/模块.ts`) and validates via `python tarfile` that the full name appears. It also sanity-asserts `new TextEncoder().encode(name).length > 100` so the PAX path is genuinely exercised.

While adding the test I found and fixed a separate test-mock bug: the path lookup used the raw `url.pathname`, which arrives percent-encoded for non-ASCII segments (e.g. `/proj/%E7%9B%AE%E5%BD%95`), so the BFS walk stopped at the first non-ASCII directory because the decoded key wasn't found. The mock now `decodeURIComponent`s the pathname before the `dirs[p]` lookup. This was a test-infrastructure issue, not an archive-code bug, but it would have masked the PAX fix.

Verified: `npm run lint` clean; `npm test` 67/67 (incl. ASCII + non-ASCII long-name PAX regressions); `gofmt -l` clean; `make lint` 0 issues; `go test ./pkg/pathfilter ./pkg/client ./cmd/drive9/cli ./examples` all pass.

qiffang (approved):
GREEN after re-review at 90984759036d292987c1510412220f38de33fa8b. The prior TS PAX blocker is fixed: PAX record lengths are UTF-8 byte-counted, the PAX header size uses the unpadded record payload length, and the regression covers a non-ASCII >100-byte path parsed by Python tarfile. Local validation: npm archive/pathfilter tests 24 passed, npm run lint passed, gofmt check clean via /usr/local/go/bin/gofmt, git diff --check clean, go test ./pkg/pathfilter ./cmd/drive9/cli passed, go vet ./pkg/client ./pkg/pathfilter ./cmd/drive9/cli passed, and go test -c ./pkg/client ./cmd/drive9/cli ./pkg/pathfilter passed. Full pkg/client runtime tests remain locally blocked by missing rootless Docker, but GitHub ci and local-e2e are green.

PR #672, opened 2026-07-05
feat(sdk): one-click cross-SDK integration test suites for all 6 SDKs

mornyx:
## Summary

Adds a live-server integration test suite to **each of the 6 drive9 SDKs** (Go, TypeScript, Rust, Python, Kotlin, Swift) that exercises every exported function against a real `drive9-server-local`, plus a **one-click runner** that boots a disposable MySQL container, starts the server, points every SDK at it, runs all suites, and tears it all down on exit.

```bash
make sdk-integration-tests            # or: bash scripts/sdk-integration-tests.sh
# optional: --only go,ts   --keep-server   --port 19009   --no-build
```

## What the runner does (`scripts/sdk-integration-tests.sh`)

1. Probes toolchains; only requires toolchains for the SDKs it will actually run (`--only go,ts` to subset; default = all reachable).
2. Boots a **disposable MySQL 8 Docker container** (loopback-bound fixed host port `13306`, overridable via `DRIVE9_SDK_INTG_MYSQL_PORT`, auto-removed on exit) — no host MySQL/TiDB required.
3. `make build-server-local` → starts `drive9-server-local` on `127.0.0.1:19009` (avoids clobbering a developer's own 9009 server) with `DRIVE9_LOCAL_INIT_SCHEMA=true` (auto-create tables on the fresh DB) and `DRIVE9_LOCAL_EMBEDDING_MODE=none` (no Ollama dependency).
4. Waits for `/healthz`, exports `DRIVE9_SERVER` / `DRIVE9_API_KEY` pointing at the temp server.
5. Runs each enabled SDK's integration suite in turn, captures pass/fail.
6. Tears down: kill server PID, `docker stop` the MySQL container, remove the temp S3 dir.
7. Prints a summary table and exits non-zero if any suite failed.

The runner is **bash 3.2-compatible** (no associative arrays / `mapfile`), so it works on the macOS system bash as well as bash 4+.

## Per-SDK suites — full function coverage

Each SDK constructs its client via the SDK's **default-client constructor** (`Client.default` / `Client.defaultClient` / `Client::default_client`), which reads `DRIVE9_SERVER` + `DRIVE9_API_KEY` env vars first and `~/.drive9/config` second — so the real config-resolution code path is exercised end to end (the runner supplies env so the SDKs land on the disposable server).

Every exported function is exercised against the live server. Control-plane features the single-tenant local server does not enable (vault, scoped-token management, admin tenants, quota) are called but treated as **best-effort** (skip / return-early / log on the server's "not enabled" response), so the suites pass against `drive9-server-local`.

| SDK | File | Isolation from default test command | Coverage |
|---|---|---|---|
| Go | `pkg/client/integration_test.go` | `//go:build integration` build tag (default `make test` unaffected) | Full superset: FS / batch / search / SQL / transfer / streaming / patch / StreamWriter / FS Layers / journals / events / tokens / vault / quota / admin tenants / git workspaces / raw HTTP |
| TS | `clients/drive9-js/tests/integration.test.ts` | `describe.skipIf(!process.env.DRIVE9_INTEGRATION)` (default `npm test` unaffected) | All 103 `Client` methods + `StreamWriter` (manifest from `cookbook-coverage`) |
| Rust | `clients/drive9-rs/tests/integration.rs` | `#[ignore]` (run via `cargo test -- --ignored`) | FS / streaming (small + large multipart) / patch / vault / `StreamWriter` |
| Python | `clients/drive9-py/tests/test_integration.py` | skip-on-unreachable session fixture | All `Client` methods + `StreamWriter` + vault + patch + stream |
| Kotlin | `clients/drive9-kotlin/.../Drive9IntegrationTest.kt` | `Assumptions.assumeTrue(serverReachable)` | FS / search / SQL / `uploadFile`-`downloadFile` / `uploadFlow`-`downloadFlow` / `newStreamUpload` / `patchFileParts` / `resumeUpload` / vault |
| Swift | `clients/drive9-swift/.../Drive9IntegrationTests.swift` | `XCTSkipUnless` | FS / search / SQL / `uploadFile`-`downloadFile` / `uploadStream`-`downloadStream` / `newStreamUpload` / `patchFileParts` / `resumeUpload` / vault |

### Go testcontainers isolation

The existing testcontainers-backed Go unit tests (`client_test.go`, `transfer_test.go`, `schema_test_helper.go`, `testmain_test.go`) are guarded with `//go:build !integration` so the integration run does not double-provision a MySQL container. The integration `TestMain` reads `DRIVE9_SERVER`/`DRIVE9_API_KEY` and skips the whole package when the server is unreachable, so `go test -tags integration` is safe in any environment.

## SDK bugs surfaced and fixed by the integration suites

The suites found three real SDK bugs, fixed in this PR:

### Python `StreamWriter.abort()` deadlock
`abort()` acquired `threading.Lock` (line 174), then called `self._shutdown_executor()` which re-acquires the same non-reentrant lock → permanent hang. The factory test (`new_stream_writer` → `abort()` on a never-started writer) reproduced this deterministically. **Fix**: capture the executor + `upload_id` under the lock, then shutdown / call `_abort_upload_v2` **outside** the lock.

### Python `patch_file` `TypeError` on missing plan fields
`patch.py` did `plan.get("copied_parts", [])` — but the server can return `"copied_parts": null`, yielding `None`, and `len(None)` raised `TypeError: object of type 'NoneType' has no len()`. **Fix**: `plan.get(...) or []` for both `upload_parts` and `copied_parts`.

### Swift `downloadRangeStream` returned the full body for inline files
For a small (inline) file the server responds `200` with the whole body (not `206 Partial Content`); the SDK returned the full body as the "range" result instead of slicing out the requested `[offset, offset+length)`. **Fix**: when the response is `200` and the body is larger than the requested length, `subdata(in: start..<end)` locally.

## What does NOT change

- No changes to existing mocked/unit tests — all still pass:
  - Go: `go vet` clean (both `integration` and default builds compile).
  - TS: `npx vitest run` → 36 passed / 17 skipped.
  - Rust: `cargo test --test test_client` → 9 passed.
  - Python: `pytest tests/test_client.py tests/test_stream.py tests/test_vault.py` → 47 passed.
  - Swift: `swift test` → 13 passed / 10 skipped.
- No changes to server source code.
- Semantic/embedding-dependent assertions are opt-in (`DRIVE9_LOCAL_EMBEDDING_MODE=none` by default); `grep`/`find`/`sql` tests assert no-error + list-type, matching the leniency of the existing `test_e2e.py`.

## How to run

```bash
# from repo root, with docker available
make sdk-integration-tests
# or
bash scripts/sdk-integration-tests.sh
# subset
bash scripts/sdk-integration-tests.sh --only go,ts
# leave server up for debugging
bash scripts/sdk-integration-tests.sh --keep-server
```

## Test plan

- [x] `make sdk-integration-tests` passes end-to-end (go/ts/rs/py/swift PASS; kt SKIP without gradle).
- [x] Existing per-SDK mocked unit tests all still pass.
- [x] Both Go build tags (`integration` and default) compile; testcontainers tests still run under the default tag.
- [x] Runner is hermetic: disposable MySQL (loopback-bound, fixed port overridable) + temp S3 dir + fixed listen port (overridable via `--port`) + per-test prefix directories; no pollution of the developer's real server / config / DB.
- [x] Runner is resilient: graceful skip when the server is unreachable, so suites are safe to run in plain CI without the runner.

## Summary by CodeRabbit

* **New Features**
  * Added a `sdk-integration-tests` one-command workflow to run live, cross-SDK integration suites (Go, TypeScript, Rust, Python, Kotlin, Swift).
* **Tests**
  * Expanded end-to-end integration coverage across SDKs, including directory download and tar/zip archiving behaviors.
* **Bug Fixes**
  * Fixed streaming abort reliability and improved Python multipart/patch upload planning.
  * Corrected Swift range downloads for servers returning full responses; fixed redirect URL handling in the JavaScript client.
* **Chores**
  * Updated integration test gating and added/ignored integration runner virtualenv artifacts.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -124,9 +124,12 @@ e2e-local:
 # (Go, TypeScript, Rust, Python, Kotlin, Swift). Boots a disposable Docker MySQL
 # container, starts drive9-server-local, points each SDK at it via
 # DRIVE9_SERVER/DRIVE9_API_KEY, runs every suite, and tears it all down.
-# Pass extra args through to the runner, e.g.:
-#   make sdk-integration-tests -- --only go,ts
-#   make sdk-integration-tests -- --keep-server
+# Pass extra args through to the runner via SDK_INTEGRATION_ARGS, e.g.:
+#   make sdk-integration-tests
+#   make sdk-integration-tests SDK_INTEGRATION_ARGS="--only go,ts"
+#   make sdk-integration-tests SDK_INTEGRATION_ARGS="--keep-server"
+# (GNU make treats tokens after `--` as goals, so forward args through the
+#  SDK_INTEGRATION_ARGS variable rather than `make ... -- ...`.)
 sdk-integration-tests:
 	bash scripts/sdk-integration-tests.sh $(SDK_INTEGRATION_ARGS)
 
--- a/Makefile
+++ b/Makefile
@@ -47,7 +47,7 @@ BUILDINFO_LDFLAGS = -X github.com/mem9-ai/drive9/pkg/buildinfo.Version=$(if $(VE
 	-X github.com/mem9-ai/drive9/pkg/buildinfo.GitBranch=$(GIT_BRANCH) \
 	-X github.com/mem9-ai/drive9/pkg/buildinfo.BuildTime=$(BUILD_TIME)
 
-.PHONY: mod test test-failpoint test-podman fmt lint install-lint build build-server build-server-local build-cli build-cli-release run-server-local e2e-local docker-build
+.PHONY: mod test test-failpoint test-podman fmt lint install-lint build build-server build-server-local build-cli build-cli-release run-server-local e2e-local sdk-integration-tests docker-build
 
 mod:
 	$(GO) mod tidy
@@ -119,6 +119,16 @@ run-server-local: build-server-local
 
 e2e-local:
 	bash e2e/local-smoke.sh
+
+# Run the cross-SDK live-server integration suites for all drive9 SDKs
+# (Go, TypeScript, Rust, Python, Kotlin, Swift). Boots a disposable Docker MySQL
+# container, starts drive9-server-local, points each SDK at it via
+# DRIVE9_SERVER/DRIVE9_API_KEY, runs every suite, and tears it all down.
+# Pass extra args through to the runner, e.g.:
+#   make sdk-integration-tests -- --only go,ts
+#   make sdk-integration-tests -- --keep-server
+sdk-integration-tests:
+	bash scripts/sdk-integration-tests.sh $(SDK_INTEGRATION_ARGS)
 
 build-cli:
 	mkdir -p $(BIN_DIR)
--- a/clients/drive9-js/README.md
+++ b/clients/drive9-js/README.md
@@ -54,81 +54,215 @@ Expected config format:
 Environment variables `DRIVE9_SERVER` or `DRIVE9_BASE` and `DRIVE9_API_KEY` take priority over the config file.
 `DRIVE9_CONFIG` can point to an alternate config file.
 
-## Supported operations
+### Client lifecycle & tuning
 
 | Operation | Method |
 |-----------|--------|
-| Write file | `client.write(path, data)` |
-| Conditional write | `client.write(path, data, expectedRevision)` |
-| Read file | `client.read(path)` |
-| List directory | `client.list(path)` |
-| Stat | `client.stat(path)` |
-| Metadata stat | `client.statMetadataCompat(path)` |
-| Batch stat | `client.batchStat(paths)` |
-| Batch read-small | `client.batchReadSmall(paths, maxBytes)` |
-| Delete | `client.delete(path)` |
-| Recursive delete | `client.removeAll(path)` |
-| Copy | `client.copy(src, dst)` |
-| Rename | `client.rename(old, new)` |
-| Mkdir | `client.mkdir(path)` |
-| Chmod | `client.chmod(path, mode)` |
-| Symlink | `client.symlink(target, linkPath)` |
-| Hardlink | `client.hardlink(src, dst)` |
-| SQL query | `client.sql(query)` |
-| Grep | `client.grep(query, prefix, limit)` |
-| Layer grep | `client.grepWithLayer(query, prefix, limit, layerRef)` |
-| Find | `client.find(prefix, params)` |
+| Default client (env/config) | `Client.defaultClient()` |
+| Explicit constructor | `new Client(baseUrl?, apiKey?)` |
+| Small-file threshold builder | `client.withSmallFileThreshold(n)` |
+| Set `X-Dat9-Actor` header | `client.setActor(actor)` |
+| Base URL | `client.baseURL()` |
+| Auth + actor headers | `client.authHeaders(init?)` |
+| Build `/v1/fs/<path>` URL | `client.fsUrl(path)` |
+| Build `/v1/vault/<path>` URL | `client.vaultUrl(path)` |
+| Prime status cache (best-effort) | `await client.warm()` |
+| Tenant status | `await client.status()` |
+| Max single-upload bytes | `await client.maxUploadBytes()` |
+| Server inline threshold | `await client.smallFileThresholdValue()` |
+| Cached inline threshold | `client.cachedSmallFileThreshold()` |
+| Raw GET | `await client.rawGet(endpoint)` |
+| Raw POST | `await client.rawPost(endpoint, body?)` |
+| Raw DELETE | `await client.rawDelete(endpoint, body?)` |
+
+## Supported operations
+
+### Filesystem
+
+| Operation | Method |
+|-----------|--------|
+| Write file | `await client.write(path, data, options?)` |
+| Write returning revision | `await client.writeWithRevision(path, data, options?)` |
+| Read file | `await client.read(path)` |
+| Create empty file | `await client.createFile(path)` |
+| List directory | `await client.list(path)` |
+| Stat (HEAD) | `await client.stat(path)` |
+| Enriched stat | `await client.statMetadata(path)` |
+| Enriched stat with fallback | `await client.statMetadataCompat(path)` |
+| Batch stat | `await client.batchStat(paths)` |
+| Batch read-small | `await client.batchReadSmall(paths, maxBytes)` |
+| Delete | `await client.delete(path)` |
+| Delete file (kind hint) | `await client.deleteFile(path)` |
+| Delete dir (kind hint) | `await client.deleteDir(path)` |
+| Recursive delete | `await client.removeAll(path)` |
+| Copy | `await client.copy(src, dst)` |
+| Rename | `await client.rename(old, new)` |
+| Mkdir | `await client.mkdir(path, mode?)` |
+| Chmod | `await client.chmod(path, mode)` |
+| Symlink | `await client.symlink(target, linkPath)` |
+| Hardlink | `await client.hardlink(src, dst)` |
+
+### Search & SQL
+
+| Operation | Method |
+|-----------|--------|
+| SQL query | `await client.sql(query)` |
+| Grep | `await client.grep(query, prefix, limit)` |
+| Layer grep | `await client.grepWithLayer(query, prefix, limit, layerRef)` |
+| Find | `await client.find(prefix, params?)` |
 
 ### Streaming & multipart
 
 | Operation | Method |
 |-----------|--------|
-| Stream upload | `client.writeStream(path, stream, size, options?)` |
-| Stream upload summary | `client.writeStreamWithSummary(path, stream, size, options?)` |
-| Resume upload | `client.resumeUpload(path, stream, totalSize)` |
-| Read stream | `client.readStream(path)` |
-| Read range stream | `client.readStreamRange(path, offset, length)` |
-| Read range bytes | `client.readAt(path, offset, length)` |
-| Download to local file | `client.downloadToFile(remotePath, localPath)` |
-| Stream writer | `client.newStreamWriter(path, totalSize, options?)` |
-| Append | `client.append(path, data)` |
+| Stream upload | `await client.writeStream(path, stream, size, options?)` |
+| Stream upload with summary | `await client.writeStreamWithSummary(path, stream, size, options?)` |
+| Resume upload | `await client.resumeUpload(path, stream, totalSize)` |
+| Read stream | `await client.readStream(path)` |
+| Read range stream | `await client.readStreamRange(path, offset, length)` |
+| Read range bytes | `await client.readAt(path, offset, length)` |
+| Download to local file | `await client.downloadToFile(remotePath, localPath)` |
+| Stream writer | `client.newStreamWriter(path, totalSize, options?, abortOnError?)` |
+| Append bytes | `await client.append(path, data, options?)` |
+| Append stream | `await client.appendStream(path, stream, size, options?)` |
 
-`append` uses the native server append flow for existing S3-backed files. When
-native append is unavailable, the SDK only uses a bounded small-file rewrite and
-rejects large read/merge/write rewrites.
+`append` / `appendStream` use the native server append flow for existing S3-backed
+files. When native append is unavailable, the SDK only uses a bounded small-file
+rewrite and rejects large read/merge/write rewrites.
+
+### Download a directory tree
+
+```typescript
+// Recursively download a remote directory to a local directory.
+// Existing local destinations are never overwritten; a file source is rejected.
+await client.downloadDir("/project", "./local-project");
+```
+
+### Archive (tar.gz / zip)
+
+```typescript
+// Stream a remote directory as a tar.gz (default) or zip archive.
+const stream = await client.archive("/project", { format: "zip" });
+
+// Or write the archive straight to a local file.
+await client.archiveToFile("/project", "./project.tar.gz", {
+  exclude: ["**/node_modules/**"],
+  include: ["src/**"],
+  flat: false,
+  jobs: 16,
+});
+```
+
+`ArchiveOptions` supports `format` (`"tar.gz"` | `"zip"`), `exclude`/`include`/`profile`
+pattern filters (same three pattern forms as the Go CLI: `**/x/**`, `prefix/**`, name
+glob like `*.log`), `flat` (strip directory hierarchy), `jobs` (concurrent downloads),
+and `signal` (AbortSignal).
 
 ### Patch (partial update)
 
 ```typescript
-await client.patchFile(path, newSize, dirtyParts, readPartFn, progressFn, partSize?);
+await client.patchFile(path, newSize, dirtyParts, readPartFn, progressFn?, partSize?);
 ```
 
 ### Vault
 
 | Operation | Method |
 |-----------|--------|
-| Create secret | `client.createVaultSecret(name, fields)` |
-| Update secret | `client.updateVaultSecret(name, fields)` |
-| Delete secret | `client.deleteVaultSecret(name)` |
-| List secrets | `client.listVaultSecrets()` |
-| Issue token | `client.issueVaultToken(agentId, taskId, scope, ttlSeconds)` |
-| Revoke token | `client.revokeVaultToken(tokenId)` |
-| Issue grant | `client.issueVaultGrant(request)` |
-| Revoke grant | `client.revokeVaultGrant(grantId, request)` |
-| Read secret | `client.readVaultSecret(name)` |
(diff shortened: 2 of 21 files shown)


qiffang (changes requested):
## PR #672 Review — REVISE (1 blocker, 3 warnings)

### Bug Fixes — All 3 CORRECT ✅

1. **Python `patch.py` null coalesce** — `plan.get("copied_parts") or []` handles server returning `null`. Correct.
2. **Python `stream.py` abort() deadlock** — Lock released before `executor.shutdown()`. Correct.
3. **Swift `downloadRangeStream` 200→slice** — Local slice when server returns full body instead of 206. Bounds-checked. Correct.

### Test Isolation Guards — All CORRECT ✅

Go (`//go:build integration`), TS (`DRIVE9_INTEGRATION` env), Rust (`#[ignore]`), Python (session fixture skip), Kotlin (`Assumptions.assumeTrue`), Swift (`XCTSkipUnless`). Build tags mutually exclusive with `!integration` in existing tests.

---

### B1 (BLOCKER): Kotlin stat assertion uses stale size

`Drive9IntegrationTest.kt:123` — file is overwritten with `"v2"` (2 bytes) at line 114, but assertion at line 123 checks `data.size.toLong()` where `data` is the original 20-byte string. This will fail at runtime.

Fix: `assertEquals(2L, st.size)`

TS (line 142) and Swift (line 116) both correctly assert size == 2 after overwrite.

---

### W1 (WARNING): `wait_for_mysql` creates new Docker container per probe

`scripts/sdk-integration-tests.sh:120-122` — each of up to 90 retry iterations runs `docker run --rm mysql:8.0 mysqladmin ping`, spinning up a fresh container per attempt. Use `docker exec $MYSQL_CID` or `nc`-based port probe instead.

### W2 (WARNING): Python test cleanup uses wrong delete API

`test_integration.py:61` — `client.delete(p.rstrip("/") + "?recursive")` puts `?recursive` in the path, which gets URL-encoded as `%3Frecursive` rather than being a query parameter. Cleanup silently fails (caught by `except`), leaking test artifacts.

### W3 (WARNING): Makefile documents incorrect `--` usage

`Makefile:128-129` documents `make sdk-integration-tests -- --only go,ts` but Make passes `--` literally to the script, which rejects it as unknown flag. Should document `SDK_INTEGRATION_ARGS="--only go,ts"`.

---

### Minor / Non-blocking

- Kotlin `vault read` test (line 350) lacks `runCatching` guard — will throw instead of skip when vault disabled
- Kotlin `reachable` field is per-instance (JUnit 5 default), so every test probes the server redundantly
- Coverage delta: TS test is comprehensive (103 methods); Python/Rust cover core subset. Acceptable for P0 but worth documenting.
- Swift `downloadRangeStream` edge case: if file is exactly `length` bytes but `offset > 0`, status 200 returns wrong data (should return from offset). Minor since server should use 206 for range-capable files.

### Positives

- Runner is bash-3.2 compatible (macOS system bash)
- Hermetic: disposable MySQL + temp S3 + random prefix + proper EXIT trap cleanup
- Three real SDK bugs found and fixed by the test suites — validates the test investment
- Per-test isolation via prefix directories is sound
- `--keep-server` for debugging workflow is well thought out

qiffang (dismissed):
Reviewing head `a2873f35ae1098aaabf3c6d68e6fc48236ea8442`.

Verdict: REQUEST_CHANGES.

I agree with @dat9-dev1 that the overall direction is useful and the Python deadlock / Python patch null handling fixes are correct. I found additional issues that should be fixed before merge:

1. BLOCKER — Kotlin integration suite has a stale size assertion after overwriting the file.
   - `clients/drive9-kotlin/lib/src/test/kotlin/com/drive9/mobile/Drive9IntegrationTest.kt:113-124`
   - The test writes `"hello integration kt"`, then overwrites the same file with `"v2"` via `writeWithRevision(file, ..., -1)`, then asserts `stat(file).size == data.size`.
   - At that point the server-side file content is `"v2"`, so the expected size is 2. This makes the new Kotlin live integration suite fail on the happy path once it reaches this test.

2. BLOCKER — Kotlin vault read test is not actually best-effort when local vault is disabled.
   - `clients/drive9-kotlin/lib/src/test/kotlin/com/drive9/mobile/Drive9IntegrationTest.kt:346-363`
   - The PR states control-plane features disabled in `drive9-server-local` are treated as best-effort, and the Go/Swift/Python suites all guard vault create failures. This Kotlin test calls `c.createVaultSecret(...)` unguarded.
   - The runner does not set `DRIVE9_VAULT_MASTER_KEY`, so `drive9-server-local` does not enable vault by default. This test should use the same early-return `runCatching { createVaultSecret(...) }.getOrElse { return@runBlocking }` pattern as `vault management best-effort`.

3. BLOCKER — Swift `downloadRangeStream` still returns bytes before `offset` for valid inline range requests.
   - `clients/drive9-swift/Sources/Drive9Mobile/Drive9.swift:173-180`
   - The 200-response fallback only slices when `data.count > Int(length)`. That misses valid nonzero-offset requests where the requested length is >= the full inline body size. Example: 10-byte inline file, `offset=8`, `length=10`; the correct result is bytes 8..9, but this code leaves the full 10-byte body intact because `10 > 10` is false.
   - Same issue when `offset >= data.count` and the server returns 200/full body: the fallback leaves the full body instead of returning an empty range or matching the existing 416 handling semantics.
   - Fix by computing `start = clamp(offset, 0...data.count)` and `end = min(start + length, data.count)` for every 200 inline response with `offset > 0 || data.count > length`, then slicing to `start..<end` when valid. Add a Swift test that asserts range content, not only `count`.

Warnings / follow-ups:

- `scripts/sdk-integration-tests.sh:181-188` publishes MySQL with `-p ${MYSQL_CONTAINER_PORT}:3306`, which binds on all host interfaces. Bind to loopback (`127.0.0.1:${MYSQL_CONTAINER_PORT}:3306`) for a local disposable test DB.
- `scripts/sdk-integration-tests.sh:102-109` starts a fresh `mysql:8.0` helper container for every readiness probe and uses `--network host`. This is heavy and fragile on Docker Desktop/macOS. Prefer `docker exec "$MYSQL_CID" mysqladmin ping ...` inside the MySQL container, or a host TCP probe.
- `Makefile:128-129` documents `make sdk-integration-tests -- --only go,ts`, but the target only forwards `$(SDK_INTEGRATION_ARGS)`. In GNU make, args after `--` are still goals, so this usage is misleading. Document `make sdk-integration-tests SDK_INTEGRATION_ARGS="--only go,ts"` or implement goal forwarding.

Local validation performed:

- `bash -n scripts/sdk-integration-tests.sh` passed.
- `/usr/local/go/bin/go test -c ./pkg/client` passed.
- `/usr/local/go/bin/go test -tags=integration -c ./pkg/client` passed.
- `python3 -m py_compile drive9/stream.py drive9/patch.py tests/test_integration.py` passed in `clients/drive9-py`.
- `swift test` could not run locally because this machine's XCTest/xcrun setup is broken (`XCTest not available`), so Swift finding above is code-path review plus existing test coverage gap.
- Python unit pytest could not run locally because `responses` is not installed in this environment.

mornyx:
Thanks @qiffang for the thorough review — all three blockers and the warnings/follow-ups are addressed in `6528395` (CI is green again: all 9 checks pass).

## Blockers

**1. Kotlin stale size assertion** — `Drive9IntegrationTest.kt:122-124` now expects `2L` after `writeWithRevision(file, v2, -1)` overwrote the file (was asserting the pre-overwrite `data.size`).

**2. Kotlin vault read not best-effort** — `vault read best-effort` now guards the create with `runCatching { c.createVaultSecret(...) }.getOrElse { return@runBlocking }`, matching the `vault management best-effort` pattern, so it skips cleanly when the local vault backend is disabled instead of throwing.

**3. Swift `downloadRangeStream` 200-fallback** — the condition `data.count > Int(length)` was wrong: it skipped slicing for nonzero-offset requests where the requested length >= the inline body size (e.g. 8-byte file, `offset=8, length=10` left the full 8 bytes) and for offsets past EOF (returned the full body). Now every 200 response is sliced with `start = min(Int(offset), total)`, `end = min(start + Int(length), total)`, returning empty when `offset >= total` (matching the 416 path semantics). The test now asserts the actual slice content for: (a) `offset=5, length=10` on a 10000-byte file, (b) `offset=8, length=10` on a 10-byte file (expects `ij`), and (c) `offset=100, length=10` past EOF (expects empty).

## Warnings / follow-ups

- **Go SSE test**: `TestIntegrationEvents` now gates the test write on the `OnCurrent` lifecycle callback (both watchers), so the event is not absorbed into the SSE head sequence during the initial replay.
- **TS SSE test**: `watchEvents` is now called with `{ signal: ctrl.signal }`, so `ctrl.abort()` actually stops the SSE loop instead of always waiting for the 12s timeout.
- **Runner MySQL bind**: `docker run -p` now uses `127.0.0.1::3306` (loopback only) instead of binding on all interfaces.
- **Runner readiness probe**: switched from a separate `docker run --network host mysql:8.0 mysqladmin` helper container to `docker exec $MYSQL_CID mysqladmin ping`, which is lighter and avoids the Docker-Desktop host-networking fragility.
- **Swift cleanup URL**: `cleanup()` now strips only the trailing slash (was `trimmingCharacters(in: /)` which dropped the leading slash too, producing `/v1/fsit-swift-...` and leaving test dirs behind).
- **Makefile docs**: updated the comment to `make sdk-integration-tests SDK_INTEGRATION_ARGS="--only go,ts"` (GNU make treats tokens after `--` as goals, so the previous `make ... -- ...` form never forwarded args).

## Bonus: new function coverage

While rebasing onto latest main I added live-server coverage for the two new features that landed on main after the original PR:

- **Go**: `TestIntegrationDownloadDir` (DownloadDir / DownloadDirCtx, refuse-overwrite, file-source rejection) and `TestIntegrationArchiveDir` (tar.gz default, zip + `**/excluded.txt` + `**/node_modules/**` exclude, `sub/**` include-only, unsupported-format error). Tagged the main-brought `download_dir_test.go` / `archive_test.go` with `//go:build !integration` so the integration build does not lose `newTestClient`.
- **TS**: `downloadDir` test (full tree, refuse-overwrite, file-source rejection) and `archive`/`archiveToFile` test (tar.gz stream, zip archiveToFile with exclude, `sub/**` include). Added `listTarGzNames` (ustar reader) and `listZipNames` (python3 zipfile over a temp file) helpers.

## SDK bug the new downloadDir test found

The TS `downloadDir` test failed with `Failed to parse URL from /v1/fs/.../dl/a.txt`: `readStreamImpl` followed a 302/307 redirect by `fetch(location)` with the raw `Location` header value, but the local S3 mock returns a **relative** path (e.g. `/s3/...`), and `fetch` requires an absolute URL. Fixed by resolving the `Location` against the original request URL (`new URL(location, baseUrl).toString()`) before following. Existing MSW tests still pass.

Local validation: `make sdk-integration-tests --no-build` → go/ts/rs/py/swift all PASS (kt SKIP, no gradle on this host); existing per-SDK mocked unit tests all still pass; both Go build tags (`integration` and default) compile.

srstack:
## Review

This PR adds live-server integration suites to all 6 SDKs (Go/TS/Rust/Python/Kotlin/Swift), a one-click bash runner that boots a disposable MySQL + drive9-server-local, and three real SDK bug fixes. The source fixes all check out: the Python `abort()` deadlock fix (capture executor + upload_id under the lock, shut down outside it), the Python `patch.py` `... or []` null-guard, the Swift inline-`200` range slice, and the JS relative-`Location` resolution (`new URL(location, baseUrl)`) are each correct. The test isolation via build tags (`//go:build !integration` / `integration`), env-gated skips, and per-test prefix dirs is clean. Findings are minor and concentrated in the runner.

### Findings

**1. `scripts/sdk-integration-tests.sh:4367` — the MySQL container uses a fixed name (`drive9-sdk-intg-mysql`) with `docker run --rm` but no pre-run `docker rm -f`, so a second run after `--keep-server` (or any crash where `--rm` didn't fire) aborts immediately.**
`docker run -d --rm --name "$MYSQL_CONTAINER_NAME"` fails with `Conflict. The container name ... is already in use` whenever a container with that name still exists. `--keep-server` explicitly leaves the container running (line 4545 clears `MYSQL_CID` so teardown skips it), so the very next normal invocation collides and exits 1 under `set -e`. Cleanup only does `docker stop` (relying on `--rm`); nothing removes a stale/leftover container first. Fix: `docker rm -f "$MYSQL_CONTAINER_NAME" >/dev/null 2>&1 || true` before `docker run`.

**2. `scripts/sdk-integration-tests.sh:4206-4209` — fixed host ports (MySQL `13306`, listen `19009`) contradict the PR description's "random host port," and make concurrent runs / a busy host port fail.**
Two runs on one host, or any process already bound to `13306`/`19009`, collide. The env-var overrides (`DRIVE9_SDK_INTG_MYSQL_PORT`, `--port`) mitigate this manually, but the default isn't the "disposable / random" behavior the summary advertises. Either pick an ephemeral port (e.g. `-p 127.0.0.1:0:3306` then read back the mapping) or reword the description.

**3. `scripts/sdk-integration-tests.sh:4234-4238` — `--only` silently ignores unknown SDK tokens.**
`--only go,typescript` (or any typo) sets `SDKS="go typescript"`; `sdk_enabled` never matches `typescript`, so it runs nothing for TS and the summary prints `not selected` with no warning that the token was unrecognized. A one-line validation loop against `$ALL_SDKS` that errors on an unknown token would prevent a silent no-op run.

**4. `clients/drive9-py/drive9/stream.py:1299-1302` — leftover `try: _abort_upload_v2(...) finally: pass` scaffolding.**
Now that `_shutdown_executor()` moved out of the `finally`, the `finally: pass` does nothing; the `try` can be dropped so the call stands alone. Purely cosmetic — no behavior change.

Overall this is solid, well-isolated test tooling. None of the findings block the core value; #1 is the one most likely to bite a developer on a second local run.

mornyx:
Thanks @srstack — all four findings addressed in `3b14210` (CI green: all SDK jobs + ci + e2e smoke pass).

1. **Stale container name** — `bootstrap_db` now does `docker rm -f "$MYSQL_CONTAINER_NAME" >/dev/null 2>&1 || true` before `docker run`, so a second run after `--keep-server` (or a crash where `--rm` didn't fire) no longer aborts with *container name already in use*.

2. **'random host port' wording** — the runner does use fixed default ports (`13306` MySQL, `19009` server), both overridable (`DRIVE9_SDK_INTG_MYSQL_PORT` / `--port`). I went with rewording the description rather than ephemeral ports: picking `-p 127.0.0.1:0:3306` and reading back the mapped port would mean the server's `DRIVE9_LOCAL_DSN` / `DRIVE9_LISTEN_ADDR` and the per-SDK `DRIVE9_SERVER` all have to be wired through the discovered port at runtime, which is more moving parts for little gain on a local disposable run. The description now says *loopback-bound fixed host port, overridable* and the test-plan line says *fixed listen port (overridable via `--port`)* instead of *random*.

3. **`--only` silently ignores unknown tokens** — there's now a validation loop over `$SDKS` against `$ALL_SDKS` that exits 2 with `unknown SDK in --only: '<token>' (known: go ts rs py kt swift)`. Verified: `--only go,typescript` now errors instead of running nothing.

4. **Leftover `try: _abort_upload_v2(...) finally: pass`** — dropped; the abort call now stands alone now that `_shutdown_executor` moved out of the finally. `pytest tests/test_stream.py` still 4/4 green.

srstack:
## Re-review (after `3b14210c` — address review feedback)

Three of the four items are resolved:

- **Stale MySQL container (#1):** fixed — `docker rm -f "$MYSQL_CONTAINER_NAME"` now runs before `docker run`, so a second run after `--keep-server` (or a crash where `--rm` didn't fire) no longer aborts on a name conflict.
- **`--only` silent no-op (#3):** fixed — tokens are validated against `$ALL_SDKS` with `grep -qw`, and an unknown SDK now exits `2` with a clear message instead of silently running nothing.
- **Python `finally: pass` (#4):** fixed — the leftover scaffolding is gone; `_abort_upload_v2(upload_id)` is called directly outside the lock.

Remaining (non-blocking):

- **Fixed default ports (#2, `scripts/sdk-integration-tests.sh:4202-4205`):** still `13306`/`19009`. Overridable via `--port` / `DRIVE9_SDK_INTG_MYSQL_PORT`, so concurrent runs are workable — just not the "random host port" the description advertises. Fine to leave as-is if you update the wording.

Verified the earlier source fixes still stand (Python abort deadlock, patch null-guard, Swift inline-range slice, JS relative-Location). LGTM.

srstack (approved):
LGTM. The three review rounds' findings are all resolved: stale-container guard, --only validation, and the Python abort cleanup. Solid, well-isolated cross-SDK integration test tooling plus three legit SDK bug fixes. Remaining port-default note is non-blocking.

PR #675, opened 2026-07-06
ci: publish drive9-csi image in parallel with prod-cd

qiffang:
## Summary

- Adds a parallel job to `prod-cd` that dispatches [`drive9-ai/k8s-csi`](https://github.com/drive9-ai/k8s-csi)'s `publish-image.yml`, producing a `ghcr.io/drive9-ai/drive9-csi:drive9-<drive9sha7>-csi-<csisha7>` image pinned to the drive9 commit being promoted.
- Dispatch logic lives in a new reusable/dispatchable workflow `csi-publish-trigger.yml`, so it can also be **triggered manually** from the drive9 Actions UI for an arbitrary `drive9_ref` — useful for one-off builds and hotfixes.
- Runs in **parallel** with the `deploy` job: a failed prod rollout does not cancel the CSI build (and vice versa). SHA resolution runs as a tiny standalone job that both consume.

## Prerequisite (👉 needs to happen before merge)

Add repo secret **`CSI_DISPATCH_TOKEN`** — a GitHub PAT with `actions:write` on `drive9-ai/k8s-csi`. Two options:

- **Fine-grained PAT (recommended)**: Repository access → `drive9-ai/k8s-csi`; Permissions → `Actions: Read and write`, `Contents: Read`.
- **Classic PAT**: scopes `repo` + `workflow`.

`GITHUB_TOKEN` cannot be used — it can't dispatch workflows in another repo.

## Companion PR

Requires https://github.com/drive9-ai/k8s-csi/pull/NEW (adds `workflow_dispatch` input `drive9_ref` to `publish-image.yml`).

## How it triggers

| Trigger | Behavior |
|---|---|
| `prod-cd` manual run | Resolves drive9 SHA from the image tag being promoted → dispatches k8s-csi build in parallel |
| Manual `csi-publish-trigger` from Actions UI | Prompts for `drive9_ref` (defaults to workflow commit); dispatches directly |
| Manual `publish-image.yml` in k8s-csi repo | Same as before, plus optional `drive9_ref` input |

## Test plan

- [ ] Configure `CSI_DISPATCH_TOKEN` secret
- [ ] Merge the companion PR in `drive9-ai/k8s-csi` first
- [ ] Run `prod-cd` with `region=aws-ap-southeast-1` (any region — CSI job runs regardless of region)
- [ ] Verify a new run appears in [k8s-csi publish-image runs](https://github.com/drive9-ai/k8s-csi/actions/workflows/publish-image.yml) with the drive9 SHA in the tag
- [ ] Verify the resulting image lands at `ghcr.io/drive9-ai/drive9-csi:drive9-<sha7>-csi-<sha7>`
- [ ] Verify manual `csi-publish-trigger` run works with a custom `drive9_ref`

## Summary by CodeRabbit

* **New Features**
  * Added automated CSI image publishing during production promotion.
  * Production deploys now carry forward the matching build reference so the published image stays aligned with the release being promoted.
  * Added a reusable manual trigger for publishing the CSI image separately when needed.

### DIFF
--- a/.github/workflows/csi-publish-trigger.yml
+++ b/.github/workflows/csi-publish-trigger.yml
@@ -1,0 +1,78 @@
+name: Publish CSI image (drive9-ai/k8s-csi)
+
+# Dispatches drive9-ai/k8s-csi's `publish-image.yml` to build the drive9-csi
+# container image pinned to a specific drive9 commit. Runs in parallel with
+# `prod-cd` (invoked as a reusable workflow) and can also be triggered
+# manually from the Actions UI.
+#
+# Requires repo secret `CSI_DISPATCH_TOKEN`: a PAT with `actions:write` on
+# `drive9-ai/k8s-csi` (fine-grained PAT recommended). The default
+# `GITHUB_TOKEN` cannot dispatch workflows in a different repository.
+
+on:
+  workflow_dispatch:
+    inputs:
+      drive9_ref:
+        description: "drive9 commit SHA to bake into the CSI image (default: this workflow's commit)"
+        required: false
+        type: string
+      csi_ref:
+        description: "k8s-csi ref to build from (default: main)"
+        required: false
+        type: string
+        default: main
+  workflow_call:
+    inputs:
+      drive9_ref:
+        description: "drive9 commit SHA to bake into the CSI image"
+        required: true
+        type: string
+      csi_ref:
+        description: "k8s-csi ref to build from"
+        required: false
+        type: string
+        default: main
+    secrets:
+      CSI_DISPATCH_TOKEN:
+        required: true
+
+jobs:
+  dispatch:
+    name: Dispatch k8s-csi publish-image workflow
+    runs-on: ubuntu-latest
+    steps:
+      - name: Resolve drive9 ref
+        id: resolve
+        env:
+          INPUT_DRIVE9_REF: ${{ inputs.drive9_ref }}
+          FALLBACK_SHA: ${{ github.sha }}
+        run: |
+          set -euo pipefail
+          REF="${INPUT_DRIVE9_REF:-$FALLBACK_SHA}"
+          case "$REF" in
+            *[![:alnum:]_.-]*)
+              echo "Invalid drive9_ref: $REF" >&2
+              exit 1
+              ;;
+          esac
+          printf 'drive9_ref=%s\n' "$REF" >> "$GITHUB_OUTPUT"
+          printf 'Dispatching k8s-csi build for drive9 ref: %s\n' "$REF"
+
+      - name: Trigger drive9-ai/k8s-csi publish-image.yml
+        env:
+          GH_TOKEN: ${{ secrets.CSI_DISPATCH_TOKEN }}
+          DRIVE9_REF: ${{ steps.resolve.outputs.drive9_ref }}
+          CSI_REF: ${{ inputs.csi_ref || 'main' }}
+        run: |
+          set -euo pipefail
+          if [ -z "${GH_TOKEN:-}" ]; then
+            echo "CSI_DISPATCH_TOKEN secret is not set on this repository." >&2
+            echo "Create a PAT with actions:write on drive9-ai/k8s-csi and add it as CSI_DISPATCH_TOKEN." >&2
+            exit 1
+          fi
+          gh workflow run publish-image.yml \
+            --repo drive9-ai/k8s-csi \
+            --ref "$CSI_REF" \
+            -f drive9_ref="$DRIVE9_REF"
+          echo "Dispatched drive9-ai/k8s-csi publish-image.yml (csi ref: $CSI_REF, drive9_ref: $DRIVE9_REF)"
+          echo "Watch: https://github.com/drive9-ai/k8s-csi/actions/workflows/publish-image.yml"
--- a/.github/workflows/csi-publish-trigger.yml
+++ b/.github/workflows/csi-publish-trigger.yml
@@ -49,12 +49,19 @@ jobs:
         run: |
           set -euo pipefail
           REF="${INPUT_DRIVE9_REF:-$FALLBACK_SHA}"
-          case "$REF" in
-            *[![:alnum:]_.-]*)
-              echo "Invalid drive9_ref: $REF" >&2
-              exit 1
-              ;;
-          esac
+          # Require a hex commit SHA (7-40 chars). Earlier
+          # alnum+dot+dash validation would let branch/tag names
+          # through (e.g. `main`, `v1.2.3`), but k8s-csi bakes the
+          # value into the image tag `drive9-<ref7>-csi-<sha7>` and
+          # into the container image. A branch head can move between
+          # dispatch and build, producing a non-reproducible image
+          # under a misleading tag. Refuse anything that isn't a
+          # canonical SHA up-front so the failure mode is clear at
+          # dispatch time, not silently downstream.
+          if ! printf '%s' "$REF" | grep -Eq '^[0-9a-f]{7,40}$'; then
+            echo "Invalid drive9_ref (must be a 7-40 char hex commit SHA): $REF" >&2
+            exit 1
+          fi
           printf 'drive9_ref=%s\n' "$REF" >> "$GITHUB_OUTPUT"
           printf 'Dispatching k8s-csi build for drive9 ref: %s\n' "$REF"
 
--- a/.github/workflows/prod-cd.yml
+++ b/.github/workflows/prod-cd.yml
@@ -262,3 +262,79 @@ jobs:
             kubectl --kubeconfig ~/.kube/tencent-config -n drive9-tidbcloud \
               rollout undo deployment/drive9-server
           fi
+
+  resolve-drive9-sha:
+    # Resolves the drive9 short SHA from the image tag being promoted, so the
+    # csi-publish job can dispatch a matching k8s-csi build in parallel with
+    # the deploy job. Kept independent of `deploy` so both run concurrently.
+    name: Resolve drive9 SHA for CSI build
+    runs-on: ubuntu-latest
+    permissions:
+      id-token: write
+      contents: read
+    outputs:
+      drive9_sha: ${{ steps.resolve.outputs.drive9_sha }}
+    steps:
+      - name: Configure AWS credentials
+        uses: aws-actions/configure-aws-credentials@v4
+        with:
+          role-to-assume: ${{ secrets.AWS_ASSUME_ARN_PROD }}
+          aws-region: ap-southeast-1
+
+      - name: Update kubeconfig for dev
+        run: |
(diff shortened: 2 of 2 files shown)


qiffang:
**adversary-2 review** — 3 findings net-new to CodeRabbit's automated review. I'm the second adversary reviewer per the standard #review workflow; leaving aggregate framing to @adversary-1 (task #19 owner).

### 1. `csi-publish-trigger.yml` validation regex is looser than what k8s-csi enforces (medium)

Outer workflow allows `drive9_ref` matching `[[:alnum:]_.-]*` (lines 47-52), while the k8s-csi child in [PR #14](https://github.com/drive9-ai/k8s-csi/pull/14) validates with strict `[[:alnum:]]*`. Manual-dispatch UX becomes split:

- Input `main` (branch name, all alnum) → passes outer, passes inner → produces `drive9-main-csi-<sha7>` (misleading — trace tag suggests a commit).
- Input `v1.0.0` (contains `.`) → passes outer, **fails inner** with a validation error that surfaces from the *dispatched* workflow rather than the one the user just clicked. Debug is split across two repos.

Suggest tightening the outer to a trailing-hex-SHA match (`^[0-9a-f]{7,40}$`), same as the inner, so failures surface immediately in the caller.

### 2. `${DRIVE9_REF:0:7}` on the k8s-csi side silently truncates a full 40-char input (low, UX)

If the manual-dispatch user pastes a full 40-char SHA into `drive9_ref` (which passes validation), the resulting trace tag is `drive9-<first7of40>-csi-<sha7>` — but the log message just says `drive9_ref=<full 40>`. When the user then goes hunting for a tag that matches the full SHA they typed, they can't find one. Suggest changing the log to explicitly show the truncation:

```
Dispatching k8s-csi build for drive9 ref: <full input> (trace slice: <first7>)
```

Applied in [k8s-csi PR #14 side](https://github.com/drive9-ai/k8s-csi/pull/14) as well.

### 3. `resolve-drive9-sha` accepts non-SHA `image_tag` suffixes (medium, extends CodeRabbit)

CodeRabbit's finding on lines 291-327 flagged the template-injection risk on `${{ inputs.image_tag }}` (correct — must be routed through `env:` like `csi-publish-trigger.yml` does). Extending that: the same block accepts any `[[:alnum:]_.-]*` string as the tag and then extracts `SHA="${TAG##*-}"`. A tag like `prod-release1` or `prod-20260707` passes length + alnum checks but is not a real drive9 commit SHA. Prod deploy proceeds while k8s-csi eventually fails to `git checkout` the fake SHA, silently pinning a wrong image or aborting mid-parallel.

Suggest: after extracting, gate with `case "$SHA" in *[!0-9a-f]*)` (hex-only) or require a separate explicit `drive9_ref` input for manual overrides so the SHA source is unambiguous.

---

**Verified against upstream state**:
- k8s-csi PR #14 side `[[:alnum:]]*` validation confirmed at `.github/workflows/publish-image.yml:58-65`.
- CodeRabbit's actionable comments (2 findings) confirmed present and correct.
- No stale `DRIVE9_REF` env var references outside expected Docker build-arg / Makefile / docs (verified via `rg --hidden` on the k8s-csi PR branch per strategy-2 aggregate `#drive9:b8b00db9`).

Not blocking merge as long as CodeRabbit's template-injection finding is addressed; the three above are hardening. Deferring the merge-vs-request-changes call to @adversary-1 as task #19 owner.

qiffang:
Aggregate review on current head `8019ff2`.

Verdict: request changes before merge. The direction is right, but the current workflow does not yet guarantee that the CSI image is baked against the same Drive9 image/commit that prod-cd actually deploys.

1. **The deployed image and the CSI-baked Drive9 ref are resolved independently.**

`deploy` still resolves the image from dev in its own job, while `resolve-drive9-sha` separately queries the same dev deployment for `csi-publish`. Because those jobs run independently, a dev rollout between the two reads can produce:

- prod deploy promotes image/tag A
- csi-publish bakes Drive9 ref B

That breaks the core invariant this PR is adding. Fix by making one upstream resolver job produce the full image/tag and extracted Drive9 ref, then have both `deploy` and `csi-publish` consume that shared output. Deploy and CSI can still run in parallel after the resolver.

2. **The new resolver directly interpolates `inputs.image_tag` into shell.**

CodeRabbit flagged this and I confirmed it. The validation happens after shell interpolation, so a malicious/manual workflow input can break out of the assignment before the `case` guard runs. Route the input through `env:` and read it as a shell variable. If you centralize the resolver per finding 1, fix the existing deploy-side image_tag handling at the same time so there is only one safe path.

3. **The extracted Drive9 ref should be validated as a hex commit SHA, not just alnum.**

`SHA="${TAG##*-}"` plus `[[:alnum:]]` accepts suffixes like `release1` or `20260707`, which are not commit SHAs. Prod deploy would continue while the CSI build either fails later in the other repo or builds against a misleading ref. Tighten to a hex SHA shape, e.g. `[0-9a-f]{7,40}`, or add an explicit `drive9_ref` input if manual overrides need separate semantics.

4. **Merge ordering / contract note:** drive9 #675 assumes k8s-csi #14 is already available on the `csi_ref` it dispatches to. Since this workflow defaults `csi_ref` to `main`, merge k8s-csi #14 first, or explicitly point `csi_ref` at a ref containing the new `drive9_ref` input until #14 is merged.

I also reviewed k8s-csi #14. Its side needs the same SHA-shape tightening so branch/tag names cannot become reproducibility-breaking `drive9-main-csi-*` images.

Verification run locally:

```text
git diff --check origin/main...HEAD   # drive9 PR #675 worktree
git diff --check origin/main...HEAD   # k8s-csi PR #14 worktree
```

qiffang:
**adversary-2 re-review on `74d106b`: ✅ LGTM. All 3 blockers fixed correctly.**

## Trace against my original findings

### 1. P0 template-injection — FIXED ✓

Old (line 60):
```bash
INPUT_TAG='${{ inputs.image_tag }}'
```

New (line 74-80):
```yaml
env:
  INPUT_TAG: ${{ inputs.image_tag }}
  ECR_REGISTRY: ${{ env.ECR_REGISTRY }}
  ...
```

Value is now plain data at runtime, cannot break out of `$INPUT_TAG` variable expansion. Same env-var routing pattern already used in `csi-publish-trigger.yml`. Bonus: `ECR_REGISTRY`, `ECR_REPOSITORY`, `DEPLOY_NAMESPACE`, `K8S_DEPLOYMENT` also routed through env-vars — defense-in-depth on the other user-influenced substitutions in the same run block. Comment on line 66-73 explains the pattern for future maintainers.

### 2. P1 shared-resolver race — FIXED ✓

New topology:
```
resolve → deploy  (parallel branches)
       ↘ csi-publish
```

The `resolve` job produces outputs `image` / `registry` / `drive9_sha` from a single kubectl read. Both `deploy` (line 118 `IMAGE: ${{ needs.resolve.outputs.image }}`) and `csi-publish` (line 358 `drive9_ref: ${{ needs.resolve.outputs.drive9_sha }}`) consume the same output.

Verified: `deploy` no longer runs its own kubectl to resolve the image — it just pulls from `$IMAGE` which is passed in as `needs.resolve.outputs.image`. No independent read, no race window. Both still run in parallel after `resolve`.

Comment on line 27-35 correctly cites the prior race and why consolidation fixes it.

### 3. P1 hex-only SHA extraction — FIXED ✓

New (line 118-124):
```bash
SHA="${TAG##*-}"
if ! printf '%s' "$SHA" | grep -Eq '^[0-9a-f]{7,40}$'; then
  echo "Could not extract hex drive9 SHA from tag: $TAG (extracted: $SHA)" >&2
  echo "Expected pattern: <ref-name>-<7-40 hex chars>" >&2
  exit 1
fi
```

Same pattern applied to `csi-publish-trigger.yml` (line 60-63) — consistent hex-only requirement across the outer + inner validation. `prod-release1` / `v1.2.3` / `main` all fail-fast now at the correct layer with a clear error message. My original finding #3 also flagged the outer/inner regex mismatch — that's resolved by the coordinated tightening.

## Trace applied sharpened checklist

- ✅ **Wire audit**: no wire-visible changes, all env-var routing
- ✅ **Semantic-preservation**: `resolve` job accurately produces what both consumers need; no data loss between kubectl output → deploy image / csi-publish SHA
- ✅ **Discriminator + payload-value enumeration**: `INPUT_TAG` path branches (set vs empty), fallback path branches (kubectl succeeds vs fails), `SHA` extraction path branches (hex match vs miss). All paths traced. All fail closed on ambiguity.
- ✅ **Call-chain trace**: `resolve` produces outputs → `deploy` + `csi-publish` both consume via `needs.resolve.outputs.*`. No independent side-reads.
- ✅ **Regression guardrail**: the failing paths (`prod-release1`, `v1.2.3`) now emit explicit error messages; anyone re-adding the loose match would need to explicitly override the grep.

## Minor observations (non-blocker)

- **Uppercase hex not accepted**: `^[0-9a-f]{7,40}$` rejects `AFF1023D...`. Standard `git rev-parse` output is lowercase, so this is correct in practice — but if a user pastes an uppercase SHA from an external tool, they'll get a confusing "not hex" error. Optional: `^[0-9a-fA-F]{7,40}$` or `grep -Eiq` for case-insensitive. Marginal.
- **Comment quality**: comments on line 27-35, 66-73, 105-113, 118-124 all cite specific prior bugs and why the fix chose the current pattern. This is the level of documentation I'd want on any security-sensitive workflow. Kudos.

## Verdict

✅ **LGTM. All 3 blockers cleanly resolved.** Approve merge pending @adversary-1 sign-off as task #19 owner.

CI check to verify locally / on GHA: run prod-cd with `image_tag=abc'; touch /tmp/pwned; echo '` — must fail on the alnum validation before ever executing anything. Also run with `image_tag=prod-release1` — must fail on the hex check. Both fail-modes were possible before this commit; both are blocked now.

qiffang:
Post-fix re-review on merged head `74d106b`: the previous blockers are resolved from my lane.

Checked:
- `inputs.image_tag` now enters shell via `env: INPUT_TAG`, with related deployment variables also routed through env.
- `prod-cd` now has a single `resolve` job whose outputs feed both `deploy` and `csi-publish`, so there is no independent dev-deployment side read that can produce deploy image A and CSI ref B.
- Drive9 ref extraction/validation is tightened to `^[0-9a-f]{7,40}$`; branch/tag-like values such as `main`, `prod-release1`, and `v1.2.3` fail.
- `csi-publish-trigger.yml` applies the same hex-only gate before cross-repo dispatch.

Local verification:
- YAML parse check passed for both changed workflows.
- Regex spot-check accepted lowercase 7/40-char hex and rejected `main`, `prod-release1`, `v1.2.3`, uppercase hex, and short refs.
- GitHub `ci` is green on the merged head.

No remaining blocker from my review lane. Operational note: running the new cross-repo dispatch needs `CSI_DISPATCH_TOKEN` configured on `mem9-ai/drive9` with Actions write permission on `drive9-ai/k8s-csi`.

PR #680, opened 2026-07-06
fix: clean up provisioned clusters on tenant create failure

srstack:
## Summary
- clean up TiDB Cloud clusters when tenant creation fails after a cluster has already been created
- route post-cluster local failures through the existing cleanup path instead of leaving failed tenants with leaked clusters
- update provision tests for partial-cluster errors and password encryption failures

## Tests
- `TESTCONTAINERS_RYUK_DISABLED=true go test ./pkg/server -run 'TestProvisionTiDBCloudNativeCleansClusterWhen(OrgBindingMissing|ProvisionReturnsClusterAndError|PasswordEncryptFails)|TestProvisionTiDBCloudNativeCreateTimeQuotaLocalPersistenceErrorIsInternal|TestProvisionCleansPartialClusterBeforeMarkingFailed' -count=1`\n- `TESTCONTAINERS_RYUK_DISABLED=true go test ./pkg/server`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of TiDB Cloud Native provisioning failures to reliably deprovision partially created clusters.
  * Ensures tenant provisioning metadata and related credentials are cleaned up consistently across multiple failure paths (including password encryption and status transition errors).
  * Clears tenant `cluster_id` and connection fields after cleanup to prevent stale provisioning data.
  * Expanded and stabilized automated tests to validate these cleanup outcomes deterministically.

### DIFF
--- a/pkg/meta/meta.go
+++ b/pkg/meta/meta.go
@@ -1415,6 +1415,13 @@ func (s *Store) GetTenantTiDBCloudOrgBinding(ctx context.Context, tenantID strin
 	return &rec, nil
 }
 
+func (s *Store) DeleteTenantTiDBCloudOrgBinding(ctx context.Context, tenantID string) (err error) {
+	start := time.Now()
+	defer observeMeta(ctx, "delete_tidbcloud_org_binding", start, &err)
+	_, err = s.db.ExecContext(ctx, `DELETE FROM tenant_tidbcloud_org_bindings WHERE tenant_id = ?`, tenantID)
+	return err
+}
+
 func (s *Store) CreateTenantPool(ctx context.Context, p *TenantPool) (err error) {
 	start := time.Now()
 	defer observeMeta(ctx, "create_tidbcloud_pool", start, &err)
--- a/pkg/meta/meta.go
+++ b/pkg/meta/meta.go
@@ -1415,13 +1415,6 @@ func (s *Store) GetTenantTiDBCloudOrgBinding(ctx context.Context, tenantID strin
 	return &rec, nil
 }
 
-func (s *Store) DeleteTenantTiDBCloudOrgBinding(ctx context.Context, tenantID string) (err error) {
-	start := time.Now()
-	defer observeMeta(ctx, "delete_tidbcloud_org_binding", start, &err)
-	_, err = s.db.ExecContext(ctx, `DELETE FROM tenant_tidbcloud_org_bindings WHERE tenant_id = ?`, tenantID)
-	return err
-}
-
 func (s *Store) CreateTenantPool(ctx context.Context, p *TenantPool) (err error) {
 	start := time.Now()
 	defer observeMeta(ctx, "create_tidbcloud_pool", start, &err)
@@ -2464,6 +2457,52 @@ func (s *Store) UpdateTenantConnection(ctx context.Context, id string, cluster *
 		return ErrNotFound
 	}
 	return nil
+}
+
+func (s *Store) UpdateTenantClusterReference(ctx context.Context, id string, cluster *Tenant) (err error) {
+	start := time.Now()
+	defer observeMeta(ctx, "update_tenant_cluster_reference", start, &err)
+	if cluster == nil {
+		return fmt.Errorf("tenant cluster reference is required")
+	}
+	res, err := s.db.ExecContext(ctx, `UPDATE tenants
+		SET db_host = ?, db_port = ?, db_user = ?, db_name = ?, db_tls = ?,
+			provider = ?, cluster_id = ?, branch_id = ?, claim_url = ?, claim_expires_at = ?, updated_at = ?
+		WHERE id = ?`,
+		cluster.DBHost, cluster.DBPort, cluster.DBUser, cluster.DBName, boolToInt(cluster.DBTLS),
+		cluster.Provider, nullStr(cluster.ClusterID), cluster.BranchID, nullStr(cluster.ClaimURL), cluster.ClaimExpiresAt,
+		time.Now().UTC(), id)
+	if err != nil {
+		return err
+	}
+	n, _ := res.RowsAffected()
+	if n == 0 {
+		return ErrNotFound
+	}
+	return nil
+}
+
+func (s *Store) ClearTenantProvisionMetadata(ctx context.Context, tenantID string) (err error) {
+	start := time.Now()
+	defer observeMeta(ctx, "clear_tenant_provision_metadata", start, &err)
+	now := time.Now().UTC()
+	return s.InTx(ctx, func(tx *sql.Tx) error {
+		res, err := tx.ExecContext(ctx, `UPDATE tenants
+			SET db_host = '', db_port = 0, db_user = '', db_password = ?, db_name = '', db_tls = 1,
+				cluster_id = NULL, branch_id = '', claim_url = NULL, claim_expires_at = NULL, updated_at = ?
+			WHERE id = ?`,
+			[]byte{}, now, tenantID)
+		if err != nil {
+			return err
+		}
+		if err := requireAffected(res); err != nil {
+			return err
+		}
+		if _, err := tx.ExecContext(ctx, `DELETE FROM tenant_tidbcloud_org_bindings WHERE tenant_id = ?`, tenantID); err != nil {
+			return err
+		}
+		return nil
+	})
 }
 
 func (s *Store) UpdateTenantDBCredentialIf(ctx context.Context, id, fromDBUser, dbUser string, dbPasswordCipher []byte) (updated bool, err error) {
--- a/pkg/server/provision_test.go
+++ b/pkg/server/provision_test.go
@@ -169,21 +169,21 @@ func (f *fakeProvisioner) ProvisionWithCredentialsAndQuota(_ context.Context, te
 }
 
 func (f *fakeProvisioner) Deprovision(_ context.Context, cluster *tenant.ClusterInfo) error {
-	f.deprovisionCalls.Add(1)
 	if cluster != nil {
 		out := *cluster
 		f.lastDeprovision = &out
 	}
+	f.deprovisionCalls.Add(1)
 	return f.deprovisionErr
 }
 
 func (f *fakeProvisioner) DeprovisionWithCredentials(_ context.Context, cluster *tenant.ClusterInfo, req tenant.CredentialProvisionRequest) error {
-	f.deprovisionCalls.Add(1)
 	f.lastCredentialReq = req
 	if cluster != nil {
 		out := *cluster
 		f.lastDeprovision = &out
 	}
+	f.deprovisionCalls.Add(1)
 	return f.deprovisionErr
 }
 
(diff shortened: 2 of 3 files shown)


srstack:
## Review

This PR replaces `persistFailedProvisionClusterInfo` (which saved a failed tenant's partial cluster connection info to the DB row) with `cleanupProvisionedClusterAfterProvisionFailure`, which marks the tenant failed and **synchronously deprovisions** the already-created TiDB Cloud cluster. Eight failure paths in `provisionTenant` now route through this helper, so a partially-created cluster gets deleted instead of leaked. The core goal is sound, but two issues surfaced.

### Findings

**1. `pkg/server/server.go:4857` — cleanup deletes the cluster but orphans the `tenant_tidbcloud_org_bindings` row.**
On native failure paths where the org binding was already persisted (`UpsertTenantTiDBCloudOrgBinding` at 4749) and a *later* step fails — `connection_persist_error` (4795), `status_update_error` (4803), `quota_error` (4815), `api_key_persist_error` (4832) — cleanup deprovisions the cluster and marks the tenant failed but never deletes the binding row. The only two paths that delete that row (`MarkTenantDeleted`, `FinalizeTenantDelete` in meta.go) are reachable solely through the owner-authenticated tenant-delete endpoint, and the owner API key is only issued at 4827 (after all four failing steps), so the owner can never authenticate a delete. No background reconciler reclaims it (`resumeForkCleanup` skips non-fork tenants). Result: a permanent binding row pointing at a now-deleted cluster. The PR fixes the cluster leak but introduces a metadata leak on those paths.

**2. `pkg/server/server.go:4882` — the remote cluster DELETE now runs synchronously in the HTTP request handler on a detached, deadline-free context.**
`cleanupProvisionedClusterAfterProvisionFailure` calls `deprovisionTenantCluster` directly (not via `startServerWorker`, unlike the schema-init path). `cleanupCtx := backgroundWithTrace(ctx)` derives from `context.Background()` with no deadline and drops the request's cancellation. `DeprovisionWithCredentials` issues an HTTP DELETE with digest auth (two round trips) against a client with a 60s timeout — so on a slow/hung TiDB Cloud API, the caller's error response can block up to ~120s, and a client disconnect won't abort it. The code it replaced only did a fast local DB write. Consider running the deprovision in a background worker.

### Non-blocking cleanup

**`pkg/server/server.go:4717`** — the `if cluster != nil && ClusterID != "" { cleanup } else { mark failed }` guard duplicates logic the helper already performs (it marks failed then early-returns on nil/empty cluster). Calling the helper unconditionally is behaviorally identical and removes the hand-synced else branch. Relatedly, several other failure paths (auto-embedding insert @4663, quota-not-supported @4697, and the `_ =`-discarded status updates @4681/4685) still hand-roll the mark-failed block, leaving the consolidation partial and inconsistent.

_Note: dropping `OrganizationID` in `clusterInfoFromTenant` does **not** break deprovision — `DeprovisionWithCredentials` only uses `ClusterID` + credentials. That path is fine._

srstack:
Addressed review feedback in `fb157d42`:

- Removed the duplicate nil/empty-cluster branch and route provision errors through `cleanupProvisionedClusterAfterProvisionFailure` consistently.
- Moved remote cluster deprovision into `startServerWorker` with a 2 minute cleanup timeout so tenant-create error responses are not blocked by TiDB Cloud DELETE.
- Added `DeleteTenantTiDBCloudOrgBinding` and clean up native org bindings after successful async deprovision to avoid stale metadata pointing at deleted clusters.
- Updated tests to wait for async cleanup and assert org binding cleanup on post-binding failures.

Validated with:
- `TESTCONTAINERS_RYUK_DISABLED=true go test ./pkg/server -run 'TestProvisionTiDBCloudNativeCleansClusterWhen(OrgBindingMissing|ProvisionReturnsClusterAndError|PasswordEncryptFails)|TestProvisionTiDBCloudNativeCreateTimeQuota(LocalPersistenceErrorIsInternal|RequiresQuotaProvisioner)|TestProvisionCleansPartialClusterBeforeMarkingFailed' -count=1`\n- `TESTCONTAINERS_RYUK_DISABLED=true go test ./pkg/meta`\n- `TESTCONTAINERS_RYUK_DISABLED=true go test ./pkg/server`

mornyx (approved):
Reviewed PR #680: fix: clean up provisioned clusters on tenant create failure.

**Summary of review:**

This PR ensures that when tenant creation fails after a TiDB Cloud cluster has already been provisioned, the cluster is deprovisioned instead of being leaked. Previously, post-cluster failures (password encrypt, connection persist, status update, API key persist) would mark the tenant as failed but leave the remote cluster running. The old `persistFailedProvisionClusterInfo` function (which saved connection info for manual cleanup) is removed.

**Key changes verified:**

1. **Unified cleanup path**: All post-cluster-create failure points now call `cleanupProvisionedClusterAfterProvisionFailure` instead of inline `UpdateTenantStatus(TenantFailed)`. The function first marks the tenant as `Failed` synchronously, then launches cluster deprovisioning asynchronously via `startServerWorker` with a 2-minute timeout.

2. **Background cleanup with timeout**: `provisionFailureClusterCleanupTimeout = 2 * time.Minute` prevents the cleanup goroutine from hanging indefinitely. The cleanup runs in the existing `forkWorkerWG` pool, so `Server.Close()` waits for it to finish.

3. **Org binding cleanup**: For `tidb_cloud_native` provider, after successful deprovision, `DeleteTenantTiDBCloudOrgBinding` is called to clean up the org binding row. New `meta.DeleteTenantTiDBCloudOrgBinding` method added.

4. **Credential handling**: For native provider, if `cred == nil` during cleanup, it logs a warning and skips deprovisioning (can't deprovision without credentials). This is correct — the cluster may leak in this case, but the tenant is marked failed and the situation is logged for operational follow-up.

5. **`persistFailedProvisionClusterInfo` removed**: Previously saved partial cluster connection info to the tenant row for manual cleanup. Now replaced by automatic deprovisioning. The test `TestProvisionPersistsPartialClusterBeforeMarkingFailed` is renamed to `TestProvisionCleansPartialClusterBeforeMarkingFailed` and asserts the cluster is deprovisioned with empty `cluster_id` on the tenant row.

6. **Test helpers**: `waitForDeprovisionCalls` and `waitForTiDBCloudOrgBindingNotFound` added for async test assertions with 2s deadline polling. Tests cover: provision returns cluster+error, password encrypt fails, org binding missing, quota local persistence error.

7. **`startServerWorker` during shutdown**: Codex noted that if `forkWorkerClosed` is true (server shutting down), the cleanup worker won't start and the cluster may leak. This is an acceptable edge case — the tenant is already marked `Failed` synchronously, so it won't appear active. The cluster can be cleaned up by another pod instance or operational tooling. Not a blocker.

**Verification:**
- `go vet ./pkg/server/... ./pkg/meta/...` — passes
- CI: CodeRabbit pass (ci/e2e pending at review time)

srstack:
## Re-review (after the async-cleanup revision)

Thanks for the update — the two commits address both earlier findings well: cleanup now runs in a background worker with a 2-minute timeout (no longer blocks the request), and the success path deletes the org binding via the new `ClearTenantProvisionMetadata`, while the deprovision-failure path persists the cluster reference via `UpdateTenantClusterReference`. A few new issues surfaced in the revised code.

### Findings

**1. `pkg/server/provision_test.go:1202` — `TestProvisionTiDBCloudNativePersistsClusterReferenceWhenCleanupFails` has a race and will flake.** (CONFIRMED)
The test waits only on `waitForDeprovisionCalls(t, prov, 1)` then reads the tenant row and asserts `cluster_id`/`db_host`/`db_user`/`db_name` are populated. But provision fails at the first step (`ProvisionWithCredentials`), so `UpdateTenantConnection` never runs — those columns are written *only* by the async worker's `UpdateTenantClusterReference` at server.go:4882, which runs *after* `deprovisionTenantCluster` increments the counter (fake at provision_test.go:186). So the counter can hit 1 before the DB write commits, and the `QueryRow` at 1206 can read an empty row → flake under CI scheduling. Fix: poll for `cluster_id` being set (like `waitForTiDBCloudOrgBindingNotFound` does for absence), rather than relying on the deprovision counter.

**2. `pkg/server/server.go:4882` — on deprovision failure, the leaked cluster + orphan org binding are never retried or reconciled.** (CONFIRMED)
When `deprovisionTenantCluster` fails, the worker persists the cluster reference but leaves `cluster_id` set and the `tenant_tidbcloud_org_bindings` row intact. No reconciler covers live `TenantFailed` tenants (`resumeForkCleanup` skips non-fork tenants at fork.go:1123; the pending/provisioning resumers only handle those statuses). So a transient TiDB Cloud DELETE failure permanently leaks the cluster — the exact outcome this PR targets. If the persisted reference is meant to enable a later retry, nothing consumes it yet; worth a follow-up sweeper or a note in the PR that this is best-effort.

**3. `pkg/server/server.go:4877` — cleanup is silently dropped if the server is shutting down.** (CONFIRMED)
`startServerWorker` logs `server_worker_start_after_close` and returns without running `fn` when `forkWorkerClosed` is true (server.go:1154-1158). A provision that fails during graceful shutdown marks the tenant `TenantFailed` synchronously but never deprovisions the cluster, and no restart reconciler picks it up → leak during the shutdown window.

**4. `pkg/meta/meta.go:2462` — `UpdateTenantClusterReference` is a near-verbatim copy of `UpdateTenantConnection`.** (cleanup)
The two differ only by omitting `db_password` from the SET clause. Both ~10-line `UPDATE tenants SET ...` statements must be kept column- and arg-aligned with the schema by hand; adding a connection column means editing both, and missing one silently diverges the two paths. Consider a shared helper parameterized on whether to include the password.

**5. `pkg/meta/meta.go:2472` — `UpdateTenantClusterReference` clobbers `db_tls`, `claim_url`, `claim_expires_at`.** (PLAUSIBLE, low)
The `t` struct built at server.go:4860 leaves `DBTLS=false`, `ClaimURL=\"\"`, `ClaimExpiresAt` zero, so on a late failure that already persisted `db_tls=1` and real claim values (via `UpdateTenantConnection`), this UPDATE flips `db_tls→0` and NULLs the claim fields. Impact is minor — those columns have no reader post-failure and deprovision uses only `cluster_id`+credentials — so this is an audit/fidelity blemish, not a functional break. Copying the known fields into `t` would keep the persisted reference faithful.

srstack:
Addressed the latest re-review in `0529089c`:

- Fixed the test race in `TestProvisionTiDBCloudNativePersistsClusterReferenceWhenCleanupFails` by polling for the persisted cluster reference instead of relying only on the deprovision call counter.
- Persist the non-sensitive cluster reference synchronously before starting async deprovision. This covers the shutdown-window case where `startServerWorker` may reject new work, and also keeps cluster metadata available if remote DELETE later fails.
- Preserve fidelity for cleanup references by setting `BranchID`, `DBTLS`, `ClaimURL`, and `ClaimExpiresAt` from the provisioned cluster / provider settings.
- Cleanup success still clears tenant provision metadata; cleanup failure keeps the reference for operational follow-up. A retry sweeper is intentionally left as follow-up scope rather than expanding this PR.

Validated with:
- `TESTCONTAINERS_RYUK_DISABLED=true go test ./pkg/server -run 'TestProvisionTiDBCloudNative(CleansClusterWhen(OrgBindingMissing|ProvisionReturnsClusterAndError|PasswordEncryptFails)|PersistsClusterReferenceWhenCleanupFails|CreateTimeQuota(LocalPersistenceErrorIsInternal|RequiresQuotaProvisioner))|TestProvisionCleansPartialClusterBeforeMarkingFailed' -count=1`\n- `TESTCONTAINERS_RYUK_DISABLED=true go test ./pkg/meta`\n- `TESTCONTAINERS_RYUK_DISABLED=true go test ./pkg/server`

srstack:
## Re-review (after `0529089c` — persist cleanup references before async deprovision)

The latest commit resolves three of the five items from the previous round:

- **Test flake (was #1):** fixed — the new `waitForTenantClusterReference` polling helper now gates the reads in `TestProvisionTiDBCloudNativePersistsClusterReferenceWhenCleanupFails` and `TestProvisionCleansPartialClusterBeforeMarkingFailed`, so the async worker write is awaited instead of raced.
- **`db_tls`/claim clobber (was #5):** fixed — the cleanup `t` struct now sets `BranchID`, `DBTLS` (via the extracted `dbTLSForProvisionedTenant` helper, which also de-dups the inline env check), `ClaimURL`, and `ClaimExpiresAt`, so the persisted reference is faithful.
- **Shutdown data-loss (was #3):** mitigated — `UpdateTenantClusterReference` now runs synchronously at server.go:626 *before* `startServerWorker`, so even if the worker is dropped during shutdown the cluster_id/reference is recorded in the DB.

Two residual items:

**1. `pkg/server/server.go:4875` — a failed deprovision still leaves the cluster with no automatic retry.** (design limitation, reduced severity)
When `deprovisionTenantCluster` fails (or the worker is skipped at shutdown), the cluster reference is persisted but nothing ever retries the delete: `resumeForkCleanup` skips non-fork tenants (`t.Kind != meta.TenantKindFork` → continue, fork.go:1123), and no other reconciler covers live `TenantFailed` tenants. This is now softer than before — the reference is recorded for manual/observability recovery via the error logs and cluster_id — but the cluster still leaks until an operator intervenes. If auto-retry isn't in scope, consider a one-line note in the PR that failed-deprovision cleanup is best-effort + operator-recoverable.

**2. `pkg/server/server.go:642` — the worker's failure-branch `UpdateTenantClusterReference` duplicates the synchronous persist at line 626.** (minor)
`t` is unchanged between the two calls, so on the deprovision-failure path the reference is written twice with identical data. It's only non-redundant as a retry if the synchronous write at 626 failed; otherwise it's a wasted DB write on an already-failing path. Optional to drop.

Also still standing from before (non-blocking): `UpdateTenantClusterReference` (meta.go:2462) remains a near-verbatim copy of `UpdateTenantConnection` differing only by the `db_password` column — a shared helper would remove the hand-alignment risk.

Overall the core behavior is solid now: cleanup is off the request path, the org binding is removed on success, and the reference is durably recorded on failure.

PR #683, opened 2026-07-06
feat: add alicloud private endpoint domain env var + cloud provider constants

srstack:
## Summary

Add `DRIVE9_TIDBCLOUD_ALICLOUD_PRIVATE_ENDPOINT_DOMAIN` env var for alicloud private endpoint domain override, similar to the existing `DRIVE9_TIDBCLOUD_TENCENT_PRIVATE_ENDPOINT_HOST` for tencentcloud.

In private mode, alicloud (like tencentcloud) doesn't have unified private endpoint DNS — only per-node DNS — so a statically configured override domain is needed.

## Changes

- Add `EnvTiDBCloudAlicloudPrivateEndpointDomain` constant and env var
- Add `alicloudPrivateEndpointHost` field to Provisioner struct
- Add cloud provider name constants (`cloudProviderTencentCloud`, `cloudProviderAlicloud`, `cloudProviderAWS`)
- Add `privateEndpointOverrideHost()` helper method that returns the correct override host per cloud provider
- Replace all direct `p.tencentPrivateEndpointHost` references with `p.privateEndpointOverrideHost()`
- Validate alicloud private endpoint domain is set when `usePrivate=true` and cloud provider is alicloud

## Summary by CodeRabbit

* **New Features**
  * Extended private-endpoint provisioning to select the correct override host based on the configured cloud provider (Tencent, Alicloud).
  * Added a required Alicloud private-endpoint domain setting when private endpoints are enabled (alongside existing Tencent settings).
* **Bug Fixes**
  * Fixed cluster and branch provisioning/polling so the chosen private-endpoint override host is applied consistently and takes priority over response-provided hosts.
* **Tests**
  * Added coverage for private-endpoint environment parsing and override-host priority behavior.

### DIFF
--- a/pkg/meta/quota.go
+++ b/pkg/meta/quota.go
@@ -8,6 +8,9 @@ import (
 	"fmt"
 	"sync/atomic"
 	"time"
+
+	"github.com/mem9-ai/drive9/pkg/logger"
+	"go.uber.org/zap"
 )
 
 // QuotaConfig holds per-tenant quota limits stored in the central server DB.
@@ -148,6 +151,9 @@ func (s *Store) GetQuotaConfigVersion(ctx context.Context, tenantID string) (str
 		 FROM tenant_quota_config WHERE tenant_id = ?`, tenantID,
 	).Scan(&maxStorageBytes, &maxFileSizeBytes, &maxFileCount, &maxMediaLLMFiles, &maxMonthlyCostMC)
 	if err == sql.ErrNoRows {
+		logger.Info(ctx, "quota_config_not_found_using_defaults",
+			zap.String("tenant_id", tenantID))
+		err = nil
 		return "", nil
 	}
 	if err != nil {
--- a/pkg/tenant/tidbcloudnative/provisioner.go
+++ b/pkg/tenant/tidbcloudnative/provisioner.go
@@ -29,15 +29,15 @@ import (
 )
 
 const (
-	EnvTiDBCloudNativeAPIURL               = "DRIVE9_TIDBCLOUD_NATIVE_API_URL"
-	EnvTiDBCloudNativeCloudProvider        = "DRIVE9_TIDBCLOUD_NATIVE_CLOUD_PROVIDER"
-	EnvTiDBCloudNativeRegion               = "DRIVE9_TIDBCLOUD_NATIVE_REGION"
-	EnvTiDBCloudNativeDefaultDatabaseName  = "DRIVE9_TIDBCLOUD_NATIVE_DEFAULT_DATABASE_NAME"
-	EnvTiDBCloudDefaultSpendingLimit       = "DRIVE9_TIDBCLOUD_DEFAULT_SPENDING_LIMIT"
-	EnvTiDBCloudNativePublicKey            = "DRIVE9_TIDBCLOUD_NATIVE_PUBLIC_KEY"
-	EnvTiDBCloudNativePrivateKey           = "DRIVE9_TIDBCLOUD_NATIVE_PRIVATE_KEY"
-	EnvTiDBCloudNativeUsePrivateEndpoint   = "DRIVE9_TIDBCLOUD_NATIVE_USE_PRIVATE_ENDPOINT"
-	EnvTiDBCloudTencentPrivateEndpointHost = "DRIVE9_TIDBCLOUD_TENCENT_PRIVATE_ENDPOINT_HOST"
+	EnvTiDBCloudNativeAPIURL                  = "DRIVE9_TIDBCLOUD_NATIVE_API_URL"
+	EnvTiDBCloudNativeCloudProvider           = "DRIVE9_TIDBCLOUD_NATIVE_CLOUD_PROVIDER"
+	EnvTiDBCloudNativeRegion                  = "DRIVE9_TIDBCLOUD_NATIVE_REGION"
+	EnvTiDBCloudNativeDefaultDatabaseName     = "DRIVE9_TIDBCLOUD_NATIVE_DEFAULT_DATABASE_NAME"
+	EnvTiDBCloudDefaultSpendingLimit          = "DRIVE9_TIDBCLOUD_DEFAULT_SPENDING_LIMIT"
+	EnvTiDBCloudNativePublicKey               = "DRIVE9_TIDBCLOUD_NATIVE_PUBLIC_KEY"
+	EnvTiDBCloudNativePrivateKey              = "DRIVE9_TIDBCLOUD_NATIVE_PRIVATE_KEY"
+	EnvTiDBCloudNativeUsePrivateEndpoint      = "DRIVE9_TIDBCLOUD_NATIVE_USE_PRIVATE_ENDPOINT"
+	EnvTiDBCloudTencentPrivateEndpointHost    = "DRIVE9_TIDBCLOUD_TENCENT_PRIVATE_ENDPOINT_HOST"
 	EnvTiDBCloudAlicloudPrivateEndpointDomain = "DRIVE9_TIDBCLOUD_ALICLOUD_PRIVATE_ENDPOINT_DOMAIN"
 
 	DefaultDatabaseName = "tidbcloud_fs"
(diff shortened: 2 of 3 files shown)


srstack:
## Review

This PR adds the `DRIVE9_TIDBCLOUD_ALICLOUD_PRIVATE_ENDPOINT_DOMAIN` env var (mirroring the existing tencentcloud override), introduces `cloudProvider*` name constants, and replaces the ~8 direct `p.tencentPrivateEndpointHost` references with a `privateEndpointOverrideHost()` helper that dispatches per cloud provider. The refactor is clean: every call site was converted (the only remaining `p.tencentPrivateEndpointHost` is inside the helper itself, which is correct), and the alicloud required-when-`usePrivate` validation faithfully mirrors the tencent one.

### Findings

**1. `pkg/tenant/tidbcloudnative/provisioner.go:42` (const block) & `:85` (struct) — the file is not gofmt-clean.**
`gofmt -l` flags the file. Adding `EnvTiDBCloudAlicloudPrivateEndpointDomain` (now the longest identifier in the `const` block) shifted the `=` alignment column, but the existing lines were left at the old alignment; similarly the `Provisioner` struct only re-aligned the bottom fields, not the untouched ones above (`apiURL`, `cloudProvider`, …). If CI runs a gofmt/goimports check this fails the build. Fix: `gofmt -w pkg/tenant/tidbcloudnative/provisioner.go`.

**2. `pkg/tenant/tidbcloudnative/provisioner.go:49` — `cloudProviderAWS` is declared but never used.**
`privateEndpointOverrideHost()` only branches on tencentcloud and alicloud; AWS falls through to `default: return ""`, and no other code references the constant (grep across the package finds only its declaration). It compiles (unused package-level consts are legal) but it's dead code. Either drop it, or add an explicit `case cloudProviderAWS: return ""` to document that AWS intentionally has no override.

Nothing else surfaced — the per-provider dispatch is at the right altitude (generalizes the mechanism rather than special-casing each call site), casing is handled consistently (`strings.ToLower` in the helper, `EqualFold` in validation), and the repeated `privateEndpointOverrideHost()` calls are trivial string switches not worth caching.

srstack:
## Re-review (after `caf7a4c5` + `e3e283ec`)

Both prior findings are resolved, and the added tests are solid:

- **gofmt (was #1):** fixed — the const block and `Provisioner` struct are now fully aligned; `gofmt -l` is clean.
- **`cloudProviderAWS` dead constant (was #2):** fixed — `privateEndpointOverrideHost()` now has an explicit `case cloudProviderAWS: return ""`, documenting that AWS intentionally has no override.

The new behavior change in `e3e283ec` — override host now *always* takes priority over the API-returned private host, instead of only filling in when the API host is empty — is applied **consistently** across all three connection-building sites (`ProvisionWithCredentialsAndQuota`, `clusterInfoFromResponse`, `fillBranchEndpoint`), and the `clusterConnectionIncomplete` / `branchConnectionIncomplete` checks already branch on `overridePrivateHost != ""`, so the completeness gating stays in sync (port + user_prefix still required, host check skipped when overriding). That matches the stated rationale — for tencentcloud/alicloud the API returns only per-node private DNS, so the static override should win. The new tests (`...OverrideTakesPriority`, `...RequiresAlicloudPrivateEndpointDomain` with the aws/tencent/alicloud matrix) cover the priority and validation paths well.

Package builds clean. No new findings — LGTM.

mornyx (approved):
Reviewed PR #683: feat: add alicloud private endpoint domain env var + cloud provider constants.

**Summary of review:**

Adds  env var for alicloud private endpoint override, parallel to the existing tencentcloud host override. Introduces  helper that dispatches per cloud provider, replacing all direct  references. Also includes a small fix in  where  now explicitly handles  as a non-error with defaults.

**Key areas verified:**

1. **Override priority change**: The old code used  only as a fallback when . The new code uses  as the **primary** host when set, taking priority over the response-provided host. This is intentional and correct — the override domain is statically configured because alicloud/tencentcloud don't have unified private DNS, so the API response host would be per-node and not usable directly.

2. ** / **: When override host is set, these functions no longer require  to be non-empty (line 1627-1628), since it will be replaced anyway. Correct — avoids unnecessary polling when the override is configured.

3. **Validation**:  validates that  is set when  and . Parallel to the existing tencentcloud validation. AWS requires no override.

4. ** dispatch**: Clean switch on  — tencentcloud → tencentPrivateEndpointHost, alicloud → alicloudPrivateEndpointHost, aws/default → "". Consistent with validation logic.

5. **All call sites updated**: , , , , , ,  — all now use . No remaining direct references to .

6. ** fix**:  now handles  by logging at Info level and returning  instead of propag the error. Callers () treat  error as success and continue with the (empty) version string — correct behavior for tenants without quota config. Codex noted this Info log fires on every 30s refresh cycle for tenants without quota config, which could be log noise. Non-blocking — suggest moving to Debug or adding a once-per-tenant guard in a follow-up.

7. **Tests**: New tests cover alicloud domain required/optional, tencentcloud host required/optional, AWS no override needed, override takes priority over response host (both cluster and branch paths).

**Verification:**
- `go vet ./pkg/tenant/... ./pkg/meta/...` — passes
- `go test ./pkg/tenant/tidbcloudnative -count=1` — passes
- Codex (glm-5.2, xhigh) — one P2 finding (quota.go Info log spam on periodic refresh, non-blocking)

mornyx (approved):
Reviewed PR #683: feat: add alicloud private endpoint domain env var + cloud provider constants.

**Summary of review:**

Adds DRIVE9_TIDBCLOUD_ALICLOUD_PRIVATE_ENDPOINT_DOMAIN env var for alicloud private endpoint override, parallel to the existing tencentcloud host override. Introduces privateEndpointOverrideHost() helper that dispatches per cloud provider, replacing all direct tencentPrivateEndpointHost references. Also includes a small fix in quota.go where GetQuotaConfigVersion now explicitly handles sql.ErrNoRows as a non-error with defaults.

**Key areas verified:**

1. **Override priority change**: The old code used tencentPrivateEndpointHost only as a fallback when the response Private.Host was empty. The new code uses privateEndpointOverrideHost() as the primary host when set, taking priority over the response-provided host. This is intentional and correct -- the override domain is statically configured because alicloud/tencentcloud don't have unified private DNS, so the API response host would be per-node and not usable directly.

2. **clusterProvisionMetadataIncomplete / branchConnectionIncomplete**: When override host is set, these functions no longer require Private.Host to be non-empty (line 1627-1628), since it will be replaced anyway. Correct -- avoids unnecessary polling when the override is configured.

3. **Validation**: NewProvisionerFromEnv validates that DRIVE9_TIDBCLOUD_ALICLOUD_PRIVATE_ENDPOINT_DOMAIN is set when usePrivate=true and cloudProvider=alicloud. Parallel to the existing tencentcloud validation. AWS requires no override.

4. **privateEndpointOverrideHost() dispatch**: Clean switch on strings.ToLower(p.cloudProvider) -- tencentcloud returns tencentPrivateEndpointHost, alicloud returns alicloudPrivateEndpointHost, aws/default returns empty string. Consistent with validation logic.

5. **All call sites updated**: ProvisionWithCredentialsAndQuota, clusterInfoFromResponse, fillBranchEndpoint, waitForClusterProvisionMetadata, waitForBatchClusterProvisionMetadataGroup, CreateBranchWithCredentials, waitForBranchActive -- all now use privateEndpointOverrideHost(). No remaining direct references to tencentPrivateEndpointHost.

6. **quota.go fix**: GetQuotaConfigVersion now handles sql.ErrNoRows by logging at Info level and returning empty string + nil instead of propagating the error. Callers (quotaConfigCache.refresh) treat nil error as success and continue with the (empty) version string -- correct behavior for tenants without quota config. Codex noted this Info log fires on every 30s refresh cycle for tenants without quota config, which could be log noise. Non-blocking -- suggest moving to Debug or adding a once-per-tenant guard in a follow-up.

7. **Tests**: New tests cover alicloud domain required/optional, tencentcloud host required/optional, AWS no override needed, override takes priority over response host (both cluster and branch paths).

**Verification:**
- go build ./pkg/... ./cmd/... -- passes
- go vet ./pkg/tenant/... ./pkg/meta/... -- passes
- git diff --check origin/main...HEAD -- clean
- go test ./pkg/tenant/tidbcloudnative -count=1 -- passes
- Codex (glm-5.2, xhigh) -- one P2 finding (quota.go Info log spam on periodic refresh, non-blocking)

PR #686, opened 2026-07-07
fix(fuse): implement directory fsync

qiffang:
## Summary
- Implement FUSE FsyncDir so directory fsync no longer falls through to go-fuse default ENOSYS.
- Sync local-overlay directories through the existing local sync hook and return OK for remote-persistent directories where Drive9 has no local directory metadata to flush.
- Add perf accounting for fsyncdir and regressions for remote path OK, local overlay sync, and SQLite -shm local-overlay mmap-safe open flags.

## SQLite/WAL/mmap findings
- On 81.70.84.180 with latest released CLI d8a7f2c, sqlite3 CLI with PRAGMA mmap_size=0 + WAL + synchronous=FULL + wal_checkpoint(FULL) passed on Drive9 mount.
- Python sqlite3 with mmap_size=0 + WAL passed on Drive9 mount: 100 inserts committed and read back.
- Direct C mmap probe showed current main-DB mmap gap is the FOPEN_DIRECT_IO path; PR #685 is the right fix for that path via CAP_DIRECT_IO_ALLOW_MMAP on Linux 6.6+.
- This PR intentionally does not switch SQLite main DB away from direct-io, to avoid reintroducing stale page-cache races and to stay compatible with PR #685.

## Validation
- git diff --check
- /usr/local/go/bin/go test ./pkg/fuse -run 'Test(SQLiteDirectIOPathMatching|OpenReadOnlySQLiteUsesDirectIO|OpenWritableSQLiteUsesDirectIO|SQLiteWALIndexSidecar|FsyncDir)' -count=1

Note: full ./pkg/fuse currently also hits unrelated existing mount_wait_test failures on this macOS environment.

## Summary by CodeRabbit

* **New Features**
  * Added directory syncing via FUSE, including proper handling for directories served through the local overlay.
  * Extended FUSE performance reporting with a new `fsyncdir` operation.

* **Bug Fixes**
  * Improved SQLite WAL sidecar handling for `-shm` by ensuring the `Create` request uses `O_TRUNC`, and by validating consistent cache-friendly open behavior.

* **Tests**
  * Added coverage confirming directory sync avoids remote calls and performs the expected single local sync for local-overlay paths.

### DIFF
--- a/pkg/fuse/dat9fs.go
+++ b/pkg/fuse/dat9fs.go
@@ -8467,6 +8467,38 @@ func (fs *Dat9FS) ReleaseDir(input *gofuse.ReleaseIn) {
 	fs.dirHandles.Delete(input.Fh)
 }
 
+func (fs *Dat9FS) FsyncDir(cancel <-chan struct{}, input *gofuse.FsyncIn) (status gofuse.Status) {
+	perfStart := fs.perfStart()
+	defer func() { fs.perfRecordFuse(perfFuseFsyncDir, perfStart, status, 0) }()
+
+	dh, ok := fs.dirHandles.Get(input.Fh)
+	if !ok {
+		return gofuse.OK
+	}
+	ctx, cf := fuseCtx(cancel)
+	defer cf()
+
+	if overlay, local, st := fs.localOverlayForDirPath(ctx, dh.Path); local {
+		if st != gofuse.OK {
+			return st
+		}
+		abs, err := overlay.abs(dh.Path)
+		if err != nil {
+			return localErrToFuseStatus(err)
+		}
+		dir, err := os.Open(abs)
+		if err != nil {
+			return localErrToFuseStatus(err)
+		}
+		defer dir.Close()
+		if err := syncOpenLocalFile(dir); err != nil {
+			return localErrToFuseStatus(err)
+		}
+	}
+
+	return gofuse.OK
+}
+
 func (fs *Dat9FS) listDir(ctx context.Context, dirPath string) ([]DirEntry, error) {
 	if overlay, local, st := fs.localOverlayForPath(ctx, dirPath); local {
 		if st != gofuse.OK {
--- a/pkg/fuse/dat9fs.go
+++ b/pkg/fuse/dat9fs.go
@@ -8490,7 +8490,7 @@ func (fs *Dat9FS) FsyncDir(cancel <-chan struct{}, input *gofuse.FsyncIn) (statu
 		if err != nil {
 			return localErrToFuseStatus(err)
 		}
-		defer dir.Close()
+		defer func() { _ = dir.Close() }()
 		if err := syncOpenLocalFile(dir); err != nil {
 			return localErrToFuseStatus(err)
 		}
--- a/pkg/fuse/dat9fs_test.go
+++ b/pkg/fuse/dat9fs_test.go
@@ -7815,6 +7815,64 @@ func TestCodingAgentLocalOverlayCreateReadWriteDoesNotUseRemote(t *testing.T) {
 	}
 }
 
+func TestFsyncDirRemotePathReturnsOK(t *testing.T) {
+	var remoteCalls atomic.Int32
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		remoteCalls.Add(1)
+		http.Error(w, "remote should not be used for directory fsync", http.StatusInternalServerError)
+	}))
+	defer ts.Close()
+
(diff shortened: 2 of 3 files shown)


qiffang:
Adversary review / meaning check:

This PR is meaningful, but only if we scope it as a narrow FUSE compatibility/correctness patch. It should not be positioned as performance work or as a complete directory-fsync durability implementation.

Why it matters:
- go-fuse defaults `FSYNCDIR` to `ENOSYS`; safe-write users such as SQLite/Git/tooling that call `fsync(parent_dir)` can see a false unsupported-operation failure.
- Local-overlay directories are real host directories; syncing them through `file.Sync()` is the right baseline behavior.
- The SQLite `-shm` `FOPEN_KEEP_CACHE` assertions are useful regression coverage for the mmap-safe local-overlay sidecar path.

What is not solved here:
- No impact on cold sequential read throughput, small-file write latency, read concurrency, or the SQLite main DB direct-io/mmap issue from #685.
- Remote-persistent `FsyncDir` returns OK without draining `writeBack` / `pendingIndex` / `commitQueue`; that is fine only as an explicit compatibility no-op, not as a remote durability barrier.
- For coding-agent local-overlay `.git`, parent directory fsync after temp-file rename is the point where Git expects metadata durability. This implementation syncs the host dir but does not force the Drive9 git-state checkpoint before returning.
- The local-overlay path reopens by path, so it syncs the current path rather than the original open dirfd if the directory was renamed/replaced before `fsync(dirfd)`. Probably not the common SQLite/Git path, but it is not POSIX-perfect.

Verdict: YES, there is value. I would merge after making the scope explicit in a comment/test or follow-up: “remote dir fsync is compatibility OK/no-op; not remote-durable.” If the intended product claim is “`fsync(parent)` makes create/rename durable across restore/remount,” then this PR is incomplete and needs to wait/drain pending child commits plus checkpoint `.git` state on local-overlay directory fsync.

Local validation I ran on head `08f2cec`:
- `go test ./pkg/fuse -run 'Test(SQLiteDirectIOPathMatching|OpenReadOnlySQLiteUsesDirectIO|OpenWritableSQLiteUsesDirectIO|SQLiteWALIndexSidecar|FsyncDir)' -count=1`

mornyx (approved):
## Review — PR #686

**Verdict: APPROVED**

### Summary

Implements `FsyncDir` in the FUSE filesystem so directory fsync no longer falls through to go-fuse's default ENOSYS. This was causing issues for SQLite WAL operations that call fsync on the directory containing the database.

### Changes reviewed

**pkg/fuse/dat9fs.go — new FsyncDir method (line 8470)**

Logic:
1. Perf accounting via perfFuseFsyncDir (proper addition to perf constants + names array)
2. Look up dir handle from fs.dirHandles — if not found, return OK (idempotent no-op)
3. Check localOverlayForDirPath:
   - Local overlay path: open underlying local directory, call syncOpenLocalFile (file.Sync()), return status
   - Remote-persistent path: return OK — Drive9 has no local directory metadata to flush

This is correct: the default ENOSYS was problematic because some callers (notably SQLite WAL checkpoint) treat fsync failure as a real error. Returning OK for remote directories is the right behavior since directory metadata is server-side.

**pkg/fuse/perf.go — new perfFuseFsyncDir constant**

Added in correct position in both the const block and the perfFuseOpNames array.

**pkg/fuse/dat9fs_test.go — 3 test additions**

1. TestFsyncDirRemotePathReturnsOK — verifies no remote API calls for remote-persistent directories, returns OK
2. TestFsyncDirLocalOverlaySyncsDirectory — verifies exactly 1 syncOpenLocalFile call for local-overlay directories
3. TestSQLiteWALIndexSidecarUsesTransientLocalOverlay — updated to assert FOPEN_KEEP_CACHE for -shm Create (with O_TRUNC) and Open

### Static verification

- go build ./pkg/fuse/... passes
- go-fuse default FsyncDir returns ENOSYS (confirmed in defaultraw.go) — this PR correctly overrides it
- syncOpenLocalFile = file.Sync() (line 1303) — standard fsync for local files
- localOverlayForDirPath delegates to localOverlayForPathWithHint with dir hint — consistent with other directory operations
- Handle not found returns OK (graceful, not an error)
- defer close uses errcheck-satisfied pattern
- CI: CodeRabbit pass, ci pass (6m19s), e2e pass (7m27s)

### Notes

- Codex (glm-5.2, xhigh) timed out after 10 minutes — small PR (3 files, +99/-1), static analysis is straightforward
- O_TRUNC addition to -shm Create test is good hardening: SQLite may open -shm with O_TRUNC, test now verifies FOPEN_KEEP_CACHE still returned (needed for mmap)

PR #690, opened 2026-07-07
fix(fuse): make community.ltp_syscalls fully PASS (truncate02 + getdents01)

mornyx:
## Background

The `community.ltp_syscalls` blackbox module was FAIL on `main`:

```
blackbox summary: PASS=0 FAIL=1 SKIP=0 XFAIL=0 WARN=0
community.ltp_syscalls: FAIL (product regression)
```

Two distinct LTP test failures were the cause:

```
truncate02.c:55: TFAIL: truncate(testfile, 256) failed: ENOENT (2)
truncate02.c:55: TFAIL: truncate(testfile, 512) failed: ENOENT (2)
getdents01.c:124: TFAIL: Some entries not found   (×3 tcases)
```

Everything else in the module (chmod01, chown01, close01, ftruncate01,
rename01, unlink01, write01, ...) passed.

### 1. `truncate02` — path-based `truncate(path, size)` on a newly-created, uncommitted file returns spurious `ENOENT`

`applyRemoteTruncate` (`pkg/fuse/dat9fs.go`) builds the truncated buffer by
reading the file's first `size` bytes from the **remote server** before
re-uploading. But a file created via `open(O_CREAT)` in write-back mode has
data only in an open dirty handle and the shadow store — it has no remote base
revision (`entry.Revision <= 0`) and does not exist on the server until Flush
commits it. So the remote read returns NotFound, `httpToFuseStatus` maps it to
ENOENT, and the caller sees a spurious ENOENT for an open, just-written file.

This is exactly LTP `truncate02`'s sequence: `create -> write 1024 ->
truncate(path, 256)` with the fd still open — both 256 and 512 tcases failed.

The `newSize == 0` path already had `adoptSingleCallerPathTruncate` to hand the
truncation to the open dirty handle instead of the remote; `newSize > 0` had no
equivalent.

### 2. `getdents01` — `readdir` does not return `.` and `..` entries

go-fuse's raw `RawFileSystem` layer does **not** synthesize the `.` and `..`
directory entries (only the higher-level nodefs connector does). Drive9's
`ReadDir` / `ReadDirPlus` only iterated the server's directory listing, which
never contains `.` / `..`. `listDir` and `validDirEntryName` also explicitly
stripped them. As a result `getdents` / `readdir` on a Drive9 FUSE mount never
yielded `.` or `..`, and `getdents01` reported `Entry '.' not found` /
`Entry '..' not found`.

### Fix 1: delegate path-truncate of an uncommitted new file to its open dirty handle

`pkg/fuse/dat9fs.go`:

- Added `adoptOpenHandlePathTruncate(entry, ino, callerPID, newSize)`: when
  `entry.Revision <= 0` and the file has an open dirty handle, apply the
  truncation to that handle (`truncateWritableHandleLocked`) and sync the
  shadow store, instead of contacting the remote server. The handle commits
  the new size on Flush, keeping the file as `PendingNew` rather than queuing
  an invalid `PendingOverwrite` against a non-existent base revision.
- Generalizes the existing `adoptSingleCallerPathTruncate` (which only handled
  `newSize == 0`) to any `newSize`.
- When multiple dirty handles exist on the same uncommitted file, adopts **any
  one** of them (preferring the caller-owned handle) and stops after the first
  adoption — a path-truncate collapses the file to a single size, so one
  adoption is sufficient; sibling dirty handles reconcile against the shadow
  store / pending index. This keeps multi-open uncommitted new files on the
  local path instead of falling back to the remote (which has no copy and
  would return ENOENT).
- Added the new branch at the top of `applyRemoteTruncate`, guarded by
  `entry.Revision <= 0 && !fs.layerEnabled()`.

### Fix 2: emit POSIX `.` and `..` entries in readdir

`pkg/fuse/dat9fs.go`:

- In both `ReadDir` and `ReadDirPlus`, emit synthetic `.` and `..` entries
  before the real `dh.Entries` loop. `.` references the directory's own inode
  (`dh.Ino`); `..` references the parent inode.
- **Stable offset scheme**: `.` at Off=1, `..` at Off=2, real entry `i` at
  `Off = i + 3`. Resume maps offsets `≤ 2` back to the synthetic block (re-emitting
  any of `.`/`..` the kernel hasn't consumed yet) and offsets `≥ 3` to real-entry
  index `offset - 3`. A small readdir buffer that only fits `.`/`..` no longer
  causes the next page to skip the first real children — matching go-fuse's
  nodefs pagination behavior.
- **ReadDirPlus**: advertise `.`/`..` with `EntryOut.NodeId = FUSE_UNKNOWN_INO`,
  fill no attributes, and do **not** call `IncrementLookup` — matching go-fuse's
  nodefs convention. The kernel does not create dentries for `FUSE_UNKNOWN_INO`,
  so no later FORGET is issued and the directory / parent `Nlookup` accounting
  stays correct (cannot be driven negative).
- **Parent path resolution**: `parentDirIno` strips the trailing slash from
  `pathutil.ParentPath` (`/parent/` → `/parent`) so the lookup matches the
  inode-table convention (directories stored without trailing slashes; root
  is `/`). Previously `..` resolved to inode 0 for any non-root nested directory.
- Added `dotDotEntries(dh)`, `parentDirIno(dirPath)`, and `stripTrailingSlash`.

## Tests

`pkg/fuse/dat9fs_test.go`:

- `TestSetAttr_WriteBackPathTruncateCreatedNonZero` — reproduces the LTP
  `truncate02` `create -> write -> truncate(path, non-zero)` sequence on a
  write-back-created file with no remote base revision. Asserts: SetAttr OK,
  dirty handle truncated to the new size, **zero** remote GET/PUT during the
  truncate, shadow store holds the truncated prefix, and a subsequent `Read`
  via the handle returns the truncated content.
- `TestSetAttr_WriteBackPathTruncateCreatedNonZeroMultipleHandles` — two
  dirty opens of a just-created uncommitted file; path-truncate must succeed
  locally with zero remote reads (covers the multi-dirty-handle case).
- `TestReadDirEmitsDotAndDotDot` — asserts readdir on a subdirectory yields
  `[. .. file.txt]` on the first page, that `.` / `..` carry `S_IFDIR` and
  reference the directory / parent inodes, that they are not repeated on the
  second page (resumed at offset 3), and a small-buffer pagination case proving
  `file.txt` is still reached when a page only fits `.`/`..`.
- `TestReadDirEmitsDotAndDotDotAtRoot` — asserts `..` at the root resolves to
  the root inode itself (root is its own parent).
- `TestReadDirPlusDotDotUsesUnknownIno` — asserts READDIRPLUS serves `.`/`..`
  with `FUSE_UNKNOWN_INO` and does not bump dir/parent `Nlookup`; `..` resolves
  to the real parent inode via the no-trailing-slash convention.

`go test ./pkg/fuse/` (truncate + readdir + vault + overlay suites) green.
`make lint` 0 issues.

## Verification

Rebuilt `bin/drive9-server-local` + `bin/drive9` on this branch and re-ran the
blackbox module with `--server-mode local`:

```
blackbox summary: PASS=1 FAIL=0 SKIP=0 XFAIL=0 WARN=0
Overall: PASS
community.ltp_syscalls: ✅ PASS (34.5s, passed)
TFAIL count: 0
```

Before/after per-test:

| Test | main | this branch |
|---|---|---|
| `truncate02` (256) | ❌ ENOENT | ✅ TPASS |
| `truncate02` (512) | ❌ ENOENT | ✅ TPASS |
| `getdents01` (×3 tcases) | ❌ `.`/`..` not found | ✅ All entries found |

## Performance note

- **truncate fix**: net positive. The new branch only fires for
  `entry.Revision <= 0` (newly-created, uncommitted files); the common
  path-truncate of already-committed files (`Revision > 0`) is untouched. For
  new files it saves a remote `ReadAtCtx` round-trip that was guaranteed to
  fail (the file doesn't exist remotely yet).
- **readdir fix**: tiny, per-first-page cost (one small allocation +
  `pathutil.ParentPath` + 1–2 inode-map lookups + 2 serialized dirents ~64 B).
  `IncrementLookup` is skipped for `.`/`..`. Later pages are unaffected. The
  only theoretical cost is that the first readdir page holds 2 fewer real
  entries, so very large directory scans may add one extra readdir syscall.

## Scope note

The `.`/`..` fix only touches `Dat9FS` (raw FUSE, used by `profile=none` and
the blackbox harness). `VaultFS` (`pkg/fuse/vaultfs.go`) has the same gap and
is left to a follow-up.

### DIFF
--- a/pkg/fuse/dat9fs.go
+++ b/pkg/fuse/dat9fs.go
@@ -1584,6 +1584,70 @@ func (fs *Dat9FS) adoptSingleCallerPathTruncate(remotePath string, callerPID uin
 	return adopted
 }
 
+// adoptOpenHandlePathTruncate applies a path-based truncate of an uncommitted
+// newly-created file (entry.Revision <= 0) to its owning open dirty handle
+// instead of the remote server, which has no copy of the data yet. It returns
+// OK when a dirty handle adopted the truncation, ENOENT when no eligible dirty
+// handle exists (callers fall back to the remote path), or another status on
+// a hard failure. Unlike adoptSingleCallerPathTruncate this handles any
+// newSize and does not require the truncating PID to own the handle, so a
+// different process can truncate a just-created file by path before Flush.
+func (fs *Dat9FS) adoptOpenHandlePathTruncate(entry *InodeEntry, ino uint64, callerPID uint32, newSize int64) gofuse.Status {
+	if fs == nil || fs.openHandles == nil || entry == nil || entry.Path == "" {
+		return gofuse.ENOENT
+	}
+	if newSize < 0 {
+		return gofuse.Status(syscall.EINVAL)
+	}
+
+	var matching []*FileHandle
+	for _, fh := range fs.openHandles.SnapshotPath(entry.Path) {
+		if fh == nil {
+			continue
+		}
+		fh.Lock()
+		dirty := fh.Dirty != nil
+		fh.Unlock()
+		if dirty {
+			matching = append(matching, fh)
+		}
+	}
+	if len(matching) == 0 {
+		return gofuse.ENOENT
+	}
+
+	adopted := false
+	for _, fh := range matching {
+		var abortStreamer func()
+		fh.Lock()
+		// Prefer the handle owned by the truncating caller when there is
+		// exactly one matching handle; otherwise adopt any dirty handle so a
+		// sibling-process truncate still succeeds against the local data.
+		ownerMatch := shouldAdoptSingleHandlePathTruncate(fh, callerPID, len(matching))
+		if ownerMatch || len(matching) == 1 {
+			var err error
+			abortStreamer, err = fs.truncateWritableHandleLocked(fh, newSize)
+			if err != nil {
+				log.Printf("open-handle path truncate failed for %s: %v", fh.Path, err)
+				fh.Unlock()
+				continue
+			}
+			adopted = true
+		}
+		fh.Unlock()
+		if abortStreamer != nil {
+			abortStreamer()
+		}
+	}
+	if !adopted {
+		return gofuse.ENOENT
+	}
+	if newSize == 0 {
+		fs.markDirtySize(ino, 0)
+	}
+	return gofuse.OK
+}
+
 func (fs *Dat9FS) refreshCommittedRevisionForOpenHandles(path string, revision int64, skip *FileHandle) {
 	if fs == nil || fs.openHandles == nil || path == "" || revision <= 0 {
 		return
@@ -2443,6 +2507,24 @@ func (fs *Dat9FS) applyRemoteTruncate(ctx context.Context, entry *InodeEntry, in
 	}
 	if newSize > int64(int(^uint(0)>>1)) {
 		return gofuse.Status(syscall.EFBIG)
+	}
+
+	// A newly-created file (e.g. open(O_CREAT) in write-back mode) may have
+	// data only in an open dirty handle and the shadow store, with no remote
+	// base revision yet. A path-based truncate(path, size) on such a file must
+	// not consult the remote server — the file does not exist there and the
+	// read would return NotFound, surfacing a spurious ENOENT to the caller
+	// (observed with LTP truncate02: create -> write -> truncate(path)).
+	// Delegate the truncation to the owning dirty handle, mirroring the
+	// newSize==0 adoption path; the handle commits the new size on Flush.
+	if entry.Revision <= 0 && !fs.layerEnabled() {
+		if st := fs.adoptOpenHandlePathTruncate(entry, ino, pid, newSize); st == gofuse.OK {
+			fs.invalidateReadCacheAndTargets(entry.Path)
+			fs.cacheFileForPath(entry.Path, newSize, entry.Mtime, 0)
+			return gofuse.OK
+		} else if st != gofuse.ENOENT {
+			return st
+		}
 	}
 
 	apiPath := fs.remotePath(entry.Path)
--- a/pkg/fuse/dat9fs.go
+++ b/pkg/fuse/dat9fs.go
@@ -1620,11 +1620,15 @@ func (fs *Dat9FS) adoptOpenHandlePathTruncate(entry *InodeEntry, ino uint64, cal
 	for _, fh := range matching {
 		var abortStreamer func()
 		fh.Lock()
-		// Prefer the handle owned by the truncating caller when there is
-		// exactly one matching handle; otherwise adopt any dirty handle so a
-		// sibling-process truncate still succeeds against the local data.
+		// Prefer the handle owned by the truncating caller; otherwise adopt
+		// any dirty handle. This keeps a multi-open, uncommitted new file on
+		// the local path instead of falling back to the remote server (which
+		// has no copy yet and would return ENOENT). Adopting one dirty handle
+		// is enough: a path-truncate collapses the file to a single size; any
+		// sibling dirty handles will reconcile against the shadow store /
+		// pending index and re-sync on their next write.
 		ownerMatch := shouldAdoptSingleHandlePathTruncate(fh, callerPID, len(matching))
-		if ownerMatch || len(matching) == 1 {
+		if !adopted && (ownerMatch || len(matching) >= 1) {
 			var err error
 			abortStreamer, err = fs.truncateWritableHandleLocked(fh, newSize)
 			if err != nil {
@@ -1637,6 +1641,10 @@ func (fs *Dat9FS) adoptOpenHandlePathTruncate(entry *InodeEntry, ino uint64, cal
 		fh.Unlock()
 		if abortStreamer != nil {
 			abortStreamer()
+		}
+		// Stop once one handle has adopted the truncate.
+		if adopted {
+			break
 		}
 	}
 	if !adopted {
@@ -8374,18 +8382,22 @@ func (fs *Dat9FS) ReadDir(cancel <-chan struct{}, input *gofuse.ReadIn, out *gof
 	}
 
 	// POSIX readdir yields "." and ".." before the real entries. go-fuse's raw
-	// layer does not synthesize them (only nodefs does), so emit them here at
-	// the first page (offset 0). Pagination offsets for real entries start at
-	// len(synthetic)+1 so subsequent pages skip this block.
-	if input.Offset == 0 {
-		for _, e := range fs.dotDotEntries(dh) {
-			if !out.AddDirEntry(e) {
-				return gofuse.OK
-			}
+	// layer does not synthesize them (only nodefs does), so emit them here.
+	// Use a stable offset scheme: "." occupies offset 1, ".." offset 2, and
+	// real entry i occupies offset i + dotDirCount + 1, so the kernel can
+	// resume past the synthetic entries without skipping real children.
+	dots := fs.dotDotEntries(dh)
+	for d := int(input.Offset); d < len(dots); d++ {
+		if !out.AddDirEntry(dots[d]) {
+			return gofuse.OK
 		}
 	}
 
-	for i := int(input.Offset); i < len(dh.Entries); i++ {
+	realStart := int(input.Offset) - len(dots)
+	if realStart < 0 {
+		realStart = 0
+	}
+	for i := realStart; i < len(dh.Entries); i++ {
 		e := dh.Entries[i]
 		if !validDirEntryName(e.Name) {
 			continue
@@ -8394,7 +8406,7 @@ func (fs *Dat9FS) ReadDir(cancel <-chan struct{}, input *gofuse.ReadIn, out *gof
 			Name: e.Name,
 			Ino:  e.Ino,
 			Mode: e.Mode,
-			Off:  uint64(i + 1),
+			Off:  uint64(i + len(dots) + 1),
 		}) {
 			break
 		}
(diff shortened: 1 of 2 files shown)


srstack:
## Review

This PR makes `community.ltp_syscalls` fully pass by fixing two real FUSE bugs: (1) a path-based `truncate(path, size>0)` on an uncommitted write-back-created file no longer round-trips to the remote (which returns spurious ENOENT) — it now delegates to the open dirty handle via the generalized `adoptOpenHandlePathTruncate`; and (2) `readdir` now emits the POSIX `.`/`..` entries that go-fuse's raw layer doesn't synthesize. Both fixes are well-scoped and thoroughly tested (zero-remote-IO assertion on the truncate path, dot-entry presence + pagination + root-parent coverage on readdir). `go build ./pkg/fuse/` is clean.

The truncate fix is correct: the new branch is gated on `entry.Revision <= 0 && !fs.layerEnabled()`, returns ENOENT to fall through to the unchanged remote path when no dirty handle exists, and only truncates the caller-owned (or sole) handle. Committed files (`Revision > 0`) are untouched.

### One finding (PLAUSIBLE, low severity)

**`pkg/fuse/dat9fs.go:8299-8303` / `8340-8344` — the synthetic `.`/`..` offsets collide with real-entry offsets, so a readdir buffer that fills *within* the synthetic block silently skips real entries.**
`dotDotEntries` assigns `.`→`Off:1` and `..`→`Off:2`, but the real-entry loop is unchanged and emits `Off: uint64(i+1)` starting at `1` — so real `entries[0]` also gets `Off:1` and `entries[1]` gets `Off:2`. FUSE treats each dirent's `Off` as the resume cookie. If `AddDirEntry` fills the buffer right after `.`/`..` but before the first real entry, the loop `return gofuse.OK`s with the last emitted `Off` = 1 or 2; the kernel resumes the next readdir at offset 1 or 2, and the real loop `for i := int(input.Offset)` restarts at index 1 or 2 — skipping `entries[0]` (and `entries[1]`). Net: missing directory entries across that page boundary.

This is why the diff comment at ReadDir ("*Pagination offsets for real entries start at `len(synthetic)+1`*") doesn't match the code — real offsets still start at 1, not 3. In practice the trigger is narrow (kernels issue ≥4 KiB readdir buffers, and two ~32-byte synthetic dirents plus a real one fit easily), so getdents01 and normal listings pass — hence PLAUSIBLE, not confirmed. But the offset space collides by construction. Cleanest fix: reserve the synthetic offsets and shift the real ones (e.g. real `Off = i + 1 + numSynthetic` on page 0, and subtract the synthetic count when resuming), so a mid-synthetic page break resumes at real index 0.

Everything else looks good — the scope note (VaultFS left to follow-up) and performance analysis are accurate, and `IncrementLookup` is correctly skipped for `.`/`..` so no extra kernel refs are pinned.

mornyx:
Thanks for the review — all flagged issues addressed in `77c0b99`:

### Multi-dirty-handle truncate fallback (CodeRabbit, Codex, Copilot)
`adoptOpenHandlePathTruncate` now adopts **any one** dirty handle when `len(matching) > 1`, instead of requiring `shouldAdoptSingleHandlePathTruncate` (which is always false for `matchCount != 1`). It stops after the first adopting handle — a path-truncate collapses the file to a single size, so one adoption is sufficient; sibling dirty handles reconcile against the shadow store / pending index and re-sync on their next write. This keeps multi-open uncommitted new files on the local path instead of falling back to the remote (which has no copy and would return ENOENT).

Added `TestSetAttr_WriteBackPathTruncateCreatedNonZeroMultipleHandles` (two dirty opens, path-truncate succeeds locally with zero remote reads).

### Readdir offset collisions (Codex)
Rewrote the offset scheme: `.` at Off=1, `..` at Off=2, real entry `i` at `Off = i + 3`. Resume maps offsets `≤ 2` back to the synthetic block (re-emitting whatever `.`/`..` the kernel hasn't consumed yet) and offsets `≥ 3` to real-entry index `offset - 3`. A small readdir buffer that only fits `.`/`..` no longer causes the next page to skip the first real children.

Added a small-buffer pagination case to `TestReadDirEmitsDotAndDotDot` proving `file.txt` is still reached across pages. Fixed the second-page test to resume at offset 3 (was 1, which masked the bug).

### ReadDirPlus IncrementLookup for `.`/`..` (Codex)
Switched to go-fuse's nodefs convention: `ReadDirPlus` advertises `.`/`..` with `EntryOut.NodeId = FUSE_UNKNOWN_INO`, fills no attributes, and does **not** call `IncrementLookup`. The kernel does not create dentries for `FUSE_UNKNOWN_INO`, so no later FORGET is issued — the directory / parent `Nlookup` cannot be driven negative. (Previously I emitted real inodes without incrementing, which would underflow on FORGET.)

Added `TestReadDirPlusDotDotUsesUnknownIno`: asserts `.`/`..` carry `FUSE_UNKNOWN_INO` and that dir/parent `Nlookup` is unchanged after `ReadDirPlus`.

### `..` parent path convention (Codex)
`parentDirIno` now strips the trailing slash from `pathutil.ParentPath` (`/parent/` → `/parent`) so the lookup matches the inode-table convention (directories stored without trailing slashes; root is `/`). Previously `..` resolved to inode 0 for any non-root nested directory.

Test now registers `/parent` (no trailing slash) and asserts `..` resolves to the real parent inode.

### Test pagination offset (Copilot)
Fixed — the "second page" test now resumes at offset 3 (past `.`/`..`), not 1. Using 1 actually masked the offset-collision bug the small-buffer case now catches.

---

Verification: `go test ./pkg/fuse/` green for all readdir + truncate suites; `make lint` 0 issues. Rebuilt and re-ran the blackbox module: `PASS=1 FAIL=0`, `TFAIL` count = 0 (session `20260707-124157`).

srstack:
## Re-review (after `77c0b99f` + `bd475f75`)

The offset-collision finding from the last round is fully resolved.

**Offset scheme (fixed):** `dotDotEntries` keeps `.`→`Off:1`, `..`→`Off:2`, and real entry `i` now uses `Off: i + len(dots) + 1` (= i+3), so the synthetic and real offset spaces no longer overlap. Both `ReadDir` and `ReadDirPlus` resume correctly: the synthetic loop iterates from `input.Offset` into `dots`, and `realStart := input.Offset - len(dots)` (clamped ≥0) resumes real entries without skipping. I traced the page-boundary cases (buffer fills after `.`, after `..`, after real[0]) and no entry is dropped or duplicated.

**ReadDirPlus lookup accounting (improved):** `.`/`..` are now emitted with `FUSE_UNKNOWN_INO` (Ino=0, `entryOut.NodeId = FUSE_UNKNOWN_INO`) and no `IncrementLookup`/attr-fill, matching go-fuse's nodefs pattern — so the kernel creates no dentry for them and directory/parent `Nlookup` accounting stays correct. That's more robust than the first version, which referenced the real dir/parent inodes.

**Tests:** the reworked suite now exercises exactly the boundary I flagged — a 64-byte page that only fits `.`/`..`, then resumes at the last returned offset (`smallPageLastOffset`) and asserts the real children still appear. The fragile `reflect`/`unsafe` `ReadDirPlus` wire-parsing was dropped per `bd475f75`.

The truncate fix (multi-handle adoption, `Revision <= 0 && !layerEnabled()` gate, remote fallback on ENOENT) remains correct. Nothing else outstanding.

LGTM.

srstack (approved):
LGTM. The readdir offset-collision finding is fully resolved — real entries now use Off = i + len(dots) + 1 so the synthetic/real offset spaces no longer overlap, both ReadDir and ReadDirPlus resume correctly via realStart clamping, and the reworked small-buffer test covers the exact page-boundary case. ReadDirPlus now emits ./.. with FUSE_UNKNOWN_INO and no IncrementLookup, matching go-fuse's nodefs accounting. Truncate fix remains correct.

qiffang (approved):
## Adversarial Review — GREEN ✅ (1 warning)

Two independent POSIX conformance fixes for LTP truncate02 + getdents01. Both root causes clearly identified, fixes well-scoped, tests thorough.

### Fix 1: `adoptOpenHandlePathTruncate` — Correct

Guard `entry.Revision <= 0 && !fs.layerEnabled()` is tight. Fallback to ENOENT preserves the existing remote path for committed files. Shadow store sync and `cacheFileForPath` keep attr caches consistent.

**W1 (dead code):** `ownerMatch` is never load-bearing. `shouldAdoptSingleHandlePathTruncate` returns `false` when `matchCount != 1`, but `!adopted && (ownerMatch || len(matching) >= 1)` always enters the branch because `len(matching) >= 1` is invariantly true (early return at top). The function always adopts the first dirty handle. Simplify to `if !adopted {`.

### Fix 2: readdir `.`/`..` — Correct

Stable offset scheme (`.`=1, `..`=2, real=`i+3`) handles pagination correctly. ReadDirPlus uses `FUSE_UNKNOWN_INO`, skips `IncrementLookup` — Nlookup stays correct. `parentDirIno` strips trailing slash matching inode-table convention. Root-is-own-parent handled.

### Tests — Complete

5 new tests covering single/multi-handle truncate, ReadDir ordering/pagination/root, ReadDirPlus FUSE_UNKNOWN_INO invariant. VaultFS gap deferred to follow-up (acceptable).

qiffang (approved):
## Review Summary — GREEN

PR quality is high. Both fixes have clear root-cause analysis, targeted fixes, and thorough regression tests.

### Fix 1: `truncate02` — path-truncate of uncommitted new file ✅

- Root cause accurate: `applyRemoteTruncate` reads remote for `Revision <= 0` files → ENOENT
- Guard `entry.Revision <= 0 && !fs.layerEnabled()` is well-scoped
- `adoptOpenHandlePathTruncate` logic is clean: prefer caller-owned handle, break after first adoption

**Nits:**
1. `!adopted && (ownerMatch || len(matching) >= 1)` — `len(matching) >= 1` is always true at this point (early return above), so condition is effectively `!adopted`. The `ownerMatch` check is redundant. Not a bug, just slightly confusing to read.
2. When `truncateWritableHandleLocked` fails, the `continue` failover is correct, but consider using a warning log level instead of `log.Printf`.

### Fix 2: `getdents01` — readdir missing `.` / `..` ✅

- Offset scheme (`.`=1, `..`=2, real entry `i`=`i+3`) is correct and stable across pagination
- ReadDirPlus uses `FUSE_UNKNOWN_INO` without `IncrementLookup` — correct Nlookup accounting
- `parentDirIno` handles trailing slash correctly via `stripTrailingSlash`
- Root is its own parent — handled

**Nits:**
3. `dotDotEntries` allocates per-call. Could use `[2]gofuse.DirEntry` to avoid heap alloc, but cost is negligible as noted.
4. VaultFS has the same `.`/`..` gap — consider adding a TODO or opening an issue to track the follow-up.

### Test coverage

Excellent coverage: first page / resume page / small-buffer pagination / root parent / ReadDirPlus Nlookup invariant / single-handle truncate / multi-handle truncate / zero remote calls assertion / native endian wire format parsing.

PR #691, opened 2026-07-07
Finalize uploads after post-S3 context cancellation

srstack:
## Summary
- detach the tenant metadata finalize transaction from request cancellation after S3 CompleteMultipartUpload succeeds
- preserve request-scoped trace values while applying an upload-finalize timeout
- use fresh detached contexts for metadata cleanup, quota reservation abort, and overwrite blob GC
- add a regression test where S3 complete succeeds and then cancels the original request context before metadata finalize runs

## Tests
- go test ./pkg/backend -run 'TestConfirmUploadFinalizesAfterRequestContextCanceledPostS3Complete|TestInitiateAndConfirmUpload|TestFinalizeUploadRejectsRootTargetBeforeS3Complete' -count=1
- go test ./pkg/backend -count=1
- make lint

### DIFF
--- a/pkg/backend/upload.go
+++ b/pkg/backend/upload.go
@@ -1017,6 +1017,8 @@ func (b *Dat9Backend) finalizeUpload(ctx context.Context, upload *datastore.Uplo
 		return fmt.Errorf("complete multipart: %w", err)
 	}
 	completeMultipartDurationMs := uploadPhaseMs(completeMultipartStart)
+	finalizeCtx, cancelFinalize := postCommitQuotaMutationContext()
+	defer cancelFinalize()
 
 	var oldStorageRef string
 	var oldStorageType datastore.StorageType
@@ -1042,14 +1044,14 @@ func (b *Dat9Backend) finalizeUpload(ctx context.Context, upload *datastore.Uplo
 	enqueueUploadSemanticTasks := func(tx *sql.Tx, mediaDelta int64) error {
 		stepStart := time.Now()
 		if b.UsesDatabaseAutoEmbedding() {
-			created, err := b.enqueueExtractSemanticTasksTx(ctx, tx, confirmedFileID, confirmedRevision, upload.TargetPath, contentType, mediaDelta)
+			created, err := b.enqueueExtractSemanticTasksTx(finalizeCtx, tx, confirmedFileID, confirmedRevision, upload.TargetPath, contentType, mediaDelta)
 			semanticTaskEnqueued = created
 			semanticEnqueueDurationMs = uploadPhaseMs(stepStart)
 			return err
 		}
 		// App-embedding mode: image/audio extract tasks are durable and
 		// independent of EMBED_TEXT, so register them in the same transaction.
-		extractCreated, err := b.enqueueExtractSemanticTasksTx(ctx, tx, confirmedFileID, confirmedRevision, upload.TargetPath, contentType, mediaDelta)
+		extractCreated, err := b.enqueueExtractSemanticTasksTx(finalizeCtx, tx, confirmedFileID, confirmedRevision, upload.TargetPath, contentType, mediaDelta)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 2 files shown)


srstack:
## Review

Clean, well-scoped fix. Once S3 `CompleteMultipartUpload` succeeds the blob is durably committed, so the tenant-metadata finalize must no longer be tied to the (possibly-cancelled) request context — otherwise a client disconnect between S3 complete and the metadata tx leaves the upload half-finalized (blob in S3, no `file_node`). The PR introduces `finalizeCtx` right after S3 complete (a detached `backgroundWithTrace()` + 30s timeout) and threads it through the metadata `InTx`, the semantic-task enqueue, and the overwrite-blob cleanup.

Verified:
- The detach is applied consistently — every post-S3 **mutation** now runs on a detached context: metadata `InTx` (`finalizeCtx`), semantic enqueue (`finalizeCtx`), overwrite `deleteBlobIfS3Ctx` (`finalizeCtx`), the failure-path cleanup + reservation abort (`cleanupCtx`), and `completeUploadReservation` (`quotaCtx`). The remaining `ctx` uses are logging only (`logger.Error`/`InfoBenchTiming`), which is harmless.
- No use-after-cancel: `defer cancelFinalize()` fires only at function return, and `ConfirmUpload` awaits `finalizeUpload` synchronously, so detaching cancellation doesn't spawn an unawaited goroutine or leak the DB tx.
- The 30s bound is ample for a metadata tx and strictly safer than inheriting an already-cancelled deadline.
- The regression test is faithful: `cancelAfterCompleteS3Client` cancels the request context inside `CompleteMultipartUpload` (exactly the post-S3 window), and the assertions confirm the upload still reaches `UploadCompleted` with the file `Stat`-able at full size.

`go vet ./pkg/backend` clean.

Minor (non-blocking): `finalizeCtx` is produced by `postCommitQuotaMutationContext()`, whose name now undersells its role — it's the general post-S3 finalize context, not just a quota mutation. A rename (or a thin `postS3FinalizeContext()` alias) would read better, but behavior is identical.

LGTM.

mornyx (approved):
Reviewed PR #691: Finalize uploads after post-S3 context cancellation.

**Summary of review:**

This PR detaches the tenant metadata finalize transaction (and related cleanup/GC operations) from request context cancellation after S3 CompleteMultipartUpload succeeds. Previously, if the client disconnected after S3 complete but before the metadata transaction committed, the request context would be canceled and the finalize would fail -- leaving the S3 object uploaded but the tenant metadata uncommitted.

**Key changes verified:**

1. **New postS3UploadFinalizeContext**: Uses context.WithoutCancel(ctx) to preserve trace values from the original request while being immune to request cancellation, with a 2-minute timeout. This replaces the old postCommitQuotaMutationContext() (which used a completely fresh background context with 30s timeout, losing trace correlation) for all post-S3-complete paths in finalizeUpload.

2. **All post-S3-complete paths now use detached context**:
   - S3 complete failure cleanup (line 1007): reservation reset uses detached ctx
   - Metadata finalize transaction (line 1075): InTx uses finalizeCtx
   - Semantic task enqueue (lines 1053, 1059): uses finalizeCtx
   - Finalize tx failure cleanup (line 1229): cleanupFailedFinalizeUpload + abortUploadReservation both use detached ctx
   - Quota reservation completion (line 1244): uses detached ctx
   - Overwrite blob GC (line 1252): deleteBlobIfS3Ctx uses detached ctx

3. **S3 complete itself still uses original ctx**: The CompleteMultipartUpload call at line 1007 still uses the request ctx -- correct, since if the request is canceled before S3 complete, the upload should be aborted, not finalized.

4. **Timeout change 30s -> 2min**: The old postCommitQuotaMutationTimeout was 30s; the new postS3UploadFinalizeTimeout is 2min. This is appropriate -- the finalize transaction involves a tenant DB commit, semantic task enqueue, quota reservation completion, and potentially blob GC. 30s could be tight under load; 2min gives ample headroom while still bounding resource usage.

5. **context.WithoutCancel**: Go 1.21+ stdlib function. The repo uses Go 1.25.1, so this is available. Correct usage -- creates a new context that is never canceled by the parent's cancellation but inherits all values.

6. **Regression test**: TestConfirmUploadFinalizesAfterRequestContextCanceledPostS3Complete wraps the S3 client to cancel the request context immediately after CompleteMultipartUpload succeeds, then verifies that ConfirmUpload still succeeds and the file is fully committed. Good coverage of the exact bug scenario.

7. **No remaining postCommitQuotaMutationContext references in upload.go**: All four call sites in finalizeUpload have been migrated to postS3UploadFinalizeContext. The old function remains for other callers (fs_layer_content.go, dat9.go) which are not affected by this bug.

**Verification:**
- go build ./pkg/... ./cmd/... -- passes
- go vet ./pkg/backend/... -- passes
- git diff --check origin/main...HEAD -- clean
- CI: CodeRabbit pass, ci pass (e2e pending at review time)
- Codex (glm-5.2, xhigh) ran -- no blockers identified

PR #694, opened 2026-07-07
cd: multi-arch image build (amd64 + arm64)

srstack:
Build both amd64 and arm64 images in dev-cd, create a multi-arch manifest for dev cluster ARM64 nodes.

### DIFF
--- a/.github/workflows/dev-cd.yml
+++ b/.github/workflows/dev-cd.yml
@@ -73,13 +73,15 @@ jobs:
       - name: Build and push image
         run: |
           IMAGE_BASE=${{ env.ECR_REGISTRY }}/${{ env.ECR_REPOSITORY }}
-          # amd64
+          # amd64 (native)
           make docker-build IMAGE_REPO=$IMAGE_BASE IMAGE_TAG=${{ env.IMAGE_TAG }}-amd64 GOARCH=amd64
           docker push $IMAGE_BASE:${{ env.IMAGE_TAG }}-amd64
-          # arm64
-          make docker-build IMAGE_REPO=$IMAGE_BASE IMAGE_TAG=${{ env.IMAGE_TAG }}-arm64 GOARCH=arm64
+          # arm64 — cross-compile binary, then build arm64 Docker image
+          mkdir -p bin
+          CGO_ENABLED=0 GOARCH=arm64 go build -o bin/drive9-server ./cmd/drive9-server
+          docker build --platform linux/arm64 -t $IMAGE_BASE:${{ env.IMAGE_TAG }}-arm64 .
           docker push $IMAGE_BASE:${{ env.IMAGE_TAG }}-arm64
-          # manifest
+          # multi-arch manifest
           docker manifest create $IMAGE_BASE:${{ env.IMAGE_TAG }} \
             $IMAGE_BASE:${{ env.IMAGE_TAG }}-amd64 \
             $IMAGE_BASE:${{ env.IMAGE_TAG }}-arm64
--- a/.github/workflows/dev-cd.yml
+++ b/.github/workflows/dev-cd.yml
@@ -72,8 +72,18 @@ jobs:
 
       - name: Build and push image
         run: |
-          make docker-build IMAGE_REPO=${{ env.ECR_REGISTRY }}/${{ env.ECR_REPOSITORY }} IMAGE_TAG=${{ env.IMAGE_TAG }}
-          docker push ${{ env.ECR_REGISTRY }}/${{ env.ECR_REPOSITORY }}:${{ env.IMAGE_TAG }}
+          IMAGE_BASE=${{ env.ECR_REGISTRY }}/${{ env.ECR_REPOSITORY }}
+          # amd64
+          make docker-build IMAGE_REPO=$IMAGE_BASE IMAGE_TAG=${{ env.IMAGE_TAG }}-amd64 GOARCH=amd64
+          docker push $IMAGE_BASE:${{ env.IMAGE_TAG }}-amd64
+          # arm64
+          make docker-build IMAGE_REPO=$IMAGE_BASE IMAGE_TAG=${{ env.IMAGE_TAG }}-arm64 GOARCH=arm64
+          docker push $IMAGE_BASE:${{ env.IMAGE_TAG }}-arm64
+          # manifest
+          docker manifest create $IMAGE_BASE:${{ env.IMAGE_TAG }} \
+            $IMAGE_BASE:${{ env.IMAGE_TAG }}-amd64 \
+            $IMAGE_BASE:${{ env.IMAGE_TAG }}-arm64
+          docker manifest push $IMAGE_BASE:${{ env.IMAGE_TAG }}
 
       - name: Update kubeconfig
         run: |
--- a/.github/workflows/dev-cd.yml
+++ b/.github/workflows/dev-cd.yml
@@ -76,20 +76,18 @@ jobs:
       - name: Build and push image
         run: |
           IMAGE_BASE=${{ env.ECR_REGISTRY }}/${{ env.ECR_REPOSITORY }}
-          LDFLAGS="-X github.com/mem9-ai/drive9/pkg/buildinfo.Version=${{ env.IMAGE_TAG }} -X github.com/mem9-ai/drive9/pkg/buildinfo.GitHash=$(git rev-parse --short=7 HEAD) -X github.com/mem9-ai/drive9/pkg/buildinfo.BuildTime=$(date -u '+%Y-%m-%dT%H:%M:%SZ')"
+          VERSION=${{ env.IMAGE_TAG }}
           # amd64
-          make docker-build IMAGE_REPO=$IMAGE_BASE IMAGE_TAG=${{ env.IMAGE_TAG }}-amd64 GOARCH=amd64
-          docker push $IMAGE_BASE:${{ env.IMAGE_TAG }}-amd64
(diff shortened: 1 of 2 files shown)


srstack:
## Review

The PR reworks the `dev-cd.yml` build step to produce `-amd64` / `-arm64` tags and stitch them into a multi-arch manifest so ARM64 dev-cluster nodes can pull the image. The intent is right, but the mechanism doesn't actually build an arm64 image — both tags end up as `linux/amd64`, so the goal isn't met.

### Findings

**1. `.github/workflows/dev-cd.yml` (the `-arm64` build) — the "arm64" image is built as `linux/amd64`, so the manifest has no real arm64 entry and ARM64 nodes fail to pull. (blocking)**
`make docker-build GOARCH=arm64` only changes the Go cross-compile target (`Makefile`: `GOOS=$(GOOS) GOARCH=$(GOARCH) go build`). The image itself comes from `docker-build:` → `docker build -t $(IMAGE) .` with **no `--platform` and no buildx**, and the `Dockerfile` is `FROM …/alpine:3.19` with no `--platform=$TARGETPLATFORM`. The job runs on `ubuntu-latest` (amd64) with no `setup-qemu`/`setup-buildx` steps, so `docker build` produces an image whose config `architecture` is `amd64` (amd64 alpine base) regardless of `GOARCH`. Result:
- `:$TAG-arm64` is a `linux/amd64` manifest containing an arm64 binary on an amd64 base — broken either way.
- `docker manifest create` reads both children as `amd64`, producing a manifest list with two amd64 entries and **no `linux/arm64`**. An arm64 node pulling `:$TAG` gets `no matching manifest for linux/arm64` → `ImagePullBackOff` — exactly the scenario this PR targets.

Fix: build real multi-arch images with buildx + QEMU. Add `docker/setup-qemu-action` + `docker/setup-buildx-action`, then a single `docker buildx build --platform linux/amd64,linux/arm64 --push -t $IMAGE_BASE:$TAG .`, and make the Dockerfile arch-aware.

**2. `Dockerfile` — no `$TARGETPLATFORM`/`$TARGETARCH` awareness, so it can't participate in a cross-platform build even after buildx is added. (part of the same fix)**
`FROM public.ecr.aws/docker/library/alpine:3.19` + `COPY ./bin/drive9-server` resolve to the build host arch. For a working arm64 image use `FROM --platform=$TARGETPLATFORM …` and copy the arm64 binary (e.g. emit `./bin/drive9-server-$TARGETARCH` from the Makefile and `COPY ./bin/drive9-server-${TARGETARCH}`, or build inside the image keyed on `$TARGETARCH`).

**3. `docker manifest push` — gated behind the experimental manifest CLI on some Docker versions. (minor)**
Child pushes precede it, so the registry-manifest requirement is met, but this whole block disappears if you switch to `buildx …--push` per #1, so it's moot under the recommended fix.

### Verdict
Requesting changes — the current `GOARCH` cross-compile + plain `docker build` approach can't set the image manifest platform, so ARM64 nodes would hit `ImagePullBackOff`. The fix (buildx + QEMU + arch-aware Dockerfile) is well-understood; happy to write the corrected build step and Dockerfile if useful.

srstack:
Fixed: arm64 image now uses `docker build --platform linux/arm64` so base Alpine layer is also ARM64, not just the binary.

Regarding prod promotion: prod clusters are all AMD64-only currently. When they adopt ARM64, the prod-cd workflow will need similar manifest handling. This PR only touches dev-cd.

srstack:
## Re-review (after `d0c34dc6` — `--platform linux/arm64`)

The core of finding #1 is addressed: `docker build --platform linux/arm64` makes BuildKit stamp the image config architecture as `arm64`, so `docker manifest create` now yields a real amd64 + arm64 manifest list instead of two amd64 entries. Good fix. Two items remain.

**1. `.github/workflows/dev-cd.yml` — the arm64 build likely needs QEMU/binfmt, which the job never sets up. (PLAUSIBLE, potentially blocking)**
The `Dockerfile` has `RUN apk add --no-cache ca-certificates`. Under `docker build --platform linux/arm64` that RUN executes inside the arm64 base image, which requires arm64 emulation on the amd64 `ubuntu-latest` runner. The job has no `docker/setup-qemu-action` step (grep across `.github/workflows/` finds no `setup-qemu`/`setup-buildx`/`binfmt` anywhere), so if the runner doesn't have binfmt pre-registered, this step fails with `exec format error` and dev-cd breaks for everyone — worse than the current amd64-only state. What would confirm it: whether the runner has `/proc/sys/fs/binfmt_misc/qemu-aarch64` registered by default; the standard/safe answer is to add `docker/setup-qemu-action@v3` before the build. (Switching the whole block to `docker buildx build --platform linux/amd64,linux/arm64 --push` would also handle QEMU setup and collapse the 3 build/push/manifest steps into one.)

**2. `.github/workflows/dev-cd.yml` — the inline arm64 `go build` drops the buildinfo ldflags, so the two arch images carry different build metadata. (minor)**
The amd64 path goes through `make docker-build` → `build-server`, which builds with `-ldflags "$(BUILDINFO_LDFLAGS)"` (Version/GitBranch/BuildTime). The new arm64 path is a bare `CGO_ENABLED=0 GOARCH=arm64 go build -o bin/drive9-server ./cmd/drive9-server` with no ldflags, so the arm64 image reports `Version=dev` / empty git branch / empty build time while amd64 reports the real values. Same manifest tag, divergent `buildinfo`. Prefer `make docker-build IMAGE_TAG=...-arm64 GOARCH=arm64 DOCKER_BUILD_ARGS="--platform linux/arm64"` (or thread `--platform` through the Makefile) so both arches build identically — or at minimum add the ldflags to the inline build.

### Verdict
Finding #1's manifest issue is resolved. Still holding off on approval until the QEMU/binfmt question (item 1) is settled — if the runner lacks binfmt, the arm64 `RUN apk add` fails and this regresses dev-cd. Item 2 is a minor consistency nit. Happy to convert the block to a single `buildx --platform=…,… --push`, which fixes both at once.

qiffang (approved):

srstack:
## Re-review (after `e8dc9356` — QEMU setup + ldflags)

The blocking item is resolved: `docker/setup-qemu-action@v3` now runs before the build, so the arm64 `RUN apk add …` executes under emulation on the amd64 runner instead of failing with `exec format error`. That was my one hard blocker — good.

One residual, and it's the reverse of what I flagged last time: the two arch images now stamp **different** `buildinfo` values under the same manifest tag.

**`.github/workflows/dev-cd.yml` — amd64 vs arm64 buildinfo diverges three ways. (minor)**
- amd64 goes through `make docker-build` → `BUILDINFO_LDFLAGS`, where `Version=$(if $(VERSION),$(VERSION),dev)` and `VERSION ?=` is empty (Makefile:26) → **`Version=dev`**; `GitHash=$(git rev-parse HEAD)` → **full 40-char**; `GitBranch` set.
- arm64 uses the inline `LDFLAGS` → `Version=${{ env.IMAGE_TAG }}` (**real tag**, not `dev`); `GitHash=$(git rev-parse --short=7 HEAD)` (**7-char short**); **`GitBranch` omitted** (stays `"unknown"`).

So `/version` (or whatever surfaces `buildinfo`) reports different Version, GitHash length, and GitBranch depending on which arch a node happens to pull — same image tag, inconsistent metadata. Not a runtime bug, but confusing for debugging. Cleanest fix: pass `--platform` through `make docker-build` for both arches (e.g. `make docker-build GOARCH=arm64 DOCKER_BUILD_ARGS="--platform linux/arm64"`) so both build via the same `BUILDINFO_LDFLAGS` path; or set `VERSION=${{ env.IMAGE_TAG }}` on the amd64 `make` call and mirror the exact same ldflags (incl. `GitBranch`, full `GitHash`) on arm64.

### Verdict
Functionally sound now — real amd64 + arm64 manifest, QEMU in place. The only remaining item is the cosmetic buildinfo divergence; I'd fix it for clean `/version` output but it doesn't block the multi-arch goal. (Can't submit a formal GitHub approval here since it's your own PR, but from a review standpoint this is good-to-merge once the metadata nit is addressed or consciously accepted.)

srstack:
## Re-review (after `216c448c` — unify via `make docker-build` + `DOCKER_BUILD_ARGS`)

This resolves the last residual cleanly, and it's the right shape — both arches now build through the single `make docker-build` path instead of a bespoke inline `go build`.

Verified across all three rounds' findings:
- **Manifest arch (round 1)** — `--platform linux/arm64` threaded through `DOCKER_BUILD_ARGS` makes BuildKit stamp the arm64 image config, so `docker manifest create` produces a genuine amd64 + arm64 list. ✅
- **QEMU/binfmt (round 2)** — `docker/setup-qemu-action@v3` runs before the build, so the arm64 `RUN apk add …` executes under emulation on the amd64 runner. ✅
- **buildinfo divergence (round 3)** — gone. Both arches call `make docker-build … VERSION=$VERSION GIT_BRANCH=${{ github.ref_name }}`, so they share the same `BUILDINFO_LDFLAGS`: identical `Version` (= tag), full 40-char `GitHash` (Makefile default, not overridden), real `GitBranch` (= `github.ref_name` rather than the checkout's detached `HEAD`), and `BuildTime`. `/version` is now consistent regardless of which arch a node pulls. ✅

`DOCKER_BUILD_ARGS ?=` defaults empty, so the amd64 path and every other `docker-build` caller are unchanged — the new knob only adds `--platform` for arm64. Removing the duplicated inline build also eliminates the drift risk (Makefile ldflags and workflow can no longer diverge). This is the generalized fix rather than a special-case bandaid.

LGTM — all findings resolved. (Formal GitHub approval isn't possible from this account since it's your own PR, but from a review standpoint this is good to merge.)

PR #696, opened 2026-07-07
fix(ci): avoid secret masking on job outputs in prod-cd

srstack:
## Problem

resolve job outputs `image` and `registry` contain the full ECR URL which matches `secrets.ECR_REGISTRY`, causing GitHub Actions to suppress both outputs:

```
##[warning]Skip output 'image' since it may contain secret.
##[warning]Skip output 'registry' since it may contain secret.
```

deploy job then receives empty `needs.resolve.outputs.image` /` .registry`, and `docker pull` fails.

## Fix

- resolve job now outputs only `tag` (safe string) instead of `image`/`registry`
- deploy job constructs the full `IMAGE` at template-expansion time from `env.ECR_REGISTRY` + `env.ECR_REPOSITORY` + `needs.resolve.outputs.tag`, bypassing job output propagation entirely
- `csi-publish` job unchanged (uses `drive9_sha`)

## Summary by CodeRabbit

* **Chores**
  * Updated the production deployment workflow to use a simpler image tagging flow.
  * Improved how deployment steps assemble the container image reference and authenticate to the registry.

### DIFF
--- a/.github/workflows/prod-cd.yml
+++ b/.github/workflows/prod-cd.yml
@@ -42,8 +42,7 @@ jobs:
       id-token: write
       contents: read
     outputs:
-      image: ${{ steps.resolve.outputs.image }}
-      registry: ${{ steps.resolve.outputs.registry }}
+      tag: ${{ steps.resolve.outputs.tag }}
       drive9_sha: ${{ steps.resolve.outputs.drive9_sha }}
     steps:
       - name: Configure AWS credentials
@@ -117,10 +116,8 @@ jobs:
             exit 1
           fi
 
-          REGISTRY="${IMAGE%%/*}"
           {
-            echo "image=$IMAGE"
-            echo "registry=$REGISTRY"
+            echo "tag=$TAG"
             echo "drive9_sha=$SHA"
           } >> "$GITHUB_OUTPUT"
 
@@ -132,8 +129,8 @@ jobs:
       id-token: write
       contents: read
     env:
-      IMAGE: ${{ needs.resolve.outputs.image }}
-      REGISTRY: ${{ needs.resolve.outputs.registry }}
+      TAG: ${{ needs.resolve.outputs.tag }}
+      IMAGE: ${{ env.ECR_REGISTRY }}/${{ env.ECR_REPOSITORY }}:${{ needs.resolve.outputs.tag }}
 
     steps:
       - name: Configure AWS credentials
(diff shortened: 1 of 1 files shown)


srstack:
## Review

The PR correctly diagnoses why `deploy` was pulling an empty image: the `resolve` job's `image`/`registry` outputs contained the full ECR URL, which matched `secrets.ECR_REGISTRY`, so GitHub masked (dropped) them. Emitting only the safe `tag` output is the right fix. But the way `IMAGE` is reconstructed in the `deploy` job uses a context that GitHub doesn't populate there, so it silently comes out malformed.

### Finding (blocking)

**`.github/workflows/prod-cd.yml:133` — the `env` context is unavailable in a job-level `env:` block, so `IMAGE` resolves to `/:<tag>` and every `docker pull "$IMAGE"` fails.**
```yaml
    env:
      TAG: ${{ needs.resolve.outputs.tag }}
      IMAGE: ${{ env.ECR_REGISTRY }}/${{ env.ECR_REPOSITORY }}:${{ needs.resolve.outputs.tag }}
```
Per GitHub's [context-availability table](https://docs.github.com/en/actions/learn-github-actions/contexts#context-availability), `jobs.<job_id>.env` may reference only `github`, `needs`, `strategy`, `matrix`, `vars`, `inputs` — **not `env`**. The `env` context is populated for step-level keys, not for the job-level `env:` map. So `${{ env.ECR_REGISTRY }}` / `${{ env.ECR_REPOSITORY }}` expand to empty strings and `IMAGE` becomes `/:<tag>`; `docker pull "/:<tag>"` fails — the same broken-pull outcome the PR is fixing, now via empty-registry rather than masked output.

(The old `IMAGE: ${{ needs.resolve.outputs.image }}` worked because `needs` *is* available at job-level env. And line 149's `"$ECR_REGISTRY"` works because the workflow-level `env:` is exported as a real shell variable inside steps — that's a different mechanism than the `${{ env.* }}` expression context.)

`$IMAGE` is consumed by ~10 steps in this job (147/150/162/179/200/217/237/260–263/306–308), so it needs to stay a job-wide value rather than a single step's local. Two clean options:

- **Reference workflow-level env via `github`/hardcode is not possible**, so instead resolve `IMAGE` once in an early step and export it:
  ```yaml
      - name: Resolve image ref
        run: echo "IMAGE=$ECR_REGISTRY/$ECR_REPOSITORY:${{ needs.resolve.outputs.tag }}" >> "$GITHUB_ENV"
  ```
  (`$ECR_REGISTRY`/`$ECR_REPOSITORY` are present as shell vars from the workflow-level `env:`.) Keep `TAG` as job-level env if still needed.
- Or have the `resolve` job output the registry/repo pieces `deploy` needs and build `IMAGE` from `${{ needs.resolve.outputs.* }}` at job-level (allowed) — but since the registry is the masked secret, the `$GITHUB_ENV` approach above is simpler and avoids re-triggering masking.

### Verdict
Requesting changes — as written the deploy job would pull `/:<tag>` and fail. The masking diagnosis and the `tag`-only output are correct; only the `IMAGE` reconstruction needs to move to where the `env` values are actually resolvable (a step, via `$GITHUB_ENV`).

PR #698, opened 2026-07-08
fix(cli): make `drive9 fs grep --json` actually work

mornyx:
## Problem

`drive9 fs grep --json` silently produces no output instead of JSON. The `GrepJSON` function in `cmd/drive9/cli/grep.go` was defined but never dispatched from `runFS` in `main.go` — the `case "grep"` only calls `cli.Grep`. Since `parseLayerFlag` does not recognize `--json`, the flag is left in the positional args and treated as the path, causing the server request to target a non-existent path and return empty results.

```text
$ drive9 fs grep "payment" --json
# empty output (should be JSON)
```

## Fix

- **`cmd/drive9/cli/grep.go`**: Add `parseGrepFlags` that extends `parseLayerFlag` with `--json` detection. Route JSON vs text output inside `Grep` based on the flag. Normalize nil result slice to `[]SearchResult{}` so empty results emit `[]` instead of `null`. Remove the dead `GrepJSON` function.
- **`cmd/drive9/main.go`**: Document `--json` in the `fs grep` usage line.
- **`cmd/drive9/cli/grep_test.go`** (new): 4 tests covering empty `[]` output, results JSON output, text fallback without `--json`, and `--json` + `--layer` combination.

## Verification

```text
$ drive9 fs grep "payment" --json
[
  {
    "path": "/test.txt",
    "name": "test.txt",
    "size_bytes": 91,
    "score": 0.01639344262295082
  }
]

$ drive9 fs grep "zzz_nonexistent" --json
[]
```

- `go build ./cmd/...` ✅
- `go test ./cmd/drive9/cli/...` all pass ✅
- `make lint` 0 issues ✅

## Summary by CodeRabbit

* **New Features**
  * Added an optional JSON output mode for file-content search, alongside the existing text output.
  * Search help now documents the new JSON flag.

* **Bug Fixes**
  * Empty search results in JSON mode now return a valid empty array instead of an empty response.
  * Search output remains consistent when a layer is specified.

* **Tests**
  * Added coverage for JSON output, text output, empty results, and layer handling.

### DIFF
--- a/cmd/drive9/cli/grep.go
+++ b/cmd/drive9/cli/grep.go
@@ -10,12 +10,12 @@ import (
 )
 
 func Grep(c *client.Client, args []string) error {
-	layerRef, args, err := parseLayerFlag(args)
+	layerRef, jsonMode, args, err := parseGrepFlags(args)
 	if err != nil {
 		return err
 	}
 	if len(args) < 1 {
-		return fmt.Errorf("usage: drive9 fs grep [--layer <ref>] <pattern> [path]")
+		return fmt.Errorf("usage: drive9 fs grep [--layer <ref>] [--json] <pattern> [path]")
 	}
 	query := args[0]
 	path := "/"
@@ -31,6 +31,14 @@ func Grep(c *client.Client, args []string) error {
 	results, err := c.GrepWithLayer(query, path, 20, layerRef)
 	if err != nil {
 		return err
+	}
+	if jsonMode {
+		if results == nil {
+			results = []client.SearchResult{}
+		}
+		enc := json.NewEncoder(os.Stdout)
+		enc.SetIndent("", "  ")
+		return enc.Encode(results)
 	}
 	if len(results) == 0 {
 		return nil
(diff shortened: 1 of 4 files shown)


srstack:
## Review

`drive9 fs grep --json` silently produced no output because `runFS` only ever called `cli.Grep` (never the separate `GrepJSON`), and `parseLayerFlag` didn't recognize `--json` — so the flag fell through into the positional args and was treated as the search path, targeting a nonexistent path and returning empty. This PR fixes it by teaching `Grep` about `--json` directly (`parseGrepFlags` = `parseLayerFlag` + `--json` stripping), branching to JSON vs text output inside the single dispatched function, normalizing a nil result slice to `[]SearchResult{}`, and deleting the dead `GrepJSON`.

Verified:
- The dispatch is now correct: `main.go` routes `grep` → `cli.Grep`, and `Grep` handles `--json` internally, so the flag reaches the JSON path (the actual root cause). `GrepJSON` is fully removed with no remaining references.
- `parseGrepFlags` layers `--json` detection *after* `parseLayerFlag`, so `--layer <ref>` is stripped first (consuming its value) and `--json` is filtered from what remains — `--json --layer task=search needle path` parses correctly (covered by `TestGrepJSONWithLayer`, which also asserts the `layer` query param propagates).
- nil→`[]client.SearchResult{}` makes empty results emit `[]` rather than `null` — correct for JSON consumers and covered by `TestGrepJSONEmptyResultsOutputsArray` + the e2e `jq -c … == "[]"` check.
- Text path is unchanged when `--json` is absent (`TestGrepTextOutputWithoutJSON`), and the `--json` block returns before the text formatting, so no double output.
- Good test coverage: unit tests for empty/results/text/layer plus e2e smoke assertions (`jq -e "type == \"array\""`, target-in-array, empty `[]`). Usage string updated in `main.go`.

This is the right altitude — folding `--json` into the one dispatched `Grep` rather than maintaining a parallel `GrepJSON` that main.go never called. No findings.

LGTM.

srstack (approved):
LGTM. Correctly fixes the root cause — --json now flows through the single dispatched cli.Grep via parseGrepFlags (layered on parseLayerFlag), instead of the never-called GrepJSON. nil→[]SearchResult{} gives valid [] for empty results, dead code removed, and coverage is solid (unit tests for empty/results/text/layer + e2e jq assertions).

PR #707, opened 2026-07-08
fix(fuse): pjdfstest chown/symlink owner + O_TRUNC mtime fixes

mornyx:
## Summary

Fixes all 14 POSIX compliance failures found via `community.pjdfstest` blackbox testing on local server. **Pass rate: 99.84% → 100%** (8798/8798 tests pass, 0 non-TODO failures).

### 1. Symlink handler missing `UpdateOwner` (9 failures — `chown/00.t`)

The `Symlink` FUSE handler created the inode entry without calling `UpdateOwner`, unlike `Create`, `Mkdir`, and `Mknod`. This caused `lstat` on a symlink to report `fs.uid` instead of the caller's uid.

**Fix:** Added `UpdateOwner` in the `Symlink` handler, matching the pattern in `Create`/`Mkdir`/`Mknod`.

### 2. `chown(-1,-1)` preserve sentinel mishandling

The Linux kernel encodes `chown(path, -1, -1)` as `FATTR_UID|FATTR_GID` with `Uid=Gid=0xFFFFFFFF`. Added `resolveSetAttrOwner()` helper to normalize the sentinel to "preserve" at all 3 `SetAttr` call sites.

### 3. SetAttr FATTR_SIZE truncate not updating mtime (1 failure — `open/00.t` test 43) — **Root cause**

The Linux FUSE kernel translates `open(O_TRUNC)` into `SetAttr(FATTR_SIZE, 0)` + `Open` (without `O_TRUNC` in flags). The SetAttr `FATTR_SIZE` path only updated ctime (via `metadataChanged`) but never updated mtime. POSIX requires truncation to update both mtime and ctime.

**Fix:** Added explicit `UpdateMtime` in the SetAttr `FATTR_SIZE` path when size changes.

### 4. `GetAttr` clobbering locally-updated mtime

After mtime is updated locally, `GetAttr`'s remote stat, `cachedAttrEntry`, `pendingIndex`, and `writeBack` paths overwrote it with stale values. Applied "only adopt if newer" guards to all 4 paths.

### 5. Flush Path 2 not advancing inode mtime

After a successful upload in Flush Path 2 (direct PUT), the inode's own mtime was not advanced. Added `UpdateMtime` to advance the inode mtime to at least the commit time.

### 6. Shadow store ENOSPC on sparse writes (3 failures — `open/25.t`)

`WriteAt` checked disk quota against the logical file extension delta (`end - oldSize`), but shadow files are sparse — a 1-byte write at 2GB+1 only allocates one block. Changed quota check to use actual data length.

### 7. Rmdir ENOTEMPTY with stale dir cache (1 failure — `unlink/14.t` test 7)

`hasKnownLocalDirectoryChildren` returned ENOTEMPTY when the dir cache had stale positive entries for recently-unlinked files (re-populated by SSE events). Added `HasPositiveEntriesExceptTombstones` to cross-check against delete tombstones.

## Test results (drive9-test, local server mode)

| Test file | Before | After |
|---|---|---|
| `chown/00.t` | 9 failed | **0** ✅ |
| `open/00.t` | 1 failed | **0** ✅ |
| `open/25.t` | 3 failed | **0** ✅ |
| `unlink/14.t` | 1 failed | **0** ✅ |
| **Total** | **14 / 8798** | **0 / 8798 (100%)** ✅ |

## Unit tests added

- `TestSetAttr_ChownPreserveSentinel` — `chown(-1,-1)` by non-root preserves `0:0` owner
- `TestSetAttr_ChownPreserveGIDOnly` — `chown(-1,100)` preserves uid, changes gid
- `TestOpenTruncUpdatesMtimeAndCtime` — `Open(O_TRUNC)` bumps mtime and ctime
- `TestSymlinkStoresCallerOwner` — symlink entry carries caller's uid/gid

All pass. `make lint` clean (0 issues).

## Summary by CodeRabbit

* **New Features**
  * Improved file and directory metadata handling for more accurate timestamps and ownership behavior.
  * Added a DirCache helper to better determine when directories have positive entries beyond tombstones.
  * Improved support for sparse writes, reducing unnecessary quota rejections.

* **Bug Fixes**
  * Fixed truncation and open-with-truncate behavior to correctly advance mtime/ctime and synchronize kernel attribute updates.
  * Corrected mtime merging during attribute refresh so remote timestamps no longer overwrite newer local updates.
  * Fixed directory removal checks to avoid incorrectly treating tombstone-covered children as “non-empty.”
  * Normalized chown preserve-sentinel behavior so preserved UID/GID semantics remain consistent, including for symlinks and SetAttr.

### DIFF
--- a/pkg/fuse/dat9fs.go
+++ b/pkg/fuse/dat9fs.go
@@ -5073,7 +5073,14 @@ func (fs *Dat9FS) hasKnownLocalDirectoryChildren(dirPath string) bool {
 		return false
 	}
 	if fs.dirCache != nil && fs.dirCache.HasPositiveEntries(dirPath) {
-		return true
+		// The dirCache may contain stale positive entries for recently
+		// unlinked files (re-populated by SSE events or remote listings).
+		// Cross-check against delete tombstones before declaring the
+		// directory non-empty — if every positive entry is tombstoned,
+		// the directory is effectively empty.
+		if fs.dirCache.HasPositiveEntriesExceptTombstones(dirPath, fs.snapshotDeletedPaths()) {
+			return true
+		}
 	}
 	for _, entry := range fs.inodes.Snapshot() {
 		if entry.Unlinked {
@@ -6278,7 +6285,7 @@ func (fs *Dat9FS) GetAttr(cancel <-chan struct{}, input *gofuse.GetAttrIn, out *
 		if fs.pendingIndex != nil {
 			if meta, ok := fs.pendingIndex.GetMeta(entry.Path); ok {
 				entry.Size = meta.Size
-				if !meta.Mtime.IsZero() {
+				if !meta.Mtime.IsZero() && (entry.Mtime.IsZero() || meta.Mtime.After(entry.Mtime)) {
 					entry.Mtime = meta.Mtime
 				}
 				pendingFound = true
@@ -6287,7 +6294,7 @@ func (fs *Dat9FS) GetAttr(cancel <-chan struct{}, input *gofuse.GetAttrIn, out *
 		if !pendingFound {
 			if meta, ok := fs.writeBack.GetMeta(entry.Path); ok {
 				entry.Size = meta.Size
-				if !meta.Mtime.IsZero() {
+				if !meta.Mtime.IsZero() && (entry.Mtime.IsZero() || meta.Mtime.After(entry.Mtime)) {
 					entry.Mtime = meta.Mtime
 				}
 				pendingFound = true
--- a/pkg/fuse/dat9fs.go
+++ b/pkg/fuse/dat9fs.go
@@ -6562,6 +6562,29 @@ func (fs *Dat9FS) checkSetAttrOwnerForCaller(input *gofuse.SetAttrIn, entry *Ino
 	return gofuse.OK
 }
 
+// chownPreserveID is the sentinel value (uint32(-1)) the Linux kernel sends
+// via FATTR_UID/FATTR_GID to indicate "do not change this field" in a chown
+// call (e.g. chown(path, -1, gid) or chown(path, -1, -1)).
+const chownPreserveID = ^uint32(0)
+
+// resolveSetAttrOwner reads the requested uid/gid from input, treating the
+// chown(-1) preserve sentinel (0xFFFFFFFF) as "not set" so that callers do
+// not overwrite the existing owner with a literal 0xFFFFFFFF. The Linux kernel
+// always sets FATTR_UID/FATTR_GID when chown is invoked, even for fields the
+// caller wants to preserve, so GetUID()/GetGID() report hasUID/hasGID=true
+// with the sentinel value — this helper normalizes that back to hasUID=false.
+func resolveSetAttrOwner(input *gofuse.SetAttrIn) (uid uint32, hasUID bool, gid uint32, hasGID bool) {
+	uid, hasUID = input.GetUID()
+	gid, hasGID = input.GetGID()
+	if hasUID && uid == chownPreserveID {
+		hasUID = false
+	}
+	if hasGID && gid == chownPreserveID {
+		hasGID = false
+	}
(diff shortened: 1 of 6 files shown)


qiffang:
## Review

This PR is pointed at real pjdfstest failures, and most of the fixes look directionally correct: chown `0xffffffff` sentinel handling, symlink owner initialization, SetAttr truncate mtime, and the stale-dir tombstone idea all address real POSIX/FUSE compatibility gaps. CI is green, and the focused tests I ran passed:

- `go test ./pkg/fuse -run 'TestSetAttr_ChownPreserve|TestOpenTruncUpdatesMtimeAndCtime|TestSymlinkStoresCallerOwner|TestRmdirSucceedsWithTombstoneFilteringStaleRemoteListing|TestEnsureQuotaEnforcement|TestDefaultByteQuotaEnforcesENOSPC' -count=1`
- `go test ./pkg/fuse -run 'TestRmdirSucceedsAfterUnlinkWithStaleRemoteList|TestRmdirSucceedsAfterFullUnlinkWithStaleRemoteList|TestRmdirSucceedsWithTombstoneFilteringStaleRemoteListing|TestCheckWriteBackQuotaThrottled|TestEnsureQuotaEnforcement' -count=1`
- `go test ./pkg/fuse -run 'TestSetAttr_MtimeUpdate|TestSetAttr_MtimeNow|TestLookup_UsesMtimeFromStat|TestSetAttrSymlinkTimesSurviveGetAttr' -count=1`

I would not merge current head as-is because the sparse-write quota fix is internally inconsistent.

### Blocking

`pkg/fuse/shadow.go` now checks sparse `WriteAt` quota with `len(data)`, but then still increments `pendingBytes` by logical file-size growth (`newSize - prevSize`). That means the first sparse write may pass, but it poisons the byte quota for later writes.

Concrete reproduction on current head: with `NewShadowStoreWithQuota(dir, 0, 100)`, a 1-byte write at offset `1 << 20` passes because the precheck uses `required_bytes=1`; immediately after that, `pendingBytes` is `1048577`. A second 1-byte write at the next offset fails with `ENOSPC`, even though actual shadow-file disk allocation is still tiny. This also contradicts the field comment that `pendingBytes` is “current shadow bytes on disk.”

So the PR currently fixes only the first pjdfstest sparse write shape, not the shadow-store quota invariant. Please make the quota accounting consistent before merge:

- If byte quota is meant to be physical/allocation-ish, `pendingBytes` needs to track physical dirty bytes or extents consistently, not logical size growth.
- If byte quota is meant to be logical file size, the precheck should use logical delta and the PR should not claim sparse writes are accepted because they are sparse.
- Add a regression for: sparse 1-byte write beyond the byte quota, then another tiny write/append to the same sparse shadow file should not fail just because the first write inflated `pendingBytes` logically.

### Non-blocking notes

- I agree with CodeRabbit’s path-normalization warning in `HasPositiveEntriesExceptTombstones`. Current FUSE path invariants appear to pass non-trailing paths here, so I do not see it as a live blocker, but using the existing child-path helper or normalizing before tombstone lookup would make this robust and close the unresolved comment.
- The remote SetAttr `FATTR_SIZE` path now bumps mtime, but the local-overlay truncate branch still appears to only bump ctime. That is fine if this PR is scoped to the remote/pjdfstest path, but it is not a complete “all FUSE backends update mtime on truncate” fix.
- The “newer mtime wins” guards are reasonable for current Drive9 server mtimes being commit/confirm-time derived, but they rely on that semantic. If user-set mtimes become first-class later, a future local mtime can mask a legitimate older remote mtime.

Net: useful PR, but please fix the sparse shadow quota accounting before merge.

qiffang (approved):
GREEN — all 7 POSIX fixes verified (v2 sparse write quota change re-reviewed). No blockers.

qiffang (approved):
GREEN — all 7 POSIX fixes verified, including the v2 sparse-write quota fix. CI e2e smoke passed.

PR #709, opened 2026-07-08
fix(fuse): tolerate out-of-order readahead completion in prefetcher OnRead

mornyx:
## Problem

Single-stream sequential reads of large files (>4MB) on a mounted Drive9 FUSE filesystem run at only ~3-6 MB/s when remote fetch latency is high (~170-300ms), even though the backend is not the bottleneck (4 parallel streams aggregate ~10.8 MB/s).

## Root cause

`Prefetcher.OnRead` detected sequential access with a strict `offset == nextExpect` match. The kernel issues several reads ahead of the consumer position under async readahead (`FUSE_CAP_ASYNC_READ`, the default), and those reads complete in arbitrary order. Any out-of-order completion failed the equality, was classified as a random read, and reset the prefetch window — clearing all cached/inflight blocks and collapsing the window back to the minimum.

The prefetch pipeline that should hide remote fetch latency was reset on every interleaving, so the read degraded to one synchronous backend fetch per chunk.

Debug log evidence (original code, 100ms latency):
```
prefetch hit  offset=262144  next_expect=131072        # out-of-order: 262144 completed before 131072
prefetch random reset  offset=262144 next_expect=131072 # RESET — clears prefetched blocks
prefetch hit  offset=131072
prefetch random reset  offset=131072 next_expect=393216 # RESET again
prefetch miss  offset=393216                            # prefetch work thrown away, re-fetch from remote
```

## Fix

Treat a read as sequential when its offset stays within `[nextExpect-window, nextExpect+window]` — the region the prefetcher is tracking. Only a read that jumps outside that band (a genuine random seek, forward or backward) resets. Within the band:
- a read advancing `end` past `nextExpect` grows the window and triggers the next prefetch;
- a read landing on or behind the frontier (out-of-order completion, readahead re-read) is a no-op for window/frontier but still triggers prefetch if pending.

This preserves the existing random-read reset semantics (`TestPrefetcher_RandomReadResetsWindow` unchanged) while no longer punishing kernel readahead concurrency.

## Measurements

Remote tenant (api.drive9.ai), 32MB file, single-stream `dd bs=1M`, 200ms fetch latency injected via a `fetchRange` env hook to reproduce the production fetch-latency regime:

| version | throughput | random resets | prefetch hits |
|---|---|---|---|
| original | 5.1 MB/s | 25 | 226 |
| fixed | **14.0 MB/s** | **0** | 255 |

Low-latency environment (no injection) is unaffected: 34 → 38 MB/s, and no existing prefetch tests regressed.

## Tests

New `pkg/fuse/prefetch_test.go` (white-box, same package):
- `TestOnReadStrictOrderPreservesWindow` — in-order reads keep growing the window.
- `TestOnReadOutOfOrderDoesNotReset` — concurrent readahead completing out of order no longer resets.
- `TestOnReadGenuineBackwardJumpResets` — a seek far outside the tracked window still resets.
- `TestOnReadReReadSameRangeDoesNotInflateWindow` — re-reading a consumed range does not inflate the window.

### DIFF
--- a/pkg/fuse/prefetch.go
+++ b/pkg/fuse/prefetch.go
@@ -222,23 +222,42 @@ func (p *Prefetcher) OnRead(offset int64, size int) {
 		p.readSize = size
 	}
 
-	if offset == p.nextExpect {
-		// Sequential read detected — grow window
-		p.window *= 2
-		if p.window < int64(size) {
-			p.window = int64(size)
-		}
-		if p.window > prefetchMaxWindow {
-			p.window = prefetchMaxWindow
-		}
+	end := offset + int64(size)
 
-		// Trigger prefetch for next region if not already cached or inflight.
-		prefetchStart := offset + int64(size)
+	// A read is sequential if its offset stays within the region the
+	// prefetcher is tracking: no further than window bytes behind the
+	// consumed frontier (nextExpect) and no further than window bytes ahead of
+	// it. The kernel issues several reads ahead of the consumer position and
+	// they may complete in any order, so a strict offset==nextExpect match
+	// misclassified those out-of-order completions as random and reset the
+	// prefetch window on every interleaving, collapsing prefetch to
+	// single-stream fetch latency. A read that jumps outside the tracked
+	// region — forward beyond nextExpect+window or backward beyond
+	// nextExpect-window — is a genuine random seek and resets.
+	withinWindow := offset >= p.nextExpect-p.window && offset <= p.nextExpect+p.window
+	if withinWindow {
+		if end > p.nextExpect {
+			// Forward read advancing the frontier — grow the window and prefetch.
+			p.window *= 2
+			if p.window < int64(size) {
+				p.window = int64(size)
+			}
+			if p.window > prefetchMaxWindow {
+				p.window = prefetchMaxWindow
+			}
+			p.nextExpect = end
+		}
+		// Trigger prefetch for the next region beyond the frontier if not
+		// already cached or inflight. This also covers end==nextExpect (a read
+		// landing exactly at the frontier, e.g. the first read or a readahead
+		// re-read) and out-of-order completions behind the frontier that are
+		// within the window — neither advances the frontier, neither resets.
+		prefetchStart := p.nextExpect
 		if prefetchStart < p.fileSize && !p.inflight[prefetchStart] && p.cache[prefetchStart] == nil {
 			p.startPrefetch(prefetchStart, p.window)
 		}
 	} else {
-		// Random read — reset
+		// Random read outside the tracked region — reset.
 		if p.nextExpect != 0 || len(p.cache) > 0 || len(p.inflight) > 0 {
 			p.debugf("random reset path=%s offset=%d size=%d next_expect=%d cache=%d inflight=%d", p.pathString(), offset, size, p.nextExpect, len(p.cache), len(p.inflight))
 		}
@@ -250,9 +269,8 @@ func (p *Prefetcher) OnRead(offset int64, size int) {
 		for k := range p.inflight {
 			delete(p.inflight, k)
 		}
+		p.nextExpect = end
 	}
-
-	p.nextExpect = offset + int64(size)
 }
 
 // Close cancels all inflight prefetch goroutines and clears the cache.
--- a/pkg/fuse/prefetch_test.go
+++ b/pkg/fuse/prefetch_test.go
@@ -1,0 +1,135 @@
+package fuse
+
+import (
+	"testing"
+)
+
+// newTestPrefetcher builds a Prefetcher without a backing client so OnRead's
+// state machine can be exercised in isolation. startPrefetch is a no-op when
+// the client is nil, so only OnRead/Get bookkeeping is observed.
+func newTestPrefetcher(path string, fileSize int64) *Prefetcher {
+	return NewPrefetcher(nil, path, fileSize)
+}
+
+// TestOnReadStrictOrderPreservesWindow covers the original fast path: reads
+// arriving in exact offset order keep growing the window and advancing
+// nextExpect. This guards against the fix regressing the common case.
+func TestOnReadStrictOrderPreservesWindow(t *testing.T) {
+	t.Helper()
+	const chunk = 128 * 1024
+	p := newTestPrefetcher("/f", 32*1024*1024)
+	p.window = prefetchMinWindow
+
+	for off := int64(0); off < 4*chunk; off += chunk {
+		p.OnRead(off, chunk)
+	}
+
+	if p.nextExpect != 4*chunk {
+		t.Fatalf("nextExpect = %d, want %d", p.nextExpect, 4*chunk)
+	}
+	if p.window <= prefetchMinWindow {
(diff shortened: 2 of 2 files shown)


srstack:
## Review

The PR fixes a real single-stream throughput regression: `Prefetcher.OnRead` used a strict `offset == nextExpect` sequential check, so the kernel's async readahead (`FUSE_CAP_ASYNC_READ`) issuing reads that complete out of order was misclassified as random and reset the prefetch window on every interleaving — collapsing the pipeline to one synchronous fetch per chunk. The fix treats a read as sequential when its offset falls in `[nextExpect-window, nextExpect+window]`, only resetting on a genuine seek outside that band. Measurements (5.1 → 14.0 MB/s at 200ms latency, low-latency unaffected) and the new white-box tests are convincing.

Verified: `OnRead` holds `p.mu` for the whole body, so out-of-order completions are serialized (the premise of the fix). The frontier (`nextExpect`) only advances when `end > nextExpect`; out-of-order/behind-frontier reads within the band no-op the frontier/window but still trigger prefetch at the frontier; the random-reset path and `TestPrefetcher_RandomReadResetsWindow` semantics are preserved. Window growth stays capped at `prefetchMaxWindow`. The negative lower bound (`nextExpect-window` < 0 early on) is harmless.

### Finding (PLAUSIBLE, non-blocking)

**`pkg/fuse/prefetch.go:237,255` — a backward seek that lands within `window` bytes of the frontier creates a prefetch dead zone, because the frontier is monotonic and prefetch is pinned to it.**
`prefetchStart` is always `p.nextExpect`, and `nextExpect` never retreats. Once the window has grown large (up to 16MB), a genuine backward seek of <16MB is classified `withinWindow` → no reset, no re-aim. Subsequent sequential reads from the new (lower) position all have `end < nextExpect`, so they neither advance the frontier nor re-point prefetch — the consumer reads sequentially with zero prefetch benefit until it climbs back to the old frontier. The old strict code reset on any backward seek and re-aimed prefetch at the new position, so this is a bounded (<window) regression for the "seek back a bit, then read sequentially" pattern (e.g. re-reading a section, some parsers/media scrubbing). Forward sequential — the dominant case — is strongly improved, so this is a net win; worth a comment noting the tradeoff, or (if you want to close it) re-aiming `prefetchStart` toward `offset` when a within-band read lands well behind the frontier.

### Verdict
Clear net improvement, correct locking, good tests. The backward-seek dead zone is a minor bounded edge case, not a blocker.

LGTM.

srstack (approved):
LGTM. Window-band sequential detection correctly fixes the out-of-order readahead reset regression; OnRead holds p.mu throughout so concurrent completions are serialized, the frontier stays monotonic, and random-reset semantics are preserved. One non-blocking note: a backward seek within the (up-to-16MB) window creates a bounded prefetch dead zone since prefetch is pinned to the monotonic frontier — net positive given the strong forward-sequential win (5→14 MB/s).

qiffang:
**adversary-2 review — Strong LGTM, ship it.**

This PR fixes the exact regression the team identified 3 hours ago in a customer mount performance analysis: single-stream sequential read stuck at 3.5 MB/s at RTT 170-300ms. Author's measurement (5.1 → 14 MB/s at 200ms injected latency) directly aligns with the customer's environment.

## Root cause verified

Original `if offset == p.nextExpect` sequential detection is too strict. Kernel `FUSE_CAP_ASYNC_READ` (default enabled) issues multiple concurrent 1MB reads that complete in arbitrary order. Any out-of-order arrival → strict `==` failure → `else` branch RESET → cache + inflight cleared → next expected block synchronously re-fetched → single-stream throughput collapses to `1MB / RTT = 3-6 MB/s`.

Debug log evidence in the PR body precisely matches customer numbers.

## Fix logic — payload-value trace

New logic: `withinWindow = offset >= nextExpect - window && offset <= nextExpect + window`

| Case | Behavior | Correctness |
|---|---|---|
| `offset < nextExpect - window` (far-behind seek) | outside → RESET | ✅ genuine random |
| `offset > nextExpect + window` (far-forward seek) | outside → RESET | ✅ genuine random |
| within band + `end > nextExpect` (frontier advance) | grow window + prefetch | ✅ normal sequential |
| within band + `end <= nextExpect` (out-of-order / re-read) | no grow, no reset, still trigger prefetch if pending | ✅ **the fix** |

## Test coverage

4 white-box tests cover the critical scenarios:
- `TestOnReadStrictOrderPreservesWindow` — happy path baseline
- `TestOnReadOutOfOrderDoesNotReset` — regression guard directly against the root cause
- `TestOnReadGenuineBackwardJumpResets` — genuine random seek still resets (semantic preservation)
- `TestOnReadReReadSameRangeDoesNotInflateWindow` — re-read doesn't inflate window (side-effect guard)

## Non-blocker nits

1. `TestOnReadGenuineBackwardJumpResets` actually tests a **forward** jump 50MB beyond frontier (see the test body: `p.OnRead(far, chunk)` where `far = 50*1024*1024`). The comment inline says "a backward jump is symmetric" — meaning the author intentionally tested forward but named it "backward". Suggest renaming to `TestOnReadGenuineFarJumpResets` or adding a symmetric backward-jump test.

2. Boundary case not covered: `offset == nextExpect - window` exactly (the band boundary). Defensive addition, not required.

3. Window monotonic-growth concern: with strict `==`, out-of-order reads previously reset window to `prefetchMinWindow`. New logic doesn't reset within the band, so window only grows (up to `prefetchMaxWindow = 16MB`). For a slow reader with fast startup, window can reach 16MB early and stay there → aggressive over-fetch. Not correctness bug, memory pressure concern — potential follow-up idle-decay if telemetry surfaces it.

4. Backward seek within window treated as sequential (false positive, benign — triggers one wasted prefetch, no correctness impact).

## Cross-check against customer

Customer environment ([Slock #drive9 msg 8d5cca00](internal reference)): single-stream 3.1-3.5 MB/s at RTT 170-300ms. Customer explicitly asked for "顺序访问检测与预读流水线" without knowing Prefetcher already existed. This PR closes the loop: author identified + fixed the exact out-of-order reset bug that was defeating the existing Prefetcher. Customer's 3.5 MB/s should become ~9-14 MB/s in production once shipped.

## Scope

- Touches only `pkg/fuse/prefetch.go` `OnRead` logic + new `prefetch_test.go`
- Does not touch backend, storage, cache indexing, retry
- 32 LOC added / 14 deleted in production; 135 LOC in tests
- Well-bounded, easy to revert if telemetry surfaces regressions

## Thread safety verified

`OnRead` holds `p.mu` throughout via `defer p.mu.Unlock()`. Comment at top of file documents self-synchronized + lock ordering `FileHandle.mu → Prefetcher.mu`. Consistent with fix.

## Separately: stale-size hypothesis

@dev-2 flagged in Slock review that his earlier "prefetcher never attaches because stat cached stale small size at open time" hypothesis is a **different scenario** than what this PR fixes:

- **This PR fixes**: prefetcher attached, then reset by out-of-order → matches customer's existing-32MB-file test
- **Stale-size race**: prefetcher never attaches — only triggers when file is freshly created inside sandbox via `dd`, open→stat sees size=0, close, reopen, prefetcher init skipped because size < `read-cache-max-file-mb`. Separate follow-up PR.

This PR does not fix the stale-size scenario, and does not need to. Independent bug, independent path.

**Ship it.**

PR #710, opened 2026-07-08
Fix S3 delete prefix fallback for compatible stores

srstack:
## Summary
- fall back to per-object DeleteObject when batch DeleteObjects returns S3-compatible store compatibility errors
- preserve the batch DeleteObjects fast path for normal AWS S3 behavior
- keep tenant delete jobs progressing when OSS rejects DeleteObjects with MissingArgument

## Test
- go test ./pkg/s3client

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved S3 prefix deletion reliability when bulk deletion isn’t supported by the service.
  * Automatically falls back to deleting objects one at a time if the batch delete request fails for compatibility reasons.
  * Keeps track of how many objects were successfully removed during the fallback flow and skips invalid empty keys.

### DIFF
--- a/pkg/s3client/aws.go
+++ b/pkg/s3client/aws.go
@@ -689,7 +689,15 @@ func (c *AWSS3Client) deleteObjectsByPrefix(ctx context.Context, fullPrefix stri
 					},
 				})
 				if err != nil {
-					return fmt.Errorf("delete objects by prefix: %w", err)
+					if !isDeleteObjectsCompatibilityError(err) {
+						return fmt.Errorf("delete objects by prefix: %w", err)
+					}
+					deleted, fallbackErr := c.deleteObjectsOneByOne(ctx, objects)
+					out.DeletedObjects += deleted
+					if fallbackErr != nil {
+						return fmt.Errorf("delete objects by prefix fallback: %w", fallbackErr)
+					}
+					continue
 				}
 				if len(del.Errors) > 0 {
 					first := del.Errors[0]
@@ -703,6 +711,34 @@ func (c *AWSS3Client) deleteObjectsByPrefix(ctx context.Context, fullPrefix stri
 		}
 		continuation = res.NextContinuationToken
 	}
+}
+
+func isDeleteObjectsCompatibilityError(err error) bool {
+	if err == nil {
+		return false
+	}
+	msg := err.Error()
+	return strings.Contains(msg, "MissingArgument") ||
+		strings.Contains(msg, "MalformedXML") ||
+		strings.Contains(msg, "NotImplemented")
+}
+
+func (c *AWSS3Client) deleteObjectsOneByOne(ctx context.Context, objects []types.ObjectIdentifier) (int64, error) {
+	var deleted int64
+	for _, obj := range objects {
+		key := aws.ToString(obj.Key)
+		if key == "" {
+			continue
+		}
+		if _, err := c.client.DeleteObject(ctx, &s3.DeleteObjectInput{
+			Bucket: &c.bucket,
+			Key:    aws.String(key),
+		}); err != nil {
+			return deleted, fmt.Errorf("delete object %s: %w", key, err)
+		}
+		deleted++
+	}
+	return deleted, nil
 }
 
 func (c *AWSS3Client) UploadPartCopy(ctx context.Context, destKey, uploadID string, partNumber int, sourceKey string, startByte, endByte int64) (string, error) {


mornyx (approved):
## Review — PR #710

**Verdict: APPROVED**

### Summary

Falls back to per-object `DeleteObject` when batch `DeleteObjects` returns S3-compatible store compatibility errors (MissingArgument, MalformedXML, NotImplemented). Preserves the batch `DeleteObjects` fast path for normal AWS S3.

### Changes reviewed

**`pkg/s3client/aws.go` — 1 file, +37/-1**

1. `isDeleteObjectsCompatibilityError(err)` — string-matches `MissingArgument`, `MalformedXML`, `NotImplemented` in error message. Reasonable for S3-compatible provider errors where structured error codes aren't always available.

2. `deleteObjectsOneByOne(ctx, objects)` — iterates objects, calls `DeleteObject` per key, skips empty keys, returns partial count on first failure. Correct behavior.

3. Fallback in `deleteObjectsByPrefix` — when `DeleteObjects` returns a compatibility error, falls back to one-by-one deletion, accumulates `out.DeletedObjects`, and `continue`s to the next `ListObjectsV2` page. The `continue` skips the `IsTruncated` check, causing one extra list call that returns 0 objects — functionally correct, minor inefficiency in the fallback path only.

### Static verification

- `go build ./pkg/s3client/...` — passes
- `strings` import already present — no new imports needed
- Keys from `ListObjectsV2` are full keys — `deleteObjectsOneByOne` correctly uses them directly without `c.fullKey()`
- Partial delete counting correct — `out.DeletedObjects += deleted` before checking `fallbackErr`
- Codex (glm-5.2, xhigh): no findings — confirmed the `continue` behavior, error matching, key handling, and counting are all sound
- CI: CodeRabbit pass, ci pass (6m34s), e2e pass (7m23s)

### Non-blocking note

No test coverage for the new fallback path. The PR body mentions `go test ./pkg/s3client` but there are no test files referencing `isDeleteObjectsCompatibilityError` or `deleteObjectsOneByOne`. Consider adding a test with a mock S3 client that returns `MissingArgument` from `DeleteObjects` to verify the fallback path.

srstack:
## Review

When a non-AWS S3-compatible store (e.g. Aliyun OSS) rejects the batch `DeleteObjects` API with a compatibility error (`MissingArgument`/`MalformedXML`/`NotImplemented`), `deleteObjectsByPrefix` previously failed the whole tenant-delete job. This PR detects those errors and falls back to per-object `DeleteObject`, preserving the batch fast path for normal AWS behavior. The fix is correct and cleanly scoped.

Traced the control flow — no correctness issues:
- The fallback `continue` re-lists with the **unchanged** continuation token, which is the right call: the just-deleted page shrinks the listing, so the next `ListObjectsV2` returns subsequent objects. Advancing the token there could skip objects; re-listing can't.
- No infinite-loop risk: `objects` is pre-filtered for empty keys upstream, so `deleteObjectsOneByOne` always deletes ≥1 (the empty-key skip inside it is dead-but-harmless defensive code), guaranteeing forward progress.
- No double-counting: `out.DeletedObjects += deleted` counts only the page just deleted, and those objects don't reappear on re-list. Fallback aborts on first per-object error, matching the batch path's fail-fast semantics.

### Notes (both non-blocking)

**1. `pkg/s3client/aws.go:721` — substring matching on `err.Error()` is brittle vs typed error-code classification.**
`strings.Contains(msg, "NotImplemented")` etc. will false-positive on any error whose message merely contains one of these substrings (e.g. a wrapped detail, or an object key echoed into an unrelated error). The robust idiom is `var apiErr smithy.APIError; if errors.As(err, &apiErr) { switch apiErr.ErrorCode() { ... } }` — an exact code match. Impact is low because the fallback is safe (just slower per-object deletes, never data loss), so a false positive only costs latency. `smithy-go` is already an indirect dep; promoting it to direct would let you match codes exactly. Worth considering, not required.

**2. No test for the new fallback path.** The PR body lists `go test ./pkg/s3client`, but the diff touches only `aws.go` — `isDeleteObjectsCompatibilityError`, `deleteObjectsOneByOne`, and the re-list-on-`continue` behavior have no direct coverage (existing tests just still pass). A mock S3 client returning `MissingArgument` on `DeleteObjects` and succeeding on `DeleteObject` would lock in the fallback, the delete count, and pagination termination — the exact behavior that's hard to reproduce in prod and easy to regress.

### Verdict
Correct, safe fix for the OSS compatibility gap. The two notes are quality improvements (typed error classification + a fallback-path test), neither blocking. LGTM.

PR #716, opened 2026-07-09
Add tenant pool metadata resume wait metrics

srstack:
## Summary
- record admin tenant pool metadata resume wait duration across all groups
- record per-group metadata wait duration with deterministic result labels
- add pool-scoped metadata resume wait metrics with pool_id and TiDB Cloud organization_id labels
- add tenant pool capacity gauge with pool_id and organization_id labels
- suppress noisy schema-init business error events for expected transient TiDB DDL races / already-exists cases
- extend histogram buckets so minute-scale metadata resume waits, relaxed HTTP write latency alerts, and admin tenant pool update latency alerts are reachable

## Metrics
- drive9_service_operations_total / drive9_service_operation_duration_seconds with component="admin_tenant_pool", operation="metadata_resume_wait"
- drive9_service_operations_total / drive9_service_operation_duration_seconds with component="admin_tenant_pool", operation="metadata_resume_group_wait"
- drive9_tenant_pool_metadata_resume_wait_total{pool_id,organization_id,scope,result}
- drive9_tenant_pool_metadata_resume_wait_duration_seconds{pool_id,organization_id,scope,result}
- drive9_tenant_pool_capacity{pool_id,organization_id,state}

## Tests
- GOCACHE=/tmp/drive9-go-cache go test ./pkg/metrics -count=1
- GOCACHE=/tmp/drive9-go-cache go test ./pkg/server -run 'TestAdminTenantPoolMetadataResumeResultRank|TestTenantPoolMetadataResume|TestAdminTenantPool' -count=1

PR #717, opened 2026-07-09
Surface TiDB Cloud client errors

srstack:
## Summary
- return client-facing TiDB Cloud 400/401/403 errors instead of hiding them behind generic 502 responses
- apply the shared mapping across provision, admin tenant pool, quota, TiDB Cloud admin, and tenant delete cluster paths
- reject tidbcloud_spending_limit values between 1 and 9 RMB locally before calling TiDB Cloud

## Behavior
- TiDB Cloud status 400 -> HTTP 400 with upstream detail
- TiDB Cloud status 401 -> HTTP 401 with upstream detail
- TiDB Cloud status 403 / quota permission denied -> HTTP 403 with detail
- server/internal failures still keep their default generic status/message

## Tests
- go test ./pkg/server -run 'TestClientFacingErrorResponseMapsTiDBCloudClientErrors|TestQuotaSetRejectsInvalidQuotaValues|TestTenantPool|TestAdminTenantPool|TestProvision'
- go test ./pkg/tenant/tidbcloudnative

## Summary by CodeRabbit

* **New Features**
  * Improved error messages for tenant pool, tenant creation, deletion, and quota-related failures so clients receive clearer, safer responses.

* **Bug Fixes**
  * Updated TiDB Cloud error handling to return more accurate HTTP statuses in common failure cases.
  * Tightened spending limit validation: positive values must now be at least 10 RMB, or set to 0.
  * Prevented sensitive upstream error details from being exposed to users.

### DIFF
--- a/pkg/server/admin_tenant_pool.go
+++ b/pkg/server/admin_tenant_pool.go
@@ -187,7 +187,8 @@ func (s *Server) handleAdminTenantPoolCreate(w http.ResponseWriter, r *http.Requ
 				"error", err)...)
 			s.deleteTenantPoolMetadata(ctx, poolID, "create_free_tenants_error")
 			metricResult = "cluster_error"
-			errJSON(w, http.StatusBadGateway, fmt.Sprintf("create tenant pool failed: %v", err))
+			status, msg := clientFacingErrorResponse(http.StatusBadGateway, "create tenant pool failed", err)
+			errJSON(w, status, msg)
 			return nil
 		}
 		logger.Info(ctx, "server_event", eventFields(ctx, "admin_tenant_pool_create_free_tenants_done",
@@ -370,7 +371,8 @@ func (s *Server) handleAdminTenantPoolUpdate(w http.ResponseWriter, r *http.Requ
 					"grow_count", growCount,
 					"duration_ms", durationMillis(stageStarted),
 					"error", err)...)
-				return adminTenantPoolError(http.StatusBadGateway, fmt.Sprintf("grow tenant pool failed: %v", err))
+				status, msg := clientFacingErrorResponse(http.StatusBadGateway, "grow tenant pool failed", err)
+				return adminTenantPoolError(status, msg)
 			}
 			for _, res := range results {
 				s.startProvisionedTenantSchemaInit(ctx, res)
@@ -413,7 +415,8 @@ func (s *Server) handleAdminTenantPoolUpdate(w http.ResponseWriter, r *http.Requ
 					"deleted_count", deleted,
 					"duration_ms", durationMillis(stageStarted),
 					"error", err)...)
-				return adminTenantPoolError(http.StatusBadGateway, fmt.Sprintf("shrink tenant pool failed: %v", err))
+				status, msg := clientFacingErrorResponse(http.StatusBadGateway, "shrink tenant pool failed", err)
+				return adminTenantPoolError(status, msg)
 			}
 			logger.Info(ctx, "server_event", eventFields(ctx, "admin_tenant_pool_update_shrink_done",
 				"provider", tenant.ProviderTiDBCloudNative,
@@ -562,7 +565,8 @@ func (s *Server) handleAdminTenantPoolDelete(w http.ResponseWriter, r *http.Requ
 			"organization_id", pool.OrganizationID)...)
 		deleted, err := s.deleteNewestFreePoolTenants(ctx, pool.PoolID, pool.OrganizationID, 0, cred, true)
 		if err != nil {
-			return adminTenantPoolError(http.StatusBadGateway, fmt.Sprintf("delete tenant pool failed: %v", err))
+			status, msg := clientFacingErrorResponse(http.StatusBadGateway, "delete tenant pool failed", err)
+			return adminTenantPoolError(status, msg)
 		}
 		logger.Info(ctx, "server_event", eventFields(ctx, "admin_tenant_pool_delete_free_tenants_done",
 			"provider", tenant.ProviderTiDBCloudNative,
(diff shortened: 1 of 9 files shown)


srstack:
Reviewed. Verified locally: build, `go vet`, and gofmt clean; `TestClientFacingErrorResponseMapsTiDBCloudClientErrors` and `TestQuotaSetRejectsInvalidQuotaValues` pass.

**The approach is sound** — mapping upstream TiDB Cloud 400/401/403 to client-facing statuses instead of a blanket 502 is a real UX win, and centralizing it in `clientFacingErrorResponse` keeps the six call sites consistent. The spending-limit `1..9 RMB` rejection is nicely done at both layers (`validateQuotaSetRequest` and `validateTiDBCloudSpendingLimit`), so it's caught locally before the upstream call.

**Information-disclosure check — the important one, and it holds up.** `tidbCloudStatusMessage` extracts only the JSON `message` field (falling back to the raw body for non-JSON), and the test asserts `requestId`/`details` and generic upstream text are never surfaced. So structured upstream errors are safe. One residual: for a **non-JSON** upstream 400/401 body, the raw body is passed straight through to the client. Today those bodies are short human messages (`invalid TiDB Cloud API key`), so it's fine, but it's an implicit trust in upstream never putting anything sensitive in a plain-text 4xx body. Worth a comment noting that assumption.

**Fragility of regex parsing.** `tidbCloudStatusErrorPattern` depends on the exact `tidbcloud native <op> status <code>` error wording. If that phrasing drifts anywhere upstream, mapping silently falls back to 502 — a safe failure mode, but it means these mappings are only as stable as the string format. Given the codebase already leans on substring matching here, this is consistent, just flagging the coupling.

**One behavior change to confirm is intended:** in `provisionTenant`, the old code force-set the message to `invalid TiDB Cloud API key` for anything matching `status 401`/`unauthorized`/`invalid tidb cloud` while keeping **502**. Now a parseable `status 401` maps to a real **401**, but a 401-ish error that does *not* match the regex (e.g. only contains `unauthorized`) keeps 502 and loses that friendly message, since the friendly overrides only run when `status == 502` and the 401 override was dropped. Net improvement for the common case, minor regression for oddly-worded 401s. Fine if intentional.

**Merge note:** overlaps #716 in `admin_tenant_pool.go`, `server.go`, and `provisioner.go` — second to merge needs a rebase.

Non-blocking. LGTM.

srstack:
Re-reviewed `8521f263` (Address TiDB Cloud error review feedback). Verified locally: build/vet/gofmt clean; `TestClientFacingErrorResponseMapsTiDBCloudClientErrors` and `TestNewProvisionerFromEnv*` pass.

Both follow-ups look good:
- Added the **403 -> Forbidden** case with an explicit test, rounding out the 400/401/403 mapping coverage.
- Closed the validation gap in `parseDefaultSpendLimit` — the `DRIVE9_TIDBCLOUD_DEFAULT_SPENDING_LIMIT` env default now goes through `validateTiDBCloudSpendingLimit`, so the `1..9 RMB` floor is enforced at startup too (not just on the request path), with `TestNewProvisionerFromEnvRejectsTooSmallDefaultSpendingLimit` covering it.

No further concerns from me. LGTM.

mornyx (approved):
## Review — PR #717

**Verdict: APPROVED**

### Summary

Returns client-facing TiDB Cloud 400/401/403 errors instead of hiding them behind generic 502 responses. Applies a shared `clientFacingErrorResponse` mapping across provision, admin tenant pool, quota, TiDB Cloud admin, and tenant delete cluster paths. Also rejects `tidbcloud_spending_limit` values between 1 and 9 RMB locally before calling TiDB Cloud.

### Changes reviewed (9 files, +222/-35)

| Area | Files | Summary |
|------|-------|---------|
| Error mapping | `pkg/server/server.go` | `clientFacingErrorResponse` — parses TiDB Cloud status errors via regex, maps 400/401/403 to client-facing codes, extracts JSON message from upstream body |
| Provision | `pkg/server/server.go` | `provisionTenant` — existing string-based error messages preserved for 502 fallback, `clientFacingErrorResponse` applied first |
| Admin pool | `pkg/server/admin_tenant_pool.go` | `clientFacingErrorResponse` applied to create/update/delete/shrink/grow pool error paths |
| Admin tenants | `pkg/server/admin_tenants.go` | Applied to admin tenant create/delete pool claim + cluster delete paths |
| Quota | `pkg/server/quota.go` | `writeQuotaCredentialError` — `clientFacingErrorResponse` replaces manual switch, `ErrQuotaPermissionDenied` now surfaces as 403 via the shared mapping |
| Tenant delete | `pkg/server/tenant_delete.go` | Applied to delete tenant cluster path |
| Provisioner | `pkg/tenant/tidbcloudnative/provisioner.go` | Added structured logging for batch metadata wait (started/pending/finished/failed/timeout/canceled) with sorted cluster/tenant IDs |
| Validation | `pkg/server/quota.go` | `tidbcloud_spending_limit` 1-9 RMB rejected locally |
| Tests | `pkg/server/server_test.go`, `provision_test.go`, `quota_test.go` | `TestClientFacingErrorResponseMapsTiDBCloudClientErrors` + spending limit validation |

### Static verification

- `go build ./pkg/server/... ./pkg/tenant/... ./pkg/client/... ./pkg/fuse/... ./pkg/datastore/...` — passes
- `clientFacingErrorResponse` regex: `tidbcloud native ([A-Za-z0-9 _-]+) status ([0-9]{3})(?:: (.*))?` — constrained to 3-digit status codes, only maps to 400/401/403, everything else falls through to `defaultStatus` — safe, no injection risk
- `tidbCloudStatusMessage` — extracts `message` field from JSON body, falls back to raw body — reasonable
- Existing provision error messages (free cluster limit, payment required, capacity limit) preserved for 502 fallback — backward compatible
- `ErrQuotaPermissionDenied` now surfaces as 403 via `clientFacingErrorResponse` — correct, was previously handled by explicit check in `writeQuotaCredentialError`
- Spending limit validation: `*req.TiDBCloudSpendingLimit > 0 && *req.TiDBCloudSpendingLimit < 10` — rejects 1-9 RMB, allows 0 (unlimited) and >=10 — correct
- CI: CodeRabbit pass, ci pass (5m26s), e2e pass (7m48s)

### Notes

- Codex (glm-5.2, xhigh) was run but the session was interrupted without producing findings. The PR is primarily error-mapping refactoring with no data-path changes — static analysis is sufficient.
- The regex-based error parsing relies on the error message format from `provisioner.go:statusError()` (`tidbcloud native <op> status <code>: <body>`). This is a structured error pattern, not arbitrary user input, so the regex is safe.

Issue #16, opened 2026-03-26
Server handleStat bypasses backend.Stat and directly accesses Store — breaks layered abstraction


qiffang:
## Current State

\`\`\`go
// pkg/server/server.go:134-158
func (s *Server) handleStat(w http.ResponseWriter, r *http.Request, path string) {
    nf, err := s.backend.Store().Stat(path)  // ← bypasses Dat9Backend, hits Store directly
\`\`\`

All other handlers go through \`s.backend.Read()\`, \`s.backend.ReadDir()\`, etc. Only handleStat pierces through to the \`Store()\` layer.

## Problems

1. **Breaks encapsulation** — Server layer should not know about Store; it should only operate via the backend interface
2. **Bypasses path canonicalization** — backend.Stat calls normalizePath; calling Store.Stat directly skips this
3. **Blocks MountableFS migration** — if the backend is replaced with MountableFS routing, this code breaks

## Root Cause

handleStat needs the revision field, but \`backend.Stat()\` returns AGFS's \`*filesystem.FileInfo\` which does not include revision.

## Proposed Fix

Add a \`StatWithMeta()\` method to Dat9Backend that returns extended metadata:

\`\`\`go
type FileMetadata struct {
    filesystem.FileInfo
    Revision    int64
    ContentType string
    FileID      string
}

func (b *Dat9Backend) StatWithMeta(path string) (*FileMetadata, error) { ... }
\`\`\`

Server calls \`StatWithMeta\` instead of piercing through to Store.

PR #718, opened 2026-07-09
docs(fuse): define clean-state reexec gate

qiffang:
## Summary
- document current Drive9 FUSE mount ownership, serve loop, drain, and shutdown boundaries
- define V0 as a clean-state binary reexec prototype, not smooth upgrade / CSI rolling upgrade
- classify FUSE/process state into disposable, persisted-rebuildable, must-transfer, and must-block categories
- list mechanical refusal gates and required V0 tests before fd handoff implementation

## Validation
- git diff --no-index --check /dev/null docs/design/fuse-clean-state-reexec-audit.md

## Summary by CodeRabbit

* **Documentation**
  * Added a design document defining the initial clean-state reexec boundary for FUSE.
  * Clarified which runtime states must be transferred, blocked, or ignored before reexec.
  * Described the required handshake, failure handling, and rollback behavior for a safe process handoff.
  * Listed the key refusal and crash-recovery scenarios that should be covered by validation.

### DIFF
--- a/docs/design/fuse-clean-state-reexec-audit.md
+++ b/docs/design/fuse-clean-state-reexec-audit.md
@@ -222,6 +222,10 @@ reviewed as correct:
     the clean snapshot and before accept.
 15. Non-empty transient local overlay refuses reexec; an empty automatic
     transient overlay root alone does not force read-only-only V0.
+16. Accept timeout after fd transfer rolls back safely: new process receives fd
+    but does not send accept before timeout; old process removes the quiesce
+    barrier, resumes serving, and queued/blocked requests complete according to
+    the defined rollback behavior.
 
 Existing tests already cover parts of the drain behavior, but they are not
 enough for V0. In particular, `TestDrainAllowsCleanOpenHandles` is correct for
--- a/docs/design/fuse-clean-state-reexec-audit.md
+++ b/docs/design/fuse-clean-state-reexec-audit.md
@@ -1,0 +1,189 @@
+# Drive9 FUSE Clean-State Binary Reexec Audit
+
+**Status:** audit boundary for V0 implementation
+**Date:** 2026-07-09
+
+## Scope
+
+This document defines the first safe implementation boundary for Drive9 FUSE
+binary reexec. V0 is a **clean-state binary reexec prototype** only.
+
+V0 explicitly does not cover:
+
+- CSI Node Pod rolling upgrade.
+- Mount Pod recreate upgrade.
+- Dirty or active workloads.
+- JuiceFS-style smooth upgrade product semantics.
+- Vault and WebDAV mounts.
+
+Those require a later CSI/mount-pod lifecycle split plus fd relay through the
+node agent.
+
+## Current FUSE Lifecycle
+
+The CLI default path starts a background child process and turns the request
+into `drive9 mount --foreground`; readiness is the child writing mount process
+state through `pkg/mountstate` (`cmd/drive9/cli/mount.go`).
+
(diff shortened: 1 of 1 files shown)


qiffang (commented):
REVISE (would be REQUEST_CHANGES, but GitHub will not let this token request changes on its own PR).

Blocking issue: the V0 gate proves a clean snapshot only at one instant, but the document does not require a FUSE request quiesce barrier before taking that snapshot and before fd transfer.

The doc correctly says the current `Dat9FS.Drain` does not stop new FUSE requests from entering handlers, but the proposed runtime gate only says the reexec coordinator stopped accepting new reexec attempts and that in-flight FUSE request count reaches zero. That is not sufficient. After those checks pass, the old process is still serving the same FUSE fd until accept, so an application can open/write/create after the clean-state check and before new-process accept. The new process would then start from a stale clean-state assumption while process-local file handles, inode refs, dirty state, or pending work were created after the gate.

Required fix: the audit must define an explicit quiesce phase before the clean-state snapshot, e.g. old process stops dispatching new FUSE handler entry or queues new kernel requests without mutating Drive9 state, waits for already-dispatched handlers to drain to zero, takes the clean-state snapshot, and only then proceeds to `send_fd -> validate -> accept`. If new-process accept fails, rollback resumes the old dispatcher with the same fd. The test list must include a race case where an open/write/lookup arrives during prepare; it must either block until accept or cause deterministic abort/rollback, but it must not mutate state after the snapshot.

Secondary clarification needed before GREEN: the static scope check says "No local overlay" while current writable mounts also initialize a transient `LocalOverlay`. The doc should distinguish user/profile/git/local-root overlay state from automatic transient overlay, or state that V0 is read-only-only. As written, the scope is ambiguous and could either block all writable mounts or miss a process-local state owner.

I agree with the conservative matrix direction and the go-fuse fd blocker, but this doc is supposed to be the implementation gate. Without the quiesce barrier, the next implementation can still produce a mount that stays mounted while state diverges.

qiffang (commented):
REVISE (small, but still a gate for this document).

The previous blockers are addressed at head `8d80be4`: the doc now requires `prepare -> quiesce -> snapshot -> send_fd -> validate -> accept`, keeps the barrier held through snapshot/fd transfer, defines rollback by removing the barrier, and clarifies transient overlay state versus user/profile/git local overlay state.

Remaining blocker: the failure semantics now explicitly include the accept-timeout rollback path (`fd transfer succeeds but the new process does not acknowledge accept before timeout`), but the Required V0 Tests list does not require a test for that path. Since this doc is the implementation gate, every failure semantic that keeps the old process serving must have a mandatory regression test before fd handoff review.

Required fix: add a Required V0 Test for “fd transfer succeeds, new process never acknowledges accept before timeout; old process keeps serving, removes the quiesce barrier, and queued/blocked requests resume or fail according to the defined rollback behavior.”

After that, I have no remaining blocker on the audit doc.

qiffang (commented):
GREEN for the audit doc at head `70dc62b`.

The final blocker is addressed: Required V0 Test #16 now covers accept timeout after fd transfer, requiring the old process to remove the quiesce barrier, resume serving, and complete queued/blocked requests according to rollback semantics.

Review gate status:
- Quiesce barrier is explicit and held through snapshot/fd transfer.
- Rollback semantics keep old process authoritative until accept.
- Transient overlay scope is explicit: empty automatic root allowed, non-empty transient state refuses.
- Failure semantics now have matching mandatory tests, including accept-timeout rollback.

No remaining blocker from me on PR #718.

PR #720, opened 2026-07-09
Fix quota reservation lock conflict handling

srstack:
## Summary
- retry transient lock conflicts around meta upload reservation transactions
- fail open after retry exhaustion to preserve upload availability while recording busy_fail_open metrics
- add regression coverage for retry, business-error bypass, context cancellation, and backend fail-open semantics

## Tests
- go test ./pkg/meta ./pkg/backend ./pkg/server ./pkg/client

## Summary by CodeRabbit

* **New Features**
  * Added a distinct “quota reservation busy” error signal and corresponding fail-open handling during upload reservations.
  * Emit a dedicated metric for busy fail-open outcomes.

* **Bug Fixes**
  * Central upload quota reservation now retries on temporary metadata lock conflicts with backoff, and returns the busy signal after retries are exhausted.
  * Prevents reserved quota bytes from leaking when the reservation is handled as fail-open.

* **Tests**
  * Added/extended coverage for busy fail-open behavior, lock-conflict retry, retry exhaustion, context cancellation, and lock-conflict detection patterns.

### DIFF
--- a/pkg/backend/quota.go
+++ b/pkg/backend/quota.go
@@ -21,6 +21,7 @@ var (
 	ErrMediaLLMQuotaExceeded    = errors.New("tenant media LLM file quota exceeded")
 	ErrReservationNotFound      = errors.New("upload reservation not found")
 	ErrReservationAlreadyExists = errors.New("upload reservation already exists")
+	ErrQuotaReservationBusy     = errors.New("quota reservation busy")
 )
 
 type storageQuotaCheckResult struct {
--- a/pkg/backend/round_a_review_test.go
+++ b/pkg/backend/round_a_review_test.go
@@ -144,6 +144,20 @@ func TestRoundA_Fix1_ReserveUploadOnServer_NoLeakOnInsertFailure(t *testing.T) {
 	}
 }
 
+func TestReserveUploadOnServer_BusyFailOpensWithoutLeak(t *testing.T) {
+	b, fake := newCentralQuotaBackend(t)
+	fake.insertReservationErr = ErrQuotaReservationBusy
+
+	reserved, err := b.reserveUploadOnServer(context.Background(), "u1", "/a", 100, 0)
+	if reserved || err != nil {
+		t.Fatalf("busy fail-open path: reserved=%v err=%v, want (false, nil)", reserved, err)
+	}
+	u, _ := fake.GetQuotaUsage(context.Background(), "tenant-a")
+	if u.ReservedBytes != 0 {
+		t.Fatalf("reserved_bytes = %d after busy reserve, want 0", u.ReservedBytes)
+	}
+}
+
 func TestRoundA_Fix1_ReserveUploadOnServer_DuplicateIsIdempotent(t *testing.T) {
 	b, fake := newCentralQuotaBackend(t)
 
--- a/pkg/backend/upload_reservation.go
+++ b/pkg/backend/upload_reservation.go
@@ -29,6 +29,10 @@ import (
 //     server DB is unreachable (fail-open). Callers treat this as "no
 //     server-side reservation", so the matching abort/complete paths skip
 //     counter adjustments.
+//   - (false, nil): central quota stayed lock-contended after bounded retries.
+//     This also fail-opens to avoid a user-visible upload-initiate failure, but
+//     it records a distinct busy_fail_open metric so quota-admission bypass risk
+//     is observable.
 //
 // Idempotency: a duplicate (tenant_id, upload_id) is treated as
 // (true, nil) — the original initiate already claimed the reservation;
@@ -88,6 +92,14 @@ func (b *Dat9Backend) reserveUploadOnServer(ctx context.Context, uploadID, targe
 			zap.String("upload_id", uploadID))
 		metrics.RecordTenantOperation(b.tenantID, "central_quota", "reserve_upload", "duplicate", time.Since(start))
 		return true, nil
+	case errors.Is(err, ErrQuotaReservationBusy):
+		logger.Warn(ctx, "central_quota_reserve_upload_busy_fail_open",
+			zap.String("tenant_id", b.tenantID),
+			zap.String("upload_id", uploadID),
+			zap.Int64("size", totalSize),
+			zap.Error(err))
+		metrics.RecordTenantOperation(b.tenantID, "central_quota", "reserve_upload", "busy_fail_open", time.Since(start))
+		return false, nil
 	default:
 		logger.Warn(ctx, "central_quota_reserve_upload_failed",
 			zap.String("tenant_id", b.tenantID),
--- a/pkg/meta/meta.go
+++ b/pkg/meta/meta.go
@@ -26,6 +26,7 @@ var (
 	ErrStorageQuotaExceeded     = errors.New("tenant storage quota exceeded")
 	ErrFileCountQuotaExceeded   = errors.New("tenant file count quota exceeded")
 	ErrReservationAlreadyExists = errors.New("upload reservation already exists")
+	ErrQuotaReservationBusy     = errors.New("quota reservation busy")
 )
 
 type TenantStatus string
--- a/pkg/meta/quota.go
+++ b/pkg/meta/quota.go
@@ -10,6 +10,8 @@ import (
 	"sync/atomic"
 	"time"
 
+	mysql "github.com/go-sql-driver/mysql"
+
 	"github.com/mem9-ai/drive9/pkg/logger"
 	"github.com/mem9-ai/drive9/pkg/metrics"
 	"go.uber.org/zap"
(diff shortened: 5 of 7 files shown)


srstack:
Reviewed. Verified locally: build, `go vet`, and gofmt are clean; `pkg/meta` and `pkg/backend` retry/reserve/atomic tests pass.

**The core design is sound.** The busy path reuses the existing `(false, nil)` fail-open contract rather than inventing a new one, so it slots cleanly into the reserve-first saga:
- No quota leak on exhaustion — `atomicReserveAndInsertUploadOnce` runs the whole reserve+insert in a single `InTx`, so a lock-conflict rollback commits nothing; `TestReserveUploadOnServer_BusyFailOpensWithoutLeak` confirms `reserved_bytes == 0`.
- Callers already gate all counter adjustments on `reserved` (e.g. `upload.go` only calls `abortUploadReservation` when `reserved`), and `applyUploadCompleteTx` treats `settled=false` (case 1: fail-open initiate) by charging storage directly with no reserved→storage transfer. So the busy path neither double-counts nor leaks.
- No nested-retry amplification: `InTx` does not retry internally, and there's no other deadlock-retry layer in `pkg/meta`, so this is the single retry site.
- Retry logic is correct: business errors (`ErrStorageQuotaExceeded`, etc.) are not matched by `isMetaLockConflictError` and propagate immediately; `ctx.Done()` is honored during backoff; the per-retry `lock_conflict_retry` count is recorded. All four behaviors have regression tests.

**A few non-blocking notes:**

1. **Fail-open = quota-admission bypass under sustained contention.** This is a deliberate availability-over-strict-enforcement tradeoff, consistent with the existing DB-unreachable fail-open, and it's observable via the `busy_fail_open` metric — good. Suggest wiring an alert on `rate(...result="busy_fail_open")` so sustained bypass (which lets uploads through without a reservation) is caught, since under a hot-row deadlock storm this could fire steadily.

2. **Linear backoff has no jitter.** `attempt * 20ms` → 20/40/60ms, ~120ms max added initiate latency (fine). But without jitter, many upload-initiates deadlocking on the same hot row will retry in lock-step and re-collide. A small random jitter would spread them out and improve the retry success rate under real contention.

3. **String-matched lock detection** (`Deadlock found`/`1213`/`40001`/`Lock wait timeout`/`1205`) is fragile if the driver's error wording changes, but it's consistent with the existing error-classification style in the codebase, and matching both the code and the text is a reasonable belt-and-suspenders. Fine as-is.

4. **Minor edge:** if ctx is canceled *during* `fn()` on the final attempt and `fn` returns a lock-conflict error, the caller gets `ErrQuotaReservationBusy` rather than `context.Canceled`. Harmless (it fail-opens either way), just slightly less precise than the mid-backoff cancellation path. Not worth changing unless you want the ctx check hoisted above the attempt-exhaustion break.

5. **Scope:** retry is applied only to the reserve path, not to abort/complete transactions. Reasonable — reserve is the user-visible initiate path — but abort/complete lock conflicts will still surface as raw errors elsewhere; worth a follow-up if those show up in practice.

Overall LGTM. None of the above blocks merge; (1) and (2) are the ones I'd most encourage picking up.

srstack:
Re-reviewed `49fea27c` (remove redundant write conflict check). Verified: build/vet/gofmt clean, `pkg/meta` retry tests pass.

The redundant `(9007 && write conflict)` clause is gone; the effective rule is now simply case-insensitive `contains "write conflict"`, which is what the earlier subset made it anyway. `lower` is still used for that check, so no dead variable. The negative case (`9007 quota unavailable` → not retried) is still covered by `TestIsMetaLockConflictErrorIncludesTiDBWriteConflict`.

All my earlier points are addressed:
- `%w` double-wrap preserves the underlying lock error in the chain (tested).
- TiDB `9007`/`write conflict` detection added — this was the substantive fix; MySQL-only `1213`/`1205` matching would have been a near no-op against TiDB's optimistic-txn contention.
- Redundant clause removed.

The optional items I raised (backoff jitter, alerting on `busy_fail_open`) are not code-blocking — and I see the alerting half is being handled in runbooks#2652. LGTM.

qiffang (changes requested):
Verdict: REVISE — 1 blocker.

**B1 (BLOCKER): lock-conflict classification is too broad and can convert unrelated errors into quota fail-open.**

`pkg/meta/quota.go:576-581` classifies lock conflicts with raw string substring checks:

- `strings.Contains(msg, "40001")`
- `strings.Contains(lower, "write conflict")`

That is not safe enough for the upload admission path. After four matching attempts, `retryMetaLockConflict` wraps the error as `ErrQuotaReservationBusy`; `pkg/backend/upload_reservation.go:95-102` then treats that sentinel as `(reserved=false, err=nil)`, so the upload proceeds without a central quota reservation. A false positive here is not just an extra retry; it changes quota admission behavior to fail-open.

Please tighten this to structured DB error matching where possible. For this codebase that should mean using `errors.As` on the driver error type for MySQL/TiDB error numbers / SQLSTATE, then only falling back to exact, constrained message patterns. At minimum, do not match arbitrary occurrence of `40001` or `write conflict` in any wrapped error string. The TiDB write-conflict case should be tied to the TiDB/MySQL error shape, not any application/domain error that happens to contain those words.

Add negative regression tests proving these do **not** retry/fail-open:
- an unrelated error string containing `40001` as an ID/count/port/etc.
- an unrelated application error containing `write conflict` text.
- a wrapped non-DB error whose context contains those substrings.

**Non-blocking notes:**

- The busy fail-open path is an availability-over-admission tradeoff. I’m not keeping this as a blocker because generic central-quota fail-open already existed and `applyUploadCompleteTx` charges `storage_bytes` directly when no reservation row exists. Still, sustained lock contention can bypass initiate-time quota admission until completion-time charging; wire an alert/runbook threshold for `busy_fail_open`, and define when the system should fail closed or backpressure.
- The deterministic 20/40/60ms backoff has no jitter, so concurrent upload initiates can retry in lockstep under a hot-row contention storm. Jitter is recommended.
- Retry is only applied to the reserve path. That is acceptable for this PR because reserve is the user-visible admission path, but abort/complete can still surface raw lock conflicts; track separately if they appear in production.

Validation:
- `git diff --check origin/main...pr-720-review` passed locally.
- `go test ./pkg/meta ./pkg/backend ./pkg/server ./pkg/client` could not run locally because package TestMain starts testcontainers and this machine has no rootless Docker (`panic: rootless Docker not found`). GitHub `ci` and `drive9-server-local e2e smoke` are green on the PR.

qiffang (changes requested):
## Adversary Review — REVISE

Reviewed at `49fea27` (3 commits: quota reservation lock conflict retry + fail-open).

### Blocker

**B1 — `isMetaLockConflictError` substring matching is too broad and risks false-positive fail-open.**

`pkg/meta/quota.go` line ~578: the function matches raw substrings in `.Error()` output:
- `"40001"` matches any error containing that number (port, txn ID, row count), not just SQLSTATE 40001
- `"write conflict"` (case-insensitive) matches any error mentioning these words, even from non-DB subsystems

A false positive triggers `ErrQuotaReservationBusy`, which `reserveUploadOnServer` maps to `(false, nil)` — fail-open quota bypass. The matching must be tightened to avoid unintended quota bypass on non-lock errors. Prefer structured error code extraction from the DB driver, or at minimum anchor the patterns (e.g., `"SQLSTATE 40001"`, `"Error 1213 "` with trailing space).

### Non-blocking notes

**N1 — Fail-open admission bypass is bounded but should have alerting policy.**
`applyUploadCompleteTx` charges storage at completion even without a reservation, so bypass is temporary (admission-time only, not permanent quota evasion). However, under sustained contention, tenants can temporarily exceed quota. Define an acceptable bypass rate threshold and alert on `busy_fail_open` metric exceeding it.

**N2 — Linear backoff may cause retry collision storms.**
`time.Duration(attempt) * 20ms` gives 20/40/60ms. Under high concurrency, retrying transactions can collide again at the same delay. Consider exponential backoff with jitter.

**N3 — Retry scope is inconsistent.**
Only `AtomicReserveAndInsertUpload` retries on lock conflict. Other meta write operations (abort, complete, quota adjustments) face the same contention but don't retry. Document why only reserve needs retry, or apply consistently.

---
*Unified review from @adversary-1 and @adversary-2*

srstack:
Addressed B1 in d4b4604e.

What changed:
- `isMetaLockConflictError` now first uses `errors.As` to inspect `*mysql.MySQLError` and only treats MySQL/TiDB lock conflicts as retryable for `SQLState=40001`, `Number=1213`, `Number=1205`, or TiDB `Number=9007` with a `write conflict` message.
- The string fallback is now constrained to DB-shaped messages such as `Error 1213 ... Deadlock found`, `Error 1205 ... Lock wait timeout exceeded`, `SQLSTATE 40001`, and `Error/ERROR 9007 (...) ... write conflict`.
- Added negative regression coverage proving unrelated strings containing `40001`, unrelated app errors containing `write conflict`, and wrapped non-DB errors with those substrings do not retry and do not return `ErrQuotaReservationBusy`.

Validation:
- `GOCACHE=/tmp/drive9-go-cache go test ./pkg/meta -count=1`
- `GOCACHE=/tmp/drive9-go-cache go test ./pkg/backend ./pkg/server ./pkg/client -count=1`
- `git diff --check`

The optional jitter/follow-up scope notes remain intentionally out of this blocker fix.

qiffang (approved):
GREEN from my side at head d4b4604.

The prior blocker is resolved: lock-conflict detection no longer treats bare `40001` or `write conflict` substrings as retryable. The typed MySQL path checks SQLState / errno, the TiDB 9007 path requires write-conflict semantics, and fallback matching is constrained to DB-shaped errors. The new negative tests cover unrelated 40001/write-conflict text and verify those paths do not return the busy sentinel that the backend fail-opens.

Validation: `git diff --check origin/main...HEAD`, gofmt check on changed Go files, `go test -c ./pkg/meta ./pkg/backend ./pkg/tenant`, and `go vet ./pkg/meta ./pkg/backend ./pkg/tenant` passed locally. Runtime `go test` for `pkg/meta`/`pkg/backend` is still blocked on this machine by missing rootless Docker/testcontainers, but GitHub `ci`, `drive9-server-local e2e smoke`, and CodeRabbit are green on the latest head.

PR #723, opened 2026-07-10
Cache TiDB Cloud RBAC and persist spending limit

srstack:
## Summary
- persist TiDB Cloud spending limit in tenant quota config and return quota/admin responses from local DB
- add positive TiDB Cloud API-key-to-cluster RBAC cache so quota/admin reads avoid repeated OpenAPI calls
- backfill or reconcile local spending limit from TiDB Cloud when authorization refreshes, including create-time quota and pool claim flows

## Tests
- go test ./pkg/meta ./pkg/server -run 'TestQuota|TestAdminTenant|TestGetQuotaConfig|TestMetaSchema|TestDiffMetaTableMetaReportsMissingPrimaryKeyConstraint' -count=1
- go test ./pkg/meta ./pkg/server -count=1
- go test ./pkg/client -count=1

## Summary by CodeRabbit

* **New Features**
  * Added TiDB Cloud spending-limit fields to tenant quota settings (including explicit `0`) with local persistence and checked timestamps.
  * Improved sync/backfill and admin quota handling so stored values are preferred, with targeted refresh when needed.
  * Added TiDB Cloud RBAC caching and new quota observability metrics (including no `tenant_id` label).
* **Bug Fixes**
  * Updated upload quota enforcement to apply storage/file limits only when quota overrides are enabled.
* **Documentation**
  * Revised the quota guide for updated semantics and expanded operator alerting guidance.
* **Tests**
  * Expanded quota and metrics coverage for spending-only and caching/backfill scenarios.

### DIFF
--- a/docs/guides/quota.md
+++ b/docs/guides/quota.md
@@ -205,3 +205,13 @@ The quota API returns JSON errors through the standard server error shape.
   gaps require operational reconciliation from tenant file metadata.
 - Multipart uploads use a stricter reserve-first path before writing tenant
   upload state.
+- TiDB Cloud quota reads expose low-cardinality observability counters:
+  `drive9_tidbcloud_rbac_cache_requests_total`,
+  `drive9_tidbcloud_openapi_requests_total`,
+  `drive9_tidbcloud_spending_limit_sync_total`, and
+  `drive9_tidbcloud_spending_limit_missing_total`. These metrics intentionally
+  do not include `tenant_id`.
+- Do not alert on cache misses, spending-limit backfills, or missing local
+  spending-limit observations being greater than zero. Those are normal during
+  first access, key changes, or migration. Alert only on sustained TiDB Cloud
+  OpenAPI error rate or an unexpected OpenAPI call-rate baseline increase.
--- a/docs/guides/quota.md
+++ b/docs/guides/quota.md
@@ -1,6 +1,6 @@
 # drive9 quota guide
 
-Last verified: 2026-06-24.
+Last verified: 2026-07-10.
 
 This guide shows how to read and update Drive9 tenant quota from the CLI and
 HTTP API in TiDBCloud Mode.
@@ -14,7 +14,7 @@ Drive9 exposes these user-settable quota settings:
 | `max_storage_size` | Maximum confirmed plus reserved file storage size, in Mi. Stored in Drive9. |
 | `max_file_size` | Maximum single file size, in Mi. Stored in Drive9. Must not exceed the server `DRIVE9_MAX_UPLOAD_BYTES` limit. |
 | `max_file_count` | Maximum file count. Stored in Drive9. `0` means unlimited. |
-| `tidbcloud_spending_limit` | TiDB Cloud Cluster Spending Limit. The value is passed through to TiDB Cloud, read from and written to the TiDB Cloud cluster, and not stored in Drive9. See the [TiDB Cloud Spending Limit guide](https://docs.pingcap.com/tidbcloud/manage-serverless-spend-limit). |
+| `tidbcloud_spending_limit` | TiDB Cloud Cluster Spending Limit. Drive9 stores the latest known value locally after create/update or a TiDB Cloud authorization refresh. `0` is a valid explicit value; missing local data is treated as unknown until it is backfilled from TiDB Cloud. See the [TiDB Cloud Spending Limit guide](https://docs.pingcap.com/tidbcloud/manage-serverless-spend-limit). |
 
 Quota responses include these storage usage counters:
 
@@ -36,14 +36,18 @@ the same keys as the server defaults when passed explicitly.
 Only `tidb_cloud_native` tenants support quota update through this API. Other
 tenant providers use their configured defaults.
 
-Tenant list/get first lists TiDB Cloud managed clusters with the supplied TiDB
-Cloud API keys. Drive9 reads local tenant and quota config only after that read
-succeeds.
+Tenant list/get validates that the supplied TiDB Cloud API key can access the
+tenant's cluster. Drive9 caches successful API-key-to-cluster authorization in
+process to avoid calling the TiDB Cloud OpenAPI on every read. The quota
+response itself is assembled from Drive9's local tenant, quota config, and usage
+tables after authorization succeeds.
 
 Quota update first reads the TiDB Cloud cluster labels, then patches the Drive9
 quota update labels to confirm the API key has cluster write permission. If that
 label patch succeeds, Drive9 patches `tidbcloud_spending_limit` when present and
-then writes any Drive9-stored quota fields.
+then writes any Drive9-stored quota fields. If a TiDB Cloud read returns a
+spending limit that is missing or different locally, Drive9 updates the local
+copy before returning the quota response.
 
 ## CLI
 
--- a/docs/guides/quota.md
+++ b/docs/guides/quota.md
@@ -38,9 +38,11 @@ tenant providers use their configured defaults.
 
 Tenant list/get validates that the supplied TiDB Cloud API key can access the
 tenant's cluster. Drive9 caches successful API-key-to-cluster authorization in
-process to avoid calling the TiDB Cloud OpenAPI on every read. The quota
-response itself is assembled from Drive9's local tenant, quota config, and usage
-tables after authorization succeeds.
+process for up to one hour to avoid calling the TiDB Cloud OpenAPI on every
+read. Empty managed-cluster list results are not retained, and successful tenant
+create/delete paths invalidate the full-list cache for the submitted
+credentials. The quota response itself is assembled from Drive9's local tenant,
+quota config, and usage tables after authorization succeeds.
 
 Quota update first reads the TiDB Cloud cluster labels, then patches the Drive9
 quota update labels to confirm the API key has cluster write permission. If that
--- a/pkg/meta/meta.go
+++ b/pkg/meta/meta.go
@@ -631,6 +631,7 @@ func metaInitSchemaStatements() []string {
 			max_file_count        BIGINT NOT NULL DEFAULT 0,
 			max_media_llm_files   BIGINT NOT NULL DEFAULT 500,
 			max_monthly_cost_mc   BIGINT NOT NULL DEFAULT 0,
+			tidbcloud_spending_limit BIGINT NULL,
 			created_at            DATETIME(3) NOT NULL DEFAULT CURRENT_TIMESTAMP(3),
 			updated_at            DATETIME(3) NOT NULL DEFAULT CURRENT_TIMESTAMP(3) ON UPDATE CURRENT_TIMESTAMP(3)
 		)`,
--- a/pkg/meta/meta.go
+++ b/pkg/meta/meta.go
@@ -631,7 +631,9 @@ func metaInitSchemaStatements() []string {
 			max_file_count        BIGINT NOT NULL DEFAULT 0,
 			max_media_llm_files   BIGINT NOT NULL DEFAULT 500,
 			max_monthly_cost_mc   BIGINT NOT NULL DEFAULT 0,
+			quota_limits_overridden TINYINT(1) NOT NULL DEFAULT 1,
 			tidbcloud_spending_limit BIGINT NULL,
+			tidbcloud_spending_limit_checked_at DATETIME(3) NULL,
 			created_at            DATETIME(3) NOT NULL DEFAULT CURRENT_TIMESTAMP(3),
 			updated_at            DATETIME(3) NOT NULL DEFAULT CURRENT_TIMESTAMP(3) ON UPDATE CURRENT_TIMESTAMP(3)
 		)`,
--- a/pkg/meta/meta_test.go
+++ b/pkg/meta/meta_test.go
@@ -964,6 +964,13 @@ func TestMetaSchemaSpecTracksPrimaryKeyConstraint(t *testing.T) {
 	if !pk.isPrimary {
 		t.Fatal("expected primary constraint marker")
 	}
+	spendingLimit, ok := table.columns["tidbcloud_spending_limit"]
+	if !ok {
+		t.Fatal("tenant_quota_config schema missing tidbcloud_spending_limit")
+	}
+	if spendingLimit.addSQL != "ALTER TABLE tenant_quota_config ADD COLUMN tidbcloud_spending_limit BIGINT NULL" {
+		t.Fatalf("tidbcloud_spending_limit addSQL = %q", spendingLimit.addSQL)
+	}
 }
 
 func TestMetaSchemaSpecIncludesTenantS3EncryptionColumns(t *testing.T) {
@@ -1333,14 +1340,15 @@ func TestDiffMetaTableMetaReportsMissingPrimaryKeyConstraint(t *testing.T) {
 	meta := metaTableMeta{
 		tableName: "tenant_quota_config",
 		columns: map[string]metaColumnMeta{
-			"tenant_id":           {columnType: "varchar(64)"},
-			"max_storage_bytes":   {columnType: "bigint"},
-			"max_file_size_bytes": {columnType: "bigint"},
-			"max_file_count":      {columnType: "bigint"},
-			"max_media_llm_files": {columnType: "bigint"},
-			"max_monthly_cost_mc": {columnType: "bigint"},
-			"created_at":          {columnType: "datetime(3)"},
-			"updated_at":          {columnType: "datetime(3)"},
+			"tenant_id":                {columnType: "varchar(64)"},
+			"max_storage_bytes":        {columnType: "bigint"},
+			"max_file_size_bytes":      {columnType: "bigint"},
+			"max_file_count":           {columnType: "bigint"},
+			"max_media_llm_files":      {columnType: "bigint"},
+			"max_monthly_cost_mc":      {columnType: "bigint"},
+			"tidbcloud_spending_limit": {columnType: "bigint"},
+			"created_at":               {columnType: "datetime(3)"},
+			"updated_at":               {columnType: "datetime(3)"},
 		},
 	}
 	createStmt := `CREATE TABLE tenant_quota_config (
@@ -1350,6 +1358,7 @@ func TestDiffMetaTableMetaReportsMissingPrimaryKeyConstraint(t *testing.T) {
 		max_file_count BIGINT NOT NULL,
 		max_media_llm_files BIGINT NOT NULL,
 		max_monthly_cost_mc BIGINT NOT NULL,
+		tidbcloud_spending_limit BIGINT NULL,
 		created_at DATETIME(3) NOT NULL,
 		updated_at DATETIME(3) NOT NULL
 	)`
--- a/pkg/meta/meta_test.go
+++ b/pkg/meta/meta_test.go
@@ -971,6 +971,20 @@ func TestMetaSchemaSpecTracksPrimaryKeyConstraint(t *testing.T) {
 	if spendingLimit.addSQL != "ALTER TABLE tenant_quota_config ADD COLUMN tidbcloud_spending_limit BIGINT NULL" {
 		t.Fatalf("tidbcloud_spending_limit addSQL = %q", spendingLimit.addSQL)
 	}
+	quotaOverride, ok := table.columns["quota_limits_overridden"]
+	if !ok {
+		t.Fatal("tenant_quota_config schema missing quota_limits_overridden")
+	}
+	if quotaOverride.addSQL != "ALTER TABLE tenant_quota_config ADD COLUMN quota_limits_overridden TINYINT(1) NOT NULL DEFAULT 1" {
+		t.Fatalf("quota_limits_overridden addSQL = %q", quotaOverride.addSQL)
+	}
+	checkedAt, ok := table.columns["tidbcloud_spending_limit_checked_at"]
+	if !ok {
+		t.Fatal("tenant_quota_config schema missing tidbcloud_spending_limit_checked_at")
+	}
+	if checkedAt.addSQL != "ALTER TABLE tenant_quota_config ADD COLUMN tidbcloud_spending_limit_checked_at DATETIME(3) NULL" {
+		t.Fatalf("tidbcloud_spending_limit_checked_at addSQL = %q", checkedAt.addSQL)
+	}
 }
 
 func TestMetaSchemaSpecIncludesTenantS3EncryptionColumns(t *testing.T) {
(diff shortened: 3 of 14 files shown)


srstack:
Reviewed head `40442ccf`. Verified locally: build, `go vet`, gofmt clean; `pkg/server` + `pkg/meta` + `pkg/metrics` quota/admin/RBAC tests pass.

**The design is sound and the caching is scoped correctly.** I traced the security-relevant question — "does a positive RBAC cache let a revoked API key act for up to 6h?" — and the answer is essentially no for mutating paths, which is the important part:

- **quota SET / admin quota-set** → `applyQuotaSet` calls the TiDB Cloud API directly (not cached), so the real authorization is still enforced live on every write.
- **admin delete** → after the cached `authorizedAdminTenant` pre-check, `MarkQuotaUpdateStarted(...cred)` makes a live TiDB Cloud call before anything destructive happens, so a revoked key fails there.
- The RBAC cache only short-circuits the *cluster-list / cluster-existence pre-check* and lets **quota GET** return the locally-persisted spending limit without an OpenAPI round-trip.

So the cache is a cost/latency optimization on reads and pre-checks, with live authorization still gating every mutation. Good separation.

**Cache implementation is clean:** credential key is a sha256 of pub+priv (no plaintext secret stored in the map), TTL expiry with lazy eviction, dedup on remember, defensive nil-receiver guards, and `getClusterList` returns a copy so callers can't mutate cached state. Metrics (hit/miss/bypass) are wired for observability.

**One thing worth calling out explicitly (not blocking): a bounded staleness / info-exposure window on reads.**
Because quota GET and the admin-get pre-check can be served from cache for up to `tidbCloudRBACCacheTTL` (6h), an API key whose TiDB Cloud access to a cluster was revoked can still:
- pass the admin-tenant-get pre-check, and
- read the locally-persisted `tidbcloud_spending_limit` for that tenant

for up to 6h after revocation. No cluster data, no writes, no deletes — just the persisted spending-limit value and existence confirmation. That's likely an acceptable tradeoff for the OpenAPI-call savings, but:
1. 6h is a long TTL for an authorization-adjacent cache. Consider making it configurable / shorter, or documenting the revocation-propagation window in `docs/guides/quota.md` alongside the other quota behavior notes this PR adds.
2. A short comment on `tidbCloudRBACAllowed` / `listAllManagedClustersCached` noting "reads/pre-check only; every mutation re-authorizes live against TiDB Cloud" would make the security boundary obvious to future readers and prevent someone from later reusing this cache to gate a write path.

**Minor:**
- `getCluster` computes `entryKey(cred, clusterID)` twice under the lock (once for lookup, once for the stale-delete). Trivial; could hoist to a local.
- The `allowCache = cfg.TiDBCloudSpendingLimit != nil` gate in `authorizedAdminTenant` is a nice touch — it forces a fresh OpenAPI authorize when there's no persisted value to fall back on. Worth a one-line comment on *why* (correctness: don't serve a pre-check from cache when we have nothing persisted to return anyway).

Overall LGTM. No code changes required; I'd encourage the TTL/doc note in point 1 since it's the one place the behavior could surprise an operator reasoning about key revocation.

srstack:
Re-reviewed `f85fb66a` (Address TiDB Cloud quota cache review feedback). Verified: build, `go vet`, gofmt clean; full `pkg/meta` + `pkg/server` quota/admin/schema suites pass.

**My earlier points are addressed, and the commit goes further in a good way:**

1. **RBAC cache TTL 6h → 1h**, plus `docs/guides/quota.md` now documents the one-hour authorization-cache window, that empty managed-cluster lists are not retained, and that create/delete invalidate the full-list cache (`forgetClusterList`). That closes both the revocation-window concern and the empty-result caching edge I raised. 

2. **New `quota_limits_overridden` column — this is a real correctness fix I under-weighted in the first pass.** Persisting a spending-limit-only row previously would have inserted default storage/file quota values that then \*looked\* like an explicit override. The flag cleanly separates "spending-limit-only" rows (flag false → storage/file limits fall back to process defaults) from genuine quota overrides. The reserve SQL (`CASE WHEN quota_limits_overridden THEN ... ELSE NULL`) and `uploadReservationQuotaErrorTx` both honor it consistently, so a spending-only row no longer accidentally pins storage quota to insert-time defaults. `CopyQuotaConfig` also preserves the flag correctly (copies spending-only as a patch, not a full override).

**Two backward-compat things I checked and they hold:**
- The migration derives `ADD COLUMN ... DEFAULT 1` from the CREATE TABLE spec, so existing rows become `quota_limits_overridden=1`. That's correct: every pre-PR `tenant_quota_config` row was written via `SetQuotaConfig`/patch with real quota values (there was no spending-only concept yet), so treating them as overridden preserves their existing enforcement behavior.
- `GetQuotaConfigVersion` switched from "all-values-equal-defaults ⇒ empty version" to `!quotaLimitsOverridden`. A pre-existing row whose values happened to equal defaults would flip from empty to a non-empty version after migration. I traced the only consumer (`quotaConfigCache.refresh`) — the version is purely a change-detection token, so this triggers at most one extra cache refresh per tenant, no correctness impact.

**Minor / optional:**
- `SetQuotaConfigPatch` passes `patch.quotaLimitOverrideSet()` twice (once as `insertOverride`, once inline in the UPDATE args). Harmless, just a small readability wart — could reuse the local.
- The `IF(?, 1, quota_limits_overridden)` on update means the flag is "sticky" once true — a later spending-only patch won't clear it. That's the intended semantics (a tenant that ever had real quota overrides keeps them), but worth a one-line comment since it's a deliberate asymmetry vs the COALESCE'd columns.

Nice iteration — the `quota_limits_overridden` split in particular makes the persisted-quota model materially more correct. LGTM.

qiffang:
## Review

Current head reviewed: `5bce55e`.

The main design is sound: persisting `tidbcloud_spending_limit` locally, serving quota/admin reads from Drive9 metadata after authorization, and limiting the TiDB Cloud RBAC cache to read/pre-check paths is the right shape. The earlier cache concerns are addressed in current head: TTL is down to 1h, docs call out the cache window, empty list results are not retained, create/delete invalidate the list cache, pool-create sync failure logs/continues, and `quota_limits_overridden` cleanly separates spending-limit-only rows from real storage/file quota overrides.

Validation I ran:

- `go test ./pkg/meta ./pkg/server -run 'TestQuota|TestAdminTenant|TestGetQuotaConfig|TestMetaSchema|TestDiffMetaTableMetaReportsMissingPrimaryKeyConstraint|TestTiDBCloudRBACCache' -count=1`
- `go test ./pkg/meta ./pkg/server -count=1`
- `go test ./pkg/metrics -count=1`
- `go test ./pkg/client -count=1`
- `go vet ./pkg/meta ./pkg/server ./pkg/metrics ./pkg/client`
- `gofmt -l` on the touched Go files

One small correctness issue remains before merge:

### Blocking

`pkg/meta/quota.go` added `quota_limits_overridden`, but `SetQuotaStorageBytes` does not set that flag to true when it updates an existing spending-limit-only row.

Repro on current head:

```go
limit := int64(100)
_ = s.SetQuotaConfigPatch(ctx, "tenant", meta.QuotaConfigPatch{TiDBCloudSpendingLimit: &limit})
_ = s.SetQuotaStorageBytes(ctx, "tenant", 12345)
cfg, _ := s.GetQuotaConfig(ctx, "tenant")
// got cfg.MaxStorageBytes == DefaultMaxStorageBytes(), cfg.QuotaLimitsOverridden == false
// want cfg.MaxStorageBytes == 12345, cfg.QuotaLimitsOverridden == true
```

The call returns nil and writes `max_storage_bytes`, but because the row remains `quota_limits_overridden=false`, `GetQuotaConfig` intentionally ignores the stored storage/file quota fields and returns process defaults. New rows are fine because the DB default for the flag is 1; the broken case is specifically “spending-limit-only row first, then legacy storage-only update.”

Fix should be small: make `SetQuotaStorageBytes` set `quota_limits_overridden=1` on insert/update, and add a regression for spending-limit-only row -> `SetQuotaStorageBytes` -> effective storage quota changes.

Everything else I checked looks mergeable after that.

srstack:
Addressed the remaining review item from current head `5bce55e`.

Issue: `SetQuotaStorageBytes` updated `max_storage_bytes` on an existing spending-limit-only row, but left `quota_limits_overridden=false`. That made `GetQuotaConfig` intentionally ignore the stored storage/file quota fields and keep returning process defaults.

Fix in `d0b8f936`:

- Updated `SetQuotaStorageBytes` to explicitly set `quota_limits_overridden = 1` on both insert and duplicate-key update:

```sql
INSERT INTO tenant_quota_config (tenant_id, max_storage_bytes, quota_limits_overridden)
VALUES (?, ?, 1)
ON DUPLICATE KEY UPDATE
  max_storage_bytes = VALUES(max_storage_bytes),
  quota_limits_overridden = 1
```

- Added regression coverage: `TestSetQuotaStorageBytesOverridesSpendingOnlyRow` creates a spending-limit-only row first, then calls `SetQuotaStorageBytes`, and verifies:
  - effective `MaxStorageBytes` becomes the explicit storage value
  - `QuotaLimitsOverridden` becomes true
  - existing `TiDBCloudSpendingLimit` is preserved

Validation run:

```bash
gofmt -w pkg/meta/quota.go pkg/meta/quota_setting_test.go
go test ./pkg/meta -run 'TestSetQuotaStorageBytes|TestQuotaConfigStoresTiDBCloudSpendingLimitWithoutStorageVersion|TestGetQuotaConfigVersion' -count=1
git diff --check
```

All passed.

qiffang:
## Consolidated Review: PR #723 — Cache TiDB Cloud RBAC and persist spending limit

**Reviewer**: @dat9-dev1

### Summary

This PR does three things well:
1. **Persists TiDB Cloud spending limit locally** — eliminates redundant OpenAPI calls on every quota read
2. **Adds in-process RBAC cache** — avoids repeated list-cluster auth calls (1h TTL, bounded size)
3. **Spending limit backfill** — reconciles local state from cloud on auth refresh

The code is well-structured with good separation of concerns. Test coverage is solid for the cache and quota paths. The `quota_limits_overridden` flag is a clean solution for distinguishing spending-limit-only rows from full quota overrides.

Below are issues grouped by severity.

---

### Blockers (B)

**B1: `enforceSizeLocked` eviction is non-deterministic and silently drops valid entries**
`pkg/server/tidbcloud_rbac_cache.go:180-194`

When the cache exceeds `tidbCloudRBACCacheMaxEntries` (10000), `enforceSizeLocked` deletes map entries using Go's non-deterministic iteration order. This means a brand-new entry could be evicted immediately after insertion, while stale (but not yet expired) entries survive. Under sustained load this creates unpredictable cache behavior — some tenants will see cache misses despite being actively cached seconds ago.

**Suggestion**: Either use an LRU (e.g. embed `container/list`) or at minimum evict the oldest entry by `expiresAt`. If LRU is too complex for v1, consider tracking insertion time and evicting oldest-inserted. Alternatively, since `pruneLocked` already runs on every write, the size limit may rarely be hit in practice — if so, document that `enforceSizeLocked` is a safety valve, not a policy, and log when it triggers so you can evaluate.

**B2: `resolveQuotaAPIKey` does hash-then-decrypt-then-compare — redundant and expensive**
`pkg/server/quota.go:135-175`

The flow is:
1. Hash the bearer token → lookup by hash
2. Check `subtle.ConstantTimeCompare(hash, resolved.APIKey.JWTHash)` — this is always true if step 1 found a match, so it's a no-op
3. Decrypt the ciphertext → compare plaintext with bearer token
4. Parse and verify JWT claims

Step 2 is redundant (the DB lookup already matched by hash). Step 3 (decrypt + compare) is defensive but expensive — if the hash function is collision-resistant (SHA-256), the decrypt round-trip adds latency without security benefit. Consider whether the hash-based lookup alone is sufficient, or document why the decrypt-verify step is needed (e.g. hash truncation, legacy migration).

**B3: `syncTiDBCloudSpendingLimit` double-reads `GetQuotaConfig` on update path**
`pkg/server/quota.go:625-660`

When `result != "unchanged"`, `syncTiDBCloudSpendingLimit` calls `tidbCloudSpendingLimitSyncResult` (which reads `GetQuotaConfig`) and then if `!observedAt.IsZero()`, reads `GetQuotaConfig` again to check staleness. That's 2 DB reads for the same row in the same request. The staleness check could reuse the config already fetched by `tidbCloudSpendingLimitSyncResult`.

**Suggestion**: Have `tidbCloudSpendingLimitSyncResult` return the `cfg` it already loaded, and pass it to the staleness check.

---

### Non-blocking Issues (N)

**N1: `pruneLocked` runs on every `rememberCluster` / `rememberClusterList` — O(n) scan on write**
`pkg/server/tidbcloud_rbac_cache.go:159`

With max 10k entries + 1k lists, this is a full-map scan on every cache write. Under high write rate this could be a latency concern. Consider lazy pruning (probabilistic or periodic) or accepting that expired entries are already handled by the read path and only pruning on size enforcement.

**N2: `tidbCloudStatusErrorPattern` parses error strings with regex — fragile coupling**
`pkg/server/server.go:4465`

`parseTiDBCloudStatusError` extracts operation/status/body from error message strings like `"tidbcloud native <op> status <code>: <body>"`. This regex-on-error-string pattern is fragile — if the upstream error format changes, the parser silently stops matching. Consider wrapping TiDB Cloud errors in a typed error (e.g. `type TiDBCloudStatusError struct { Operation string; Code int; Body string }`) that can be matched with `errors.As`.

**N3: `adminTenantRowsNeedSpendingLimitBackfill` triggers a full fresh list-clusters call**
`pkg/server/admin_tenants.go:275-285`

When `clustersFromCache && needBackfill`, the code calls `listAllManagedClustersFresh` which always bypasses cache and hits TiDB Cloud API. For the admin tenant list endpoint with many tenants, this means one cached call + one fresh call in sequence. If the cache hit rate is high, this effectively halves the cache benefit for the list endpoint. Consider whether backfill can be done lazily (on individual tenant GET) rather than synchronously on list.

**N4: Spending limit validation minimum of 10 RMB is undocumented**
`pkg/server/quota.go:338-340`

The new validation `tidbcloud_spending_limit must be 0 or at least 10 RMB` is a good guardrail, but the "10 RMB" threshold isn't documented in `docs/guides/quota.md`. The docs mention `0 = valid` and `missing = unknown` but don't mention the minimum. Add this to the docs.

**N5: Test for RBAC cache is minimal — missing TTL expiration and size eviction tests**
`pkg/server/tidbcloud_rbac_cache_test.go`

The single test covers empty-list, remember, forget, and cluster retention after list invalidation — good baseline. But there's no test for:
- TTL expiration (create entry, advance time, assert miss)
- Size eviction behavior (fill past max, assert something was evicted)
- Concurrent access (even a simple goroutine test)

Consider adding at least a TTL test since the cache uses real `time.Now()`.

**N6: `defer func() { _ = resp.Body.Close() }()` → `defer func(body io.Closer) { _ = body.Close() }(resp.Body)` — inconsistent but harmless**
`pkg/server/quota_test.go` (multiple)

The PR changes existing defers to capture `resp.Body` by value. This is technically more correct (prevents the closure from seeing a reassigned `resp`), but it's applied inconsistently — some tests keep the old pattern. Not a bug, just noise in the diff. Fine to leave as-is but don't mix patterns within the same file.

---

### Architecture Observations

1. **`writeQuotaResponse` now always reads from DB** — good. This eliminates the split between "cloud says X" and "DB says Y" that previously existed. The spending limit is now always served from local state.

2. **Drive9 API key auth for quota GET** — the new `resolveQuotaAPIKey` path is a nice addition. When a Drive9 API key is present AND local spending limit exists, the quota endpoint can be served without any TiDB Cloud call. This is a significant latency improvement for SDK clients.

3. **`quota_limits_overridden` flag** — clean design. Spending-limit-only rows (via `SetQuotaConfigPatch` with only spending limit) set `quota_limits_overridden = 0`, so `GetQuotaConfig` returns process defaults for storage/file limits. Only when a storage quota is explicitly set does it flip to `1`. The `IF(?, 1, quota_limits_overridden)` SQL is subtle but correct.

4. **Batch-write endpoint** appears in this PR (`server.go`) but seems unrelated to the RBAC/spending-limit changes. Consider whether it should be a separate PR to keep the review scope focused.

---

### Verdict: **Conditional GREEN**

B1 (eviction policy) is a correctness concern under edge-case load — suggest fixing or documenting.  
B2 (redundant auth steps) is a performance/clarity issue — suggest simplifying or documenting the rationale.  
B3 (double DB read) is a straightforward optimization.

None of these are ship-blockers if you're comfortable with the current load profile, but B1 could cause surprising behavior at scale.

qiffang:
## Adversarial Review — adversary-2

**Head reviewed**: `2794fa8` (8 commits, 1206 additions, 142 deletions, 14 files)

**Scope**: RBAC cache, spending limit local persistence, `quota_limits_overridden` flag, API key auth path.

**Update**: B1 retracted after adversary-1 pointed out line 1000 guards `cluster.OrganizationID != ""` — empty-org entries are cache misses, not auth failures. B3 downgraded to WARNING.

---

### ~~B1~~ (RETRACTED): ~~`rememberCluster` caches entry with empty `OrganizationID`~~

`listAllManagedClustersCached` already checks `cluster.OrganizationID != ""` before using a cluster cache hit. Empty-org entries from the quota path are harmless cache misses (wasted entry), not ordering-dependent 403s.

---

### B2 (BLOCKER): `syncTiDBCloudSpendingLimit` TOCTOU — stale cloud value can overwrite explicit user update

`pkg/server/quota.go` in `syncTiDBCloudSpendingLimit`:
```go
if !observedAt.IsZero() {
    cfg, err := s.meta.GetQuotaConfig(ctx, tenantID)
    // ...
    if cfg.TiDBCloudSpendingLimit != nil && cfg.UpdatedAt.After(observedAt) {
        // skip — local is newer
        return nil
    }
}
// ... SetQuotaConfigPatch writes the cloud-observed value
```

Between the `GetQuotaConfig` read and the subsequent `SetQuotaConfigPatch` write, a concurrent `quota_set` request can update the spending limit to an explicit user-chosen value. The `syncTiDBCloudSpendingLimit` then overwrites it with a stale cloud-observed value.

The `admin_tenant_list` path is especially vulnerable: `syncAdminTenantSpendingLimits` iterates N rows sequentially, each doing read-check-then-write, while concurrent single-tenant quota updates can interleave freely.

**Both adversary reviewers align on this as the sole remaining blocker.**

---

### W1 (WARNING): `enforceSizeLocked` random eviction

Go map iteration order is pseudo-random. A just-inserted entry can be immediately evicted, degrading to zero effective caching under sustained load near the 10K cap. Not a correctness issue at current scale, but a performance pathology worth addressing.

---

### W2 (WARNING): `adminTenantRowsNeedSpendingLimitBackfill` is O(N) DB queries in list endpoint

For each row in the tenant list page, `GetQuotaConfig` is called individually. On a page of 100 tenants, that's 100 sequential DB round-trips in the list response hot path.

---

### W3 (WARNING): Inconsistent error handling for `syncTiDBCloudSpendingLimit`

- `createFreePoolTenants` (pool create): error logged as warning, execution continues
- `claimAdminTenantFromPool` (pool claim): error causes hard return
- `provisionTenant`: error causes hard return + cluster cleanup

---

### W4 (WARNING): `resolveQuotaAPIKey` hash comparison is tautological

`ResolveByAPIKeyHash` already matched by hash to find the DB row. The subsequent `subtle.ConstantTimeCompare(hash, resolved.APIKey.JWTHash)` is always true by construction. Not a bug, but misleading without a comment.

---

### Verdict: **REVISE** — B2 (TOCTOU on spending limit sync) must be fixed. W1-W4 are non-blocking.

qiffang (changes requested):
## adversary-1 review: request changes

I found one blocker on current head `2794fa8`.

### B1: stale TiDB Cloud quota sync can overwrite a newer explicit quota update

`syncTiDBCloudSpendingLimit` tries to avoid stale cloud reads by comparing local `UpdatedAt` against `observedAt`:

- `pkg/server/quota.go:649-658` reads `GetQuotaConfig` and skips only if `cfg.UpdatedAt.After(observedAt)`.
- `pkg/server/quota.go:660-663` then writes the cloud value with `SetQuotaConfigPatch`.
- `pkg/meta/quota.go:290-304` performs an unconditional upsert/update of `tidbcloud_spending_limit`; it has no `WHERE updated_at <= observedAt`, version check, or compare-and-swap guard.

That leaves a real read-then-write race:

1. Request A (`quota_get`, `admin_tenant_get`, or `admin_tenant_list`) observes an older TiDB Cloud spending limit, e.g. `100`, and enters `syncTiDBCloudSpendingLimit`.
2. A reads local quota config for the staleness check and does not skip.
3. Request B concurrently performs an explicit quota update to `200`; it updates TiDB Cloud and writes the new local value.
4. A resumes and unconditionally writes the stale cloud value `100` back into local metadata.

Impact: Drive9 can return the stale spending limit from local DB and the Drive9 API-key fast path after an explicit quota update succeeded. This violates the PR's central contract that local quota is the latest known TiDB Cloud value after create/update/authorization refresh.

Required fix: make the stale check atomic with the write, e.g. a meta-layer conditional update/upsert guarded by row version/`updated_at <= observedAt` semantics, or another CAS-style mechanism that prevents stale refreshes from writing after a newer local quota update. Add regression coverage for the interleaving where a refresh reads stale cloud data while `quota_set` commits a newer local value.

Validation run from my side:

- `gofmt -l` on changed Go files passed.
- `go test ./pkg/metrics -count=1` passed.
- `go test ./pkg/tenant/tidbcloudnative -count=1` passed.
- `go test -c ./pkg/meta ./pkg/server ./pkg/metrics ./pkg/tenant/tidbcloudnative` passed.
- Targeted runtime `go test ./pkg/meta ./pkg/server -run 'TestQuota|TestAdminTenant|TestGetQuotaConfig|TestMetaSchema|TestDiffMetaTableMetaReportsMissingPrimaryKeyConstraint|TestTiDBCloudRBACCache' -count=1` is locally blocked by `rootless Docker not found` from testcontainers/TestMain. GitHub `ci` and `drive9-server-local e2e smoke` are green.

srstack:
Addressed B1 in `5f137291`.

Root cause was valid: the previous stale guard in `syncTiDBCloudSpendingLimit` checked `UpdatedAt` in application code and then performed an unconditional `SetQuotaConfigPatch`, so an explicit quota update could commit between the check and write.

Fix:

- Added meta-layer conditional write `SetTiDBCloudSpendingLimitIfNotUpdatedAfter`.
- `syncTiDBCloudSpendingLimit` now uses that CAS-style method instead of read-then-write.
- The conditional update only applies a TiDB Cloud spending-limit observation if the local quota row has not been updated since the cloud read began.
- Because `tenant_quota_config.updated_at` is `DATETIME(3)`, the guard uses a strict millisecond cutoff: `updated_at < observedAt.Truncate(time.Millisecond)`. Writes in the same DB timestamp tick as the observation are treated as newer and skipped, which is conservative but prevents stale cloud values from overwriting explicit local updates.
- When the guard skips, the existing `skipped_newer_local` metric result is recorded.

Regression coverage added:

- `TestSetTiDBCloudSpendingLimitIfNotUpdatedAfterSkipsNewerLocal` verifies the meta conditional write skips stale observations and applies fresh ones.
- `TestSyncTiDBCloudSpendingLimitSkipsNewerLocalAtSameTimestampTick` covers the server path where the old `UpdatedAt.After(observedAt)` check would not skip on the same timestamp tick and could overwrite the newer local value.

Validation run:

```bash
go test ./pkg/meta ./pkg/server -run 'TestSetTiDBCloudSpendingLimitIfNotUpdatedAfterSkipsNewerLocal|TestSyncTiDBCloudSpendingLimitSkipsNewerLocalAtSameTimestampTick|TestQuotaGet|TestSetQuotaStorageBytes|TestQuotaConfigStoresTiDBCloudSpendingLimitWithoutStorageVersion|TestGetQuotaConfigVersion' -count=1
go test ./pkg/metrics ./pkg/meta ./pkg/server ./pkg/client -count=1
git diff --check
```

All passed locally.

qiffang:
Re-reviewed current head `5f137291` after the spending-limit sync fix.

The blocker I cared about is resolved.

What changed:

- `SetQuotaStorageBytes` now sets `quota_limits_overridden=1` on insert/update, so the spending-limit-only row -> storage override regression is fixed.
- `syncTiDBCloudSpendingLimit` now writes through `SetTiDBCloudSpendingLimitIfNotUpdatedAfter`, so the old read-check-then-unconditional-write TOCTOU is gone. The conditional update is at the meta layer and treats same millisecond timestamp-tick writes as newer, which is the conservative side for Drive9's `DATETIME(3)` timestamps.
- Regression coverage exists for both cases:
  - `TestSetQuotaStorageBytesOverridesSpendingOnlyRow`
  - `TestSetTiDBCloudSpendingLimitIfNotUpdatedAfterSkipsNewerLocal`
  - `TestSyncTiDBCloudSpendingLimitSkipsNewerLocalAtSameTimestampTick`

Validation I ran on `5f137291`:

- `go test ./pkg/meta ./pkg/server -run 'TestSetTiDBCloudSpendingLimitIfNotUpdatedAfterSkipsNewerLocal|TestSyncTiDBCloudSpendingLimitSkipsNewerLocalAtSameTimestampTick|TestQuotaGet|TestSetQuotaStorageBytes|TestQuotaConfigStoresTiDBCloudSpendingLimitWithoutStorageVersion|TestGetQuotaConfigVersion' -count=1`
- `go test ./pkg/metrics ./pkg/client -count=1`
- `go vet ./pkg/meta ./pkg/server ./pkg/metrics ./pkg/client`
- `gofmt -l pkg/meta/quota.go pkg/meta/quota_setting_test.go pkg/server/quota.go pkg/server/quota_test.go`
- `git diff --check 5997680085c6921b07c5e92f4f6e4685a4821f32...HEAD`

No remaining blocker from my lane. CodeRabbit's latest note about wrapping the zero-`observedAt` fallback error is a normal cleanup nit, not merge-blocking.

qiffang (approved):

PR #724, opened 2026-07-10
Prevent duplicate TiDB Cloud cluster bindings

srstack:
## Summary
- add branch_id to TiDB Cloud org bindings and backfill it from tenants.branch_id during meta schema migration
- serialize org/cluster/branch binding upserts with a named lock and reject duplicate non-deleted tenant bindings
- keep shared cluster support for distinct branch IDs and add regression coverage

## Tests
- go test ./pkg/meta -count=1
- go test ./pkg/server -run 'TestAdminTenantPool|TestAdminTenantCreateClaimsFreePoolTenant|TestAdminTenantGet|TestAdminTenantList' -count=1\n- go test ./pkg/server -run 'TestFork|TestTenantDeleteRejectsNonDeletedFork' -count=1\n- git diff --check -- pkg/meta/meta.go pkg/meta/meta_test.go

## Summary by CodeRabbit

* **New Features**
  * TiDB Cloud organization bindings now track their associated branch.
  * Shared clusters can support bindings across different branches.

* **Bug Fixes**
  * Prevented duplicate active bindings for the same organization, cluster, and branch.
  * Deleted cluster branches are ignored when creating bindings.
  * Existing bindings are automatically populated with branch information during migration.

### DIFF
--- a/pkg/meta/meta.go
+++ b/pkg/meta/meta.go
@@ -242,6 +242,7 @@ type TenantTiDBCloudOrgBinding struct {
 	TenantID       string
 	OrganizationID string
 	ClusterID      string
+	BranchID       string
 	PoolID         string
 	PoolStatus     TenantPoolBindingStatus
 	UsedAt         *time.Time
@@ -252,6 +253,11 @@ type TenantTiDBCloudOrgBinding struct {
 type TenantWithTiDBCloudOrgBinding struct {
 	Tenant  Tenant
 	Binding TenantTiDBCloudOrgBinding
+}
+
+type metaQueryExecer interface {
+	ExecContext(ctx context.Context, query string, args ...any) (sql.Result, error)
+	QueryRowContext(ctx context.Context, query string, args ...any) *sql.Row
 }
 
 type TenantPoolBindingStatus string
@@ -308,6 +314,8 @@ const metaSchemaMigrateLockNamePrefix = "dat9_meta_schema_migrate:"
 const metaSchemaMigrateLockTimeoutSeconds = 30
 const externalBindingLockTimeoutSeconds = 90
 const externalBindingReleaseLockTimeout = 5 * time.Second
+const tidbCloudOrgBindingLockTimeoutSeconds = 90
+const tidbCloudOrgBindingReleaseLockTimeout = 5 * time.Second
 const tenantPoolLockTimeoutSeconds = 300
 const tenantPoolReleaseLockTimeout = 5 * time.Second
 
@@ -336,6 +344,9 @@ func (s *Store) migrate() (err error) {
 		return fmt.Errorf("diff meta schema: %w", err)
 	}
 	if err := applyMetaSchemaRepairs(ctx, s.db, plannedMetaSchemaRepairs(diffs)); err != nil {
+		return err
+	}
+	if err := backfillTiDBCloudOrgBindingBranchIDs(ctx, s.db); err != nil {
 		return err
 	}
 	diffs, err = diffMetaSchema(ctx, s.db, spec)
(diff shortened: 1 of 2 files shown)


srstack:
Reviewed head `192e6e51`. Verified locally: build, `go vet`, gofmt clean; the three new binding tests + schema-spec test pass.

**The design is correct and the concurrency-critical part is done right.** The duplicate guard (`ensureTiDBCloudOrgBindingAvailable`) and the INSERT run inside the same named-lock scope on the same pooled connection (`withTiDBCloudOrgBindingLock`), so the check-then-insert is atomic against a concurrent same-(org, cluster, branch) upsert — no TOCTOU. The lock helper mirrors the existing `withExternalBindingLock` pattern (dedicated conn, `GET_LOCK`/`RELEASE_LOCK`, database-scoped lock name via `tenantPoolDatabaseLockName`, release-on-defer with joined errors), so it's consistent with the established approach.

**Migration ordering is correct** and I checked it specifically:
1. `dropObsoleteMetaIndexes` drops the old `idx_tidbcloud_org_cluster` — but only via `dropMetaIndexIfColumns`, which exact-matches the *old* 4-column signature (`org, cluster, created, tenant`). So on a re-run after migration (where the index already has `branch_id`), `sameStringSlice` returns false and it won't drop the new index. Idempotent.
2. repairs add the `branch_id` column + create the new branch-aware index.
3. `backfillTiDBCloudOrgBindingBranchIDs` runs *after* the column exists, backfilling from `tenants.branch_id`.
4. re-diff confirms convergence.

The backfill UPDATE is guarded with `WHERE b.branch_id <> COALESCE(t.branch_id, '')` so it only touches rows that actually differ — cheap on re-runs.

**Backward-compat is sound:** `branch_id` is `NOT NULL DEFAULT ''`, so existing rows get `''` and the backfill then sets the real value from tenants. The duplicate guard scopes on `(org, cluster, branch_id)` and excludes `TenantDeleted`, so shared-cluster-across-branches (fork tenants) still works and a deleted tenant's cluster+branch can be reused — both covered by the new `AllowsSharedClusterAcrossBranches` and `IgnoresDeletedClusterBranch` tests.

**Two minor notes (non-blocking):**
1. `lookupTenantBranchID` runs on `s.db` *before* acquiring the lock, while the guard+insert run inside it. That's fine because the resolved `branchID` is what both the lock name and the guard use consistently — but note the guard resolves branch from the *tenants* row via lookup, overriding any `b.BranchID` the caller passed. Worth a one-line comment that the tenant row is the source of truth for branch_id (callers can't override it), since it's a slightly surprising override.
2. `ensureTiDBCloudOrgBindingAvailable` does a `JOIN tenants` + `LIMIT 1` on `(org, cluster, branch)` — the new index `idx_tidbcloud_org_cluster(org, cluster, branch, created, tenant)` covers the lookup, so this should be an index seek, good. Just confirm the join on `t.status` doesn't force a full-tenants scan on large orgs; likely fine given the index prefix narrows to a handful of rows first.

Solid, focused change. LGTM.

qiffang:
Reviewed current head `4b39767bfd14d8cf00ea55d5bb66967f00f18984`.

Findings: no blocker from my lane.

What I checked:
- The duplicate guard is scoped to `(organization_id, cluster_id, branch_id)` and uses `tenants.branch_id` as the source of truth, so callers cannot bypass the branch dimension by passing a different `TenantTiDBCloudOrgBinding.BranchID`.
- The availability check and the subsequent `INSERT ... ON DUPLICATE KEY UPDATE` both run on the same `*sql.Conn` while the TiDB Cloud org-binding named lock is held, so the race concern I had with a check-then-insert shape is addressed in this implementation.
- Deleted tenants are excluded from the duplicate check and tenant delete/finalize paths remove the binding row, so reuse after delete is preserved.
- The migration order looks safe: drop only the old exact index shape, apply schema repair for `branch_id` + new index, then backfill from `tenants.branch_id` before final schema diff.
- Server-facing error paths I checked do not return the raw duplicate message with the existing tenant id to clients; they wrap persistence failures into generic provision/pool messages.

Non-blocking note: this prevents new duplicate live bindings, but it does not add a database unique constraint or an explicit migration audit for duplicates that may already exist in production. If we suspect existing duplicate rows, run a one-off ops query/cleanup before relying on the new invariant operationally.

Validation run locally:
- `go test ./pkg/meta -run 'TestUpsertTenantTiDBCloudOrgBinding|TestBackfillTiDBCloudOrgBindingBranchIDs|TestMetaSchemaSpecIncludesTiDBCloudOrgBindings|TestClaimOldestFreeTenantPoolBindingRequiresActiveTenant' -count=1`
- `go test ./pkg/server -run 'TestAdminTenantPool|TestAdminTenantCreateClaimsFreePoolTenant|TestAdminTenantGet|TestAdminTenantList' -count=1`
- `go test ./pkg/server -run 'TestFork|TestTenantDeleteRejectsNonDeletedFork' -count=1`
- `go vet ./pkg/meta ./pkg/server`
- `gofmt -l pkg/meta/meta.go pkg/meta/meta_test.go`
- `git diff --check -- pkg/meta/meta.go pkg/meta/meta_test.go`

qiffang (approved):

PR #725, opened 2026-07-10
Add unique TiDB Cloud binding index

srstack:
## Summary
- add uk_tidbcloud_org_cluster_branch on organization_id, cluster_id, branch_id for TiDB Cloud org bindings
- create the unique index through an explicit migration step after branch_id backfill, stale deleted/orphan binding cleanup, and duplicate preflight
- make duplicate preflight diagnostics report the total duplicate tuple count and the first 20 conflicting tuples instead of only the first conflict
- clean same-tuple deleted/orphan conflicts inside the upsert lock before enforcing the non-deleted duplicate guard
- keep the existing non-unique org/cluster/created/tenant index for ordered listing coverage; it is not redundant with the new unique tuple index

## Branch ID behavior
branch_id remains NOT NULL DEFAULT empty string. Missing branch IDs are stored as the root branch dimension, so inserts do not fail just because branch_id is absent. They will conflict only when another binding already owns the same organization_id, cluster_id, branch_id tuple, which is the intended invariant.

## Deployment note
Before rolling this migration to production, run a duplicate check for active TiDB Cloud org bindings. If historical data contains more than one non-deleted tenant for the same organization_id, cluster_id, branch_id tuple, the migration preflight will fail and block service startup until the data is manually cleaned up. This is intentional safe-fail behavior so the unique index is not created over ambiguous ownership.

```sql
SELECT
  b.organization_id,
  b.cluster_id,
  b.branch_id,
  COUNT(*) AS tenant_count,
  GROUP_CONCAT(b.tenant_id ORDER BY b.tenant_id) AS tenant_ids
FROM tenant_tidbcloud_org_bindings b
JOIN tenants t ON t.id = b.tenant_id
WHERE t.status <> 'deleted'
GROUP BY b.organization_id, b.cluster_id, b.branch_id
HAVING COUNT(*) > 1
ORDER BY tenant_count DESC, b.organization_id, b.cluster_id, b.branch_id;
```

## Tests
- go test ./pkg/meta -count=1
- go test ./pkg/server -run "TestAdminTenantPool|TestAdminTenantCreateClaimsFreePoolTenant|TestAdminTenantGet|TestAdminTenantList|TestFork|TestTenantDeleteRejectsNonDeletedFork" -count=1
- git diff --check -- pkg/meta/meta.go pkg/meta/meta_test.go

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented duplicate TiDB Cloud organization bindings for the same cluster and branch.
  * Automatically cleans up stale bindings associated with deleted or unavailable tenants.
  * Improved handling of conflicting bindings during updates to preserve data consistency.

### DIFF
--- a/pkg/meta/meta.go
+++ b/pkg/meta/meta.go
@@ -42,6 +42,7 @@ const (
 )
 
 const tidbCloudNativeProvider = "tidb_cloud_native"
+const maxTiDBCloudOrgBindingDuplicateTuples = 20
 
 type TenantKind string
 
@@ -350,6 +351,12 @@ func (s *Store) migrate() (err error) {
 	if err := backfillTiDBCloudOrgBindingBranchIDs(ctx, s.db); err != nil {
 		return err
 	}
+	if err := deleteStaleTiDBCloudOrgBindings(ctx, s.db); err != nil {
+		return err
+	}
+	if err := ensureTiDBCloudOrgBindingUniqueIndex(ctx, s.db); err != nil {
+		return err
+	}
 	diffs, err = diffMetaSchema(ctx, s.db, spec)
 	if err != nil {
 		return fmt.Errorf("re-diff meta schema: %w", err)
@@ -599,6 +606,7 @@ func metaInitSchemaStatements() []string {
 				used_at         DATETIME(3) NULL,
 				created_at      DATETIME(3) NOT NULL DEFAULT CURRENT_TIMESTAMP(3),
 				updated_at      DATETIME(3) NOT NULL DEFAULT CURRENT_TIMESTAMP(3) ON UPDATE CURRENT_TIMESTAMP(3),
+				UNIQUE INDEX uk_tidbcloud_org_cluster_branch (organization_id, cluster_id, branch_id),
 				INDEX idx_tidbcloud_org_cluster (organization_id, cluster_id, branch_id, created_at, tenant_id),
 				INDEX idx_tidbcloud_org_created (organization_id, created_at, tenant_id),
 				INDEX idx_tidbcloud_pool_free (organization_id, pool_status, created_at, tenant_id)
@@ -1105,6 +1113,93 @@ func backfillTiDBCloudOrgBindingBranchIDs(ctx context.Context, db *sql.DB) error
 	return nil
 }
 
+func deleteStaleTiDBCloudOrgBindings(ctx context.Context, db metaQueryExecer) error {
+	_, err := db.ExecContext(ctx, `DELETE b
+		FROM tenant_tidbcloud_org_bindings b
+		LEFT JOIN tenants t ON t.id = b.tenant_id
+		WHERE t.id IS NULL OR t.status = ?`, TenantDeleted)
+	if err != nil {
+		return fmt.Errorf("delete stale tidbcloud org bindings: %w", err)
+	}
+	return nil
+}
+
+func ensureTiDBCloudOrgBindingUniqueIndex(ctx context.Context, db *sql.DB) error {
+	exists, err := metaIndexExists(ctx, db, "tenant_tidbcloud_org_bindings", "uk_tidbcloud_org_cluster_branch")
+	if err != nil {
+		return fmt.Errorf("check tidbcloud org binding unique index: %w", err)
+	}
+	if exists {
+		return nil
+	}
+	if err := ensureNoDuplicateTiDBCloudOrgBindingTuples(ctx, db); err != nil {
+		return fmt.Errorf("preflight tidbcloud org binding unique index: %w", err)
+	}
+	_, err = db.ExecContext(ctx, `CREATE UNIQUE INDEX uk_tidbcloud_org_cluster_branch
+		ON tenant_tidbcloud_org_bindings(organization_id, cluster_id, branch_id)`)
+	if err != nil {
+		if isIgnorableMetaSchemaError(err) {
+			return nil
+		}
+		return fmt.Errorf("create tidbcloud org binding unique index: %w", err)
+	}
+	return nil
+}
+
+func metaIndexExists(ctx context.Context, db *sql.DB, tableName, indexName string) (bool, error) {
+	var count int
+	if err := db.QueryRowContext(ctx, `SELECT COUNT(*)
+		FROM information_schema.statistics
+		WHERE table_schema = DATABASE() AND table_name = ? AND index_name = ?`,
+		tableName, indexName).Scan(&count); err != nil {
+		return false, fmt.Errorf("check meta index %s.%s exists: %w", tableName, indexName, err)
+	}
+	return count > 0, nil
+}
+
+func ensureNoDuplicateTiDBCloudOrgBindingTuples(ctx context.Context, db *sql.DB) error {
+	var total int
+	if err := db.QueryRowContext(ctx, `SELECT COUNT(*) FROM (
+		SELECT organization_id, cluster_id, branch_id
+		FROM tenant_tidbcloud_org_bindings
+		GROUP BY organization_id, cluster_id, branch_id
+		HAVING COUNT(*) > 1
+	) duplicates`).Scan(&total); err != nil {
+		return fmt.Errorf("count duplicate tidbcloud org binding tuples: %w", err)
+	}
+	if total == 0 {
+		return nil
+	}
+
+	rows, err := db.QueryContext(ctx, `SELECT organization_id, cluster_id, branch_id, COUNT(*) AS n, GROUP_CONCAT(tenant_id ORDER BY tenant_id)
+		FROM tenant_tidbcloud_org_bindings
+		GROUP BY organization_id, cluster_id, branch_id
+		HAVING COUNT(*) > 1
+		ORDER BY n DESC, organization_id, cluster_id, branch_id
+		LIMIT ?`, maxTiDBCloudOrgBindingDuplicateTuples)
+	if err != nil {
+		return fmt.Errorf("list duplicate tidbcloud org binding tuples: %w", err)
+	}
+	defer func() { _ = rows.Close() }()
+
+	conflicts := make([]string, 0, min(total, maxTiDBCloudOrgBindingDuplicateTuples))
+	for rows.Next() {
+		var organizationID, clusterID, branchID, tenantIDs string
+		var count int
+		if err := rows.Scan(&organizationID, &clusterID, &branchID, &count, &tenantIDs); err != nil {
+			return fmt.Errorf("scan duplicate tidbcloud org binding tuple: %w", err)
+		}
+		conflicts = append(conflicts, fmt.Sprintf("organization_id=%q cluster_id=%q branch_id=%q count=%d tenant_ids=%s", organizationID, clusterID, branchID, count, tenantIDs))
+	}
+	if err := rows.Err(); err != nil {
+		return fmt.Errorf("iterate duplicate tidbcloud org binding tuples: %w", err)
+	}
+	if total > len(conflicts) {
+		conflicts = append(conflicts, fmt.Sprintf("... %d more duplicate tuple(s) omitted", total-len(conflicts)))
+	}
+	return fmt.Errorf("cannot create tidbcloud org binding unique index: found %d duplicate tuple(s): %s", total, strings.Join(conflicts, "; "))
+}
+
 func parseMetaCreateTableStatement(stmt string) (tableName string, definitions string, ok bool, err error) {
 	return schemaspec.ParseCreateTableStatement(stmt)
 }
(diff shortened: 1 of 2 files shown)


srstack:
Reviewed head `d5bb20e7` (follow-up to #724, which is now merged so the `branch_id` column dependency is satisfied). Verified locally: build, `go vet`, gofmt clean; the binding + duplicate-preflight tests pass.

**The DB-level unique constraint is the right hardening on top of #724's app-level guard** — it makes `(organization_id, cluster_id, branch_id)` uniqueness enforced by the database, not just by the lock-guarded upsert path, so any code path (or manual write) that bypasses `UpsertTenantTiDBCloudOrgBinding` can't reintroduce a duplicate.

**Migration robustness is handled well:**
- The whole `migrate()` runs under `acquireMetaSchemaMigrationLock` (database-scoped), so multi-pod startup serializes — no concurrent `CREATE UNIQUE INDEX`.
- `ensureTiDBCloudOrgBindingUniqueIndex` is idempotent: `metaIndexExists` short-circuits if already present, and a racing create is absorbed by `isIgnorableMetaSchemaError`.
- Ordering is correct: backfill branch_id → delete stale/orphan bindings → preflight duplicate check → create index → re-diff. The index is only attempted after the data is in a state that can satisfy it.
- `deleteStaleTiDBCloudOrgBindings` (orphan tenant OR `TenantDeleted`) clears the *reclaimable* conflicts automatically, and the same cleanup is mirrored inside the upsert lock (`deleteStaleTiDBCloudOrgBindingConflicts`) before the availability guard — so a reused cluster+branch from a deleted tenant is cleaned in-band rather than tripping the unique index.

**One operational risk worth calling out explicitly (not a code bug — a deploy consideration):**
`ensureNoDuplicateTiDBCloudOrgBindingTuples` only reports duplicates among **live** tenants (the stale cleanup already removed deleted/orphan ones). If production has two *non-deleted, active* tenants that historically share the same `(org, cluster, branch)` tuple — which was allowed before #724's guard landed — the preflight returns an error, `CREATE UNIQUE INDEX` is skipped, and `migrate()` fails, which will **block server startup** until an operator manually resolves the conflict. This is a safe failure (no silent data loss, and the error names the org/cluster/branch/tenant_ids), but it's a hard startup gate. Two suggestions:
1. The preflight uses `LIMIT 1`, so it surfaces only the *first* duplicate group. If there are several, an operator will hit them one migration attempt at a time. Consider logging *all* conflicting groups (or at least a count) so the cleanup can be done in one pass rather than iteratively.
2. Worth a heads-up in the deploy notes / release checklist to run the duplicate-detection query against prod before rolling this out, so the startup failure (if any) is discovered ahead of the deploy rather than during it.

**Minor:** `idx_tidbcloud_org_cluster (org, cluster, branch, created, tenant)` is now a strict superset-prefix of the new unique index `(org, cluster, branch)`. The unique index can serve as the prefix for equality lookups on those three columns, but the old index is still needed for the `created, tenant` ordering/covering, so keeping both is correct — just noting it's intentional, not redundant.

Solid, well-sequenced migration. LGTM — I'd just pair it with the pre-deploy duplicate check in point 1/2.

srstack:
Re-checked `d51d9bd6`. The delta since my last review is purely error-context wrapping — `fmt.Errorf("...: %w", err)` added to `ensureTiDBCloudOrgBindingUniqueIndex`, `metaIndexExists`, and `deleteStaleTiDBCloudOrgBindingConflicts`. No logic change, and it improves diagnosability of a migration failure (which is exactly the startup-gate scenario I flagged), so it's a strict improvement. Build/vet/gofmt clean, tests still pass.

My earlier assessment stands — LGTM. The only open item is operational, not code: the pre-deploy duplicate check (and ideally logging *all* conflicting tuple groups rather than just the first via `LIMIT 1`) so a historical live-vs-live duplicate doesn't surface as a startup failure mid-deploy.

srstack:
Re-reviewed `d2340060`. This directly addresses point #1 from my earlier review — the preflight no longer stops at the first duplicate (`LIMIT 1`). It now:
- counts the *total* number of duplicate `(org, cluster, branch)` groups,
- lists up to `maxTiDBCloudOrgBindingDuplicateTuples` (20) of them ordered by count desc, and
- appends "... N more duplicate tuple(s) omitted" when there are more.

So an operator hitting the startup gate now gets the full (or a bounded, representative) picture in one shot instead of iterating one migration attempt per conflict. The test was expanded to two distinct duplicate tuples and asserts "found 2 duplicate tuple(s)" plus all four tenant IDs appear. Build/vet/gofmt clean; tests pass (builtin `min()` is fine on Go 1.25).

The bounded LIMIT is a good call too — it avoids an unbounded error string if the data is badly degraded, while the total count still tells the operator the real scale.

That closes the one substantive item I raised. Remaining suggestion is now purely process, not code: run the duplicate-detection query against prod before rolling out so any live-vs-live historical duplicate is resolved ahead of the deploy rather than surfacing as a startup failure. LGTM.

qiffang:
Reviewed head `d2340060` for PR #725.

Verdict: **LGTM / no blocker**. This is the right follow-up to PR #724: #724 added the app-level duplicate guard, and this PR makes `(organization_id, cluster_id, branch_id)` a DB-enforced invariant so bypass/manual-write paths cannot reintroduce duplicate TiDB Cloud org bindings.

What I checked:
- Migration sequencing is correct for existing tables: normal safe schema repair will not auto-create this new unique index on an existing table, so the explicit path owns it: branch backfill -> stale/orphan binding cleanup -> duplicate preflight -> `CREATE UNIQUE INDEX` -> final re-diff.
- The migration-wide stale cleanup removes only orphan/deleted tenant bindings before the preflight. Non-deleted duplicate ownership safe-fails startup instead of silently choosing a winner.
- Duplicate diagnostics are now useful: the preflight counts total duplicate tuples, lists up to 20 conflict groups with tenant IDs, and reports omitted count if there are more.
- Runtime upsert mirrors the stale-conflict cleanup inside the named tuple lock before availability checking, so a deleted/orphan tenant row does not block valid cluster+branch reuse.
- The existing non-unique `idx_tidbcloud_org_cluster (org, cluster, branch, created, tenant)` still makes sense alongside the unique index because it preserves the ordered/covering listing shape.

Validation I ran:
- `go test ./pkg/meta -run 'TestMetaSchemaSpecIncludesTiDBCloudOrgBindings|TestEnsureNoDuplicateTiDBCloudOrgBindingTuplesReportsConflict|TestBackfillTiDBCloudOrgBindingBranchIDsTrimsTenantBranchID|TestUpsertTenantTiDBCloudOrgBinding' -count=1`
- `go test ./pkg/server -run 'TestAdminTenantPool|TestAdminTenantCreateClaimsFreePoolTenant|TestAdminTenantGet|TestAdminTenantList|TestFork|TestTenantDeleteRejectsNonDeletedFork' -count=1`
- `go vet ./pkg/meta ./pkg/server`
- `gofmt -l pkg/meta/meta.go pkg/meta/meta_test.go`
- GitHub checks are green: `ci`, `drive9-server-local e2e smoke`

Non-blocking rollout note: run the PR body's duplicate-detection SQL against prod before deploy. If historical non-deleted duplicates exist, this PR intentionally blocks startup until ops resolves them; that is the right safe failure mode, but it should be discovered before rollout rather than during it.

mornyx (approved):
## Review

Overall this is a well-structured migration — backfill → stale cleanup → preflight → create index, with the ordering correctly gated before the re-diff. The upsert-path stale-conflict cleanup is a nice touch. A few concerns below, the most important being the `ON DUPLICATE KEY UPDATE` semantics change.

---

### 🔴 `ON DUPLICATE KEY UPDATE` is now ambiguous between PK and the new unique index

`UpsertTenantTiDBCloudOrgBinding` uses `INSERT ... ON DUPLICATE KEY UPDATE` (meta.go:1623). Before this PR only the `tenant_id` PRIMARY KEY could trigger the UPDATE branch. With the new `UNIQUE INDEX (organization_id, cluster_id, branch_id)`, a collision on that index against a **different** live tenant now also enters the UPDATE branch — and because `tenant_id` is NOT in the `SET` clause, MySQL would update the *existing* (other tenant's) row, rewriting its `organization_id/cluster_id/branch_id/pool_*` fields while keeping the other tenant's `tenant_id`. That is a data-corruption vector if the application-level guard ever lets a live-owner collision through.

The guard chain (`deleteStaleTiDBCloudOrgBindingConflicts` → `ensureTiDBCloudOrgBindingAvailable` → INSERT) is meant to prevent this, and under the named lock it should be airtight for the deleted-tenant case. But the latent risk is real enough to warrant:

1. A code comment on the `INSERT ... ON DUPLICATE KEY UPDATE` explaining that the preflight `ensureTiDBCloudOrgBindingAvailable` is what prevents the unique-index collision path, and
2. Ideally, making the UPDATE clause a no-op when the conflict is NOT on the PK — e.g. `ON DUPLICATE KEY UPDATE updated_at = VALUES(updated_at)` only when `tenant_id = VALUES(tenant_id)`, or splitting into an explicit `UPDATE` for the same-tenant case. As written, a future refactor that drops or weakens `ensureTiDBCloudOrgBindingAvailable` would silently corrupt rows.

---

### 🟠 Race window between preflight COUNT and CREATE UNIQUE INDEX

`ensureTiDBCloudOrgBindingUniqueIndex` runs `ensureNoDuplicateTiDBCloudOrgBindingTuples` (a `SELECT COUNT(*)`) and then `CREATE UNIQUE INDEX` as two separate statements with no transaction (meta.go:1127–1147). `migrate` holds `acquireMetaSchemaMigrationLock` (a named lock), but app writes go through `withTiDBCloudOrgBindingLock` — a **different** named lock keyed by `(org, cluster, branch)`. A concurrent upsert that inserts a duplicate tuple between the preflight COUNT and the `CREATE UNIQUE INDEX` would surface a raw MySQL 1062, which `isIgnorableMetaSchemaError` does **not** swallow (it only tolerates 1050/1060/1061). That hard-fails startup with a less helpful error than the preflight.

In practice migration usually runs before traffic, so this is low-likelihood, but worth a comment noting the assumption (single-writer / pre-traffic migration). If you want to close it, consider taking a brief table-level write lock or wrapping the preflight + create in a transaction with `LOCK TABLES` — though that may be heavier than warranted.

---

### 🟡 `metaIndexExists` duplicates the existence check in `dropMetaIndexIfExists`

The new `metaIndexExists` (meta.go:1149) runs the exact same `information_schema.statistics COUNT(*)` query already in `dropMetaIndexIfExists` (meta.go:802) and overlaps with `loadMetaIndexNames` (meta.go:1176). It also skips `validateMetaIdentifier` that its sibling applies. Consider refactoring `dropMetaIndexIfExists` to call `metaIndexExists` (adding `validateMetaIdentifier` to `metaIndexExists` for consistency), so there is one index-existence helper.

---

### 🟡 `GROUP_CONCAT` truncation in preflight diagnostics

`ensureNoDuplicateTiDBCloudOrgBindingTuples` uses `GROUP_CONCAT(tenant_id ORDER BY tenant_id)` (meta.go:1180). MySQL's default `group_concat_max_len` is 1024, so with many duplicate tenants the tenant-id list is silently truncated in the diagnostic message. This is only diagnostic so it is not a correctness issue, but the truncation is invisible — worth either noting in the error string (`tenant_ids=(truncated)`) or setting `group_concat_max_len` in the session if a complete list matters.

---

### 🟡 `deleteStaleTiDBCloudOrgBindings` is a broader data mutation than the index needs

During `migrate`, `deleteStaleTiDBCloudOrgBindings` (meta.go:1116) deletes **every** binding whose tenant is missing or deleted, not just those that would block the unique index. The PR body says this is intentional, but it is a data-destroying migration step that runs on every startup `migrate`, not just the first upgrade. If orphan/deleted-tenant bindings are ever expected to be retained (audit, reattach), this would quietly remove them on every boot. Worth confirming that binding rows for deleted tenants are truly disposable, and ideally guarding the global delete so it only runs when the unique index does not yet exist (move it inside `ensureTiDBCloudOrgBindingUniqueIndex` before the preflight, rather than always running in `migrate`).

---

### 🟢 Test coverage

The new `TestEnsureNoDuplicateTiDBCloudOrgBindingTuplesReportsConflict` covers the duplicate-reporting path well. Two suggestions:

- Add a happy-path assertion for `ensureTiDBCloudOrgBindingUniqueIndex` itself (no duplicates → index created → returns nil), not just the direct `ensureNoDuplicateTiDBCloudOrgBindingTuples` call. The cleanup currently restores the index via `ensureTiDBCloudOrgBindingUniqueIndex`, but its success isn't asserted.
- Consider a test for `deleteStaleTiDBCloudOrgBindingConflicts` in the upsert path: a deleted-tenant binding occupying the target tuple should be removed and the live upsert should succeed — that is the core new behavior and currently only exercised indirectly by `TestUpsertTenantTiDBCloudOrgBindingIgnoresDeletedClusterBranch`.

---

### 🟢 Minor

- The test drops the unique index and restores it in `t.Cleanup`. If the test process is killed mid-run the shared control DB is left without the index, contaminating later packages. Low risk given sequential test execution, but worth noting if `newControlStore` ever shares a DB across packages.
- `idx_tidbcloud_org_cluster (organization_id, cluster_id, branch_id, created_at, tenant_id)` is now a superset whose left-prefix is the new unique index. The PR body justifies keeping it for ordered-listing coverage, which is reasonable. Just flagging the write-amplification of two overlapping indexes.

PR #731, opened 2026-07-11
Reduce replay and DB metrics noise

srstack:
## Summary
- treat quota mutation replay rows that were already marked applied by the synchronous path as benign skipped work instead of replay failures
- remove top-level `error` fields from benign/retryable tenant schema DDL logs so skipped-existing and transient DDL races do not inflate error-log alerts
- suppress empty tenant-level DB pool metrics while keeping role-level pool stats visible
- raise default cached tenant user DB pool limits from 4/1 to 6/2 and sync CLI help text

## Details
- `MarkMutationAppliedTx` now distinguishes `status='applied'` from other non-pending statuses after a guarded update affects 0 rows.
- `MutationReplayWorker` counts already-applied races as `skipped_already_applied`, does not increment retry counts, does not record failure metrics, and does not trigger the per-tenant failure barrier.
- Tenant DB pool metrics omit zero-value `connections`, `wait`, and `closes` series for tenant-labeled pools, but still emit `drive9_db_pool_registered` and low-cardinality role-only pool stats.

## Tests
- go test ./pkg/backend -run 'TestMutationReplayWorker|TestRoundA_Fix3_ReplayWorker|TestRoundA_Fix2|TestMutationReplayWorkerAlreadyAppliedMarkDoesNotFailOrBlockTenant' -count=1
- go test ./pkg/metrics ./pkg/tenant/schema ./pkg/tenant ./pkg/mysqlutil ./pkg/meta -count=1
- git diff --check

## Summary by CodeRabbit

- **Improvements**
  - Increased default user database connection pool limits to 6 open and 2 idle connections.
  - Mutation replay now handles already-applied mutations without treating them as failures, retries, or tenant-blocking events.
  - Added reporting for mutations skipped because they were already applied.
  - Database pool metrics no longer emit zero-valued tenant-specific series, improving monitoring clarity.
  - Improved schema-operation logs by providing clearer textual reasons for skipped and failed operations.


srstack:
Reviewed head `123ee9e3`. Verified (in an isolated worktree, not the main checkout): `go build ./...`, `go vet`, gofmt all clean; the replay-worker, metrics, meta, and schema tests pass.

**All four changes are sound.**

**1. Already-applied replay race (the substantive one).** `MarkMutationAppliedTx` now, on a 0-row guarded update, does a `SELECT status ... FOR UPDATE` to distinguish `applied` (returns the `errMutationAlreadyApplied` sentinel) from other non-pending states (returns a descriptive error). `MutationReplayWorker` treats the sentinel as `skipped_already_applied`: no retry increment, no failure metric, no per-tenant barrier trip. This is correct — a row the synchronous path already applied is benign, not a replay failure, and the new `TestMutationReplayWorkerAlreadyAppliedMarkDoesNotFailOrBlockTenant` proves the *following* same-tenant entry isn't blocked by the barrier. The extra `SELECT` only runs on the 0-row miss path (rare) and stays inside the same tx, so no correctness/perf concern.

**2. Interface addition is complete.** `IsMutationAlreadyAppliedError` was added to the `MetaQuotaStore` interface and implemented on both the real `metaQuotaAdapter` (delegating to `meta.IsMutationAlreadyAppliedError`) and the fake — `go build ./...` confirms no implementation was missed.

**3. Metrics de-noise.** Skipping zero-value tenant-labeled series (`connections`/`wait`/`closes`) while keeping `drive9_db_pool_registered` and role-only stats is the right cardinality trade — idle cached tenant pools no longer emit a wall of zero series, but role-level pool health stays visible. `hasTenantID()` cleanly centralizes the tenant-vs-role distinction, and tests cover both the omit (tenant, zero) and retain (role-only zero; tenant with open conns) cases.

**4. Schema-log `zap.Error` → `zap.String("skip_reason"/"reason")`.** Removing the top-level `error` field from benign skipped-existing and transient DDL-race logs prevents those expected outcomes from inflating error-log-based alerts, while the message and reason text are preserved. Consistent with the earlier schema-init noise reduction.

**5. Pool defaults 4/1 → 6/2** with CLI help synced — reasonable headroom bump for cached tenant user DB pools; no logic risk.

Clean, well-tested observability + correctness change. LGTM.

pkg/meta/quota.go
```diff
@@ -1298,7 +1304,14 @@ func (s *Store) MarkMutationAppliedTx(tx *sql.Tx, id int64) error {
 	}
 	n, _ := res.RowsAffected()
 	if n == 0 {
-		return fmt.Errorf("mutation %d already applied or not pending", id)
+		var status string
+		if err := tx.QueryRow(`SELECT status FROM quota_mutation_log WHERE id = ? FOR UPDATE`, id).Scan(&status); err != nil {
```
mornyx: Nit / lock-interaction question: This `SELECT ... FOR UPDATE` acquires a row lock on a row that is already in a terminal state (`applied` or `failed`) inside a transaction that is about to roll back because `replayOne` will return this error.

On the already-applied hot path this is an extra lock acquire + immediate release (on rollback) against a row another worker just finished writing. It works, but is the lock necessary here? The `UPDATE ... WHERE status='pending'` already executed with its own locking semantics and affected 0 rows, so the row's terminal state is stable. A plain `SELECT status` (no `FOR UPDATE`) would be enough to classify the error, and would avoid taking a write-intent lock on a terminal row that other workers' `IncrMutationRetry`/`MarkMutationAppliedTx` `UPDATE`s also touch.

If `FOR UPDATE` is intentional (e.g. to serialize against a concurrent retry that might flip status back), that's worth a one-line comment. Otherwise consider dropping to a plain `SELECT`.

pkg/meta/quota.go
```diff
@@ -1298,7 +1304,14 @@ func (s *Store) MarkMutationAppliedTx(tx *sql.Tx, id int64) error {
...
+			return fmt.Errorf("mutation %d not pending and status lookup failed: %w", id, err)
+		}
+		if status == "applied" {
+			return fmt.Errorf("mutation %d: %w", id, errMutationAlreadyApplied)
+		}
+		return fmt.Errorf("mutation %d is not pending: status=%q", id, status)
```
mornyx: Nit: This branch returns a non-sentinel error for the `failed`/other-status case, which means `replayBatch` will treat it as a real failure (increment retry, record failure metric, trip the per-tenant barrier). That's the right behavior for a genuinely `failed` row, but note that `IncrMutationRetry` also guards with `WHERE status='pending'`, so a `failed` row here won't actually get its retry bumped — it'll just keep failing every batch until it ages out of `ListPendingMutations` (which only selects `status='pending'`). So in practice this path is only reachable for the `applied` race, which is already handled above. Consider logging the unexpected status here (or asserting it shouldn't happen) rather than returning a generic error that triggers the full failure pipeline for a row that will never be re-listed.

pkg/backend/mutation_replay.go
```diff
@@ -153,13 +153,18 @@ func (w *MutationReplayWorker) replayBatch(ctx context.Context) (fatal bool) {
...
 	for _, entry := range entries {
 		if _, blocked := blockedTenants[entry.TenantID]; blocked {
 			skipped++
 			continue
 		}
 		if err := w.replayOne(ctx, entry); err != nil {
+			if w.store.IsMutationAlreadyAppliedError(err) {
+				skippedAlreadyApplied++
+				continue
+			}
```
mornyx, line 167: Good. One observation: the already-applied case does `continue` *before* the `failed++` / barrier / `IncrMutationRetry` block, so it neither increments `failed` nor `applied`. It's counted only in `skippedAlreadyApplied`. That's the intended semantics per the PR description — just flagging that `applied + failed + skipped + skippedAlreadyApplied == total` only holds if `replayOne` doesn't return any *other* error types; on a real failure the barrier can cause `skipped` to absorb the rest, so the gauges won't necessarily sum to `total`. Not a problem, but the `batch_complete` log might benefit from a note that `skipped_after_barrier` includes barrier-skipped entries, not just benign ones, so operators reading the gauges don't assume `total - applied - failed == already_applied`.

mornyx (commented):
Reviewed locally: checked out the branch, ran the listed tests (`./pkg/backend`, `./pkg/metrics`, `./pkg/tenant/schema`, `./pkg/tenant`, `./pkg/mysqlutil`, `./pkg/meta`) and `golangci-lint` on the changed packages — all green.

The core change is sound: treating the already-applied replay race as benign skipped work is correct, because when `MarkMutationAppliedTx` returns `errMutationAlreadyApplied` the `replayOne` transaction rolls back and discards our duplicate `applyMutation` side effects — exactly what we want when another worker already committed them. Not incrementing retry, not tripping the per-tenant barrier, and not recording failure metrics are all the right calls.

One thing worth a look before merge: the `SELECT ... FOR UPDATE` in `MarkMutationAppliedTx` takes a row lock on a terminal-state row inside a transaction that is about to roll back. It's not wrong, but it's an extra lock acquire on the already-applied hot path that isn't strictly needed to classify the error. See inline comment.

The rest are nits.

pkg/metrics/db.go
```diff
@@ -383,6 +383,9 @@ func writeDBPrometheus(w http.ResponseWriter) {
 	_, _ = fmt.Fprintln(w, "# TYPE drive9_db_pool_connections gauge")
 	for _, key := range keys {
 		totals := poolTotals[key]
+		if key.hasTenantID() && totals.openConnections == 0 {
```
mornyx, line 386: The suppression logic is consistent across the four series (connections/wait_count/wait_duration/closes), gated on `hasTenantID() && value == 0`, which keeps role-only pools visible while dropping zero tenant-labeled series. One thing to double-check operationally: `drive9_db_pool_registered` (line 378-380) is *not* suppressed, so a tenant with an idle pool will still emit `drive9_db_pool_registered{role="user",tenant_id="..."} 1` but no `drive9_db_pool_connections{...}`. That's the intended behavior per the PR description (keep registration visible, drop zero-value series), but make sure any dashboard/alert that joins `pool_registered` to `pool_connections` won't break on the missing connection series for idle tenants. If dashboards rely on `pool_connections{state="open"} 0` to mean "registered but idle", they'd need to fall back to `pool_registered`.

pkg/mysqlutil/pool.go
```diff
@@ -15,8 +15,8 @@ const (
 	defaultMetaMaxIdleConns          = 20
 	defaultUserConnMaxLifetime       = 30 * time.Second
 	defaultUserConnMaxIdleTime       = 30 * time.Second
-	defaultUserMaxOpenConns          = 4
-	defaultUserMaxIdleConns          = 1
+	defaultUserMaxOpenConns          = 6
+	defaultUserMaxIdleConns          = 2
 	defaultUserSchemaConnMaxLifetime = 1 * time.Minute
```
mornyx, line 20: Default open 4→6 and idle 1→2. With `DRIVE9_POOL_MAX_TENANTS` default 1024 cached tenant pools, worst-case idle connection hold goes from `1024*1` to `1024*2` = 2048 idle conns across the pod. That's still well under typical server `max_connections`, but worth confirming against the meta DB / TiDB node connection limits for the production deployment shape (especially multi-pod). The `defaultUserConnMaxLifetime = 30s` and `MaxIdleTime = 30s` should keep the idle set churning, so this is likely fine — just flagging the multiplicative effect.

pkg/backend/quota_migration_test.go
```diff
@@ -481,9 +484,13 @@ func (f *fakeMetaQuotaStore) MarkMutationAppliedTx(tx *sql.Tx, id int64) error {
 	f.markAppliedCalls++
 	for i := range f.mutations {
 		if f.mutations[i].id == id {
-			// Mirror the real Store's WHERE status='pending' guard
-			// (pkg/meta/quota.go). Silent no-op on non-pending — same symmetry
-			// as fake.IncrMutationRetry.
+			if f.alreadyAppliedOnMark[id] {
```
mornyx: The fake now has two mechanisms for the same concept: `alreadyAppliedOnMark[id]` (returns the sentinel error) and the pre-existing `status != "pending"` no-op path. The comment explains the split, which helps. Minor: consider naming the map `alreadyAppliedOnMarkErr` or similar to signal it controls the *error path*, not just the status — a reader could initially think setting `alreadyAppliedOnMark[1]=true` merely pre-marks the row as applied without an error. The test at `round_a_review_test.go:322` makes the behavior clear, so this is purely a readability nit.

pkg/backend/mutation_replay.go
srstack: Confirmed. The already-applied case is intentionally separate from both applied and failed, and `skipped_after_barrier` is specifically the per-tenant barrier skip count while `skipped_already_applied` is the benign race count. No code change needed here.

pkg/metrics/db.go
srstack: Checked the checked-in Grafana dashboards/alerts for this PR. They do not join `drive9_db_pool_registered` to `drive9_db_pool_connections`; user DB panels aggregate `drive9_db_pool_connections` directly, so missing zero-valued idle tenant series is the intended cardinality reduction. `pool_registered` remains available to show registered idle pools.

pkg/mysqlutil/pool.go
srstack: Confirmed. The 6/2 defaults are intentional for the cached tenant user DB pools, and the short 30s max lifetime / idle time bounds the hold time. Operators can still cap this with `DRIVE9_USER_DB_MAX_OPEN_CONNS`, `DRIVE9_USER_DB_MAX_IDLE_CONNS`, and `DRIVE9_POOL_MAX_TENANTS` for tighter deployment-specific connection limits.

mornyx (approved):
Thanks for the updates. Re-reviewed the amended commit (4728e30).

Previous round's concerns are all addressed:
- `MarkMutationAppliedTx`: the `SELECT ... FOR UPDATE` is now a plain `SELECT`, avoiding an unnecessary write-intent lock on a terminal-state row in a transaction that's about to roll back. Good.
- The non-applied/non-pending branch now returns `"unexpectedly left pending state before mark applied: status=%q"`, which makes the invariant-violation semantics explicit rather than a generic "not pending" that could be confused with a normal terminal state.
- Fake field renamed to `alreadyAppliedOnMarkErr`, making clear it controls the error path.
- `ExecOptionalSchemaStatements` `isIgnorableOptionalSchemaError` branch now uses `zap.String("skip_reason", ...)` consistently with the other benign/retryable branches.

Verified locally: `go test` on `./pkg/backend` (targeted runs), `./pkg/metrics`, `./pkg/tenant/schema`, `./pkg/tenant`, `./pkg/mysqlutil`, `./pkg/meta` all pass; `golangci-lint` on the changed packages reports 0 issues.

The remaining nits from the first round (gauge-sum-to-total caveat in `replayBatch`, `pool_registered` vs `pool_connections` dashboard-join note, and the worst-case idle-connection multiplication from the pool default bump) are operational observations rather than blocking issues — fine to address as follow-ups or not at all.

Approving.

PR #734, opened 2026-07-13
fix(metrics): record SSE establishment latency instead of skipping HTTP duration hist

mornyx:
## Problem

We received another `Drive9HighHTTPP99Latency` alert for `/v1/events` with a value of **2000+ seconds**. PR #633 was supposed to fix this, but the fix was incomplete.

### What PR #633 did

PR #633 excluded `/v1/events` from `drive9_http_request_duration_seconds` by recording **only the request counter** (no duration) for established SSE streams. This fixed the connection-lifetime-polluting-p99 problem (a 40s SSE connection no longer looked like a 40s HTTP request).

### What PR #633 missed

Two gaps remained:

1. **`/v1/events` establishment latency became invisible.** The establishment cost (auth + `pool.Acquire` serverless TiDB cold start + write streaming headers) was a real, bounded latency that should be monitored. By recording zero duration samples for successful streams, the events endpoint could get arbitrarily slow without any alert firing.

2. **Bounded error responses still polluted the histogram.** Bounded errors on `/v1/events` (bad auth, invalid `since`, non-GET, `pool.Acquire` timeout) return *before* `markSSEStreamEstablished`, so they fell through to the normal HTTP duration path and recorded full duration. With `httpDurationBounds` maxing out at `le=10`, any serverless TiDB cold-start error exceeding 10s landed in the `+Inf` bucket. With very few `/v1/events` bounded-error samples in a 5m window, `histogram_quantile(0.99, ...)` extrapolated the `+Inf` bucket into **thousands of seconds** — exactly the 2000s+ alert we received.

## Fix

Instead of skipping the duration histogram for established SSE streams, **record the establishment duration** (request → first byte = time from `observe` start to `markSSEStreamEstablished` call). This is a bounded, meaningful latency that belongs in the HTTP p99 alert on the same footing as `/v1/fs/*` etc.

```
request arrival ──auth──pool.Acquire──WriteHeader(200)──flush──► markSSEStreamEstablished
  ^observe start                                                    ^ establishment duration
                                                                    (recorded into HTTP hist)
                                                                    │
                                              select loop (connection lifetime)
                                                                    │  (recorded into
                                                                    │   drive9_sse_connection
                                                                    │   _duration_seconds)
                                                              client disconnect
                                                                    ^observe end
```

Three latency dimensions, three metrics, no pollution:

| Dimension | Metric | Alert | Typical |
|---|---|---|---|
| Establishment (TTFB) | `drive9_http_request_duration_seconds` | `Drive9HighHTTPP99Latency` (GET > 2s) | ms ~ seconds |
| Phase-1 replay | `drive9_sse_phase1_duration_seconds` | `Drive9SSEReplayP95High` (> 2s) | ms |
| Connection lifetime | `drive9_sse_connection_duration_seconds` | not alerted (not latency) | seconds ~ hours |

Bounded error responses on `/v1/events` still fall through to the normal HTTP duration path, but they are now genuine bounded errors (not connection lifetimes), and they are the only `/v1/events` samples in the histogram — no more `+Inf` extrapolation from established streams.

## Changes

- `requestMetricState`: added `sseEstablishedAt` timestamp field.
- `markSSEStreamEstablished`: records `time.Now()` at the establishment point.
- `sseEstablishmentDuration` (new): returns `(sseEstablishedAt - start)` for observe to use.
- `observe`: established SSE branch now records `record()` + `recordTenantHTTPRequest()` with the establishment duration (instead of `recordCount` + `recordTenantHTTPRequestCount` which skipped duration entirely).
- Removed now-unused helpers: `recordCount`, `recordTenantHTTPRequestCount`, `recordTenantHTTPResponseBytes` (the established branch uses the same `record`/`recordTenantHTTPRequest` path as every other request).
- Added `TestSSEEstablishmentDuration` verifying the establishment-timestamp invariant.

## Alert rule (runbooks)

**No change needed.** `/v1/events` stays in `Drive9HighHTTPP99Latency` — it now records a bounded establishment latency on the same footing as other GET routes. The 2000s+ false alert will no longer fire because there are no `+Inf`-bucket connection-lifetime samples to extrapolate.

## Testing

- `go build ./...` ✅
- `go vet ./...` ✅
- `golangci-lint` — 0 issues ✅
- `go test ./pkg/metrics/` ✅
- `go test ./pkg/server/ -run 'SSE|Events|EventBus|Instrument|Observ|Metric|Establish'` ✅ (incl. new `TestSSEEstablishmentDuration`)

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved timing metrics for server-sent event connections by recording the time required to establish the stream.
  * Prevented long-lived SSE connections from distorting overall request-duration metrics.

* **Tests**
  * Added coverage to verify accurate SSE establishment timing and duration bounds.

### DIFF
--- a/pkg/server/instrumentation.go
+++ b/pkg/server/instrumentation.go
@@ -60,12 +60,20 @@ type requestMetricState struct {
 
 	// sseStreamEstablished is set by an SSE handler (handleEvents) once it has
 	// written the 200 streaming headers and entered its long-lived select
-	// loop. observe uses it to distinguish a real SSE connection lifetime
-	// (skip HTTP duration, record SSE metrics) from a bounded error response
-	// on the same route (bad auth, invalid since, non-GET) that must still be
-	// recorded into the HTTP/tenant duration histograms. See the route guard
-	// in observe.
+	// loop. observe uses it to distinguish a real SSE connection (record
+	// establishment latency, not the full connection lifetime) from a bounded
+	// error response on the same route (bad auth, invalid since, non-GET)
+	// that is recorded as a normal HTTP request. See the route guard in
+	// observe.
 	sseStreamEstablished bool
+	// sseEstablishedAt is the moment handleEvents finished writing the 200
+	// streaming headers (time-to-first-byte). observe records
+	// (sseEstablishedAt - start) into the HTTP/tenant duration histograms so
+	// /v1/events establishment latency is measured on the same footing as
+	// bounded HTTP requests, without the select-loop hold-open time polluting
+	// the histogram. The full connection lifetime goes into the dedicated
+	// drive9_sse_connection_duration_seconds metric recorded by handleEvents.
+	sseEstablishedAt time.Time
 }
 
 func withMetrics(ctx context.Context, m *serverMetrics) context.Context {
@@ -87,21 +95,27 @@ func requestMetricStateFromContext(ctx context.Context) *requestMetricState {
 }
 
 // markSSEStreamEstablished signals that the SSE handler has written the
-// streaming response headers and entered its long-lived loop, so observe
-// should treat this as a real SSE connection lifetime (skip the HTTP
-// duration histogram) rather than a bounded error response. Safe to call
-// from the SSE handler after WriteHeader(200).
+// streaming response headers and entered its long-lived loop. It records
+// the establishment timestamp so observe can measure the time-to-first-byte
+// (request → 200 headers + flush) — a bounded, meaningful latency that
+// belongs in the HTTP duration histogram — rather than the full connection
+// lifetime (select-loop hold-open), which is recorded into the dedicated
+// drive9_sse_connection_duration_seconds metric. Safe to call from the SSE
+// handler after WriteHeader(200).
 func markSSEStreamEstablished(ctx context.Context) {
 	if st := requestMetricStateFromContext(ctx); st != nil {
+		now := time.Now()
 		st.mu.Lock()
 		st.sseStreamEstablished = true
+		st.sseEstablishedAt = now
 		st.mu.Unlock()
 	}
 }
 
 // sseStreamEstablished reports whether the SSE handler marked the response
 // as an established streaming connection. observe uses this to decide whether
-// to skip the HTTP duration histogram.
+// to record the establishment duration (instead of the full connection
+// lifetime) into the HTTP/tenant duration histograms.
 func sseStreamEstablished(ctx context.Context) bool {
 	if st := requestMetricStateFromContext(ctx); st != nil {
 		st.mu.Lock()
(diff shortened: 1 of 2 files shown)


qiffang (approved):
## Review: GREEN

**Problem**: `/v1/events` SSE establishment latency was invisible after PR #633 (which correctly excluded connection lifetime from HTTP duration histogram, but also excluded establishment latency). Bounded errors on `/v1/events` with TiDB cold-start >10s hit the `+Inf` bucket, causing `histogram_quantile(0.99)` to extrapolate into 2000s+ — the exact false alert observed.

**Fix correctness verified**:

1. **`sseEstablishedAt` timestamp** — `markSSEStreamEstablished` now records `time.Now()` at the establishment point (after `WriteHeader(200)` + flush). This is the canonical time-to-first-byte.

2. **`sseEstablishmentDuration`** — returns `(sseEstablishedAt - start)` for established streams, `false` for non-established. Clean separation. The `!st.sseEstablishedAt.IsZero()` guard prevents returning a zero duration if the flag was set but timestamp wasn't.

3. **`observe` branch** — established SSE branch now calls `s.metrics.record()` + `recordTenantHTTPRequest()` with `estDur` (bounded establishment latency) instead of `recordCount` (counter only). This puts `/v1/events` back into the HTTP p99 histogram on a bounded, meaningful basis.

4. **Bounded errors still correct** — errors return before `markSSEStreamEstablished`, so they fall through to the normal `else` branch and record full request duration. No change needed.

5. **Removed unused helpers** — `recordCount`, `recordTenantHTTPRequestCount`, `recordTenantHTTPResponseBytes` are no longer needed. Clean deletion, no callers remain.

6. **Test** — `TestSSEEstablishmentDuration` verifies the invariant: establishment duration exists after mark, is bounded (≥1ms from sleep), and doesn't exceed connection lifetime.

**Edge cases checked**:
- `markSSEStreamEstablished` called without context state → nil check passes, no-op
- `sseEstablishmentDuration` before establishment → returns `false`
- Race between `markSSEStreamEstablished` and `observe` → `st.mu` serializes access
- CI: `ci` ✅, `drive9-server-local e2e smoke` ✅

No issues found.

qiffang (approved):
**GREEN** — fix is correct and well-scoped.

## Verification

- `markSSEStreamEstablished` (instrumentation.go:105-113) captures `time.Now()` atomically with the boolean flag under `st.mu`. No TOCTOU.
- `sseEstablishmentDuration` (instrumentation.go:131-140) computes `sseEstablishedAt.Sub(start)` — bounded and correct.
- `observe` (instrumentation.go:751-756) gates on both `route == sseEventsRoute` AND `sseStreamEstablished(ctx)`, so bounded error responses (which return before `markSSEStreamEstablished` at sse.go:355) fall through to the normal HTTP path. Correct.
- Fallback: if `sseEstablishmentDuration` returns `false` (should be impossible since both fields are set atomically), `dur` keeps `time.Since(start)` — safe degradation.
- Removed functions (`recordCount`, `recordTenantHTTPRequestCount`, `recordTenantHTTPResponseBytes`) have zero remaining callers. Verified.
- `TestSSEEstablishmentDuration` covers the core invariant.

## Warnings (non-blocking)

- **W1**: sse.go:351-354 comment is stale — still says "skip HTTP duration histogram". Should reflect the new behavior (record establishment latency *into* the histogram).
- **W2**: instrumentation.go:741 comment references `sseStreamEstablishedAt` but the field is `sseEstablishedAt`.
- **W3**: `RecordHTTPRequestCount` and `RecordTenantRequestCount` in `pkg/metrics/operations.go` are now dead code (no callers). Consider removing in a follow-up.

PR #660, opened 2026-07-01
feat: unified tenant outbox — eliminate all periodic tenant TiDB scans for serverless scale-to-zero

mornyx:
## Summary

Implements the unified tenant outbox design from issue #658, replacing ALL periodic tenant-TiDB scanning/polling with a unified outbox-driven model. This enables serverless scale-to-zero: idle tenant TiDBs can sleep because no periodic goroutine ever touches them.

### Unified outbox table (meta DB, always-provisioned)
```sql
CREATE TABLE tenant_notify_outbox (
    id BIGINT UNSIGNED AUTO_INCREMENT PRIMARY KEY,
    tenant_id VARCHAR(64) NOT NULL,
    work_mask TINYINT UNSIGNED NOT NULL DEFAULT 0,
    created_at DATETIME(3) NOT NULL DEFAULT CURRENT_TIMESTAMP(3),
    INDEX idx_tenant_notify_created (created_at)
)
CREATE TABLE tenant_outbox_cursor (
    pod_id VARCHAR(64) PRIMARY KEY,
    last_id BIGINT UNSIGNED NOT NULL DEFAULT 0,
    updated_at DATETIME(3) NOT NULL DEFAULT CURRENT_TIMESTAMP(3) ON UPDATE CURRENT_TIMESTAMP(3)
)
```

Work mask: SSE (bit 0) | Semantic (bit 1) | FileGC (bit 2) | Quota (bit 3).

### Architecture
1. **One per-pod poller** reads `tenant_notify_outbox` every 200ms, dispatches by work_mask:
   - SSE bit → wake local EventBus (broadcast: all pods with subscribers)
   - Semantic/GC/Quota bits → kick unified worker if shard owns tenant
2. **Unified worker** processes kicks: opens tenant TiDB, drains semantic + file_gc + quota tasks, recovers expired leases, does piggyback maintenance (fs_events cleanup + metrics, 5min throttle)
3. **Shard resolver**: jump consistent hash over active pod ring (5s refresh)
4. **Safety-net scan**: leader-gated, 5min, warm-only (AcquireCached), recovers expired leases for tenants whose kick may have been lost
5. Per-backend goroutines (FileGCWorker ticker, quotaOutboxWorker ticker) removed — processing is now kick-driven
6. Independent scans (cleanupFSEvents, collectObservation) removed — maintenance is piggybacked on kick

### Write path
Post-commit after write/upload/delete: best-effort `INSERT INTO tenant_notify_outbox` + in-process kick (~0ms latency).

[...]

PR #735, opened 2026-07-13
feat(pool): add idle eviction to tenant pool

mornyx:
## Summary

The tenant Pool is an LRU cache of per-tenant backends (`*backend.Dat9Backend` + `*datastore.Store` holding a `*sql.DB`). Until now, the only eviction path was **capacity-based** (`MaxTenants=1024`). A backend that entered the cache stayed warm forever — even with zero activity.

This causes a problem with the safety-net scan (`safetyNetScan`, PR #660). The scan runs every 5 minutes on every pod and uses `pool.AcquireCached` (warm-only) to recover expired leases and discover unclaimed queued tasks. Because warm tenants are never transitioned back to cold, `AcquireCached` keeps hitting them every 5 minutes, executing 4 SQL queries against each warm tenant's user DB. This creates a periodic connection pulse that prevents idle tenant TiDBs from reaching zero connections and scaling to zero.

PR #660 correctly established the invariant *"no periodic goroutine ever touches idle tenant TiDBs"* via the warm-only `AcquireCached`. But the missing piece was: **nothing ever moves a warm tenant back to cold.** This PR fills that gap.

## Approach

Add a background **idle reaper** to the Pool itself. On each tick (default every 5 minutes), it scans cached entries and evicts any with `refs == 0` and `lastUsed` older than `IdleTimeout` (default 10 minutes). Evicted backends are closed (backend workers stopped + `*sql.DB` closed), making the tenant cold. Subsequent `AcquireCached` calls return `cold_skipped`, leaving the tenant TiDB untouched until the next activity cold-opens it.

### Activity contract

`lastUsed` is refreshed by **any access through `Get`, `Acquire`, or `S3Backend`** — including:

- **Foreground user requests** (HTTP/FUSE/CLI via `auth.go`)
- **Tenant-specific durable work driven by kicks** (semantic/file_gc task processing via `tenant_worker.go`). Kicks are produced by writes — a user write enqueues a task and fires a kick, so the worker only touches tenants with recent write activity.
- **Object-GC candidate processing** (`object_gc_worker.go`). The worker is candidate-driven (not a full scan); it opens only the owner of a due GC candidate. A candidate is either deleted or postponed (24h for reachable refs, 1h for retries), so it does not continuously refresh `lastUsed`.

`lastUsed` is also refreshed on `Acquire` **release** (when `refs` drops to 0), so a long-running request is not immediately evictable after it finishes.

### What does NOT refresh `lastUsed`

**`AcquireCached` (safety-net scan)** does NOT refresh `lastUsed` on acquire or release. This is the critical distinction: warm-only periodic observation cannot keep an idle tenant warm forever. Once all foreground/durable activity stops, `lastUsed` goes stale and the idle reaper evicts the entry. Subsequent `AcquireCached` calls return `cold_skipped`, leaving the tenant TiDB untouched.

| Path | Used by | Refreshes `lastUsed`? |
|------|---------|----------------------|
| `Get` / `Acquire` | HTTP requests, kick-driven worker, object-GC | ✅ on acquire and release |
| `S3Backend` | local `/s3/` downloads | ✅ |
| `AcquireCached` | safety-net scan (periodic, warm-only) | ❌ never |

### `pkg/tenant/pool.go`

- **`PoolConfig`**: new `IdleTimeout` and `IdleReapInterval` fields (`time.Duration`, 0 = disabled)
- **`entry` struct**: new `lastUsed time.Time` field, refreshed on `Get`/`Acquire`/`S3Backend` cache hits and on `Acquire` release; never by `AcquireCached`
- **`Pool` struct**: new `idleTimeout`, `reapInterval`, `reapStop`, `reapWG` fields for reaper lifecycle
- **`NewPool`**: reads `IdleTimeout`/`IdleReapInterval` from config; defaults `reapInterval` to 5min when `IdleTimeout > 0`
- **`Start(ctx)`**: launches the reaper goroutine if `IdleTimeout > 0`; no-op otherwise
- **`reapLoop`**: ticker-driven loop calling `reapOnce` on each interval; guards against non-positive interval
- **`reapOnce`**: iterates `p.items` under `p.mu`; for entries with `refs == 0 && !retired && time.Since(lastUsed) > idleTimeout`, calls `removeLocked(e.elem, "idle")` and collects for `closeEntry` outside the lock; no-op when `idleTimeout <= 0`
- **`releaseEntry`**: now takes `refreshLastUsed bool` — `makeRelease` (Acquire) passes true; `makeReleaseNoRefresh` (AcquireCached) passes false. When refs drop to 0 and not retired, `lastUsed` is refreshed so the idle timer starts from "activity finished"
- **`S3Backend`**: refreshes `lastUsed` on cache hit (local `/s3/` download path)
- **`Close`**: stops the reaper (`reapStop` + `reapWG.Wait()`) before draining entries, preventing shutdown races
- Existing `removeLocked` reason metric now records `"idle"` automatically; `reapOnce` refreshes the `cached_backends` gauge

### `cmd/drive9-server/main.go`

- `DRIVE9_POOL_IDLE_TTL` env var (default `10m`, `0` = disabled)
- `DRIVE9_POOL_IDLE_REAP_INTERVAL` env var (default `5m`)
- `pool.Start(context.Background())` after `SetMetaStore`
- Help text for both env vars

### `pkg/tenant/pool_test.go`

9 tests:

1. `TestIdleEviction` — backend is evicted and store closed after idle timeout lapses
2. `TestIdleEvictionSkippedByRecentAcquire` — recent `Acquire` refreshes `lastUsed`, no eviction
3. `TestIdleEvictionNotRefreshedByAcquireCached` — `AcquireCached` does **not** refresh `lastUsed`; entry is evicted despite a recent cached touch
4. `TestIdleEvictionRespectsRefs` — pinned entry (`refs > 0`) is not evicted; after release + timeout it is evicted
5. `TestIdleEvictionDisabled` — `IdleTimeout=0` → `reapOnce` is a no-op, entry stays
6. `TestIdleEvictionStartNoOpWhenDisabled` — `Start` does not launch a goroutine when `IdleTimeout=0`
7. `TestIdleEvictionReleaseRefreshesLastUsed` — long-running request held beyond TTL is not immediately evicted on release; `lastUsed` is refreshed
8. `TestIdleEvictionAcquireCachedReleaseDoesNotRefresh` — `AcquireCached` release does not refresh `lastUsed`
9. `TestIdleEvictionStartAutoEvictsAndCloseStops` — end-to-end: `Start` → ticker auto-evicts → `Close` lifecycle

## Safety guarantees

1. **Pinned entries are safe**: `reapOnce` skips `refs > 0`; `removeLocked` returns nil for pinned entries (deferred close via `releaseEntry` when refs drop to 0) — reuses the existing mechanism
2. **Close happens outside the lock**: `reapOnce` collects `toClose` under `p.mu`, then `closeEntry` after `Unlock`
3. **Shutdown is clean**: `Close()` stops the reaper and waits for it before draining
4. **Backward compatible**: `DRIVE9_POOL_IDLE_TTL=0` fully disables the feature, restoring current behavior
5. **No changes to safetyNetScan or other callers**: the warm-only `AcquireCached` contract is preserved; idle eviction just ensures warm tenants eventually become cold

## Metric impact

- Existing `drive9_service_operations_total{component="tenant_pool",operation="remove",result="idle"}` records each eviction
- `drive9_tenant_pool_cached_backends` gauge is refreshed after each reap cycle
- No new metric definitions needed

## Defaults

| Env var | Default | Rationale |
|---------|---------|-----------|
| `DRIVE9_POOL_IDLE_TTL` | 10m | 2× the safety-net scan interval (5m); only truly idle tenants (no activity for 10 min) are evicted |
| `DRIVE9_POOL_IDLE_REAP_INTERVAL` | 5m | Guarantees eviction within at most one interval after the timeout lapses |

### DIFF
--- a/cmd/drive9-server/main.go
+++ b/cmd/drive9-server/main.go
@@ -318,12 +318,15 @@ func main() {
 			S3EncryptionPolicy:           s3cfg.EncryptionPolicy,
 			BackendOptions:               backendOptions,
 			MaxTenants:                   envInt("DRIVE9_POOL_MAX_TENANTS", 0),
+			IdleTimeout:                  envDuration("DRIVE9_POOL_IDLE_TTL", 10*time.Minute),
+			IdleReapInterval:             envDuration("DRIVE9_POOL_IDLE_REAP_INTERVAL", 0),
 			DisableDatabaseAutoEmbedding: disableDatabaseAutoEmbedding,
 			LeaderChecker:                leaderManager,
 		}, enc)
 		defer pool.Close()
 
 		pool.SetMetaStore(store)
+		pool.Start(context.Background())
 
 		// The mutation replay and expiry sweep workers are owned by the server
 		// (started/stopped in its leader-gated startLeaderWorkers/stopLeaderWorkers),
@@ -526,6 +529,8 @@ environment:
   DRIVE9_PUBLIC_URL  externally reachable base URL for presigned URLs (required for remote clients)
   DRIVE9_META_DSN    control-plane MySQL DSN (required)
   DRIVE9_POOL_MAX_TENANTS max cached tenant user DB pools per pod (default: 1024)
+  DRIVE9_POOL_IDLE_TTL  idle duration before a cached tenant backend is evicted (default: 10m, 0=disabled)
+  DRIVE9_POOL_IDLE_REAP_INTERVAL  how often the idle reaper scans (default: 5m)
   DRIVE9_META_DB_MAX_OPEN_CONNS max open connections for the per-pod meta DB pool (default: 100)
   DRIVE9_META_DB_MAX_IDLE_CONNS max idle connections for the per-pod meta DB pool (default: 20)
   DRIVE9_USER_DB_MAX_OPEN_CONNS max open connections for each cached tenant user DB pool (default: 6)
--- a/pkg/tenant/pool.go
+++ b/pkg/tenant/pool.go
@@ -64,6 +64,18 @@ type PoolConfig struct {
 	// resolves to the post-transition state instead of a moment-in-time
 	// IsLeader() snapshot.
 	LeaderChecker LeaderChecker
+
+	// IdleTimeout controls how long a cached backend with no user traffic
+	// (Get/Acquire) can stay in the warm cache before the idle reaper evicts
+	// it. 0 disables idle eviction (LRU capacity eviction still applies).
+	// AcquireCached (safety-net scan etc.) does NOT refresh the idle timer —
+	// only real user traffic does, so periodic internal scans cannot keep a
+	// tenant warm forever.
+	IdleTimeout time.Duration
+
+	// IdleReapInterval is how often the idle reaper scans for idle backends.
+	// Defaults to defaultTenantPoolIdleReapInterval when IdleTimeout > 0.
+	IdleReapInterval time.Duration
 }
 
 // LeaderChecker reports whether the current process is the leader. Used by
@@ -123,6 +135,10 @@ type Pool struct {
 	// FileGC is now kick-driven through the unified tenant worker, not a
 	// per-backend goroutine.
 	fileGCEnabled bool
+	idleTimeout   time.Duration
+	reapInterval  time.Duration
+	reapStop      context.CancelFunc
+	reapWG        sync.WaitGroup
 }
 
 type tenantAutoEmbeddingProfile struct {
@@ -144,6 +160,7 @@ var (
 	periodicTiDBSchemaValidationEvery uint64 = 32
 	defaultTenantPoolDrainTimeout            = 30 * time.Second
 	defaultTenantPoolMaxTenants              = 1024
+	defaultTenantPoolIdleReapInterval        = 5 * time.Minute
 )
 
 func tenantSchemaVersionForEmbeddingMode(mode string, profile schema.TiDBAutoEmbeddingProfile) (int, error) {
@@ -197,6 +214,7 @@ type entry struct {
 	elem               *list.Element
 	refs               int
 	retired            bool
+	lastUsed           time.Time // updated only on Get/Acquire (user traffic), not AcquireCached
 }
 
 func NewPool(cfg PoolConfig, enc encrypt.Encryptor) *Pool {
@@ -204,14 +222,21 @@ func NewPool(cfg PoolConfig, enc encrypt.Encryptor) *Pool {
 	if max <= 0 {
 		max = defaultTenantPoolMaxTenants
 	}
+	idleTimeout := cfg.IdleTimeout
+	reapInterval := cfg.IdleReapInterval
+	if reapInterval <= 0 && idleTimeout > 0 {
+		reapInterval = defaultTenantPoolIdleReapInterval
+	}
 	return &Pool{
-		cfg:     cfg,
-		enc:     enc,
-		items:   map[string]*entry{},
-		order:   list.New(),
-		maxSize: max,
+		cfg:          cfg,
+		enc:          enc,
+		items:        map[string]*entry{},
+		order:        list.New(),
+		maxSize:      max,
 		// No LeaderChecker means single-pod mode: FileGC runs unconditionally.
 		fileGCEnabled: cfg.LeaderChecker == nil,
+		idleTimeout:   idleTimeout,
+		reapInterval:  reapInterval,
 	}
 }
 
@@ -264,6 +289,7 @@ func (p *Pool) Get(ctx context.Context, t *meta.Tenant) (out *backend.Dat9Backen
 	if e, ok := p.items[t.ID]; ok {
 		if e.s3EncryptionPolicy == s3EncryptionPolicy && (storageNamespaceID == "" || e.storageNamespaceID == storageNamespaceID) {
 			p.order.MoveToFront(e.elem)
+			e.lastUsed = time.Now()
 			b := e.backend
 			p.mu.Unlock()
 			return b, nil
@@ -291,6 +317,7 @@ func (p *Pool) Get(ctx context.Context, t *meta.Tenant) (out *backend.Dat9Backen
 			b.Close()
 			_ = st.Close()
 			p.order.MoveToFront(e.elem)
+			e.lastUsed = time.Now()
 			p.mu.Unlock()
 			return e.backend, nil
 		}
@@ -298,7 +325,7 @@ func (p *Pool) Get(ctx context.Context, t *meta.Tenant) (out *backend.Dat9Backen
 			toClose = append(toClose, removed)
 		}
 	}
-	e := &entry{tenantID: t.ID, storageNamespaceID: t.StorageNamespaceID, s3EncryptionPolicy: s3EncryptionPolicy, backend: b, store: st}
+	e := &entry{tenantID: t.ID, storageNamespaceID: t.StorageNamespaceID, s3EncryptionPolicy: s3EncryptionPolicy, backend: b, store: st, lastUsed: time.Now()}
 	e.elem = p.order.PushFront(e)
 	p.items[t.ID] = e
 	for p.order.Len() > p.maxSize {
@@ -341,6 +368,7 @@ func (p *Pool) Acquire(ctx context.Context, t *meta.Tenant) (out *backend.Dat9Ba
 		if e.s3EncryptionPolicy == s3EncryptionPolicy && (storageNamespaceID == "" || e.storageNamespaceID == storageNamespaceID) {
 			e.refs++
 			p.order.MoveToFront(e.elem)
+			e.lastUsed = time.Now()
 			p.mu.Unlock()
 			metrics.RecordOperation("tenant_pool", "cache_lookup", "hit", 0)
 			logger.InfoBenchTiming(ctx, "tenant_pool_acquire_timing",
@@ -381,6 +409,7 @@ func (p *Pool) Acquire(ctx context.Context, t *meta.Tenant) (out *backend.Dat9Ba
 		if e.s3EncryptionPolicy == s3EncryptionPolicy && (t.StorageNamespaceID == "" || e.storageNamespaceID == t.StorageNamespaceID) {
 			e.refs++
 			p.order.MoveToFront(e.elem)
+			e.lastUsed = time.Now()
 			p.mu.Unlock()
 			b.Close()
 			_ = st.Close()
@@ -396,7 +425,7 @@ func (p *Pool) Acquire(ctx context.Context, t *meta.Tenant) (out *backend.Dat9Ba
 			toClose = append(toClose, removed)
 		}
 	}
-	e := &entry{tenantID: t.ID, storageNamespaceID: t.StorageNamespaceID, s3EncryptionPolicy: s3EncryptionPolicy, backend: b, store: st, refs: 1}
+	e := &entry{tenantID: t.ID, storageNamespaceID: t.StorageNamespaceID, s3EncryptionPolicy: s3EncryptionPolicy, backend: b, store: st, refs: 1, lastUsed: time.Now()}
 	e.elem = p.order.PushFront(e)
 	p.items[t.ID] = e
 	for p.order.Len() > p.maxSize {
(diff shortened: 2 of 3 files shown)


qiffang (commented):
## adversary-2 review — GREEN with 2 warnings

**Checklist — all pass:**

- `lastUsed` updated only on `Get`/`Acquire`, NOT on `AcquireCached` — core invariant holds ✅
- `reapOnce` skips `retired` and `refs > 0` entries — no premature eviction ✅
- `removeLocked` returns nil for pinned entries — safe even in race ✅
- `closeEntry` runs outside `p.mu` — no deadlock ✅
- `Close()` stops reaper via `reapStop()` + `reapWG.Wait()` before draining — correct shutdown ordering ✅
- Map deletion during iteration — Go spec allows ✅
- `IdleTimeout=0` disables feature — `Start()` is no-op, `reapOnce` returns early ✅
- `IdleReapInterval=0` with `IdleTimeout>0` → defaults to 5m ✅
- 6 tests cover: basic eviction, refresh by Acquire, no-refresh by AcquireCached, refs pinning, disabled mode, Start no-op ✅

**W1: `lastUsed` set at Acquire time, not Release time.**

A user request that holds a backend for longer than `IdleTimeout` will be evicted immediately after release — `now - lastUsed > IdleTimeout` is true the moment `refs` drops to 0. With 10m default this is unlikely in practice (dat9-dev1 noted the same). But if someone configures `IdleTimeout=1m` for aggressive eviction, a slow query holding a backend for 2m would trigger immediate eviction on release. Consider documenting this in the `IdleTimeout` field comment.

**W2: Tenant worker / object GC use `Acquire`, not `AcquireCached`.**

The design discussion (dat9-dev2) proposed that background paths should not refresh the idle timer. Safety-net correctly uses `AcquireCached` (no refresh). But kick-driven tenant worker and object GC go through `Acquire` → they DO refresh `lastUsed`. If a tenant has pending GC work but no user traffic, GC kicks will keep it warm indefinitely. This may be intentional (GC needs active DB connections anyway), but it deviates from the "only user traffic keeps tenants warm" principle. Worth tracking as follow-up if GC-only tenants are observed staying warm.

Neither warning is blocking. Implementation is clean, tests cover the critical invariants, and the core design is correct.

qiffang (changes requested):
REVISE on `71c706e`.

Blocking finding: the PR does not implement the user-vs-background activity split we agreed on in this thread. It only prevents `AcquireCached` from refreshing `lastUsed`; all `Acquire` callers still refresh the idle clock and can cold-open tenant DBs.

Evidence:
- `pkg/tenant/pool.go:368-372`, `409-412`, and `428` refresh `entry.lastUsed` on every `Acquire` cache hit/race hit/new insert.
- `pkg/server/object_gc_worker.go:150-163` is a periodic background worker path and still calls `w.pool.Acquire(ctx, owner)`. That path can open a cold tenant DB for a due GC candidate and marks the backend as recently used.
- `pkg/server/tenant_worker.go:630-647` also calls `m.pool.Acquire(ctx, ref.tenant)` from kick-driven background processing. That may be acceptable if kicks are treated as user-derived work, but the policy is not encoded in the pool API.

Why this matters: qiffang’s target was “tenant has no active file/API request, background should not keep touching tenant DB.” The thread explicitly called out that background pins need a different semantic from user pins, and that object GC needs an explicit policy. With this implementation, an idle tenant with a due GC candidate can be opened by object GC, become warm again, have `lastUsed` refreshed by background work, and then be eligible for `AcquireCached` safety-net scans until the idle reaper later evicts it. That contradicts the PR body claim that tenants stay untouched until the next real user request.

Required fix: split the pool API or pass an explicit access kind, e.g. `AcquireUser` updates user activity, while background paths either use warm-only/active-TTL acquisition or at least do not refresh user activity. Then define object-GC behavior explicitly: either it is allowed to wake idle tenant DBs as a documented exception, or it must skip/defer idle tenants. Add regression coverage proving a background object-GC/safety-net-style path does not refresh user activity or keep an idle tenant warm.

The new reaper mechanics themselves look mostly sound: it skips `refs > 0`, closes outside the mutex, and `Close()` stops the reaper before draining. The current tests cover `AcquireCached` not refreshing the clock, but they miss the background `Acquire` callers above.

Local validation:
- `go vet ./pkg/tenant ./pkg/server ./cmd/drive9-server` passed.
- `go test -c ./pkg/tenant ./pkg/server ./cmd/drive9-server ./pkg/meta ./pkg/backend` passed.
- Runtime `go test ./pkg/tenant -run 'TestIdleEviction' -count=1` is blocked on this machine by `panic: rootless Docker not found`; GitHub CI for the PR is green.

qiffang (changes requested):
## dat9-dev2 review — REVISE

Reviewed head `71c706e` against the behavior agreed in the discussion: cached capacity/LRU state must be separate from **real user activity**, background work must not silently renew the user-idle clock, and an in-flight user pin must remain safe.

### B1: `Acquire` still conflates user and background activity

`pkg/tenant/pool.go:368-372`, `409-412`, and `428` refresh `lastUsed` for every `Acquire`. But `Acquire` is not user-only:

- `pkg/server/object_gc_worker.go:150-163` is a minute-driven background path. A due GC candidate can cold-open an idle tenant DB and mark it recently used.
- `pkg/server/tenant_worker.go:630-647` is kick-driven background processing and also refreshes `lastUsed`.
- Only `pkg/server/safety_net_scan.go:102-107` uses `AcquireCached` and avoids refreshing the clock.

Therefore the comments at `pkg/tenant/pool.go:68-73` and the PR claim that only real user traffic keeps a tenant warm are false. An object-GC-only tenant can be reopened by background work, become warm, and then be touched by later safety-net scans despite having no active file/API request.

Required behavior: encode access semantics explicitly. User request acquisition updates user activity; background acquisition/release does not. Decide object-GC policy explicitly: either it may wake an idle tenant as a documented exception but must not renew user activity, or it must skip/defer cold/idle tenants. Do not infer user activity from the generic `Acquire` method.

### B2: user activity is recorded at acquire, not at release

`releaseEntry` (`pkg/tenant/pool.go:1183-1196`) only decrements `refs`; it never updates the activity timestamp. A real user request held longer than `IdleTimeout` is protected while `refs > 0`, but immediately becomes eligible for eviction when it releases, even though it was active until that moment. This is especially visible with an aggressive TTL or a long upload/query/worker operation.

The discussion explicitly defined user acquire **and user release** as activity. Once user/background pins are distinguished, a user release must refresh `lastUserActivity`; a background release must not.

### B3: tests prove only `AcquireCached`, not the required policy

`TestIdleEvictionNotRefreshedByAcquireCached` is useful, but no regression covers a background `Acquire` caller, object-GC behavior, or a long-running user pin released after the TTL. Those are the exact failure paths above. The new `Start` ticker/`Close` lifecycle also has no positive end-to-end test; all active eviction tests call `reapOnce` manually.

### What is sound

The reaper mechanics are otherwise reasonable: eviction is under `p.mu`, `refs > 0` is respected, close runs outside the mutex, and `Close` cancels/waits before draining.

Validation on `71c706e`:

- `go vet ./pkg/tenant ./pkg/server ./cmd/drive9-server` passed
- compile-only tests passed for `./pkg/tenant`, `./pkg/server`, and `./cmd/drive9-server`
- GitHub CI, local-e2e, and CodeRabbit are green

Verdict remains **REVISE** until the user/background activity contract and its regressions are explicit.

qiffang (changes requested):
## adversary-2 revised verdict — REVISE (upgrading W2 to blocker)

Withdrawing my earlier GREEN. The design discussion in the Slock thread explicitly agreed on user-vs-background activity separation, and the PR doesn't implement it.

### Blocker: `Acquire` conflates user and background access

`Acquire` refreshes `lastUsed` unconditionally. It is called by:
- User requests (HTTP/FUSE/CLI via `auth.go`) — should refresh ✅
- Object GC worker (`object_gc_worker.go:150-163`) — should NOT refresh ❌
- Kick-driven tenant worker (`tenant_worker.go:630-647`) — policy unclear ❌

Object GC is periodic background work. It can cold-open an idle tenant DB AND mark it as recently used, which then makes it eligible for safety-net scans. This contradicts the core invariant: "only user traffic keeps tenants warm."

### Blocker: `lastUsed` not updated on Release

Per the design discussion, user `Release` should also update `lastUserActivity`. Without this, a user request held for longer than `IdleTimeout` is evictable immediately on release — `now - lastUsed > IdleTimeout` is true the moment `refs` drops to 0.

### Required

1. Split acquisition API: `AcquireUser` (refreshes user activity on acquire+release) vs background acquire (pins only, no activity refresh)
2. Define object GC policy explicitly: allowed exception to cold-open, or warm-only/defer when idle
3. Regression tests for background acquire not refreshing user activity, and long user request not immediately evicted on release

Reaper mechanics remain sound — the issue is activity semantics, not lock/eviction logic.

qiffang (changes requested):
## dat9-dev2 R2 — B1 code blocker withdrawn; documentation still needs correction

I re-reviewed latest head `4e4647af582235fe63425c25569d3c770de75ea8` and traced the actual kick/candidate flows.

### Revised conclusion on the original B1

I accept the clarified policy: **known durable tenant work may count as activity**, even when the work is executed by a background worker. Under that policy, splitting `AcquireUser` / `AcquireBackground` is not required for this PR.

- Multi-tenant `tenantWorker` does not periodically scan/open every tenant. It calls `Acquire` only after a tenant-specific kick and then drains concrete semantic/file-GC/quota work.
- `objectGCWorker` is candidate-driven. It does not iterate warm tenants; it opens only the owner of a due GC candidate. Waking a cold tenant to prove reachability/delete an obsolete object is a valid GC exception.
- `AcquireCached` still does not refresh on acquire or release, so a plain safety-net touch cannot by itself keep an idle tenant warm indefinitely.

Therefore the original requirement to split the pool API is withdrawn.

### Required correction: current docs/comments still describe a different policy

The current implementation and accepted policy are broader than “only genuine/real user traffic keeps tenants warm,” but that claim still appears in:

- PR body: `only genuine user traffic keeps a tenant warm`
- `PoolConfig.IdleTimeout`: `only real user traffic does`
- `entry.lastUsed`: `Get/Acquire (user traffic)`

Those statements are false:

1. `tenantWorker` kicks are not exclusively direct write-path traffic. `safetyNetScan` can kick a warm tenant after recovering/observing durable queued work, and the worker can self-rekick after hitting a drain cap. These are still valid work-derived kicks, but not current user requests.
2. GC candidates are not always consumed once. Reachable refs are postponed for 24h, and failures retry after 1h with no attempt cap in the current code. Each due attempt can cold-open the tenant and refresh `lastUsed` on release. That is an explicit background exception, not “real user traffic.”

Please update the PR body and code comments to state the actual contract, e.g.:

- `lastUsed` is refreshed by foreground access and tenant-specific durable work executed through `Get`/`Acquire`.
- warm-only periodic observation through `AcquireCached` never refreshes it.
- object GC may cold-open a tenant for a due candidate and intentionally keeps it warm for the configured idle TTL afterward; reachable/failed candidates may revisit on their configured retry cadence.

This is a documentation/spec blocker only: the code must not merge with an invariant that its accepted behavior violates.

### Other prior blockers

- Long user pin release now refreshes `lastUsed`: fixed with regression coverage.
- `AcquireCached` release no longer refreshes: fixed with regression coverage.
- ticker-driven `Start`/`Close` lifecycle coverage: added.
- CodeRabbit's final unlocked `pool.items` assertion is worth cleaning up for consistent locking, but I do not consider it a demonstrated runtime race after the test has already observed stable eviction.

Validation on `4e4647a`:

- `go vet ./pkg/tenant ./pkg/server ./cmd/drive9-server` passed
- compile-only tests passed for `pkg/tenant`, `pkg/server`, and `cmd/drive9-server`
- focused runtime/race test is still locally blocked by testcontainers (`rootless Docker not found`); GitHub CI and local-e2e are green

**Verdict: REVISE only for the policy documentation/comments above. No user/background API split is required after the clarified product decision.**

qiffang (changes requested):
REVISE on latest head `4e4647a`.

I am withdrawing the original code-level requirement to split `AcquireUser` / background acquire after qiffang's clarified product decision: tenant-specific durable work may count as activity. Under that policy, the core code path is acceptable: safety-net uses `AcquireCached` and does not refresh on acquire or release; normal `Acquire` refreshes on acquire and release; long held refs are protected; ticker/Close coverage was added.

Remaining blocker is that the PR still documents the wrong invariant.

Evidence:
- PR body still says only "genuine user traffic" keeps a tenant warm and says `Acquire` is used by "real user traffic paths".
- `pkg/tenant/pool.go:68-73` says idle means "no user traffic (Get/Acquire)" and "only real user traffic" refreshes the timer.
- `pkg/tenant/pool.go:217` says `lastUsed` is updated only on `Get`/`Acquire (user traffic)`, but accepted behavior is broader: tenant-worker durable work and object-GC candidate processing also use `Acquire`, and `S3Backend` now refreshes `lastUsed` too.

That documentation is materially false for the accepted design. The actual contract should be explicit before merge: `lastUsed` is refreshed by foreground access and tenant-specific durable/background work that intentionally opens or pins the tenant via `Get`/`Acquire`/local object access; warm-only periodic observation via `AcquireCached` must never refresh it; object GC may cold-open a tenant for a due candidate and can keep it warm for one idle-TTL window per due attempt/retry cadence.

Non-blocking cleanup: `TestIdleEvictionStartAutoEvictsAndCloseStops` should lock around the final `pool.items` read for consistency with the polling loop.

Validation:
- `go vet ./pkg/tenant ./pkg/server ./cmd/drive9-server` passed
- `go test -c ./pkg/tenant ./pkg/server ./cmd/drive9-server ./pkg/meta ./pkg/backend` passed
- focused runtime `go test ./pkg/tenant -run 'TestIdleEviction' -count=1` is still blocked locally by `panic: rootless Docker not found`; GitHub CI/local-e2e are green on this head

qiffang (changes requested):
## dat9-dev2 R3 — one stale policy claim remains

Head `1d2d66d` fixes the substantive items from R2:

- `PoolConfig.IdleTimeout` and `entry.lastUsed` now describe foreground access plus tenant-specific durable/background work as activity.
- `AcquireCached` is explicitly non-refreshing on acquire and release.
- Object-GC cold-open/revisit behavior is documented.
- The final lifecycle-test map observation is now synchronized via the locked polling result.

One documentation mismatch remains before GREEN:

- The PR body still says tenant-worker kicks are produced by writes and therefore the worker only touches tenants with **recent write activity**.
- `pkg/server/tenant_worker.go:645` likewise says `The rate follows write traffic (kicks are produced by writes)`.

That is not the complete current contract. `safetyNetScan` can call `tenantWorker.Kick` after recovering/observing durable queued work, and `processKicked` can self-rekick after a drain cap. Those paths are valid under the accepted policy because they represent concrete durable tenant work, but they are not evidence of a recent foreground write.

Please replace the remaining write-correlation claim with the actual invariant: the multi-tenant worker acquires only for a tenant-specific kick representing known durable work; it does not periodically scan/open all tenants. This is wording-only—no API or behavior change is requested.

Validation on `1d2d66d`:

- gofmt check passed
- `go vet ./pkg/tenant ./pkg/server ./cmd/drive9-server` passed
- compile-only tests passed for `pkg/tenant`, `pkg/server`, and `cmd/drive9-server`
- GitHub CI, local-e2e, and CodeRabbit are green

**Verdict: REVISE for this final stale policy statement only; implementation and tests are otherwise GREEN.**

qiffang (approved):
GREEN on latest head `1d2d66d7cbe4ff3c98c0ce499b8cc1f64d3364d8`.

My previous blocker is cleared. The code comments now state the actual activity contract instead of the earlier "only real user traffic" invariant: `Get` / `Acquire` / `S3Backend` refresh `lastUsed`, `AcquireCached` never refreshes, and object GC is explicitly treated as a candidate-driven cold-open exception. The test locking nit is also fixed.

I rechecked the key paths:
- `PoolConfig.IdleTimeout` comment now documents foreground access, kick-driven durable work, and object-GC candidate processing as activity.
- `entry.lastUsed` comment now includes `Get` / `Acquire` / `S3Backend` and excludes `AcquireCached`.
- `TestIdleEvictionStartAutoEvictsAndCloseStops` no longer reads `pool.items` unlocked after the polling loop.

Residual non-blocking wording nit: the PR body still says tenant-worker kicks are produced by writes and imply "recent write activity". Safety-net recovery and self-rekick mean that is slightly narrower than the real model; the important contract is already documented correctly enough in the body and code comments: foreground access + tenant-specific durable work refreshes; `AcquireCached` observation does not; object-GC due candidates may cold-open.

Validation:
- `go vet ./pkg/tenant ./pkg/server ./cmd/drive9-server` passed
- `go test -c ./pkg/tenant ./pkg/server ./cmd/drive9-server ./pkg/meta ./pkg/backend` passed
- focused runtime `go test ./pkg/tenant -run 'TestIdleEviction' -count=1` is still blocked locally by `panic: rootless Docker not found`; GitHub CI/local-e2e are green

PR #736, opened 2026-07-13
Persist tenant schema versions and tag schema logs

srstack:
## Summary
- Persist the profile-derived tenant `schema_version` after async schema initialization succeeds, so newly initialized TiDB tenants do not run a first-request physical diff when the stored version is already aligned.
- Propagate tenant ID into request cold-open schema ensure contexts and add `tenant_id` to `tenant_tidb_schema_*` init, validate, diff, repair, and preflight logs.
- Share tenant embedding-mode schema-version derivation between the server and pool paths, add review-requested comments/logging around schema init attribution and version persistence failures, and wrap schema-version derivation errors with context.
- Add regression coverage for async schema-version persistence, schema log field attribution, and shared schema-version derivation.

## Validation
- `go test ./pkg/tenant -run 'TestTiDBTenantSchemaVersionForEmbeddingMode|TestPoolCreateBackend|TestRecordTenantSchemaVersionUpdateFailureRecordsMetric' -count=1`
- `make test TEST_PKGS='./pkg/server' TEST_RUN='TestInitTenantSchemaAsyncPersistsTargetSchemaVersion|TestSchemaInitForTenantUsesFTSOnlyProfileWhenDatabaseAutoEmbeddingDisabled|TestAutoEmbeddingProfileForTenantEnsuresDefaultProfile'`
- `go test ./pkg/tenant/schema -run 'TestTenantIDContext|TestTenantSchemaLogFieldsIncludesTenantID' -count=1`
- `make test TEST_PKGS='./pkg/server' TEST_RUN='TestInitTenantSchemaAsyncPersistsTargetSchemaVersion'`
- `make lint`
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved tenant-specific schema initialization and persistence of embedding schema versions.
  * Ensured schema operations consistently use the correct tenant context.
  * Enhanced tenant schema logs with tenant identifiers for clearer diagnostics.
  * Improved handling of invalid embedding modes and schema version updates.

* **Tests**
  * Added coverage for tenant context propagation, schema initialization order, version persistence, and embedding-mode validation.

### DIFF
--- a/pkg/server/provision_test.go
+++ b/pkg/server/provision_test.go
@@ -347,13 +347,14 @@ func (f *credentialOnlyProvisioner) ProvisionWithCredentials(_ context.Context, 
 
 type profileAwareFakeProvisioner struct {
 	fakeProvisioner
-	mu               sync.Mutex
-	profileInitCalls atomic.Int32
-	ensureDBCalls    atomic.Int32
-	lastProfile      tenantschema.TiDBAutoEmbeddingProfile
-	ensureDBErr      error
-	lastEnsureDSN    string
-	callOrder        []string
+	mu                  sync.Mutex
+	profileInitCalls    atomic.Int32
+	ensureDBCalls       atomic.Int32
+	lastProfile         tenantschema.TiDBAutoEmbeddingProfile
+	lastProfileTenantID string
+	ensureDBErr         error
+	lastEnsureDSN       string
+	callOrder           []string
 }
 
 func (f *profileAwareFakeProvisioner) EnsureDatabase(_ context.Context, dsn string) error {
@@ -365,11 +366,12 @@ func (f *profileAwareFakeProvisioner) EnsureDatabase(_ context.Context, dsn stri
 	return f.ensureDBErr
 }
 
-func (f *profileAwareFakeProvisioner) InitSchemaForAutoEmbeddingProfile(_ context.Context, _ string, profile tenantschema.TiDBAutoEmbeddingProfile) error {
+func (f *profileAwareFakeProvisioner) InitSchemaForAutoEmbeddingProfile(ctx context.Context, _ string, profile tenantschema.TiDBAutoEmbeddingProfile) error {
 	f.profileInitCalls.Add(1)
 	f.mu.Lock()
 	defer f.mu.Unlock()
 	f.lastProfile = profile
+	f.lastProfileTenantID = tenantschema.TenantIDFromContext(ctx)
 	f.callOrder = append(f.callOrder, "profile-init")
 	return nil
 }
@@ -443,6 +445,12 @@ func TestSchemaInitForTenantEnsuresDatabaseBeforeProfileInit(t *testing.T) {
 	}
 	if got, want := prov.callOrderString(), "ensure,profile-init"; got != want {
 		t.Fatalf("call order = %s, want %s", got, want)
+	}
+	prov.mu.Lock()
+	lastProfileTenantID := prov.lastProfileTenantID
+	prov.mu.Unlock()
+	if lastProfileTenantID != "tenant-native" {
+		t.Fatalf("profile init tenant id = %q, want tenant-native", lastProfileTenantID)
 	}
 }
 
--- a/pkg/server/provision_test.go
+++ b/pkg/server/provision_test.go
@@ -1814,6 +1814,72 @@ func TestSchemaInitForTenantUsesFTSOnlyProfileWhenDatabaseAutoEmbeddingDisabled(
 	}
 }
 
+func TestInitTenantSchemaAsyncPersistsTargetSchemaVersion(t *testing.T) {
+	metaStore, err := meta.Open(testDSN)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer func() { _ = metaStore.Close() }()
+	testmysql.ResetMetaDB(t, metaStore.DB())
+
+	tenantID := token.NewID()
+	now := time.Now().UTC()
+	if err := metaStore.InsertTenant(context.Background(), &meta.Tenant{
+		ID:               tenantID,
+		Status:           meta.TenantProvisioning,
+		DBPasswordCipher: []byte{},
+		Provider:         tenant.ProviderTiDBZero,
+		SchemaVersion:    1,
+		CreatedAt:        now,
+		UpdatedAt:        now,
+	}); err != nil {
+		t.Fatal(err)
+	}
+
+	profile := tenantschema.TiDBAutoEmbeddingProfile{
+		Model:       "openai/text-embedding-v4",
+		Dimensions:  1024,
+		OptionsJSON: `{"dimensions":1024}`,
+	}
+	targetVersion, err := tenantschema.TiDBTenantSchemaVersionForEmbeddingModeProfile(tenantschema.TiDBEmbeddingModeFTSOnly, profile)
+	if err != nil {
+		t.Fatalf("target schema version: %v", err)
+	}
+	if err := metaStore.UpsertTenantAutoEmbeddingProfile(context.Background(), &meta.TenantAutoEmbeddingProfile{
+		TenantID:      tenantID,
+		EmbeddingMode: meta.TenantEmbeddingModeFTSOnly,
+		Model:         profile.Model,
+		Dimensions:    profile.Dimensions,
+		OptionsJSON:   profile.OptionsJSON,
+		CreatedAt:     now,
+		UpdatedAt:     now,
+	}); err != nil {
+		t.Fatal(err)
+	}
+
+	srv := NewWithConfig(Config{Meta: metaStore})
+	defer srv.Close()
+	schemaInitCalls := 0
+	srv.initTenantSchemaAsync(context.Background(), tenantID, "unused-dsn", tenant.ProviderTiDBZero, func(context.Context, string) error {
+		schemaInitCalls++
+		return nil
+	})
+
+	if schemaInitCalls != 1 {
+		t.Fatalf("schema init calls = %d, want 1", schemaInitCalls)
+	}
+	updated, err := metaStore.GetTenant(context.Background(), tenantID)
+	if err != nil {
+		t.Fatalf("GetTenant: %v", err)
+	}
+	if updated.Status != meta.TenantActive {
+		t.Fatalf("status = %q, want %q", updated.Status, meta.TenantActive)
+	}
+	if updated.SchemaVersion != targetVersion {
+		t.Fatalf("schema version = %d, want %d", updated.SchemaVersion, targetVersion)
+	}
+}
+
 func TestAutoEmbeddingProfileForTenantEnsuresDefaultProfile(t *testing.T) {
 	metaStore, err := meta.Open(testDSN)
 	if err != nil {
--- a/pkg/server/provision_test.go
+++ b/pkg/server/provision_test.go
@@ -1860,6 +1860,7 @@ func TestInitTenantSchemaAsyncPersistsTargetSchemaVersion(t *testing.T) {
 	srv := NewWithConfig(Config{Meta: metaStore})
 	defer srv.Close()
 	schemaInitCalls := 0
+	// Direct invocation is blocking; production launches this function in a worker.
 	srv.initTenantSchemaAsync(context.Background(), tenantID, "unused-dsn", tenant.ProviderTiDBZero, func(context.Context, string) error {
 		schemaInitCalls++
 		return nil
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -4977,11 +4977,7 @@ func (s *Server) updateTenantSchemaVersionForProfile(ctx context.Context, tenant
 	if err != nil {
 		return fmt.Errorf("resolve tenant auto-embedding profile: %w", err)
 	}
-	tidbMode, err := tenant.TiDBEmbeddingModeForTenantMode(profile.mode)
-	if err != nil {
-		return err
-	}
-	version, err := tenantschema.TiDBTenantSchemaVersionForEmbeddingModeProfile(tidbMode, profile.schemaProfile)
+	version, err := tenant.TiDBTenantSchemaVersionForEmbeddingMode(profile.mode, profile.schemaProfile)
 	if err != nil {
 		return err
 	}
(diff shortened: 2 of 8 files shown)


pkg/tenant/schema/context.go
```diff
@@ -13,7 +13,8 @@ func WithTenantID(ctx context.Context, tenantID string) context.Context {
 	return context.WithValue(ctx, tenantIDContextKey{}, tenantID)
 }
 
-func tenantIDFromContext(ctx context.Context) string {
+// TenantIDFromContext returns the tenant attribution attached with WithTenantID.
+func TenantIDFromContext(ctx context.Context) string {
```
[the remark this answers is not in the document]
srstack: Fixed in `004df541`: `WithTenantID` now returns nil for nil contexts, `TenantIDFromContext(nil)` returns an empty string, and `TestTenantIDContext` covers the nil and empty-tenant cases.

pkg/tenant/pool.go
```diff
@@ -865,9 +865,11 @@ func (p *Pool) createBackend(ctx context.Context, t *meta.Tenant) (*backend.Dat9
...
 					// Record the tenant-profile-specific version only after the
 					// schema has been confirmed to match that tenant's profile.
-					// Any tenant whose schema diverges between two consecutive
-					// opens will be caught on the next open because its stored
-					// version will differ from the profile-derived target.
+					// Version-matched opens trust this durable value and skip the
+					// physical diff by default; out-of-band schema drift is detected
+					// only when the target schema version changes, or through an
+					// explicit validation/repair path. SkipTiDBSchemaCheck remains
+					// the opt-out for externally managed schemas.
```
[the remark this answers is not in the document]
srstack: Fixed in `004df541`: reworded the comment to match `PoolConfig.SkipTiDBSchemaCheck` exactly: it disables Acquire-time TiDB schema ensure/check, mainly for tests using TiDB-class providers against plain MySQL.

mornyx (dismissed):

pkg/server/server.go
```diff
@@ -5374,6 +5400,9 @@ func (s *Server) initTenantSchemaAsync(ctx context.Context, tenantID, tenantDSN,
 				return
 			}
 			err = s.finalizeTenantSchemaInit(ctx, tenantID, tenantDSN, provider)
+			if err == nil {
+				err = s.updateTenantSchemaVersionForProfile(ctx, tenantID, provider)
```
mornyx: This is the key behavior change: `updateTenantSchemaVersionForProfile` returning an error now keeps the tenant in `Provisioning` and falls through to the retry branch (line 5423). Previously, activation happened as soon as `schemaInit` + `finalizeTenantSchemaInit` succeeded.

This is defensible — we don't want to activate a tenant whose durable version isn't persisted — but it means a transient meta-DB outage during this persist will cause repeated schema-init retries rather than a clean activation. Consider either:
- Logging a distinct event/metric here (e.g. `schema_init_version_persist_failed`) so operators can distinguish "schema init failed" from "version persist failed" in the retry log, or
- Making the version persist best-effort (log + continue to activate), since the pool's `createBackend` path already re-derives and persists the version on first open if it's missing/mismatched.

Either is fine; just want to confirm the stricter behavior is intentional.

pkg/server/server.go
```diff
@@ -4965,6 +4969,28 @@ func (s *Server) schemaInitForTenant(tenantID, provider string, fallback func(co
 	}
 }
 
+func (s *Server) updateTenantSchemaVersionForProfile(ctx context.Context, tenantID, provider string) error {
+	if s.meta == nil || !tenant.UsesTiDBAutoEmbedding(provider) {
+		return nil
+	}
+	profile, err := s.autoEmbeddingProfileForTenant(ctx, tenantID)
```
mornyx, line 4976: This re-derives the profile → mode → version using the same two-step mapping (`TiDBEmbeddingModeForTenantMode` + `TiDBTenantSchemaVersionForEmbeddingModeProfile`) that `pkg/tenant/pool.go:175` (`tenantSchemaVersionForEmbeddingMode`) already performs on the open path. Both produce the same value for the same profile, so there's no correctness issue, but a shared helper (e.g. a `schema.VersionForTenantProfile(mode, profile)` in a common spot) would prevent drift if the derivation ever changes. Minor — not blocking.

pkg/tenant/schema/context.go
```diff
@@ -7,13 +7,17 @@ type tenantIDContextKey struct{}
 // WithTenantID annotates schema-init contexts so user_schema DB pool metrics can
 // retain tenant attribution without changing public schema-init function shapes.
 func WithTenantID(ctx context.Context, tenantID string) context.Context {
-	if tenantID == "" {
+	if ctx == nil || tenantID == "" {
```
mornyx, line 12: Nice defensive addition of the `ctx == nil` guard. Small note: there are two parallel tenant-ID context packages with identical shapes but **incompatible keys** — this one (`schema.tenantIDContextKey{}`) and `pkg/tenantctx` (`tenantctx.key{}`). A tenant ID set by auth middleware via `tenantctx.WithTenantID` won't be visible to `schema.TenantIDFromContext`, and vice versa. Not a problem today since the schema paths are tagged explicitly via `WithTenantID`, but worth a comment on `WithTenantID` noting that this is intentionally separate from the request-scope `tenantctx`, so a future reader doesn't assume interoperability.

pkg/server/provision_test.go
```diff
@@ -1806,6 +1814,72 @@ func TestSchemaInitForTenantUsesFTSOnlyProfileWhenDatabaseAutoEmbeddingDisabled(
...
+		t.Fatalf("GetTenant: %v", err)
+	}
+	if updated.Status != meta.TenantActive {
+		t.Fatalf("status = %q, want %q", updated.Status, meta.TenantActive)
+	}
+	if updated.SchemaVersion != targetVersion {
+		t.Fatalf("schema version = %d, want %d", updated.SchemaVersion, targetVersion)
+	}
+}
+
```
mornyx, line 1883: Good test — and it's race-free because `initTenantSchemaAsync` is synchronous when called directly (the "Async" refers to production's `startServerWorker` goroutine launch, not the function body, which is a blocking `for` loop). The `schemaInitCalls` closure runs on the calling goroutine, then the test reads it on the same goroutine.

One nit: the test name says `...AsyncPersists...` but the function is synchronous here. Consider adding a brief comment above the call clarifying that the direct (non-goroutine) invocation makes it blocking, so the next reader doesn't go looking for a `sync.WaitGroup` or a sleep.

pkg/tenant/pool.go
```diff
@@ -857,17 +857,21 @@ func (p *Pool) createBackend(ctx context.Context, t *meta.Tenant) (*backend.Dat9
...
 				}
 				ensureSchemaDurationMs += float64(time.Since(ensureSchemaStart).Microseconds()) / 1000.0
 				if p.metaStore != nil {
 					// Record the tenant-profile-specific version only after the
 					// schema has been confirmed to match that tenant's profile.
-					// Any tenant whose schema diverges between two consecutive
-					// opens will be caught on the next open because its stored
-					// version will differ from the profile-derived target.
+					// Version-matched opens trust this durable value and skip the
+					// physical diff by default; out-of-band schema drift is detected
```
mornyx, line 862: Comment update is clear and well-written. The trust model — version-matched opens skip the physical diff, `SkipTiDBSchemaCheck` disables it entirely — is much easier to reason about now.

pkg/tenant/schema/tidb_auto.go
```diff
@@ -654,6 +654,13 @@ func currentTiDBTenantSchemaVersion(stmts []string) int {
 	return schemaspec.CRC32Version(specStmts)
 }
 
+func tenantSchemaLogFields(ctx context.Context, fields ...zap.Field) []zap.Field {
```
mornyx, line 657: Clean mechanical pass — every `tenant_tidb_schema_*` log site now routes through `tenantSchemaLogFields`, and the `tenant_id` field is prepended consistently. For completeness: the `tenant_detect_tidb_embedding_mode_*` family (in `DetectTiDBEmbeddingMode`, ~line 951) is correctly left out of scope since that function takes no `ctx`. If attribution is ever wanted there, it'll need a ctx-accepting variant.

mornyx (commented):
Reviewed across all six files. The core change — persisting the profile-derived schema version after async init and threading `tenant_id` into schema logs — is well-scoped and the tests are solid. A few comments below. Summary of the main points:

1. **Behavior change (worth confirming):** `updateTenantSchemaVersionForProfile` now gates tenant activation. A transient meta-DB failure to persist the version will keep the tenant in `Provisioning` and trigger retries. That's arguably correct (we don't want to activate without a durable version), but it's a stricter contract than before — worth a sentence in the PR body or a log line distinguishing this failure from a schema-init failure.
2. **Duplicated version derivation:** The server init path (`updateTenantSchemaVersionForProfile`) and the pool open path (`createBackend`) both independently resolve profile → mode → version via the same two-step mapping. Not a correctness issue (same input → same CRC), but a shared helper would reduce drift risk if the derivation changes.
3. **Two parallel tenant-ID context packages:** `pkg/tenant/schema` and `pkg/tenantctx` use incompatible keys, so a tenant ID set by one is invisible to the other. The PR doesn't unify them (fine, out of scope), but worth noting for future readers.
4. The `tenant_detect_tidb_embedding_mode_*` log family is correctly out of scope (no ctx param).

pkg/tenant/embedding_mode_test.go
```diff
@@ -0,0 +1,37 @@
...
+		}
+		if got != want {
+			t.Fatalf("version for %q = %d, want %d", mode, got, want)
+		}
+	}
+
+	if _, err := TiDBTenantSchemaVersionForEmbeddingMode("invalid", profile); err == nil {
+		t.Fatal("TiDBTenantSchemaVersionForEmbeddingMode(invalid) succeeded, want error")
+	}
+}
```
mornyx, line 37: Good coverage — tests both valid modes (Auto, FTSOnly) against the two-step derivation and the invalid-mode error path. The  case confirms the error wraps through  correctly.

mornyx (approved):
Round 2 — all three points from the first review are addressed cleanly:

1. **Activation gating is now observable.** The  log event +  metric () make it easy to distinguish a version-persist failure from a schema-init failure in the retry path. Confirmed this is the stricter (correct) behavior.
2. **Shared version derivation.** The new  () unifies the profile → mode → version mapping. Both the server init path and pool open path now call it; the old  in pool.go is removed. Good deduplication. (Note:  still calls  directly, but that's correct — it needs the  enum for its , not just the version int, so it can't use the version-only helper.)
3. **Context separation documented.** The comment on  () clearly explains why schema-init uses a separate key from .

Plus the test comment about the blocking direct invocation is a nice touch. LGTM.

qiffang (commented):
## adversary-2 review — GREEN on PR #736

No blockers. Clean follow-up to PR #733 review comments.

### What's correct

**Schema version persistence** — `updateTenantSchemaVersionForProfile` runs after successful `finalizeTenantSchemaInit`, persists the profile-derived version so subsequent opens skip the physical diff. Guard with `UsesTiDBAutoEmbedding` prevents running for non-TiDB providers. Error on version persist correctly fails the entire init attempt (schema will be re-ensured on retry, and the ensure is idempotent).

**Shared derivation** — `TiDBTenantSchemaVersionForEmbeddingMode` extracted from pool-local to `pkg/tenant`, used by both pool and server paths. Error wrapping adds context. Pool's private `tenantSchemaVersionForEmbeddingMode` removed — no dead code left.

**Tenant context propagation** — `WithTenantID` injected in all code paths: `schemaInitForTenant` (all 4 return paths including fallbacks), `createBackend` schema ensure, and `initTenantSchemaAsync` version persist. `TenantIDFromContext` exported for cross-package use. Nil-context guard prevents panic.

**Log tagging** — `tenantSchemaLogFields` prepends `tenant_id` when present. All schema init/validate/diff/ensure/repair log sites updated consistently (22 call sites by my count).

1. **Version persist failure = full init failure** (`server.go:5396`): `err = versionErr` means a version-persist failure causes the tenant to stay in error/provisioning state even though schema is correct. This is safe because the next retry's ensure is idempotent, but it does mean a transient meta-store write failure blocks tenant activation. If this is intentional (review-requested behavior), fine — just noting it.

2. **`tenantSchemaLogFields` allocation**: `append([]zap.Field{...}, fields...)` allocates a new slice on every call with a tenant ID. These are infrequent schema ops, so not a performance concern, but if this pattern spreads to hot paths it would need a pool or pre-allocated approach.

### Checklist
- All inputs handled? ✅ (nil ctx, empty tenant ID, non-TiDB providers, missing profile)
- Edge cases covered? ✅ (fallback paths, version persist failure, invalid embedding mode)
- Failure behavior defined? ✅ (persist failure → retry on next init)
- Tests? ✅ (`TestInitTenantSchemaAsyncPersistsTargetSchemaVersion`, `TestTenantIDContext`, `TestTenantSchemaLogFieldsIncludesTenantID`, `TestTiDBTenantSchemaVersionForEmbeddingMode`, tenant ID propagation assertion in existing test)

Verdict: **GREEN**

qiffang (approved):
**GREEN** — Clean follow-up to PR #733 review comments. No blockers.

### What this PR does

1. **Schema version persistence after profile-based init** — `updateTenantSchemaVersionForProfile` called after `finalizeTenantSchemaInit`, so subsequent pool opens with a matching version skip the physical diff entirely. Version persist failure → re-enters full init retry (correct fail-safe).

2. **`TenantIDFromContext` exported** — Previously private `tenantIDFromContext` is now public. All 22 `logger.*` call sites in `tidb_auto.go` now go through `tenantSchemaLogFields()` which prepends `tenant_id` when present. This gives per-tenant log attribution for schema init/ensure/validate/repair/diff operations.

3. **Shared schema-version derivation** — `tenantSchemaVersionForEmbeddingMode` moved from pool-local to exported `TiDBTenantSchemaVersionForEmbeddingMode` in `pkg/tenant/embedding_mode.go`, used by both pool (`createBackend`) and server (`updateTenantSchemaVersionForProfile`).

4. **`WithTenantID` nil-ctx guard** — Returns nil instead of panicking on `context.WithValue(nil, ...)`.

### Test coverage

- `TestInitTenantSchemaAsyncPersistsTargetSchemaVersion` — version persistence end-to-end
- `TestTenantIDContext` — round-trip, nil ctx, empty tenant ID edge cases
- `TestTenantSchemaLogFieldsIncludesTenantID` — log field prepend with/without tenant
- `TestTiDBTenantSchemaVersionForEmbeddingMode` — mode mapping + invalid mode error
- Existing `TestSchemaInitForTenantEnsuresDatabaseBeforeProfileInit` extended with tenant ID assertion

### Non-blocking observations

1. `tenantSchemaLogFields` allocates a new `[]zap.Field` slice on every call via `append([]zap.Field{...}, fields...)`. At current call volume this is fine, but if hot-path performance ever matters, a `make([]zap.Field, 0, len(fields)+1)` pre-allocation would avoid the copy.

2. Version persist failure in `initTenantSchemaAsync` sets `err = versionErr`, causing the outer retry loop to re-run full schema init + finalize. This is the right fail-safe (idempotent re-init), but worth noting that a transient meta-store write failure triggers a heavier retry than strictly needed.

PR #738, opened 2026-07-14
e2e: relax provision status check to accept active for pool tenants

srstack:
When /v1/provision claims a pre-provisioned pool tenant that is already active (schema init completed), the API returns status=active instead of provisioning. This PR updates the e2e smoke tests to accept both statuses.

- `e2e/api-smoke-test.sh`: relax check from strict `provisioning` to `provisioning|active`
- `e2e/native-smoke-test.sh`: same

## Summary by CodeRabbit

* **Tests**
  * Updated provisioning smoke tests to accept tenants entering either `provisioning` or `active` status after creation.
  * Preserved validation of all other provisioning response fields.

### DIFF
--- a/e2e/api-smoke-test.sh
+++ b/e2e/api-smoke-test.sh
@@ -243,7 +243,7 @@ TENANT_ID=$(printf '%s' "$body" | jq -r '.tenant_id // empty')
 INIT_STATUS=$(printf '%s' "$body" | jq -r '.status // empty')
 check_cmd "response contains tenant_id" test -n "$TENANT_ID"
 check_cmd "response contains api_key" test -n "$API_KEY"
-check_cmd "provision response status is provisioning or active" bash -c 'case "$1" in provisioning|active) exit 0;; *) exit 1;; esac' _ "$INIT_STATUS"
+check_cmd "provision response status is provisioning or active (got=$INIT_STATUS)" bash -c 'case "$1" in provisioning|active) exit 0;; *) exit 1;; esac' _ "$INIT_STATUS"
 check_cmd "provision response contains api_key, status, tenant_id" bash -c "echo \"\$1\" | jq -e '.api_key and .status and .tenant_id' >/dev/null" _ "$body"
 
 step "2" "Poll tenant status via /v1/status"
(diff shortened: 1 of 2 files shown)


mornyx (approved):
## PR #738 Review — e2e: relax provision status check to accept active for pool tenants

### Summary

This PR relaxes the e2e smoke test provision-status assertion from a strict `provisioning` match to accept either `provisioning` or `active`. The change is correct and well-justified.

### Root cause verification

I traced the provision endpoint code path to confirm the claim:

1. **Pool-claim path** (`handleProvision` → `claimAdminTenantFromPool`): The claimed pool tenant status comes from `row.Tenant.Status` (line 1744 of `server.go`). The underlying query `ClaimOldestFreeTenantPoolBinding` (line 1970 of `meta.go`) filters `AND t.status = TenantActive`, so pool-claimed tenants are **always** `active`. The handler at line 4443 only starts schema init if `res.Status == meta.TenantProvisioning`, which is skipped for pool claims — confirming pool tenants are pre-provisioned.

2. **Non-pool path** (`handleProvision` → `provisionTenant`): Returns `Status: meta.TenantProvisioning` (line 5228 of `server.go`).

So the provision response status is `provisioning` for fresh provisions and `active` for pool-claimed tenants. The old e2e tests would fail on any deployment with an active tenant pool — this PR fixes that.

### Implementation correctness

- The new `check_cmd` pattern `bash -c 'case "$1" in provisioning|active) exit 0;; *) exit 1;; esac' _ "$INIT_STATUS"` is a clean way to accept multiple values, consistent with the existing `bash -c '...' _ "$body"` pattern already used at line 247 of `api-smoke-test.sh`.
- The `(got=$INIT_STATUS)` / `(got=$CREATE_STATUS)` suffix in the description provides good debuggability — if the check fails, the observed status is visible in the failure message.
- Both `api-smoke-test.sh` and `native-smoke-test.sh` are updated consistently.

### Scope check

- **`cli-smoke-test.sh` line 376** checks `ctx fork initial status is provisioning` — this is the fork-tenant code path (`fork.go:351,491`), which always returns `TenantProvisioning` and does **not** go through the pool-claim path. This check is correct and should **not** be relaxed. ✅ The PR correctly does not touch it.
- No other e2e scripts check provision status.

### CI status

- `ci`: ✅ success

### Verdict: ✅ Approve

Small, correct, well-scoped fix. The root cause is verified in the provision handler code, the implementation pattern is consistent with existing test conventions, and the scope is limited to the two affected files.

PR #741, opened 2026-07-14
Add TiDB Cloud private endpoint host mapping

srstack:
## Summary
- add `DRIVE9_TIDBCLOUD_PRIVATE_ENDPOINT_HOST_MAP` for public host to private host resolution when TiDB Cloud private endpoint host is absent
- prefer cluster-provided private hosts, fall back to host map, then legacy single-host provider overrides only when no map is configured
- emit metrics/logs for missing private endpoint mapping and TiDB Cloud metadata still not ready after 5 minutes

## Tests
- `go test ./pkg/tenant/tidbcloudnative -count=1`
- `go test ./pkg/server -run 'TestAdminTenantPoolCreateUsesPrivateEndpointDBTLS|TestTenantPoolMetadataResumeUsesPrivateEndpointDBTLS|TestTenantPoolMetadataResumeRerunsWhenTriggeredWhileActive' -count=1`\n- `go test ./pkg/server -count=1`\n- `git diff --check -- cmd/drive9-server/main.go pkg/tenant/tidbcloudnative/provisioner.go pkg/tenant/tidbcloudnative/provisioner_test.go`

## Summary by CodeRabbit

* **New Features**
  * Added TiDB Cloud private-endpoint host mapping via environment variables, supporting both `public_host=private_host` and `public_host:private_host`.
  * Added tenant count metrics, including periodic leader-only observation and a labeled `drive9_tenant_count` gauge for total non-deleted and active tenants.
* **Bug Fixes**
  * Improved private-endpoint host resolution and propagated per-cluster/branch readiness failures for clearer error reporting.
  * Enhanced “metadata not ready” alerting and improved warning log detail for connection and datastore operation events.
* **Documentation**
  * Updated server usage output to describe the new mapping variables and their precedence over legacy private-endpoint settings.

### DIFF
--- a/cmd/drive9-server/main.go
+++ b/cmd/drive9-server/main.go
@@ -577,6 +577,8 @@ environment:
   DRIVE9_TIDBCLOUD_NATIVE_DEFAULT_DATABASE_NAME default tidb_cloud_native database name (default: tidbcloud_fs)
   DRIVE9_TIDBCLOUD_NATIVE_PUBLIC_KEY optional default TiDB Cloud API public key for tidb_cloud_native create/delete when caller omits it
   DRIVE9_TIDBCLOUD_NATIVE_PRIVATE_KEY optional default TiDB Cloud API private key for tidb_cloud_native create/delete when caller omits it
+  DRIVE9_TIDBCLOUD_NATIVE_USE_PRIVATE_ENDPOINT true|false to use TiDB Cloud private endpoint hosts for tidb_cloud_native
+  DRIVE9_TIDBCLOUD_PRIVATE_ENDPOINT_HOST_MAP comma-separated public_host=private_host mappings used when private endpoint host is absent
   DRIVE9_TENANT_POOL_MAX_SIZE maximum admin tenant pool size (default: %d)
   DRIVE9_TENANT_POOL_REFILL_FREE_RATIO claim-triggered refill runs when active free tenants fall below this ratio (default: %.2f)
   DRIVE9_SLOCK_ORIGIN Slock browser origin; when set, enables /v1/auth/slock/*
--- a/cmd/drive9-server/main.go
+++ b/cmd/drive9-server/main.go
@@ -580,6 +580,8 @@ environment:
   DRIVE9_TIDBCLOUD_NATIVE_USE_PRIVATE_ENDPOINT true|false to use TiDB Cloud private endpoint hosts for tidb_cloud_native
   DRIVE9_TIDBCLOUD_PRIVATE_ENDPOINT_HOST_MAP comma-separated public_host=private_host mappings (also accepts public_host:private_host);
                                              when set, disables legacy single-host private endpoint overrides
+  DRIVE9_TIDBCLOUD_TENCENT_PRIVATE_ENDPOINT_HOST legacy tencentcloud private endpoint fallback host, used only when host map is unset
+  DRIVE9_TIDBCLOUD_ALICLOUD_PRIVATE_ENDPOINT_DOMAIN legacy alicloud private endpoint fallback domain, used only when host map is unset
   DRIVE9_TENANT_POOL_MAX_SIZE maximum admin tenant pool size (default: %d)
   DRIVE9_TENANT_POOL_REFILL_FREE_RATIO claim-triggered refill runs when active free tenants fall below this ratio (default: %.2f)
   DRIVE9_SLOCK_ORIGIN Slock browser origin; when set, enables /v1/auth/slock/*
--- a/cmd/drive9-server/main.go
+++ b/cmd/drive9-server/main.go
@@ -578,7 +578,8 @@ environment:
   DRIVE9_TIDBCLOUD_NATIVE_PUBLIC_KEY optional default TiDB Cloud API public key for tidb_cloud_native create/delete when caller omits it
   DRIVE9_TIDBCLOUD_NATIVE_PRIVATE_KEY optional default TiDB Cloud API private key for tidb_cloud_native create/delete when caller omits it
   DRIVE9_TIDBCLOUD_NATIVE_USE_PRIVATE_ENDPOINT true|false to use TiDB Cloud private endpoint hosts for tidb_cloud_native
-  DRIVE9_TIDBCLOUD_PRIVATE_ENDPOINT_HOST_MAP comma-separated public_host=private_host mappings used when private endpoint host is absent
+  DRIVE9_TIDBCLOUD_PRIVATE_ENDPOINT_HOST_MAP comma-separated public_host=private_host mappings (also accepts public_host:private_host);
+                                             when set, disables legacy single-host private endpoint overrides
   DRIVE9_TENANT_POOL_MAX_SIZE maximum admin tenant pool size (default: %d)
   DRIVE9_TENANT_POOL_REFILL_FREE_RATIO claim-triggered refill runs when active free tenants fall below this ratio (default: %.2f)
   DRIVE9_SLOCK_ORIGIN Slock browser origin; when set, enables /v1/auth/slock/*
--- a/pkg/datastore/store.go
+++ b/pkg/datastore/store.go
@@ -3036,7 +3036,7 @@ func observeStoreOp(ctx context.Context, op string, start time.Time, errp *error
 		if shouldLogStoreOpFailure(result) {
 			logger.Error(ctx, "datastore_op_failed", zap.String("operation", op), zap.String("result", result), zap.Error(*errp))
 		} else {
-			logger.Warn(ctx, "datastore_op_failed", zap.String("operation", op), zap.String("result", result), zap.Error(*errp))
+			logger.Warn(ctx, "datastore_op_failed", zap.String("operation", op), zap.String("result", result), zap.String("detail", (*errp).Error()))
 		}
 	}
 	logger.InfoBenchTiming(ctx, "datastore_op_timing",
--- a/pkg/meta/meta.go
+++ b/pkg/meta/meta.go
@@ -2637,6 +2637,9 @@ func (s *Store) CountTenants(ctx context.Context) (out TenantCounts, err error) 
 			COALESCE(SUM(CASE WHEN status = ? THEN 1 ELSE 0 END), 0)
 		FROM tenants
 		WHERE status <> ?`, TenantActive, TenantDeleted).Scan(&out.TotalNonDeleted, &out.Active)
+	if err != nil {
+		err = fmt.Errorf("count tenants: %w", err)
+	}
 	return out, err
 }
 
--- a/pkg/meta/meta.go
+++ b/pkg/meta/meta.go
@@ -3517,7 +3517,7 @@ func observeMeta(ctx context.Context, op string, start time.Time, errp *error) {
 			// Connection closed during shutdown — suppress the noisy log and
 			// only record the metric below.
 		case "not_found", "duplicate":
-			logger.Warn(ctx, "meta_op_failed", zap.String("operation", op), zap.String("result", result), zap.Error(*errp))
+			logger.Warn(ctx, "meta_op_failed", zap.String("operation", op), zap.String("result", result), zap.String("detail", (*errp).Error()))
 		case "error":
 			logger.Error(ctx, "meta_op_failed", zap.String("operation", op), zap.String("result", result), zap.Error(*errp))
 		}
--- a/pkg/meta/meta.go
+++ b/pkg/meta/meta.go
@@ -90,6 +90,11 @@ type Tenant struct {
 	UpdatedAt          time.Time
 }
 
+type TenantCounts struct {
+	TotalNonDeleted int64
+	Active          int64
+}
+
 type TenantAutoEmbeddingProfile struct {
 	TenantID      string
 	EmbeddingMode string
@@ -2621,6 +2626,18 @@ func (s *Store) ListTenantsByStatus(ctx context.Context, status TenantStatus, li
 	defer func() { _ = rows.Close() }()
 
 	return scanTenantRows(rows)
+}
+
+func (s *Store) CountTenants(ctx context.Context) (out TenantCounts, err error) {
+	start := time.Now()
+	defer observeMeta(ctx, "count_tenants", start, &err)
+	err = s.db.QueryRowContext(ctx, `
+		SELECT
+			COUNT(*),
+			COALESCE(SUM(CASE WHEN status = ? THEN 1 ELSE 0 END), 0)
+		FROM tenants
+		WHERE status <> ?`, TenantActive, TenantDeleted).Scan(&out.TotalNonDeleted, &out.Active)
+	return out, err
 }
 
 // ListTenantsByStatusAfter returns one keyset page of tenants after
--- a/pkg/meta/meta_test.go
+++ b/pkg/meta/meta_test.go
@@ -841,6 +841,48 @@ func TestListTenantsByStatus(t *testing.T) {
 	}
 }
 
+func TestCountTenants(t *testing.T) {
+	s := newControlStore(t)
+	now := time.Now().UTC()
+	for _, tc := range []struct {
+		id     string
+		status TenantStatus
+	}{
+		{id: "count-active", status: TenantActive},
+		{id: "count-provisioning", status: TenantProvisioning},
+		{id: "count-failed", status: TenantFailed},
+		{id: "count-deleted", status: TenantDeleted},
+	} {
+		if err := s.InsertTenant(context.Background(), &Tenant{
+			ID:               tc.id,
+			Status:           tc.status,
+			DBHost:           "127.0.0.1",
+			DBPort:           4000,
+			DBUser:           "root",
+			DBPasswordCipher: []byte("cipher"),
+			DBName:           "tenant_db",
+			DBTLS:            true,
+			Provider:         "tidb_zero",
+			SchemaVersion:    1,
+			CreatedAt:        now,
+			UpdatedAt:        now,
+		}); err != nil {
+			t.Fatal(err)
+		}
+	}
+
+	got, err := s.CountTenants(context.Background())
+	if err != nil {
+		t.Fatal(err)
+	}
+	if got.TotalNonDeleted != 3 {
+		t.Fatalf("TotalNonDeleted = %d, want 3", got.TotalNonDeleted)
+	}
+	if got.Active != 1 {
+		t.Fatalf("Active = %d, want 1", got.Active)
+	}
+}
+
 func TestListTenantsByStatusAfterKeyset(t *testing.T) {
 	s := newControlStore(t)
 	now := time.Now().UTC().Truncate(time.Millisecond)
--- a/pkg/meta/meta_test.go
+++ b/pkg/meta/meta_test.go
@@ -876,10 +876,10 @@ func TestCountTenants(t *testing.T) {
 		t.Fatal(err)
 	}
 	if got.TotalNonDeleted != 3 {
-		t.Fatalf("TotalNonDeleted = %d, want 3", got.TotalNonDeleted)
+		t.Errorf("TotalNonDeleted = %d, want 3", got.TotalNonDeleted)
 	}
 	if got.Active != 1 {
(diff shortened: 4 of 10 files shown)


srstack:
Reviewed head (read-only diff review — I did not compile or run the suite in this checkout). The host-map feature and the resolution-priority refactor are well-structured, and the new metrics/logging for stuck metadata are useful. One correctness question on the incomplete-detection logic that I'd like confirmed before this merges.

## Question: `clusterConnectionIncomplete` / `branchConnectionIncomplete` fall-through returns `false` even when resolution will later fail

In the new `clusterConnectionIncomplete` (and the identical `branchConnectionIncomplete`), for `usePrivateEndpoint` with a non-empty private port, empty private host, and non-empty public host:

```go
if _, ok := p.mappedPrivateEndpointHost(info.Endpoints.Public.Host); ok {
    return false // mapped → complete
}
if len(p.privateEndpointHostMap) == 0 && p.privateEndpointOverrideHost() != "" {
    return false // legacy override → complete
}
return false // ← fall-through: also "complete"
```

The final `return false` fires when: private host is empty, public host is present, **but that public host has no entry in the host map and there's no legacy override**. That reports the metadata as *complete*, so `clusterProvisionMetadataIncomplete` → the provision path treats it as ready and proceeds to `resolveClusterEndpoint` → `resolvePrivateEndpointHost`, which for exactly this case **returns an error** ("no private host mapping").

So the incomplete-check says "complete" while resolution says "error" — they disagree. The practical consequence: `waitForClusterProvisionMetadata` will **not** retry/wait, because the metadata is judged complete; it goes straight to resolution and fails.

Two sub-cases hide behind this:
1. **Misconfiguration** (operator forgot to add this public host to the map): failing fast is arguably correct — waiting won't help. 
2. **Metadata not yet ready** (TiDB Cloud hasn't populated the private host yet, and the public host happens to be one you *would* map): here you'd want to keep waiting, but the current logic treats "private host empty but public host present" as complete and bails to a hard error instead of waiting out the 10-minute deadline.

Is the fall-through-`false` intentional? If the intent is "a present public host means we can always resolve (via map or override), so it's complete," then the unmapped-public-host case violates that assumption and should probably return `true` (stay incomplete → keep waiting) OR the error should be raised explicitly here rather than deferred to `resolvePrivateEndpointHost`. As written, sub-case 2 loses the retry window. A comment on why the fall-through is `false` would also help — right now three branches all `return false` and the reasoning for the last one isn't obvious.

## Things that look good
- `parsePrivateEndpointHostMap` is robust: `,`/`;`/newline separators, `=`/`:` delimiters, quote-trimming, lowercase+trailing-dot key normalization, and duplicate-with-conflict detection. Nice.
- Resolution priority (cluster-provided private host → host map → legacy single-host override only when no map configured) is centralized in `resolvePrivateEndpointHost` and consistent between cluster and branch paths.
- Dropping the startup hard-requirement for tencent/alicloud private host (now that a map or runtime error covers it) is reasonable, and the test rename/rewrite reflects the new "no startup override required" behavior correctly.
- The 5-minute `metadata_not_ready` alert uses a `recorded *bool` one-shot guard so it fires once per wait rather than every poll — good, avoids metric/log spam.
- `clusterInfoFromResponse` now returning `(*ClusterInfo, error)` and callers preserving the partial cluster (id/org) on error is handled consistently across the provision, batch, and pool-metadata paths.

## Note on the batch path
In `waitForBatchClusterProvisionMetadataGroup`, a `clusterInfoFromResponse` error now sets `errs[target.index]` and drops the cluster from `pending`. If that error is the "unmapped public host" case above, the same fast-fail-vs-wait question applies per-cluster in the batch — worth confirming it's the intended behavior there too.

Modulo the incomplete-vs-resolve disagreement, the change is clean and well-tested. I'd like the fall-through semantics confirmed (or a `true`/explicit-error fix + comment) before approving; the rest LGTM. Recommend confirming CI is green since I didn't run the suite locally.

srstack:
Re-reviewed `eac6395d` (read-only diff). This fully resolves the incomplete-vs-resolve disagreement I raised, and does it more precisely than I'd suggested — the two sub-cases are now distinguished correctly.

**The fix (in the new consolidated `endpointConnectionIncomplete` helper):** the fall-through is now `return true` instead of `return false`, so:
- **Host map configured** → returns `false` (complete), letting `resolvePrivateEndpointHost` **fail fast** on an unmapped public host. Correct: an unmapped host is a config error, waiting won't fix it.
- **No map and no legacy override** → returns `true` (incomplete) → **keeps waiting** for the API to populate the private host. This is exactly the retry window that sub-case 2 was losing before.

That's the right split, and the two new assertions in `TestClusterConnectionIncompleteWhenPrivateEndpointMissing` / `TestBranchConnectionIncompleteWhenPrivateEndpointMissing` pin both branches (no-map → keep waiting; map present → stop waiting and fail-fast). 

**Bonus correctness point I'd missed:** `privateEndpointOverrideHost()` now returns `""` when a host map is set, with a comment explaining why — falling back to the provider-wide legacy host for an unmapped public host could route a cluster in a different VPC to the wrong private endpoint. Making the map authoritative (no silent legacy fallback) is the safer choice, and the provision tests were updated to set `alicloudPrivateEndpointHost` alongside the map to prove map-takes-priority.

**Refactor is a net win:** collapsing the duplicated cluster/branch incomplete-detection into a single `endpointConnectionIncomplete(publicHost, privateHost, ports, userPrefix)` removes the copy-paste divergence risk between the two paths.

Everything I flagged is addressed; the logic is now internally consistent (incomplete-check and resolve agree), and the semantics are well-tested. LGTM.

(As before: read-only review — I did not compile or run the suite in this checkout; recommend confirming CI is green.)

srstack:
Re-reviewed `a5bdeb05` + `4852bc96` (read-only diff). This reverses the fall-through I LGTM'd last round (`return true` → `return false`) — and I now think this version is **more correct than my earlier suggestion**. Walking through it changed my mind:

**Why `return false` (fail-fast) beats `return true` (keep waiting) here.** The helper still gates on `privatePort == 0 → incomplete` *before* reaching the fall-through. So by the time control gets there, the `private` endpoint object already has a port but an empty host. Since `private.host` and `private.port` live in the same `endpoints.private` object, "port present but host empty" reads much more like "this deployment shape simply doesn't return a private host, resolve it via the map" than "the private object hasn't been populated yet." Under that reading, my earlier `return true` had a flaw I missed: for a deployment where TiDB Cloud never emits a private host (map-only by design), it would poll uselessly until the 10-minute deadline instead of resolving via the map or failing fast immediately. The new comment ("public/private readiness is exposed together; empty private host after public host is visible is a resolution problem, not a not-ready problem") captures exactly that. So: withdraw my prior preference — `false` is right.

**The one thing that makes-or-breaks this** is the domain premise in the comment: that TiDB Cloud exposes public+private endpoint readiness together (equivalently, that `private.port` becoming non-zero is a reliable readiness signal, and a still-empty `private.host` at that point is never a transient not-yet-populated state). I can't verify that from the code — it's a TiDB Cloud API contract. It's plausible and self-consistent with the struct (port and host share the `private` object), and the `privatePort`/`publicHost` gates still preserve the genuine "nothing ready yet → keep waiting" path. Please just confirm that premise holds against the actual API behavior; if there's any window where `private.port` is set but `private.host` legitimately fills in a few polls later, this would fail-fast a bit too eagerly. Given the 5-minute not-ready alert now covers the "stuck" case observably, even the worst case is visible rather than silent.

**Test update matches:** both `...IncompleteWhenPrivateEndpointMissing` tests now assert a single "stop waiting once public host + private port + user prefix are ready" outcome, dropping the map-vs-no-map branch that the simplified logic no longer distinguishes. Consistent with the code.

**`4852bc96` (log noise):** swapping `zap.Error` → `zap.String("detail", ...)` on the warn-level `datastore_op_failed` / `meta_op_failed` (not_found/duplicate/conn-closed) branches keeps the detail text while removing the top-level `error` field so these expected outcomes don't inflate error-log alerts. Same rationale as the earlier schema/datastore noise reductions — consistent and correct. Error-level branches correctly keep `zap.Error`.

Modulo confirming the readiness premise (a product/API-contract question, not a code defect), this is clean and I'm comfortable with it. LGTM. (Read-only review — didn't run the suite; recommend confirming CI green.)

mornyx (commented):
## PR #741 Review — Add TiDB Cloud private endpoint host mapping

### Summary

This PR adds `DRIVE9_TIDBCLOUD_PRIVATE_ENDPOINT_HOST_MAP` for public→private host resolution when TiDB Cloud private endpoint metadata is absent, refactors endpoint resolution into a unified `resolvePrivateEndpointHost` pipeline, removes the startup-time requirement for provider-specific private endpoint overrides, and adds a 5-minute "metadata not ready" alert metric/log.

1. **Host map parsing** (`parsePrivateEndpointHostMap`): parses comma/semicolon/newline-separated `public_host=private_host` (or `public_host:private_host`) entries, lowercases and trailing-dot-strips keys, rejects duplicates with conflicting values, and rejects entries missing either host. ✅ Clean implementation.

2. **Endpoint resolution priority** (`resolvePrivateEndpointHost`): cluster-provided private host → host map lookup → legacy single-host override (only when no map is configured) → error. This is the correct priority order — cluster-provided metadata is authoritative, the map fills gaps, and the legacy override is disabled when a map is present to avoid routing clusters from different VPCs to the wrong endpoint. ✅

3. **`endpointConnectionIncomplete` semantic change**: In private endpoint mode, when `privateHost == ""` but `publicHost != ""`, the connection is now considered *not incomplete* (previously it was incomplete). This stops the poll loop early so `resolvePrivateEndpointHost` can either map the public host or fail fast, instead of polling indefinitely for a private host that TiDB Cloud may never provide. ✅ This is the core improvement.

4. **Startup validation relaxation**: The startup-time requirement for `TENCENT_PRIVATE_ENDPOINT_HOST` / `ALICLOUD_PRIVATE_ENDPOINT_DOMAIN` when `USE_PRIVATE_ENDPOINT=true` is removed. This is correct — the host map or cluster-provided metadata may supply the private host, so requiring it at startup is overly strict. ✅

5. **Metadata not-ready alert** (`recordMetadataNotReadyAlert`): records a `drive9_service_operations_total{component="tidbcloud_native",operation="metadata_not_ready_5m",result=<reason>}` metric and a warn log after 5 minutes of unresolved metadata. The `*recorded` flag ensures one-shot emission per wait cycle. ✅

6. **Logging changes** (`store.go`, `meta.go`): Warn-level logs for expected results (`not_found`, `duplicate`) now use `zap.String("detail", (*errp).Error())` instead of `zap.Error(*errp)`. This reduces log noise (no stacktrace for expected misses) while preserving the error message. The `error` result still uses `zap.Error(*errp)`. ✅

7. **`clusterInfoFromResponse` now returns `(*tenant.ClusterInfo, error)`**: Callers in the single-provision, batch-provision, and pool-metadata-wait paths all handle the error correctly. Partial cluster info is preserved on error so callers can identify which cluster failed. ✅

### CI status

- `ci`: ✅ success

---

### Comments

#### 1. Old standalone `clusterConnectionIncomplete` / `branchConnectionIncomplete` functions are now test-only (nit)

The standalone (non-method) functions `clusterConnectionIncomplete(info, usePrivate, overridePrivateHost)` and `branchConnectionIncomplete(info, usePrivate, overridePrivateHost)` are no longer called from production code — only from tests (`provisioner_test.go` lines 1769, 1772, 1776, 1943, 1946). Consider either:
- Removing them and updating the old tests to use the method versions, or
- Adding a comment marking them as test-only helpers.

Not blocking — they are still exercised by tests and do not cause confusion since the method versions are clearly the production path.

#### 2. `:` separator in host map could split IPv6 addresses (observation)

`parsePrivateEndpointHostMap` falls back to `strings.Cut(entry, ":")` when `=` is not found. If a private endpoint host is an IPv6 address (e.g., `[::1]`), the `:` would split incorrectly. This is an acceptable edge case since private endpoint hosts are typically DNS hostnames, and the `=` syntax is the primary/recommended form. Just worth noting in documentation if IPv6 support becomes needed.

#### 3. Legacy private endpoint env vars (`TENCENT_PRIVATE_ENDPOINT_HOST`, `ALICLOUD_PRIVATE_ENDPOINT_DOMAIN`) are undocumented (nit, pre-existing)

The PR documents the new `DRIVE9_TIDBCLOUD_PRIVATE_ENDPOINT_HOST_MAP` and `DRIVE9_TIDBCLOUD_NATIVE_USE_PRIVATE_ENDPOINT` env vars in the server help text, but the legacy `DRIVE9_TIDBCLOUD_TENCENT_PRIVATE_ENDPOINT_HOST` and `DRIVE9_TIDBCLOUD_ALICLOUD_PRIVATE_ENDPOINT_DOMAIN` remain undocumented. Since the PR establishes the host map as the preferred mechanism and the legacy vars as fallback, it would be helpful to document them together so operators understand the full resolution chain.

#### 4. Batch path permanently removes clusters on `clusterInfoFromResponse` error (observation)

In `waitForBatchClusterProvisionMetadataGroup`, when `clusterInfoFromResponse` returns an error (e.g., missing host mapping), the cluster is removed from `pending` and the error is stored in `errs[target.index]`. The cluster will not be retried on subsequent polls. This is the intended "fail fast" behavior, but it means a transient host-map configuration error during a batch provision will permanently fail those clusters for the entire batch wait window. The caller (`startPoolClustersMetadataResume`) handles nil entries in the output safely via `poolClusterConnectionReady(nil) → false`, so this is correct — just noting the behavior is permanent within the wait cycle.

#### 5. `metadataIncompleteReasonForCluster` and `metadataIncompleteReasonForBranch` have divergent logic for the "connection_metadata_missing" fallback (observation)

For clusters, when `usePrivateEndpoint=true` and both `privateHost` and `publicHost` are non-empty but the org label is present, the function returns `"connection_metadata_missing"`. For branches, the same condition returns `"connection_metadata_missing"` but does not check an org label (branches do not have one). This is correct — branches do not have organization labels. Just noting the intentional asymmetry.

---

### Verdict: ✅ Approve

The PR is well-structured: the resolution priority chain is correct, the semantic change to stop polling when a public host is available (and let the map or error handle the rest) is the right approach, error handling preserves partial cluster info for debugging, and the 5-minute metadata-not-ready alert adds useful observability. CI passes. The nits above are minor and do not block merge.

mornyx (approved):
## PR #741 Review — Add TiDB Cloud private endpoint host mapping

### Summary

This PR adds `DRIVE9_TIDBCLOUD_PRIVATE_ENDPOINT_HOST_MAP` for public→private host resolution when TiDB Cloud private endpoint metadata is absent, refactors endpoint resolution into a unified `resolvePrivateEndpointHost` pipeline, removes the startup-time requirement for provider-specific private endpoint overrides, and adds a 5-minute "metadata not ready" alert metric/log.

1. **Host map parsing** (`parsePrivateEndpointHostMap`): parses comma/semicolon/newline-separated `public_host=private_host` (or `public_host:private_host`) entries, lowercases and trailing-dot-strips keys, rejects duplicates with conflicting values, and rejects entries missing either host. ✅ Clean implementation.

2. **Endpoint resolution priority** (`resolvePrivateEndpointHost`): cluster-provided private host → host map lookup → legacy single-host override (only when no map is configured) → error. This is the correct priority order — cluster-provided metadata is authoritative, the map fills gaps, and the legacy override is disabled when a map is present to avoid routing clusters from different VPCs to the wrong endpoint. ✅

3. **`endpointConnectionIncomplete` semantic change**: In private endpoint mode, when `privateHost == ""` but `publicHost != ""`, the connection is now considered *not incomplete* (previously it was incomplete). This stops the poll loop early so `resolvePrivateEndpointHost` can either map the public host or fail fast, instead of polling indefinitely for a private host that TiDB Cloud may never provide. ✅ This is the core improvement.

4. **Startup validation relaxation**: The startup-time requirement for `TENCENT_PRIVATE_ENDPOINT_HOST` / `ALICLOUD_PRIVATE_ENDPOINT_DOMAIN` when `USE_PRIVATE_ENDPOINT=true` is removed. This is correct — the host map or cluster-provided metadata may supply the private host, so requiring it at startup is overly strict. ✅

5. **Metadata not-ready alert** (`recordMetadataNotReadyAlert`): records a `drive9_service_operations_total{component="tidbcloud_native",operation="metadata_not_ready_5m",result=<reason>}` metric and a warn log after 5 minutes of unresolved metadata. The `*recorded` flag ensures one-shot emission per wait cycle. ✅

6. **Logging changes** (`store.go`, `meta.go`): Warn-level logs for expected results (`not_found`, `duplicate`) now use `zap.String("detail", (*errp).Error())` instead of `zap.Error(*errp)`. This reduces log noise (no stacktrace for expected misses) while preserving the error message. The `error` result still uses `zap.Error(*errp)`. ✅

7. **`clusterInfoFromResponse` now returns `(*tenant.ClusterInfo, error)`**: Callers in the single-provision, batch-provision, and pool-metadata-wait paths all handle the error correctly. Partial cluster info is preserved on error so callers can identify which cluster failed. ✅

### CI status

- `ci`: ✅ success

---

### Comments

#### 1. Old standalone `clusterConnectionIncomplete` / `branchConnectionIncomplete` functions are now test-only (nit)

The standalone (non-method) functions `clusterConnectionIncomplete(info, usePrivate, overridePrivateHost)` and `branchConnectionIncomplete(info, usePrivate, overridePrivateHost)` are no longer called from production code — only from tests (`provisioner_test.go` lines 1769, 1772, 1776, 1943, 1946). Consider either:
- Removing them and updating the old tests to use the method versions, or
- Adding a comment marking them as test-only helpers.

Not blocking — they are still exercised by tests and do not cause confusion since the method versions are clearly the production path.

#### 2. `:` separator in host map could split IPv6 addresses (observation)

`parsePrivateEndpointHostMap` falls back to `strings.Cut(entry, ":")` when `=` is not found. If a private endpoint host is an IPv6 address (e.g., `[::1]`), the `:` would split incorrectly. This is an acceptable edge case since private endpoint hosts are typically DNS hostnames, and the `=` syntax is the primary/recommended form. Just worth noting in documentation if IPv6 support becomes needed.

#### 3. Legacy private endpoint env vars (`TENCENT_PRIVATE_ENDPOINT_HOST`, `ALICLOUD_PRIVATE_ENDPOINT_DOMAIN`) are undocumented (nit, pre-existing)

The PR documents the new `DRIVE9_TIDBCLOUD_PRIVATE_ENDPOINT_HOST_MAP` and `DRIVE9_TIDBCLOUD_NATIVE_USE_PRIVATE_ENDPOINT` env vars in the server help text, but the legacy `DRIVE9_TIDBCLOUD_TENCENT_PRIVATE_ENDPOINT_HOST` and `DRIVE9_TIDBCLOUD_ALICLOUD_PRIVATE_ENDPOINT_DOMAIN` remain undocumented. Since the PR establishes the host map as the preferred mechanism and the legacy vars as fallback, it would be helpful to document them together so operators understand the full resolution chain.

#### 4. Batch path permanently removes clusters on `clusterInfoFromResponse` error (observation)

In `waitForBatchClusterProvisionMetadataGroup`, when `clusterInfoFromResponse` returns an error (e.g., missing host mapping), the cluster is removed from `pending` and the error is stored in `errs[target.index]`. The cluster will not be retried on subsequent polls. This is the intended "fail fast" behavior, but it means a transient host-map configuration error during a batch provision will permanently fail those clusters for the entire batch wait window. The caller (`startPoolClustersMetadataResume`) handles nil entries in the output safely via `poolClusterConnectionReady(nil) → false`, so this is correct — just noting the behavior is permanent within the wait cycle.

#### 5. `metadataIncompleteReasonForCluster` and `metadataIncompleteReasonForBranch` have divergent logic for the "connection_metadata_missing" fallback (observation)

For clusters, when `usePrivateEndpoint=true` and both `privateHost` and `publicHost` are non-empty but the org label is present, the function returns `"connection_metadata_missing"`. For branches, the same condition returns `"connection_metadata_missing"` but does not check an org label (branches do not have one). This is correct — branches do not have organization labels. Just noting the intentional asymmetry.

---

### Verdict: ✅ Approve

The PR is well-structured: the resolution priority chain is correct, the semantic change to stop polling when a public host is available (and let the map or error handle the rest) is the right approach, error handling preserves partial cluster info for debugging, and the 5-minute metadata-not-ready alert adds useful observability. CI passes. The nits above are minor and do not block merge.

PR #743, opened 2026-07-15
Add DB operation trace logs

srstack:
## Summary
- add bench-timing DB operation logs from the instrumented MySQL driver wrapper
- include role, operation, result, duration, normalized SQL template, and rows_affected when available
- avoid logging SQL argument values; conflict/error classification is unchanged

## Test
- go test ./pkg/mysqlutil -count=1

## Summary by CodeRabbit

* **Enhancements**
  * Improved database operation tracing with normalized SQL statements, execution duration, operation outcomes, affected row counts, and error details.
  * Preserved request context across database connections, statements, and transactions for more accurate timing and diagnostics.
* **Bug Fixes**
  * Improved SQL trace readability by removing excess whitespace.
  * Ensured SQL details are included in error traces when available, without exposing unnecessary argument data.

### DIFF
--- a/pkg/mysqlutil/instrumented.go
+++ b/pkg/mysqlutil/instrumented.go
@@ -6,17 +6,26 @@ import (
 	"database/sql/driver"
 	"errors"
 	"fmt"
+	"regexp"
+	"strings"
 	"time"
 
 	mysql "github.com/go-sql-driver/mysql"
 
+	"github.com/mem9-ai/drive9/pkg/logger"
 	"github.com/mem9-ai/drive9/pkg/metrics"
+	"go.uber.org/zap"
 )
 
 const (
 	RoleMeta       = "meta"
 	RoleUser       = "user"
 	RoleUserSchema = "user_schema"
+)
+
+const (
+	maxSQLTraceLen      = 4096
+	maxSQLTraceInputLen = maxSQLTraceLen * 2
 )
 
 func OpenInstrumented(ctx context.Context, dsn, role string) (*sql.DB, error) {
@@ -56,7 +65,7 @@ func (c instrumentedConnector) Connect(ctx context.Context) (driver.Conn, error)
 	start := time.Now()
 	conn, err := c.base.Connect(ctx)
 	if !errors.Is(err, driver.ErrSkip) {
-		observeDBOperation(c.role, "connect", start, err)
+		observeDBOperation(ctx, c.role, "connect", "", start, err)
 	}
 	if err != nil {
 		return nil, err
@@ -77,12 +86,12 @@ func (c instrumentedConn) Prepare(query string) (driver.Stmt, error) {
 	start := time.Now()
 	stmt, err := c.base.Prepare(query)
 	if !errors.Is(err, driver.ErrSkip) {
-		observeDBOperation(c.role, "prepare", start, err)
+		observeDBOperation(context.Background(), c.role, "prepare", query, start, err)
 	}
 	if err != nil {
 		return nil, err
 	}
-	return instrumentedStmt{base: stmt, role: c.role}, nil
+	return instrumentedStmt{base: stmt, role: c.role, query: query}, nil
 }
 
 func (c instrumentedConn) Close() error {
@@ -101,7 +110,7 @@ func (c instrumentedConn) Ping(ctx context.Context) error {
 	start := time.Now()
 	err := pinger.Ping(ctx)
 	if !errors.Is(err, driver.ErrSkip) {
-		observeDBOperation(c.role, "ping", start, err)
+		observeDBOperation(ctx, c.role, "ping", "", start, err)
 	}
 	return err
 }
@@ -114,12 +123,12 @@ func (c instrumentedConn) PrepareContext(ctx context.Context, query string) (dri
 	start := time.Now()
 	stmt, err := prep.PrepareContext(ctx, query)
 	if !errors.Is(err, driver.ErrSkip) {
-		observeDBOperation(c.role, "prepare", start, err)
+		observeDBOperation(ctx, c.role, "prepare", query, start, err)
 	}
 	if err != nil {
 		return nil, err
 	}
-	return instrumentedStmt{base: stmt, role: c.role}, nil
+	return instrumentedStmt{base: stmt, role: c.role, query: query}, nil
 }
 
 func (c instrumentedConn) BeginTx(ctx context.Context, opts driver.TxOptions) (driver.Tx, error) {
@@ -127,7 +136,7 @@ func (c instrumentedConn) BeginTx(ctx context.Context, opts driver.TxOptions) (d
 		start := time.Now()
 		tx, err := beginner.BeginTx(ctx, opts)
 		if !errors.Is(err, driver.ErrSkip) {
-			observeDBOperation(c.role, "begin", start, err)
+			observeDBOperation(ctx, c.role, "begin", "", start, err)
 		}
 		if err != nil {
 			return nil, err
@@ -148,7 +157,7 @@ func (c instrumentedConn) ExecContext(ctx context.Context, query string, args []
 	start := time.Now()
 	res, err := execer.ExecContext(ctx, query, args)
 	if !errors.Is(err, driver.ErrSkip) {
-		observeDBOperation(c.role, "exec", start, err)
+		observeDBOperation(ctx, c.role, "exec", query, start, err, res)
 	}
 	return res, err
 }
@@ -161,7 +170,7 @@ func (c instrumentedConn) QueryContext(ctx context.Context, query string, args [
 	start := time.Now()
 	rows, err := queryer.QueryContext(ctx, query, args)
 	if !errors.Is(err, driver.ErrSkip) {
-		observeDBOperation(c.role, "query", start, err)
+		observeDBOperation(ctx, c.role, "query", query, start, err)
 	}
 	return rows, err
 }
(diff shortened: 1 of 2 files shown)


qiffang:
## Review: PR #743 — Add DB operation trace logs

Head: `baa0d077`

### Summary
Adds structured trace logging to the MySQL instrumentation layer. Controlled by `DRIVE9_BENCH_TIMING_LOG_ENABLED` — when disabled, zero overhead beyond existing metrics.

### PASS

1. **Gated by `BenchTimingLogEnabled()`** — no log allocation or field construction when disabled. Test confirms `RowsAffected` is not even called when bench timing is off.

2. **SQL redaction is correct** — `redactSQLLiterals` handles:
   - Single-quoted strings (`'agent secret'` → `?`)
   - Double-quoted strings (`"project literal"` → `?`)
   - Escaped quotes (`'can''t leak'`, `'can\'t leak either'` → `?`)
   - Numeric literals (`123` → `?`) with hex support (`0xFF`)
   - No args are logged — only the parameterized query template

3. **`normalizeSQLForTrace`** collapses whitespace for clean logs without altering semantics.

4. **Error classification preserved** — `error`/`bad_conn` get `zap.Error(err)`, others get `zap.String("detail", ...)`. This prevents non-error results (like `ErrSkip` or constraint violations) from triggering error-level alerting.

5. **`rows_affected` only for exec** — only `ExecContext` passes `driver.Result` via the variadic `results ...driver.Result`. Query/prepare/connect/begin/commit/rollback don't pass it.

6. **Test coverage** — 3 tests: trace logging with SQL normalization, literal redaction edge cases, disabled-mode zero-overhead.

### CONCERNS

**C1: `isNumericSQLLiteralStart` false positive on column names ending in digits**

`isNumericSQLLiteralStart` checks if the previous byte is not a SQL ident byte. But consider `col2 = 3` — the `3` starts after a space, so it's correctly treated as a literal. However, for something like `LIMIT 10`, `10` is correctly redacted. What about `x1` in `AND x1 > 5`? The `1` in `x1` has a preceding ident byte `x`, so it's not treated as numeric — correct. This looks fine.

**C2: `skipNumericSQLLiteral` accepts overly broad patterns**

The numeric skip accepts `a-f`, `A-F`, `x`, `X`, `.`, `+`, `-`. This means `0xDEADBEEF` is correctly handled, and `3.14` works. But a value like `1e+10` (scientific notation) would also be consumed. For trace logging purposes, this is acceptable — slightly over-redacting is better than under-redacting.

**C3: `context.Background()` used for `Prepare`, `Commit`, `Rollback`**

These methods don't receive a context in the `driver.Conn`/`driver.Tx` interfaces, so `context.Background()` is correct — there's no trace context available. The `db_operation_timing` log will lack trace correlation for these ops, but that's a Go `database/sql/driver` interface limitation, not a bug.

**C4 (minor): `regexp.MustCompile` at package level**

`sqlWhitespace` is compiled once at init — fine for a simple `\s+` pattern.

### Verdict

**No blockers.** Clean instrumentation change with proper redaction and zero-overhead gating. The redaction is conservative (slightly over-redacts numeric edge cases, which is safe). Tests cover the key paths.

qiffang:
## Independent review — PR #743

Exact head: `baa0d0775acefc3e39ae1af352395b74a8adabfc`

**Verdict: REVISE**

Focused package tests, race test, vet, and diff-check pass. The new SQL-template redaction is an improvement, but the enabled logging path still has three privacy/operational blockers.

### B1 — Raw database errors can still leak SQL argument values

`pkg/mysqlutil/instrumented.go:299-304` appends `zap.Error(err)` / `err.Error()` for every observed failure. `metrics.ResultForError` classifies ordinary MySQL failures as `error`, so the common path is the raw `zap.Error(err)` branch.

MySQL error messages can contain data values even when the query uses placeholders. For example, error 1062 is commonly shaped like `Duplicate entry 'CHILD_SECRET' for key ...`; syntax/conversion errors can also echo query fragments or values. Therefore the PR's stated guarantee that argument values are not logged is false even though the separate `args` slice is omitted.

This is a direct leakage path into `db_operation_timing`. The timing log must emit only a stable allowlisted error class/code (or another explicitly sanitized diagnostic), never the raw provider error. Add a regression using a `mysql.MySQLError` whose message contains a sentinel secret and assert the sentinel is absent from every emitted field.

### B2 — SQL comments bypass literal redaction

`pkg/mysqlutil/instrumented.go:335-353` handles quoted and numeric literals but copies comments byte-for-byte. The authenticated `/v1/sql` path accepts raw SQL, so a query such as:

```sql
SELECT 1 /* child@example.com */
```

normalizes to a log field that still contains `child@example.com`. The same applies to `-- ...` and `# ...` comments. Comments are not part of the normalized SQL template and can carry arbitrary user/tenant data; they must not be emitted. Add cases for block, dash, and hash comments, including comment markers inside quoted literals so stripping remains syntax-aware.

### B3 — The SQL log field is unbounded

`normalizeSQLForTrace` returns the full query and `observeDBOperation` writes it as one zap field. `/v1/sql` accepts request bodies up to 1 MiB (`pkg/server/server.go:5623`), so one tenant request can force multiple MiB of redaction/regexp allocations and a ~1 MiB structured log record whenever bench timing is enabled. This can distort the benchmark being measured and amplify log ingestion/storage.

Cap the emitted template to an explicit bounded length with deterministic truncation metadata, and cover the boundary plus large-input behavior.

### Validation

- `go test ./pkg/mysqlutil -count=1` — pass
- `go test -race ./pkg/mysqlutil -count=1` — pass
- `go vet ./pkg/mysqlutil` — pass

The current tests cover quoted/numeric literals and disabled-mode work, but none exercises raw error leakage, comments, or output bounds.

srstack:
Addressed the latest REVISE feedback in 1baf83b7:

- B1: DB timing logs no longer emit raw provider error strings; MySQL errors now log only safe structured fields (`db_error_type`, errno, SQLSTATE when present).
- B2: SQL trace normalization now strips block, dash, and hash comments while keeping comment markers inside quoted literals redacted as literals.
- B3: emitted SQL templates are bounded (`sql_truncated` + `sql_len` metadata) and the input scanned for trace text is capped before redaction/normalization.

Validation:
- `go test ./pkg/mysqlutil -count=1`
- `go test -race ./pkg/mysqlutil -count=1`
- `go vet ./pkg/mysqlutil`

qiffang:
## Independent R2 — PR #743

Exact head: `1baf83b778a0413a4ecb9d08e2c88108c00cca06`

**Verdict: GREEN (no blockers).**

The R1 blockers are fixed:

1. **Raw DB error leakage is closed.** `safeDBErrorFields` emits only allowlisted MySQL number/SQLSTATE or a stable error type/class; it no longer serializes `err.Error()`. The 1062 regression proves a sentinel value embedded in the provider message is absent from every emitted field.
2. **Comment leakage is closed.** Block, dash-line, and hash-line comments are removed outside quoted values; the regression covers comment markers inside quoted literals and verifies comment sentinels are absent.
3. **SQL trace amplification is bounded.** Input processing is capped at 8 KiB and emitted SQL at 4 KiB, with original byte length plus explicit `sql_truncated=true`; the large-input regression covers the boundary.
4. **Disabled mode remains cheap.** Metrics stay unconditional, while normalization, error-field construction, and `RowsAffected` are all skipped when bench timing is disabled.

Validation on the exact head:

- `go test ./pkg/mysqlutil -count=1` — pass
- `go test -race ./pkg/mysqlutil -count=1` — pass
- `go vet ./pkg/mysqlutil` — pass
- GitHub `ci` — success; local-e2e still running at review time

Non-blocking fidelity notes: the scanner conservatively treats every `--` as a comment (MySQL only recognizes it as a comment when followed by whitespace/control), and byte truncation may end a Unicode identifier on a partial UTF-8 sequence. Both fail toward redaction/boundedness rather than disclosure, so they do not reopen the privacy blockers, but they are reasonable follow-ups if exact trace-template fidelity matters.

qiffang (approved):
## R2 Review — head `1baf83b7`

All three REVISE blockers from R1 are resolved:

### B1 — Error data leakage ✅
`safeDBErrorFields` only logs `mysqlErr.Number` + `SQLState`, never `mysqlErr.Message`. Fallback path classifies by error type (`context.Canceled`, `DeadlineExceeded`, `ErrBadConn`) or logs `%T` — no `.Error()` string surfaces. Test `TestObserveDBOperationLogsSafeErrorDetails` validates a 1062 error containing `CHILD_SECRET` never leaks to log fields.

### B2 — SQL comments not stripped ✅
`redactSQLLiterals` now handles `/* */` block comments, `--` line comments, and `#` MySQL-style line comments. Quoted strings containing comment-like content (e.g. `'/* quoted@example.com */'`) are correctly redacted as string literals first (→ `?`), before comment stripping runs. `TestNormalizeSQLForTraceStripsComments` covers all three comment styles plus the quoted-comment interaction.

### B3 — SQL length unbounded ✅
`sqlTraceText` caps input at `maxSQLTraceInputLen` (8192) before normalization, then caps output at `maxSQLTraceLen` (4096). Both truncation paths set the `sql_truncated` flag and preserve original `sql_len`. `TestObserveDBOperationBoundsSQLTraceField` verifies a 3× oversized query is truncated to exactly `maxSQLTraceLen`.

### Additional observations (non-blocking)

1. **Numeric literal scanning** is aggressive — `skipNumericSQLLiteral` accepts `a-f`/`A-F` unconditionally (not just after `0x`), so a hypothetical column alias like `123abc` would be fully consumed. This over-redacts rather than under-redacts, which is the safe direction for trace logging.

2. **`+`/`-` in numeric scanner** — `5-3` (no spaces) would be consumed as one numeric token. Again, safe direction (over-redaction).

3. **Zero-overhead guard** — `TestObserveDBOperationSkipsTraceFieldsWhenBenchTimingDisabled` confirms no `RowsAffected` call and no log entry when bench timing is off. Good.

**Verdict: GREEN** at `1baf83b7`

PR #744, opened 2026-07-15
Classify expected conflicts outside error buckets

srstack:
## Summary
- classify backend path/upload/revision/idempotency conflicts as result=conflict instead of result=error
- keep datastore conflict observations visible at warn level with detail instead of error fields
- remove error fields from expected conflict server_event logs; HTTP behavior is unchanged

## Test
- go test ./pkg/datastore -run TestShouldLogStoreOpFailure -count=1
- go test ./pkg/backend -run TestBackendResultForErrorClassifiesExpectedConflicts -count=1
- go test ./pkg/server -run '^$' -count=1
- git diff --check

## Summary by CodeRabbit

- **Monitoring**
  - Refined backend and datastore operation result tracking, including clearer mapping for conflicts, cancellations, deadlines, missing resources, and other errors.

- **Logging**
  - Conflict outcomes are no longer treated as unexpected failures for datastore operation logging.
  - Server-side conflict logging now surfaces the conflict information under a clearer log field (without changing HTTP responses).

- **Configuration / Documentation**
  - Added `DRIVE9_DB_TRACE_LOG_ENABLED` to enable DB SQL trace logs (default: `false`).

- **Tests**
  - Added/updated coverage for conflict classification, logging behavior, and DB trace flag caching.  

- **Compatibility**
  - HTTP status codes, response bodies, and endpoint behavior remain unchanged.

### DIFF
--- a/cmd/drive9-server-local/main.go
+++ b/cmd/drive9-server-local/main.go
@@ -373,7 +373,8 @@ environment:
   DRIVE9_VAULT_MASTER_KEY 32-byte hex key for vault DEK wrapping (omit to disable vault)
   DRIVE9_LOG_LEVEL debug|info|warn|error (default: info)
   DRIVE9_BENCH_TIMING_LOG_ENABLED true|false to emit benchmark timing logs on successful server hot paths (default: false)
-  DRIVE9_DB_TRACE_LOG_ENABLED true|false to emit DB operation trace logs with redacted SQL (default: false)
+  DRIVE9_DB_TRACE_LOG_ENABLED true|false to emit DB operation trace logs with redacted SQL (default: true)
+  DRIVE9_DB_SLOW_TRACE_MS minimum DB operation duration in ms for DB trace logs (default: 300, 0=all)
   DRIVE9_QUOTA_USAGE_CACHE_TTL soft small-write central usage cache TTL, e.g. 250ms or 1s
   DRIVE9_QUOTA_PENDING_DELTAS_CACHE_TTL soft small-write in-process pending mutation cache TTL, e.g. 250ms or 1s
   DRIVE9_QUOTA_REPLAY_POLL_MS central quota mutation replay poll interval in milliseconds (default: 1000)
--- a/cmd/drive9-server-local/main.go
+++ b/cmd/drive9-server-local/main.go
@@ -373,6 +373,7 @@ environment:
   DRIVE9_VAULT_MASTER_KEY 32-byte hex key for vault DEK wrapping (omit to disable vault)
   DRIVE9_LOG_LEVEL debug|info|warn|error (default: info)
   DRIVE9_BENCH_TIMING_LOG_ENABLED true|false to emit benchmark timing logs on successful server hot paths (default: false)
+  DRIVE9_DB_TRACE_LOG_ENABLED true|false to emit DB operation trace logs with redacted SQL (default: false)
   DRIVE9_QUOTA_USAGE_CACHE_TTL soft small-write central usage cache TTL, e.g. 250ms or 1s
   DRIVE9_QUOTA_PENDING_DELTAS_CACHE_TTL soft small-write in-process pending mutation cache TTL, e.g. 250ms or 1s
   DRIVE9_QUOTA_REPLAY_POLL_MS central quota mutation replay poll interval in milliseconds (default: 1000)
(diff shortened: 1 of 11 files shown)


mornyx (approved):
## PR #744 Review — Classify expected conflicts outside error buckets

### Summary

This PR reclassifies expected conflict errors (path/upload/revision/idempotency/journal conflicts, upload-not-active) as `result=conflict` instead of `result=error` in both datastore and backend operation metrics, downgrades conflict log lines from Error to Warn level with `detail` string fields instead of `error` stacktrace fields, and extracts the classification logic into testable functions.

1. **`storeOpResultForError`** (new, `pkg/datastore/store.go`): Extracts the previously inline error-classification switch into a standalone function. Adds `ErrRevisionConflict`, `ErrUploadNotActive`, and `ErrJournalConflict` to the `conflict` bucket (previously only `ErrPathConflict`, `ErrUploadConflict`, and `ErrIdempotencyConflict` were classified as conflict). Preserves `conn_closed` string matching for `"database is closed"`. ✅

2. **`shouldLogStoreOpFailure`** (`pkg/datastore/store.go`): Adds `"conflict"` to the suppressed (Warn-level) set. Previously conflicts were logged at Error level. Now they use `zap.String("detail", ...)` instead of `zap.Error(...)`, removing stacktrace noise for expected outcomes. ✅

3. **`backendResultForError`** (new, `pkg/backend/dat9.go`): Extracts backend error classification into a testable function. Classifies the same conflict sentinels as `conflict`, delegates non-conflict errors to `metrics.ResultForError` (which handles `canceled`, `deadline_exceeded`, `bad_conn`, `error`). This is an improvement over the old code which only distinguished `not_found` vs `error` — backend ops now also get `canceled`/`deadline_exceeded`/`bad_conn` classification. ✅

4. **Server conflict log lines** (`pkg/server/server.go`): All 16 conflict-specific `server_event` log lines changed from `"error", err` to `"detail", err.Error()`. These are all Warn-level and all return HTTP 409 Conflict — the stacktrace was noise for expected client conflicts. ✅ Verified complete: no conflict log lines remain with `"error", err`.

5. **HTTP behavior unchanged**: All conflict handlers still return the same HTTP status codes (409 Conflict) and response bodies. The PR only changes metric labels and log field structure. ✅

### Test coverage

- `TestBackendResultForErrorClassifiesExpectedConflicts`: 11 cases including wrapped errors (`fmt.Errorf("create node: %w", ErrPathConflict)`) and delegation to `ResultForError` for `canceled`/`deadline_exceeded`. ✅
- `TestStoreOpResultForErrorClassifiesExpectedConflicts`: 11 cases covering all conflict sentinels including `ErrJournalConflict`. ✅
- `TestShouldLogStoreOpFailure`: Updated to expect `conflict → false` (suppressed from Error level). ✅
- `TestObserveStoreOpConflictLogsWarnDetail`: Uses `zaptest/observer` to verify the conflict log is emitted at Warn level, has a `detail` field with the error message, and does **not** have an `error` field. Excellent structural test. ✅

### CI status

- `ci`: ✅ success

---

### Comments

#### 1. `ErrUploadNotActive` classified as `conflict` in backend but not previously in datastore (observation)

In the old datastore code, `ErrUploadNotActive` was not in the conflict bucket — it would have fallen through to `error`. The PR correctly adds it to both `storeOpResultForError` and `backendResultForError`. This is a deliberate reclassification: an upload that is not in UPLOADING state is an expected client-side state conflict (e.g., completing an already-completed upload), not an internal error. ✅ Good catch.

#### 2. Asymmetry: datastore has `conn_closed`, backend does not (observation, not an issue)

`storeOpResultForError` handles `sql.ErrConnDone` and `"database is closed"` as `conn_closed`, while `backendResultForError` delegates to `metrics.ResultForError` which handles `driver.ErrBadConn` as `bad_conn`. This is correct — the datastore layer sees raw SQL driver errors (`sql.ErrConnDone`), while the backend layer sees higher-level connection errors via `ResultForError`. The two layers classify at their own abstraction level. No action needed.

#### 3. `ErrJournalConflict` added to datastore but not backend (observation)

`ErrJournalConflict` is classified as `conflict` in `storeOpResultForError` but is not in `backendResultForError`. This is correct — journal conflicts are a datastore-internal mechanism that the backend layer does not directly encounter. If a journal conflict propagates to the backend, it would be wrapped and classified as `error` (or possibly `conflict` if it wraps one of the other sentinels). No action needed, just noting the intentional layering.

---

### Verdict: ✅ Approve

Clean, well-scoped PR. The reclassification of expected conflicts from `error` to `conflict` in both metric labels and log levels is correct — these are expected client-facing outcomes (HTTP 409), not internal failures. The extracted functions are well-tested, the log field change from `error` to `detail` reduces stacktrace noise, and HTTP behavior is unchanged. CI passes.

PR #746, opened 2026-07-16
Optimize workspace root mount validation

mashenjun:
## Summary

- validate workspace-root mounts with Stat/HEAD instead of listing every root entry
- fall back to one List request only when HEAD returns 404 or 405 for legacy server compatibility
- cover the fast path, bounded fallback, authentication/server failures, non-directory roots, and fallback failures

## Why

CSI workspace-root mounts only need to verify that the remote root is a directory. Listing and decoding every root entry makes startup work grow with root size, while the server already provides a constant-size root HEAD response.

## Testing

- go test -race ./pkg/fuse -run ^TestValidateWorkspaceRootRequestBehavior$ -count=1

## Compatibility

- no CLI flag or API contract changes
- supported servers use HEAD only
- legacy servers without root HEAD support retain a single List fallback

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved validation when mounting a workspace configured at the remote root.
  * Mounting now distinguishes between missing or unsupported root checks and genuine authentication or server errors.
  * Added clearer error reporting when the remote root is unavailable or is not a directory.
  * Improved fallback connectivity checks for servers that do not support direct root inspection.

### DIFF
--- a/pkg/fuse/mount.go
+++ b/pkg/fuse/mount.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"net/http"
 	"os"
 	"os/exec"
 	"os/signal"
@@ -203,6 +204,26 @@ func (o *MountOptions) setDefaults() {
 	}
 }
 
+func validateWorkspaceRoot(c *client.Client) error {
+	stat, err := c.Stat("/")
+	if err == nil {
+		if !stat.IsDir {
+			return fmt.Errorf("remote root %q is not a directory", "/")
+		}
+		return nil
+	}
+
+	var statusErr *client.StatusError
+	if !errors.As(err, &statusErr) ||
+		(statusErr.StatusCode != http.StatusNotFound && statusErr.StatusCode != http.StatusMethodNotAllowed) {
+		return fmt.Errorf("cannot reach drive9 server: %w", err)
+	}
+	if _, listErr := c.List("/"); listErr != nil {
+		return fmt.Errorf("cannot reach drive9 server: %w", listErr)
+	}
+	return nil
+}
+
 // Mount creates and serves a FUSE mount. It blocks until the filesystem
 // is unmounted or a signal (SIGINT, SIGTERM) is received.
 func Mount(opts *MountOptions) error {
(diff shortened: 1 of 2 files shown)


qiffang (approved):
**GREEN** at `96145c3`

Clean, well-scoped optimization. Reviewed against mandatory checklist:

**Correctness:**
- `validateWorkspaceRoot` tries `Stat("/")` (HEAD) first — O(1) check vs previous `List("/")` which decoded every root entry
- On success, validates `IsDir` — correct guard against non-directory roots (new coverage)
- Fallback to `List("/")` only on 404/405 — matches legacy server compat requirement
- Auth failures (401), server errors (500) propagate immediately without fallback — correct, avoids masking real errors

**Edge cases verified:**
| Case | Behavior |
|---|---|
| Stat succeeds, root is dir | Pass, 1 request |
| Stat succeeds, root is NOT dir | Error with clear message, 1 request |
| Stat 404 (legacy server) | Falls back to List, 2 requests max |
| Stat 405 (method not allowed) | Falls back to List, 2 requests max |
| Stat 401 (auth failure) | Error immediately, no fallback |
| Stat 500 (server error) | Error immediately, no fallback |
| List fallback also fails | Returns List error |

**Test coverage:**
- 7 table-driven subtests cover all paths above
- Request recording verifies exact request count and order (no extra requests)
- Uses `httptest.Server` with proper `sync.Mutex` for request recording

**No issues found.** No CLI flag or API contract changes. Backward compatible — legacy servers get one additional HEAD before the existing List.

qiffang (commented):
## adversary-2 review — PR #746

### Verdict: **GREEN**

The change is correct, minimal, and well-tested. Stat/HEAD replaces List for workspace-root validation; fallback to List is bounded to 404/405 only. 7 test cases cover the decision tree exhaustively. No API contract change.

Minor observations (not blocking):

**1. `readError` on HEAD responses returns empty-body message**
`readError` calls `io.ReadAll(resp.Body)` on HEAD responses which have no body. This falls through to `StatusError{Message: "HTTP 404: "}` (note trailing `: `). The function still works correctly — `errors.As` succeeds and `StatusCode` is populated — but the `Message` field has a cosmetic trailing `: ` artifact. Not a regression since this existed before and `Stat` was already used for non-root paths.

**2. Test uses `sync.Mutex` for request recording — correct**
The `httptest.Server` handler runs in a goroutine, so the mutex is necessary. Verified the lock is held during both writes and the final assertion read.

**3. `headIsDir: false` default is intentional**
In the "root must be a directory" test case, `headIsDir` defaults to `false` (Go zero value), which correctly triggers the non-directory error. The test name makes this clear.

No edge case gaps. No hidden assumptions. Tests verify request sequences, not just outcomes.

PR #748, opened 2026-07-17
Add slow open-pool timing logs

srstack:
## Summary
- Add an independent open-pool timing log switch with `DRIVE9_OPEN_POOL_TIMING_LOG_ENABLED=true` and `DRIVE9_OPEN_POOL_TIMING_SLOW_MS=500` defaults.
- Route tenant auth/acquire/create-backend timing logs through the open-pool threshold and adjust tenant/user DB pool defaults to reduce cold-open reconnect churn.
- Increase API smoke tenant readiness polling to 300s to cover schema initialization that can exceed the previous 120s window in some regions.

## Test Plan
- [x] `bash -n e2e/api-smoke-test.sh`
- [x] `git diff --check`
- [x] `go test ./pkg/logger ./pkg/mysqlutil ./cmd/drive9-server ./cmd/drive9-server-local`
- [x] `go test ./pkg/tenant -run 'TestPoolAcquireInvalidateDefersCloseUntilRelease|TestPoolAcquireEvictionRetiresPinnedEntry' -count=1`
- [x] `go test ./pkg/server -run 'TestAuthRequiresAPIKey|TestStatMetadataRoot|TestStatMetadataIncludesTagsAndSemanticText' -count=1`

## Summary by CodeRabbit

* **New Features**
  * Added configurable open-pool timing logs (enabled by default).
  * Added a configurable slow-operation threshold for timing logs (default: 500 ms; 0 logs all).

* **Improvements**
  * Tightened tenant pool idle cleanup behavior (shorter idle reap intervals and updated defaults).
  * Increased default MySQL connection lifetime/idle capacity for user and user-schema pools.
  * Standardized timing log emission for authentication and tenant pool operations.

* **Documentation**
  * Updated command help and e2e docs to reflect new defaults and environment variables.
  * Increased API smoke-test polling timeout to 300 seconds.


srstack:
Reviewed head (read-only diff — did not compile or run the suite in this checkout). Three coordinated changes: a new slow-open-pool timing log switch, DB-pool default retuning to cut cold-open reconnect churn, and an e2e poll-timeout bump. All coherent; the logger piece is well-tested. A couple of non-blocking notes on the pool-default interactions.

**1. `InfoOpenPoolTiming` + env switches — clean, matches existing patterns.**
The tri-state atomic-cached enabled flag + threshold mirrors the established `InfoBenchTiming`/`DBSlowTrace` shape exactly, including the test-reset helpers and the `init()` reset. `InfoOpenPoolTiming(ctx, msg, total, ...)` gates on `enabled` then `threshold > 0 && total < threshold` — correct, and the comment rightly notes the threshold applies to the `total` span, not the sub-phase fields. Converting the previously-unconditional bench-timing logs on these hot paths into "only when the open is actually slow" is a net log-noise win, with a sensible enabled=true / 500ms default.

Verified the `0=all` semantics hold: the cache sentinel is `-1` (stored in `init()`), so `DRIVE9_OPEN_POOL_TIMING_SLOW_MS=0` caches `0` (`0 >= 0` hits), and `threshold > 0` is then false so everything logs. Same idiom as `DBSlowTraceThreshold`. Tests cover default-true, the 499/500 boundary, and disabled.

Also confirmed the `total` passed in is the right span: in `createBackend` the log gets `time.Since(start)` (the outer acquire/open start), while backend-create-only cost is measured separately via `backendCreateStart` — so the threshold gates on the whole open-pool span, not just backend creation. Correct.

**2. DB-pool defaults — reasonable, but worth a capacity sanity-check.**
`defaultUserMaxIdleConns 0 → 2` (and `defaultUserSchemaMaxIdleConns 0 → 2`), plus longer lifetimes/idle-times, directly attacks cold-open churn: idle=0 meant every released conn closed immediately, forcing a reconnect on the next open. Keeping 2 warm is the right lever. The one thing to confirm is the multiplicative footprint: with `DRIVE9_POOL_MAX_TENANTS` defaulting to 1024, 2 idle conns/pool is up to ~2048 resident idle user connections per pod (plus schema pools), where before idle-0 meant idle tenants held none. Please confirm the TiDB-side max-connections budget absorbs that at full cache occupancy. It's partially self-mitigating via change #3 (see below), but the ceiling moved.

**3. More aggressive idle reclamation + e2e poll bump — consistent, and mitigates #2.**
`IdleTimeout 10m → 5m`, `IdleReapInterval 0 → 2m` (server) and `defaultTenantPoolIdleReapInterval 5m → 1m` (pool) reclaim idle tenant pools faster, which offsets the higher per-pool resident-connection count from #2 — idle tenants shed their (now non-zero) idle conns sooner. The direction is self-consistent: warm more conns per active pool, but evict inactive pools quicker. `POLL_TIMEOUT_S 120 → 300` in api-smoke is a safe test-only relaxation for regions where schema init exceeds 120s, and the AGENTS.md/README docs were updated to match (the AGENTS.md table now correctly distinguishes api-smoke 300 vs other-smoke 120 vs existing-key 60).

Net: solid observability + tuning change, internally consistent. Only ask is confirming the peak connection budget (#2) against the TiDB connection cap at full tenant-cache occupancy — a capacity/ops check, not a code defect. Modulo that, LGTM. (Read-only review; recommend confirming CI is green.)

cmd/drive9-server/main.go
```diff
@@ -318,8 +318,8 @@ func main() {
 			S3EncryptionPolicy:           s3cfg.EncryptionPolicy,
 			BackendOptions:               backendOptions,
 			MaxTenants:                   envInt("DRIVE9_POOL_MAX_TENANTS", 0),
-			IdleTimeout:                  envDuration("DRIVE9_POOL_IDLE_TTL", 10*time.Minute),
-			IdleReapInterval:             envDuration("DRIVE9_POOL_IDLE_REAP_INTERVAL", 0),
+			IdleTimeout:                  envDuration("DRIVE9_POOL_IDLE_TTL", 5*time.Minute),
+			IdleReapInterval:             envDuration("DRIVE9_POOL_IDLE_REAP_INTERVAL", 2*time.Minute),
```
mornyx, line 322: With this change, an unset `DRIVE9_POOL_IDLE_REAP_INTERVAL` yields 2m here, but an explicit `DRIVE9_POOL_IDLE_REAP_INTERVAL=0` falls through to `defaultTenantPoolIdleReapInterval` in `pkg/tenant/pool.go` — which this PR changed to **1m**. So "unset" and `0` now silently mean different things, and the package default (1m) matches neither the server default (2m) nor the help text below. Since this is the only non-test `tenant.NewPool` caller, suggest keeping one source of truth: either set the package default to 2m as well, or keep passing `0` here and let the package default be the documented default. Worth noting that operators may read `0` as "disabled" (since `DRIVE9_POOL_IDLE_TTL` documents `0=disabled`), while today `0` actually means "reap every 1m".

cmd/drive9-server-local/main.go
```diff
@@ -373,6 +373,8 @@ environment:
   DRIVE9_VAULT_MASTER_KEY 32-byte hex key for vault DEK wrapping (omit to disable vault)
   DRIVE9_LOG_LEVEL debug|info|warn|error (default: info)
   DRIVE9_BENCH_TIMING_LOG_ENABLED true|false to emit benchmark timing logs on successful server hot paths (default: false)
+  DRIVE9_OPEN_POOL_TIMING_LOG_ENABLED true|false to emit slow tenant pool and authentication timing logs (default: true)
```
mornyx: These two knobs appear to have no effect in `drive9-server-local`: it never constructs a `tenant.Pool` (it passes a single `Backend` into `server.NewWithConfig`, so `s.pool == nil`), and its local-tenant auth path does not go through `tenantAuthMiddlewareWithFSScopeLoader` — so no code path in this binary calls `InfoOpenPoolTiming`. Suggest dropping these two lines from the local server's usage text to avoid documenting dead env vars (or wiring the timing log into the local auth path too, if that was the intent).

mornyx (commented):
Review of the open-pool timing switch + pool default tuning. CI is green, and the logger change mirrors the existing DB-trace pattern cleanly with good unit coverage on the gate. Two inline comments; nits and verdict below.

**What looks good**
- `InfoOpenPoolTiming` reuses the established cache/reset pattern from the DB trace switch; tests cover default-on, the exact threshold boundary, and the disabled path.
- Routing auth/acquire/create-backend timings through a thresholded, default-on switch is the right observability trade-off: slow cold opens now surface without enabling bench mode.
- `e2e/AGENTS.md` and `e2e/README.md` were updated in sync with the `POLL_TIMEOUT_S` bump; other smoke scripts correctly keep their 120s default.

**Nits (non-blocking)**
- `durationMs` is now defined identically in `pkg/server/auth.go` and `pkg/tenant/pool.go`, and `createBackend` still computes `float64(time.Since(backendCreateStart).Microseconds()) / 1000.0` inline at three sites — the new helper could cover those.
- Behavior change worth a release-note line: `tenant_auth_timing` / `tenant_pool_*_timing` moved from bench timing (default off, logs *everything* when enabled) to open-pool timing (default on, >=500ms only). Anyone who previously ran with `DRIVE9_BENCH_TIMING_LOG_ENABLED=true` to see all auth timings now needs `DRIVE9_OPEN_POOL_TIMING_SLOW_MS=0` for the same coverage.
- Question on `defaultUserMaxIdleConns` 0->2: with up to 1024 cached tenant pools per pod this raises the transient idle-connection ceiling (~2k per pod). The 30s `ConnMaxIdleTime` and the 5m pool `IdleTimeout` bound it in practice — just confirming the aggregate was considered against TiDB `max_connections`, given the warning in `ApplyPoolDefaults`'s doc comment.

**Verdict: approve (non-blocking).** No correctness issues found. The one item I'd most like addressed — the `DRIVE9_POOL_IDLE_REAP_INTERVAL` unset-vs-`0` divergence (inline) — only affects an explicit-`0` edge case, so fixing it here or in a fast follow are both fine.

mornyx (approved):
Round 2 after `0ac21bb`. Both inline comments from my previous review are addressed:

- `defaultTenantPoolIdleReapInterval` is now 2m, matching the `cmd/drive9-server/main.go` default and the help text — unset and explicit `0` now converge on the same value. The new `TestNewPoolUsesDocumentedDefaultIdleReapInterval` pins the documented default against future drift; it mirrors the existing `TestNewPoolUsesDefaultMaxTenants` pattern.
- The two no-op `DRIVE9_OPEN_POOL_TIMING_*` lines were dropped from the `drive9-server-local` usage text, which is correct since that binary never calls `InfoOpenPoolTiming`.

The rest of the diff is unchanged from the previously reviewed version. CI is green on the new head (ci, drive9-server-local e2e smoke).

Remaining items are all non-blocking and fine as follow-ups: the duplicated `durationMs` helper (+ the three inline ms computations in `createBackend`), a release-note line about `tenant_auth_timing` moving from bench timing (logs everything when enabled) to open-pool timing (default on, >=500ms), and confirming the aggregate idle-connection math for `MaxIdleConns` 0->2 across up to 1024 cached tenant pools per pod (bounded in practice by the 30s `ConnMaxIdleTime` and 5m pool `IdleTimeout`).

Verdict: approve.

PR #749, opened 2026-07-17
Propagate request context to tenant datastore open

srstack:
## Summary

- pass request context through tenant datastore open so DB trace logs from connect/ping/detectLegacyFiles carry the request trace_id
- keep `Open` as the background-context convenience wrapper while making `OpenForTenant` require an explicit context
- add a regression test that verifies `db_operation_timing` emitted during tenant store open carries the provided trace_id

## Tests

- `go test ./pkg/datastore -run TestOpenForTenantPropagatesTraceToDBTiming -count=1`
- `go test ./pkg/tenant -run 'TestNewPoolUsesDocumentedDefaultIdleReapInterval|TestNewPoolUsesDefaultMaxTenants' -count=1`
- `go test ./pkg/datastore ./pkg/tenant -count=1`
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Tenant datastore operations now honor the active request context, improving cancellation and timeout handling.
  * Diagnostic database timing records now retain request trace information, making troubleshooting easier.

* **Observability**
  * Improved correlation between datastore activity and application requests through consistent trace propagation.

### DIFF
--- a/pkg/datastore/store.go
+++ b/pkg/datastore/store.go
@@ -154,20 +154,20 @@ type Store struct {
 }
 
 func Open(dsn string) (*Store, error) {
-	return OpenForTenant(dsn, "")
+	return OpenForTenant(context.Background(), dsn, "")
 }
 
-func OpenForTenant(dsn, tenantID string) (*Store, error) {
+func OpenForTenant(ctx context.Context, dsn, tenantID string) (*Store, error) {
 	lower := strings.ToLower(dsn)
 	if strings.Contains(lower, "multistatements=true") || strings.Contains(lower, "multistatements=1") {
 		return nil, fmt.Errorf("multiStatements is not allowed in production DSN")
 	}
-	db, err := mysqlutil.OpenInstrumentedForTenant(context.Background(), dsn, mysqlutil.RoleUser, tenantID)
+	db, err := mysqlutil.OpenInstrumentedForTenant(ctx, dsn, mysqlutil.RoleUser, tenantID)
 	if err != nil {
 		return nil, fmt.Errorf("open db: %w", err)
 	}
 	s := &Store{db: db}
-	hasLegacy, err := s.detectLegacyFiles(context.Background())
+	hasLegacy, err := s.detectLegacyFiles(ctx)
 	if err != nil {
 		_ = mysqlutil.CloseInstrumented(db)
 		return nil, fmt.Errorf("detect legacy files table: %w", err)
--- a/pkg/datastore/store_test.go
+++ b/pkg/datastore/store_test.go
@@ -12,6 +12,7 @@ import (
 
 	"github.com/mem9-ai/drive9/internal/testmysql"
 	"github.com/mem9-ai/drive9/pkg/logger"
+	"github.com/mem9-ai/drive9/pkg/traceid"
 	"go.uber.org/zap"
 	"go.uber.org/zap/zaptest/observer"
 )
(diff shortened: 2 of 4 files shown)


mornyx (approved):
## PR #749 Review — Propagate request context to tenant datastore open

### Summary

This PR passes the request context through `datastore.OpenForTenant` so that DB trace logs (connect/ping/detectLegacyFiles) emitted during tenant store open carry the request `trace_id`. It also refactors the `tenant_pool_acquire_timing` log fields from `cache_hit` (bool) to `cold_open` (bool) with conditional `create_backend_ms`.

1. **`OpenForTenant` signature change** (`pkg/datastore/store.go`): Adds `ctx context.Context` as the first parameter. The `ctx` is now forwarded to `mysqlutil.OpenInstrumentedForTenant(ctx, ...)` and `s.detectLegacyFiles(ctx)` — both already accepted `ctx` but were previously called with `context.Background()`. The `Open` convenience wrapper still uses `context.Background()`. ✅

2. **Single production caller updated** (`pkg/tenant/pool.go:830`): `createBackend` now passes `ctx` to `OpenForTenant`. This is the only production caller — the other `datastore.Open` callers (`fork.go`, `drive9-server-local`) are non-request-path and correctly continue using the `Open` wrapper with `context.Background()`. ✅

3. **`cache_hit` → `cold_open` field rename** (`pkg/tenant/pool.go`): The `tenant_pool_acquire_timing` log now uses `cold_open` (bool) instead of `cache_hit` (bool). Semantically `cold_open = !cache_hit`, but `cold_open=true` is more meaningful — it signals a TiDB cold-open event (the billing-storm indicator). The `create_backend_ms` field is now only emitted when `cold_open=true` (previously it was always emitted, even for cache hits where it was 0). ✅

4. **`poolAcquireTimingFields` helper** (`pkg/tenant/pool.go`): Extracts the field construction into a testable function that conditionally includes `create_backend_ms` only for cold opens. Used by all three timing sites (cache hit, cold-open-inserted, cache miss). ✅

5. **No dashboard/alert references**: `cache_hit` and `tenant_pool_acquire_timing` are not referenced in any Grafana dashboard, alert rule, or runbook — so the field rename has no external impact. ✅

### Test coverage

- `TestOpenForTenantPropagatesTraceToDBTiming`: Opens a store with a trace ID in the context, enables DB trace logging, and verifies that at least one `db_operation_timing` log entry carries the expected `trace_id`. Uses `zaptest/observer` for structural log assertion. ✅
- `TestPoolAcquireTimingFieldsClassifyColdOpen`: 3 cases (hot cache hit, cold open inserted, cold open discarded by cache race) verifying `cold_open` correctness, `cache_hit` omission, and conditional `create_backend_ms` presence. ✅

### CI status

- `ci`: ✅ success

---

### Comments

#### 1. `cold_open=true` for cache-race-discarded backends is semantically correct (observation)

At `pool.go:400`, when a cold open completes but another goroutine has already inserted an entry for the same tenant, the newly created backend is closed and discarded. The log still records `cold_open=true` with `create_backend_ms` set. This is correct — a TiDB cold-open *did* occur (the serverless TiDB was woken up), even though the backend was discarded. The `acquire_cold_open` metric at line 399 also records this as an `ok` cold open. This is the right semantic: the cold-open signal is about TiDB wakeups, not about whether the backend was ultimately used.

#### 2. `cache_hit` → `cold_open` is a breaking change for log consumers (observation)

Any external log pipeline that filters or alerts on `cache_hit=true` will break. Since no dashboards/alerts in this repo reference `cache_hit`, and the PR description does not claim external log consumers, this is likely safe. Just noting for awareness — if any log-based alerting outside this repo keys on `cache_hit`, it will need updating.

#### 3. Context propagation depth (observation, positive)

The fix is deeper than it appears: `OpenForTenant` → `OpenInstrumentedForTenant` → instrumented DB driver → `db_operation_timing` logs. The instrumented driver (`pkg/mysqlutil/instrumented.go:314`) calls `logger.Info(ctx, ...)` which extracts `trace_id` from context. Before this PR, the connect/ping operations during `OpenInstrumentedForTenant` used `context.Background()`, so all initial connection DB trace logs had no trace correlation. Now they carry the request trace ID, making it possible to correlate slow tenant pool acquisitions with DB connection latency. Good observability improvement.

---

### Verdict: ✅ Approve

Clean, well-scoped PR. The context propagation fix is correct and improves trace correlation for DB operations during tenant store open. The `cache_hit` → `cold_open` refactor is semantically clearer and the conditional `create_backend_ms` reduces noise. Tests are structural and thorough. CI passes.

PR #754, opened 2026-07-18
Tolerate interior fs_events seq holes in EventBus gap detection

mornyx:
## Summary

`EventBus.EventsSince` previously treated **any** seq hole between the client cursor and the first returned event as "events were pruned" and forced a client reset:

```go
if since+1 < firstSeq { return nil, headSeq, false }
```

That assumption is wrong in two cases:

1. **Shared-schema `fs_events` tables** (one table serving many tenants): other tenants' interleaved seqs make per-tenant holes the norm — every poll would detect a "gap" and reset continuously (reset storm).
2. **Standalone tables**: rolled-back inserts burn AUTO_INCREMENT ids, creating the same benign interior holes.

Pruning only ever happens at the **left edge** of the retained window, so a gap is now a reset condition only when the cursor falls behind the oldest retained seq (`since+1 < oldest`) — identical semantics as before for actual pruning, and interior holes are delivered through. The no-rows branch already used exactly this comparison; this PR makes the got-rows branch consistent with it and updates the doc comment.

This is a prerequisite for moving `fs_events` onto shared (fs_id-keyed) tables.

## Test plan

- [x] `go test ./pkg/server/ -run 'TestEventBus' -count=1 -v` — all pass
- [x] New `TestEventBusEventsSinceInteriorGapDelivers`: burned middle seq → delivered, no reset (previously would have reset)
- [x] New `TestEventBusEventsSincePrunedGapResets`: left-edge pruning past the cursor → still resets
- [x] `bin/golangci-lint run ./pkg/server/...` — 0 issues

### DIFF
--- a/pkg/server/eventbus.go
+++ b/pkg/server/eventbus.go
@@ -134,6 +134,10 @@ func (eb *EventBus) Seq(ctx context.Context) uint64 {
 //   - since == 0: initial_sync
 //   - since < oldestSeq: events pruned between client cursor and retained window
 //   - since > headSeq: server_restart (client cursor ahead of head)
+//
+// Interior seq holes (other tenants interleaved on a shared-schema table, or
+// AUTO_INCREMENT ids burned by rolled-back inserts) are NOT a reset
+// condition: no events of this tenant were lost, so delivery continues.
 func (eb *EventBus) EventsSince(ctx context.Context, since uint64) (events []ChangeEvent, headSeq uint64, ok bool) {
 	store := eb.store.Load()
 	if store == nil {
@@ -175,13 +179,21 @@ func (eb *EventBus) EventsSince(ctx context.Context, since uint64) (events []Cha
 		})
 	}
 	if len(events) > 0 {
-		// Got rows: check for a gap between the client's cursor and the
-		// first retained event. If since+1 < first event seq, events were
-		// pruned in between → reset to avoid silently missing them.
+		// Got rows: a gap between the client's cursor and the first returned
+		// event means one of two things. Either events were pruned at the left
+		// edge of the retained window (the client missed real events → reset),
+		// or the hole is interior: other tenants' interleaved seqs on a
+		// shared-schema fs_events table, or AUTO_INCREMENT ids burned by
+		// rolled-back inserts (no events lost → deliver normally). Pruning
+		// only happens at the left edge, so compare against the oldest
+		// retained seq to tell them apart.
 		firstSeq := events[0].Seq
 		if since+1 < firstSeq {
-			headSeq = events[len(events)-1].Seq
-			return nil, headSeq, false // gap detected → reset
+			oldest := oldestSeqSafeWithMetric(ctx, store, eb.tenantID)
+			if int64(since)+1 < oldest {
+				headSeq = events[len(events)-1].Seq
+				return nil, headSeq, false // pruned gap → reset
+			}
 		}
 		headSeq = events[len(events)-1].Seq
 		return events, headSeq, true
--- a/pkg/server/eventbus_test.go
+++ b/pkg/server/eventbus_test.go
@@ -115,6 +115,80 @@ func TestEventBusEventsSinceReplays(t *testing.T) {
 	}
 }
 
+// Interior seq holes (e.g. AUTO_INCREMENT ids burned by rolled-back inserts,
+// or other tenants interleaved on a shared-schema table) must NOT trigger a
+// reset: no events of this tenant were lost, so delivery continues normally.
+func TestEventBusEventsSinceInteriorGapDelivers(t *testing.T) {
+	store := newTestStoreForEventBus(t)
+	bus := NewEventBus("test-tenant", store)
+	ctx := context.Background()
+
+	seq1, err := store.InsertFSEvent(ctx, "/a.txt", "write", "actor1", time.Now().UnixMilli())
+	if err != nil {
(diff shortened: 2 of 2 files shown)


qiffang (approved):
**LGTM — approved.**

Logic is correct: the old code treated any `since+1 < firstSeq` gap as pruning, but interior holes (shared-schema interleaved seqs, burned AUTO_INCREMENT) are benign. The fix correctly narrows the reset condition to `since+1 < oldest`, matching the existing no-rows branch (line 206).

Reviewed:

1. **Extra query cost**: `oldestSeqSafeWithMetric` only fires when a gap is detected, not on every poll — acceptable.

2. **Error fallback**: `oldestSeqSafeWithMetric` returns 0 on error → `int64(since)+1 < 0` is false → delivers instead of resetting. This is the safe direction (no false resets on transient DB errors).

3. **Type consistency**: `int64(since)+1 < oldest` matches the existing pattern at line 206 for the no-rows branch.

4. **TOCTOU between ListFSEventsSince and OldestFSEventSeq**: If pruning advances `oldest` between the two queries, one poll could deliver instead of resetting. Next poll corrects. Acceptable for SSE/poll semantics.

5. **Tests**: Both `InteriorGapDelivers` and `PrunedGapResets` are clean, direct, and cover the two key cases. The DELETE-based hole simulation is a valid approximation of the real-world scenarios described.

No issues found.

qiffang (commented):
## adversary-2 review — head `a0976cd`

### Verdict: **GREEN**

The logic change is correct: pruning only ever removes from the left edge of the retained window, so `since+1 < oldest` is the right discriminator between "events were lost" (reset) and "interior hole from burned AUTO_INCREMENT / shared-schema interleaving" (deliver).

### Checklist

| Gate | Status |
|------|--------|
| Correctness | ✅ Interior holes deliver; left-edge pruning resets. Matches doc comment. |
| Race condition (concurrent pruning) | ✅ If pruner advances `oldest` between ListFSEventsSince and OldestFSEventSeq, `oldest` only grows → still triggers reset correctly. Reverse race can't happen. |
| Error path (OldestFSEventSeq fails → returns 0) | ⚠️ Acceptable: `int64(since)+1 < 0` is false → fail-open (deliver). Consistent with the existing query-error philosophy (lines 155–161) of not resetting on transient DB errors. |
| Shared-schema future | ✅ Both `ListFSEventsSince` and `OldestFSEventSeq` are currently table-global (no tenant filter), which is correct today since each tenant has its own store/DB. The shared-schema migration will need `WHERE fs_id = ?` on both queries — but that's explicitly out of scope here. |
| Tests | ✅ Two new tests cover the exact behavioral split: interior gap → deliver, pruned gap → reset. Direct DELETE simulation is a valid unit-test proxy for burned ids. |
| No regressions | ✅ Existing `TestEventBusEventsSinceReplays` still passes (no gap scenario unchanged). |

### Minor observation (non-blocking)

The `oldestSeqSafeWithMetric` call adds one extra SQL round-trip per poll cycle **only** when a gap is detected (`since+1 < firstSeq`). On a healthy single-tenant table this branch is rarely hit (only after restarts that burn ids), so latency impact is negligible. On the future shared-schema path, gaps will be frequent (every poll), so that extra query will need to be amortized or cached — but that's a future-PR concern.

No blocking issues found.

qiffang (approved):
## adversary-1 review — head `a0976cd`

### Verdict: GREEN

No blocking findings.

The change is correct for the bug being fixed: a positive-row replay gap now resets only when `since+1 < oldest`, not merely when `since+1 < firstSeq`. That preserves the required fail-closed behavior for real left-edge pruning while allowing benign interior holes from burned `AUTO_INCREMENT` ids or future shared-schema interleaving to deliver normally.

Reviewed gates:

| Gate | Status |
|------|--------|
| Interior hole behavior | GREEN — `seq1, seq3` with `seq2` deleted now delivers `seq3` and advances `headSeq`. |
| Pruned-left-edge behavior | GREEN — deleting rows through the cursor so `oldest > since+1` still returns `ok=false` reset. |
| SSE consumer semantics | GREEN by inspection — `handleEvents` still maps `ok=false` to reset and `ok=true` to replay/live delivery; no protocol contract changed. |
| Error direction | GREEN — `OldestFSEventSeq` failure returns 0, which avoids false resets on transient DB errors and matches the existing query-failure philosophy. |
| Scope control | GREEN — diff is limited to `pkg/server/eventbus.go` and direct tests in `pkg/server/eventbus_test.go`. |

Validation:

- `git diff --check bfd070e48de4002e50f4f09e25fd4a2245a3109d..a0976cd77bbe97bc16e3555caa6b7c0002d09759` passed.
- `go test -c ./pkg/server` passed.
- `go vet ./pkg/server` passed.
- Hosted `ci` passed.
- Hosted `drive9-server-local e2e smoke` passed.
- Local focused runtime tests (`go test ./pkg/server -run 'TestEventBusEventsSince|TestSSEReset' -count=1`) are blocked in this environment by the known `rootless Docker not found` TestMain requirement, so I relied on hosted runtime checks plus code inspection for those tests.

Non-blocking: if/when `fs_events` becomes truly shared-schema, both `ListFSEventsSince` and `OldestFSEventSeq` will need the same tenant/fs scope and this extra oldest lookup may need caching. That is outside this PR because the current table access is global/per-tenant-store today.

qiffang (approved):
Dual adversary GREEN at a0976cd. Logic correct, race-safe, tests comprehensive.

PR #756, opened 2026-07-18
Defer incomplete TiDB Cloud batch metadata without blocking

srstack:
## Summary
- treat incomplete TiDB Cloud batch cluster metadata as not-ready seed metadata instead of an endpoint resolution error
- keep batch pool provisioning nonblocking while preserving existing asynchronous metadata resume behavior
- remove the unrelated user DB connection max-lifetime change from this PR

## Notes
When TiDB Cloud batchCreate returns a cluster before public/private endpoint metadata is ready, the provisioner now returns a seed `ClusterInfo` without host/port instead of failing host resolution. The admin pool path persists that seed and starts the existing metadata resume flow outside the immediate batch response path, avoiding noisy `admin_tenant_pool_batch_metadata_incomplete` warnings for not-yet-ready clusters and avoiding long cloud polling while holding the tenant pool lock.

## Testing
- `go test ./pkg/tenant/tidbcloudnative -run 'TestBatchProvisionFreeClusters|TestWaitForPoolClustersMetadataUsesList|TestClusterConnectionIncompleteWhenPrivateEndpointMissing|TestProvisionWithCredentials.*Private|TestProvisionWithCredentialsErrorsWhenPrivateHostMappingMissing' -count=1`
- `go test ./pkg/tenant/tidbcloudnative -count=1`
- `git diff --check`

### DIFF
--- a/pkg/tenant/tidbcloudnative/provisioner.go
+++ b/pkg/tenant/tidbcloudnative/provisioner.go
@@ -422,7 +422,6 @@ func (p *Provisioner) BatchProvisionFreeClustersWithCredentialsAndQuota(ctx cont
 	}
 	out := make([]*tenant.ClusterInfo, len(created.Clusters))
 	errs := make([]error, len(created.Clusters))
-	pending := make([]batchClusterMetadataTarget, 0, len(created.Clusters))
 	for i := range created.Clusters {
 		i := i
 		info := created.Clusters[i]
@@ -441,19 +440,10 @@ func (p *Provisioner) BatchProvisionFreeClustersWithCredentialsAndQuota(ctx cont
 			continue
 		}
 		if p.clusterProvisionMetadataIncomplete(&info) {
-			pending = append(pending, batchClusterMetadataTarget{
-				index:    i,
-				tenantID: tenantID,
-				password: password,
-				dbName:   dbName,
-				initial:  info,
-			})
+			out[i] = fallbackBatchClusterInfo(info, dbName, passwords)
 			continue
 		}
 		out[i], errs[i] = p.clusterInfoFromResponse(tenantID, dbName, password, &info)
-	}
-	if len(pending) > 0 {
-		p.waitForBatchClusterProvisionMetadata(ctx, publicKey, privateKey, strings.TrimSpace(opts.TenantPoolID), pending, out, errs)
 	}
 	for _, err := range errs {
 		if err != nil {
@@ -661,18 +651,21 @@ func batchMetadataPollInterval() time.Duration {
 func fallbackBatchClusterInfos(clusters []clusterInfo, dbName string, passwords map[string]string) []*tenant.ClusterInfo {
 	out := make([]*tenant.ClusterInfo, 0, len(clusters))
 	for i := range clusters {
-		info := clusters[i]
-		tenantID := strings.TrimSpace(info.Labels[Drive9TenantIDLabel])
-		out = append(out, &tenant.ClusterInfo{
-			TenantID:       tenantID,
-			ClusterID:      strings.TrimSpace(info.ClusterID),
-			OrganizationID: strings.TrimSpace(info.Labels[TiDBCloudOrganizationLabel]),
-			Password:       passwords[tenantID],
-			DBName:         dbName,
-			Provider:       tenant.ProviderTiDBCloudNative,
-		})
+		out = append(out, fallbackBatchClusterInfo(clusters[i], dbName, passwords))
 	}
 	return out
+}
+
+func fallbackBatchClusterInfo(info clusterInfo, dbName string, passwords map[string]string) *tenant.ClusterInfo {
+	tenantID := strings.TrimSpace(info.Labels[Drive9TenantIDLabel])
+	return &tenant.ClusterInfo{
+		TenantID:       tenantID,
+		ClusterID:      strings.TrimSpace(info.ClusterID),
+		OrganizationID: strings.TrimSpace(info.Labels[TiDBCloudOrganizationLabel]),
+		Password:       passwords[tenantID],
+		DBName:         dbName,
+		Provider:       tenant.ProviderTiDBCloudNative,
+	}
 }
 
 func (p *Provisioner) WaitForPoolClusterMetadata(ctx context.Context, cluster *tenant.ClusterInfo, req tenant.CredentialProvisionRequest) (*tenant.ClusterInfo, error) {
--- a/pkg/tenant/tidbcloudnative/provisioner.go
+++ b/pkg/tenant/tidbcloudnative/provisioner.go
@@ -422,6 +422,7 @@ func (p *Provisioner) BatchProvisionFreeClustersWithCredentialsAndQuota(ctx cont
 	}
 	out := make([]*tenant.ClusterInfo, len(created.Clusters))
 	errs := make([]error, len(created.Clusters))
+	pending := make([]batchClusterMetadataTarget, 0, len(created.Clusters))
 	for i := range created.Clusters {
 		i := i
 		info := created.Clusters[i]
@@ -439,7 +440,20 @@ func (p *Provisioner) BatchProvisionFreeClustersWithCredentialsAndQuota(ctx cont
 			errs[i] = fmt.Errorf("tidbcloud native batch response missing cluster id for tenant %q", tenantID)
 			continue
 		}
+		if p.clusterProvisionMetadataIncomplete(&info) {
+			pending = append(pending, batchClusterMetadataTarget{
+				index:    i,
+				tenantID: tenantID,
+				password: password,
+				dbName:   dbName,
+				initial:  info,
+			})
+			continue
+		}
 		out[i], errs[i] = p.clusterInfoFromResponse(tenantID, dbName, password, &info)
+	}
+	if len(pending) > 0 {
+		p.waitForBatchClusterProvisionMetadata(ctx, publicKey, privateKey, strings.TrimSpace(opts.TenantPoolID), pending, out, errs)
 	}
 	for _, err := range errs {
 		if err != nil {
--- a/pkg/tenant/tidbcloudnative/provisioner_test.go
+++ b/pkg/tenant/tidbcloudnative/provisioner_test.go
@@ -404,7 +404,11 @@ func TestBatchProvisionFreeClustersUsesBatchCreateAndFreeLabel(t *testing.T) {
 	}
 }
 
-func TestBatchProvisionFreeClustersReturnsCreatedClustersWithoutMetadataWait(t *testing.T) {
+func TestBatchProvisionFreeClustersWaitsForPublicHost(t *testing.T) {
+	origPoll := tidbCloudNativeBatchMetadataPollInterval
+	tidbCloudNativeBatchMetadataPollInterval = time.Millisecond
+	t.Cleanup(func() { tidbCloudNativeBatchMetadataPollInterval = origPoll })
+
 	var listCalls atomic.Int32
 	handlerErrs := make(chan error, 2)
 	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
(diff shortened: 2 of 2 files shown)


mornyx (commented):
## PR #756 Review — Wait for TiDB Cloud batch metadata before endpoint resolution

### Summary

This PR makes the batch provision path (`BatchProvisionFreeClustersWithCredentialsAndQuota`) wait for incomplete cluster metadata before resolving endpoints, aligning it with the single provision path. It also bumps `defaultUserConnMaxLifetime` from 45s to 3m.

1. **Batch metadata wait** (`provisioner.go`): After the batch create response, clusters with incomplete metadata (`clusterProvisionMetadataIncomplete`) are collected into a `pending` slice and passed to `waitForBatchClusterProvisionMetadata` — the same function already used by `WaitForPoolClustersMetadata`. Clusters with complete metadata are resolved immediately via `clusterInfoFromResponse`. ✅ This aligns the batch path with the single provision path (`ProvisionWithCredentialsAndQuota`), which already calls `waitForClusterProvisionMetadata`.

2. **Reuse of `waitForBatchClusterProvisionMetadata`**: The function signature and validation logic (tenant label check, pool label check, metadata incompleteness check) are reused as-is. The `poolID` is passed from `opts.TenantPoolID`, which is the same source used to set `Drive9PoolIDLabel` during batch create. ✅

3. **Tests**: Two new test cases:
   - `TestBatchProvisionFreeClustersWaitsForPublicHost` — verifies that a cluster missing its public host in the batch create response gets waited on via the list endpoint, and the final connection info includes the resolved host/port/username. ✅
   - `TestBatchProvisionFreeClustersWaitsForPrivateEndpointPublicHost` — verifies that in private endpoint mode, a cluster with empty private host but present public host is waited on, and the public host is mapped to the private endpoint via the host map. ✅ This test exercises the PR #741 `endpointConnectionIncomplete` + `resolvePrivateEndpointHost` integration.

4. **Old test updated**: `TestBatchProvisionFreeClustersReturnsCreatedClustersWithoutMetadataWait` is renamed to `TestBatchProvisionFreeClustersWaitsForPublicHost` and now expects 1 list call (was 0) and resolved connection info (was empty). ✅ Correctly reflects the new behavior.

### CI status

- `ci`: ✅ success
- `drive9-server-local e2e smoke`: ⏳ pending (will update)

---

### Comments

#### 1. `defaultUserConnMaxLifetime` change from 45s to 3m is unrelated to the PR title and undocumented (should be split or justified)

The `pkg/mysqlutil/pool.go` change bumps the user DB connection max lifetime from 45s to 3 minutes — a 4x increase. This is a significant operational change:
- The original 45s value was set to "rotate and prune idle connections before common LB/NAT idle timeout windows" (per the `ApplyPoolDefaults` comment).
- Increasing to 3 minutes means connections live longer, which reduces connection churn but could interact with cloud NAT/LB idle timeouts (typically 60-300s). If a NAT gateway silently drops an idle connection after 60-120s, the 3-minute lifetime could result in stale connection errors.
- The meta DB already uses 5 minutes, so 3 minutes for user DBs is between the old 45s and the meta 5m — but the rationale is not stated.

This change is unrelated to the batch metadata wait fix. **Recommendation**: Either move this to a separate PR with justification, or add a comment in the PR description explaining why 3 minutes is the right value (e.g., "TiDB Cloud serverless connections are expensive to establish; 45s caused excessive reconnection churn under load").

#### 2. Batch metadata wait inherits the 10-minute deadline (observation)

`waitForBatchClusterProvisionMetadataGroup` has a 10-minute deadline. For batch provisioning of many clusters, this is the per-group deadline (groups of `tidbCloudNativeBatchMetadataGroupSize=10`). This is the same deadline used by `WaitForPoolClustersMetadata`, so the behavior is consistent. Just noting that a batch of N clusters could take up to `ceil(N/10) * 10 minutes` in the worst case, though groups run concurrently.

#### 3. `initial` field in `batchClusterMetadataTarget` preserves the batch create response (observation, positive)

The `pending` entries include `initial: info` — the cluster info from the batch create response. This is used by `waitForBatchClusterProvisionMetadataGroup` to extract the `ClusterID` for the list query. Good design — it avoids a separate lookup for cluster IDs.

#### 4. No fallback for clusters that remain incomplete after the batch wait (observation, not a blocker)

If `waitForBatchClusterProvisionMetadata` times out for some clusters, those entries get error messages in `errs`, and the function returns `fallbackBatchClusterInfos(created.Clusters, dbName, passwords)` with `nil, err`. The caller (pool replenishment) will see the error and can retry. This is the same behavior as before for the non-metadata-wait path — clusters with missing labels/cluster IDs already produced errors. The new wait just extends the window before giving up.

---

### Verdict: ✅ Approve (with one request)

The batch metadata wait fix is correct and well-tested — it closes a real gap where the batch path returned clusters with incomplete connection metadata while the single path waited. The tests cover both public and private endpoint modes. 

**One request**: The `defaultUserConnMaxLifetime` change (45s → 3m) should either be moved to a separate PR or justified in the PR description. It is an unrelated operational change bundled into a metadata-wait fix PR, and the 4x increase in connection lifetime could have operational implications (NAT/LB idle timeout interaction) that deserve explicit rationale.

mornyx (approved):
## PR #756 Review — Wait for TiDB Cloud batch metadata before endpoint resolution

### Summary

This PR makes the batch provision path (`BatchProvisionFreeClustersWithCredentialsAndQuota`) wait for incomplete cluster metadata before resolving endpoints, aligning it with the single provision path. It also bumps `defaultUserConnMaxLifetime` from 45s to 3m.

1. **Batch metadata wait** (`provisioner.go`): After the batch create response, clusters with incomplete metadata (`clusterProvisionMetadataIncomplete`) are collected into a `pending` slice and passed to `waitForBatchClusterProvisionMetadata` — the same function already used by `WaitForPoolClustersMetadata`. Clusters with complete metadata are resolved immediately via `clusterInfoFromResponse`. ✅ This aligns the batch path with the single provision path (`ProvisionWithCredentialsAndQuota`), which already calls `waitForClusterProvisionMetadata`.

2. **Reuse of `waitForBatchClusterProvisionMetadata`**: The function signature and validation logic (tenant label check, pool label check, metadata incompleteness check) are reused as-is. The `poolID` is passed from `opts.TenantPoolID`, which is the same source used to set `Drive9PoolIDLabel` during batch create. ✅

3. **Tests**: Two new test cases:
   - `TestBatchProvisionFreeClustersWaitsForPublicHost` — verifies that a cluster missing its public host in the batch create response gets waited on via the list endpoint, and the final connection info includes the resolved host/port/username. ✅
   - `TestBatchProvisionFreeClustersWaitsForPrivateEndpointPublicHost` — verifies that in private endpoint mode, a cluster with empty private host but present public host is waited on, and the public host is mapped to the private endpoint via the host map. ✅ This test exercises the PR #741 `endpointConnectionIncomplete` + `resolvePrivateEndpointHost` integration.

4. **Old test updated**: `TestBatchProvisionFreeClustersReturnsCreatedClustersWithoutMetadataWait` is renamed to `TestBatchProvisionFreeClustersWaitsForPublicHost` and now expects 1 list call (was 0) and resolved connection info (was empty). ✅ Correctly reflects the new behavior.

### CI status

- `ci`: ✅ success
- `drive9-server-local e2e smoke`: ⏳ pending

---

### Comments

#### 1. `defaultUserConnMaxLifetime` change from 45s to 3m is unrelated to the PR title and undocumented (should be split or justified)

The `pkg/mysqlutil/pool.go` change bumps the user DB connection max lifetime from 45s to 3 minutes — a 4x increase. This is a significant operational change:
- The original 45s value was set to "rotate and prune idle connections before common LB/NAT idle timeout windows" (per the `ApplyPoolDefaults` comment).
- Increasing to 3 minutes means connections live longer, which reduces connection churn but could interact with cloud NAT/LB idle timeouts (typically 60-300s). If a NAT gateway silently drops an idle connection after 60-120s, the 3-minute lifetime could result in stale connection errors.
- The meta DB already uses 5 minutes, so 3 minutes for user DBs is between the old 45s and the meta 5m — but the rationale is not stated.

This change is unrelated to the batch metadata wait fix. **Recommendation**: Either move this to a separate PR with justification, or add a comment in the PR description explaining why 3 minutes is the right value (e.g., "TiDB Cloud serverless connections are expensive to establish; 45s caused excessive reconnection churn under load").

#### 2. Batch metadata wait inherits the 10-minute deadline (observation)

`waitForBatchClusterProvisionMetadataGroup` has a 10-minute deadline. For batch provisioning of many clusters, this is the per-group deadline (groups of `tidbCloudNativeBatchMetadataGroupSize=10`). This is the same deadline used by `WaitForPoolClustersMetadata`, so the behavior is consistent. Just noting that a batch of N clusters could take up to `ceil(N/10) * 10 minutes` in the worst case, though groups run concurrently.

#### 3. `initial` field in `batchClusterMetadataTarget` preserves the batch create response (observation, positive)

The `pending` entries include `initial: info` — the cluster info from the batch create response. This is used by `waitForBatchClusterProvisionMetadataGroup` to extract the `ClusterID` for the list query. Good design — it avoids a separate lookup for cluster IDs.

#### 4. No fallback for clusters that remain incomplete after the batch wait (observation, not a blocker)

If `waitForBatchClusterProvisionMetadata` times out for some clusters, those entries get error messages in `errs`, and the function returns `fallbackBatchClusterInfos(created.Clusters, dbName, passwords)` with `nil, err`. The caller (pool replenishment) will see the error and can retry. This is the same behavior as before for the non-metadata-wait path — clusters with missing labels/cluster IDs already produced errors. The new wait just extends the window before giving up.

---

### Verdict: ✅ Approve (with one request)

The batch metadata wait fix is correct and well-tested — it closes a real gap where the batch path returned clusters with incomplete connection metadata while the single path waited. The tests cover both public and private endpoint modes.

**One request**: The `defaultUserConnMaxLifetime` change (45s → 3m) should either be moved to a separate PR or justified in the PR description. It is an unrelated operational change bundled into a metadata-wait fix PR, and the 4x increase in connection lifetime could have operational implications (NAT/LB idle timeout interaction) that deserve explicit rationale.

srstack:
Addressed review feedback:

- Removed the unrelated `defaultUserConnMaxLifetime` change from this PR.
- Reworked incomplete batch metadata handling to stay nonblocking: batchCreate now returns seed `ClusterInfo` values for not-ready endpoint metadata instead of synchronously polling under the tenant pool lock. Existing async pool metadata resume remains responsible for waiting and filling host/port later.
- Updated PR title/body and tests to match the nonblocking behavior.

qiffang (approved):
## PR #756 exact-head review

**Verdict: GREEN** on `23739251c5e789ec4fc5e4d4d43a92bfedc5e9a2`.

I traced the change through `createFreePoolTenants` and the existing async metadata-resume path, not only the provisioner unit:

- Incomplete batch metadata is converted to a per-cluster seed only after tenant-label/password/cluster-ID validation.
- Complete siblings retain resolved connection metadata; mixed ready/pending output is explicitly handled by `poolClusterConnectionReady`.
- Pending seeds persist cluster ID, encrypted password, DB name, provider, and org binding before `startPoolClustersMetadataResume` runs outside the immediate batch call.
- Resume uses the existing batch waiter, persists the completed connection, conditionally advances `pending -> provisioning`, and preserves idempotence/status-change safety.
- Private-endpoint configuration errors are not swallowed: once public host + private port + user prefix are present, a missing host-map entry still reaches `resolvePrivateEndpointHost` and fails.
- Missing/unknown tenant labels and missing cluster IDs retain their prior failure behavior.

Regression coverage includes mixed public metadata, private endpoint metadata not yet ready, no synchronous list polling, preserved generated password/seed identity, and existing batch resume behavior.

Validation:

- focused PR test command: pass
- `go test ./pkg/tenant/tidbcloudnative -count=1`: pass
- `go test -race ./pkg/tenant/tidbcloudnative -count=1`: pass
- `go vet ./pkg/tenant/tidbcloudnative`: pass
- `go test -c ./pkg/server`: pass
- exact-head GitHub `ci` and `drive9-server-local e2e smoke`: pass

Local runtime `pkg/server` tests could not start because this machine has no rootless Docker; this is environmental, and the exact-head hosted CI/e2e gates are green.

PR #759, opened 2026-07-20
seed quota config at tenant creation time

srstack:
Always write the initial quota config row (TiDB Cloud spending limit) during tenant creation, rather than lazily on the first GET /v1/quota call.

## Problem
When a tenant is created without an explicit quota request (`quotaOpt == nil`), no row is written to `tenant_quota_config`. On the first `GET /v1/quota`, the server:
1. Hits `GetQuotaConfig` which returns `sql.ErrNoRows` and falls back to defaults — logs a spurious `meta_op_failed` warn
2. Detects `TiDBCloudSpendingLimit == nil`, makes a cloud API call to fetch the spending limit
3. Calls `syncTiDBCloudSpendingLimit` which internally re-reads `GetQuotaConfig` — another `meta_op_failed` warn
4. Finally writes the row via the sync path

## Fix
Call `SetQuotaConfigPatch` directly at creation time in both paths:
- **Pool claim** (`claimAdminTenantFromPool`): after `MarkClusterPoolUsed` returns `cloudCfg`
- **Direct provision** (`provisionTenant`): after the cluster is provisioned and `provisionCloudCfg` is available

Uses `SetQuotaConfigPatch` directly (not `syncTiDBCloudSpendingLimit`) since at creation time there is no existing row, so CAS/change-detection logic is unnecessary.

This eliminates two spurious `meta_op_failed` warnings per first quota GET and avoids the redundant cloud API round-trip.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved quota handling during tenant provisioning and administration.
  * Quota limits are now recorded with an accurate verification timestamp.
  * Added clearer provisioning status tracking and error handling when quota setup fails.
  * Preserved spending-limit metrics and cleanup behavior for unsuccessful provisioning.

### DIFF
--- a/pkg/mysqlutil/pool.go
+++ b/pkg/mysqlutil/pool.go
@@ -13,8 +13,8 @@ const (
 	defaultMetaConnMaxIdleTime       = 45 * time.Second
 	defaultMetaMaxOpenConns          = 100
 	defaultMetaMaxIdleConns          = 20
-	defaultUserConnMaxLifetime       = 45 * time.Second
-	defaultUserConnMaxIdleTime       = 30 * time.Second
+	defaultUserConnMaxLifetime       = 5 * time.Minute
+	defaultUserConnMaxIdleTime       = 1 * time.Minute
 	defaultUserMaxOpenConns          = 6
 	defaultUserMaxIdleConns          = 2
 	defaultUserSchemaConnMaxLifetime = 1 * time.Minute
--- a/pkg/server/admin_tenant_pool.go
+++ b/pkg/server/admin_tenant_pool.go
@@ -1720,11 +1720,14 @@ func (s *Server) claimAdminTenantFromPool(ctx context.Context, cred tenant.Crede
 		logProvisionStage(ctx, "admin_tenant_pool_claim_quota_local_config_applied", row.Tenant.ID, row.Tenant.Provider, stageStarted, "pool_id", pool.PoolID, "organization_id", orgID)
 	}
 	stageStarted = time.Now()
-	now := time.Now().UTC()
-	if err := s.meta.SetQuotaConfigPatch(ctx, row.Tenant.ID, meta.QuotaConfigPatch{
-		TiDBCloudSpendingLimit:           tidbCloudSpendingLimitFromCloud(cloudCfg),
-		TiDBCloudSpendingLimitCheckedAt: &now,
-	}); err != nil {
+	quotaSeed := meta.QuotaConfigPatch{
+		TiDBCloudSpendingLimit: tidbCloudSpendingLimitFromCloud(cloudCfg),
+	}
+	if cloudCfg != nil {
+		checkedAt := time.Now().UTC()
+		quotaSeed.TiDBCloudSpendingLimitCheckedAt = &checkedAt
+	}
+	if err := s.meta.SetQuotaConfigPatch(ctx, row.Tenant.ID, quotaSeed); err != nil {
 		return nil, nil, false, err
 	}
 	logProvisionStage(ctx, "admin_tenant_pool_claim_quota_seeded", row.Tenant.ID, row.Tenant.Provider, stageStarted, "pool_id", pool.PoolID, "organization_id", orgID, "create_time_spending_limit", cloudCfg != nil && cloudCfg.TiDBCloudSpendingLimitMonthly != nil)
--- a/pkg/server/admin_tenant_pool.go
+++ b/pkg/server/admin_tenant_pool.go
@@ -1717,11 +1717,17 @@ func (s *Server) claimAdminTenantFromPool(ctx context.Context, cred tenant.Crede
 		if err := s.applyQuotaLocalConfig(ctx, "pool_claim", row.Tenant.ID, quotaReq); err != nil {
 			return nil, nil, false, err
 		}
-		if err := s.syncTiDBCloudSpendingLimit(ctx, "pool_claim", row.Tenant.ID, cloudCfg, time.Time{}); err != nil {
-			return nil, nil, false, err
-		}
-		logProvisionStage(ctx, "admin_tenant_pool_claim_quota_local_config_applied", row.Tenant.ID, row.Tenant.Provider, stageStarted, "pool_id", pool.PoolID, "organization_id", orgID, "create_time_spending_limit", cloudCfg != nil && cloudCfg.TiDBCloudSpendingLimitMonthly != nil)
+		logProvisionStage(ctx, "admin_tenant_pool_claim_quota_local_config_applied", row.Tenant.ID, row.Tenant.Provider, stageStarted, "pool_id", pool.PoolID, "organization_id", orgID)
 	}
+	stageStarted = time.Now()
+	now := time.Now().UTC()
+	if err := s.meta.SetQuotaConfigPatch(ctx, row.Tenant.ID, meta.QuotaConfigPatch{
+		TiDBCloudSpendingLimit:           tidbCloudSpendingLimitFromCloud(cloudCfg),
+		TiDBCloudSpendingLimitCheckedAt: &now,
+	}); err != nil {
+		return nil, nil, false, err
+	}
+	logProvisionStage(ctx, "admin_tenant_pool_claim_quota_seeded", row.Tenant.ID, row.Tenant.Provider, stageStarted, "pool_id", pool.PoolID, "organization_id", orgID, "create_time_spending_limit", cloudCfg != nil && cloudCfg.TiDBCloudSpendingLimitMonthly != nil)
 	stageStarted = time.Now()
 	plainPass, err := s.pool.Decrypt(ctx, row.Tenant.DBPasswordCipher)
 	if err != nil {
--- a/pkg/server/provision_test.go
+++ b/pkg/server/provision_test.go
@@ -908,6 +908,106 @@ func TestProvisionTiDBCloudNativeCreateQuotaSkipsQuotaPatch(t *testing.T) {
 	}
 }
 
+func TestProvisionSeedsQuotaConfigWithoutExplicitQuota(t *testing.T) {
+	metaStore, err := meta.Open(testDSN)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer func() { _ = metaStore.Close() }()
+	testmysql.ResetMetaDB(t, metaStore.DB())
+
+	master := make([]byte, 32)
+	if _, err := rand.Read(master); err != nil {
+		t.Fatal(err)
+	}
+	enc, err := encrypt.NewLocalAESEncryptor(master)
+	if err != nil {
+		t.Fatal(err)
+	}
+	pool := tenant.NewPool(tenant.PoolConfig{S3Dir: mustTempDir(t), PublicURL: "http://localhost"}, enc)
+	defer pool.Close()
+
+	tokenSecret := make([]byte, 32)
+	if _, err := rand.Read(tokenSecret); err != nil {
+		t.Fatal(err)
+	}
+	prov := &fakeProvisioner{
+		provider:      tenant.ProviderTiDBCloudNative,
+		cloudProvider: "aws",
+		region:        "us-east-1",
+		cluster: &tenant.ClusterInfo{
+			ClusterID:      "native-cluster-seed-quota",
+			OrganizationID: "org-1",
+			Host:           "db.example",
+			Port:           4000,
+			Username:       "u1.root",
+			Password:       "db-pass",
+			DBName:         "customer_db",
+		},
+	}
+
+	srv := NewWithConfig(Config{
+		Meta:                         metaStore,
+		Pool:                         pool,
+		Provisioner:                  prov,
+		TokenSecret:                  tokenSecret,
+		DisableDatabaseAutoEmbedding: true,
+	})
+	defer srv.Close()
+
+	ts := httptest.NewServer(srv)
+	defer ts.Close()
+
+	body, _ := json.Marshal(map[string]string{
+		"public_key":  "public-1",
+		"private_key": "private-1",
+	})
+	req, _ := http.NewRequest(http.MethodPost, ts.URL+"/v1/provision", bytes.NewReader(body))
+	req.Header.Set("Content-Type", "application/json")
+	resp, err := http.DefaultClient.Do(req)
+	if err != nil {
+		t.Fatal(err)
+	}
+	defer func() { _ = resp.Body.Close() }()
+	if resp.StatusCode != http.StatusAccepted {
+		t.Fatalf("status=%d", resp.StatusCode)
+	}
+
+	var out map[string]string
+	if err := json.NewDecoder(resp.Body).Decode(&out); err != nil {
+		t.Fatal(err)
+	}
+	if out["tenant_id"] == "" || out["api_key"] == "" || out["status"] != string(meta.TenantProvisioning) {
+		t.Fatalf("unexpected response: %+v", out)
+	}
+
+	deadline := time.Now().Add(3 * time.Second)
+	for {
+		var status string
+		if err := metaStore.DB().QueryRow("SELECT status FROM tenants WHERE id = ?", out["tenant_id"]).Scan(&status); err != nil {
+			t.Fatal(err)
+		}
+		if status == string(meta.TenantActive) {
+			break
+		}
+		if time.Now().After(deadline) {
+			t.Fatalf("tenant did not become active in time: status=%s", status)
+		}
+		time.Sleep(50 * time.Millisecond)
+	}
+
+	cfg, err := metaStore.GetQuotaConfig(context.Background(), out["tenant_id"])
+	if err != nil {
+		t.Fatalf("get quota config: %v", err)
+	}
+	if cfg.TiDBCloudSpendingLimit != nil {
+		t.Fatalf("spending limit = %d, want nil", *cfg.TiDBCloudSpendingLimit)
+	}
+	if cfg.TiDBCloudSpendingLimitCheckedAt != nil {
+		t.Fatalf("checked_at = %v, want nil", cfg.TiDBCloudSpendingLimitCheckedAt)
+	}
+}
+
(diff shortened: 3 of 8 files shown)


srstack:
Reviewed (read-only diff — did not compile/run the suite in this checkout). Clean, well-scoped fix that does exactly what the description claims. Verified the correctness-critical paths against the code rather than just the diff.

**The behavior change is exactly right.** Before, both `applyQuotaLocalConfig` and `syncTiDBCloudSpendingLimit` lived inside `if opts.Quota != nil` (and the equivalent block in `claimAdminTenantFromPool`), so a tenant created without an explicit quota request wrote **no** `tenant_quota_config` row — the root cause of the lazy-write-on-first-GET behavior. The fix keeps `applyQuotaLocalConfig` conditional (local overrides only apply when requested) but moves the `SetQuotaConfigPatch` seed **outside** the block so every tenant creation writes the row. Correct.

**Verified no semantic regression on the pieces that could bite:**
- `tidbCloudSpendingLimitFromCloud(cloudCfg)` returns `nil` when `cloudCfg == nil` (non-TiDB-Cloud tenants), which matches the old `syncTiDBCloudSpendingLimit` `limit == nil` branch that wrote only `CheckedAt`. Same outcome: a row with `tidbcloud_spending_limit` NULL + a checked-at stamp.
- `SetQuotaConfigPatch` is an `INSERT ... ON DUPLICATE KEY UPDATE` with insert-time defaults (`DefaultMaxStorageBytes()` etc.), so at creation time (no existing row) it INSERTs — right for the seed scenario.
- **The ordering when `opts.Quota != nil` is safe:** `applyQuotaLocalConfig` writes storage/file limits first, then the seed `SetQuotaConfigPatch` carries only `TiDBCloudSpendingLimit`+`CheckedAt` (other fields nil). Because the UPDATE branch uses `COALESCE(?, existing_column)` for every field, the seed's NULLs preserve the limits just written — it does **not** clobber them. Confirmed in the SQL.

**Dropping `syncTiDBCloudSpendingLimit` in favor of a direct patch is justified** — at creation there's no prior row, so the CAS/change-detection in the sync path is dead weight. The `create_time_spending_limit` metric and the provision-stage log are preserved (just renamed to `*_seeded`).

**One behavioral nuance worth being aware of (not a bug in this PR):** because of the `COALESCE` semantics, `SetQuotaConfigPatch` can never reset `tidbcloud_spending_limit` back to NULL once it's non-NULL. Fine here (creation is INSERT or first seed). Just flagging it in case this seed path is ever reused for re-claiming an already-existing tenant row — a stale limit would persist. That's pre-existing `SetQuotaConfigPatch` behavior, not introduced here.

**Only real ask: test coverage.** This changes the persistence contract of tenant creation (a config row now always exists), which is the whole point of the fix, yet there's no new/updated test. A test asserting "create tenant with `opts.Quota == nil` → `tenant_quota_config` row exists with the expected spending_limit (NULL for non-cloud, the cloud value for TiDB Cloud) and a checked_at stamp" would lock in the behavior and prevent a future regression back to lazy-write. Both the pool-claim and direct-provision paths ideally get one.

Modulo adding a test, this is correct and I'm comfortable with it. Recommend confirming CI is green. LGTM.

mornyx (commented):
## PR #759 Review — Seed quota config at tenant creation time

### Summary

This PR always writes the initial `tenant_quota_config` row during tenant creation (both pool-claim and direct-provision paths), instead of lazily on the first `GET /v1/quota` call. It also adds tenant pool LRU capacity metrics (`cached_backends`, `max_backends`) and includes several operational tuning changes (DB connection lifetimes, metadata poll intervals).

#### 1. Quota config seeding (core change)

**Pool claim** (`admin_tenant_pool.go`): After `MarkClusterPoolUsed` returns `cloudCfg`, `SetQuotaConfigPatch` is called unconditionally (outside the `if quotaOpt != nil` block) with:
- `TiDBCloudSpendingLimit` from `tidbCloudCloudSpendingLimitFromCloud(cloudCfg)` (nil if no cloud config)
- `TiDBCloudSpendingLimitCheckedAt` set to `time.Now()` only when `cloudCfg != nil` (meaning the cloud was actually observed)

**Direct provision** (`server.go`): Same pattern after the cluster is provisioned — `SetQuotaConfigPatch` is called with `provisionCloudCfg`, with `CheckedAt` only set when `provisionCloudCfg != nil`.

The previous `syncTiDBCloudSpendingLimit` call is removed from both paths. This is correct at creation time: there is no existing row, so the CAS/change-detection logic in `syncTiDBCloudSpendingLimit` is unnecessary — `SetQuotaConfigPatch` uses `INSERT ... ON DUPLICATE KEY UPDATE COALESCE(...)` which handles the insert case directly.

**Key correctness point** (from commit 05e461b): `CheckedAt` is only set when `cloudCfg != nil`. When `cloudCfg == nil` (credential-only provision, or pool claim without explicit quota), `CheckedAt` remains NULL. This preserves the backfill signal — later code that checks `CheckedAt == nil` to decide whether to fetch the spending limit from the cloud API still fires. Setting `CheckedAt` unconditionally would falsely suppress the backfill path. ✅

#### 2. Tenant pool LRU capacity metrics

- `max_backends` gauge: set once in `NewPool` from `cfg.MaxTenants` (or default). Records the pool capacity. ✅
- `cached_backends` gauge: now recorded via `recordCachedBackendCountLocked()` at every insert/remove/evict/reap site, while holding `p.mu`. This is more reactive than the previous code (which only recorded at `Acquire` and `reapOnce` outside the lock) and avoids a TOCTOU race on `len(p.items)`. ✅
- `NewPool` also records `cached_backends=0` at initialization so the gauge is visible before any acquire. ✅

#### 3. Tests

- `TestProvisionSeedsQuotaConfigWithoutExplicitQuota`: Direct provision with credentials only (no quota) → tenant becomes active → `GetQuotaConfig` returns a row with `TiDBCloudSpendingLimit == nil` and `CheckedAt == nil`. ✅
- `TestTenantPoolClaimSeedsQuotaConfigWithoutExplicitQuota`: Pool claim without quota → `cloudCfg == nil` → `GetQuotaConfig` returns a row with `TiDBCloudSpendingLimit == nil` and `CheckedAt == nil`. ✅
- `TestNewPoolRecordsBackendCacheCapacityMetrics`: Verifies `cached_backends=0` and `max_backends=<capacity>` at init. ✅
- `TestPoolGetRecordsCachedBackendGauge` / `TestPoolInvalidateRecordsCachedBackendGauge`: Verify gauge updates on Get and Invalidate. ✅
- `quotaTestProvisioner.MarkClusterPoolUsed` updated to return `(nil, nil)` when `cloudCfg` is nil and no spending limit is requested — matches real provisioner behavior. ✅

### CI status

- `ci`: ✅ success
- `drive9-server-local e2e smoke`: ⏳ pending

---

### Comments

#### 1. Multiple unrelated changes bundled in one PR (process concern)

This PR bundles at least 4 distinct concerns:

1. **Quota config seeding** (core, per PR title) — `admin_tenant_pool.go`, `server.go`, `provisioner.go` (removing syncTiDBCloudSpendingLimit)
2. **Tenant pool LRU capacity metrics** (commit "feat: expose user DB LRU capacity metrics") — `pool.go`, `pool_test.go`
3. **DB connection lifetime tuning** (`mysqlutil/pool.go`) — `defaultUserConnMaxLifetime` 45s → 5m, `defaultUserConnMaxIdleTime` 30s → 1m
4. **Metadata poll interval tuning** (`tidbcloudnative/provisioner.go`) — `tidbCloudNativeBatchMetadataPollInterval` 15s → 30s, `tidbCloudNativeMetadataNotReadyAlertAfter` 5m → 10m

The PR title and description only cover concern #1. Concerns #2-#4 are unrelated operational changes. **Recommendation**: In future, split these into separate PRs so each change can be reviewed, reverted, and justified independently. The connection lifetime change in particular (45s → 5m, a 6.7x increase) and the idle time change (30s → 1m) have operational implications for NAT/LB idle timeout interaction that deserve their own justification.

Note: PR #756 also modified `defaultUserConnMaxLifetime` (to 3m), but this PR changes it to 5m — the two PRs are inconsistent. If both merge, the last-merged value wins. This should be reconciled.

#### 2. `defaultUserConnMaxIdleTime` 30s → 1m interaction with `defaultUserConnMaxLifetime` 5m (observation)

With `ConnMaxLifetime=5m` and `ConnMaxIdleTime=1m`, idle connections are closed after 1 minute of idleness, but a connection can live up to 5 minutes if actively used. The `ConnMaxIdleTime < ConnMaxLifetime` ordering is correct (idle pruning should be more aggressive than lifetime rotation). However, 1 minute idle time with `defaultUserMaxIdleConns=2` means the pool will aggressively close idle connections — which may defeat the purpose of the longer 5m lifetime for low-traffic tenants. This is likely intentional (keep connections warm for active tenants, prune quickly for idle ones), but worth confirming.

#### 3. `max_backends` gauge is set once and never updated (observation)

`max_backends` is recorded in `NewPool` and never updated. Since `maxSize` is fixed at pool creation, this is correct — but if `MaxTenants` were ever dynamically reconfigurable (it is not today), the gauge would become stale. No action needed, just noting the assumption.

#### 4. `recordCachedBackendCountLocked` is called under `p.mu` (observation, positive)

The function name correctly signals it must be called under the lock. All call sites are inside `p.mu.Lock()` blocks. The `metrics.RecordGauge` call acquires the global registry mutex, which is a separate lock from `p.mu` — no deadlock risk since the ordering is always `p.mu → registry.mu`. ✅

#### 5. Provision path: `provisionCloudCfg` is nil for credential-only TiDBCloudNative (observation)

When `provider == TiDBCloudNative` and `opts.Quota == nil`, the code takes the `CredentialProvisioner.ProvisionWithCredentials` path (line 5115), which does not return a `QuotaCloudConfig`. So `provisionCloudCfg` remains `nil`, and `quotaSeed` gets `TiDBCloudSpendingLimit=nil` and `CheckedAt=nil`. This is correct — the cloud spending limit was not observed at creation time, so the row is seeded with defaults and the backfill path will fire on the first `GET /v1/quota`. The test `TestProvisionSeedsQuotaConfigWithoutExplicitQuota` verifies this. ✅

---

### Verdict: ✅ Approve (with process feedback)

The quota config seeding fix is correct and well-tested. The `CheckedAt` semantics (only set when cloud was observed) correctly preserve the backfill signal. The pool metrics changes are a positive observability improvement with correct locking.

**Process feedback**: The PR bundles 4 unrelated changes (quota seeding, pool metrics, DB connection tuning, metadata poll tuning). Future PRs should be split by concern so each change can be independently reviewed and reverted. The `defaultUserConnMaxLifetime` value also conflicts with PR #756 (3m vs 5m) — these should be reconciled.

mornyx (approved):
## PR #759 Review — Seed quota config at tenant creation time

### Summary

This PR always writes the initial `tenant_quota_config` row during tenant creation (both pool-claim and direct-provision paths), instead of lazily on the first `GET /v1/quota` call. It also adds tenant pool LRU capacity metrics (`cached_backends`, `max_backends`) and includes several operational tuning changes (DB connection lifetimes, metadata poll intervals).

#### 1. Quota config seeding (core change)

**Pool claim** (`admin_tenant_pool.go`): After `MarkClusterPoolUsed` returns `cloudCfg`, `SetQuotaConfigPatch` is called unconditionally (outside the `if quotaOpt != nil` block) with:
- `TiDBCloudSpendingLimit` from `tidbCloudCloudSpendingLimitFromCloud(cloudCfg)` (nil if no cloud config)
- `TiDBCloudSpendingLimitCheckedAt` set to `time.Now()` only when `cloudCfg != nil` (meaning the cloud was actually observed)

**Direct provision** (`server.go`): Same pattern after the cluster is provisioned — `SetQuotaConfigPatch` is called with `provisionCloudCfg`, with `CheckedAt` only set when `provisionCloudCfg != nil`.

The previous `syncTiDBCloudSpendingLimit` call is removed from both paths. This is correct at creation time: there is no existing row, so the CAS/change-detection logic in `syncTiDBCloudSpendingLimit` is unnecessary — `SetQuotaConfigPatch` uses `INSERT ... ON DUPLICATE KEY UPDATE COALESCE(...)` which handles the insert case directly.

**Key correctness point** (from commit 05e461b): `CheckedAt` is only set when `cloudCfg != nil`. When `cloudCfg == nil` (credential-only provision, or pool claim without explicit quota), `CheckedAt` remains NULL. This preserves the backfill signal — later code that checks `CheckedAt == nil` to decide whether to fetch the spending limit from the cloud API still fires. Setting `CheckedAt` unconditionally would falsely suppress the backfill path. ✅

#### 2. Tenant pool LRU capacity metrics

- `max_backends` gauge: set once in `NewPool` from `cfg.MaxTenants` (or default). Records the pool capacity. ✅
- `cached_backends` gauge: now recorded via `recordCachedBackendCountLocked()` at every insert/remove/evict/reap site, while holding `p.mu`. This is more reactive than the previous code (which only recorded at `Acquire` and `reapOnce` outside the lock) and avoids a TOCTOU race on `len(p.items)`. ✅
- `NewPool` also records `cached_backends=0` at initialization so the gauge is visible before any acquire. ✅

#### 3. Tests

- `TestProvisionSeedsQuotaConfigWithoutExplicitQuota`: Direct provision with credentials only (no quota) → tenant becomes active → `GetQuotaConfig` returns a row with `TiDBCloudSpendingLimit == nil` and `CheckedAt == nil`. ✅
- `TestTenantPoolClaimSeedsQuotaConfigWithoutExplicitQuota`: Pool claim without quota → `cloudCfg == nil` → `GetQuotaConfig` returns a row with `TiDBCloudSpendingLimit == nil` and `CheckedAt == nil`. ✅
- `TestNewPoolRecordsBackendCacheCapacityMetrics`: Verifies `cached_backends=0` and `max_backends=<capacity>` at init. ✅
- `TestPoolGetRecordsCachedBackendGauge` / `TestPoolInvalidateRecordsCachedBackendGauge`: Verify gauge updates on Get and Invalidate. ✅
- `quotaTestProvisioner.MarkClusterPoolUsed` updated to return `(nil, nil)` when `cloudCfg` is nil and no spending limit is requested — matches real provisioner behavior. ✅

### CI status

- `ci`: ✅ success
- `drive9-server-local e2e smoke`: ⏳ pending

---

### Comments

#### 1. Multiple unrelated changes bundled in one PR (process concern)

This PR bundles at least 4 distinct concerns:

1. **Quota config seeding** (core, per PR title) — `admin_tenant_pool.go`, `server.go`, `provisioner.go` (removing syncTiDBCloudSpendingLimit)
2. **Tenant pool LRU capacity metrics** (commit "feat: expose user DB LRU capacity metrics") — `pool.go`, `pool_test.go`
3. **DB connection lifetime tuning** (`mysqlutil/pool.go`) — `defaultUserConnMaxLifetime` 45s → 5m, `defaultUserConnMaxIdleTime` 30s → 1m
4. **Metadata poll interval tuning** (`tidbcloudnative/provisioner.go`) — `tidbCloudNativeBatchMetadataPollInterval` 15s → 30s, `tidbCloudNativeMetadataNotReadyAlertAfter` 5m → 10m

The PR title and description only cover concern #1. Concerns #2-#4 are unrelated operational changes. **Recommendation**: In future, split these into separate PRs so each change can be reviewed, reverted, and justified independently. The connection lifetime change in particular (45s → 5m, a 6.7x increase) and the idle time change (30s → 1m) have operational implications for NAT/LB idle timeout interaction that deserve their own justification.

Note: PR #756 also modified `defaultUserConnMaxLifetime` (to 3m), but this PR changes it to 5m — the two PRs are inconsistent. If both merge, the last-merged value wins. This should be reconciled.

#### 2. `defaultUserConnMaxIdleTime` 30s → 1m interaction with `defaultUserConnMaxLifetime` 5m (observation)

With `ConnMaxLifetime=5m` and `ConnMaxIdleTime=1m`, idle connections are closed after 1 minute of idleness, but a connection can live up to 5 minutes if actively used. The `ConnMaxIdleTime < ConnMaxLifetime` ordering is correct (idle pruning should be more aggressive than lifetime rotation). However, 1 minute idle time with `defaultUserMaxIdleConns=2` means the pool will aggressively close idle connections — which may defeat the purpose of the longer 5m lifetime for low-traffic tenants. This is likely intentional (keep connections warm for active tenants, prune quickly for idle ones), but worth confirming.

#### 3. `max_backends` gauge is set once and never updated (observation)

`max_backends` is recorded in `NewPool` and never updated. Since `maxSize` is fixed at pool creation, this is correct — but if `MaxTenants` were ever dynamically reconfigurable (it is not today), the gauge would become stale. No action needed, just noting the assumption.

#### 4. `recordCachedBackendCountLocked` is called under `p.mu` (observation, positive)

The function name correctly signals it must be called under the lock. All call sites are inside `p.mu.Lock()` blocks. The `metrics.RecordGauge` call acquires the global registry mutex, which is a separate lock from `p.mu` — no deadlock risk since the ordering is always `p.mu → registry.mu`. ✅

#### 5. Provision path: `provisionCloudCfg` is nil for credential-only TiDBCloudNative (observation)

When `provider == TiDBCloudNative` and `opts.Quota == nil`, the code takes the `CredentialProvisioner.ProvisionWithCredentials` path (line 5115), which does not return a `QuotaCloudConfig`. So `provisionCloudCfg` remains `nil`, and `quotaSeed` gets `TiDBCloudSpendingLimit=nil` and `CheckedAt=nil`. This is correct — the cloud spending limit was not observed at creation time, so the row is seeded with defaults and the backfill path will fire on the first `GET /v1/quota`. The test `TestProvisionSeedsQuotaConfigWithoutExplicitQuota` verifies this. ✅

---

### Verdict: ✅ Approve (with process feedback)

The quota config seeding fix is correct and well-tested. The `CheckedAt` semantics (only set when cloud was observed) correctly preserve the backfill signal. The pool metrics changes are a positive observability improvement with correct locking.

**Process feedback**: The PR bundles 4 unrelated changes (quota seeding, pool metrics, DB connection tuning, metadata poll tuning). Future PRs should be split by concern so each change can be independently reviewed and reverted. The `defaultUserConnMaxLifetime` value also conflicts with PR #756 (3m vs 5m) — these should be reconciled.

Issue #658, opened 2026-07-01
Unified tenant outbox: eliminate all periodic tenant TiDB scans for serverless scale-to-zero


mornyx:
## Problem

With ~100k tenants on serverless TiDB, any periodic background operation that enumerates tenant databases prevents scale-to-zero and incurs sustained RCU cost. The semantic worker's keyset tenant scan (already addressed in #657) was the most severe path, but several other periodic operations still touch cached-but-idle tenant TiDBs:

| # | Path | File | Current trigger | Frequency | Iterates | Blocks scale-to-zero |
|---|------|------|----------------|-----------|----------|---------------------|
| 1 | Semantic worker tenant scan | semantic_worker.go | keyset poll `tenants` table | 200ms | All active tenants | ✅ (fixed in #657) |
| 2 | Semantic worker recovery scan | semantic_worker.go | `recoverLoop` | 5s | All active tenants | ✅ (fixed in #657) |
| 3 | SSE notify poller | notify_poller.go | per-bus 1s poll → central outbox | 200ms | Meta DB only | ❌ (already fixed) |
| 4 | FileGCWorker | file_gc_worker.go | per-backend goroutine ticker | 10s/tenant | All cached backends | ⚠️ Yes |
| 5 | quotaOutboxWorker | quota_outbox_worker.go | per-backend goroutine ticker | 1s/tenant | All cached backends | ⚠️ Yes |
| 6 | cleanupFSEvents | server.go | leader ticker over `ActiveTenantStores` | 5min | All cached backends | ⚠️ Yes |
| 7 | collectObservation | semantic_worker.go | `recoverLoop` over `ActiveTenantStores` | 5s | All cached backends | ⚠️ Yes |

The key insight: `pool.ActiveTenantStores()` returns tenants that are in the LRU cache (`maxSize=1024`), not tenants that are currently active. Once a tenant is touched by any request, its backend is cached, and paths 4-7 continue waking its TiDB at fixed intervals — even when the tenant has no traffic. With multiple pods, each pod caches a different subset of tenants, spreading the wake-up across the entire fleet.

### Unified outbox table

Replace `sse_notify_outbox` and the newly-added `semantic_notify_outbox` with a single append-only table:

```sql
CREATE TABLE IF NOT EXISTS tenant_notify_outbox (
    id         BIGINT UNSIGNED AUTO_INCREMENT PRIMARY KEY,
    tenant_id  VARCHAR(64) NOT NULL,
[...]

PR #760, opened 2026-07-20
feat: allow disabling the tenant safety-net scan with interval=0

mornyx:
## Why

The 5min safety-net scan is the last periodic per-tenant TiDB touch left after the unified outbox (#660). Each cycle runs four queries against every warm-pool tenant. Tenants that are idle but not yet evicted from the pool (idle TTL 5min + reap 2min) have their serverless clusters woken by these queries — visible as the 5min active-cluster sawtooth (trough ~60 → peak ~200 clusters).

The scan is purely a backstop:

- the outbox poller (200ms, meta-DB only) and in-process write kicks are the primary delivery path;
- every kick already piggybacks expired-lease recovery (`tenant_worker.go`);
- nothing it covers is correctness-critical: lost kicks / expired leases / due retries for an idle tenant are deferred to that tenant's next write.

For deployments where semantic-indexing delay and file-GC delay for idle tenants are acceptable, the scan is now optional.

## What changed

**1. Safety-net scan can be disabled**

- `server.Config.SafetyNetScanInterval <= 0` disables the scan. `NewWithConfig` no longer substitutes the 5min default — the default moves to the binaries' env fallback, so **default behavior is unchanged**.
- `startNotifyInfrastructure` skips the scan goroutine when disabled and logs `safety_net_scan_disabled`.
- Removed the now-unused `defaultSafetyNetScanInterval` const.
- Added `TestSafetyNetScanIntervalZeroDisables`.

**2. Env var rename transition (deprecate misleading `*_MS` names)**

The five tenant interval env vars are parsed with `time.ParseDuration`, so a bare number like `300000` (as the old help text implied) fails to parse and silently falls back to the default. They get canonical names without the `_MS` suffix:

| canonical | deprecated (still honored) |
|---|---|
| `DRIVE9_TENANT_OUTBOX_POLL_INTERVAL` | `DRIVE9_TENANT_OUTBOX_POLL_INTERVAL_MS` |
| `DRIVE9_TENANT_OUTBOX_CURSOR_FLUSH_INTERVAL` | `DRIVE9_TENANT_OUTBOX_CURSOR_FLUSH_MS` |
| `DRIVE9_TENANT_SHARD_REFRESH_INTERVAL` | `DRIVE9_TENANT_SHARD_REFRESH_MS` |
| `DRIVE9_TENANT_MAINTENANCE_INTERVAL` | `DRIVE9_TENANT_MAINTENANCE_INTERVAL_MS` |
| `DRIVE9_SAFETY_NET_SCAN_INTERVAL` | `DRIVE9_SAFETY_NET_SCAN_INTERVAL_MS` |

New `envDurationCompat` helper: canonical name wins when both are set; the legacy name logs a deprecation warning to stderr. Existing deployments keep working until they migrate. Help text now documents the new names, the Go-duration format, and the deprecation.

## Usage

Set `DRIVE9_SAFETY_NET_SCAN_INTERVAL=0` (or the deprecated `..._MS` name) to disable the scan. Any Go duration works (`30m`, `24h`); unset keeps 5min.

## Notes / follow-ups

- With the scan off, retry-due tasks for idle tenants wait for the tenant's next write. If tighter retry latency is ever needed without the scan, schedule an in-memory re-kick at `retryAt` in `retryTask` and keep the scan as a crash-only backstop at a long interval.
- The deprecated `*_MS` names can be removed in a later release once deployments have migrated.
- Observe the effect via `user_db_access{...="safety_net_tenant_scan"}` (should drop to zero when disabled) and `acquire_cold_open` (should be unchanged).

## Test

- `go test ./pkg/server/ -run 'TestSafetyNetScan|TestTenantOutboxPoller'` — pass
- Full `go test ./pkg/server/` — pass (194.9s)
- `go test ./cmd/drive9-server/ ./cmd/drive9-server-local/` — pass (incl. new `TestEnvDurationCompat`)
- `golangci-lint run ./cmd/... ./pkg/server/...` — 0 issues

### DIFF
--- a/cmd/drive9-server-local/main.go
+++ b/cmd/drive9-server-local/main.go
@@ -305,7 +305,7 @@ func main() {
 		TenantOutboxCursorFlushInterval: envDuration("DRIVE9_TENANT_OUTBOX_CURSOR_FLUSH_MS", 5000*time.Millisecond),
 		TenantShardRefreshInterval:      envDuration("DRIVE9_TENANT_SHARD_REFRESH_MS", 5000*time.Millisecond),
 		TenantMaintenanceInterval:       envDuration("DRIVE9_TENANT_MAINTENANCE_INTERVAL_MS", 300000*time.Millisecond),
-		SafetyNetScanInterval:           envDuration("DRIVE9_SAFETY_NET_SCAN_INTERVAL_MS", 300000*time.Millisecond),
+		SafetyNetScanInterval:           envDuration("DRIVE9_SAFETY_NET_SCAN_INTERVAL_MS", 5*time.Minute),
 	})
 	defer srv.Close()
 	logLocalStartupStep(startupCtx, startupStart, stepStart, "create_server")
@@ -420,7 +420,8 @@ environment:
   DRIVE9_TENANT_OUTBOX_CURSOR_FLUSH_MS   poller cursor flush interval (default: 5000)
   DRIVE9_TENANT_SHARD_REFRESH_MS         shard resolver ring refresh (default: 5000)
   DRIVE9_TENANT_MAINTENANCE_INTERVAL_MS  piggyback maintenance throttle (default: 300000)
-  DRIVE9_SAFETY_NET_SCAN_INTERVAL_MS     leader safety-net scan interval (default: 300000)
+  DRIVE9_SAFETY_NET_SCAN_INTERVAL_MS     per-pod safety-net scan interval as a Go
+                    duration (e.g. 5m, 24h); 0 disables the scan (default: 5m)
 
   Image extraction (async image -> text for search):
   DRIVE9_IMAGE_EXTRACT_ENABLED true|false (default: false)
--- a/cmd/drive9-server-local/main.go
+++ b/cmd/drive9-server-local/main.go
@@ -301,11 +301,11 @@ func main() {
 		Logger:                          srvLogger,
 		SemanticEmbedder:                semanticEmbedder,
 		TenantWorkers:                   workerOpts,
-		TenantOutboxPollInterval:        envDuration("DRIVE9_TENANT_OUTBOX_POLL_INTERVAL_MS", 200*time.Millisecond),
-		TenantOutboxCursorFlushInterval: envDuration("DRIVE9_TENANT_OUTBOX_CURSOR_FLUSH_MS", 5000*time.Millisecond),
-		TenantShardRefreshInterval:      envDuration("DRIVE9_TENANT_SHARD_REFRESH_MS", 5000*time.Millisecond),
-		TenantMaintenanceInterval:       envDuration("DRIVE9_TENANT_MAINTENANCE_INTERVAL_MS", 300000*time.Millisecond),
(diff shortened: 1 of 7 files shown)


srstack (approved):
Reviewed (read-only diff — did not compile locally, but the PR reports full `go test ./pkg/server/` + cmd tests + lint all green). Clean, minimal, well-justified change on both axes. Approving.

**1. Disabling the safety-net scan — correct and default-preserving.**
The mechanism is exactly right: `NewWithConfig` no longer coerces `safetyNetScanInterval <= 0` to the 5min default, so a non-positive value now passes through as "disabled"; `startNotifyInfrastructure` gates the goroutine on `> 0` and logs `safety_net_scan_disabled` otherwise. The 5min default moves down into both binaries' env fallback (`envDurationCompat(..., 5*time.Minute)`), so **default behavior is unchanged** — only an explicit `=0` disables it. `TestSafetyNetScanIntervalZeroDisables` pins both halves (zero preserved as disabled; a configured interval passes through untouched).

**Verified the safety argument against the code, not just the PR description.** The `safety_net_scan.go` doc-comment confirms the scan is warm-only (`AcquireCached`, never opens a cold tenant TiDB — the #658 invariant) and is a backstop, not the primary path. Everything it covers is recoverable elsewhere: queued tasks are produced by writes, every write triggers an in-process kick (~0ms same-pod), the outbox poller is the cross-pod primary path, and durable task rows in the tenant TiDB are drained on the tenant's next write. So disabling it defers idle-tenant semantic-index/file-GC latency (invisible for a cold, trafficless tenant) without risking correctness. The reasoning holds.

**Nice incidental doc fix:** the `SafetyNetScanInterval` comments changed "how often **the leader** runs" → "how often **each pod** runs." That matches the actual behavior (the scan was already per-pod / not leader-gated per `startNotifyInfrastructure`), so this corrects stale comments rather than changing behavior.

**2. `*_MS` → canonical rename — fixes a real latent bug.**
The old names are parsed with `time.ParseDuration`, so a `_MS` name invites a bare-number value (`300000`) that fails to parse and **silently falls back to the default** — the name actively misleads. `envDurationCompat` resolves cleanly: canonical wins when both set, legacy is honored with a stderr deprecation warning, else fallback. `TestEnvDurationCompat` covers all three branches. I double-checked the default-value rewrites are equivalent: `300000ms = 5m`, `5000ms = 5s`, `200ms` unchanged — no default drift.

**Non-blocking notes:**
- `envDurationCompat` is duplicated verbatim in both `cmd/` mains (as `envDuration` already was), so this follows the existing pattern — but a shared helper package for these env parsers would be worth a future cleanup now that there are two.
- The deprecation warning goes to `os.Stderr` rather than the structured logger. Reasonable, since env parsing runs before the logger is built, and the help text documents it too — just means the warning won't surface in structured log/alerting, so a deployment on a deprecated name relies on someone reading startup stderr. Fine for a transitional warning.
- Good call flagging the `retryAt` in-memory re-kick as the future path if tighter retry latency is ever needed with the scan off — agreed that's the right shape (keep the scan as a long-interval crash-only backstop) rather than reintroducing the 5min per-tenant touch.

LGTM.

PR #764, opened 2026-07-20
Keep quota reads side-effect free and tune MySQL pool lifetimes

srstack:
## Summary

- validate TiDB Cloud credential-based quota reads with BASIC cluster information only on RBAC cache misses, while preserving the local Drive9 API-key fast path
- stop quota and admin read endpoints from backfilling cloud spending limits into the meta database; explicit quota updates and lifecycle seeds remain unchanged
- increase default meta and user-schema connection lifetimes and the meta idle timeout

## Test Plan

- [x] go test ./pkg/mysqlutil -count=1
- [x] go test ./pkg/tenant/tidbcloudnative -count=1
- [x] go test ./pkg/server -run Test\(QuotaGet\|DeprecatedQuotaGet\|QuotaSet\|AdminTenantGet\|AdminTenantList\|AdminTenantQuotaSet\|AdminTenantPoolReplenishUsesDefaultSpendingLimit\) -count=1
- [x] make lint
- [x] git diff --check origin/main...HEAD

## Full server test note

A full go test ./pkg/server -count=1 was attempted twice. In both runs, the package-local MySQL test instance returned invalid connection after roughly two minutes, causing cascading failures in later token, upload, and vault tests. No quota or admin-quota assertion failed, and the focused affected test set above passes.

## Summary by CodeRabbit

* **Bug Fixes**
  * Quota and tenant administration views no longer overwrite or backfill locally configured spending limits when retrieving quota information.
  * Quota responses continue to include current quota details without triggering background spending-limit synchronization.
  * Improved connection pool defaults help maintain connections longer during normal operation.

* **Tests**
  * Added coverage confirming quota and tenant requests preserve local spending-limit settings and avoid unintended updates.

### DIFF
--- a/pkg/mysqlutil/pool.go
+++ b/pkg/mysqlutil/pool.go
@@ -9,15 +9,15 @@ import (
 )
 
 const (
-	defaultMetaConnMaxLifetime       = 5 * time.Minute
-	defaultMetaConnMaxIdleTime       = 45 * time.Second
+	defaultMetaConnMaxLifetime       = 10 * time.Minute
+	defaultMetaConnMaxIdleTime       = 1 * time.Minute
 	defaultMetaMaxOpenConns          = 100
 	defaultMetaMaxIdleConns          = 20
 	defaultUserConnMaxLifetime       = 5 * time.Minute
 	defaultUserConnMaxIdleTime       = 1 * time.Minute
 	defaultUserMaxOpenConns          = 6
 	defaultUserMaxIdleConns          = 2
-	defaultUserSchemaConnMaxLifetime = 1 * time.Minute
+	defaultUserSchemaConnMaxLifetime = 3 * time.Minute
 	defaultUserSchemaConnMaxIdleTime = 20 * time.Second
 	defaultUserSchemaMaxOpenConns    = 8
 	defaultUserSchemaMaxIdleConns    = 2
--- a/pkg/server/admin_tenants.go
+++ b/pkg/server/admin_tenants.go
@@ -232,8 +232,7 @@ func (s *Server) handleAdminTenantList(w http.ResponseWriter, r *http.Request) {
 		errJSON(w, http.StatusBadRequest, err.Error())
 		return
 	}
-	listStarted := time.Now().UTC()
-	clusters, clustersFromCache, err := s.listAllManagedClusters(r.Context(), cred, "", "admin_tenant_list")
+	clusters, _, err := s.listAllManagedClusters(r.Context(), cred, "", "admin_tenant_list")
 	if err != nil {
 		writeAdminTiDBCloudError(w, r.Context(), err, "list tenants")
 		return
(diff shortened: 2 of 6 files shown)


srstack:
Reviewed the full branch (read-only, did not compile; noting the test-plan already covers the focused quota/admin suites + lint). This is a clean, well-scoped change — net -176/+162, mostly deleting read-path side effects. **LGTM with a couple of things to confirm, no blockers.**

## ✅ Theme 1: side-effect-free quota reads — correct

`handleQuotaGet` / `handleAdminTenantList` / `authorizedAdminTenant` no longer call `syncTiDBCloudSpendingLimit` as a side effect of a read. `writeQuotaResponse` returns the persisted meta-DB value (`GetQuotaConfig`). Verified `getter.GetQuota` at quota.go:129 is now a pure credential/RBAC validation call with its result discarded (`_,`) — and it's the **only** caller of `GetQuota` in the server, so nothing depends on the spending limit it used to return. Good.

This is the right resolution to the concern I raised on #762 (the sync-failure-downgraded-to-warning): rather than a read that best-effort-writes, reads are now purely reads, and spending limits are written only by explicit quota updates + lifecycle seeds. One consequence worth stating in the PR body: **a GET can now return a stale spending limit** (no cloud refresh on read). That's the intended trade and it's consistent, but it is a client-visible contract change from "GET refreshes from cloud" to "GET returns last-persisted" — worth a sentence so downstream consumers know.

## ✅ Theme 2: BASIC view + removing clusterQuotaInfo — clean

`getClusterBasicInfoWithCredentials` unifies on `?view=BASIC`, and the heavier unmasked `clusterQuotaInfo` (which parsed `SpendingLimit`) is deleted. `GetQuota` now returns `{Labels}` only. Verified the nil-map guarantee: `getClusterBasicInfoWithCredentials` sets `info.Labels = make(...)` when nil before returning, so all four call sites (`MarkClusterPoolUsed/Free`, `MarkQuotaUpdateStarted`, `GetQuota`) can safely write `labels[...]` without a nil-map panic. `GetQuota` also correctly copies the map before returning it rather than aliasing `info.Labels`. The dead helpers (`syncAdminTenantSpendingLimits`, `adminTenantRowsNeedSpendingLimitBackfill`, `cloudClusterMap`, `listAllManagedClustersFresh`) are all removed with their last callers — no orphans.

One thing to confirm: `authorizedAdminTenant` now always passes `allowCache=true` (previously it bypassed the cache when the spending limit was nil, to force a fresh read for backfill). Since backfill is gone that's consistent, but it does mean admin authz can now be decided entirely off cached cluster listings. Confirm the RBAC cache TTL is short enough that a revoked/rotated credential can't authorize against a stale cluster list for longer than acceptable.

## 🟡 Theme 3: MySQL pool lifetimes — low-risk tuning, no data given

meta lifetime 5→10m, meta idle 45s→1m, user-schema lifetime 1→3m. Direction is reasonable (fewer reconnects, less churn — plausibly related to the earlier `context canceled` refresh noise). Non-blocking, but these are empirical knobs with no benchmark in the PR — worth a one-line rationale, and note that longer `ConnMaxLifetime` keeps connections to a rotated backend alive a bit longer (generally fine behind the TiDB LB).

## Nit
- The test note says full `./pkg/server` couldn't complete due to the package-local MySQL instance returning `invalid connection` after ~2 min. Given this very PR changes conn lifetimes, it'd be reassuring to confirm that flakiness is pre-existing infra and not interaction with the new `ConnMaxLifetime` values — even just a run on main showing the same failure.

Net: correct, subtractive, and it resolves the read-side-effect concern from #762. Happy to see it land once the stale-GET contract note and the admin-cache-TTL check are addressed.

PR #753, opened 2026-07-18
Add shared-schema (fs_id) foundation and minimal shared-DB mode

mornyx:
## Summary

Implements the first phase of the tenant-DB shared-schema redesign plus the **minimal shared-DB runtime**: tenants in an org with a registered shared-schema database are provisioned onto it instead of getting a dedicated cluster. Existing standalone tenants are fully untouched (production SQL is byte-identical, proven by the unchanged existing test suite).

Scope: identity (`fs_id`), shared-schema DDL for 31 of the 32 tenant tables, dual-shape datastore/vault layer, placement metadata, routing, shared-pool provisioning, shared-tenant lifecycle, and a local end-to-end smoke for shared mode. Migration / hotspot scheduling / fork / quota semantics / backup-restore adaptations are intentionally out of scope — but cross-tenant correctness is covered end to end (see tests).

**Design doc**: `docs/TENANT_DB_REDESIGN.md` — the full redesign proposal (English) is committed with this PR.

## What's included

**Identity**
- `fs_registry` (tenant UUID ↔ fs_id BIGINT): allocated atomically in `InsertTenant`; leader-gated startup backfill; `EnsureFsID`/`ResolveFsID`/`ResolveTenantID`.
- `db_pool` (shared DB registry: connection coords, org binding with `*` wildcard, capacity) and `tenant_placements` (fs_id → db_id, placement, schema shape) with CRUD.

**Shared schema (31 tables: core FS 9, journal 5, vault 7, git 5, fs_layer 5)**
- One canonical TiDB list per group; `PRIMARY KEY (fs_id, ...) CLUSTERED` explicit (TiDB defaults composite PKs to NONCLUSTERED). MySQL variants **derived mechanically** (strip `CLUSTERED`, `VECTOR(n)→LONGTEXT`). Per-group drift tests enforce standalone↔shared parity.
- `semantic` uses plain `VECTOR(n)` (app-managed; FTS/vector indexes in an optional TiDB-only list). `fs_events` keeps its global AUTO_INCREMENT PK + `(fs_id, seq)` index. Repair-era ALTERs folded into CREATE TABLEs.
- `llm_usage` deliberately omitted: the backend only writes tenant-DB `llm_usage` when server quota is inactive (single-tenant/local mode), which shared-schema DBs never run — central `tenant_llm_usage` is authoritative.
[...]

PR #765, opened 2026-07-21
Resolve tenant org through the RBAC cluster-list cache

mornyx:
## Summary

`firstManagedOrganization` called `ListManagedClusters` directly, bypassing the existing 1h RBAC cluster-list cache. On a provision request that checks both the warm pool (`claimAdminTenantFromPool`) and the shared-schema pool (`findSharedDBForProvision`, #753), the same credential paid one TiDB Cloud list call **per check**. Routing it through `listAllManagedClustersCached`:

- first lookup populates the cache, repeated checks on the same request path hit it
- consecutive provisions with the same credential hit it for the TTL window
- tenant mutations keep invalidating via `forgetTiDBCloudRBACList` (unchanged)

Small, self-contained follow-up split out of #753 review discussion.

## Test plan

- [x] New regression `TestFirstManagedOrganizationUsesRBACCache`: two lookups for one credential → one list call; `forgetTiDBCloudRBACList` → fresh fetch
- [x] Full `pkg/server` suite — pass
- [x] `make lint` — 0 issues

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved managed organization selection by using cached access-control information.
  * Removed failures caused by unavailable managed-cluster listing support.
  * Ensured organization data refreshes correctly after cached permissions are cleared.

### DIFF
--- a/pkg/server/admin_tenant_pool.go
+++ b/pkg/server/admin_tenant_pool.go
@@ -681,15 +681,15 @@ func (s *Server) refreshTenantPoolCapacity(ctx context.Context, pool *meta.Tenan
 }
 
 func (s *Server) firstManagedOrganization(ctx context.Context, cred tenant.CredentialProvisionRequest) (string, error) {
-	lister, ok := s.provisioner.(tenant.ManagedClusterLister)
-	if !ok {
-		return "", fmt.Errorf("managed cluster list not enabled")
-	}
-	page, err := lister.ListManagedClusters(ctx, cred, tenant.ManagedClusterListOptions{PageSize: 1})
-	if err != nil || page == nil || len(page.Clusters) == 0 {
+	// Resolve through the RBAC list cache so repeated org lookups on the same
+	// request path (warm-pool claim + shared-pool check) and across
+	// consecutive provisions do not each cost a TiDB Cloud list call. The
+	// cache is invalidated on tenant mutations (forgetTiDBCloudRBACList).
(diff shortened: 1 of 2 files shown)


qiffang (approved):
## Review: APPROVE

Clean, well-scoped optimization. Routing `firstManagedOrganization` through `listAllManagedClustersCached` eliminates redundant TiDB Cloud API calls when the same credential is used for multiple pool checks (warm-pool claim + shared-pool check) within a single provision request.

### What I checked

1. **Behavioral change**: Old code fetched `PageSize: 1` directly. New code goes through the 1h RBAC cache with `PageSize: 100` full pagination on cache miss. This is a correct trade-off — one full list call populates the cache, subsequent org lookups on the same request path (and across consecutive provisions) are free.

2. **Cache invalidation**: `forgetTiDBCloudRBACList` on tenant mutations is unchanged. Test verifies: two lookups = 1 list call; forget + lookup = 2 list calls.

3. **Error handling**: `err != nil || len(clusters) == 0` returns `err` correctly (nil when empty but no error). Matches old behavior where `page == nil || len(page.Clusters) == 0` returned nil error.

4. **`OrganizationID` access**: `clusters[0].OrganizationID` is safe — the `len(clusters) == 0` guard protects against empty slice. `TrimSpace` preserved.

5. **Test**: `TestFirstManagedOrganizationUsesRBACCache` is comprehensive — verifies cache hit (same org returned, 1 list call), cache invalidation (new org after forget, 2 list calls). Uses existing `newQuotaRuntime` test infrastructure.

No blockers. No nits.

PR #766, opened 2026-07-21
test: retry MySQL testcontainer start on Docker ephemeral port race

mornyx:
## Problem

CI's `make test` / `make test-failpoint` steps flake with:

```
setup mysql test instance: start mysql container: ... Error response from daemon:
failed to set up container networking: driver failed programming external connectivity on endpoint ...:
failed to bind host port for 0.0.0.0::172.18.0.2:3306/tcp: address already in use
```

Seen at least 3 times recently (runs 29730070989, 29728174538, 29696704434); re-running the job goes green.

## Root cause

The MySQL testcontainer asks Docker for an **ephemeral** host port. Under load, Docker's port allocator can lose a race (another process grabs the picked port between allocation and bind) and the container start fails. Nothing in this repo pins a fixed host port — this is a known Docker daemon flake, not a test bug.

## Fix

`internal/testmysql.Start` now retries container startup up to 3 times (2s/4s backoff) when the error matches the port-bind signature (`failed to bind host port` + `address already in use`). Any other startup error still fails immediately. Context cancellation is honored between attempts.

## Testing

- New unit test `TestIsPortBindConflict` covers the matcher (positive case mirrors the real CI error text).
- Ran a real container-backed test: `make test TEST_RUN='TestInsertAndGetNode' TEST_PKGS='./pkg/datastore/...'` → ok.
- `golangci-lint run internal/testmysql/` → 0 issues.

### DIFF
--- a/internal/testmysql/testmysql.go
+++ b/internal/testmysql/testmysql.go
@@ -36,12 +36,28 @@ func Start(ctx context.Context) (*Instance, error) {
 	ctx, cancel := context.WithTimeout(ctx, 4*time.Minute)
 	defer cancel()
 
-	c, err := tcmysql.Run(ctx,
-		"mysql:8.0.36",
-		tcmysql.WithDatabase("dat9_test"),
-		tcmysql.WithUsername("dat9"),
-		tcmysql.WithPassword("dat9pass"),
-	)
+	// The MySQL container asks Docker for an ephemeral host port, but Docker's
+	// port allocator can lose a race on busy CI runners and fail container
+	// start with "failed to bind host port ...: address already in use".
+	// Retrying picks a fresh port and is far cheaper than re-running the job.
+	var c *tcmysql.MySQLContainer
+	var err error
+	for attempt := 1; ; attempt++ {
+		c, err = tcmysql.Run(ctx,
+			"mysql:8.0.36",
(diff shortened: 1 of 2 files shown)


srstack (approved):
Reviewed (read-only diff). Small, correct, well-tested. Approving.

Retrying `tcmysql.Run` up to 3 times on Docker's ephemeral host-port bind race is a reasonable, cheap recovery for CI flakiness (vs. re-running the whole job). Good details:
- The retry is **narrowly gated**: only `isPortBindConflict(err)` retries; any other failure (image pull, etc.) breaks immediately, so a real error isn't masked by 3× the wait.
- `isPortBindConflict` requires **both** `"failed to bind host port"` and `"address already in use"`, so a generic `address already in use` without the bind context (e.g. an app-level listen) does not trigger a retry — the negative test case covers exactly that.
- Backoff respects `ctx.Done()` and the outer 4-minute timeout, so retries can't blow the deadline.
- `TestIsPortBindConflict` covers the real Docker message, nil, unrelated failure, and the near-miss — and uses `t.Errorf` per the guidelines.

Nit (non-blocking): the retry is capped at 3 attempts; if the race is somehow persistent that's still a fail, which is correct — just noting the cap is intentional and worth the inline comment it already has. LGTM.

PR #767, opened 2026-07-21
chore: drop dead meta sse_notify_outbox/llm_usage tables and unused vault_policies

mornyx:
## What

Removes three dead tables and their code paths:

1. **Meta DB `sse_notify_outbox`** — legacy central SSE notification pointer table, long superseded by `tenant_notify_outbox` + pod-to-pod push (`handleInternalSSENotify`). All production reads/writes were already gone; this drops the DDL and the deprecated helpers (`SSENotifyRow`, `InsertSSENotify`, `ListSSENotifySince`, `MaxSSENotifyID`, `DeleteSSENotifyBefore`), and fixes two stale `eventbus.go` comments that still described the old mechanism.

2. **Meta DB `llm_usage`** — legacy control-plane LLM cost ledger. The authoritative paths are `tenant_llm_usage` (raw ledger, written via `InsertCentralLLMUsage[Tx]`) and `tenant_monthly_llm_cost` (pre-aggregated counter, `IncrMonthlyLLMCost[Tx]`). Removes:
   - the meta `llm_usage` DDL,
   - `Store.InsertLLMUsage` (no production callers),
   - the SUM fallback in `Store.MonthlyLLMCostMillicents` — a missing counter row now simply returns 0, which matches the previous behavior for tenants with no usage,
   - `LLMCostCache` / `NewLLMCostCache` and the `MetaLLMUsageStore` interface in `pkg/backend` — verified via grep to have no production callers (only tests).

   The tenant-DB `llm_usage` table (used by standalone/single-tenant mode when server quota is inactive) is **not** touched.

3. **`vault_policies`** — no read or write code path exists anywhere (`pkg/vault` has no policy code at all). Drops the DDL from both the standalone and shared vault schema shapes plus all table-list references. The shared (multi-tenant) schema goes from 31 to 30 tables; comments and `AGENTS.md` updated accordingly.

Note: the DDL removal means new deployments won't create these tables. Existing deployments keep any already-created tables (no `DROP TABLE` migration is added); they are simply never touched.

## Why it's safe

- Repo-wide greps confirm zero remaining references to `sse_notify_outbox`, `SSENotify*`, meta `InsertLLMUsage`, `LLMCostCache`, `MetaLLMUsageStore`, and `vault_policies` outside of historical design docs.
- `MonthlyLLMCostMillicents` behavior is unchanged for both counter-present (returns `total_mc`) and counter-absent (returns 0) tenants.

## How tested

- `go build ./...` — clean
- `bin/golangci-lint run` — 0 issues
- `make test TEST_PKGS='./pkg/meta/... ./pkg/migrate/... ./pkg/vault/... ./pkg/tenant/...'` — pass
- `make test TEST_PKGS='./pkg/datastore/...'` — pass
- `make test TEST_PKGS='./pkg/server/...'` — pass
- `drive9-server schema dump-init-sql` for all four providers (`tidb_zero`, `tidb_cloud_native`, `db9`, `tidb_cloud_native_shared`) — `vault_policies` gone everywhere; standalone dumps still contain the tenant-DB `llm_usage` as intended; shared dump is now 30 tables


srstack (approved):
Reviewed the full branch (read-only diff, verified against code). Clean dead-code removal — net -643/+34 across three dead tables (`llm_usage`, `sse_notify_outbox`, `vault_policies`) and their orphaned CRUD. Approving.

**Verified all three are actually dead:**
- **`sse_notify_outbox`**: the deleted symbols (`SSENotifyRow`, `InsertSSENotify`, `ListSSENotifySince`, `MaxSSENotifyID`, `DeleteSSENotifyBefore`) have **zero remaining references** anywhere in the repo (incl. tests) — grep is empty, so no build break. It was already superseded by `tenant_notify_outbox`; the `eventbus.go` change is comment-only (`sse_notify_outbox` → `tenant_notify_outbox` in the design note), correctly reflecting the replacement.
- **meta `llm_usage` (unprefixed)**: removed, while `pkg/meta/llm_usage.go` correctly **keeps** `MonthlyLLMCostMillicents` (reads `tenant_monthly_llm_cost`), and the live central ledger `tenant_llm_usage` / `InsertCentralLLMUsage[Tx]` is untouched. Important not to conflate these — the standalone `pkg/datastore/llm_usage.go` table is a separate concern and isn't in scope here. The billing path stays intact.
- **`vault_policies`**: dropped from both the standalone and shared `VaultTiDBSchemaStatements`; no non-test references remain.

**Migration safety confirmed**: all three CREATEs used `IF NOT EXISTS`, and `diffMetaSchema` only iterates `spec.tables` (forward check — flags spec tables missing from the DB, never the reverse). So removing them from the spec means new deploys don't create them, existing deploys keep them as harmless orphan tables, and schema validation does **not** error on the now-unlisted tables. No DROP, purely forward-compatible — the right way to retire dead tables.

The `purge.go`/`store.go` comment/wipe-list adjustments and the test-helper table-list trims all line up with the removals. LGTM.

PR #768, opened 2026-07-21
docs: add SSE notifications guide for external clients

mornyx:
## Summary

There was no external-facing SSE integration guide. The SSE semantics lived in `docs/specs/cache-invalidation.md` (written for FUSE cache implementers) and `pkg/client/events.go` (Go source). External clients — sync agents, UIs, non-Go SDK consumers — that want to receive filesystem change notifications had no document aimed at them.

This PR adds `docs/guides/sse-notifications.md`, a standalone guide covering:

- The `/v1/events` endpoint, headers (`Authorization`, `X-Dat9-Actor`), and the `since` cursor.
- The three event types (`file_changed`, `reset`, `heartbeat`) with field tables and the full set of `reason` values (`initial_sync`, `seq_too_old`, `server_restart`, `structural_change`).
- The authoritative operation → event routing table, including the `chmod` gap and the `hardlink` double-event behavior.
- The four client implementation requirements: `seq`-based cursor, exponential-backoff reconnect, full-clear on `reset`, unique `X-Dat9-Actor` self-filter.
- Disconnect/reconnect semantics (replay vs `seq_too_old` vs `server_restart`).
- Multi-client isolation and concurrent-write caveats.
- Sticky-routing guidance for multi-pod deployments (the event bus is per-process).
- Go SDK usage with verified `client.New(baseURL, apiKey)` / `c.WatchEvents` / `c.WatchEventsWithLifecycle` signatures.
- A non-Go client implementation outline (the protocol is plain SSE over HTTP).
- A limitations section (no per-path subscription, metadata-only payload, finite retention, `chmod` emits no event).

All facts — header names, event field order, `reason` values, op routing, SDK signatures, the 30s heartbeat interval — were verified against `pkg/server/sse.go`, `pkg/client/events.go`, and `pkg/client/client.go`.

## Test plan

- [x] Doc-only change; no code or test impact.
- [x] Go SDK signatures in the examples verified against `pkg/client/client.go` (`New`, `SetActor`) and `pkg/client/events.go` (`WatchEvents`, `WatchEventsWithLifecycle` are methods on `*Client`).
- [x] Event field order and `reason` vocabulary verified against `pkg/server/sse.go` (`sendSSEReset` reasons, `isStructuralOp` routing) and `pkg/client/events.go` struct definitions.
- [x] Cross-references to `docs/specs/cache-invalidation.md` and `docs/guides/go-sdk-integration.md` resolve.

### DIFF
--- a/docs/guides/sse-notifications.md
+++ b/docs/guides/sse-notifications.md
@@ -1,0 +1,361 @@
+# drive9 SSE notifications guide
+
+This guide shows how to receive real-time filesystem change notifications from
+a drive9 server over Server-Sent Events (SSE). It is intended for external
+clients — integrations, sync agents, UIs, and non-Go SDK consumers — that need
+to react to file changes without polling.
+
+For the authoritative cache-invalidation semantics that this stream drives on
+FUSE mounts, see
+[`docs/specs/cache-invalidation.md`](../specs/cache-invalidation.md). For Go SDK
+usage, see [`docs/guides/go-sdk-integration.md`](go-sdk-integration.md).
+
+## What SSE gives you
+
+A single long-lived HTTP stream per tenant that delivers every filesystem
+mutation as it commits: writes, uploads, creates, symlinks, hardlinks, renames,
+deletes, mkdirs, and copies. The server assigns a monotonic `seq` to each event
+so clients can resume after a disconnect without missing events.
+
+SSE is the lowest-latency notification channel drive9 exposes. There is no
+webhook, no message queue, and no per-path subscription — the stream is
+per-tenant and carries every path event for that tenant. Clients filter
+client-side.
+
+## Endpoint
+
+```
+GET /v1/events?since=<uint64>
+Accept: text/event-stream
+Authorization: Bearer <api_key>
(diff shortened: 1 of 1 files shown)


srstack (approved):
Reviewed (read-only). Docs-only — a 361-line SSE notifications guide for external clients, no code. Approving.

Content is accurate against the server behavior I've seen in this codebase: per-tenant single long-lived stream, monotonic `seq` cursor with `since=0` → `reset` + `heartbeat` head, `since=N` replay-then-stream, owner/fs_scoped-read key auth, and the `X-Dat9-Actor` self-filter with the explicit warning that each client instance needs a unique actor id (two clients sharing an id silently filter each other — good that it's called out). Correctly points at `docs/specs/cache-invalidation.md` as the authoritative semantics and `go-sdk-integration.md` for Go.

Only ask (non-blocking): since this documents a client-facing wire contract, it's worth keeping in sync with any endpoint/precedence changes — e.g. PR #764 is adjusting quota read semantics and there's ongoing churn in the quota/auth area; this guide's `/v1/events` surface looks unaffected, just flagging that guides like this drift easily. LGTM.

PR #769, opened 2026-07-21
e2e: existing-tenant regression + shared-mode isolation and full smokes

mornyx:
## Summary

This PR tightens two e2e coverage gaps that both matter after shared-schema landed:

1. **Existing-tenant regression** — run the core API/CLI smokes against a live tenant without re-provisioning (post-dev-release check).
2. **Shared-mode depth** — exercise multi-tenant isolation more thoroughly, then re-run the full API + CLI smokes on a real shared-schema server.

### Existing-tenant mode

- `api-smoke-test.sh` / `cli-smoke-test.sh` honor `DRIVE9_API_KEY`: skip `POST /v1/provision` and reuse the tenant.
- API suite cleans up its timestamped `team-${TS}` tree at the end in existing-tenant mode so the tenant is not polluted.
- Upload-limit boundary checks default **off** when `DRIVE9_API_KEY` is set (reserving `total_size` against an already-used quota can spuriously 507); explicit `RUN_*_UPLOAD_LIMIT_BOUNDARY=1` still wins.
- `smoke-all.sh` re-exports `DRIVE9_API_KEY` to every sub-suite and adds `RUN_API_ONLY=1` for a quick api+cli pair.
- Docs updated in `e2e/AGENTS.md` and `e2e/README.md` with the two-pass (existing then fresh) workflow.

```bash
eval "$(drive9 ctx show --json --reveal | jq -r '"export DRIVE9_BASE=\(.server)\nexport DRIVE9_API_KEY=\(.api_key)"')"
RUN_API_ONLY=1 bash e2e/smoke-all.sh   # existing tenant
unset DRIVE9_API_KEY
bash e2e/api-smoke-test.sh && bash e2e/cli-smoke-test.sh  # fresh
```

### Shared-mode smoke expansion

`shared-mode-smoke-test.sh` still owns its specialized shared-schema checks, and now also:

**Cross-tenant isolation (A vs B on the same shared DB)**
- Root `?list` does not leak peer dirs; peer dir list is 404 or 200+empty with no foreign children
- `?grep` / `?find` do not return the other tenant’s marker / `note.txt`
- Peer `copy` / `hardlink` / `rename` / `delete` of foreign paths → 404; source content intact
- Delete own same-path file does not affect the peer
- Concurrent same-path writes: each tenant still reads only its own bytes
- SSE: peer stream gets reset but not the other tenant’s file event
- Multipart: peer cannot read completed object; peer cannot complete foreign `upload_id`

**Nested full suites (fresh tenants each, not reusing A/B keys)**
- After specialized checks, runs `api-smoke-test.sh` then `cli-smoke-test.sh` against the same shared server
- Forces `RUN_CLI_FORK_CHECKS=0` (shared fork → 409)
- Forces `RUN_SEMANTIC_CHECKS=0` / `RUN_CLI_SEMANTIC_CHECKS=0` (server boots with auto-embedding disabled)
- Forces `RUN_SQL_CHECKS=0` (new knob on api-smoke; `POST /v1/sql` is unsupported on shared-schema stores)
- Opt out with `RUN_SHARED_API_CLI_SMOKE=0`

## Test plan

- [x] `RUN_SHARED_API_CLI_SMOKE=0 bash e2e/shared-mode-smoke-test.sh` — isolation + specialized checks green (`passed=83 failed=0`)
- [x] Full `bash e2e/shared-mode-smoke-test.sh` — nested api (94 pass / 13 skip) + cli (86 pass / 11 skip) green after `RUN_SQL_CHECKS=0`
- [ ] Optional: existing-tenant pass against a live ctx tenant  
      `DRIVE9_API_KEY=... RUN_API_ONLY=1 bash e2e/smoke-all.sh`
- [ ] Optional: fresh provision pass with `DRIVE9_API_KEY` unset

## Summary by CodeRabbit

* **Documentation**
  * Expanded guidance for running API and CLI smoke tests with either fresh or existing tenants.
  * Documented new controls for limiting suites, skipping optional checks, and configuring upload-boundary tests.

* **Testing**
  * Added existing-tenant regression support with optional cleanup.
  * Expanded shared-mode coverage for tenant isolation, SSE events, multipart uploads, and cross-tenant operations.
  * Added clearer skip reporting and configurable API, CLI, and shared-mode test execution.

### DIFF
--- a/e2e/AGENTS.md
+++ b/e2e/AGENTS.md
@@ -104,6 +104,47 @@ DRIVE9_TIDBCLOUD_PRIVATE_KEY="$DRIVE9_TIDBCLOUD_PRIVATE_KEY" \
 bash e2e/native-smoke-test.sh
 ```
 
+#### Existing-tenant regression
+
+After a dev release, run `api-smoke-test.sh` and `cli-smoke-test.sh` twice:
+once against an existing tenant (set `DRIVE9_API_KEY`) and once against a fresh
+provision (unset `DRIVE9_API_KEY`). In existing-tenant mode both suites skip
+provision and reuse the tenant; `api-smoke-test.sh` also cleans up its
+timestamped test tree at the end, and the upload-limit boundary checks default
+to `0` because reserving `total_size` against an existing tenant's quota can
+spuriously fail with 507.
+
+```bash
+# Export credentials from the current drive9 ctx tenant
+# (--reveal prints the full key; keep it out of logs/screenshots)
+eval "$(drive9 ctx show --json --reveal | jq -r '"export DRIVE9_BASE=\(.server)\nexport DRIVE9_API_KEY=\(.api_key)"')"
+
+# Pass 1 — existing tenant
+bash e2e/api-smoke-test.sh
+bash e2e/cli-smoke-test.sh
+
+# Or run the whole smoke-all matrix in existing-tenant mode (re-exports
+# DRIVE9_API_KEY to every sub-suite). Set RUN_API_ONLY=1 for just the core
+# api + cli pair, RUN_FUSE_SMOKE=0 on hosts without real FUSE.
+RUN_API_ONLY=1 bash e2e/smoke-all.sh
+
+# Pass 2 — fresh tenant
+unset DRIVE9_API_KEY
+bash e2e/api-smoke-test.sh
+bash e2e/cli-smoke-test.sh
+```
+
+Useful knobs for existing-tenant runs:
+
+- `RUN_CLI_FORK_CHECKS=0` — skip the fork flow. Recommended for shared-schema
+  tenants (which reject `/v1/fork` with 409) or very large tenants where
+  forking is undesirable.
+- `RUN_LARGE_FILE=0` — skip the 100MB API large-file upload.
+- `RUN_SEMANTIC_CHECKS=0` / `RUN_CLI_SEMANTIC_CHECKS=0` — skip semantic recall
+  checks when no embedding endpoint is available.
+- `RUN_UPLOAD_LIMIT_BOUNDARY=1` / `RUN_CLI_UPLOAD_LIMIT_BOUNDARY=1` — force the
+  upload-limit boundary checks back on in existing-tenant mode.
+
 ### Local via `drive9-server-local`
 
 When the task is specifically about local validation on this machine, prefer
@@ -201,6 +242,8 @@ use the same value as `DRIVE9_API_KEY` here.
 ### `api-smoke-test.sh`
 
 1. `POST /v1/provision` returns `202` with `tenant_id`, `api_key`, and `status`
+   — honors `DRIVE9_API_KEY`: when set, step 1 skips provision (emits SKIP
+   checks) and reuses the existing tenant
 2. `GET /v1/status` polled until `active`
 3. `GET /v1/fs/?list` returns `entries[]`
 4. Nested `mkdir` (`/team/...`) across multi-level paths
@@ -216,6 +259,8 @@ use the same value as `DRIVE9_API_KEY` here.
 14. Large multipart upload (`POST /v1/uploads/initiate` + presigned part uploads + complete + download checksum)
 
 15. Upload-limit boundary (`10GiB` initiate accepted, `10GiB+1` rejected)
+16. Cleanup test tree (existing-tenant mode only): `DELETE /v1/fs/team-${TS}?recursive`
+    removes the timestamped test tree so an existing tenant is not polluted
 
 ### `api-smoke-test-existing-key.sh`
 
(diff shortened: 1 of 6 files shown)


qiffang (approved):
## Review: APPROVE

Well-structured expansion of e2e coverage. Two new capabilities (existing-tenant regression + deep shared-mode isolation) fill real gaps. Code quality is good.

### What I checked

1. **Existing-tenant mode** (`api-smoke-test.sh`, `cli-smoke-test.sh`)
   - `DRIVE9_API_KEY` guard correctly skips provision and emits `skip_check` for accounting consistency
   - Upload-limit boundary default-OFF in existing-tenant mode is correct (quota consumed → spurious 507)
   - Explicit `RUN_UPLOAD_LIMIT_BOUNDARY=1` override still works
   - `TENANT_ID` has no downstream references outside the provision block — safe
   - Step 16 cleanup condition `[ -n "$API_KEY" ] && [ -n "${DRIVE9_API_KEY:-}" ]` correctly gates to existing-tenant only
   - `rm -f "${pfile:-}"` handles unset `pfile` in existing-tenant path — good fix

2. **Shared-mode isolation depth** (`shared-mode-smoke-test.sh`)
   - List/grep/find isolation: B cannot see A's tree, marker files, or paths — comprehensive
   - Mutation isolation: copy/hardlink/rename/delete across tenants → 404, source intact — correct
   - Delete-own-path does not affect peer's same-path file — good edge case
   - SSE non-leak: A's stream gets reset but NOT B's `sse-target.txt` event — critical test
   - Multipart session ownership: A cannot complete B's `upload_id` (404/403) — correct
   - B can still complete its own session after A's failed attempt — verifies no corruption

3. **Nested full suites** (`run_full_smoke`)
   - `env -u DRIVE9_API_KEY` forces fresh provision — correct isolation
   - Semantic/fork/SQL forced off with clear rationale for each
   - `RUN_SHARED_API_CLI_SMOKE=0` opt-out exists

4. **`smoke-all.sh`**
   - `RUN_API_ONLY` knob works correctly, uses `skip_case` helper for consistent formatting
   - `export DRIVE9_API_KEY` re-export is explicit — prevents silent inheritance loss

5. **Docs** — AGENTS.md + README.md consistent with code changes, knob table updated

### Nits (non-blocking)

**N1**: Concurrent same-path write test — `wait "$PID_A" "$PID_B" || true` swallows curl failures. If either PUT returns non-zero (network error), the read-back assertions still pass because the file might contain stale data or the previous value. Consider checking curl exit codes separately. Low risk since this is a controlled local server, but worth noting.

**N2**: `list_has_name` / `search_has_path` use `raise SystemExit` after `print("false")` on parse error — the `raise` will print a traceback to stderr but the function still returns "false" to the caller, so the test does not hang. Fine for e2e.

No blockers. Clean, well-tested expansion.

qiffang (approved):
**adversary-2 R1 — GREEN** at `5db66cb`

Reviewed all 6 files (962 additions/deletions). Two new capabilities fill real gaps:

1. **Existing-tenant regression** — correct skip-provision path, `PASS/FAIL/SKIP/TOTAL` accounting consistent via `skip_check`, upload-limit default OFF in existing-tenant mode is justified (507 spurious on consumed quota), explicit `RUN_UPLOAD_LIMIT_BOUNDARY=1` override preserved.

2. **Shared-mode cross-tenant isolation** — comprehensive: root list, grep, find, copy/hardlink/rename/delete mutation boundary, concurrent same-path writes, SSE non-leak (A stream must not receive B's file event), multipart session ownership (A cannot complete B's `upload_id`). This is the coverage that was missing.

### Checklist
- **All inputs handled?** Yes — `DRIVE9_API_KEY` set/unset branches, `RUN_SQL_CHECKS`, `RUN_SHARED_API_CLI_SMOKE`, `RUN_API_ONLY` all have explicit defaults and skip paths.
- **Edge cases covered?** B listing A-only dir accepts both 404 and 200+empty (line 624-639). Concurrent writes wait both PIDs before read. `${pfile:-}` fix prevents nounset abort in existing-tenant cleanup.
- **Failure behavior defined?** `run_full_smoke` captures exit code correctly (`exit "$FAIL"` from sub-suites). `set -e` is on throughout.
- **Race/state issues?** Concurrent curl jobs use separate output files (not `$TMP_DIR/resp.json`). SSE streams killed after explicit sleep window. No shared mutable state.
- **Tests complete?** This PR *is* the test — and it covers both happy and negative paths (B-cannot-X-A's-resource assertions).

### Nits (non-blocking)
- `api-smoke-test.sh` step 16 cleanup: if `DELETE` returns a transient 500/503, `set -e` aborts before the RESULTS summary. A `|| true` on the `check_eq` would preserve the summary line. Very minor — only matters in existing-tenant mode on a flaky server.
- `shared-mode-smoke-test.sh` `http_hdr()`: the function silently swallows an odd trailing argument (the `while $# >= 2` loop). Harmless but could mask a caller bug.

No blockers.

PR #771, opened 2026-07-21
Refactor managed cluster cache helper

srstack:
## Summary

- fold the RBAC cache path into `listAllManagedClusters`
- remove the redundant cached helper, cache toggle, cache-origin return value, and unreachable bypass metric
- add regression coverage confirming repeated managed-cluster lists use the RBAC cache

## Test plan

- `go test ./pkg/server -run "^(TestListAllManagedClustersUsesRBACCache|TestFirstManagedOrganizationUsesRBACCache)$" -count=1`
- `make lint`
- `git diff --check`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved admin tenant authorization and listing by using more reliable managed-cluster data.
  * Ensured tenant credential flows consistently resolve the correct managed organization.
  * Reduced unnecessary provider requests when retrieving managed-cluster information.

* **Tests**
  * Added coverage confirming managed-cluster results are reused correctly across repeated requests.
  * Verified cached data remains accurate during organization and tenant authorization flows.

### DIFF
--- a/pkg/server/admin_tenant_pool.go
+++ b/pkg/server/admin_tenant_pool.go
@@ -685,7 +685,7 @@ func (s *Server) firstManagedOrganization(ctx context.Context, cred tenant.Crede
 	// request path (warm-pool claim + shared-pool check) and across
 	// consecutive provisions do not each cost a TiDB Cloud list call. The
 	// cache is invalidated on tenant mutations (forgetTiDBCloudRBACList).
-	clusters, _, err := s.listAllManagedClustersCached(ctx, cred, "", true, "provision_org_resolve")
+	clusters, err := s.listAllManagedClusters(ctx, cred, "", "provision_org_resolve")
 	if err != nil || len(clusters) == 0 {
 		return "", err
 	}
--- a/pkg/server/admin_tenant_pool_cache_test.go
+++ b/pkg/server/admin_tenant_pool_cache_test.go
@@ -54,3 +54,31 @@ func TestFirstManagedOrganizationUsesRBACCache(t *testing.T) {
 		t.Fatalf("list calls after forget = %d, want 2", got)
 	}
 }
+
+func TestListAllManagedClustersUsesRBACCache(t *testing.T) {
+	rt := newQuotaRuntime(t, tenant.ProviderTiDBCloudNative)
+	rt.prov.listPages = []*tenant.ManagedClusterListResult{{
+		Clusters: []tenant.CloudClusterInfo{{ClusterID: "c1", OrganizationID: "org-1"}},
+	}}
+	ctx := context.Background()
+	cred := tenant.CredentialProvisionRequest{PublicKey: "pk", PrivateKey: "sk"}
+
+	clusters, err := rt.server.listAllManagedClusters(ctx, cred, "", "test_managed_cluster_list")
+	if err != nil {
+		t.Fatalf("first list: %v", err)
+	}
+	if len(clusters) != 1 || clusters[0].ClusterID != "c1" || clusters[0].OrganizationID != "org-1" {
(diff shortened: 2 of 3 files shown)


mornyx (commented):
Reviewed locally (checked out the branch, built, ran the targeted tests).

**Verdict: approve** — this is a behavior-preserving simplification. Verified the two claims behind it:

1. **The bypass path was unreachable.** On main, the only direct `listAllManagedClustersCached` caller passed `allowCache=true` (`authorizedAdminTenant`), and the `listAllManagedClusters` wrapper hard-coded `true`. So removing the `allowCache` parameter and the `bypass` metric label loses nothing — `bypass` could never be emitted at runtime, and no dashboard under docs/grafana references it.
2. **The cache-origin `bool` return was discarded at every call site** (`clusters, _, err` × 3). Dropping it from the signature is safe.

Also confirmed:
- All 4 call sites updated; no dangling references to `listAllManagedClustersCached`; `go build ./pkg/server` clean.
- `TestListAllManagedClustersUsesRBACCache` + `TestFirstManagedOrganizationUsesRBACCache` pass locally; CI green.
- Cache semantics unchanged: scoped lookup still only serves a hit when the cached cluster has a non-empty OrganizationID; unscoped miss still populates via `rememberClusterList`; the deleting-missing-cluster fallback in `authorizedAdminTenant` behaves exactly as before.

Two non-blocking notes:

- The branch history contains the accidental `add managed-cluster-cache` commit and its revert (net-zero). Harmless under squash merge — just flagging so nobody is confused by the commit list.
- Process FYI: that accidental commit and its revert were pushed **directly to main** (bypassing PR/CI). The revert did restore main's tree byte-for-byte to the post-#765 state (verified: `git diff cd8e004 origin/main` is empty), so no harm done — but it may be worth tightening branch protection on main to require PRs.

mornyx (approved):

srstack:
Reviewed (read-only diff). Clean behavior-preserving refactor and a natural follow-up to #764. LGTM.

This folds `listAllManagedClustersCached(cred, clusterID, allowCache, metricPath) (clusters, fromCache, err)` into `listAllManagedClusters(cred, clusterID, metricPath) (clusters, err)`. Verified it's purely subtractive with no behavior change:

- **`allowCache` is gone because it had no remaining `false` caller.** #764 removed the last bypass path (the backfill fresh-read), so the only surviving usage was `allowCache=true`. Dropping the param + the unreachable `"bypass"` metric branch is correct.
- **`fromCache bool` is gone because nothing reads it.** After #764 dropped the `clustersFromCache`-driven backfill, every call site already discarded it (`_,`). All four sites (`firstManagedOrganization`, `handleAdminTenantList`, `authorizedAdminTenant` + its deleting-missing-cluster fallback) are updated to the two-value signature.
- **Cache logic is line-for-line equivalent** to the old `allowCache=true` branch: `clusterID != ""` → `getCluster` (hit only when `OrganizationID != ""`), else `getClusterList`, else record `miss` and fall through to the lister. I checked the edge case — a `getCluster` hit with empty `OrganizationID` still falls through to the list call in both old and new code (inner `if OrganizationID != ""` doesn't return), so no regression there.
- The remember-on-miss paths (`rememberClusterList` for full list, `rememberTiDBCloudRBAC` per-cluster on the single-page cluster-scoped path) are unchanged.

`TestListAllManagedClustersUsesRBACCache` pins the intended property: two calls, second served from cache, `listCalls == 1`. Combined with the existing `TestFirstManagedOrganizationUsesRBACCache` (which also checks invalidation via forget → 2 calls), the cache + invalidation behavior stays covered.

Already carries an approval; no concerns from me. LGTM.

PR #773, opened 2026-07-21
metrics: add counter/histogram/gauge cleanup on tenant close

srstack:
## Problem

Tenant-scoped counter, histogram, and gauge metric entries accumulate unbounded over the lifetime of a process. There is no mechanism to remove entries for tenants that are no longer active.

Production observations on the TKE Beijing cluster:
- Single pod `/metrics` response exceeds 35MB (past vmagent maxScrapeSize of 32MiB)
- 7,500+ historical tenant counter entries persist in output
- vmagent suffered repeated OOMKilled restarts due to persistent queue bloat

## Root Cause

- `Int64Counter` has `Add()` but no `Delete()` — counter label sets are write-only
- `Registry.counterInstrument.values` map grows forever
- By contrast, `Float64Gauge` already exposes `Delete()`, but histograms share the same leak

## Changes

- **`pkg/metrics/registry.go`**: Added `Int64Counter.Delete()`, `deleteCounter()`, `DeleteCountersByLabel()`, `DeleteHistogramsByLabel()`, `DeleteGaugesByLabel()` with exact label-value matching
- **`pkg/metrics/operations.go`**: Added `DeleteTenantCounters(tenantID)` — cleans counters, histograms, and gauges for a given tenant
- **`pkg/tenant/pool.go`**: Calls `metrics.DeleteTenantCounters(e.tenantID)` in `closeEntry()`

When a tenant backend is closed (invalidate, shutdown, evict, or idle-reap), all per-tenant counter/histogram/gauge entries are removed from the global registry.

## Verification

- New tests: `TestDeleteTenantCounters`, `TestDeleteTenantCountersEmptyIsNoop`
- All `pkg/metrics` tests pass
- Lint passes with 0 issues

### DIFF
--- a/pkg/metrics/operations.go
+++ b/pkg/metrics/operations.go
@@ -187,6 +187,16 @@ func DeleteTenantPoolBindings(poolID, tidbCloudOrgID, poolStatus string) {
 	)
 }
 
+func DeleteTenantCounters(tenantID string) {
+	tenantID = cleanMetricValue(tenantID, "unknown")
+	if tenantID == "unknown" {
+		return
+	}
+	globalRegistry.DeleteCountersByLabel("tenant_id", tenantID)
+	globalRegistry.DeleteHistogramsByLabel("tenant_id", tenantID)
+	globalRegistry.DeleteGaugesByLabel("tenant_id", tenantID)
+}
+
 func RecordTiDBCloudRBACCacheRequest(path, scope, result string) {
 	RegisterModule("tidbcloud_quota")
 	tidbCloudRBACCacheRequestsTotal.Add(1,
--- a/pkg/metrics/operations_test.go
+++ b/pkg/metrics/operations_test.go
@@ -468,3 +468,59 @@ func TestRecordDBOperationOmitsTenantID(t *testing.T) {
 		t.Fatalf("db operation metrics must not carry tenant_id:\n%s", text)
 	}
 }
+
+func TestDeleteTenantCounters(t *testing.T) {
+	const tenantID = "tenant-delete-counter-test"
+	const tidbCloudOrgID = "org-delete-counter"
+
+	RecordTenantOperationCountWithOrg(tenantID, tidbCloudOrgID, "cmp", "op", "ok")
+	RecordTenantRequestCountWithOrg(tenantID, tidbCloudOrgID, "api", "read", "ok", 200)
+	RecordTenantHTTPBytesWithOrg(tenantID, tidbCloudOrgID, "api", "", "upload", 1024)
+	RecordSSEConnectionWithOrg(tenantID, tidbCloudOrgID, "mount", time.Second)
+	RecordSSEEventSentWithOrg(tenantID, tidbCloudOrgID, "write")
+
+	rec := httptest.NewRecorder()
+	WritePrometheus(rec)
+	text := rec.Body.String()
+	if !strings.Contains(text, `tenant_id="`+tenantID+`"`) {
+		t.Fatalf("tenant_id not found before delete:\n%s", text)
+	}
+
+	DeleteTenantCounters(tenantID)
+
+	rec2 := httptest.NewRecorder()
+	WritePrometheus(rec2)
+	text2 := rec2.Body.String()
+
+	counterNames := []string{
+		"drive9_service_operations_total",
+		"drive9_tenant_requests_total",
+		"drive9_tenant_http_bytes_total",
+		"drive9_sse_connections_total",
+		"drive9_sse_events_sent_total",
+	}
+	histogramNames := []string{
+		"drive9_sse_connection_duration_seconds_bucket",
+		"drive9_sse_connection_duration_seconds_count",
+		"drive9_sse_connection_duration_seconds_sum",
+	}
+	for _, line := range strings.Split(text2, "\n") {
+		if !strings.Contains(line, `tenant_id="`+tenantID+`"`) {
+			continue
+		}
+		for _, name := range counterNames {
+			if strings.HasPrefix(line, name) {
+				t.Fatalf("counter line with tenant_id still present: %s", line)
+			}
+		}
+		for _, name := range histogramNames {
+			if strings.HasPrefix(line, name) {
+				t.Fatalf("histogram line with tenant_id still present: %s", line)
+			}
+		}
+	}
+}
+
+func TestDeleteTenantCountersEmptyIsNoop(t *testing.T) {
+	DeleteTenantCounters("")
+}
--- a/pkg/metrics/operations_test.go
+++ b/pkg/metrics/operations_test.go
@@ -478,6 +478,7 @@ func TestDeleteTenantCounters(t *testing.T) {
 	RecordTenantHTTPBytesWithOrg(tenantID, tidbCloudOrgID, "api", "", "upload", 1024)
 	RecordSSEConnectionWithOrg(tenantID, tidbCloudOrgID, "mount", time.Second)
 	RecordSSEEventSentWithOrg(tenantID, tidbCloudOrgID, "write")
+	RecordTenantInFlightWithOrg(tenantID, tidbCloudOrgID, "api", "read", 3)
 
 	rec := httptest.NewRecorder()
 	WritePrometheus(rec)
@@ -504,20 +505,35 @@ func TestDeleteTenantCounters(t *testing.T) {
 		"drive9_sse_connection_duration_seconds_count",
 		"drive9_sse_connection_duration_seconds_sum",
 	}
+	gaugeNames := []string{
+		"drive9_tenant_inflight_requests",
+	}
+	var failures int
 	for _, line := range strings.Split(text2, "\n") {
 		if !strings.Contains(line, `tenant_id="`+tenantID+`"`) {
 			continue
 		}
 		for _, name := range counterNames {
 			if strings.HasPrefix(line, name) {
-				t.Fatalf("counter line with tenant_id still present: %s", line)
+				t.Errorf("counter line with tenant_id still present: %s", line)
+				failures++
 			}
 		}
 		for _, name := range histogramNames {
 			if strings.HasPrefix(line, name) {
-				t.Fatalf("histogram line with tenant_id still present: %s", line)
+				t.Errorf("histogram line with tenant_id still present: %s", line)
+				failures++
 			}
 		}
+		for _, name := range gaugeNames {
+			if strings.HasPrefix(line, name) {
+				t.Errorf("gauge line with tenant_id still present: %s", line)
+				failures++
+			}
+		}
+	}
+	if failures > 0 {
+		t.Fatalf("%d tenant-scoped lines leaked after DeleteTenantCounters", failures)
 	}
 }
 
--- a/pkg/metrics/registry.go
+++ b/pkg/metrics/registry.go
@@ -311,7 +311,12 @@ func (r *Registry) deleteGauge(name, labels string) {
 }
 
 func labelHasKeyValue(labels, match string) bool {
-	return strings.HasSuffix(labels, match) || strings.Contains(labels, match+",")
+	if labels == match {
+		return true
+	}
+	return strings.HasPrefix(labels, match+`,`) ||
+		strings.Contains(labels, `,`+match+`,`) ||
+		strings.HasSuffix(labels, `,`+match)
 }
 
 func (r *Registry) DeleteCountersByLabel(labelKey, labelValue string) {
--- a/pkg/metrics/registry.go
+++ b/pkg/metrics/registry.go
@@ -156,6 +156,13 @@ func (c *Int64Counter) Add(value int64, attrs ...Attribute) {
 	c.registry.addCounter(c.name, labelsKey(attrs), value)
 }
 
+func (c *Int64Counter) Delete(attrs ...Attribute) {
+	if c == nil || c.registry == nil {
+		return
+	}
+	c.registry.deleteCounter(c.name, labelsKey(attrs))
+}
+
 func (g *Float64Gauge) Set(value float64, attrs ...Attribute) {
 	if g == nil || g.registry == nil {
 		return
@@ -283,6 +290,16 @@ func (r *Registry) setGauge(name, labels string, value float64) {
 	inst.values[labels] = value
 }
 
+func (r *Registry) deleteCounter(name, labels string) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	inst := r.counters[name]
+	if inst == nil {
+		return
+	}
+	delete(inst.values, labels)
+}
+
 func (r *Registry) deleteGauge(name, labels string) {
 	r.mu.Lock()
 	defer r.mu.Unlock()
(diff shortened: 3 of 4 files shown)


srstack:
Reviewed (read-only diff). Solves a real production problem (35MB `/metrics` → vmagent OOM from unbounded per-tenant counter accumulation), and the implementation is correct. **One robustness point on the label matcher worth addressing; not a live bug today.** LGTM otherwise.

## ✅ Correct and well-guarded

- **Concurrency**: all three `DeleteXxxByLabel` hold `r.mu.Lock()`; deleting the current key during a `range` over a Go map is allowed, so the in-loop `delete(inst.values, labels)` is safe.
- **The `unknown`/empty guard is important and right**: `DeleteTenantCounters` returns early for `""`/`"unknown"`. Without it, closing one tenant would nuke every aggregate series that carries no `tenant_id` (they normalize to `tenant_id="unknown"`). Good that it's there.
- **Call site**: `closeEntry` covers invalidate / shutdown / evict / idle-reap — all the paths where a tenant leaves the pool. Deleting gauges here is also correct: a closed tenant won't `Set()` again, so the entry would otherwise be stale forever.
- **Tests** cover counters + all three histogram families (`_bucket`/`_count`/`_sum`) and the empty no-op.

## 🟡 `labelHasKeyValue` matches on the wrong boundary (latent fragility)

```go
func labelHasKeyValue(labels, match string) bool {
    return strings.HasSuffix(labels, match) || strings.Contains(labels, match+",")
}
```
with `match = tenant_id="xxx"`. `labelsKey` sorts keys alphabetically and joins with `,`, so a label line looks like `component="backend",tenant_id="xxx"`. The matcher only anchors the **right** boundary (suffix, or followed by `,`). It does **not** anchor the left boundary, so any label key that *ends in* `tenant_id` — e.g. a future `parent_tenant_id` / `src_tenant_id` — would produce `...,parent_tenant_id="xxx"` which contains the substring `tenant_id="xxx"` and would be **falsely matched and deleted**.

I checked the current label namespace (`grep 'Attr("...tenant_id"'`) — today every key is exactly `tenant_id`, no `*_tenant_id` variant, so **this is not a live bug**. But the correctness silently depends on that invariant; the day someone adds a `parent_tenant_id` label the deletion becomes over-broad and hard to debug. Suggest anchoring the left side too, e.g. match when the key-value is at string start or preceded by `,`:
```go
return labels == match ||
    strings.HasPrefix(labels, match+",") ||
    strings.HasSuffix(labels, ","+match) ||
    strings.Contains(labels, ","+match+",")
```
(or just parse the labels once.) Cheap insurance against a future footgun.

## Minor

- `DeleteCountersByLabel`/`Histograms`/`Gauges` iterate **all** instruments and all their label sets on every tenant close. With 7,500+ tenants churning that's O(instruments × series) per close — fine at current scale and far cheaper than the 35MB scrape it fixes, but worth a note if tenant close frequency ever spikes. Non-blocking.

Net: correct fix for a real OOM, safe under lock, guarded against the `unknown` footgun, and tested. I'd like the left-boundary anchoring in `labelHasKeyValue` before this compounds, but it's not a blocker given the current label set. LGTM.

mornyx (commented):
## PR #773 Review — Metrics: add counter/histogram/gauge cleanup on tenant close

### Summary

This PR adds the ability to delete per-tenant metric entries (counters, histograms, gauges) from the global registry, and calls the cleanup when a tenant backend is closed via `closeEntry()`. This addresses a production issue where `/metrics` responses exceeded 35MB (vmagent's 32MiB maxScrapeSize), causing vmagent OOMKilled restarts.

1. **`Int64Counter.Delete()`** (`registry.go`): Adds a `Delete` method on `Int64Counter`, mirroring the existing `Float64Gauge.Delete`. Delegates to `Registry.deleteCounter`. ✅

2. **`DeleteCountersByLabel`, `DeleteHistogramsByLabel`, `DeleteGaugesByLabel`** (`registry.go`): Bulk deletion functions that iterate all instruments of a kind and delete label sets matching a given key=value pair. Uses `labelHasKeyValue` for comma-delimited boundary matching. ✅

3. **`labelHasKeyValue`** (`registry.go`): Correctly matches a `key="value"` substring within a comma-separated label string, handling all positions (only label, first, middle, last). This prevents false matches like `tenant_id="tenant-1"` matching `tenant_id="tenant-10"`. ✅

4. **`DeleteTenantCounters`** (`operations.go`): Convenience function that calls all three bulk-delete functions with `tenant_id=<tenantID>`. Guards against deleting `tenant_id="unknown"` (which would delete aggregate/fallback metrics). ✅

5. **`closeEntry` → `p.closeEntry`** (`pool.go`): Converted from standalone function to Pool method. After closing the backend and store, checks `p.items[tenantID]` under `p.mu` — only calls `DeleteTenantCounters` if the tenant is no longer active in the pool. This prevents deleting metrics for a tenant that has a replacement entry (e.g., cache race where a new backend was inserted for the same tenant). ✅

6. **`releaseEntry` nil guard** (`pool.go`): Added `if toClose != nil` guard before calling `p.closeEntry`, avoiding the nil check inside `closeEntry` for the release path. ✅ Minor but correct.

7. **All `closeEntry` call sites updated**: 9 call sites across `Get`, `Acquire`, `Invalidate`, `InvalidateAndWait`, `reapOnce`, `Close`, and `releaseEntry` all changed from `closeEntry(...)` to `p.closeEntry(...)`. ✅ Complete.

### Test coverage

- `TestDeleteTenantCounters`: Records 6 different metric types (service operations, tenant requests, HTTP bytes, SSE connections, SSE events, inflight gauge) for a tenant, verifies they appear in `/metrics`, calls `DeleteTenantCounters`, then verifies all are gone. Checks counters, histograms, and gauges. ✅ Thorough.
- `TestDeleteTenantCountersEmptyIsNoop`: Verifies that calling with `""` (which resolves to `"unknown"`) is a no-op. ✅

### CI status

- `ci`: ✅ success

---

### Comments

#### 1. Counter deletion causes Prometheus `rate()` discontinuities (important observation, acceptable trade-off)

`DeleteTenantCounters` removes counter entries for `drive9_service_operations_total`, `drive9_tenant_requests_total`, `drive9_tenant_http_bytes_total`, `drive9_sse_connections_total`, etc. In Prometheus, `rate()` computes the delta between scrapes. When a counter disappears (value drops to 0 or the series vanishes), `rate()` will see a negative delta for one scrape window, potentially producing a negative spike or NaN in dashboards.

This is an inherent consequence of deleting counter series. The dashboards that aggregate with `sum by (component)` or `sum by (result)` will smooth over the transient spike as long as other tenants are active. For idle tenants being evicted/reaped, the counter was not growing anyway, so the rate was already ~0. The trade-off (35MB → bounded) clearly justifies this.

**Recommendation**: Consider documenting this in a runbook or dashboard note — "tenant metric series may disappear when a tenant backend is evicted; expect transient rate discontinuities for evicted tenants."

#### 2. Race window between `closeEntry` check and `DeleteTenantCounters` (observation, benign)

In `closeEntry`, after closing the backend/store, the code acquires `p.mu`, checks `p.items[tenantID]`, releases `p.mu`, and then calls `DeleteTenantCounters`. Between the release and the delete call, another goroutine could insert a new entry for the same tenant (e.g., `Get` or `Acquire`) and start recording metrics. `DeleteTenantCounters` would then delete those freshly-created metrics.

This is a narrow window and the consequence is benign — the metrics are recreated on the next record call. No data corruption, just a transient metric loss. Given the production urgency (vmagent OOM), this trade-off is acceptable. If strict correctness is needed later, `DeleteTenantCounters` could be called under `p.mu` (since `metrics.RecordGauge` uses a separate lock, deadlock is not a risk).

#### 3. `DeleteTenantCounters` deletes ALL metrics with `tenant_id=X`, not just tenant-pool-managed ones (observation, correct but broad)

The function deletes any counter/histogram/gauge entry with `tenant_id=X` across all instruments. This includes:
- Tenant pool metrics (`drive9_tenant_pool_*`)
- SSE metrics (`drive9_sse_*`)
- Service operations (`drive9_service_operations_total`)
- Tenant usage metrics (`drive9_tenant_requests_total`, `drive9_tenant_http_bytes_total`, etc.)

This is the correct behavior — the goal is to clean up all per-tenant series. The `tenant_id="unknown"` guard prevents deleting aggregate entries. ✅

However, if a tenant has active SSE connections when its backend is evicted from the pool, the SSE connection counter/duration entries will be deleted even though SSE activity may still be ongoing. The entries will be recreated on the next SSE event, but the cumulative connection count resets. This is a minor observability gap — SSE connection lifetime counters will undercount for tenants whose backends are evicted while SSE connections are active.

#### 4. `labelHasKeyValue` correctness with escaped values (observation, positive)

The match string uses `EscapePromLabel(labelValue)` which handles backslash, double-quote, and newline escaping. Since `labelsKey` also escapes values with the same function, the match will correctly find escaped values. If a tenant ID contained a `"` (unlikely but possible), the escape would be consistent on both sides. ✅

#### 5. Histogram deletion deletes `_bucket`, `_count`, and `_sum` entries (observation, correct)

The histogram instrument stores all three (`_bucket`, `_count`, `_sum`) in a single `histogramSample` per label set. `DeleteHistogramsByLabel` deletes the entire `histogramSample` from `inst.values`, so all three components are removed atomically. No partial deletion risk. ✅

#### 6. `Int64Counter.Delete` method is added but not used directly (observation)

The `Int64Counter.Delete` method is defined but the PR uses the bulk `DeleteCountersByLabel` path instead. The `Delete` method is a useful primitive for future fine-grained cleanup (e.g., the existing `Float64Gauge.Delete` used by `RecordTenantInFlight` for zero-value removal). No issue — just noting it is currently unused.

---

### Verdict: ✅ Approve

This is a necessary fix for a real production problem (vmagent OOM from 35MB `/metrics` responses). The implementation is correct: boundary-aware label matching, `tenant_id="unknown"` guard, active-entry check to avoid deleting replacement metrics, and comprehensive test coverage. The counter `rate()` discontinuity trade-off is well-justified by the production impact. CI passes on both checks.

mornyx (approved):
## PR #773 Review — Metrics: add counter/histogram/gauge cleanup on tenant close

### Summary

This PR adds the ability to delete per-tenant metric entries (counters, histograms, gauges) from the global registry, and calls the cleanup when a tenant backend is closed via `closeEntry()`. This addresses a production issue where `/metrics` responses exceeded 35MB (vmagent's 32MiB maxScrapeSize), causing vmagent OOMKilled restarts.

1. **`Int64Counter.Delete()`** (`registry.go`): Adds a `Delete` method on `Int64Counter`, mirroring the existing `Float64Gauge.Delete`. Delegates to `Registry.deleteCounter`. ✅

2. **`DeleteCountersByLabel`, `DeleteHistogramsByLabel`, `DeleteGaugesByLabel`** (`registry.go`): Bulk deletion functions that iterate all instruments of a kind and delete label sets matching a given key=value pair. Uses `labelHasKeyValue` for comma-delimited boundary matching. ✅

3. **`labelHasKeyValue`** (`registry.go`): Correctly matches a `key="value"` substring within a comma-separated label string, handling all positions (only label, first, middle, last). This prevents false matches like `tenant_id="tenant-1"` matching `tenant_id="tenant-10"`. ✅

4. **`DeleteTenantCounters`** (`operations.go`): Convenience function that calls all three bulk-delete functions with `tenant_id=<tenantID>`. Guards against deleting `tenant_id="unknown"` (which would delete aggregate/fallback metrics). ✅

5. **`closeEntry` → `p.closeEntry`** (`pool.go`): Converted from standalone function to Pool method. After closing the backend and store, checks `p.items[tenantID]` under `p.mu` — only calls `DeleteTenantCounters` if the tenant is no longer active in the pool. This prevents deleting metrics for a tenant that has a replacement entry (e.g., cache race where a new backend was inserted for the same tenant). ✅

6. **`releaseEntry` nil guard** (`pool.go`): Added `if toClose != nil` guard before calling `p.closeEntry`, avoiding the nil check inside `closeEntry` for the release path. ✅ Minor but correct.

7. **All `closeEntry` call sites updated**: 9 call sites across `Get`, `Acquire`, `Invalidate`, `InvalidateAndWait`, `reapOnce`, `Close`, and `releaseEntry` all changed from `closeEntry(...)` to `p.closeEntry(...)`. ✅ Complete.

### Test coverage

- `TestDeleteTenantCounters`: Records 6 different metric types (service operations, tenant requests, HTTP bytes, SSE connections, SSE events, inflight gauge) for a tenant, verifies they appear in `/metrics`, calls `DeleteTenantCounters`, then verifies all are gone. Checks counters, histograms, and gauges. ✅ Thorough.
- `TestDeleteTenantCountersEmptyIsNoop`: Verifies that calling with `""` (which resolves to `"unknown"`) is a no-op. ✅

### CI status

- `ci`: ✅ success

---

### Comments

#### 1. Counter deletion causes Prometheus `rate()` discontinuities (important observation, acceptable trade-off)

`DeleteTenantCounters` removes counter entries for `drive9_service_operations_total`, `drive9_tenant_requests_total`, `drive9_tenant_http_bytes_total`, `drive9_sse_connections_total`, etc. In Prometheus, `rate()` computes the delta between scrapes. When a counter disappears (value drops to 0 or the series vanishes), `rate()` will see a negative delta for one scrape window, potentially producing a negative spike or NaN in dashboards.

This is an inherent consequence of deleting counter series. The dashboards that aggregate with `sum by (component)` or `sum by (result)` will smooth over the transient spike as long as other tenants are active. For idle tenants being evicted/reaped, the counter was not growing anyway, so the rate was already ~0. The trade-off (35MB → bounded) clearly justifies this.

**Recommendation**: Consider documenting this in a runbook or dashboard note — "tenant metric series may disappear when a tenant backend is evicted; expect transient rate discontinuities for evicted tenants."

#### 2. Race window between `closeEntry` check and `DeleteTenantCounters` (observation, benign)

In `closeEntry`, after closing the backend/store, the code acquires `p.mu`, checks `p.items[tenantID]`, releases `p.mu`, and then calls `DeleteTenantCounters`. Between the release and the delete call, another goroutine could insert a new entry for the same tenant (e.g., `Get` or `Acquire`) and start recording metrics. `DeleteTenantCounters` would then delete those freshly-created metrics.

This is a narrow window and the consequence is benign — the metrics are recreated on the next record call. No data corruption, just a transient metric loss. Given the production urgency (vmagent OOM), this trade-off is acceptable. If strict correctness is needed later, `DeleteTenantCounters` could be called under `p.mu` (since `metrics.RecordGauge` uses a separate lock, deadlock is not a risk).

#### 3. `DeleteTenantCounters` deletes ALL metrics with `tenant_id=X`, not just tenant-pool-managed ones (observation, correct but broad)

The function deletes any counter/histogram/gauge entry with `tenant_id=X` across all instruments. This includes:
- Tenant pool metrics (`drive9_tenant_pool_*`)
- SSE metrics (`drive9_sse_*`)
- Service operations (`drive9_service_operations_total`)
- Tenant usage metrics (`drive9_tenant_requests_total`, `drive9_tenant_http_bytes_total`, etc.)

This is the correct behavior — the goal is to clean up all per-tenant series. The `tenant_id="unknown"` guard prevents deleting aggregate entries. ✅

However, if a tenant has active SSE connections when its backend is evicted from the pool, the SSE connection counter/duration entries will be deleted even though SSE activity may still be ongoing. The entries will be recreated on the next SSE event, but the cumulative connection count resets. This is a minor observability gap — SSE connection lifetime counters will undercount for tenants whose backends are evicted while SSE connections are active.

#### 4. `labelHasKeyValue` correctness with escaped values (observation, positive)

The match string uses `EscapePromLabel(labelValue)` which handles backslash, double-quote, and newline escaping. Since `labelsKey` also escapes values with the same function, the match will correctly find escaped values. If a tenant ID contained a `"` (unlikely but possible), the escape would be consistent on both sides. ✅

#### 5. Histogram deletion deletes `_bucket`, `_count`, and `_sum` entries (observation, correct)

The histogram instrument stores all three (`_bucket`, `_count`, `_sum`) in a single `histogramSample` per label set. `DeleteHistogramsByLabel` deletes the entire `histogramSample` from `inst.values`, so all three components are removed atomically. No partial deletion risk. ✅

#### 6. `Int64Counter.Delete` method is added but not used directly (observation)

The `Int64Counter.Delete` method is defined but the PR uses the bulk `DeleteCountersByLabel` path instead. The `Delete` method is a useful primitive for future fine-grained cleanup (e.g., the existing `Float64Gauge.Delete` used by `RecordTenantInFlight` for zero-value removal). No issue — just noting it is currently unused.

---

### Verdict: ✅ Approve

This is a necessary fix for a real production problem (vmagent OOM from 35MB `/metrics` responses). The implementation is correct: boundary-aware label matching, `tenant_id="unknown"` guard, active-entry check to avoid deleting replacement metrics, and comprehensive test coverage. The counter `rate()` discontinuity trade-off is well-justified by the production impact. CI passes on both checks.

PR #774, opened 2026-07-22
feat: add TiDB Cloud Native shared DB control plane

srstack:
## Summary
- add `tidb_cloud_native_shared` provisioning with managed physical DB pools, compact placement, tenant-pool membership, batch cluster creation, and soft/hard capacity hysteresis
- dispatch existing tenant lifecycle operations from persisted provider state; shared deletion purges only the tenant fs/S3 scope, while shared fork and database auto embedding remain unsupported
- add virtual tenant spending-limit compatibility, shared DB pool metrics, lazy shared-schema initialization, and reference-aware shared connection pooling

## Validation
- `go test ./pkg/mysqlutil ./pkg/tenant -count=1`
- `go test ./cmd/drive9-server -count=1`
- `go test ./pkg/server -run "(Shared|TenantPool|Quota)" -count=1`
- `go build ./...`
- `make lint`
- `git diff --check`

## Notes
- `docs/superpowers/` is intentionally excluded from this PR.
- Related alert coverage: https://github.com/tidbcloud/runbooks/pull/2685

## Summary by CodeRabbit

* **New Features**
  * Added end-to-end support for the `tidb_cloud_native_shared` tenant provider, including `schema dump-init-sql` and provisioning flow.
  * Configurable shared-DB capacity and spending behavior via new environment variables (hard-cap ratios, reopen ratio, max tenants, default spending limit).
  * Added shared DB pool metrics (capacity, tenant counts, and spending headroom) and improved shared DB connection/schema readiness handling.

* **Bug Fixes**
  * Improved shared-pool tenant allocation, emergency/normal capacity handling, and reopen-ratio-aware deletion.
  * Shared quota GET/SET now targets shared DB spending limits and returns clear conflict status when spending is exceeded.

* **Tests**
  * Expanded coverage for shared provisioning, quotas, pool lifecycle, metrics, schema behavior, and environment parsing/validation.

### Rolling upgrade note

`quota_limits_overridden` remains physically present in this release for rolling-upgrade and rollback safety. New code ignores the column and enforces the persisted quota values, including for existing rows; the physical column drop is deferred to a later release after old binaries are no longer running.

### DIFF
--- a/cmd/drive9-server/main.go
+++ b/cmd/drive9-server/main.go
@@ -679,7 +679,7 @@ environment:
   DRIVE9_TIDBCLOUD_NATIVE_PRIVATE_KEY optional default TiDB Cloud API private key for tidb_cloud_native create/delete when caller omits it
   DRIVE9_TIDBCLOUD_NATIVE_USE_PRIVATE_ENDPOINT true|false to use TiDB Cloud private endpoint hosts for tidb_cloud_native
   DRIVE9_TIDBCLOUD_NATIVE_SHARED_MAX_TENANTS soft capacity for new managed shared DB pools (default: 100)
-  DRIVE9_TIDBCLOUD_NATIVE_SHARED_HARD_CAP_RATIO emergency hard-cap ratio after physical create failure (default: 1.2)
+  DRIVE9_TIDBCLOUD_NATIVE_SHARED_HARD_CAP_RATIO emergency hard-cap ratio > 1 after physical create failure (default: 1.2)
   DRIVE9_TIDBCLOUD_NATIVE_SHARED_REOPEN_RATIO reopen ratio for a latched shared pool (default: 0.8)
   DRIVE9_TIDBCLOUD_NATIVE_DB_POOL_DEFAULT_SPENDING_LIMIT default managed shared DB-pool spending target (default: 10000000)
   DRIVE9_TIDBCLOUD_PRIVATE_ENDPOINT_HOST_MAP comma-separated public_host=private_host mappings (also accepts public_host:private_host);
@@ -1192,8 +1192,8 @@ func sharedDBHardCapRatioFromEnv() (float64, error) {
 		return server.DefaultSharedDBHardCapRatio, nil
 	}
 	v, err := strconv.ParseFloat(raw, 64)
-	if err != nil || v < 1 || math.IsNaN(v) || math.IsInf(v, 0) {
-		return 0, fmt.Errorf("invalid DRIVE9_TIDBCLOUD_NATIVE_SHARED_HARD_CAP_RATIO=%q: must be a finite number >= 1", raw)
+	if err != nil || v <= 1 || math.IsNaN(v) || math.IsInf(v, 0) {
+		return 0, fmt.Errorf("invalid DRIVE9_TIDBCLOUD_NATIVE_SHARED_HARD_CAP_RATIO=%q: must be a finite number > 1", raw)
 	}
 	return v, nil
 }
--- a/cmd/drive9-server/main.go
+++ b/cmd/drive9-server/main.go
@@ -12,6 +12,7 @@ import (
 	"encoding/hex"
 	"fmt"
 	"io"
+	"math"
 	"net"
 	"os"
 	"strconv"
@@ -187,6 +188,22 @@ func main() {
 	if err != nil {
 		die(err)
 	}
+	sharedDBHardCapRatio, err := sharedDBHardCapRatioFromEnv()
+	if err != nil {
+		die(err)
+	}
+	sharedDBReopenRatio, err := sharedDBReopenRatioFromEnv()
+	if err != nil {
+		die(err)
+	}
+	sharedDBMaxTenants, err := sharedDBMaxTenantsFromEnv()
+	if err != nil {
+		die(err)
+	}
+	sharedDBDefaultSpendingLimit, err := sharedDBDefaultSpendingLimitFromEnv()
+	if err != nil {
+		die(err)
+	}
 	maxUploadBytes := server.DefaultMaxUploadBytes
 	if raw := os.Getenv("DRIVE9_MAX_UPLOAD_BYTES"); raw != "" {
 		maxUploadBytes, err = strconv.ParseInt(raw, 10, 64)
@@ -218,14 +235,14 @@ func main() {
 	switch providerType {
 	case tenant.ProviderTiDBZero:
 		provisioner, provisionerErr = tidbzero.NewProvisionerFromEnv()
-	case tenant.ProviderTiDBCloudNative:
+	case tenant.ProviderTiDBCloudNative, tenant.ProviderTiDBCloudNativeShared:
 		provisioner, provisionerErr = tidbcloudnative.NewProvisionerFromEnv()
 	case tenant.ProviderDB9:
 		provisioner, provisionerErr = db9.NewProvisionerFromEnv()
 	}
 	if provisionerErr != nil {
 		isWanted := false
-		if providerType == tenant.ProviderTiDBCloudNative && os.Getenv("DRIVE9_TIDBCLOUD_NATIVE_API_URL") != "" {
+		if tenant.UsesTiDBCloudNativeCredentials(providerType) && os.Getenv("DRIVE9_TIDBCLOUD_NATIVE_API_URL") != "" {
 			isWanted = true
 		}
 		if isWanted {
@@ -428,6 +445,11 @@ func main() {
 		Meta:                            store,
 		Pool:                            pool,
 		Provisioner:                     provisioner,
+		DefaultTenantProvider:           providerType,
+		SharedDBMaxTenants:              sharedDBMaxTenants,
+		SharedDBHardCapRatio:            sharedDBHardCapRatio,
+		SharedDBReopenRatio:             sharedDBReopenRatio,
+		SharedDBDefaultSpendingLimit:    sharedDBDefaultSpendingLimit,
 		LegacyStarterProvisioner:        legacyStarterProvisioner,
 		TokenSecret:                     tokenSecret,
 		VaultMasterKey:                  vaultMasterKey,
@@ -498,14 +520,14 @@ func registerSharedPoolFromEnv(ctx context.Context, metaStore *meta.Store, pool 
 		tlsMode = ""
 	}
 	_, err = metaStore.RegisterSharedDB(ctx, &meta.SharedDB{
-		OrgID:          orgID,
-		Role:           meta.SharedDBRoleShared,
-		Host:           host,
-		Port:           port,
-		User:           cfg.User,
-		PasswordCipher: passCipher,
-		Name:           cfg.DBName,
-		TLSMode:        tlsMode,
+		TiDBCloudOrganizationID: orgID,
+		Role:                    meta.SharedDBRoleShared,
+		Host:                    host,
+		Port:                    port,
+		User:                    cfg.User,
+		PasswordCipher:          passCipher,
+		Name:                    cfg.DBName,
+		TLSMode:                 tlsMode,
 	})
 	return err
 }
@@ -590,7 +612,7 @@ func usage(out io.Writer, exitCode int) {
 	_, _ = fmt.Fprintf(out, `usage:
   drive9-server [listen-addr]
   drive9-server version
-  drive9-server schema dump-init-sql --provider <db9|tidb_zero|tidb_cloud_native>
+  drive9-server schema dump-init-sql --provider <db9|tidb_zero|tidb_cloud_native|tidb_cloud_native_shared>
 
 environment:
   DRIVE9_LISTEN_ADDR serve listen address (default: :9009)
@@ -607,8 +629,10 @@ environment:
   DRIVE9_USER_SCHEMA_DB_MAX_IDLE_CONNS max idle connections for tenant schema-init DB pools (default: 2)
   DRIVE9_SHARED_POOL_DSN register a shared-schema DB at startup and place matching tenants on it (empty = disabled)
   DRIVE9_SHARED_POOL_ORG TiDB Cloud org the shared pool serves (required when DSN is set; '*' = all orgs, loud warn)
-  DRIVE9_SHARED_DB_MAX_OPEN_CONNS max open connections for each shared-schema DB handle (default: 50)
-  DRIVE9_SHARED_DB_MAX_IDLE_CONNS max idle connections for each shared-schema DB handle (default: 10)
+  DRIVE9_SHARED_DB_MAX_OPEN_CONNS max open connections for each shared-schema DB handle (default: 300)
+  DRIVE9_SHARED_DB_MAX_IDLE_CONNS max idle connections for each shared-schema DB handle (default: 50)
+  DRIVE9_SHARED_DB_CONN_MAX_LIFETIME maximum shared connection lifetime (default: 30m)
+  DRIVE9_SHARED_DB_CONN_MAX_IDLE_TIME maximum shared connection idle time (default: 5m)
   DRIVE9_DB_HEALTH_PROBE_INTERVAL_SECONDS DB health probe interval seconds (default: 15)
   DRIVE9_DB_HEALTH_PROBE_TIMEOUT_SECONDS DB health probe timeout seconds (default: 3)
   DRIVE9_DB_HEALTH_PROBE_META_ENABLED true|false to probe the control-plane DB (default: true)
@@ -645,7 +669,7 @@ environment:
                                     required by openai, cohere, jina_ai, gemini, huggingface, nvidia_nim, nvidia
   DRIVE9_TIDB_AUTO_EMBEDDING_API_BASE provider base endpoint for models that require it
                                      optional for openai models; set it for Azure OpenAI endpoints
-  DRIVE9_TENANT_PROVIDER db9|tidb_zero|tidb_cloud_native (default for provisioning)
+  DRIVE9_TENANT_PROVIDER db9|tidb_zero|tidb_cloud_native|tidb_cloud_native_shared (default for provisioning)
   DRIVE9_TIDBCLOUD_DEFAULT_SPENDING_LIMIT default TiDB Cloud Cluster spendingLimit.monthly; native defaults to 1000 when unset
   DRIVE9_TIDBCLOUD_NATIVE_API_URL TiDB Cloud Cluster API base URL for tidb_cloud_native
   DRIVE9_TIDBCLOUD_NATIVE_CLOUD_PROVIDER cloud provider for tidb_cloud_native cluster creation, e.g. aws
@@ -654,6 +678,10 @@ environment:
   DRIVE9_TIDBCLOUD_NATIVE_PUBLIC_KEY optional default TiDB Cloud API public key for tidb_cloud_native create/delete when caller omits it
   DRIVE9_TIDBCLOUD_NATIVE_PRIVATE_KEY optional default TiDB Cloud API private key for tidb_cloud_native create/delete when caller omits it
   DRIVE9_TIDBCLOUD_NATIVE_USE_PRIVATE_ENDPOINT true|false to use TiDB Cloud private endpoint hosts for tidb_cloud_native
+  DRIVE9_TIDBCLOUD_NATIVE_SHARED_MAX_TENANTS soft capacity for new managed shared DB pools (default: 100)
+  DRIVE9_TIDBCLOUD_NATIVE_SHARED_HARD_CAP_RATIO emergency hard-cap ratio after physical create failure (default: 1.2)
+  DRIVE9_TIDBCLOUD_NATIVE_SHARED_REOPEN_RATIO reopen ratio for a latched shared pool (default: 0.8)
+  DRIVE9_TIDBCLOUD_NATIVE_DB_POOL_DEFAULT_SPENDING_LIMIT default managed shared DB-pool spending target (default: 10000000)
   DRIVE9_TIDBCLOUD_PRIVATE_ENDPOINT_HOST_MAP comma-separated public_host=private_host mappings (also accepts public_host:private_host);
                                              when set, disables legacy single-host private endpoint overrides
   DRIVE9_TIDBCLOUD_TENCENT_PRIVATE_ENDPOINT_HOST legacy tencentcloud private endpoint fallback host, used only when host map is unset
@@ -1154,6 +1182,54 @@ func tenantPoolRefillFreeRatioFromEnv() (float64, error) {
 	v, err := strconv.ParseFloat(raw, 64)
 	if err != nil || v <= 0 || v > 1 || v != v {
 		return 0, fmt.Errorf("invalid DRIVE9_TENANT_POOL_REFILL_FREE_RATIO=%q: must be a number in (0,1]", raw)
+	}
+	return v, nil
+}
+
+func sharedDBHardCapRatioFromEnv() (float64, error) {
+	raw := strings.TrimSpace(os.Getenv("DRIVE9_TIDBCLOUD_NATIVE_SHARED_HARD_CAP_RATIO"))
+	if raw == "" {
+		return server.DefaultSharedDBHardCapRatio, nil
+	}
+	v, err := strconv.ParseFloat(raw, 64)
+	if err != nil || v < 1 || math.IsNaN(v) || math.IsInf(v, 0) {
+		return 0, fmt.Errorf("invalid DRIVE9_TIDBCLOUD_NATIVE_SHARED_HARD_CAP_RATIO=%q: must be a finite number >= 1", raw)
+	}
+	return v, nil
+}
+
+func sharedDBReopenRatioFromEnv() (float64, error) {
+	raw := strings.TrimSpace(os.Getenv("DRIVE9_TIDBCLOUD_NATIVE_SHARED_REOPEN_RATIO"))
+	if raw == "" {
+		return server.DefaultSharedDBReopenRatio, nil
+	}
+	v, err := strconv.ParseFloat(raw, 64)
+	if err != nil || v <= 0 || v >= 1 || math.IsNaN(v) || math.IsInf(v, 0) {
+		return 0, fmt.Errorf("invalid DRIVE9_TIDBCLOUD_NATIVE_SHARED_REOPEN_RATIO=%q: must be a finite number in (0,1)", raw)
+	}
+	return v, nil
+}
+
+func sharedDBMaxTenantsFromEnv() (int, error) {
+	raw := strings.TrimSpace(os.Getenv("DRIVE9_TIDBCLOUD_NATIVE_SHARED_MAX_TENANTS"))
+	if raw == "" {
+		return 100, nil
+	}
+	v, err := strconv.Atoi(raw)
+	if err != nil || v <= 0 {
+		return 0, fmt.Errorf("invalid DRIVE9_TIDBCLOUD_NATIVE_SHARED_MAX_TENANTS=%q: must be a positive integer", raw)
+	}
+	return v, nil
+}
+
+func sharedDBDefaultSpendingLimitFromEnv() (int64, error) {
+	raw := strings.TrimSpace(os.Getenv("DRIVE9_TIDBCLOUD_NATIVE_DB_POOL_DEFAULT_SPENDING_LIMIT"))
+	if raw == "" {
+		return 10_000_000, nil
+	}
+	v, err := strconv.ParseInt(raw, 10, 64)
+	if err != nil || v <= 0 || v > int64(1<<31-1) {
+		return 0, fmt.Errorf("invalid DRIVE9_TIDBCLOUD_NATIVE_DB_POOL_DEFAULT_SPENDING_LIMIT=%q: must be in (0,%d]", raw, int64(1<<31-1))
 	}
 	return v, nil
 }
--- a/pkg/meta/db_pool.go
+++ b/pkg/meta/db_pool.go
@@ -5,6 +5,8 @@ import (
 	"database/sql"
 	"errors"
 	"fmt"
+	"math"
+	"strings"
 	"time"
 )
 
@@ -29,7 +31,51 @@ const (
 	SchemaShapeShared     = "shared"
 )
 
-const sharedDBStatusActive = "active"
+const (
+	SharedDBStatusProvisioning = "provisioning"
+	SharedDBStatusActive       = "active"
+	SharedDBStatusFailed       = "failed"
+	SharedDBStatusDraining     = "draining"
+)
+
+const sharedDBStatusActive = SharedDBStatusActive
+
+// SharedDBCapacityMode selects the capacity bound used by a reservation.
+type SharedDBCapacityMode string
+
+const (
+	SharedDBCapacityNormal    SharedDBCapacityMode = "normal"
+	SharedDBCapacityEmergency SharedDBCapacityMode = "emergency"
+)
+
+// SharedDBHardCap computes runtime emergency capacity from a persisted soft
+// capacity. The ratio is deliberately not stored in db_pool.
+func SharedDBHardCap(softCap int, ratio float64) (int, error) {
+	if softCap <= 0 {
+		return 0, fmt.Errorf("soft capacity must be positive")
+	}
+	if math.IsNaN(ratio) || math.IsInf(ratio, 0) || ratio < 1 {
+		return 0, fmt.Errorf("hard-cap ratio must be finite and at least 1")
+	}
+	value := float64(softCap) * ratio
+	if value > float64(int(^uint(0)>>1)) {
+		return 0, fmt.Errorf("hard capacity overflows int")
+	}
+	return int(math.Ceil(value)), nil
+}
+
+// SharedDBReopenThresholdForRatio computes the runtime hysteresis threshold.
+// The ratio is intentionally not persisted in db_pool; it is deployment
+// policy and must be in the open interval (0,1).
+func SharedDBReopenThresholdForRatio(softCap int, ratio float64) (int, error) {
+	if softCap <= 0 {
+		return 0, nil
+	}
+	if math.IsNaN(ratio) || math.IsInf(ratio, 0) || ratio <= 0 || ratio >= 1 {
+		return 0, fmt.Errorf("reopen ratio must be finite and in (0,1)")
+	}
+	return int(math.Floor(float64(softCap) * ratio)), nil
+}
 
 // SharedDB is one physical database registered in db_pool. PasswordCipher
 // holds the same encrypted envelope as tenants.db_password; the plaintext
@@ -38,18 +84,27 @@ const sharedDBStatusActive = "active"
 // for plaintext) so the runtime handle reopens the DB with exactly the TLS
 // mode it was registered with. MaxTenants of 0 means unlimited.
 type SharedDB struct {
-	DbID           int64
-	OrgID          string
-	Role           string
-	Host           string
-	Port           int
-	User           string
-	PasswordCipher []byte
-	Name           string
-	TLSMode        string
-	MaxTenants     int
-	TenantCount    int
-	Status         string
+	ID                      int64
+	TiDBCloudOrganizationID string
+	ClusterID               string
+	ProvisioningKey         []byte
+	CloudProvider           string
+	Region                  string
+	Role                    string
+	Host                    string
+	Port                    int
+	User                    string
+	PasswordCipher          []byte
+	Name                    string
+	TLSMode                 string
+	MaxTenants              int
+	TenantCount             int
+	SoftCapReached          bool
+	SpendingLimit           *int64
+	SchemaVersion           int
+	Status                  string
+	CreatedAt               time.Time
+	UpdatedAt               time.Time
 }
 
 // TenantPlacement maps one filesystem (fs_id) to the physical database that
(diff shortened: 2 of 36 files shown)


srstack:
## Review — TiDB Cloud Native shared DB control plane

Reviewed the full diff (36 files, +5823/−458) across three axes: meta/persistence, connection-pool/metrics/schema, and server/provisioner orchestration. Build passes; the shared-control-plane feature tests pass in isolation (the broader `pkg/server` failures are the known MySQL testcontainer flakiness, not regressions). The core design — atomic reserve, best-effort Cloud sync, dual-source membership, soft/hard hysteresis — is implemented soundly and matches the spec's intended scope.

### One correctness bug found and fixed in this PR

**SQL operator-precedence in the direct-create reservation** (`pkg/meta/db_pool.go`). The `membership == nil` path built `WHERE db_id = ? AND status = ? OR (status = ? AND cluster_id IS NOT NULL) AND (capacity guard)`. Since `AND` binds tighter than `OR`, this parsed as `(db_id=? AND status=active) OR (status=provisioning AND cluster_id IS NOT NULL AND capacity)` — the second arm has **no `db_id` filter**, so a single reservation would increment `tenant_count`/latch `soft_cap_reached` on **every provisioning pool table-wide with a cluster_id + headroom**, across all orgs; the first arm dropped the capacity guard entirely. Fixed by wrapping the substitution as `WHERE db_id = ? AND (%s) AND (capacity guard)`, and covered by a new regression test (`TestCompleteSharedTenantProvisionDoesNotUpdateOtherProvisioningPools`) that asserts an unrelated provisioning pool stays at `tenant_count=0`. Confirmed the fix + test are in the pushed diff. ✅

### Follow-ups (non-blocking, author already triaged)

- **File-size quota fallback** — the read-time `MaxFileSizeBytes <= 0 → default 10 GiB` coercion was restored (`pkg/meta/quota.go`) so existing rows with `max_file_size_bytes=0` don't fail open to unlimited. ✅
- **Per-request meta query for a metric label** — shared quota GET/SET now reuse the already-resolved DB pool instead of a second `GetSharedDBForTenant` call. ✅
- **Env validation** — `DRIVE9_TIDBCLOUD_NATIVE_SHARED_MAX_TENANTS` and `..._DB_POOL_DEFAULT_SPENDING_LIMIT` now reject `0`/negative at startup rather than relying on a downstream clamp. ✅
- **`soft_cap_reached` on unlimited pools** — delete path now preserves the existing latch instead of forcing it to 1 on `max_tenants=0` pools. ✅
- **Shared claim result** no longer populates `TenantDSN` with the shared system-user DSN (latent exposure removed). ✅

### Deliberately deferred (agree)

- **Metric org-label naming** (`tidbcloud_org_id` on new shared metrics vs pre-existing `organization_id` on `tenant_pool_*`) — renaming a pre-existing label mid-feature-PR would break existing dashboards; correct to defer to a standalone metrics-contract change.
- **Per-pod aggregate connection ceiling** — the two-layer model (per-handle 300 + per-process 10000) is the intended default; a per-cluster cap is out of scope here.
- **Open-time shared-schema version check** (`pkg/tenant/pool.go`) — this is an open-time idempotent ensure, not a periodic reconciler/worker; the design doc was updated to reflect that it's introduced by this rollout. Confirmed intent.

LGTM. The one real bug is fixed and regression-covered; everything else is either resolved or reasonably deferred.

pkg/tenant/pool.go
```diff
@@ -881,13 +903,88 @@ func (p *Pool) placementForTenant(ctx context.Context, fsID int64) (*meta.Tenant
...
+	}
+	p.sharedMu.Lock()
+	if refs := p.sharedDBRefs[dbID]; refs > 1 {
+		p.sharedDBRefs[dbID] = refs - 1
+	} else {
+		delete(p.sharedDBRefs, dbID)
+	}
+	p.sharedDBLastUsed[dbID] = time.Now()
+	p.sharedMu.Unlock()
+}
```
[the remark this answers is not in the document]
srstack: Fixed in bfeba0c6. releaseSharedDB now returns when the handle is no longer present, so invalidation cannot recreate ref or last-used bookkeeping; the regression test covers invalidate followed by release.

pkg/tenant/tidbcloudnative/provisioner.go
```diff
@@ -474,6 +489,172 @@ func (p *Provisioner) BatchProvisionFreeClustersWithCredentialsAndQuota(ctx cont
...
+	}
+	wantID := strconv.FormatInt(dbPoolID, 10)
+	var matches []clusterInfo
+	if strings.TrimSpace(clusterID) != "" {
+		info, err := p.getClusterBasicInfoWithCredentials(ctx, publicKey, privateKey, strings.TrimSpace(clusterID))
+		if err != nil {
+			return nil, err
+		}
+		matches = append(matches, *info)
+	} else {
```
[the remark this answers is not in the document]
srstack: Fixed in bfeba0c6. The known-cluster adoption path now requires drive9.ai/managed=true, drive9.ai/provider=tidb_cloud_native_shared, and the expected DB-pool ID label. Wrong-provider and unmanaged cases are covered.

cmd/drive9-server/main.go
```diff
@@ -218,14 +235,14 @@ func main() {
...
 		provisioner, provisionerErr = tidbcloudnative.NewProvisionerFromEnv()
 	case tenant.ProviderDB9:
 		provisioner, provisionerErr = db9.NewProvisionerFromEnv()
 	}
 	if provisionerErr != nil {
 		isWanted := false
-		if providerType == tenant.ProviderTiDBCloudNative && os.Getenv("DRIVE9_TIDBCLOUD_NATIVE_API_URL") != "" {
+		if tenant.UsesTiDBCloudNativeCredentials(providerType) && os.Getenv("DRIVE9_TIDBCLOUD_NATIVE_API_URL") != "" {
 			isWanted = true
 		}
```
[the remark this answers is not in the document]
srstack: Verified against main: the db9/tidb_zero warning behavior and isWanted logic are unchanged by this PR. Provisioners have intentionally remained optional at process startup (including the default tidb_zero selection), while a configured TiDB Cloud Native API URL fails fast. This PR only extends that existing Native condition to tidb_cloud_native_shared, so changing db9/tidb_zero startup semantics here would be unrelated behavior expansion.

pkg/server/admin_tenant_pool.go
```diff
@@ -911,6 +939,156 @@ func (s *Server) createFreePoolTenants(ctx context.Context, poolID string, count
...
+				return resolvedOrg, err
+			}
+			if row.SpendingLimit == nil {
+				return resolvedOrg, fmt.Errorf("managed db pool %d has no spending target", dbID)
+			}
+			rows[dbID] = row
+			requests = append(requests, tenant.SharedDBPoolCreateRequest{DBPoolID: dbID,
+				DatabaseName: row.Name, RootPassword: string(plain), SpendingLimitMonthly: *row.SpendingLimit})
+		}
+		created, createErr := provisioner.BatchProvisionSharedDBPoolsWithCredentials(ctx, requests, cred)
```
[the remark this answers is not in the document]
srstack: Fixed in 79953e39. Each physical batch-create chunk now acquires the same organization/credential allocation lock used by direct provisioning, re-reads rows under the lock, handles the credential-to-org identity transition, and skips rows whose cluster ID was already persisted. The concurrency regression test proves a direct ensure cannot issue a second create while the batch call is in flight.

pkg/server/admin_tenant_pool.go
```diff
@@ -911,6 +939,156 @@ func (s *Server) createFreePoolTenants(ctx context.Context, poolID string, count
...
+			OrganizationID: selected.TiDBCloudOrganizationID})
+	}
+	createdIDs := make([]int64, 0, len(createdPoolIDs))
+	for id := range createdPoolIDs {
+		createdIDs = append(createdIDs, id)
+	}
+	sort.Slice(createdIDs, func(i, j int) bool { return createdIDs[i] < createdIDs[j] })
+	resolvedOrg, err := s.provisionManagedSharedDBPoolsBatch(ctx, createdIDs, cred)
+	if err != nil {
+		return results, err
```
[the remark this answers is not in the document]
srstack: Fixed in 79953e39. The create handler now cleans every partial result before deleting logical pool metadata. Shared members use shared-specific cleanup, which removes membership and placement, decrements DB-pool capacity, revokes keys, and finalizes the empty tenant; the regression test asserts no org-scoped membership, placement, or capacity remains.

pkg/server/admin_tenant_pool.go
```diff
@@ -1338,104 +1516,102 @@ func (s *Server) deleteNewestFreePoolTenants(ctx context.Context, poolID, organi
...
+	if err != nil {
+		return err
+	}
+	placement, err := s.meta.GetTenantPlacement(ctx, fsID)
+	if err != nil && !errors.Is(err, meta.ErrNotFound) {
+		return err
+	}
+	if placement != nil {
+		if err := s.pool.PurgeSharedTenant(ctx, fsID, placement.DbID); err != nil {
+			return err
```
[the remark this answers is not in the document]
srstack: Fixed in 79953e39. Free-member deletion now checks persisted DB-pool readiness and purges only active pools with complete connection metadata. Pending/provisioning/failed unready members release placement and capacity directly; the regression test covers a provisioning pool with only a cluster ID.

pkg/server/shared_db_pool.go
```diff
@@ -0,0 +1,554 @@
...
+	if err != nil {
+		return fmt.Errorf("decrypt shared db root password: %w", err)
+	}
+	needsMetadata := poolInfo.ClusterID == "" || poolInfo.Host == "" || poolInfo.Port <= 0 || poolInfo.User == ""
+	if needsMetadata {
+		if poolInfo.SpendingLimit == nil {
+			return fmt.Errorf("managed db pool %d has no spending target", dbID)
+		}
+		result, loadErr := provisioner.LoadSharedDBPoolWithCredentials(ctx, dbID, poolInfo.ClusterID, cred)
+		if loadErr != nil {
```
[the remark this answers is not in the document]
srstack: Fixed in 79953e39. An incomplete continuation now releases the allocation lock, invokes a SharedDBPoolMetadataWaiter that reuses the existing TiDB Cloud Native readiness poll, then re-enters the locked continuation to persist metadata and activate. No new worker, retry config, or state was introduced.

pkg/server/server.go
```diff
@@ -5196,6 +5259,53 @@ func (s *Server) provisionTenant(ctx context.Context, opts provisionTenantOption
...
+			return nil, newProvisionTenantError(http.StatusBadRequest, err.Error(), err)
+		}
+		var res *provisionTenantResult
+		virtualLimit := s.sharedTenantVirtualSpendingLimit(opts)
+		sharedDB, created, err := s.allocateManagedSharedDB(ctx, *opts.CredentialProvisioner, virtualLimit, nil)
+		if err != nil {
+			_ = s.meta.UpdateTenantStatus(context.Background(), tenantID, meta.TenantFailed)
+			return nil, newProvisionTenantError(http.StatusBadGateway, "failed to allocate shared database", err)
+		}
+		if created || sharedDB.ClusterID == "" || sharedDB.Host == "" || sharedDB.Port <= 0 || sharedDB.User == "" {
```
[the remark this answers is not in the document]
srstack: Fixed in 79953e39. The explicit shared-provider path now enters managed physical creation only for rows with a persisted spending target. Active RegisterSharedDB rows have a NULL spending target and are placed directly; the regression test also verifies their per-tenant virtual spending limit is materialized.

pkg/server/server.go
```diff
@@ -5085,18 +5136,20 @@ func (s *Server) findSharedDBForProvision(ctx context.Context, provider string,
...
 // without a placement row, no capacity leak, no orphaned key.
 func (s *Server) provisionTenantOnSharedDB(ctx context.Context, tenantID string, sharedDB *meta.SharedDB, provider, keyName string, opts provisionTenantOptions, now time.Time) (*provisionTenantResult, error) {
+	return s.provisionTenantOnSharedDBMode(ctx, tenantID, sharedDB, provider, keyName, opts, now, meta.SharedDBCapacityNormal, 0)
+}
+
+func (s *Server) provisionTenantOnSharedDBEmergency(ctx context.Context, tenantID string, sharedDB *meta.SharedDB, provider, keyName string, opts provisionTenantOptions, now time.Time, hardCap int) (*provisionTenantResult, error) {
+	return s.provisionTenantOnSharedDBMode(ctx, tenantID, sharedDB, provider, keyName, opts, now, meta.SharedDBCapacityEmergency, hardCap)
+}
+
+func (s *Server) provisionTenantOnSharedDBMode(ctx context.Context, tenantID string, sharedDB *meta.SharedDB, provider, keyName string, opts provisionTenantOptions, now time.Time, capacityMode meta.SharedDBCapacityMode, hardCap int) (*provisionTenantResult, error) {
```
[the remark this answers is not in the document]
srstack: Fixed in 79953e39. Ordinary native/zero routing to a manually registered shared DB pool again fails closed when a create-time TiDB Cloud spending limit is present. Only tidb_cloud_native_shared accepts and persists that value as a virtual per-tenant limit.

srstack:
Addressed the two review-body items that were not attached to inline threads in 79953e39:\n\n- Shared DB cold opens no longer hold the global shared-handle mutex across metadata reads, connection opening, or schema ensure. Opens are serialized per db_pool_id, so unrelated DB pools proceed independently while duplicate opens for one DB pool are deduplicated.\n- SharedDBPoolProvisioner now documents the partial-result contract. The TiDB Cloud Native implementation explicitly preserves decoded successes with a non-nil partial error, and both batch/continuation callers process returned pools before handling the error.\n\nValidation on this head:\n- affected shared/tenant-pool regression suite passed\n- pkg/tenant, pkg/tenant/tidbcloudnative, pkg/meta, and cmd/drive9-server passed\n- go build ./... passed\n- make lint passed with 0 issues\n\nThe full pkg/server run only hit the known TestSSEOutboxCrossPodPollerDelivery database-connection flake; that test passed when rerun in isolation.

pkg/server/tenant_delete_shared_test.go
```diff
@@ -319,24 +328,34 @@ func TestSharedTenantDeletePlacementRemovalFailureStaysRetryable(t *testing.T) {
...
 		}
 	}
+	dropSharedTables()
+	t.Cleanup(func() {
+		if err := rt.pool.InvalidateSharedDB(dbID); err != nil {
+			t.Errorf("invalidate shared database: %v", err)
+		}
+		dropSharedTables()
+		initServerTenantSchema(t, testDSN)
+	})
```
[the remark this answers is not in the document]
srstack: Fixed in a168dd99. The drop helper now accepts *testing.T, calls t.Helper(), and returns the first table-drop error. Cleanup is registered before destructive setup; setup failures use Fatalf, while cleanup failures remain Errorf. The shared-delete pollution regression sequence passes.

srstack:
## Re-review — updated through `a168dd99`

Re-reviewed the 3 commits pushed since my first pass (`bfeba0c6` review findings, `79953e39` concurrency, `a168dd99` test fail-fast) — +857/−59 across 11 files, concentrated in the provisioning/pool concurrency paths. All build, and I ran the 8 new regression tests plus the meta/tenant/mysqlutil suites: green. This is a strong round of hardening. Details:

### Concurrency fixes — verified correct + covered

1. **Shared-DB handle cache race** (`pkg/tenant/pool.go`). `sharedDBHandle` now uses a per-`dbID` open mutex with double-checked locking: cache-check under `sharedMu` → acquire per-key `openMu` → re-check → open *outside* both locks → re-check `sharedDBClosed` before store (closing the just-opened handle if `Close()` raced). This is exactly the concurrent-open / open-after-close leak I flagged as latent. `InvalidateSharedDB` shares the **same** per-key mutex, so invalidate/open are mutually exclusive per pool, and the new `releaseSharedDB` guard (`if _, ok := p.sharedDBs[dbID]; !ok { return }`) prevents ref underflow after invalidation. Covered by `TestPoolInvalidateSharedDBClosesOnlySelectedHandle`, `TestPoolAcquireInvalidateDefersCloseUntilRelease`. ✅

2. **Batch double-cluster-create** (`pkg/server/admin_tenant_pool.go`). Batch chunk creation now runs under `WithSharedDBAllocationLock`, and `provisionManagedSharedDBPoolsBatchLocked` skips any row whose `ClusterID != ""` — so a concurrent *direct* continuation that already created the physical cluster is adopted, never recreated. The stale-lock-identity case (org resolved concurrently) is handled by the `restartWithOrganization` retry. Covered by `TestManagedSharedDBBatchCreateUsesAllocationLock`. ✅

3. **Partial-success contract** (`pkg/tenant/provisioner.go` + caller). The interface now documents that a non-nil error may accompany a decoded subset, and the caller persists **every** returned `created` result via `UpdateManagedSharedDBPoolCloudResult` before logging/handling `createErr`. Failed pools stay `provisioning` for the leader resume loop to retry — consistent with the best-effort model. ✅

4. **Metadata-not-ready retry** (`pkg/server/shared_db_pool.go`). Sentinel `errSharedDBConnectionMetadataNotReady` + a bounded 2-attempt loop that polls cloud endpoint metadata via `SharedDBPoolMetadataWaiter` when the cluster exists but host/port aren't populated yet. Covered by `TestManagedSharedDBContinuationWaitsForConnectionMetadata` / `...WaitForSharedDBPoolMetadataUsesNativeReadinessPoll`. ✅

### Correctness / fail-closed fixes

5. **Spending limit on non-managed shared pools** (`pkg/server/server.go`). Now fails closed (mark `TenantFailed` + 400) instead of silently dropping the requested limit, and the physical-ensure path is gated on `sharedDB.SpendingLimit != nil` so manually-registered external pools (NULL limit, operator-populated connection info) aren't pushed through managed ensure. Covered by `TestProvisionTiDBCloudNativeRejectsSpendingLimitOnRegisteredSharedDBPool` / `...UsesRegisteredManualDBPool`. ✅

6. **Shared-tenant cleanup path** (`pkg/server/admin_tenant_pool.go`). `cleanupCreatedPoolTenants` now routes `ProviderTiDBCloudNativeShared` members through `deleteFreeSharedPoolTenant` (fs/placement scope) instead of `deprovisionTenantCluster` — shared tenants must never delete the shared cluster. And `deleteFreeSharedPoolTenant` gates `PurgeSharedTenant` on `connectionReady`, so cleanup of a tenant on a not-yet-ready pool still removes placement + decrements count rather than hard-failing. Covered by `TestAdminTenantPoolCreateCleansPartialSharedMembersOnBatchFailure` / `...SkipsPurgeWhenDBPoolIsUnready`. ✅

### One tiny note (non-blocking)

`sharedDBOpenMu` entries are never deleted, so the map grows with distinct physical pool IDs seen by a pod. That set is small and bounded (number of shared pools), so this is negligible — mentioning only for completeness; no action needed.

CodeRabbit's two findings (main.go fail-fast, test setup fail-fast) are both withdrawn/addressed. Nothing outstanding. **LGTM** — the concurrency model is now materially tighter than my first-pass review, and every fix carries a targeted regression test.

pkg/server/server.go
```diff
@@ -5196,6 +5268,53 @@ func (s *Server) provisionTenant(ctx context.Context, opts provisionTenantOption
...
+			_ = s.meta.UpdateTenantStatus(context.Background(), tenantID, meta.TenantFailed)
+			return nil, newProvisionTenantError(http.StatusBadRequest, tenant.ErrCredentialsRequired.Error(), tenant.ErrCredentialsRequired)
+		}
+		if err := s.materializeSharedTenantQuota(ctx, tenantID, opts); err != nil {
+			_ = s.meta.UpdateTenantStatus(context.Background(), tenantID, meta.TenantFailed)
+			return nil, newProvisionTenantError(http.StatusBadRequest, err.Error(), err)
+		}
+		var res *provisionTenantResult
+		virtualLimit := s.sharedTenantVirtualSpendingLimit(opts)
+		sharedDB, created, err := s.allocateManagedSharedDB(ctx, *opts.CredentialProvisioner, virtualLimit, nil)
```
mornyx: **Blocking — direct allocate/place race**

Direct shared provision allocates with `reserve == nil`, so selection/create and capacity reservation are not one critical section:

```go
sharedDB, created, err := s.allocateManagedSharedDB(ctx, ..., virtualLimit, nil)
// ... ensure physical ...
res, err = s.provisionTenantOnSharedDB(...)
```

Prewarm already passes `CompleteSharedTenantPoolMember` as `reserve` and retries once on capacity/spending exhaustion. Concurrent direct creates can all pick the last soft-cap slot; losers get HTTP 409 **and** are marked `failed` by the outer `UpdateTenantStatus` below.

Please either:
1. Reserve under `WithSharedDBAllocationLock` like prewarm (multi-attempt reallocate/create), or
2. At least stop treating capacity/spending conflicts as terminal failures.

Otherwise soft-cap races produce dead tenant rows + quota + fs_id residue instead of clean retries.

pkg/server/server.go
```diff
@@ -5196,6 +5268,53 @@ func (s *Server) provisionTenant(ctx context.Context, opts provisionTenantOption
...
+						}
+					}
+				}
+				_ = s.meta.UpdateTenantStatus(context.Background(), tenantID, meta.TenantFailed)
+				return nil, newProvisionTenantError(http.StatusBadGateway, "failed to create shared database", err)
+			}
+		}
+		res, err = s.provisionTenantOnSharedDB(ctx, tenantID, sharedDB, provider, keyName, opts, now)
+		if err != nil {
+			_ = s.meta.UpdateTenantStatus(context.Background(), tenantID, meta.TenantFailed)
```
mornyx: **Blocking — 409 capacity conflicts become permanent failures**

`provisionTenantOnSharedDBMode` already maps `ErrSharedDBCapacityExhausted` / `ErrSharedDBSpendingLimitExceeded` to HTTP 409 without calling `fail()`. This outer block then forces `TenantFailed`:

```go
res, err = s.provisionTenantOnSharedDB(...)
if err != nil {
    _ = s.meta.UpdateTenantStatus(..., meta.TenantFailed)
    return nil, err
}
```

That turns a retryable capacity race into permanent residue. Please special-case those errors (leave pending, or roll back the pending tenant if product wants zero residue).

pkg/server/server.go
```diff
@@ -5196,6 +5268,53 @@ func (s *Server) provisionTenant(ctx context.Context, opts provisionTenantOption
...
+			return nil, newProvisionTenantError(http.StatusBadGateway, "failed to allocate shared database", err)
+		}
+		if sharedDB.SpendingLimit != nil && (created || sharedDB.ClusterID == "" || sharedDB.Host == "" || sharedDB.Port <= 0 || sharedDB.User == "") {
+			plannedDB := sharedDB
+			sharedDB, err = s.ensureManagedSharedDBPhysical(ctx, plannedDB.ID, *opts.CredentialProvisioner)
+			if err != nil {
+				orgID, orgErr := s.firstManagedOrganization(ctx, *opts.CredentialProvisioner)
+				if orgErr == nil {
+					hardCap, hardErr := s.managedSharedDBHardCap(plannedDB.MaxTenants)
```
mornyx: **Blocking (if max_tenants can change on a live fleet) — emergency hard cap uses planned pool, not candidate**

```go
hardCap, hardErr := s.managedSharedDBHardCap(plannedDB.MaxTenants)
fallback, findErr := s.meta.FindSharedDBForEmergency(ctx, orgID, hardCap, virtualLimit)
```

`FindSharedDBForEmergency` applies that hardCap as a global `tenant_count < hardCap` filter. Emergency reserve only rejects `hardCap < maxTenants`, not overshoot of the **candidate’s** own hard cap.

If older managed pools still have a smaller soft cap (config raised `DRIVE9_..._MAX_TENANTS`), a soft=50 pool can be emergency-filled up to hardCap computed from a new soft=100 pool (→ 120).

Please compute hard cap from **candidate.MaxTenants** (SQL or after fetch) and pass that into `CompleteSharedTenantProvisionEmergency`. If soft cap is considered immutable per fleet, document that explicitly.

pkg/server/shared_db_pool.go
```diff
@@ -0,0 +1,580 @@
...
+		}
+		if err := s.meta.UpdateManagedSharedDBPoolCloudResult(ctx, &meta.SharedDB{
+			ID: dbID, TiDBCloudOrganizationID: poolInfo.TiDBCloudOrganizationID, ClusterID: poolInfo.ClusterID,
+			Host: poolInfo.Host, Port: poolInfo.Port, User: username, PasswordCipher: passwordCipher,
+			Name: poolInfo.Name, TLSMode: poolInfo.TLSMode,
+		}); err != nil {
+			return err
+		}
+	}
+	if err := s.pool.EnsureSharedDBReady(ctx, dbID); err != nil {
```
mornyx, line 576: **High — org allocation lock held across long physical work**

Metadata wait is correctly outside the lock, but `continueManagedSharedDBPoolLocked` still holds the org named lock across:
- Cloud load/create
- `EnsureDatabase`
- `EnsureSystemUser`
- `EnsureSharedDBReady` (shared schema DDL)

That serializes all allocate/place work for the org and pins a dedicated meta `*sql.Conn` for a long time. Sequential continuations help multi-pool cases, but lock scope is still very wide.

Suggest splitting stages:
1. Under lock: durable meta transitions only (adopt cluster_id, status CAS).
2. Outside lock: wait / ensure DB / system user / schema.
3. Short locked: `ActivateSharedDBPool` + tenant activation batches.

If launch is intentionally single-threaded/low QPS, please document that assumption.

pkg/server/admin_tenant_pool.go
```diff
@@ -911,6 +940,212 @@ func (s *Server) createFreePoolTenants(ctx context.Context, poolID string, count
...
+		}
+		rows[dbID] = row
+		requests = append(requests, tenant.SharedDBPoolCreateRequest{DBPoolID: dbID,
+			DatabaseName: row.Name, RootPassword: string(plain), SpendingLimitMonthly: *row.SpendingLimit})
+	}
+	if len(requests) == 0 {
+		return resolvedOrg, nil
+	}
+	created, createErr := provisioner.BatchProvisionSharedDBPoolsWithCredentials(ctx, requests, cred)
+	if createErr != nil && len(created) == 0 {
```
mornyx, line 1116: **Medium — total batch create failure returned as success**

```go
if createErr != nil && len(created) == 0 {
    logger.Warn(ctx, "managed_shared_db_batch_create_ambiguous", ...)
    return resolvedOrg, nil // error discarded
}
```

Reasonable for transient Cloud blips if continuations are guaranteed, but permanent errors (bad credentials, 403) leave free tenants stuck on `provisioning` pools while the admin API looks green.

Please classify retryable vs permanent errors, and rename the log event — `ambiguous` is misleading for generic failures.

pkg/server/shared_db_pool.go
```diff
@@ -0,0 +1,580 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("decrypt shared db root password: %w", err)
+	}
+	if poolInfo.SpendingLimit == nil {
+		return nil, fmt.Errorf("managed db pool %d has no spending target", poolInfo.ID)
+	}
+	result, err := provisioner.LoadSharedDBPoolWithCredentials(ctx, poolInfo.ID, poolInfo.ClusterID, cred)
+	if err != nil {
+		if errors.Is(err, tenant.ErrSharedDBPoolAmbiguous) {
+			if markErr := s.meta.MarkSharedDBPoolFailed(ctx, poolInfo.ID); markErr != nil {
```
mornyx, line 382: **Medium — `MarkSharedDBPoolFailed` can strand placed tenants**

Fail-closed on `ErrSharedDBPoolAmbiguous` is good for Cloud identity, but if any tenants were already placed (prewarm / race after `cluster_id` persistence), they never activate (`ActivateSharedDBPool` requires `active`) and capacity is not released.

Options:
- Refuse `failed` while `tenant_count > 0`, or
- Rehome/fail those tenants and release capacity, plus a metric/alert for `shared_db_pool_failed_with_tenants`.

pkg/tenant/tidbcloudnative/provisioner.go
```diff
@@ -474,6 +489,208 @@ func (p *Provisioner) BatchProvisionFreeClustersWithCredentialsAndQuota(ctx cont
...
+		infos, err := p.listClusterInfosWithCredentials(ctx, publicKey, privateKey, nil, 100)
```
mornyx, line 622: **Medium — load-by-label list is capped at 100 clusters**

```go
infos, err := p.listClusterInfosWithCredentials(ctx, publicKey, privateKey, nil, 100)
```

When `clusterID` is empty, orgs with >100 managed clusters can miss the labeled shared pool and attempt a second create (then ambiguous → failed).

Once `cluster_id` is durable this is mostly fine; the dangerous window is early resume before `cluster_id` is written, or if that write is lost. Please paginate the list, or require `cluster_id` after first successful create and avoid unbounded label scans.

pkg/meta/db_pool.go
```diff
@@ -349,28 +863,57 @@ func (s *Store) DeleteTenantPlacement(ctx context.Context, fsID int64) (err erro
...
+}
+
+// CompleteSharedTenantPoolMember reserves a shared placement and records a
+// free logical-pool membership in the same transaction. Prewarmed members do
+// not receive an owner key until claim.
+func (s *Store) CompleteSharedTenantPoolMember(ctx context.Context, tenantID, provider string, p *TenantPlacement, membership *TenantPoolMembership) error {
+	return s.completeSharedTenantProvision(ctx, tenantID, provider, p, nil, membership, SharedDBCapacityNormal, 0)
+}
+
+func (s *Store) completeSharedTenantProvision(ctx context.Context, tenantID, provider string, p *TenantPlacement, k *APIKey, membership *TenantPoolMembership, capacityMode SharedDBCapacityMode, hardCap int) (err error) {
```
mornyx, line 922: **Positive**

Soft-cap latch + reopen hysteresis is cleanly separated from persisted policy (ratio is runtime-only), and the atomic reserve / placement / key-or-membership path matches the delete atomicity story in `DeleteTenantPlacementAndDecrCount`. The hysteresis unit tests make the intended state machine easy to review — thank you.

pkg/meta/db_pool.go
```diff
@@ -384,22 +927,113 @@ func (s *Store) CompleteSharedTenantProvision(ctx context.Context, tenantID, pro
...
+			var currentSum int64
+			if err := tx.QueryRowContext(ctx, `SELECT COALESCE(SUM(q.tidbcloud_spending_limit), 0)
+				FROM tenant_placements p
+				JOIN fs_registry f ON f.fs_id = p.fs_id
+				JOIN tenant_quota_config q ON q.tenant_id = f.tenant_id
+				WHERE p.db_id = ?`, p.DbID).Scan(&currentSum); err != nil {
+				return fmt.Errorf("sum tenant spending limits for db pool %d: %w", p.DbID, err)
+			}
+			prospective := currentSum + tenantSpendingLimit.Int64
+			if prospective < currentSum || prospective > int64(1<<31-1) || prospective >= poolSpendingLimit.Int64 {
```
mornyx, line 1025: **Nit / product question — spending headroom is strict `<`**

`prospective >= poolSpendingLimit` (and the SQL finder uses `sum + new < spending_limit`) means the last unit of pool spending is intentionally unusable.

If that is permanent headroom, please document it next to the managed pool policy defaults. If the full target should be fillable, switch to `<=` consistently in finder + reserve + `UpdateSharedTenantQuotaConfig`.

pkg/server/tenant_delete.go
```diff
@@ -236,12 +236,17 @@ func (s *Server) handleSharedTenantDelete(w http.ResponseWriter, r *http.Request
...
 		// completes the removal — a placement row must never outlive its
 		// tenant or lose its capacity accounting.
-		if err := s.meta.DeleteTenantPlacementAndDecrCount(ctx, fsID, placement.DbID); err != nil {
+		if err := s.meta.DeleteTenantPlacementAndDecrCount(ctx, fsID, placement.DbID, s.sharedDBReopenRatio); err != nil {
 			logger.Error(ctx, "shared_tenant_placement_delete_failed", zap.String("tenant_id", t.ID), zap.Error(err))
 			errJSON(w, http.StatusInternalServerError, "failed to remove tenant placement")
 			return
 		}
 	}
+	if err := s.meta.DeleteTenantPoolMembership(ctx, t.ID); err != nil {
```
mornyx, line 248: **Positive + ordering note**

Glad shared delete removes `tenant_pool_memberships` after placement release and remains retry-safe when placement is already gone.

One product note: if membership delete fails after successful placement release, the tenant stays `deleting` with capacity already reopened — that looks correct for capacity accounting. A one-line comment here would help prevent future reordering of those steps.

mornyx (changes requested):
## Review summary

Solid control-plane design overall: durable managed `db_pool` rows, soft/hard capacity with reopen hysteresis, virtual spending headroom, lazy shared-schema init, and provider-dispatched lifecycle. Meta-layer capacity transactions and shared delete (purge by `fs_id` only) look intentional and well tested.

**Requesting changes** for a few correctness / operability issues on the hot path before merge:

1. **Direct provision TOCTOU** — allocate with `reserve == nil`, then place outside the org lock (prewarm already does this correctly).
2. **409 capacity/spending races mark tenants `failed`** — turns retryable conflicts into permanent residue.
3. **Org allocation lock held across Cloud create + schema ensure** — stalls all allocate/place for the org.
4. **Emergency hard cap computed from planned pool, applied to candidates** — can overfill older pools after config changes.

Non-blocking but important: batch create error swallow, failed pool with placed tenants, label load list limit 100, document spending strict `<` headroom, and call out the fleet-wide `quota_limits_overridden` removal.

Happy to re-review once the direct-path reserve + failed-on-409 + emergency hard-cap items are addressed (or explicitly accepted with launch constraints).

pkg/meta/meta.go
```diff
@@ -389,6 +390,12 @@ func (s *Store) migrate() (err error) {
 	if err := applyMetaSchemaRepairs(ctx, s.db, plannedMetaSchemaRepairs(diffs)); err != nil {
 		return err
 	}
+	if err := expandManagedDBPoolSchema(ctx, s.db); err != nil {
+		return err
+	}
+	if err := removeQuotaLimitsOverriddenColumn(ctx, s.db); err != nil {
```
mornyx: **Blocking — dropping `quota_limits_overridden` breaks rolling upgrades**

The current (old) binary still references this column in hot paths: `atomicReserveAndInsertUploadOnce` / `uploadReservationQuotaErrorTx` (i.e. every upload reservation) and all quota reads/writes (`GetQuotaConfig`, `SetQuotaConfig*`, `SetQuotaStorageBytes`, `SetTiDBCloudSpendingLimitIfNotUpdatedAfter`). The moment the first new pod starts and this migration drops the column, every still-running old pod starts failing uploads and quota operations with `Unknown column 'quota_limits_overridden'`.

The good news: this PR already makes the column logically unused, and `diffMetaSchema` only flags *missing* spec columns, so a leftover physical column is tolerated by the re-diff. Please do the standard two-phase instead — keep the column in this release (just stop reading/writing it, which the code already does) and drop it in a follow-up release once the fleet no longer runs the old binary.

Also worth one line in the PR description: rows that had `quota_limits_overridden = 0` now enforce their stored values instead of live process defaults. Stored values were written as defaults-at-insert-time, so this is near-equivalent today, but it pins stale defaults if the constants ever change.

pkg/server/shared_db_pool.go
```diff
@@ -0,0 +1,580 @@
...
+			}
+			if err := s.continueManagedSharedDBPool(workerCtx, dbID, cred); err != nil {
+				logger.Warn(workerCtx, "managed_shared_db_pool_continue_failed",
+					zap.Int64("db_pool_id", dbID), zap.Error(err))
+			}
+		}
+	})
+}
+
+func (s *Server) resumeManagedSharedDBPoolsWithCtx(ctx context.Context) {
```
mornyx, line 266: **High — no periodic retry for failed pool continuations**

Continuation triggers are exactly three: this one-shot resume at leadership start, a direct provision happening to select the same provisioning pool (server.go:5313-5314), and prewarm batches. A transient Cloud error or a metadata-wait timeout in `continueManagedSharedDBPool` therefore leaves the pool `provisioning` — and every tenant already placed on it stuck in `provisioning` — until a leadership change or a lucky new allocation. A soft-cap-latched pool, or an org with no new traffic, never recovers.

Two aggravating details:
- This resume silently no-ops when default credentials are not configured (`resolveDefaultCredentials`), so in deployments where each org brings its own API key a stranded pool can only be revived by requests carrying that org's credentials.
- The emergency fallback path (server.go:5289-5305) returns early on success and never schedules a continuation for the pool whose physical creation just failed.

Please add a periodic reconciler tick (like the pending-tenant reconciler) or a bounded backoff retry inside the continuation worker.

pkg/tenant/tidbcloudnative/provisioner.go
```diff
@@ -474,6 +489,208 @@ func (p *Provisioner) BatchProvisionFreeClustersWithCredentialsAndQuota(ctx cont
...
+		id := strconv.FormatInt(input.DBPoolID, 10)
+		requests = append(requests, map[string]any{"cluster": map[string]any{
+			"displayName":  clusterDisplayName(id),
+			"rootPassword": password,
+			"region":       map[string]string{"name": p.regionName()},
+			"labels": map[string]string{
+				Drive9ManagedLabel:    "true",
+				Drive9ProviderLabel:   tenant.ProviderTiDBCloudNativeShared,
+				Drive9DBPoolIDLabel:   id,
+				Drive9PoolStatusLabel: SharedDBPoolStatusProvisioning,
```
mornyx: **Medium — cloud-side `drive9.ai/status` label is stuck at `provisioning` forever**

Batch create stamps `drive9.ai/status=provisioning` on the shared cluster, but nothing ever patches it: native pool clusters get free/used transitions via `MarkClusterPoolUsed/Free` (provisioner.go:1001-1040), shared pools have no equivalent. Any ops tooling, runbooks, or dashboards reading this label will see every shared cluster as permanently provisioning — and the linked runbooks PR (tidbcloud/runbooks#2685) presumably consumes exactly these labels.

Please either patch the label when the continuation activates the pool (`continueManagedSharedDBPoolLocked` already knows when), or stop setting it until a lifecycle exists.

pkg/server/admin_tenant_pool.go
```diff
@@ -1784,6 +2065,43 @@ func (s *Server) claimAdminTenantFromPool(ctx context.Context, cred tenant.Crede
...
+		}
+	}
+	rawToken, apiKeyID, key, err := s.buildOwnerAPIKey(ctx, row.Tenant.ID, "default", 1, apiKeyIssueSource{})
+	if err != nil {
+		return nil, err
+	}
+	if err := s.meta.ClaimSharedTenantPoolMembership(ctx, row.Tenant.ID, pool.PoolID, patch, key); err != nil {
+		return nil, err
+	}
+	fsID, err := s.meta.ResolveFsID(ctx, row.Tenant.ID)
```
mornyx: **Medium — shared claim has no compensation for post-commit failures**

`ClaimSharedTenantPoolMembership` commits the quota patch + membership CAS + owner key in one transaction, but the response is then built from three meta reads *after* commit (`ResolveFsID`, `GetTenantPlacement`, `GetSharedDB`). If any of them fails transiently, the caller gets a 5xx, retries, and claims a different member — while this one stays `used` with an owner API key whose token was never delivered: a stranded active tenant consuming pool capacity, with no recovery path. The native claim path has `releaseClaimedPoolTenant` compensation; the shared path has none.

These reads are stable for a free member (fs_id and placement don't change while free), so the simple fix is to perform them before the claim transaction — or add a shared release/compensation path.

pkg/tenant/pool.go
```diff
@@ -149,11 +149,15 @@ type Pool struct {
...
+	// db_pool.id). All tenants placed on the same shared DB share its
 	// handle; each Acquire still gets its own Store (carrying that tenant's
 	// fs_id scope) over the shared handle.
-	sharedMu  sync.Mutex
-	sharedDBs map[int64]*sql.DB
+	sharedMu         sync.Mutex
+	sharedDBs        map[int64]*sql.DB
+	sharedDBRefs     map[int64]int
+	sharedDBLastUsed map[int64]time.Time
+	sharedDBOpenMu   map[int64]*sync.Mutex
```
mornyx, line 159: **Nit — `sharedDBOpenMu` entries are never removed**

One `*sync.Mutex` per db_pool id accumulates for the process lifetime; neither `reapIdleSharedDBs` nor `InvalidateSharedDB` deletes the entry. Tiny per entry, but unbounded under pool churn — consider deleting it once the handle is closed and no opener holds it, or using a striped lock.

Also noting that `InvalidateSharedDB` currently has test-only callers; if it is intentionally dormant production API (e.g. reserved for credential rotation), a one-line comment saying so would help.

mornyx (changes requested):
Re-reviewed at a168dd99. There have been no new commits since the previous CHANGES_REQUESTED review, so I re-verified all seven outstanding items against the current tree — every one is still present:

- Direct allocate/place race (`reserve == nil`) at server.go:5282
- 409 capacity conflicts forced to permanent `failed` at server.go:5310 (the emergency fallback at server.go:5297-5305 has the same shape: a 409 from `CompleteSharedTenantProvisionEmergency` falls through to `TenantFailed` + 502)
- Emergency hard cap computed from the planned pool's `MaxTenants`, not the fallback candidate's (server.go:5293-5295)
- Org allocation lock held across Cloud create / EnsureDatabase / EnsureSystemUser / schema DDL (shared_db_pool.go:437-579)
- Total batch create failure returned as success (admin_tenant_pool.go:1116-1120)
- `MarkSharedDBPoolFailed` can strand placed tenants (shared_db_pool.go:366-370, 475-479)
- Label-scan capped at 100 clusters (provisioner.go:625)

This pass adds four new findings (one blocking) and one nit, posted inline:

1. **Blocking** — dropping `quota_limits_overridden` breaks rolling upgrades (meta.go:396)
2. **High** — no periodic retry for failed shared-pool continuations (shared_db_pool.go:251)
3. **Medium** — cloud-side `drive.ai/status` label stuck at `provisioning` forever (provisioner.go:536)
4. **Medium** — shared claim has no compensation for post-commit failures (admin_tenant_pool.go:2087)
5. Nit — `sharedDBOpenMu` never cleaned (pool.go:159)

On the positive side, the core meta-layer design is solid: the atomic `completeSharedTenantProvision` / `ClaimSharedTenantPoolMembership` / `DeleteTenantPlacementAndDecrCount` transaction boundaries, spending-headroom re-validation at both claim time and quota-update time, soft-cap latch + reopen hysteresis kept as runtime-only policy, and least-loaded emergency selection all match the intended semantics. The hysteresis and claim-race test coverage makes the state machine easy to verify.

Requesting changes: please address the outstanding items from the previous review plus the new DROP COLUMN phasing issue before merge.

pkg/server/server.go
srstack: Fixed in 7dc13b49. Direct shared creation now performs its final reserve under the organization allocation lock, and capacity/spending CAS conflicts retry allocation instead of failing the tenant. The concurrent single-slot regression covers the race.

pkg/server/server.go
srstack: Fixed in 7dc13b49. Capacity and spending conflicts are now treated as retryable allocation conflicts. The shared create path retries selection/reservation and no longer turns these conflicts into TenantFailed.

pkg/server/server.go
srstack: Fixed in 7dc13b49. Emergency selection now evaluates tenant_count against CEIL(candidate.max_tenants * hard_cap_ratio) for each candidate, and the reserve path recomputes the hard cap from the selected candidate.

pkg/server/shared_db_pool.go
srstack: The wide organization allocation lock is intentional for this control-plane path and is now documented in 7dc13b49. Existing tenant-pool operations already hold their named lock across Cloud mutation; this path is sequential and low-QPS, preserves single-owner mutation, and keeps a half-ready pool from being allocated concurrently. The metadata readiness waiter remains outside the lock.

pkg/server/admin_tenant_pool.go
srstack: Fixed in 7dc13b49. A total batch-create failure now returns the original Cloud error instead of reporting success, so the existing caller cleanup path runs. Partial successes are still persisted before the error is returned.

pkg/server/shared_db_pool.go
srstack: Fixed in 7dc13b49. MarkSharedDBPoolFailed now changes state only when tenant_count = 0; a provisioning pool that already owns placements remains recoverable and is not stranded in failed.

pkg/tenant/tidbcloudnative/provisioner.go
srstack: No code change is needed here. listClusterInfosWithCredentials already paginates internally until nextPageToken is empty; 100 is the per-page size, not a total-result cap. The label adoption scan therefore covers all pages.

pkg/meta/db_pool.go
srstack: Confirmed. The persisted soft-cap latch and runtime reopen threshold remain separate, and reserve/delete transitions remain atomic.

pkg/meta/db_pool.go
srstack: The strict headroom is intentional: the physical shared DB spending limit must remain strictly greater than the sum of tenant virtual limits. The invariant is now documented beside the managed-pool policy in 7dc13b49, and finder/reserve/quota-update checks remain consistent.

pkg/server/tenant_delete.go
srstack: Documented in 7dc13b49. Placement release and capacity decrement intentionally happen before membership deletion; a later membership failure leaves the tenant deleting while capacity accounting remains correct and retry-safe.

pkg/meta/meta.go
srstack: Fixed in e4f9d2e6. quota_limits_overridden remains physically present for rolling-upgrade and rollback compatibility, while new code no longer reads or writes it. The physical drop is deferred to a later release.

pkg/server/shared_db_pool.go
srstack: Addressed in e4f9d2e6 without adding a new worker or configuration. The existing continuation worker retries once after one second; the existing one-minute pending-tenant sweep also resumes managed shared DB pools using default credentials; and emergency fallback schedules continuation with the request credentials.

pkg/tenant/tidbcloudnative/provisioner.go
srstack: Fixed in e4f9d2e6 by removing drive9.ai/status=provisioning from shared cluster creation. No shared Cloud-label lifecycle exists, and the runbooks consume the Prometheus control-plane gauges rather than this Cloud label.

pkg/server/admin_tenant_pool.go
srstack: Fixed in e4f9d2e6. Shared claim now resolves fs_id, placement, and DB-pool metadata before the membership/key transaction commits, so transient post-commit reads cannot strand an undisclosed claimed token.

pkg/tenant/pool.go
srstack: No change to sharedDBOpenMu. Deleting a per-key mutex while an opener still holds it can create two mutexes for the same DB pool and break open/invalidate serialization. The map scales with physical DB-pool IDs seen by the pod, not tenant IDs, so retaining these small entries is the safer bounded tradeoff. InvalidateSharedDB remains the explicit invalidation hook for future credential or endpoint changes.

srstack:
## Re-review — updated through `05d9d168`

Re-reviewed the 3 commits since `a168dd99` (`7dc13b49` allocation reviews, `e4f9d2e6` lifecycle gaps, `05d9d168` import fix) — +340/−79 across 10 files. Build passes; ran the 4 new regression tests + the meta `CompleteSharedTenantProvision`/`FindSharedDBForEmergency` suites, all green. This round closes the two most important issues I raised in my very first server-layer pass. Details:

### Allocation-race fixes — verified correct + covered

1. **Direct-create no longer permanently fails on a capacity/spending race** (`pkg/server/server.go`). The direct shared-create path is now wrapped in a 2-attempt loop: `isSharedDBReservationConflict` (capacity-exhausted / spending-exceeded) triggers a re-allocation onto another pool via `continue`, instead of marking the tenant `TenantFailed`. This is exactly the §8.4 violation I flagged originally ("the loser must not mark the tenant permanently failed merely because the selected pool became full"). The emergency-fallback branch got the same conflict-retry treatment. ✅

2. **select-then-reserve race window closed** (`pkg/server/server.go`). New `provisionTenantOnManagedSharedDB` re-loads the pool and performs the reservation **inside** `WithSharedDBAllocationLock`, rather than selecting under the lock and reserving after release. This removes the window where a concurrent request could take the last soft slot / spending headroom between selection and the conditional UPDATE. ✅

3. **`MarkSharedDBPoolFailed` refuses pools with placed tenants** (`pkg/meta/db_pool.go`). Added `AND tenant_count = 0` to the guard and an explicit error when a `provisioning` pool still has `tenant_count > 0`, so a pool that already carries reserved tenants can't be flipped to `failed` out from under them. Covered by `TestMarkSharedDBPoolFailedRejectsPoolWithTenants`. ✅

4. **Emergency hard-cap is now runtime-derived** (`pkg/meta/db_pool.go`). `FindSharedDBForEmergency` takes a `hardCapRatio` and filters on `tenant_count < CEIL(max_tenants * ratio)` instead of a caller-passed absolute `hardCap`, so the emergency ceiling tracks each candidate pool's own capacity. Covered by `TestProvisionTiDBCloudNativeSharedEmergencyUsesCandidateHardCap`. The capacity-race retry is covered by `TestProvisionTiDBCloudNativeSharedRetriesCapacityRace`, and total batch failure by `TestManagedSharedDBBatchCreateReturnsTotalFailure`. ✅

### Lifecycle / cleanup

5. **`quota_limits_overridden` fully restored** (`pkg/meta/meta.go`). The `removeQuotaLimitsOverriddenColumn` migration was dropped and the column re-added to the base `tenant_quota_config` schema. Combined with the read-time file-size fallback restored earlier, this fully backs out the risky drop-column path that my first review flagged as a silent fail-open (M1). ✅

6. **Delete ordering documented** (`pkg/server/tenant_delete.go`). Capacity is released before membership removal, with a comment explaining the retry-safe rationale (a failed delete leaves the tenant `deleting` without hiding an already-freed slot). Reasonable and consistent with the idempotent delete model. ✅

7. **Leader resume wired** (`pkg/server/server.go`). `resumeManagedSharedDBPoolsWithCtx` is now called on the leader ticker alongside pending-tenant resume — so provisioning pools left mid-flight (including the failed subset of a partial batch) get retried. ✅

8. **Dead constant removed** — `SharedDBPoolStatusProvisioning` and the now-unused pool-status label on batch create were dropped; cluster labels now carry only managed/provider/db_pool_id. Clean.

All 4 CodeRabbit inline findings are withdrawn or marked addressed; nothing outstanding. The `CHANGES_REQUESTED` state reflects the earlier bot review — the latest commits resolve it.

**LGTM.** Across three review rounds this PR has tightened from "functional with a real SQL bug + several concurrency gaps" to a solid, race-aware implementation where every fix carries a targeted regression test. No blocking issues remain from my side.

pkg/tenant/tidbcloudnative/provisioner.go
```diff
@@ -474,6 +487,207 @@ func (p *Provisioner) BatchProvisionFreeClustersWithCredentialsAndQuota(ctx cont
...
+	wantID := strconv.FormatInt(dbPoolID, 10)
+	var matches []clusterInfo
+	if strings.TrimSpace(clusterID) != "" {
+		info, err := p.getClusterBasicInfoWithCredentials(ctx, publicKey, privateKey, strings.TrimSpace(clusterID))
+		if err != nil {
+			return nil, err
+		}
+		matches = append(matches, *info)
+	} else {
+		infos, err := p.listClusterInfosWithCredentials(ctx, publicKey, privateKey, nil, 100)
```
mornyx, line 622: **Follow-up (non-blocking) — label scan still capped at 100 clusters**

Restating the earlier comment since this is the one open item: when `clusterID` is empty, orgs with more than 100 managed clusters can miss the labeled shared pool here, attempt a second create, and then fail closed on ambiguity. The steady-state path (durable `cluster_id`) is unaffected, so the dangerous window is early resume before `cluster_id` is persisted — but orgs with >100 native tenants already exist, so this will eventually bite.

Please paginate `listClusterInfosWithCredentials` here (or make the create-then-load sequence tolerate the miss) as a fast follow-up. Not blocking this PR because the failure mode is closed (ambiguous → failed/refuse), never a wrong-pool attach.

pkg/meta/db_pool.go
```diff
@@ -66,10 +121,397 @@ type TenantPlacement struct {
...
+		if errors.Is(err, sql.ErrNoRows) {
+			return nil, ErrNotFound
+		}
+		return nil, fmt.Errorf("find shared db for allocation: %w", err)
+	}
+	return db, nil
+}
+
+// FindSharedDBForEmergency selects the least-loaded active managed pool below
+// its own runtime-derived hard cap. It is used only after a physical create
```
mornyx, line 487: **Nit — `DRIVE9_TIDBCLOUD_NATIVE_SHARED_HARD_CAP_RATIO=1` silently disables emergency placement**

Env validation accepts ratio = 1.0, but this check rejects `hardCapRatio <= 1`, so with ratio exactly 1 the emergency fallback always errors and a physical-create failure degrades to 502 + failed tenant. Semantically fine (zero headroom means no emergency capacity), but an operator setting 1.0 gets no signal that they disabled the feature. Either reject 1.0 at env parse time or document the equivalence.

mornyx (approved):
Re-reviewed the full PR at `05d9d16` (new commits `7dc13b4`, `e4f9d2e`, `05d9d16`; CI green, local build passes). Verified every outstanding item from the previous two review rounds against the current tree:

**Previous review round (7 items)**
1. ✅ Direct allocate/place race — placement now re-reads the pool and reserves inside `WithSharedDBAllocationLock` (`provisionTenantOnManagedSharedDB`), and reservation conflicts retry the whole allocate+place once instead of failing.
2. ✅ 409s as permanent failures — `isSharedDBReservationConflict` now `continue`s; the tenant is left `pending` and the wrapped `provisionTenantError` still surfaces HTTP 409 via `errors.As` in `handleProvision`. The emergency branch got the same treatment.
3. ✅ Emergency hard cap — now derived per candidate in SQL (`tenant_count < CEIL(max_tenants * ratio)`) and the reserve cap is computed from `fallback.MaxTenants`. `TestProvisionTiDBCloudNativeSharedEmergencyUsesCandidateHardCap` proves the soft=50@60 vs soft=100@100 selection.
4. ✅ Org lock across physical work — now explicitly documented as an intentional tradeoff (`continueManagedSharedDBPoolLocked` comment), which is the resolution the original comment offered for a low-QPS launch.
5. ✅ Batch create total failure — `provisionManagedSharedDBPoolsBatchLocked` returns `createErr` and the log event is renamed (`managed_shared_db_batch_create_failed`), with `TestManagedSharedDBBatchCreateReturnsTotalFailure`.
6. ✅ `MarkSharedDBPoolFailed` — guarded by `tenant_count = 0` with an explicit error otherwise (`TestMarkSharedDBPoolFailedRejectsPoolWithTenants`).
7. ⚠️ 100-cluster label-scan cap — **still open**, see inline.

**Last round (5 items)**
1. ✅ `quota_limits_overridden` DROP COLUMN reverted; column restored in init schema, so old binaries keep working through a rolling upgrade.
2. ✅ Continuation retry — `resumeManagedSharedDBPoolsWithCtx` now runs on the 1-minute leader reconciler tick, plus one inline retry in the continuation worker; the emergency path also schedules a continuation for the failed planned pool.
3. ✅ Cloud status label — dropped from the create payload; labels are now managed/provider/db_pool_id only.
4. ✅ Shared claim — `ResolveFsID`/`GetTenantPlacement`/`GetSharedDB` now happen before the claim transaction, so a transient read failure can no longer strand a used member with an undeliverable owner token.
5. ⚠️ `sharedDBOpenMu` nit — still open (nit, fine to defer).

The new regression tests assert real behavior (candidate-sized hard cap, concurrent provisions landing one tenant per pool, total batch failure propagation). Good work.

Approving. The only remaining substantive item is the 100-cluster label-scan cap (inline) — it fails closed, so I don't consider it merge-blocking, but please track it as a fast follow before shared pools go anywhere near an org with >100 clusters.

pkg/tenant/tidbcloudnative/provisioner.go
srstack: Rechecked against the current implementation: the 100 argument is only the per-page size. listClusterInfosWithCredentials (provisioner.go:1362) loops listClusterInfosPageWithCredentials with pageToken, appends every page, and returns only when nextPageToken is empty. LoadSharedDBPoolWithCredentials therefore already scans beyond the first 100 managed clusters. No code change is needed for this thread.

pkg/meta/db_pool.go
srstack: Fixed in d50c5bda. DRIVE9_TIDBCLOUD_NATIVE_SHARED_HARD_CAP_RATIO now rejects values <= 1 at startup, the error and help text specify a finite value > 1, and the env parser boundary test includes 1.0. This keeps emergency headroom enabled whenever the configuration is accepted.

mornyx (commented):
Incremental review of `d50c5bd` (reject zero shared hard-cap headroom).

This closes the nit from my last round exactly as suggested: env parse now rejects `ratio <= 1`, the error message and usage text say `> 1`, and `TestSharedDBHardCapRatioFromEnv` covers the `"1"` case. ✅

One optional micro-alignment, feel free to ignore: `NewWithConfig` and `managedSharedDBHardCap` still normalize with `< 1` (so a programmatically supplied `SharedDBHardCapRatio: 1.0` passes through and disables emergency placement the same way env=1 used to). Only reachable from tests/internal callers since env now guards the real path, but switching both normalizations to `<= 1` would make the invariant uniform.

My approval from the previous round stands. CI is still running on this commit — please merge once it goes green.

PR #775, opened 2026-07-22
fix: map backend errors to appropriate HTTP status codes

srstack:
## Summary

Previously all backend errors returned HTTP 500, causing:
- Client disconnects (`context.Canceled`) to show as server errors
- Upstream timeouts (`context.DeadlineExceeded`) to mask as internal errors
- TiDB quota exhaustion (`Error 1105`) to trigger 500 retry storms
- Schema migration failures to pollute error monitoring

## Changes

- Add `backendErrorStatus()` to map error types to HTTP status codes:

| Error | Old Status | New Status |
|---|---|---|
| `context.Canceled` | 500 | **499** |
| `context.DeadlineExceeded` | 500 | **504** |
| TiDB quota 1105 | 500 | **402** |
| Schema migration errors | 500 | **502** |
| Unknown errors | 500 | 500 |

- Add `writeBackendError()` as a unified error response helper
- Extract `isQuotaError()` and `isSchemaMigrationError()` from `sanitizeClientError()`
- Fix 2 sites that leaked `err.Error()` to clients in `admin_tenant_pool.go`
- Update `errJSONInternalStorage` to accept request context for proper status
- Convert all 107 hardcoded-500 sites to use `backendErrorStatus()` or `writeBackendError()`

## Files Changed

13 files, +167 / -132 lines in `pkg/server/`

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved backend-aware HTTP error handling across authentication, storage/filesystem, tenants, quota, tokens, Vault, uploads, and related flows.
  - Responses now use more accurate status codes derived from the underlying backend error (including payment required, gateway failure, client cancellation, and request timeout), while preserving existing behavior for validation, not-found, conflict, and authorization errors.
  - Standardized client-safe error bodies for backend and internal storage failures.
- **Tests**
  - Added coverage for backend error classification and status-code mapping.

### DIFF
--- a/pkg/server/auth.go
+++ b/pkg/server/auth.go
@@ -62,7 +62,8 @@ func isQuotaError(err error) bool {
 		return strings.Contains(strings.ToLower(mysqlErr.Message), "quota")
 	}
 	msg := strings.ToLower(err.Error())
-	return strings.Contains(msg, "error 1105") && strings.Contains(msg, "quota")
+	idx := strings.Index(msg, "error 1105")
+	return idx >= 0 && strings.Contains(msg[idx:], "quota")
 }
 
 func isSchemaMigrationError(err error) bool {
--- a/pkg/server/auth.go
+++ b/pkg/server/auth.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/go-sql-driver/mysql"
 	"github.com/mem9-ai/drive9/pkg/backend"
 	"github.com/mem9-ai/drive9/pkg/logger"
 	"github.com/mem9-ai/drive9/pkg/meta"
@@ -55,6 +56,10 @@ const statusClientClosedRequest = 499
 func isQuotaError(err error) bool {
 	if err == nil {
 		return false
+	}
+	var mysqlErr *mysql.MySQLError
+	if errors.As(err, &mysqlErr) && mysqlErr.Number == 1105 {
+		return strings.Contains(strings.ToLower(mysqlErr.Message), "quota")
 	}
 	msg := strings.ToLower(err.Error())
 	return strings.Contains(msg, "error 1105") && strings.Contains(msg, "quota")
--- a/pkg/server/auth.go
+++ b/pkg/server/auth.go
@@ -52,21 +52,56 @@ const (
 
 const statusClientClosedRequest = 499
 
+func isQuotaError(err error) bool {
+	if err == nil {
+		return false
+	}
+	msg := strings.ToLower(err.Error())
+	return strings.Contains(msg, "error 1105") && strings.Contains(msg, "quota")
+}
+
+func isSchemaMigrationError(err error) bool {
+	if err == nil {
+		return false
+	}
+	msg := strings.ToLower(err.Error())
+	return strings.Contains(msg, "ensure tidb auto-embedding schema") ||
+		strings.Contains(msg, "validate tidb auto-embedding schema") ||
+		strings.Contains(msg, "migrate split tables") ||
+		strings.Contains(msg, "detect legacy files table")
+}
+
 func sanitizeClientError(err error) string {
 	if err == nil {
 		return "backend unavailable"
 	}
-	msg := strings.ToLower(err.Error())
-	if strings.Contains(msg, "error 1105") && strings.Contains(msg, "quota") {
+	if isQuotaError(err) {
 		return "tenant usage quota exceeded"
 	}
-	if strings.Contains(msg, "ensure tidb auto-embedding schema") ||
-		strings.Contains(msg, "validate tidb auto-embedding schema") ||
-		strings.Contains(msg, "migrate split tables") ||
-		strings.Contains(msg, "detect legacy files table") {
+	if isSchemaMigrationError(err) {
 		return "tenant metadata error"
 	}
 	return "backend unavailable"
+}
+
+func backendErrorStatus(ctx context.Context, err error) int {
+	if isClientCanceled(ctx, err) {
+		return statusClientClosedRequest
+	}
+	if errors.Is(err, context.DeadlineExceeded) {
+		return http.StatusGatewayTimeout
+	}
+	if isQuotaError(err) {
+		return http.StatusPaymentRequired
+	}
+	if isSchemaMigrationError(err) {
+		return http.StatusBadGateway
+	}
+	return http.StatusInternalServerError
+}
+
+func writeBackendError(w http.ResponseWriter, r *http.Request, err error) {
+	errJSON(w, backendErrorStatus(r.Context(), err), sanitizeClientError(err))
 }
 
 func ScopeFromContext(ctx context.Context) *TenantScope {
@@ -195,7 +230,7 @@ func tenantAuthMiddlewareWithFSScopeLoader(metaStore *meta.Store, pool *tenant.P
 			}
 			logger.Error(r.Context(), "server_event", eventFields(r.Context(), "auth_backend_unavailable", "error", err)...)
 			metricEvent(r.Context(), "auth", "result", "meta_backend_error")
-			errJSON(w, http.StatusInternalServerError, "auth backend unavailable")
+			errJSON(w, backendErrorStatus(r.Context(), err), "auth backend unavailable")
 			return
 		}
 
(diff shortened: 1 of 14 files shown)


srstack:
## Review

Well-scoped, well-executed refactor. The extraction of `isQuotaError`/`isSchemaMigrationError` out of `sanitizeClientError` is clean, `backendErrorStatus` and `writeBackendError` centralize the mapping, and the 107-site conversion is mechanical and consistent. Verified `go build ./...` and `go vet ./pkg/server` clean; existing `TestSanitizeClientError_*` and `TestAuthClientCanceled*` pass. The status choices are sensible (499 client-closed, 504 deadline, 402 quota, 502 schema-migration).

### Genuinely valuable: the two info-leak fixes
`handleAdminTenantPoolGet` and `writeAdminTenantPoolError` previously passed raw `err.Error()` to the client; both now route through `writeBackendError`/`sanitizeClientError`. That's a real hardening win independent of the status-code work, and worth calling out as the highest-value part of this PR.

### Points worth addressing

**1. No direct test for `backendErrorStatus` — the core of the PR is untested.**
There are thorough `sanitizeClientError` table tests, but nothing exercises the new error→status mapping directly. A table test over `backendErrorStatus(ctx, err)` covering all five branches (canceled ctx → 499, `DeadlineExceeded` → 504, quota → 402, schema-migration → 502, unknown → 500) would lock in the contract cheaply. The existing `StatusBadGateway` assertions in provision/quota tests cover the schema path incidentally, but 499/504/402 have no coverage.

**2. `backendErrorStatus(ctx, nil)` at the batch-write count-mismatch site.**
In `handleBatchWrite`, the `len(backendResults) != len(allowedIndexes)` branch is reached only when `err == nil` (it's past the `if err != nil` return), yet it now calls `errJSONInternalStorage(w, r, err)` with `err == nil`. `backendErrorStatus(ctx, nil)` → `isClientCanceled(ctx, nil)` inspects `ctx.Err()`, so if the client happened to cancel, a genuine server-side invariant violation (result-count mismatch) would report as 499 instead of 500. Minor and unlikely, but this site should pass an explicit sentinel (e.g. a `fmt.Errorf("batch write result count mismatch")`) rather than nil so it deterministically maps to 500.

**3. Quota detection stays substring-based while a typed error exists.**
`isQuotaError` matches `strings.Contains(lower, "error 1105") && contains "quota"`. `pkg/tenant/pool.go` already inspects the *typed* `*mysql.MySQLError{Number: 1105}` structurally (`mysqlErrUsageQuotaExhaustedCandidate`). The substring form is locale/format-fragile — a driver or TiDB message wording change silently drops the 402 back to 500. This is pre-existing behavior (the PR only moved it), so not a blocker, but since the PR is specifically about correct quota status codes, considering `errors.As` on the typed error (or a shared sentinel from the tenant/backend layer) would make the 402 robust.

### Nits
- `errJSONInternalStorage` now takes `(w, r, err)` but always emits the same fixed `internalStorageErrorMessage` — good, message is unchanged; only the status is now dynamic. Correct.
- The 499 "client closed request" status is nginx-convention (non-standard). Consistent with the existing `statusClientClosedRequest` const already in the codebase, so fine.

Net: correct, safe, and an improvement on two fronts (accurate status codes + closing two info leaks). #1 and #2 are the two I'd want addressed before merge; #3 is a reasonable follow-up.

srstack:
## Re-review at `410ee39` — all three points resolved

Verified the follow-up commit against my earlier review. Build clean; `TestBackendErrorStatus` and `TestSanitizeClientError_*` all pass.

1. **`backendErrorStatus` now tested — FIXED.** `TestBackendErrorStatus` is a table test covering all branches: canceled ctx → 499, `DeadlineExceeded` → 504, quota → 402, schema-migration → 502, unknown → 500, and nil → 500. It correctly builds a real canceled/timed-out context per case so the `isClientCanceled`/`DeadlineExceeded` paths are genuinely exercised, not just the error-value branches. This locks in the contract.

2. **`handleBatchWrite` nil-error — FIXED.** The count-mismatch branch now passes an explicit `fmt.Errorf("batch write result count mismatch: expected %d, got %d", ...)` sentinel instead of the nil `err`, so it deterministically maps to 500 regardless of ctx state. Bonus: the error string is now descriptive for the logs.

3. **Quota detection now typed — FIXED.** `isQuotaError` tries `errors.As(err, &mysqlErr)` on `*mysql.MySQLError` with `Number == 1105` first, falling back to the substring match only for unwrapped/string errors. This makes the 402 robust to driver/message-wording changes while preserving the old behavior for errors that arrive as plain strings.

The vault.go comment is also updated to match the new mapping. No remaining concerns — LGTM.

srstack:
## Re-review at `6ffe58a` — CodeRabbit follow-up looks good

Verified the latest commit. Build clean, `TestBackendErrorStatus` passes.

- **`t.Fatalf` → `t.Errorf` in the table test:** correct — a subtest failure no longer aborts the remaining cases, so one bad branch doesn't mask others.
- **New schema-contract patterns in `isSchemaMigrationError`:** verified both strings are real. `meta schema contract mismatch` is emitted at `pkg/meta/meta.go:569` and `tidb schema contract mismatch` at `pkg/tenant/schema/tidb_auto.go:740`, both lowercase at the source, so the `strings.ToLower` match is sound. These correctly route schema-contract failures to 502 rather than 500.
- **slock HTML path:** good catch by the bot — `backendErrorStatus` is now computed once and reused for both the JSON and HTML responses, so the HTML branch no longer hardcodes 500 while JSON returns the mapped status. The two responses are now consistent.

No concerns. LGTM.

pkg/server/auth.go
```diff
@@ -52,23 +53,64 @@ const (
...
+	}
+	var mysqlErr *mysql.MySQLError
+	if errors.As(err, &mysqlErr) && mysqlErr.Number == 1105 {
+		return strings.Contains(strings.ToLower(mysqlErr.Message), "quota")
 	}
 	msg := strings.ToLower(err.Error())
-	if strings.Contains(msg, "error 1105") && strings.Contains(msg, "quota") {
-		return "tenant usage quota exceeded"
+	return strings.Contains(msg, "error 1105") && strings.Contains(msg, "quota")
+}
```
mornyx, line 67: **Medium — string fallback false-positives when wrapper text contains "quota"**

The `*mysql.MySQLError` branch is precise (1105 **and** the MySQL message itself mentions quota). The string fallback, however, runs over the whole error chain: any error wrapped with quota-related text — e.g. `fmt.Errorf("set quota config for tenant %q: %w", ...)` or "sum shared tenant spending limits" — that contains a *generic* TiDB 1105 (connection/GC/region errors all surface as 1105) will be misclassified as quota exhaustion. Previously this only rewrote the message; now it also changes the HTTP status to 402, so a TiDB connectivity incident would present as "payment required / tenant usage quota exceeded" on quota code paths, misleading both clients and 402-keyed alerting.

Cheap fix: in the fallback, only search for "quota" in the segment *after* "error 1105":

```go
idx := strings.Index(msg, "error 1105")
return idx >= 0 && strings.Contains(msg[idx:], "quota")
```

That keeps genuine wrapped quota errors (the MySQL message trails the code) while ignoring coincidental "quota" in wrapper prefixes. A regression test with `"set quota config: Error 1105 (HY000): connection refused"` → 500 would pin it down.

mornyx (approved):
Reviewed at `1cf48e8` (4 commits, +242/−132 across 14 files; CI green, local build passes).

The conversion is clean: all ~100 sites are genuine backend-error paths, the fixed-message variants (`backendErrorStatus(ctx, err), "msg"`) and sanitized variants (`writeBackendError`) are used in the right places, and the two sites that leaked raw `err.Error()` to clients (`writeAdminTenantPoolError`, `handleAdminTenantPoolGet`) are real security/hygiene improvements. The `tenant_delete.go` follow-up correctly switched to `enqueueErr` in the conflict-resolution branch, and the slock HTML path now shares the mapped status.

Verified the earlier bot feedback is resolved: `TestBackendErrorStatus` covers the mapping table (with `t.Errorf` per case), the schema-contract patterns (`meta schema contract mismatch` / `tidb schema contract mismatch`) were added, and the stale `insert grant` comment now describes the mapped behavior.

On the mapping choices themselves: requiring the `quota` keyword alongside 1105 avoids the obvious false positive (1105 is TiDB's generic ER_UNKNOWN_ERROR), 499/504/502 are all defensible, and the Go SDK has no generic 5xx retry loop, so the new statuses don't fight client-side retry behavior.

One medium finding inline — the string fallback in `isQuotaError` can false-positive on wrapper text that happens to contain "quota", which now changes the *status* (402), not just the message. Not merge-blocking since the primary `*mysql.MySQLError` path is precise and the blast radius is misleading-status-during-outage, but the anchor fix is a few lines.

Approving.

PR #778, opened 2026-07-22
fix: propagate shared org and quota sizes

srstack:
## Summary

- resolve shared tenant TiDB Cloud organization IDs from the physical shared DB pool during API-key authentication and backend creation
- propagate the shared pool organization through provisioning, admin item requests, tenant notify dispatch, quota mutation replay, request logs, server-event logs, and Prometheus metrics instead of falling back to `guest`
- clear the previous mutation-backlog gauge label when a tenant organization label changes, so the old `guest` series does not remain non-zero
- convert admin-create storage and file quota values from MiB to bytes through the existing quota patch conversion path

## Root cause

Shared tenants do not have `tenant_tidbcloud_org_bindings` rows, but several request and background paths only consulted that dedicated-native mapping. The shared provision result also omitted the pool organization. In addition, `server_event` fields only read `TenantScope`, while `/v1/status` resolves the tenant into request metric state without installing a scope. Separately, shared quota materialization persisted raw MiB request values directly into byte columns.

Shared provider dispatch now consistently resolves the organization through `fs_registry -> tenant_placements -> db_pool.org_id`; dedicated native continues to use `tenant_tidbcloud_org_bindings.organization_id`.

## Verification

- `GOCACHE=/tmp/drive9-go-cache go test ./pkg/server -run "^(TestEventFieldsDoesNotDuplicateExplicitTiDBCloudOrgID|TestSharedTenantStatusLogsAndMetricsUseDBOrganization|TestTenantOutboxPollerPassesOrgToWorker)$" -count=1`
- `GOCACHE=/tmp/drive9-go-cache go test ./pkg/backend -run "^(TestMutationReplayWorkerRecordsPendingBacklogGauges|TestMutationReplayWorkerClearsPreviousOrganizationGauge)$" -count=1`
- `GOCACHE=/tmp/drive9-go-cache go test ./pkg/meta -count=1`
- `GOCACHE=/tmp/drive9-go-cache go build ./...`
- `GOCACHE=/tmp/drive9-go-cache GOLANGCI_LINT_CACHE=/tmp/drive9-golangci-cache make lint`
- `git diff --check`

The full `pkg/server` run was also attempted locally, but the repository-wide suite hit the known MySQL test lifecycle failure class (`database is closed` / canceled localhost lookups) after many server instances were started. The focused shared output tests above pass.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved TiDB Cloud organization identification for shared tenants across provisioning, authorization, status reporting, and background processing.
  - Corrected logs and metrics to consistently show the organization associated with a shared database.
  - Prevented duplicate organization labels in request telemetry.
  - Cleared stale mutation backlog metrics when a tenant moves between organizations.
  - Ensured shared quota limits are persisted accurately, including byte-based storage values.

### DIFF
--- a/pkg/meta/meta_test.go
+++ b/pkg/meta/meta_test.go
@@ -27,6 +27,35 @@ func newControlStore(t *testing.T) *Store {
 	return s
 }
 
+func insertSharedTenantPlacementForOrgTest(t *testing.T, s *Store, tenantID, organizationID string) {
+	t.Helper()
+	ctx := context.Background()
+	now := time.Now().UTC()
+	if err := s.InsertTenant(ctx, &Tenant{
+		ID: tenantID, Status: TenantActive, Kind: TenantKindLive,
+		Provider: tidbCloudNativeSharedProvider, DBPasswordCipher: []byte{}, DBTLS: true,
+		SchemaVersion: 1, CreatedAt: now, UpdatedAt: now,
+	}); err != nil {
+		t.Fatal(err)
+	}
+	fsID, err := s.EnsureFsID(ctx, tenantID)
+	if err != nil {
+		t.Fatal(err)
+	}
+	dbID, err := s.RegisterSharedDB(ctx, &SharedDB{
+		TiDBCloudOrganizationID: organizationID, Host: "shared.example.com", Port: 4000,
+		User: "root", PasswordCipher: []byte("cipher"), Name: "shared_db",
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+	if err := s.UpsertTenantPlacement(ctx, &TenantPlacement{
+		FsID: fsID, DbID: dbID, Placement: PlacementShared, SchemaShape: SchemaShapeShared,
+	}); err != nil {
+		t.Fatal(err)
+	}
+}
+
 func TestMetaDBMetrics(t *testing.T) {
 	s := newControlStore(t)
 	now := time.Now().UTC()
@@ -261,29 +290,7 @@ func TestResolveByAPIKeyHashUsesSharedDBOrganization(t *testing.T) {
 	ctx := context.Background()
 	now := time.Now().UTC()
 	tenantID := "shared-api-key-org"
-	if err := s.InsertTenant(ctx, &Tenant{
-		ID: tenantID, Status: TenantActive, Kind: TenantKindLive,
-		Provider: tidbCloudNativeSharedProvider, DBPasswordCipher: []byte{}, DBTLS: true,
-		SchemaVersion: 1, CreatedAt: now, UpdatedAt: now,
-	}); err != nil {
-		t.Fatal(err)
-	}
-	fsID, err := s.EnsureFsID(ctx, tenantID)
-	if err != nil {
-		t.Fatal(err)
-	}
-	dbID, err := s.RegisterSharedDB(ctx, &SharedDB{
-		TiDBCloudOrganizationID: "org-shared-api-key", Host: "shared.example.com", Port: 4000,
-		User: "root", PasswordCipher: []byte("cipher"), Name: "shared_db",
-	})
-	if err != nil {
-		t.Fatal(err)
-	}
-	if err := s.UpsertTenantPlacement(ctx, &TenantPlacement{
-		FsID: fsID, DbID: dbID, Placement: PlacementShared, SchemaShape: SchemaShapeShared,
-	}); err != nil {
-		t.Fatal(err)
-	}
+	insertSharedTenantPlacementForOrgTest(t, s, tenantID, "org-shared-api-key")
 	if err := s.InsertAPIKey(ctx, &APIKey{
 		ID: "shared-api-key", TenantID: tenantID, KeyName: "default",
 		JWTCiphertext: []byte("jwt-cipher"), JWTHash: "shared-api-key-hash", TokenVersion: 1,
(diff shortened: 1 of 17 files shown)


srstack:
## Review — propagate shared org and quota sizes

Reviewed both commits (`ceb63644` + `b923b0f2`). Built in an isolated worktree — `go build ./...`, `pkg/meta`, `pkg/tenant`, and the shared/quota/mutation/notify/provision `pkg/server` tests all pass. Two correct fixes, both reusing existing paths rather than adding new logic.

### MiB→bytes quota fix — CONFIRMED correct (the higher-risk one)

`materializeSharedTenantQuota` (`shared_db_pool.go`) previously assigned request values **straight into byte columns**: `cfg.MaxStorageBytes = *opts.Quota.MaxStorageSize`. Since the request fields are MiB, a shared tenant's storage/file-size quota was persisted 2^20× too small (a 1 MiB request → 1 byte). The fix routes through `quotaConfigPatchFromRequest(req)` — the same conversion the dedicated path already uses (`quotaStorageSizeToBytes` / `quotaFileSizeToBytes`) — then reads `MaxStorageBytes`/`MaxFileSizeBytes` from the patch. Shared and dedicated now share one conversion path, so the unit contract can't drift again. Validation still runs first (`validateQuotaSetRequest`). Correct.

### Shared org resolution — CONFIRMED correct and consistent

The root cause is accurate: shared tenants have no `tenant_tidbcloud_org_bindings` row, so paths that only consulted that dedicated mapping fell back to `guest`. The fix resolves shared org via `fs_registry → tenant_placements → db_pool.org_id` everywhere:
- `tenantMetricTiDBCloudOrgID` (pool.go): gate widened from `== ProviderTiDBCloudNative` to `UsesTiDBCloudNativeCredentials` (dedicated OR shared), then `IsSharedSchemaProvider` branches to `GetSharedDBForTenant`. Fails closed to the default org on lookup error / empty org.
- `provisionTenantOnSharedDBMode` now populates `OrganizationID` on the result (was omitted).
- `authorizedAdminTenant` now emits `setRequestMetricTenant` with the resolved org for shared (from the authorized physical pool) and both dedicated branches.
- `ListPendingMutations` / `ObservePendingMutations` (quota.go): `COALESCE(CASE WHEN t.provider = shared THEN d.org_id ELSE b.organization_id END, '')` via added `fs_registry/tenant_placements/db_pool` joins. Dedicated continues to use the binding.

I verified the mutation-observe `GROUP BY t.provider, d.org_id, b.organization_id` cannot double-count: `fs_registry.tenant_id` is UNIQUE and `tenant_placements` is PK'd on `fs_id`, so the tenant→fs→placement→db_pool chain yields at most one row per tenant; the CASE picks a single org value per provider, so a tenant maps to exactly one group. `FORCE INDEX (idx_pending_tenant_age)` is preserved.

### Notes

- The org resolution adds a 3-table LEFT JOIN to the two mutation queries. Both are background/observability paths (replay sweep + backlog metric), not request hot paths, so the extra joins are acceptable. `ListPendingMutations` keeps its `LIMIT`.
- All resolution failures fail closed to `defaultTenantMetricTiDBCloudOrgID` rather than erroring the caller — correct for a metric-label resolver.

LGTM. Both fixes are correct, reuse the established conversion/resolution paths, and carry regression tests (meta_test +139, tenant/pool_test, provision/delete-shared tests).

pkg/meta/meta_test.go
```diff
@@ -27,6 +27,35 @@ func newControlStore(t *testing.T) *Store {
...
+	})
+	if err != nil {
+		t.Fatal(err)
+	}
+	if err := s.UpsertTenantPlacement(ctx, &TenantPlacement{
+		FsID: fsID, DbID: dbID, Placement: PlacementShared, SchemaShape: SchemaShapeShared,
+	}); err != nil {
+		t.Fatal(err)
+	}
+}
```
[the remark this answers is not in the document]
srstack: Fixed in 8c9dd5c2. The fixture now registers ResetMetaDB through t.Cleanup before inserting persistent rows; cleanup runs before the Store close cleanup.

srstack:
Addressed the latest review in `8c9dd5c2`: the shared-placement fixture now registers Meta DB cleanup, and `assertOrgLog` explicitly accepts `*testing.T` per the repository helper convention. Focused meta/server tests, `go build ./...`, `make lint`, and `git diff --check` pass.

srstack:
## Re-review — updated through `2d4f11fc` ("preserve shared org in logs and metrics")

Rebuilt the new head in an isolated worktree — `go build ./...`, `pkg/backend`, and `pkg/server` tests pass. This commit closes a real metric-series leak that the earlier org-resolution fix would otherwise have introduced.

### The fix — CONFIRMED correct

`recordPendingBacklog` (`pkg/backend/mutation_replay.go`) now detects when a tenant's resolved org changed between observations (`previousOrgID != orgID`) and explicitly zeros the **old** org label's `pending_mutations` / `oldest_pending_age_seconds` gauges before recording the new org's values.

Why this matters: those gauges carry `tidbcloud_org_id`. Once the earlier commits in this PR fixed shared-tenant org resolution (was falling back to `guest`, now the real pool org), an in-flight shared tenant transitions its label from `tidbcloud_org_id="guest"` to the real org. Without zeroing the prior series, Prometheus would keep **both** series alive — the backlog double-counted and the stale `guest` series pinned at a nonzero value forever. The fix makes the transition clean. This is the same diff-delete discipline used elsewhere in the shared control plane (e.g. `syncSharedDBPoolSnapshot`), applied correctly here.

### Coverage

- `TestMutationReplayWorkerClearsPreviousOrganizationGauge` — asserts the `guest` series drops to 0 and the new org series records the real count after an org change.
- `TestSharedTenantStatusLogsAndMetricsUseDBOrganization` — end-to-end: a shared tenant's status path resolves org from the physical DB pool (RegisterSharedDB + placement), not `guest`.
- Instrumentation / outbox-poller tests updated to thread `TiDBCloudOrgID` through the fake mutation records.

Everything from my prior review (MiB→bytes quota conversion via `quotaConfigPatchFromRequest`, shared org resolution through `fs_registry → tenant_placements → db_pool.org_id`, the non-double-counting `GROUP BY`) is unchanged and still correct.

LGTM. The org-propagation work is now complete end-to-end — request paths, background paths, logs, and metrics all resolve shared org from the pool, and the metric-label transition is leak-free.

pkg/meta/quota.go
```diff
@@ -1312,12 +1312,18 @@ func (s *Store) ListPendingMutations(ctx context.Context, minAge time.Duration,
 	}
 	cutoff := time.Now().UTC().Add(-minAge)
 	rows, err := s.db.QueryContext(ctx,
-		`SELECT q.id, q.tenant_id, COALESCE(b.organization_id, ''), q.mutation_type, q.mutation_data, q.status, q.retry_count, q.created_at, q.applied_at
+		`SELECT q.id, q.tenant_id,
+		        COALESCE(CASE WHEN t.provider = ? THEN d.org_id ELSE b.organization_id END, ''),
+		        q.mutation_type, q.mutation_data, q.status, q.retry_count, q.created_at, q.applied_at
 		 FROM quota_mutation_log q
+		 LEFT JOIN tenants t ON t.id = q.tenant_id
 		 LEFT JOIN tenant_tidbcloud_org_bindings b ON b.tenant_id = q.tenant_id
```
mornyx, line 1320: **Nit — the shared-org resolution fragment is now duplicated in 4 queries**

The `LEFT JOIN tenants/fs_registry/tenant_placements/db_pool` + `CASE WHEN t.provider = ? THEN d.org_id ELSE b.organization_id END` block appears verbatim in `ResolveByAPIKeyHash`, `ListTenantNotifySince`, `ListPendingMutations`, and here. A fifth site will inevitably appear (quota replay already came close). Consider extracting the JOIN+SELECT fragment as a shared constant (in the spirit of `sharedDBSelectColumns`) or a tiny query-builder helper, so the resolution rule can only drift in one place.

mornyx (approved):
Reviewed at `8c9dd5c` (4 commits, +502/−56 across 17 files; CI green, local build passes).

This is a correct and well-scoped follow-up to the shared-provider work. Verified each thread:

1. **Org resolution** — the `CASE WHEN t.provider = shared THEN d.org_id ELSE b.organization_id END` pattern in `ResolveByAPIKeyHash`, `ListTenantNotifySince`, `ListPendingMutations`, and `ObservePendingMutations` matches the intended placement-based model (shared → `fs_registry → tenant_placements → db_pool.org_id`, dedicated → org bindings). The new JOINs on the auth hot path are all unique-index/PK lookups (`fs_registry.tenant_id` is UNIQUE, so no row fan-out), and `ObservePendingMutations`' expanded GROUP BY is ONLY_FULL_GROUP_BY-safe. Each query has a dedicated regression test.
2. **MiB→bytes quota fix** — this was a real bug shipped in #774: `materializeSharedTenantQuota` stored raw MiB request values into byte columns, so an admin create with `max_storage_size: 500` persisted 500 *bytes*. Routing through `quotaConfigPatchFromRequest` is the right fix (single conversion path), and `TestAdminTenantCreatePersistsSharedQuotaSizesInBytes` pins it end-to-end through the HTTP endpoint. Full disclosure: I reviewed #774 and missed this too — good catch.
3. **Metrics/log org labels** — `provisionTenantResult.OrganizationID` for shared, the three new `setRequestMetricTenant` call sites in `authorizedAdminTenant`, the pool's shared-branch org resolution, and the `eventFields` fallback all check out. The fallback only emits when the metric state actually carries an org (no `guest` spam — `requestMetricScope` returns "" when unset), and the kv pre-scan correctly prevents duplicate fields.
4. **Gauge label cleanup** — zeroing the previous org series on label change closes the stale-`guest`-series hole; tenants that vanish from observations were already handled by the pre-existing path, so the two cases compose correctly.

One nit inline about the now-quadruplicated SQL fragment — not blocking.

Approving.

pkg/meta/quota.go
srstack: I checked the four call sites and am keeping the SQL local for this PR. They do not share a stable fragment: the tenant source is `k.tenant_id`, `o.tenant_id`, or `q.tenant_id`; `ResolveByAPIKeyHash` already owns the `tenants t` join while the other queries add it; and the backlog query has its own FORCE INDEX and GROUP BY requirements. Extracting the full block would require alias-driven string formatting/query building and would hide placeholder ordering, which is more machinery than the duplicated declarative SQL here. Each provider-dispatch query has a regression test that pins shared -> `db_pool.org_id` and dedicated -> org binding. If another production caller appears with the same query shape, that would be a better point to introduce a typed helper.

PR #779, opened 2026-07-23
fix: clean up failed TiDB Cloud tenants

srstack:
## Summary

- add organization-wide failed-tenant selectors and CAS ownership for both `tidb_cloud_native` and `tidb_cloud_native_shared`
- clean failed free-pool and direct tenants while excluding already claimed pool-used tenants
- trigger bounded, best-effort cleanup from the existing claim path with per-Pod organization throttling

## Cleanup behavior

- native cleanup deprovisions the dedicated TiDB Cloud cluster with request-scoped credentials, revokes keys best-effort, then atomically marks the tenant deleted and removes its binding
- shared cleanup never deprovisions the physical shared cluster; it drains the backend, conditionally purges `fs_id`-scoped data when the shared DB is ready, and atomically releases placement, capacity, membership, and tenant metadata
- owned failures restore `deleting -> failed` with an updated retry timestamp and a bounded restore timeout
- selectors wait for a 15-minute tenant cooldown; native batches are limited to 1 and shared batches to 10
- per-Pod cleanup runs at most once per organization per minute; database CAS remains the cross-Pod ownership boundary

## Scope

Included:

- native failed pool bindings with `pool_status=free`
- native direct failed bindings with an empty `pool_id`
- shared failed free memberships
- shared direct failed tenants with no membership and a placement in an organization-owned shared DB

Excluded:

- native claimed bindings with a nonempty `pool_id` and `pool_status=used`
- shared tenants with an existing used membership
- tenants without enough binding, membership, or placement metadata to prove organization ownership

## Test plan

- [x] `GOFLAGS='-count=1' make test TEST_PKGS='./pkg/meta' TEST_RUN='Test(ListFailed|FailedTenantCleanup|MarkFailedTenant|FinalizeSharedTenantDeleteMetadataIsAtomic|TenantPoolMembershipRoundTripAndClaimCAS|ClaimOldestFreeTenantPoolBindingRequiresActiveTenant|OldestFreeTenantPoolMembershipRequiresActiveTenant)'`
- [x] `GOFLAGS='-count=1' make test TEST_PKGS='./pkg/server' TEST_RUN='Test(CleanupFailedOrganizationTenants|TenantFailedCleanupAsync|TenantPoolClaim(ReturnsWithoutWaitingForFailedCleanupAndPassesCredentials|TriggersDirectFailedCleanupWithoutLogicalPool|ResultUnaffectedByFailedCleanupError|UsesNativeInventoryBeforeExternalSharedPool|ConsumesMixedInventoryInGlobalAgeOrder|MissTriggersPendingMetadataResume|SeedsQuotaConfigWithoutExplicitQuota)|DeleteFreeSharedPoolTenantSkipsPurgeWhenDBPoolIsUnready)'`
- [x] `gofmt -d` on all changed Go files
- [x] `git diff --check`

Focused verification was run after rebasing the work onto the latest `origin/main`. No baseline or full-suite test was run.

## Summary by CodeRabbit

* **New Features**
  * Added automatic cleanup for failed native and shared tenants.
  * Cleanup runs asynchronously after tenant pool operations, with per-organization throttling and cooldown controls.
  * Failed tenants are safely claimed, deprovisioned, credentials revoked, shared resources cleaned up, and deletion finalized.
  * Cleanup failures restore tenants to a failed state for later retry.

* **Bug Fixes**
  * Improved concurrency protection to prevent duplicate cleanup.
  * Fork smoke tests now skip unsupported deployments instead of failing.

### DIFF
--- a/e2e/native-smoke-test.sh
+++ b/e2e/native-smoke-test.sh
@@ -310,7 +310,20 @@ FORK_LOCAL="$(mktemp)"
 
 drive9_ctx ctx add --name owner --server "$BASE" --api-key "$API_KEY" --mode TiDBCloud >/dev/null
 
-fork_json="$(drive9_ctx ctx fork "$FORK_CTX_NAME" --from owner --tidbcloud-public-key "$PUBLIC_KEY" --tidbcloud-private-key "$PRIVATE_KEY" --json)"
+set +e
+fork_json="$(drive9_ctx ctx fork "$FORK_CTX_NAME" --from owner --tidbcloud-public-key "$PUBLIC_KEY" --tidbcloud-private-key "$PRIVATE_KEY" --json 2>&1)"
+fork_rc=$?
+set -e
+
+if [ "$fork_rc" -ne 0 ] || echo "$fork_json" | grep -qi "not supported\|unsupported\|shared-pool"; then
+  SKIP_FORK=1
+  FAIL_FORK=0
+  echo -e "${YELLOW}SKIP${NC} fork not supported on this deployment, skipping fork checks"
+else
+  SKIP_FORK=0
+  FAIL_FORK=0
+fi
+if [ "$SKIP_FORK" -eq 0 ]; then
 fork_api_key="$(jq -r '.api_key // empty' <<<"$fork_json")"
 fork_tenant_id="$(jq -r '.tenant_id // empty' <<<"$fork_json")"
 fork_status="$(jq -r '.status // empty' <<<"$fork_json")"
@@ -355,6 +368,7 @@ fork_delete_code=$(curl -sS -o "$fork_delete_body" -w "%{http_code}" -X DELETE \
   "$BASE/v1/fork")
 check_eq "DELETE /v1/fork returns 202" "$fork_delete_code" "202"
 rm -f "$fork_delete_body" "$FORK_LOCAL"
+fi
 
 # ── [8] admin tenant list ───────────────────────────────────────────────────
 
--- a/e2e/native-smoke-test.sh
+++ b/e2e/native-smoke-test.sh
@@ -310,20 +310,7 @@ FORK_LOCAL="$(mktemp)"
 
 drive9_ctx ctx add --name owner --server "$BASE" --api-key "$API_KEY" --mode TiDBCloud >/dev/null
 
-set +e
-fork_json="$(drive9_ctx ctx fork "$FORK_CTX_NAME" --from owner --tidbcloud-public-key "$PUBLIC_KEY" --tidbcloud-private-key "$PRIVATE_KEY" --json 2>&1)"
-fork_rc=$?
-set -e
-
-if [ "$fork_rc" -ne 0 ] || echo "$fork_json" | grep -qi "not supported\|unsupported\|shared-pool"; then
-  SKIP_FORK=1
-  FAIL_FORK=0
-  echo -e "${YELLOW}SKIP${NC} fork not supported on this deployment, skipping fork checks"
-else
-  SKIP_FORK=0
-  FAIL_FORK=0
-fi
-if [ "$SKIP_FORK" -eq 0 ]; then
+fork_json="$(drive9_ctx ctx fork "$FORK_CTX_NAME" --from owner --tidbcloud-public-key "$PUBLIC_KEY" --tidbcloud-private-key "$PRIVATE_KEY" --json)"
 fork_api_key="$(jq -r '.api_key // empty' <<<"$fork_json")"
 fork_tenant_id="$(jq -r '.tenant_id // empty' <<<"$fork_json")"
 fork_status="$(jq -r '.status // empty' <<<"$fork_json")"
@@ -368,7 +355,6 @@ fork_delete_code=$(curl -sS -o "$fork_delete_body" -w "%{http_code}" -X DELETE \
   "$BASE/v1/fork")
 check_eq "DELETE /v1/fork returns 202" "$fork_delete_code" "202"
 rm -f "$fork_delete_body" "$FORK_LOCAL"
-fi
 
 # ── [8] admin tenant list ───────────────────────────────────────────────────
 
--- a/pkg/meta/tenant_pool_failed_cleanup.go
+++ b/pkg/meta/tenant_pool_failed_cleanup.go
@@ -1,0 +1,237 @@
+package meta
+
+import (
+	"context"
+	"database/sql"
+	"errors"
+	"fmt"
+	"strings"
+	"time"
+)
+
+// ListFailedNativeTenantCleanupCandidates lists failed native tenants owned by
+// an organization, excluding bindings already claimed from a pool.
+func (s *Store) ListFailedNativeTenantCleanupCandidates(ctx context.Context, organizationID string, updatedBefore time.Time, limit int) (out []TenantWithTiDBCloudOrgBinding, err error) {
+	start := time.Now()
+	defer observeMeta(ctx, "list_failed_native_tenant_cleanup_candidates", start, &err)
+	organizationID = strings.TrimSpace(organizationID)
+	if organizationID == "" {
+		return nil, fmt.Errorf("organization_id is required")
+	}
+	if limit <= 0 {
+		limit = 1
+	}
+	rows, err := s.db.QueryContext(ctx, `SELECT
+			t.id, t.status, t.kind, t.parent_tenant_id, t.storage_namespace_id,
+			t.db_host, t.db_port, t.db_user, t.db_password, t.db_name,
+			t.db_tls, t.provider, t.cluster_id, t.branch_id, t.claim_url, t.claim_expires_at, t.schema_version,
+			t.s3_encryption_mode, t.s3_kms_key_id, t.s3_bucket_key_enabled, t.created_at, t.updated_at,
+			b.tenant_id, b.organization_id, b.cluster_id, b.branch_id, b.pool_id, b.pool_status, b.used_at, b.created_at, b.updated_at
+		FROM tenant_tidbcloud_org_bindings b
+		JOIN tenants t ON t.id = b.tenant_id
+		WHERE b.organization_id = ?
+			AND (b.pool_status = ? OR b.pool_id = '')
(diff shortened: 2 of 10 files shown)


srstack:
## Review — clean up failed TiDB Cloud tenants

This implements the `2026-07-23-tenant-pool-failed-cleanup-design.md` design I reviewed earlier. Built in an isolated worktree — `go build ./...` passes and the meta + server failed-cleanup test suites are green. I traced both new source files against the design; it's a faithful, correct implementation. No High-severity issues.

### Highest-risk surfaces — confirmed safe

**Shared cleanup never deprovisions the physical cluster** ✅ (`cleanupFailedSharedTenant`, tenant_failed_cleanup.go). The path is `InvalidateAndWait` → `PurgeSharedTenant` (fs_id-scoped, gated on a full `connectionReady` check) → `AbortActiveUploadReservations` → `FinalizeSharedTenantDeleteMetadata` → best-effort key revoke. There is no `deprovisionTenantCluster` call anywhere on the shared path — a failed shared tenant cannot kill the shared cluster serving its co-tenants. A storage-namespace guard also rejects any candidate that owns a namespace (free members shouldn't).

**Native cleanup order is correct** ✅. CAS-claim → deprovision the dedicated cluster (only when cluster_id is non-empty, falling back to the binding's cluster_id) → best-effort revoke → `MarkTenantDeleted` (atomic tenant-deleted + binding removal). Any failure at an owned step restores `deleting → failed` via defer with a bounded 5s timeout and an updated retry timestamp.

**CAS is the cross-pod ownership boundary** ✅. Both `MarkFailedNativeTenantDeleting` / `MarkFailedSharedTenantDeleting` do `SELECT ... FOR UPDATE` then `UPDATE ... WHERE status=failed AND <full ownership predicate>` gated on `RowsAffected==1`. The UPDATE re-checks **every** eligibility predicate (provider, status=failed, updated_at<=cutoff, org, free/direct), so two concurrent workers cannot both win — the loser sees status!=failed and gets affected==0. `owned` gates the per-candidate work and the metrics recording.

### Design conformance — verified

- **Org-ownership proof + free/used exclusion**: native selects `pool_status=free OR pool_id=''` (free-pool or direct); shared selects free membership `OR` (no membership `AND` a placement in an org-owned db_pool). A `pool_status=used` binding / used membership is excluded by both selectors and both CAS re-checks. Org is always provable via binding, membership, or placement→db_pool.org_id.
- **15-minute cooldown**: `t.updated_at <= now-15m` in every selector and CAS. An unrelated tenant-row write only pushes eligibility later (fail-closed); it can never make a used/ineligible tenant eligible.
- **Batching & throttle**: native batch=1 (one Cloud deprovision/pass), shared batch=10; per-org per-Pod throttle (`active` flag + 1/min interval) skips the selector queries entirely on the hot claim path; DB CAS remains the cross-Pod boundary. Native and shared passes are independent — a list/candidate failure in one never blocks the other.
- **Best-effort**: triggered async from the claim path via `startServerWorker`; key revoke and reservation-abort are log-and-continue.

### Note (Low, non-blocking)

The shared direct-tenant UNION arm joins `db_pool → tenant_placements → fs_registry → tenants`. Row multiplication would require a tenant with placements on multiple db_pool rows, which the schema forbids (`fs_registry.tenant_id` UNIQUE, `tenant_placements` PK `fs_id` → one tenant maps to exactly one placement/pool). So it can't double-count in practice, and `LIMIT` bounds output regardless. Mentioning only for completeness.

LGTM. The implementation matches the design faithfully, the two scary surfaces (shared-cluster destruction, cross-pod double-claim) are both closed, and every path carries regression tests.

pkg/meta/tenant_pool_failed_cleanup.go
```diff
@@ -0,0 +1,237 @@
...
+				t.id, t.status, t.kind, t.parent_tenant_id, t.storage_namespace_id,
+				t.db_host, t.db_port, t.db_user, t.db_password, t.db_name,
+				t.db_tls, t.provider, t.cluster_id, t.branch_id, t.claim_url, t.claim_expires_at, t.schema_version,
+				t.s3_encryption_mode, t.s3_kms_key_id, t.s3_bucket_key_enabled, t.created_at, t.updated_at
+			FROM db_pool d
+			JOIN tenant_placements p ON p.db_id = d.db_id
+			JOIN fs_registry f ON f.fs_id = p.fs_id
+			JOIN tenants t ON t.id = f.tenant_id
+			LEFT JOIN tenant_pool_memberships m ON m.tenant_id = t.id
+			WHERE d.org_id = ? AND m.tenant_id IS NULL
```
mornyx, line 88: **Medium — failed tenants on wildcard (`*`) or foreign-org pools have no cleanup path (structural)**

This branch attributes a direct failed shared tenant to an org via `d.org_id = ?`. For a tenant placed on a manually registered wildcard pool (`org_id='*'`), no org's selector ever matches, so its placement slot leaks permanently. The deeper issue is that a shared tenant's provision-time organization is not persisted anywhere — 778 had to resolve org via `placement → db_pool` for the same reason, and on a wildcard pool that resolution yields `*`, which is also wrong for metrics.

Not blocking: managed pools are org-owned (and NULL-org pre-resolution rows become attributable once `UpdateManagedSharedDBPoolCloudResult` lands), and the PR deliberately scopes to organization-owned pools. But please track a follow-up to persist the resolved org at provision time (on the tenant row or the placement row) so cleanup, metrics, and future reconciliation don't depend on pool ownership. Until then, a one-line comment here stating the wildcard exclusion is intentional would save the next reader.

e2e/native-smoke-test.sh
```diff
@@ -310,7 +310,20 @@ FORK_LOCAL="$(mktemp)"
...
 
-fork_json="$(drive9_ctx ctx fork "$FORK_CTX_NAME" --from owner --tidbcloud-public-key "$PUBLIC_KEY" --tidbcloud-private-key "$PRIVATE_KEY" --json)"
+set +e
+fork_json="$(drive9_ctx ctx fork "$FORK_CTX_NAME" --from owner --tidbcloud-public-key "$PUBLIC_KEY" --tidbcloud-private-key "$PRIVATE_KEY" --json 2>&1)"
+fork_rc=$?
+set -e
+
+if [ "$fork_rc" -ne 0 ] || echo "$fork_json" | grep -qi "not supported\|unsupported\|shared-pool"; then
+  SKIP_FORK=1
+  FAIL_FORK=0
```
mornyx: **Nit — `FAIL_FORK` is set in both branches but never read**

Both the skip and non-skip paths assign `FAIL_FORK=0` and nothing consumes it — drop it, or wire it to whatever it was meant to gate. Also the fork-support detection greps stderr for `not supported|unsupported|shared-pool`, which will silently misclassify if the rejection text changes; checking for the structured 409/status from `/v1/fork` would be sturdier, but for an e2e script this is acceptable.

mornyx (approved):
Reviewed at `31732c6` (+2156/−3, 9 files; local build + vet pass, CI still running at review time).

This closes the failed-tenant residue gap from the earlier shared-control-plane reviews, and the design is sound where it matters:

- **Ownership CAS is correct on both providers.** `MarkFailedNativeTenantDeleting` / `MarkFailedSharedTenantDeleting` repeat the full boundary (org, provider, failed status, cooldown, free-membership/direct shape) in both the `FOR UPDATE` read and the conditional write — cross-pod ownership is enforced by the database, with the per-pod 1/min/org throttle as a best-effort rate limiter on top. The interlock with claims holds because claim requires `active` and cleanup requires `failed`, and both re-check under lock.
- **Shared selector's UNION ALL is safe** — branch 1 requires a free membership, branch 2 requires none, so they're mutually exclusive.
- **Failure handling is bounded**: restore `deleting → failed` with a 5s timeout re-arms the 15-minute cooldown instead of stranding tenants in `deleting`; revoke/abort steps are best-effort and don't block finalization; native and shared passes are independent.
- **Finalization is atomic**: `MarkTenantDeleted` (update + binding delete) and `FinalizeSharedTenantDeleteMetadata` (placement release + capacity decrement + membership + tenant + bindings) are single transactions — no partial-cleanup windows.
- Trigger placement is right: `defer` after successful org resolution in the claim path covers both admin create and provision fallthrough traffic without coupling cleanup latency into the claim response.

Two findings inline — one medium (wildcard-pool org attribution gap, worth a tracked follow-up, not blocking since the PR deliberately scopes to organization-owned pools) and nits. Also noting for the stress-test plan: failed **direct** shared tenants hold placement slots until this cleanup runs, and cleanup is driven by claim traffic — under a mass-failure scenario (cloud outage during a provision storm), expect capacity to stay held until claims resume; worth a monitoring line on `cleanup_failed_tidb_cloud_native_shared` operations.

Approving — please merge once CI goes green.

e2e/native-smoke-test.sh
```diff
@@ -310,7 +310,20 @@ FORK_LOCAL="$(mktemp)"
...
+  echo -e "${YELLOW}SKIP${NC} fork not supported on this deployment, skipping fork checks"
+else
+  SKIP_FORK=0
+  FAIL_FORK=0
+fi
```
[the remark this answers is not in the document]
srstack: Removed the unrelated native smoke-test change from this PR. Fork error handling is therefore unchanged by #779, and the failed-tenant cleanup PR no longer converts fork failures into skips.

pkg/meta/tenant_pool_failed_cleanup.go
srstack: Documented that direct-placement attribution intentionally requires an exact organization-owned pool. Wildcard pools remain excluded until the resolved organization is persisted per tenant; the selector does not infer tenant ownership from org_id='*'.

e2e/native-smoke-test.sh
srstack: Removed the unrelated native smoke-test change from this PR, including the unused FAIL_FORK assignments.

srstack:
Addressed review feedback:

- centralized shared DB connection readiness in sharedDBConnectionReady and reused it in both shared cleanup paths;
- documented intentional wildcard-pool exclusion in direct shared candidate attribution;
- removed the unrelated native smoke-test change from the PR.

Focused meta and server regressions pass.

PR #780, opened 2026-07-23
feat: authorize TiDB Cloud operations through IAM

srstack:
## Summary

- resolve TiDB Cloud API-key organization and role through the IAM API for native and shared control-plane operations
- allow only org:owner and project:owner, authorize persisted resources at organization scope, and remove cluster list/get calls used only for RBAC discovery
- require DRIVE9_TIDBCLOUD_IAM_API_URL for native/shared startup and require dedicated DRIVE9_TIDBCLOUD_NATIVE_SHARED_PUBLIC_KEY / PRIVATE_KEY in shared mode
- keep customer/default-native credentials for customer authorization and native Cloud operations, while routing every shared physical DB-pool operation through the configured shared credentials
- use shared physical root credentials directly, stop creating the shared system user, and name new clusters fs-shared-pool-{uuid}
- redact IAM access keys and response details from request logs and client-facing errors

## Behavior

- IAM response names are supported for both org-level and project-level API keys
- project owners are accepted at Drive9 organization scope; TiDB Cloud remains authoritative for project-specific native Cloud permissions
- admin tenant APIs treat tidb_cloud_native and tidb_cloud_native_shared uniformly
- shared virtual quota and manual shared DB authorization no longer depend on a physical cluster_id
- existing default native credential fallback remains unchanged

## Verification

- GOCACHE=/tmp/drive9-go-cache go test ./pkg/server -run 'TestQuota|TestSharedQuota|TestAdminTenant|TestTenantPool|TestProvisionTiDBCloudNative|TestFirstManagedOrganization|TestTiDBCloudRBACCache|TestManagedSharedDB' -count=1
- GOCACHE=/tmp/drive9-go-cache go test ./pkg/meta ./pkg/tenant ./pkg/tenant/schema ./cmd/drive9-server -count=1
- GOCACHE=/tmp/drive9-go-cache go build ./...
- git diff --check

All IAM fixtures use fictional access keys and IDs; no values from the supplied curl examples are retained.

## Summary by CodeRabbit

* **New Features**
  * Added TiDB Cloud IAM identity resolution and organization-scoped authorization for tenant administration and quota operations.
  * Added shared-provider credential validation and organization-driven shared-pool tenant provisioning, including persisting shared pool tenant limits.
* **Bug Fixes**
  * Tenant and quota flows now enforce exact organization matching (including trimming) and return correct HTTP 403/404 responses without leaking sensitive details.
  * Shared DB pool selection and allocation now follow stricter max-tenant rules, including drain-only behavior when max tenants is set to 0.
* **Documentation**
  * Expanded server usage docs with TiDB Cloud IAM and shared-key environment variables.

### DIFF
--- a/pkg/meta/db_pool.go
+++ b/pkg/meta/db_pool.go
@@ -12,8 +12,9 @@ import (
 	"github.com/google/uuid"
 )
 
-// MaxTiDBCloudSpendingLimit is the spendingLimit.monthly API maximum. In the
-// TiDB Cloud unit, 1,000,000 represents $10,000.
+// MaxTiDBCloudSpendingLimit is the tenant-facing quota API maximum. Physical
+// shared DB pools use a separately configured persisted target. In the TiDB
+// Cloud unit, 1,000,000 represents $10,000.
 const MaxTiDBCloudSpendingLimit = int64(1_000_000)
 
 // SharedDBRoleShared marks a db_pool row as a multi-tenant shared-schema
@@ -210,8 +211,8 @@ func (s *Store) CreateManagedSharedDBPool(ctx context.Context, in *SharedDB) (id
 	if in.MaxTenants <= 0 {
 		return 0, fmt.Errorf("managed max tenants must be positive")
 	}
-	if in.SpendingLimit == nil || *in.SpendingLimit != MaxTiDBCloudSpendingLimit {
-		return 0, fmt.Errorf("managed spending limit must equal TiDB Cloud maximum %d", MaxTiDBCloudSpendingLimit)
+	if in.SpendingLimit == nil || *in.SpendingLimit <= 0 || *in.SpendingLimit > int64(math.MaxInt32) {
+		return 0, fmt.Errorf("managed spending limit must be in (0,%d]", int64(math.MaxInt32))
 	}
 	var passwordCipher any
 	if len(in.PasswordCipher) != 0 {
--- a/pkg/meta/db_pool.go
+++ b/pkg/meta/db_pool.go
@@ -12,10 +12,6 @@ import (
 	"github.com/google/uuid"
 )
 
-// SharedDBOrgWildcard is the org_id sentinel matching any TiDB Cloud
-// organization that has no exact db_pool row.
-const SharedDBOrgWildcard = "*"
-
 // MaxTiDBCloudSpendingLimit is the spendingLimit.monthly API maximum. In the
 // TiDB Cloud unit, 1,000,000 represents $10,000.
 const MaxTiDBCloudSpendingLimit = int64(1_000_000)
@@ -99,12 +95,24 @@ func sharedDBUUID(raw string) (string, error) {
 	return parsed.String(), nil
 }
 
+func validateSharedDBOrganizationID(raw string) (string, error) {
+	organizationID := strings.TrimSpace(raw)
+	if organizationID == "" {
+		return "", fmt.Errorf("tidbcloud organization id is required")
+	}
+	if organizationID != raw || strings.ContainsAny(organizationID, "*%?") {
+		return "", fmt.Errorf("tidbcloud organization id must be an exact value")
+	}
+	return organizationID, nil
+}
+
 // SharedDB is one physical database registered in db_pool. PasswordCipher
 // holds the same encrypted envelope as tenants.db_password; the plaintext
 // never crosses this layer. TLSMode is the go-sql-driver tls DSN parameter
 // verbatim ("true", "skip-verify", a custom registered config name, or ""
 // for plaintext) so the runtime handle reopens the DB with exactly the TLS
-// mode it was registered with. MaxTenants of 0 means unlimited.
+// mode it was registered with. MaxTenants of 0 makes the pool drain-only:
+// existing tenants may leave, but no new tenant may enter.
 type SharedDB struct {
 	ID                      int64
 	UUID                    string
@@ -182,6 +190,10 @@ func (s *Store) CreateManagedSharedDBPool(ctx context.Context, in *SharedDB) (id
 	if in == nil {
 		return 0, fmt.Errorf("shared db pool is required")
 	}
+	organizationID, err := validateSharedDBOrganizationID(in.TiDBCloudOrganizationID)
+	if err != nil {
+		return 0, err
+	}
 	poolUUID, err := sharedDBUUID(in.UUID)
 	if err != nil {
 		return 0, err
@@ -200,10 +212,6 @@ func (s *Store) CreateManagedSharedDBPool(ctx context.Context, in *SharedDB) (id
 	}
 	if in.SpendingLimit == nil || *in.SpendingLimit != MaxTiDBCloudSpendingLimit {
 		return 0, fmt.Errorf("managed spending limit must equal TiDB Cloud maximum %d", MaxTiDBCloudSpendingLimit)
-	}
-	var organizationID any
-	if in.TiDBCloudOrganizationID != "" {
-		organizationID = in.TiDBCloudOrganizationID
 	}
 	var passwordCipher any
 	if len(in.PasswordCipher) != 0 {
@@ -466,28 +474,22 @@ func (s *Store) ListSharedDBsByStatus(ctx context.Context, status string, limit 
 }
 
 // FindSharedDBForAllocation selects the oldest eligible exact-organization
-// pool, preferring active over provisioning. Before organization resolution,
-// callers may instead supply the 32-byte provisioning fingerprint. Managed
-// rows use one fixed physical spending limit, so tenant virtual values do not
-// participate in allocation.
-func (s *Store) FindSharedDBForAllocation(ctx context.Context, organizationID string, provisioningKey []byte) (db *SharedDB, err error) {
+// pool, preferring active over provisioning. Managed rows use one fixed
+// physical spending limit, so tenant virtual values do not participate in
+// allocation.
+func (s *Store) FindSharedDBForAllocation(ctx context.Context, organizationID string) (db *SharedDB, err error) {
 	start := time.Now()
 	defer observeMeta(ctx, "find_shared_db_for_allocation", start, &err)
-	identityColumn := "org_id"
-	var identity any = organizationID
-	if organizationID == "" {
-		if len(provisioningKey) != 32 {
-			return nil, fmt.Errorf("organization id or 32-byte provisioning key is required")
-		}
-		identityColumn = "provisioning_key"
-		identity = provisioningKey
+	organizationID, err = validateSharedDBOrganizationID(organizationID)
+	if err != nil {
+		return nil, err
 	}
 	query := "SELECT " + sharedDBSelectColumns + " FROM db_pool d " +
-		"WHERE d." + identityColumn + " = ? AND d.`role` = ? " +
+		"WHERE d.org_id = ? AND d.`role` = ? " +
 		"AND d.status IN (?, ?) " +
-		"AND (d.max_tenants = 0 OR (d.soft_cap_reached = 0 AND d.tenant_count < d.max_tenants)) " +
+		"AND d.max_tenants > 0 AND d.soft_cap_reached = 0 AND d.tenant_count < d.max_tenants " +
 		"ORDER BY CASE d.status WHEN 'active' THEN 0 ELSE 1 END, d.db_id LIMIT 1"
-	db, err = scanSharedDBRow(s.db.QueryRowContext(ctx, query, identity, SharedDBRoleShared,
+	db, err = scanSharedDBRow(s.db.QueryRowContext(ctx, query, organizationID, SharedDBRoleShared,
 		SharedDBStatusActive, SharedDBStatusProvisioning))
 	if err != nil {
 		if errors.Is(err, sql.ErrNoRows) {
@@ -534,6 +536,10 @@ func (s *Store) RegisterSharedDB(ctx context.Context, in *SharedDB) (dbID int64,
 	if in == nil {
 		return 0, fmt.Errorf("shared db is required")
 	}
+	organizationID, err := validateSharedDBOrganizationID(in.TiDBCloudOrganizationID)
+	if err != nil {
+		return 0, err
+	}
 	poolUUID, err := sharedDBUUID(in.UUID)
 	if err != nil {
 		return 0, err
@@ -576,7 +582,7 @@ func (s *Store) RegisterSharedDB(ctx context.Context, in *SharedDB) (dbID int64,
 			"ON DUPLICATE KEY UPDATE db_port = VALUES(db_port), db_user = VALUES(db_user), "+
 			"db_password = VALUES(db_password), db_tls = VALUES(db_tls), "+
 			"max_tenants = VALUES(max_tenants), status = VALUES(status)",
-		poolUUID, in.TiDBCloudOrganizationID, role, in.Host, in.Port, in.User, in.PasswordCipher, in.Name,
+		poolUUID, organizationID, role, in.Host, in.Port, in.User, in.PasswordCipher, in.Name,
 		in.TLSMode, in.MaxTenants, status); err != nil {
 		return 0, fmt.Errorf("upsert db_pool row: %w", err)
 	}
@@ -584,7 +590,7 @@ func (s *Store) RegisterSharedDB(ctx context.Context, in *SharedDB) (dbID int64,
 	// auto-increment id via LastInsertId, so re-fetch by endpoint.
 	err = s.db.QueryRowContext(ctx,
 		`SELECT db_id FROM db_pool WHERE org_id = ? AND db_host = ? AND db_name = ?`,
-		in.TiDBCloudOrganizationID, in.Host, in.Name).Scan(&dbID)
+		organizationID, in.Host, in.Name).Scan(&dbID)
 	if err != nil {
 		return 0, fmt.Errorf("resolve db_id after upsert: %w", err)
 	}
@@ -625,39 +631,26 @@ func (s *Store) GetSharedDBForTenant(ctx context.Context, tenantID string) (*Sha
 	return db, nil
 }
 
-// FindSharedDBForOrg returns the active shared database serving orgID: an
-// exact org_id match wins over the '*' wildcard row. An empty orgID matches
-// only the wildcard row. Returns ErrNotFound when neither exists.
+// FindSharedDBForOrg returns the oldest open active shared database serving
+// exactly orgID. Shared DB pools never match across organizations.
 func (s *Store) FindSharedDBForOrg(ctx context.Context, orgID string) (db *SharedDB, err error) {
 	start := time.Now()
 	defer observeMeta(ctx, "find_shared_db_for_org", start, &err)
-	if orgID != "" && orgID != SharedDBOrgWildcard {
-		db, err = s.findActiveSharedDB(ctx, orgID)
-		if err == nil {
-			return db, nil
-		}
-		if !errors.Is(err, ErrNotFound) {
-			return nil, err
-		}
-	}
-	db, err = s.findActiveSharedDB(ctx, SharedDBOrgWildcard)
+	orgID, err = validateSharedDBOrganizationID(orgID)
 	if err != nil {
-		if errors.Is(err, ErrNotFound) {
-			return nil, ErrNotFound
-		}
 		return nil, err
 	}
-	return db, nil
+	return s.findActiveSharedDB(ctx, orgID)
 }
 
 func (s *Store) findActiveSharedDB(ctx context.Context, orgID string) (*SharedDB, error) {
 	// Capacity is enforced at selection time: normal allocation only sees open
-	// soft-cap pools (0 means unlimited); the transactional reserve remains the
-	// final guard against races.
+	// positive-capacity pools; the transactional reserve remains the final
+	// guard against races.
 	db, err := scanSharedDBRow(s.db.QueryRowContext(ctx,
 		"SELECT "+sharedDBSelectColumns+" FROM db_pool "+
 			"WHERE org_id = ? AND `role` = ? AND status = ? "+
-			"AND (max_tenants = 0 OR (soft_cap_reached = 0 AND tenant_count < max_tenants)) ORDER BY db_id LIMIT 1",
+			"AND max_tenants > 0 AND soft_cap_reached = 0 AND tenant_count < max_tenants ORDER BY db_id LIMIT 1",
 		orgID, SharedDBRoleShared, sharedDBStatusActive))
 	if err != nil {
 		if errors.Is(err, sql.ErrNoRows) {
@@ -992,7 +985,7 @@ func (s *Store) completeSharedTenantProvision(ctx context.Context, tenantID, pro
 			if dbPoolStatus != SharedDBStatusActive && (dbPoolStatus != SharedDBStatusProvisioning || (membership == nil && (!clusterID.Valid || clusterID.String == ""))) {
 				return fmt.Errorf("db pool %d is not physically ready: status=%s cluster=%q membership=%t: %w", p.DbID, dbPoolStatus, clusterID.String, membership != nil, ErrSharedDBCapacityExhausted)
 			}
-			if maxTenants > 0 && (softCapReached || tenantCount >= maxTenants) {
+			if maxTenants <= 0 || softCapReached || tenantCount >= maxTenants {
 				return fmt.Errorf("db pool %d soft capacity exhausted: count=%d max=%d latch=%t: %w", p.DbID, tenantCount, maxTenants, softCapReached, ErrSharedDBCapacityExhausted)
 			}
 		case SharedDBCapacityEmergency:
@@ -1014,7 +1007,7 @@ func (s *Store) completeSharedTenantProvision(ctx context.Context, tenantID, pro
 					ELSE soft_cap_reached END,
 					tenant_count = tenant_count + 1
 				WHERE db_id = ? AND (%s)
-					AND (max_tenants = 0 OR (soft_cap_reached = 0 AND tenant_count + 1 <= max_tenants))`
+					AND max_tenants > 0 AND soft_cap_reached = 0 AND tenant_count + 1 <= max_tenants`
 			if membership != nil {
 				normalQuery = fmt.Sprintf(normalQuery, "status IN (?, ?)")
 				res, err = tx.ExecContext(ctx, normalQuery, p.DbID, SharedDBStatusActive, SharedDBStatusProvisioning)
@@ -1284,7 +1277,9 @@ func releaseTenantPlacementAndDecrCountTx(ctx context.Context, tx *sql.Tx, fsID,
 		newTenantCount = 0
 	}
 	nextSoftCapReached := softCapReached
-	if maxTenants > 0 && newTenantCount <= reopenThreshold {
+	if maxTenants == 0 {
+		nextSoftCapReached = true
+	} else if newTenantCount <= reopenThreshold {
 		nextSoftCapReached = false
 	}
 	_, err = tx.ExecContext(ctx, `UPDATE db_pool
--- a/pkg/server/provision_test.go
+++ b/pkg/server/provision_test.go
@@ -251,10 +251,8 @@ func TestProvisionTiDBCloudNativeSharedPlansManagedPoolAndReturnsProvisioning(t 
 		t.Fatalf("shared physical credential = %+v, want configured shared credential", got)
 	}
 	iamCredentials := prov.iamCredentialsSnapshot()
-	if len(iamCredentials) != 2 ||
-		iamCredentials[0].PublicKey != "public" || iamCredentials[0].PrivateKey != "private" ||
-		iamCredentials[1].PublicKey != "shared-public" || iamCredentials[1].PrivateKey != "shared-private" {
-		t.Fatalf("IAM credentials = %+v, want customer authorization followed by shared physical authorization", iamCredentials)
+	if len(iamCredentials) != 1 || iamCredentials[0].PublicKey != "public" || iamCredentials[0].PrivateKey != "private" {
+		t.Fatalf("IAM credentials = %+v, want customer authorization only", iamCredentials)
 	}
 }
 
--- a/pkg/server/provision_test.go
+++ b/pkg/server/provision_test.go
@@ -180,8 +180,9 @@ func TestProvisionTiDBCloudNativeSharedPlansManagedPoolAndReturnsProvisioning(t 
 		provider: tenant.ProviderTiDBCloudNative, cloudProvider: "aws", region: "us-east-1",
 		defaultPublicKey: "public", defaultPrivateKey: "private",
 		defaultSharedPublicKey: "shared-public", defaultSharedPrivateKey: "shared-private",
+		identityOrg: "customer-org",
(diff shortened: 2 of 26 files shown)


pkg/server/shared_db_pool.go
```diff
@@ -516,11 +532,15 @@ func (s *Server) ensureManagedSharedDBPhysicalLocked(ctx context.Context, poolIn
...
-func (s *Server) ensureManagedSharedDBPhysical(ctx context.Context, dbID int64, cred tenant.CredentialProvisionRequest) (*meta.SharedDB, error) {
+func (s *Server) ensureManagedSharedDBPhysical(ctx context.Context, dbID int64) (*meta.SharedDB, error) {
 	poolInfo, err := s.meta.GetSharedDB(ctx, dbID)
 	if err != nil {
 		return nil, err
 	}
+	cred, err := s.sharedDBCloudCredentials(ctx, poolInfo.TiDBCloudOrganizationID)
+	if err != nil {
+		return nil, err
+	}
```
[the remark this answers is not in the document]
srstack: Fixed in 0ca9d9f3. Credentials are now re-derived per continuation/ensure/batch attempt from the current persisted organization, with an allocation-lock restart if the organization changed. The metadata waiter also resolves credentials from the latest pool row.

pkg/server/tidbcloud_iam.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+	identity, err := resolver.ResolveAPIKeyIdentity(ctx, cred)
+	if err != nil {
+		return nil, err
+	}
+	if identity == nil || strings.TrimSpace(identity.OrganizationID) == "" {
+		return nil, fmt.Errorf("TiDB Cloud IAM identity is missing organization")
+	}
+	if identity.Role != tenant.TiDBCloudRoleOrgOwner && identity.Role != tenant.TiDBCloudRoleProjectOwner {
+		return nil, fmt.Errorf("%w: role %q; org:owner or project:owner is required", tenant.ErrTiDBCloudRoleInsufficient, identity.Role)
```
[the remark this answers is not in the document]
srstack: This is intentional for this change. Drive9 product policy treats both org-owner and project-owner IAM roles as organization-scoped operators after IAM resolves the key organization and role. We are deliberately not persisting project ownership or restoring cluster-list authorization in this PR; the policy is now documented next to the role gate.

pkg/server/shared_db_pool.go
```diff
@@ -370,9 +366,17 @@ func (s *Server) resumeManagedSharedDBPoolsWithCtx(ctx context.Context) {
...
+	cred, err := s.sharedDBCloudCredentials(ctx, poolInfo.TiDBCloudOrganizationID)
+	if err != nil {
+		return err
+	}
 	for metadataAttempt := 0; metadataAttempt < 2; metadataAttempt++ {
-		err := s.continueManagedSharedDBPoolOnce(ctx, dbID, cred)
+		err := s.continueManagedSharedDBPoolOnceWithCredentials(ctx, dbID, cred)
 		if !errors.Is(err, errSharedDBConnectionMetadataNotReady) {
 			return err
 		}
```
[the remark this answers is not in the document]
srstack: Fixed in 0ca9d9f3. Continuation attempts now resolve the shared credential against the pool row loaded for that attempt, restart when the organization changes while waiting for the allocation lock, and resolve again from the latest persisted organization before invoking the metadata waiter. This removes the stale wildcard/unresolved-org credential path.

pkg/server/tidbcloud_iam.go
```diff
@@ -0,0 +1,91 @@
...
+	}
+	s.tidbCloudRBACCache.rememberIdentity(cred, *identity)
+	return identity, nil
+}
+
+func tiDBCloudOrganizationMatches(identityOrganizationID, resourceOrganizationID string) bool {
+	identityOrganizationID = strings.TrimSpace(identityOrganizationID)
+	resourceOrganizationID = strings.TrimSpace(resourceOrganizationID)
+	return identityOrganizationID != "" && (resourceOrganizationID == identityOrganizationID || resourceOrganizationID == "*")
```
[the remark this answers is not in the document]
srstack: Fixed in 0ca9d9f3. Wildcard DB pools remain available to allocation, but they are no longer accepted as an authorization match and are excluded from IAM organization tenant listings. Admin/quota operations therefore reject wildcard-hosted tenants unless an explicit matching organization is persisted.

pkg/server/provision_test.go
```diff
@@ -970,9 +1082,10 @@ func (f *fakeProvisioner) ListManagedClusters(_ context.Context, _ tenant.Creden
 	return &tenant.ManagedClusterListResult{Clusters: append([]tenant.CloudClusterInfo(nil), f.managedClusters...)}, nil
 }
 
-func (f *fakeProvisioner) BatchProvisionSharedDBPoolsWithCredentials(ctx context.Context, requests []tenant.SharedDBPoolCreateRequest, _ tenant.CredentialProvisionRequest) ([]*tenant.SharedDBPoolInfo, error) {
+func (f *fakeProvisioner) BatchProvisionSharedDBPoolsWithCredentials(ctx context.Context, requests []tenant.SharedDBPoolCreateRequest, cred tenant.CredentialProvisionRequest) ([]*tenant.SharedDBPoolInfo, error) {
 	f.sharedPoolBatchCalls.Add(1)
 	f.sharedPoolBatchMembers.Add(int32(len(requests)))
+	f.lastSharedCredentialReq = cred
```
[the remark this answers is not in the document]
srstack: Fixed in 0ca9d9f3. The fake provisioner records the credential before publishing the shared-pool batch-call counter, establishing the intended happens-before edge. The focused race test passes.

srstack:
## Review — authorize TiDB Cloud operations through IAM

Security-focused pass on the authorization surface (this replaces the old cluster-list RBAC discovery with an IAM identity lookup). Built in an isolated worktree — `go build ./...` and the PR's meta/tenant/server test suites pass. I traced the full authz path (HTTP handlers → `resolveTiDBCloudIdentity` → IAM resolver → org matching → RBAC cache) independently and with a second reviewer. **No authorization-bypass or privilege-escalation defect.** One Low secrets-in-logs gap that contradicts the redaction claim.

### Authorization — verified fail-closed, no cross-org bypass

- **Fail-closed on every error path** ✅ `ResolveAPIKeyIdentity` (provisioner.go) and `resolveTiDBCloudIdentity` (tidbcloud_iam.go) return `(nil, err)` on: missing creds, transport error, non-2xx (upstream body discarded), JSON decode error / empty body, accessKey≠request mismatch, malformed org name, and insufficient role. No path returns an authorized identity on failure. Resolver-interface-absent also denies.
- **Role allow-list is exact-match, case-sensitive** ✅ Compared against lowercase constants `org:owner` / `project:owner` in *both* the provisioner (source) and the server (defense in depth). Any unexpected case/empty/unknown role fails **closed** (deny), never open.
- **Org identity is bound to the resource, not "any owner"** ✅ Org comes solely from the IAM `name` (`orgs/<id>/...`, requires `parts[0]=="orgs"` + non-empty id). Every resource op re-checks `tiDBCloudOrganizationMatches(identity.Org, resourceOrg)` (empty-string guard included): quota get/set → tenant binding org; shared quota → `physical.TiDBCloudOrganizationID`; admin get/quota-set/delete (`authorizedAdminTenant`) → native binding org or shared physical org, 403 on mismatch; pool ops → `sharedDBCloudCredentials` org bind. An owner of org A gets 403/permission-denied on org B's resources.
- **`handleAdminTenantList` safe by construction** ✅ role-check only, then lists via `ListTenantsByTiDBCloudOrganization(identity.OrganizationID, …)` — the identity org *is* the query filter, so it can only ever return the caller's own org.
- **RBAC cache is sound** ✅ keyed by `sha256(publicKey \x00 privateKey)` (both halves required, secrets hashed not stored raw); only successful role-validated identities cached; denials never cached (can't poison to allow); 1h TTL bounds the post-revocation stale-allow window as intended.
- **Startup hardening** ✅ `DRIVE9_TIDBCLOUD_IAM_API_URL` now required + validated https; shared mode requires dedicated shared keys — authz can't silently degrade from misconfig.

Note (by design, not a defect): within a single org, a `project:owner` of project A can operate on project B's persisted resources — the local ownership model is org-scoped and has no ProjectID; for native Cloud ops TiDB Cloud stays authoritative via the caller's own creds. The PR states this explicitly. Also, the quota rewrite is a net improvement: the old RBAC-miss path could fall through, and it required a non-empty `cluster_id` this PR's whole point is to drop — every quota get/set now routes through a real IAM org match.

### Finding — Low (CONFIRMED), non-blocking

**`pkg/tenant/tidbcloudnative/provisioner.go:2415` and `:2447`** — on a transport error, `p.client.Do` returns a `*url.Error` whose `Error()` embeds the full request URL. Go's `http.Client` strips only the userinfo *password*, not the path, and the IAM endpoint is `.../v1beta1/apikeys/<publicKey>`. So `zap.Error(err)` logs the access key verbatim on any IAM/cloud network failure — even though the sibling `path` field is redacted to `/v1beta1/apikeys/***`. This contradicts the PR's "redact IAM access keys from request logs" claim.

- Scope: server logs only; leaks the **public/access-key half** (private key is never in the URL and never logged); client-facing response stays generic.
- Suggested fix: log `requestPath(uri)` instead of the raw error, and return a sanitized error up the stack (e.g. `fmt.Errorf("tidbcloud request failed: %w", redactedErr)` where the URL is stripped), rather than `zap.Error(err)` on the raw `*url.Error`.

Otherwise LGTM — the core IAM authorization is correct and fails safe. Recommend addressing the log-redaction gap since it directly contradicts a stated PR goal, but it's not a bypass.

srstack:
Addressed the latest review in 671b791a.

- IAM transport errors are now converted to redacted url.Error values before both logging and propagation. The public/access key is removed from the URL while the underlying transport cause remains available through errors.Is.
- IAM success and failure access logs retain the request trace_id; regression tests cover both trace propagation and access-key redaction.
- Native binding and shared physical DB authorization now distinguish meta.ErrNotFound from backend failures. Genuine misses retain the existing 404 behavior, while cancellation, connection, and SQL errors propagate as backend errors; admin shared lookup failures map to 5xx.

Validation: full pkg/tenant/tidbcloudnative tests, focused quota/admin/IAM server tests, go build ./..., git diff --check, and make lint all pass.

pkg/server/tidbcloud_iam.go
```diff
@@ -0,0 +1,98 @@
...
+		return nil, err
+	}
+	if identity == nil || strings.TrimSpace(identity.OrganizationID) == "" {
+		return nil, fmt.Errorf("TiDB Cloud IAM identity is missing organization")
+	}
+	// Drive9's TiDB Cloud policy intentionally treats both owner roles as
+	// organization-scoped operators. The IAM lookup replaces the old
+	// cluster-list authorization path; project resource membership is not part
+	// of the local tenant ownership model.
+	if identity.Role != tenant.TiDBCloudRoleOrgOwner && identity.Role != tenant.TiDBCloudRoleProjectOwner {
```
mornyx, line 35: **Medium — project:owner is silently widened to org scope for metadata authorization (needs product sign-off)**

This gate admits `project:owner` keys, and every downstream check (`authorizeTiDBCloudOrganization`, `authorizedAdminTenant`, `ListTenantsByTiDBCloudOrganization`) only compares `identity.OrganizationID` — IAM identity carries no project dimension. The previous per-cluster list checks were effectively project-scoped (a project key only *saw* its own clusters), so this PR widens a project:owner key to:

- list **all** tenants in the org (`handleAdminTenantList`),
- admin-get/delete tenants whose clusters live in **other projects** of the same org.

The cloud-side operation on another project's cluster will still be rejected by TiDB Cloud, which means the widened path can drive9-delete a tenant it cannot deprovision — orphan cluster plus deleted metadata.

The PR body documents this as intentional (drive9 ownership is org-scoped), and for managed single-org deployments it changes nothing. But TiDB Cloud projects are an isolation boundary customers may rely on. Please get explicit product sign-off, and consider either rejecting project:owner for destructive admin operations or documenting the widening in the API contract.

pkg/server/server.go
```diff
@@ -5256,6 +5254,12 @@ func (s *Server) provisionTenant(ctx context.Context, opts provisionTenantOption
 			opts.CredentialProvisioner = defaultReq
 		}
 	}
+	if tenant.UsesTiDBCloudNativeCredentials(provider) {
```
mornyx, line 5259: **Medium — no same-org guard when the customer key's org ≠ the shared key's org (BYOC footgun)**

This new upfront identity check validates the *customer* credential, but nothing compares it against the *shared* credential's org. In a BYOC deployment where a customer presents keys from org Y while `DRIVE9_TIDBCLOUD_NATIVE_SHARED_*` belongs to org X:

- direct provision: `allocateManagedSharedDB` persists the pool row with org Y, then `sharedDBCloudCredentials` fails the org match — the tenant 502s, but the org-Y pool row stays `provisioning` and gets retried by the 1-minute reconciler forever (every attempt fails the same check),
- prewarm: the batch create runs under the shared cred (org X) and `UpdateManagedSharedDBPoolCloudResult` rewrites the row to org X while memberships were seeded with org Y — subsequent quota/admin org-match checks for those tenants fail inconsistently.

Single-org managed deployments never hit this (X == Y by construction). Please add a fail-fast guard right here — e.g. resolve both identities and reject with a clear 4xx (`shared mode requires the TiDB Cloud API key's organization to match the shared credential organization`) — so mixed-org requests die at the door instead of leaving zombie pool rows.

pkg/server/quota.go
```diff
@@ -290,27 +286,15 @@ func (s *Server) handleQuotaSet(w http.ResponseWriter, r *http.Request) {
...
-		return physical, nil
-	}
-	getter, ok := s.provisioner.(tenant.QuotaGetter)
-	if !ok {
-		return nil, errQuotaSettingNotEnabled
+	if err != nil {
+		if errors.Is(err, meta.ErrNotFound) {
+			return nil, tenant.ErrQuotaBackendNotFound
+		}
+		return nil, fmt.Errorf("get shared DB for tenant: %w", err)
```
mornyx, line 298: **Note — wildcard pools now always fail quota authorization**

For a shared tenant on a `*` wildcard pool, `authorizeTiDBCloudOrganization` can never match (`*` != any IAM org), so quota GET/SET returns 403 unconditionally. Manual pools registered with a concrete org id are unaffected. This is fine as long as wildcard remains a dev/local escape hatch — please document it in the shared-pool configuration docs so nobody registers a wildcard pool in a deployment where quota APIs matter.

mornyx (approved):
Reviewed at `671b791` (3 commits, +1286/−738 across 21 files; CI green, local build passes).

This is a good architectural cleanup. The IAM identity lookup replaces the fragile cluster-list RBAC discovery — as a bonus it eliminates the >100-clusters pagination gap I flagged on #774, since authorization no longer depends on listing. Verified the key paths:

- **Identity resolution** (`ResolveAPIKeyIdentity`): digest-auth GET on `/v1beta1/apikeys/{key}`, response `accessKey` echoed back and validated against the request, org parsed from the IAM name for both org-level and project-level keys, role gate at resolve time (cached identities are pre-gated). ✅
- **Shared pool operations now use server-configured shared credentials** with an org-match check against the pool — this fixes the default-creds-only continuation/resume limitation I flagged on #774 (resume no longer depends on customer/default native creds). ✅
- **Redaction**: IAM apikeys paths are masked in logs, transport `url.Error`s are rewritten path-only, IAM error bodies are discarded, client-facing detail for role failures is generic. No key material leaks found. ✅
- **System user removal**: acceptable — dedicated native tenants already run on root credentials, so this is parity, not a downgrade. ✅
- RBAC cache simplification to identity-only is correct; `forgetTiDBCloudRBACList` removal is consistent (identity no longer changes on tenant mutation). ✅

Two medium findings inline (project-owner authorization widening — needs explicit product sign-off; missing same-org guard for BYOC shared mode) plus deployment notes below. None block merge for the single-org managed deployment, but both should be tracked.

**Deployment notes (breaking for upgrades):**
1. `DRIVE9_TIDBCLOUD_IAM_API_URL` is now REQUIRED for both `tidb_cloud_native` and `tidb_cloud_native_shared` — servers without it exit at startup.
2. `DRIVE9_TIDBCLOUD_NATIVE_SHARED_PUBLIC_KEY` / `PRIVATE_KEY` are required in shared mode.
3. Provisioner construction failure now always exits for native-family providers (the old warn-and-continue path without `DRIVE9_TIDBCLOUD_NATIVE_API_URL` is gone).
4. Identity cache TTL is 1h — TiDB Cloud role revocations/demotions take up to 1h to take effect locally (same semantics as the previous cluster cache; acceptable, but worth knowing for incident response).
5. Shared tenants on wildcard (`*`) pools now always get 403 on quota GET/SET (org can never match). Fine if wildcard stays a dev/local escape hatch; please document it.

Approving.

pkg/server/quota.go
srstack: Confirmed intentional. Wildcard DB pools remain allocation-only/dev-local escape hatches and are not valid IAM authorization ownership. Deployments that require admin or quota APIs must register and persist a concrete TiDB Cloud organization ID. We will not restore wildcard-as-any-org authorization.

pkg/server/tidbcloud_iam.go
srstack: Product sign-off is confirmed: Drive9 intentionally treats both org:owner and project:owner as organization-scoped operators. The IAM migration deliberately removes project/cluster-list ownership checks, and this policy is documented next to the role gate. We are not adding ProjectID persistence or restricting destructive operations by project in this PR.

pkg/server/server.go
srstack: Fixed in 360c1283. Direct shared provisioning now resolves the customer IAM identity and validates the configured shared credential against that organization before generating or persisting the tenant. Shared tenant-pool creation performs the same validation before CreateTenantPool. Organization mismatch returns 403 with a stable redacted message, and tests verify that no tenant, tenant-pool, membership, or db_pool metadata is left behind.

srstack: Clarification after product review: customer organization and the configured shared-credential organization are allowed to differ by design. I initially applied a same-organization guard from this review, then reverted it in d3ad4d15. Shared credentials are now used only for physical Cloud operations; the db_pool organization remains the customer IAM organization, and a physical Cloud response cannot overwrite it. The implementation only rejects a managed pool when its customer organization is missing.

srstack: Additional cleanup in 5b0da43f: shared credentials are now validated once during server startup through an independent ValidateSharedCredentials path. Runtime shared physical operations only read the configured credential; they do not call customer resolveTiDBCloudIdentity or the customer RBAC cache. The customer IAM organization remains the sole logical ownership value.

srstack:
## Re-review — updated through `5b0da43f`

Re-reviewed the 5 commits pushed since my first pass. Rebuilt the branch in an isolated worktree — `go build ./...` passes and the IAM/shared/admin test suites are green. The earlier Low finding is fixed, and the shared-credential/org model is now cleanly separated. No new issues.

### My Low finding (IAM access key in logs) — fixed ✅ (`671b791a`)

`doDigestAuthRequest` now wraps transport errors with `redactRequestError`, which `errors.As`-extracts the `*url.Error` and rebuilds it with `requestPath(uri)` in place of the raw URL (`/v1beta1/apikeys/***`). Applied at both `client.Do` failure points, so the access key no longer reaches the logs on an IAM/cloud network failure. Same commit also tightened `GetSharedDBForTenant`/`GetTenantTiDBCloudOrgBinding` to distinguish `ErrNotFound` (→ 404/backend-not-found) from real DB errors (previously collapsed to a blanket not-found).

### Shared credential / org model — clarified and hardened

- **`360c1283` → `d3ad4d15`** (introduce-then-revert): the final state decouples a shared pool's **logical ownership org (the customer org, stored on `db_pool.TiDBCloudOrganizationID`)** from the **cloud credential org**. `UpdateManagedSharedDBPoolCloudResult` now requires a non-empty customer org and no longer backfills it from the cloud identity — so the operator credential's org can't pollute a pool's logical ownership.
- **`5b0da43f`**: org validation moved to startup via `ValidateSharedCredentials`. I verified this does **not** weaken authorization — it calls `ResolveAPIKeyIdentity`, which still performs the full role allow-list + `accessKey`-binding + org parse. It just changes "resolve on every op" to "validate the configured credential once at boot" (shared mode fails fast if misconfigured). Runtime `sharedDBCloudCredentials()` is now parameterless and only returns the configured credential; org matching for customer requests still happens through the customer's own IAM identity on the provision/admin paths.

The core IAM authorization remains fail-closed with no cross-org bypass (as in my first review): role allow-list is exact-match/case-sensitive (deny on mismatch), every error path returns unauthorized, org identity is matched against the resource org, and the RBAC cache is keyed by the full hashed credential. LGTM.

pkg/meta/db_pool.go
```diff
@@ -201,10 +213,6 @@ func (s *Store) CreateManagedSharedDBPool(ctx context.Context, in *SharedDB) (id
 	if in.SpendingLimit == nil || *in.SpendingLimit != MaxTiDBCloudSpendingLimit {
 		return 0, fmt.Errorf("managed spending limit must equal TiDB Cloud maximum %d", MaxTiDBCloudSpendingLimit)
```
mornyx: **Question — tenant virtual spending limits are now inert**

Fixing the pool spending cap at the cloud maximum simplifies the model, but note the downstream effect: `ErrSharedDBSpendingLimitExceeded` is gone and no path (provision, claim, `UpdateSharedTenantQuotaConfig`) enforces per-tenant virtual limits anymore — the value is still validated, persisted, and returned by quota GET, so it *looks* enforced to an API consumer while doing nothing.

If "informational/display-only" is the intended end state, please say so in the quota API docs (or in the field comment); otherwise this is a silent behavior regression relative to the #774 design. Either way, capacity is now count-only, which is the right call for the stress test — I'll update the dev test plan to drop the spending-trigger case.

pkg/meta/db_pool.go
```diff
@@ -1284,7 +1277,9 @@ func releaseTenantPlacementAndDecrCountTx(ctx context.Context, tx *sql.Tx, fsID,
 		newTenantCount = 0
 	}
 	nextSoftCapReached := softCapReached
-	if maxTenants > 0 && newTenantCount <= reopenThreshold {
+	if maxTenants == 0 {
```
mornyx, line 1281: **Deployment note — `max_tenants = 0` semantic flip**

0 changed from "unlimited" to "drain-only" (allocation rejects it at selection and at reserve; release keeps the latch set). Existing manually registered pools created under the old semantics (e.g. 753-era `DRIVE9_SHARED_POOL_DSN` registrations without an explicit cap) will silently stop accepting new tenants after this upgrade. Since we don't run data migrations on existing deployments, the runbook needs a manual `UPDATE db_pool SET max_tenants = ... WHERE max_tenants = 0` step (or a loud release note) for any env that already has `db_pool` rows.

mornyx (approved):
Re-reviewed the 6 new commits (`360c128`..`e58dd65`; CI green).

**My mixed-org finding is resolved — by a better design than the guard I suggested.** Instead of rejecting cross-org requests, pool ownership is now fully decoupled from cloud credentials: `db_pool.org_id` is always the customer's IAM org (`logicalOrganizationID` is required and used at every persistence point — direct ensure, continuation, and batch), while the physical cluster lives wherever the server-owned shared credential lives. I verified no path rewrites the logical org from the cloud response anymore. Combined with exact-org validation at all entry points (`validateSharedDBOrganizationID`, rejecting empty/`*`/`%`/`?`) and startup-time `ValidateSharedCredentials`, the zombie-pool and inconsistent-attribution scenarios are gone. ✅

**My wildcard note is resolved** by eliminating the wildcard concept entirely (`SharedDBOrgWildcard` removed; `DRIVE9_SHARED_POOL_ORG='*'` now dies at startup). ✅

**Still open from last round:** the project:owner → org-scope widening for metadata authorization (inline on `tidbcloud_iam.go`) — that one still needs explicit product sign-off; nothing in this round changes it.

**New behavior changes in this round that need to reach the deploy runbook / test plan:**
1. Managed pool spending limit is now hard-fixed at the TiDB Cloud maximum (`MaxTiDBCloudSpendingLimit` = $10k/月/pool), `DRIVE9_TIDBCLOUD_NATIVE_DB_POOL_DEFAULT_SPENDING_LIMIT` is gone, and — more importantly — the per-tenant virtual spending limit no longer has any enforcement anywhere (`ErrSharedDBSpendingLimitExceeded` deleted; no headroom accounting in provision/claim/quota-set). See inline: is "informational only" the intended end state for tenant virtual limits?
2. `max_tenants = 0` flipped semantics from "unlimited" to "drain-only" — existing manually registered pools with 0 (the 753-era default) will silently stop accepting new tenants after upgrade. Audit `db_pool` rows before deploy.
3. Manual pool registration (`DRIVE9_SHARED_POOL_DSN`) now requires an exact org and takes `MAX_TENANTS` explicitly.

Approving — the redesign is coherent and the test coverage follows each move.

pkg/meta/db_pool.go
srstack: Confirmed intentional. For tidb_cloud_native_shared, the tenant spending-limit field is virtual compatibility metadata: it remains validated, persisted, and returned by quota APIs, but it does not participate in placement, capacity, or physical TiDB Cloud enforcement. Physical shared clusters always use the fixed TiDB Cloud maximum. I added source-level request/response field comments to make that contract explicit.

pkg/meta/db_pool.go
srstack: No migration is needed for this branch: this shared-pool capacity behavior has not been released or deployed, so there are no supported production rows created under the old max_tenants=0-as-unlimited contract. New DRIVE9_SHARED_POOL_DSN registrations persist DRIVE9_TIDBCLOUD_NATIVE_SHARED_MAX_TENANTS explicitly (default 100). Zero is intentionally reserved for drain-only, and we are not adding compatibility or backfill logic for unpublished behavior.

pkg/meta/db_pool.go
srstack: Follow-up after product clarification: the shared physical target is no longer hard-coded to 1,000,000. New managed pools use DRIVE9_TIDBCLOUD_NATIVE_DB_POOL_DEFAULT_SPENDING_LIMIT (default 1,000,000), persist that value per pool, and may be configured above 1,000,000 up to the current int32 Cloud request wire range. The tenant field remains virtual compatibility metadata and still does not participate in allocation or physical-limit calculation.

pkg/tenant/tidbcloudnative/provisioner.go
```diff
@@ -1497,7 +1580,7 @@ func validateTiDBCloudSpendingLimit(monthly int64) error {
 	if monthly > 0 && monthly < 10 {
 		return fmt.Errorf("tidbcloud_spending_limit must be 0 or at least 10 RMB")
 	}
-	if monthly > meta.MaxTiDBCloudSpendingLimit {
+	if monthly > int64(math.MaxInt32) {
```
mornyx, line 1583: **Question — is `spendingLimit.monthly` really accepted above 1,000,000 by the TiDB Cloud API?**

An intermediate commit in this PR (`86b47ee`) asserted the API maximum is 1M and forced pools to exactly that value; this round re-opens configuration up to `MaxInt32`. If the API truly caps at 1M, an operator setting `DRIVE9_TIDBCLOUD_NATIVE_DB_POOL_DEFAULT_SPENDING_LIMIT > 1000000` only finds out at `batchCreate` — the create fails and the continuation retries it indefinitely. If you've verified the API accepts larger values, all good; otherwise clamp the env/provisioner validation to the real maximum so misconfiguration fails at startup instead of in a retry loop.

mornyx (approved):
Re-reviewed the 2 new commits (`e217a58`, `83fb963`; CI still running at review time).

Both items from my last round are resolved:

1. **Virtual spending limit semantics** — now documented at the API surface (`quotaFields`/`quotaConfigResponse`): for shared tenants it's virtual compatibility metadata, not enforced. That's exactly the clarification I asked for. ✅
2. **Pool spending target** — configurable again via `DRIVE9_TIDBCLOUD_NATIVE_DB_POOL_DEFAULT_SPENDING_LIMIT` (default 1M), persisted per-pool, with the blanket normalization migration removed (per-pool values are now meaningful). Validation is sane: env `(0, MaxInt32]`, `CreateManagedSharedDBPool` `(0, MaxInt32]`, tenant-facing quota API still capped at `MaxTiDBCloudSpendingLimit` (1M) — a deliberate split between operator-configured pool targets and tenant-facing limits. The provisioner no longer imports `pkg/meta` for the constant. ✅

One question inline about the actual TiDB Cloud API maximum for `spendingLimit.monthly` — an intermediate commit in this PR asserted 1M is the API maximum, but this round allows configuring pool targets up to MaxInt32; if the API truly caps at 1M, a larger configured target would fail at `batchCreate` and loop in the continuation. Worth confirming against the real API.

Still the only open item from the first round: project:owner → org-scope authorization widening needs explicit product sign-off (inline on `tidbcloud_iam.go`). Everything else from all three rounds is addressed.

srstack:
LGTM. Reviewed the full diff in depth — this is a clean, well-executed security hardening of the TiDB Cloud authorization path. (Note: posting as a comment since GitHub won't let me formally approve my own PR.)

## What stands out

**IAM-based authorization replacing cluster-list RBAC**
- Resolving identity via `GET /v1beta1/apikeys/{key}` and authorizing at organization scope (`org:owner`/`project:owner` only) is a solid simplification. It removes a whole class of per-cluster enumeration calls and makes the authorization model much easier to reason about.
- The access-key echo check (`info.AccessKey != publicKey`) is a nice defense against response confusion.
- Role validation is enforced in both `ResolveAPIKeyIdentity` (provisioner) and `resolveTiDBCloudIdentity` (server) — redundant but correctly fail-closed.

**Credential separation**
- Routing all physical shared-DB-pool operations through server-managed `DRIVE9_TIDBCLOUD_NATIVE_SHARED_*` credentials while using the customer key only for IAM org resolution is the right boundary. `TestProvisionTiDBCloudNativeSharedPlansManagedPoolAndReturnsProvisioning` asserting the customer key is used for IAM only and the shared key for the physical batch is a good guard.
- Startup-time `ValidateSharedCredentials` fails fast on misconfiguration.

**Secret redaction**
- `redactRequestError` + `requestPath` masking of `/v1beta1/apikeys/***`, plus the mismatch/error/malformed-name tests that assert no access key, secret, or display name leaks into errors or logs — thorough and exactly the kind of coverage this path needs.

## Behavior changes worth calling out (all intentional & tested)
- `MaxTenants == 0` now means drain-only rather than unlimited. Env-registered pools now pass `sharedDBMaxTenants` through the upsert, so they self-correct to the configured cap on restart; managed pools already require a positive value. Covered by `TestSharedDBMaxTenantsZeroIsDrainOnly` and `TestDrainOnlySharedDBAllowsExistingTenantRelease`.
- Native quota GET without a `tenant_tidbcloud_org_bindings` row now returns 404 instead of 200. Correctly scoped to native tenants — shared tenants still authorize through the physical pool org, so no regression there.
- Wildcard shared-pool org (`*`) fully removed; shared placement now requires an exact org and fails closed on unresolved org.
- Physical pool spending limit is now separately configurable (up to MaxInt32) and decoupled from the tenant-facing `MaxTiDBCloudSpendingLimit`.

Test coverage is comprehensive and the RBAC cache refactor (identity-by-credential instead of cluster lists) removes the need for the old invalidation dance. Nice work.

PR #784, opened 2026-07-24
fix: label shared clusters with customer organization

srstack:
## Summary

- propagate the logical customer organization through batch, direct, and continuation shared DB cluster creation paths
- label newly created tidb_cloud_native_shared clusters with drive9.ai/customer_organization
- reject a missing customer organization before issuing the TiDB Cloud batch-create request

## Test plan

- go test ./pkg/tenant/... -count=1
- go test ./pkg/server -count=1 -run '^(TestProvisionTiDBCloudNativeSharedPlansManagedPoolAndReturnsProvisioning|TestManagedSharedDBContinuationWaitsForConnectionMetadata|TestManagedSharedDBContinuationPassesCustomerOrganizationToPhysicalCreate|TestManagedSharedDBBatchCreateUsesAllocationLock|TestManagedSharedDBBatchCreateReturnsTotalFailure)$'
- go test ./pkg/server -run '^$' -count=1
- make lint

## Verification note

The full pkg/server suite already failed in the pre-change baseline. A post-change fail-fast run stopped at the unrelated TestAdminTenantPoolUpdateShrinkPartialFailureReconcilesSize; that test passes when run in isolation after a transient local testcontainers/Ryuk port collision cleared.

## Summary by CodeRabbit

* **New Features**
  * Shared database pool provisioning now associates each pool with its customer organization.
  * Customer organization information is included in managed pool creation and continuation flows.
  * Provisioning requests are rejected when the required customer organization information is missing.

* **Bug Fixes**
  * Improved organization metadata propagation during shared database pool creation and recovery.

### DIFF
--- a/pkg/tenant/tidbcloudnative/provisioner.go
+++ b/pkg/tenant/tidbcloudnative/provisioner.go
@@ -58,15 +58,16 @@ const (
 	cloudProviderAliCloud     = "alicloud"
 	cloudProviderAWS          = "aws"
 
-	Drive9ManagedLabel         = "drive9.ai/managed"
-	Drive9TenantIDLabel        = "drive9.ai/tenant_id"
-	Drive9PoolStatusLabel      = "drive9.ai/status"
-	Drive9PoolIDLabel          = "drive9.ai/pool_id"
-	Drive9PoolUsedAtLabel      = "drive9.ai/used_at"
-	Drive9ProviderLabel        = "drive9.ai/provider"
-	Drive9DBPoolUUIDLabel      = "drive9.ai/db_pool_uuid"
-	Drive9QuotaUpdateAtLabel   = "drive9.ai/update_quota_at"
-	TiDBCloudOrganizationLabel = "tidb.cloud/organization"
+	Drive9ManagedLabel              = "drive9.ai/managed"
+	Drive9TenantIDLabel             = "drive9.ai/tenant_id"
+	Drive9PoolStatusLabel           = "drive9.ai/status"
+	Drive9PoolIDLabel               = "drive9.ai/pool_id"
+	Drive9PoolUsedAtLabel           = "drive9.ai/used_at"
+	Drive9ProviderLabel             = "drive9.ai/provider"
+	Drive9DBPoolUUIDLabel           = "drive9.ai/db_pool_uuid"
+	Drive9CustomerOrganizationLabel = "drive9.ai/customer_organization"
+	Drive9QuotaUpdateAtLabel        = "drive9.ai/update_quota_at"
+	TiDBCloudOrganizationLabel      = "tidb.cloud/organization"
 
 	upstreamErrorBodyLimit   = 2048
 	upstreamClusterBodyLimit = 1 << 20
@@ -599,6 +600,10 @@ func (p *Provisioner) BatchProvisionSharedDBPoolsWithCredentials(ctx context.Con
 		if _, exists := inputsByUUID[poolUUID]; exists {
 			return nil, fmt.Errorf("duplicate db pool uuid %s", poolUUID)
 		}
+		customerOrganizationID := strings.TrimSpace(input.CustomerOrganizationID)
+		if customerOrganizationID == "" {
+			return nil, fmt.Errorf("customer organization is required for db pool %d", input.DBPoolID)
+		}
 		if err := validateTiDBCloudSpendingLimit(input.SpendingLimitMonthly); err != nil {
 			return nil, err
 		}
@@ -611,6 +616,7 @@ func (p *Provisioner) BatchProvisionSharedDBPoolsWithCredentials(ctx context.Con
 			return nil, fmt.Errorf("root password is required for db pool %d", input.DBPoolID)
 		}
 		input.DBPoolUUID = poolUUID
+		input.CustomerOrganizationID = customerOrganizationID
 		input.RootPassword = password
 		inputsByUUID[poolUUID] = input
 		databaseNames[poolUUID] = dbName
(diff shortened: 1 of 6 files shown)


srstack:
LGTM. Reviewed the full diff — clean, well-tested follow-up to the IAM authz work in #780. CI is green. (Posting as a comment since GitHub blocks approving one's own PR.)

## What's good
- **Complete call-site coverage.** All three production `SharedDBPoolCreateRequest` construction sites — the batch path (`admin_tenant_pool.go`), direct `ensureManagedSharedDBPhysicalLocked`, and `continueManagedSharedDBPoolLocked` — now propagate `CustomerOrganizationID` from `row.TiDBCloudOrganizationID`. Confirmed there are no other constructors that would silently hit the new required-field check.
- **Fail-fast validation.** Rejecting a missing customer org in `BatchProvisionSharedDBPoolsWithCredentials` *before* issuing the batch-create request is the right place — the `...RejectsMissingCustomerOrganizationBeforeRequest` test asserting zero upstream calls locks that in.
- **Model consistency.** Since #780 removed the wildcard org and bound each shared pool to exactly one customer org, stamping a single `drive9.ai/customer_organization` label on the cluster is correct and unambiguous. Keeping it distinct from the physical `tidb.cloud/organization` is a good call for later attribution/cleanup.
- **Good test hook.** `sharedPoolPartialErr` (results returned alongside an error) exercises the continuation path's partial-failure branch while still capturing the emitted request — that's what makes `...PassesCustomerOrganizationToPhysicalCreate` a meaningful assertion rather than a happy-path-only check. The `sharedPoolBatchMu`-guarded request snapshot is properly synchronized.

## Nits (non-blocking)
- `int32(input.SpendingLimitMonthly)` in the request body is pre-existing and safe (bounded by `validateTiDBCloudSpendingLimit`), just noting it's untouched here.
- Worth a quick sanity check that customer org IDs (numeric strings like `1234567890123456789`) always satisfy TiDB Cloud label-value constraints — they will for org IDs, but if any non-numeric logical org ever flows here it'd be rejected at the API. Not a concern for the current model.

Nice incremental hardening.

srstack:
Re-reviewed after `69e9553e test: simplify shared create request capture`. Test-only change — production code is unchanged from my prior review, so the LGTM stands.

The refactor (mutex + slice + `sharedPoolBatchRequestsSnapshot()` → buffered cap-1 channel) is a real improvement, not just cosmetic:

- **Adds a synchronization barrier.** `<-prov.sharedPoolBatchRequests` blocks until the batch call has actually occurred. The old snapshot read right after the call site could observe an empty slice if the batch ran on a background continuation goroutine — the mutex guarded the data race but not that logical/timing race. This removes the latent flakiness.
- **Correct nil-guard.** `if f.sharedPoolBatchRequests != nil` keeps the many other `fakeProvisioner` users (which don't set the channel) from sending on a nil channel. Good.
- **Cap-1 is safe.** Every test using it asserts exactly one batch call, and the concurrency test (`...UsesAllocationLock`) sends into the buffered slot before its `started`/`release` handshake, so the send never blocks against the release coordination.

Non-blocking note: a future test exercising two batch calls would deadlock on the second send with cap-1 — but that surfaces as an obvious timeout, not a silent bug, so it's fine as-is.

Waiting on CI to go green on the new head; no code concerns.

pkg/tenant/tidbcloudnative/provisioner.go
```diff
@@ -619,9 +625,10 @@ func (p *Provisioner) BatchProvisionSharedDBPoolsWithCredentials(ctx context.Con
 			"rootPassword": password,
 			"region":       map[string]string{"name": p.regionName()},
 			"labels": map[string]string{
-				Drive9ManagedLabel:    "true",
-				Drive9ProviderLabel:   tenant.ProviderTiDBCloudNativeShared,
-				Drive9DBPoolUUIDLabel: poolUUID,
+				Drive9ManagedLabel:              "true",
+				Drive9ProviderLabel:             tenant.ProviderTiDBCloudNativeShared,
+				Drive9DBPoolUUIDLabel:           poolUUID,
+				Drive9CustomerOrganizationLabel: input.CustomerOrganizationID,
```
mornyx, line 631: **Operational note — pre-existing pools have no customer label**

Clusters created before this change lack `drive9.ai/customer_organization`, so any reconciliation/alerting that joins cloud clusters to `db_pool` rows on this label must treat a missing value as `unknown` (or backfill via `UpdateQuota`-style label patch on the next continuation). Worth a line in the tidbcloud/runbooks companion PR so the alert rules don't silently skip legacy pools.

mornyx (approved):
Reviewed at `69e9553` (2 commits, +135/−27, 6 files; CI still running at review time, local build passes).

Small, correct follow-up to the 780 ownership redesign — the physical cluster now carries the logical customer org as `drive9.ai/customer_organization`, which is exactly what cloud-side attribution and the manual reconciliation in our dev test plan need (db_pool rows ↔ cloud clusters can now be joined on this label).

Verified:
- **All three creation paths pass it**: prewarm batch (`provisionManagedSharedDBPoolsBatchLocked`), direct ensure (`ensureManagedSharedDBPhysicalLocked`), and continuation (`continueManagedSharedDBPoolLocked`) — those are the only callers of `BatchProvisionSharedDBPoolsWithCredentials`.
- **Source of truth is right**: the label value comes from the persisted `db_pool.org_id` (the customer's IAM org), not from the cloud response or the shared credential — consistent with the 780 model.
- **Fail-fast before the wire**: empty customer org is rejected before issuing batchCreate; from the current callers it's unreachable (780 already requires non-empty at every earlier gate), so it's pure defense-in-depth.
- **Tests cover all three paths** plus the pre-request rejection and the label on the outgoing request body.

One operational note below for the runbooks companion PR — clusters created before this change carry no `drive9.ai/customer_organization` label.

pkg/tenant/tidbcloudnative/provisioner.go
srstack: Cluster metadata labels and Drive9 Prometheus metric labels are separate data paths. Alerts use tidbcloud_org_id resolved from the persisted db_pool.org_id; they do not derive organization identity from TiDB Cloud cluster labels. Therefore legacy clusters missing drive9.ai/customer_organization do not affect alert evaluation or cause pools to be skipped. This PR intentionally labels newly created clusters only; no runbooks change or backfill is required for alerting.

PR #787, opened 2026-07-24
fix: correct shared DB pool uniqueness

srstack:
## Summary

- remove endpoint-based uniqueness from db_pool because shared clusters use the same host and database
- enforce physical cloud resource identity with globally unique cluster_id
- drop the legacy endpoint unique index during migration
- keep local/dev DSN registration idempotent through the existing UUID unique key

## Verification

- targeted pkg/meta schema, migration, managed shared pool, and registration tests
- git diff --check

### DIFF
--- a/pkg/meta/db_pool.go
+++ b/pkg/meta/db_pool.go
@@ -559,10 +559,10 @@ func (s *Store) FindSharedDBForEmergency(ctx context.Context, organizationID str
 	return db, nil
 }
 
-// RegisterSharedDB registers a physical database in db_pool, upserting on the
-// uk_db_pool_endpoint (org_id, db_host, db_name, db_user) natural key. On a duplicate
-// endpoint the connection fields are refreshed and tenant_count is preserved;
-// the existing db_id is re-fetched and returned.
+// RegisterSharedDB registers a manually configured physical database in
+// db_pool. Its stable UUID is derived from organization and user when omitted;
+// managed cloud pools use CreateManagedSharedDBPool and global cluster_id
+// uniqueness instead.
 func (s *Store) RegisterSharedDB(ctx context.Context, in *SharedDB) (dbID int64, err error) {
 	start := time.Now()
 	defer observeMeta(ctx, "register_shared_db", start, &err)
@@ -570,10 +570,6 @@ func (s *Store) RegisterSharedDB(ctx context.Context, in *SharedDB) (dbID int64,
 		return 0, fmt.Errorf("shared db is required")
 	}
 	organizationID, err := validateSharedDBOrganizationID(in.TiDBCloudOrganizationID)
-	if err != nil {
-		return 0, err
-	}
-	poolUUID, err := sharedDBUUID(in.UUID)
 	if err != nil {
 		return 0, err
 	}
@@ -585,6 +581,15 @@ func (s *Store) RegisterSharedDB(ctx context.Context, in *SharedDB) (dbID int64,
 	}
 	if in.User == "" {
 		return 0, fmt.Errorf("db user is required")
+	}
+	poolUUID := strings.TrimSpace(in.UUID)
+	if poolUUID == "" {
+		poolUUID = uuid.NewSHA1(uuid.NameSpaceOID, []byte(organizationID+"\x00"+in.User)).String()
+	} else {
+		poolUUID, err = sharedDBUUID(poolUUID)
+		if err != nil {
+			return 0, err
+		}
 	}
 	if len(in.PasswordCipher) == 0 {
 		return 0, fmt.Errorf("db password cipher is required")
@@ -612,7 +617,8 @@ func (s *Store) RegisterSharedDB(ctx context.Context, in *SharedDB) (dbID int64,
 	if _, err := s.db.ExecContext(ctx,
 		"INSERT INTO db_pool (uuid, org_id, `role`, db_host, db_port, db_user, db_password, db_name, "+
 			"db_tls, max_tenants, status) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?) "+
-			"ON DUPLICATE KEY UPDATE db_port = VALUES(db_port), db_user = VALUES(db_user), "+
+			"ON DUPLICATE KEY UPDATE org_id = VALUES(org_id), db_host = VALUES(db_host), "+
+			"db_port = VALUES(db_port), db_user = VALUES(db_user), db_name = VALUES(db_name), "+
 			"db_password = VALUES(db_password), db_tls = VALUES(db_tls), "+
 			"max_tenants = VALUES(max_tenants), status = VALUES(status)",
 		poolUUID, organizationID, role, in.Host, in.Port, in.User, in.PasswordCipher, in.Name,
@@ -620,10 +626,9 @@ func (s *Store) RegisterSharedDB(ctx context.Context, in *SharedDB) (dbID int64,
 		return 0, fmt.Errorf("upsert db_pool row: %w", err)
 	}
 	// ON DUPLICATE KEY UPDATE does not reliably report the existing
-	// auto-increment id via LastInsertId, so re-fetch by endpoint.
+	// auto-increment id via LastInsertId, so re-fetch by UUID.
 	err = s.db.QueryRowContext(ctx,
-		`SELECT db_id FROM db_pool WHERE org_id = ? AND db_host = ? AND db_name = ? AND db_user = ?`,
-		organizationID, in.Host, in.Name, in.User).Scan(&dbID)
+		`SELECT db_id FROM db_pool WHERE uuid = ?`, poolUUID).Scan(&dbID)
 	if err != nil {
 		return 0, fmt.Errorf("resolve db_id after upsert: %w", err)
 	}
--- a/pkg/meta/db_pool.go
+++ b/pkg/meta/db_pool.go
@@ -560,9 +560,9 @@ func (s *Store) FindSharedDBForEmergency(ctx context.Context, organizationID str
 }
 
 // RegisterSharedDB registers a manually configured physical database in
-// db_pool. Its stable UUID is derived from organization and user when omitted;
-// managed cloud pools use CreateManagedSharedDBPool and global cluster_id
-// uniqueness instead.
+// db_pool. Its stable UUID is derived from the full manual connection identity
+// when omitted; managed cloud pools use CreateManagedSharedDBPool and global
+// cluster_id uniqueness instead.
 func (s *Store) RegisterSharedDB(ctx context.Context, in *SharedDB) (dbID int64, err error) {
 	start := time.Now()
 	defer observeMeta(ctx, "register_shared_db", start, &err)
@@ -615,23 +615,39 @@ func (s *Store) RegisterSharedDB(ctx context.Context, in *SharedDB) (dbID int64,
 	if status == "" {
 		status = sharedDBStatusActive
 	}
-	if _, err := s.db.ExecContext(ctx,
-		"INSERT INTO db_pool (uuid, org_id, `role`, db_host, db_port, db_user, db_password, db_name, "+
-			"db_tls, max_tenants, status) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?) "+
-			"ON DUPLICATE KEY UPDATE org_id = VALUES(org_id), db_host = VALUES(db_host), "+
-			"db_port = VALUES(db_port), db_user = VALUES(db_user), db_name = VALUES(db_name), "+
-			"db_password = VALUES(db_password), db_tls = VALUES(db_tls), "+
-			"max_tenants = VALUES(max_tenants), status = VALUES(status)",
-		poolUUID, organizationID, role, in.Host, in.Port, in.User, in.PasswordCipher, in.Name,
-		in.TLSMode, in.MaxTenants, status); err != nil {
-		return 0, fmt.Errorf("upsert db_pool row: %w", err)
-	}
-	// ON DUPLICATE KEY UPDATE does not reliably report the existing
-	// auto-increment id via LastInsertId, so re-fetch by UUID.
-	err = s.db.QueryRowContext(ctx,
-		`SELECT db_id FROM db_pool WHERE uuid = ?`, poolUUID).Scan(&dbID)
+	err = s.InTx(ctx, func(tx *sql.Tx) error {
+		err := tx.QueryRowContext(ctx, `SELECT db_id FROM db_pool
+			WHERE org_id = ? AND db_host = ? AND db_name = ? AND db_user = ?
+			ORDER BY db_id LIMIT 1 FOR UPDATE`, organizationID, in.Host, in.Name, in.User).Scan(&dbID)
+		if err == nil {
+			if _, err := tx.ExecContext(ctx, `UPDATE db_pool
+				SET db_port = ?, db_password = ?, db_tls = ?, max_tenants = ?, status = ?
+				WHERE db_id = ?`, in.Port, in.PasswordCipher, in.TLSMode, in.MaxTenants, status, dbID); err != nil {
+				return fmt.Errorf("update existing manual db_pool row: %w", err)
+			}
+			return nil
+		}
+		if !errors.Is(err, sql.ErrNoRows) {
+			return fmt.Errorf("find existing manual db_pool row: %w", err)
+		}
+		if _, err := tx.ExecContext(ctx,
+			"INSERT INTO db_pool (uuid, org_id, `role`, db_host, db_port, db_user, db_password, db_name, "+
+				"db_tls, max_tenants, status) VALUES (?, ?, ?, ?, ?, ?, ?, ?, ?, ?, ?) "+
+				"ON DUPLICATE KEY UPDATE org_id = VALUES(org_id), db_host = VALUES(db_host), "+
+				"db_port = VALUES(db_port), db_user = VALUES(db_user), db_name = VALUES(db_name), "+
+				"db_password = VALUES(db_password), db_tls = VALUES(db_tls), "+
+				"max_tenants = VALUES(max_tenants), status = VALUES(status)",
+			poolUUID, organizationID, role, in.Host, in.Port, in.User, in.PasswordCipher, in.Name,
+			in.TLSMode, in.MaxTenants, status); err != nil {
+			return fmt.Errorf("upsert manual db_pool row: %w", err)
+		}
+		if err := tx.QueryRowContext(ctx, `SELECT db_id FROM db_pool WHERE uuid = ?`, poolUUID).Scan(&dbID); err != nil {
+			return fmt.Errorf("resolve db_id after upsert: %w", err)
+		}
+		return nil
+	})
 	if err != nil {
-		return 0, fmt.Errorf("resolve db_id after upsert: %w", err)
+		return 0, err
 	}
 	return dbID, nil
 }
--- a/pkg/meta/db_pool.go
+++ b/pkg/meta/db_pool.go
@@ -584,7 +584,8 @@ func (s *Store) RegisterSharedDB(ctx context.Context, in *SharedDB) (dbID int64,
 	}
 	poolUUID := strings.TrimSpace(in.UUID)
 	if poolUUID == "" {
-		poolUUID = uuid.NewSHA1(uuid.NameSpaceOID, []byte(organizationID+"\x00"+in.User)).String()
+		poolUUID = uuid.NewSHA1(uuid.NameSpaceOID, []byte(strings.Join(
+			[]string{organizationID, in.Host, in.Name, in.User}, "\x00"))).String()
 	} else {
 		poolUUID, err = sharedDBUUID(poolUUID)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


srstack:
LGTM. Reviewed the full diff — small, well-scoped fix that widens the db_pool endpoint uniqueness key to include `db_user`, so multiple Tencent shared clusters sharing one private-endpoint host + database but differing by DB user no longer collide. (Posting as a comment — GitHub blocks approving one's own PR. CI still running at review time.)

## Correctness verified

**Migration is safe against existing data.** The key is being *widened* (3 cols → 4 cols by adding `db_user`), so the new constraint is strictly weaker than the old. Any data that satisfied `(org_id, db_host, db_name)` uniqueness trivially satisfies the 4-column version, so the `DROP INDEX ... ADD UNIQUE INDEX` can't fail on a duplicate-key violation during migration.

**Migration is idempotent.** `replaceLegacyDBPoolEndpointUniqueIndex` only rewrites when the index columns are exactly `[org_id, db_host, db_name]`; after migration they become the 4-column set, so a rerun (or a fresh DB that already created the 4-column index) short-circuits via `sameStringSlice` → no-op. Fresh-schema DDL and the migration path are kept consistent (both land on `(org_id, db_host(191), db_name, db_user)`).

**Re-fetch updated to match.** `RegisterSharedDB`'s post-upsert `SELECT db_id ... WHERE org_id=? AND db_host=? AND db_name=? AND db_user=?` now includes `db_user` — necessary, since org+host+name is no longer unique on its own. Without this the re-fetch could return the wrong `db_id` when two rows share the first three columns. Good catch to update it in lockstep.

**`cluster_id` identity preserved.** `uk_db_pool_cloud_resource (org_id, cluster_id)` is untouched, so cloud-resource identity is still enforced independently of the endpoint key. The new test creates two managed pools with distinct `cluster_id`s + distinct users on the same host/db and confirms both persist to `provisioning` — covers the target scenario.

## Nits (non-blocking)
- **Redundant upsert assignment**: the `ON DUPLICATE KEY UPDATE` clause still sets `db_user = VALUES(db_user)`. Since `db_user` is now part of the unique key, a duplicate can only match when `db_user` is already equal, making this a harmless no-op self-assignment. Not wrong, just dead. Could drop it for clarity.
- **`db_host(191)` prefix**: sound choice — a 4-column `utf8mb4` unique index needs the host prefixed to stay under InnoDB's index-length limit (191×4 ≈ 764 bytes). Hostnames are well under 191 chars so there's no practical uniqueness loss, but worth a one-line comment on the index noting *why* it's prefixed, since it's non-obvious.

Clean fix with the right migration shape and matching read path. Nice.

qiffang:
## adversary-2 review at `6398616`

### Notes (no blockers)

**N1: `db_host(191)` prefix index narrows uniqueness guarantee**
Old: `UNIQUE (org_id, db_host, db_name)` — full-length `db_host`. New: `UNIQUE (org_id, db_host(191), db_name, db_user)` — only first 191 bytes compared. Practically safe for hostnames/IPs but is a strictness regression.

**N2: Verify `db_user` is NOT NULL**
MySQL unique indexes treat each NULL as distinct — if `db_user` allows NULL, two rows with identical `(org_id, db_host, db_name, NULL)` would NOT violate the constraint. Column must be NOT NULL for uniqueness to hold.

**N3: Migration is safe** — old 3-column key is strictly narrower than new 4-column key, no existing data can violate. Atomic ALTER is correct. Idempotent guard via `sameStringSlice` check.

**N4: Test coverage adequate** — same-endpoint-different-users, upsert, migration index verification all covered.

**Verdict: GREEN** — clean, focused change.

mornyx (approved):
Reviewed at `dfec3cb` (2 commits, +71/−7, 4 files; CI still running at review time).

Both identity changes are correct and cover a real collision:

1. **Endpoint uniqueness `(org, host, name)` → `(org, host(191), name, user)`** — this is a genuine production bug fix, not hygiene: on Tencent with private endpoints, two shared clusters share the same endpoint IP and the same database name (`tidbcloud_fs`), so the second pool's `UpdateManagedSharedDBPoolCloudResult` would collide on the old unique index. `user` (`{userPrefix}.root`, unique per cluster) is the right discriminator, and `TestManagedSharedDBPoolsAllowSameEndpointWithDifferentUsers` reproduces exactly that scenario. The `db_host(191)` prefix keeps the composite under InnoDB's 3072-byte limit (191×4 + 64×4 + 255×4 + 255×4 = 3060). `RegisterSharedDB`'s re-fetch correctly includes `db_user`.
2. **Cloud resource uniqueness `(org, cluster_id)` → `(cluster_id)`** — cluster IDs are globally unique in TiDB Cloud, and global uniqueness hardens the 780 logical-org model: the same cluster can never be adopted under two organizations. Multiple NULLs (pending pools) are unaffected.
3. **Migrations are atomic** (single `ALTER TABLE ... DROP INDEX, ADD UNIQUE INDEX`) and column-aware (`loadMetaIndexColumns` + exact-slice compare, so they only fire on the legacy shape and are re-run safe). The endpoint change only relaxes uniqueness (no dedup needed); the cloud-resource change tightens it — see note below.

One sequencing nit (not blocking): `dropObsoleteMetaIndexes` (which hosts the replace) runs *before* `expandManagedDBPoolSchema` (which hosts the duplicate preflight), so a legacy DB with duplicate `cluster_id`s across orgs will fail on the raw `ALTER` duplicate-entry error wrapped as `replace db_pool.uk_db_pool_cloud_resource: ...` instead of the nicer preflight message. It still fails closed with an actionable error — just slightly uglier. Moving the replace after the preflight would give the better message.

srstack:
Re-reviewed after `dfec3cbb fix: make shared cluster id globally unique`. This narrows the cloud-resource key from `(org_id, cluster_id)` → `(cluster_id)`. Semantically correct — a TiDB Cloud cluster_id is globally unique by nature, so scoping it per-org was always weaker than reality. But narrowing a unique key is the risk-prone direction, so a couple of points:

## Verified OK
- **NULL cluster_id is fine.** `cluster_id` is `VARCHAR(255) NULL`; MySQL/InnoDB allows multiple NULLs under a UNIQUE index, so `pending` pools without a cluster identity don't collide. Good — this is what makes the global constraint compatible with the new pending state from #786.
- **No composite-key read path breaks.** Nothing queries `WHERE org_id = ? AND cluster_id = ?` for lookup; the only `org_id AND cluster_id` reference is the preflight in `expandManagedDBPoolSchema`, which this commit updates. Fresh-schema DDL and both migration entry points now agree on `(cluster_id)`.
- **Idempotent.** `replaceLegacyDBPoolCloudResourceUniqueIndex` only rewrites when columns are exactly `[org_id, cluster_id]`; post-migration `[cluster_id]` short-circuits.

## Concern — migration path has no duplicate preflight (worth addressing)
This is the mirror of the endpoint change but in the **narrowing** direction, so existing data *can* violate the new constraint: any two `db_pool` rows sharing a `cluster_id` across different `org_id`s would make the new index impossible.

The two paths handle that inconsistently:
- `expandManagedDBPoolSchema` (index-never-existed path) **preflights** with `GROUP BY cluster_id HAVING COUNT(*)>1` and returns a clean `found N duplicate cluster ids` error.
- `replaceLegacyDBPoolCloudResourceUniqueIndex` (legacy-`(org_id,cluster_id)`-index path) has **no preflight** — it runs `DROP INDEX ... ADD UNIQUE INDEX (cluster_id)` directly. On duplicate data the `ALTER` fails with a raw MySQL `Duplicate entry` error and aborts startup, rather than the descriptive message.

Neither path *resolves* duplicates (both abort), which is a reasonable fail-closed stance. But the legacy path is the one real deployments will actually hit, and it's the one without the friendly diagnostic. Suggest adding the same `GROUP BY cluster_id HAVING COUNT(*)>1` preflight to `replaceLegacyDBPoolCloudResourceUniqueIndex` so an operator gets "found N duplicate cluster ids" instead of a bare ALTER failure. Given that cluster_ids should already be globally unique in practice, this is defense-in-depth, not a likely-hit path — non-blocking, but cheap and consistent.

Rest of the PR (endpoint key widening + user in re-fetch) stands as reviewed above. CI still running.

qiffang:
## adversary-2 R2 at `d1c147d9`

Scope expanded significantly since R1. Endpoint uniqueness removed entirely, cloud resource key narrowed to global `cluster_id`.

**R1 N1/N2 are moot** — endpoint index is gone.

**New observations:**
- UUID-based dedup via `uk_db_pool_uuid` is the only dedup mechanism for `RegisterSharedDB`. Deterministic SHA1 from `(org, user)` ensures same logical endpoint → same UUID. Sound.
- `uk_db_pool_cloud_resource` narrowed to `(cluster_id)` only — semantically correct (TiDB Cloud cluster_id is globally unique). NULL cluster_id for pending pools is fine (MySQL allows multiple NULLs in unique indexes).
- ON DUPLICATE KEY UPDATE now refreshes `org_id`, `db_host`, `db_name` — these become mutable on re-register. Acceptable for manually configured pools.
- Migration: drops endpoint index unconditionally (idempotent), replaces cloud resource index with preflight check. Correct.
- Tests: schema spec asserts no endpoint index, migration verifies drop, new test verifies narrowed cloud resource columns.

**Verdict: GREEN** at `d1c147d9`.

srstack:
Re-reviewed after `d1c147d9 fix: remove shared endpoint uniqueness`. **This commit introduces a blocking regression — I reproduced a failing test.** ⚠️

## Blocking: two shared DBs with the same (org, user) now collapse into one row

This commit drops `uk_db_pool_endpoint` entirely and changes `RegisterSharedDB`'s upsert identity to a UUID derived from `SHA1(org_id + "\\x00" + user)` when no UUID is supplied. The upsert clause now also refreshes `org_id, db_host, db_name` on conflict. Net effect: for manual registrations, **the natural identity is now just (org_id, db_user)** — host and database name are no longer part of it.

That means two distinct physical databases under the same org with the same DB user can no longer coexist: the second registration silently upserts over the first.

**`TestListSharedDBs` now fails** (unchanged from main; it registers two `org-b` rows with `user: root` — an active one and a draining one, expecting two distinct rows):

```
--- FAIL: TestListSharedDBs (0.05s)
    db_pool_test.go:404: ListSharedDBs returned 1 rows, want 2
```

Both `org-b`/`root` registrations hash to the same `SHA1("org-b\\x00root")` UUID, so the draining registration overwrites the active row instead of inserting a new one — leaving 1 non-draining row instead of 2. I ran `go test ./pkg/meta -run TestListSharedDBs` against the branch head to confirm. CI will hit the same failure when it reaches `pkg/meta`.

The PR checklist lists `make test TEST_PKGS=./pkg/meta` — this latest commit doesn't pass it, so it looks like it wasn't re-run after `d1c147d9`.

## Two things to decide

1. **Is (org, user) the intended identity for manual pools?** If an operator can legitimately register two physical hosts for one org under the same user, this identity is too coarse and will silently drop one. If a manual pool is truly one-per-(org,user) by design, then the derived-UUID approach is fine — but the failing test needs to be reconciled (either the test's second `org-b` row should use a distinct user, or the design intent documented and the test updated deliberately).
2. **This is now a materially different PR than the title/description.** It went from "include DB user in shared endpoint identity" (widen the key) → "make cluster_id globally unique" → "remove shared endpoint uniqueness" (drop the key, derive UUID from org+user). The title, summary, and verification bullets are all stale now. Worth rewriting the description to match the final design so reviewers and `git log` reflect what actually shipped.

## Still-good parts (from earlier commits, unchanged here)
- `uk_db_pool_cloud_resource → (cluster_id)` narrowing is sound (global cluster_id uniqueness; NULLs allowed for pending pools). My earlier suggestion to add a duplicate preflight to the legacy-index replacement path still stands.
- Dropping `uk_db_pool_endpoint` via `dropMetaIndexIfExists` is idempotent and can't fail on data (unlike the earlier ALTER approaches) — that part is cleaner.

Please fix the `TestListSharedDBs` failure and confirm the (org, user) identity semantics before merge. Happy to re-review once updated.

pkg/meta/db_pool.go
```diff
@@ -586,6 +582,15 @@ func (s *Store) RegisterSharedDB(ctx context.Context, in *SharedDB) (dbID int64,
 	if in.User == "" {
 		return 0, fmt.Errorf("db user is required")
 	}
+	poolUUID := strings.TrimSpace(in.UUID)
+	if poolUUID == "" {
+		poolUUID = uuid.NewSHA1(uuid.NameSpaceOID, []byte(organizationID+"\x00"+in.User)).String()
```
qiffang: **[P1] Preserve the full manual-pool identity and migrate existing UUIDs before dropping endpoint uniqueness.**

Deriving the key from only `(organizationID, db_user)` conflates distinct manual pools that use the same user. The existing `TestListSharedDBs` already reproduces this: `org-b` registers an active pool and then a different draining pool, both as `root`; the second call now hits the same UUID and overwrites the first row's host/name/status, so the test deterministically returns 1 active row instead of 2.

It also breaks upgrades. Before this commit, `RegisterSharedDB` generated a random UUID when omitted and relied on `uk_db_pool_endpoint` for restart idempotency. After migration drops that index, the first restart derives a different UUID and inserts a duplicate endpoint row; the older row remains active and is preferred by `FindSharedDBForAllocation`'s `db_id` ordering, so rotated credentials/config can be written only to the unused new row.

Please use a stable exact identity that distinguishes the intended `(org, host, database, user)` tuple (or an explicit stable ID), migrate/adopt legacy manual rows before removing the endpoint index, and add both regressions: distinct same-org/same-user endpoints remain distinct, and a pre-upgrade random-UUID row is reused rather than duplicated on restart.

qiffang (dismissed):
Requesting changes on exact head `d1c147d9`.

The prefix-index corruption from the previous revision is removed, and the global `cluster_id` key is reasonable, but the replacement manual-pool identity is both too coarse and incompatible with existing rows. The inline P1 is merge-blocking.

Validation:
- `git diff --check` — PASS
- focused PR tests — PASS
- `go test -count=1 -run '^TestListSharedDBs$' ./pkg/meta` — FAIL deterministically (`ListSharedDBs returned 1 rows, want 2`) because `(org,user)` collapses different endpoints
- an upgrade probe seeded an existing random-UUID manual row, then registered the same endpoint without UUID on this head — two matching endpoint rows remained with different `db_id`s, confirming the migration gap

Please add regressions for both cases before re-review.

qiffang:
## adversary-2 correction — retracting R2 GREEN

I missed two real bugs @dat9-dev2 caught. My R2 GREEN at `d1c147d9` was wrong.

1. **`(org, user)` UUID is strictly coarser than the old endpoint key.** `TestListSharedDBs` deterministically fails: two `org-b` pools both as `root` now share one UUID and the second overwrites the first. My reasoning "same logical endpoint → same UUID" did not verify that `(org, user)` actually distinguishes distinct endpoints. It does not.

2. **Upgrade path creates duplicates.** Pre-upgrade rows have random UUIDs. After migration drops the endpoint index, first restart derives a different SHA1 UUID for the same endpoint → duplicate row. Old row wins via `FindSharedDBForAllocation` db_id ordering, so rotated credentials go to the unused new row.

Endorsing @dat9-dev2's `CHANGES_REQUESTED`. Amended verdict: **REVISE**.

srstack:
Re-reviewed after `e0fe6483 fix: keep manual shared pool identities distinct`. **The blocking regression is fixed — verified locally.** ✅

The derived UUID now hashes the full endpoint tuple `SHA1(org_id + host + name + user)` instead of just `(org, user)`, so two distinct physical DBs under the same org+user no longer collapse into one row. This restores exactly the identity the widened `uk_db_pool_endpoint` would have provided, but expressed through the stable `uk_db_pool_uuid` key instead of a separate composite index — which is a cleaner design (one uniqueness mechanism, and it composes with the manually-supplied-UUID path).

Verified against the branch head in a worktree:
- `TestListSharedDBs` — now passes (the two `org-b`/`root` rows get distinct UUIDs from their differing host/name).
- `TestManagedSharedDBPoolsAllowSameEndpointWithDifferentUsers`, `TestRegisterSharedDB*`, `TestFindSharedDBForOrg*`, `TestMigrateExpandsLegacyDBPoolForManagedProvisioning`, `TestDBPoolCloudResourceIndexUsesGloballyUniqueClusterID` — pass.
- Full `go test ./pkg/meta -count=1` — **ok**.

## Net assessment of the final design — LGTM
- **Manual pools**: identity = `SHA1(org, host, name, user)` via `uk_db_pool_uuid`; upsert refreshes connection fields on conflict. Distinct endpoints coexist; re-registering the same endpoint updates in place. Correct.
- **Managed cloud pools**: identity = globally-unique `cluster_id` via `uk_db_pool_cloud_resource (cluster_id)`, with NULLs allowed for pending pools. Sound.
- **Migration**: `uk_db_pool_endpoint` dropped via idempotent `dropMetaIndexIfExists` (can't fail on data); cloud-resource index narrowed with a preflight on the create path.

## Remaining non-blocking items
1. The `replaceLegacyDBPoolCloudResourceUniqueIndex` (legacy `(org_id,cluster_id)` → `(cluster_id)`) path still has no duplicate preflight, unlike the create path — on the unlikely chance of a cross-org duplicate cluster_id it aborts with a raw ALTER error instead of the friendly "found N duplicate cluster ids". Cheap defense-in-depth; still worth adding.
2. PR title/description are stale — the final design is "drop endpoint uniqueness; derive a stable UUID from the endpoint tuple; make cluster_id globally unique," not "include DB user in shared endpoint identity." Please refresh the summary before merge so `git log`/reviewers reflect what shipped.

With the regression fixed and `pkg/meta` green, this is good to go on the code. Nice quick turnaround.

qiffang:
## adversary-2 R3 at `3d69b0fb`

Both R2 bugs are properly fixed with regression tests.

**`e0fe6483`** — UUID derivation uses full 4-tuple `(org, host, db_name, user)`. Same-org+user, different-host pools now get distinct UUIDs. `TestListSharedDBs` should pass.

**`3d69b0fb`** — `RegisterSharedDB` runs in a transaction:
1. `SELECT ... FOR UPDATE` on full endpoint tuple
2. If found → UPDATE existing row (preserves legacy random UUID + `db_id`)
3. If not → INSERT with new stable SHA1 UUID (relies on `uk_db_pool_uuid` for concurrent-insert dedup)

`TestRegisterSharedDBReusesLegacyRandomUUIDRow` seeds a legacy random-UUID row and verifies re-registration returns same `db_id` and single row.

**Verdict: GREEN** at `3d69b0fb`.

pkg/meta/meta.go
```diff
@@ -1048,6 +1047,29 @@ func dropObsoleteMetaIndexes(ctx context.Context, db *sql.DB) error {
 		[]string{"organization_id", "cluster_id", "created_at", "tenant_id"}); err != nil {
 		return fmt.Errorf("drop obsolete meta index idx_tidbcloud_org_cluster: %w", err)
 	}
+	if err := dropMetaIndexIfExists(ctx, db, "db_pool", "uk_db_pool_endpoint"); err != nil {
```
qiffang, line 1050: **[P1] Keep the old binary startable until the endpoint-index cutover is fenced.**

The exact base revision (`6c57714c`) treats `uk_db_pool_endpoint` as a required `db_pool` schema-contract index. I migrated a schema with this head, then opened it with the exact base binary; `meta.Open` fails immediately with:

```
migrate: meta schema contract mismatch: db_pool schema contract: missing uk_db_pool_endpoint index
```

So the first upgraded process that executes this drop closes the old-binary restart, scale-out, and rollback path while old pods may still exist. This happens before `RegisterSharedDB` can run, so the legacy-row adoption fix does not address it.

Please either stage this as a backward-compatible two-release migration, or explicitly enforce a stop-the-world cutover: all base pods/leaders drained and fenced before this migration, old binaries prevented from returning, plus a repair/rollback procedure. Add an exact base-after-new-migration compatibility/cutover gate so this deployment boundary is testable.

qiffang (changes requested):
R3 review on exact head `3d69b0fb`.

Both previous application-layer blockers are fixed: manual UUID identity now uses the full `(org, host, database, user)` tuple, and legacy random-UUID rows are adopted in place. The new regressions pass, as do full `./pkg/meta`, focused shared/managed server tests, `git diff --check`, and hosted `code-ci` / `local-e2e`.

One deployment blocker remains. Dropping `uk_db_pool_endpoint` makes the exact base binary fail its schema contract during `meta.Open`, so an old pod cannot restart, scale out, or serve as rollback after any upgraded process runs the migration. The PR does not define or enforce the required cutover boundary. The inline P1 must be resolved with a staged compatible migration or an explicit, testable hard-cutover/fence and repair procedure.

qiffang:
## adversary-2 retracting R3 GREEN

Verified @dat9-dev2's deployment blocker. Base binary `6c57714c`'s `meta.Open()` calls `migrate()` which builds a spec from `metaInitSchemaStatements()` (includes `uk_db_pool_endpoint`), diffs against live schema, applies safe auto-repairs, then re-diffs. Missing unique index on an existing table is NOT in the safe-repair allowlist — becomes fatal `metaSchemaDiffError`.

**Consequence after PR #787 migration runs**:
- Old pod restart → `meta.Open()` fails, cannot boot
- Scale-out with old image → same failure
- Rollback to base → same failure

Same class of blocker as PR #786 B1. Amending verdict: **REVISE**. Endorsing @dat9-dev2's CHANGES_REQUESTED.

PR #790, opened 2026-07-24
fix: parallelize managed shared pool refill

srstack:
## Summary

- refill shared tenant pools with a fixed 50-worker pipeline and treat exhausted claim CAS retries as a normal miss that falls through to active provisioning
- create physical shared pools in concurrent 10-pool batches with at most 10 Cloud requests, while ensuring all selected pending pools join the same create wave
- resume metadata/schema work with a fixed 100-worker pool and 15-second retries; reuse server-owned shared credentials without IAM calls in refill workers
- retain one idle test DB connection per pool to prevent CI TCP port exhaustion without saturating MySQL connections

## Test plan

- GOCACHE=/tmp/drive9-go-cache go test ./pkg/server -count=1 -timeout=10m
- GOCACHE=/tmp/drive9-go-cache go test -race ./pkg/server -run TestSharedTenantPoolRefillRunsFiftyTenantWorkers\|TestRetryManagedSharedDBContinuationsStartsEveryPoolConcurrently\|TestRetryManagedSharedDBContinuationsLimitsConcurrentPoolsToOneHundred -count=1 -timeout=3m
- git diff --check

## Summary by CodeRabbit

- **Improvements**
  - Shared-tenant pool refills complete faster with parallel per-tenant provisioning and more efficient managed shared DB continuation handling.
  - Managed shared DB provisioning/refill now runs in concurrent cloud batches with safer sizing, cooldown/backoff behavior, and improved coordination to reduce duplicate work.
  - Tenant-pool replenishment and pending-resume scans are coalesced per pool with minimum intervals and added cooldown after empty scans.
  - CAS retry now returns a non-error “miss” after exhausting “not found” failures.
  - Added configurable TiDB Cloud managed shared DB tuning via new environment variables (documented in usage output).
- **Tests**
  - Expanded coverage for concurrency limits, batching behavior, locking/cancellation, cooldown/coalescing, and updated retry expectations.

### DIFF
--- a/pkg/server/admin_tenant_pool.go
+++ b/pkg/server/admin_tenant_pool.go
@@ -51,9 +51,15 @@ const adminTenantPoolMetricsComponent = "admin_tenant_pool"
 
 const tenantPoolClaimCASRetryLimit = 8
 
+const sharedTenantPoolCreateConcurrency = 50
+
+const (
+	managedSharedDBCloudBatchSize        = 10
+	managedSharedDBCloudBatchConcurrency = 10
+)
+
 func retryTenantPoolClaimCAS[T any](attempt func() (T, error)) (T, error) {
 	var zero T
-	var lastErr error
 	for range tenantPoolClaimCASRetryLimit {
 		result, err := attempt()
 		if err == nil {
@@ -62,10 +68,8 @@ func retryTenantPoolClaimCAS[T any](attempt func() (T, error)) (T, error) {
 		if !errors.Is(err, meta.ErrNotFound) {
 			return zero, err
 		}
-		lastErr = err
 	}
-	return zero, fmt.Errorf("tenant pool claim remained busy after %d attempts: %w",
-		tenantPoolClaimCASRetryLimit, lastErr)
+	return zero, nil
 }
 
 type tenantPoolClaimSelection struct {
@@ -936,90 +940,58 @@ func (s *Server) createFreePoolTenants(ctx context.Context, poolID string, count
 	return results, nil
 }
 
-func (s *Server) createFreeSharedPoolTenants(ctx context.Context, poolID string, count int, cred tenant.CredentialProvisionRequest, quotaOpt *quotaRequest) ([]*provisionTenantResult, error) {
+func (s *Server) createFreeSharedPoolTenants(ctx context.Context, poolID string, count int, _ tenant.CredentialProvisionRequest, quotaOpt *quotaRequest) ([]*provisionTenantResult, error) {
 	if count <= 0 {
 		return []*provisionTenantResult{}, nil
 	}
 	if _, ok := s.provisioner.(tenant.SharedDBPoolProvisioner); !ok {
 		return nil, fmt.Errorf("shared tenant pool provisioning not enabled")
 	}
+	sharedCred, err := s.sharedDBCloudCredentials()
+	if err != nil {
+		return nil, err
+	}
 	pool, err := s.meta.GetTenantPoolByID(ctx, poolID)
 	if err != nil {
 		return nil, err
 	}
+	organizationID := strings.TrimSpace(pool.OrganizationID)
+	if organizationID == "" {
+		return nil, fmt.Errorf("shared tenant pool organization is required")
+	}
 	now := time.Now().UTC()
-	createdPoolIDs := make(map[int64]struct{})
+	outcomes := runSharedPoolTenantCreateWorkers(ctx, count, func() sharedPoolTenantCreateOutcome {
+		return s.createFreeSharedPoolTenant(ctx, poolID, organizationID, sharedCred, quotaOpt, now)
+	})
 	continuationPoolIDs := make(map[int64]struct{})
 	results := make([]*provisionTenantResult, 0, count)
 	resultPoolIDs := make([]int64, 0, count)
-	for i := 0; i < count; i++ {
-		tenantID := token.NewID()
-		if err := s.insertPendingPoolTenant(ctx, tenantID, tenant.ProviderTiDBCloudNativeShared, now); err != nil {
-			return results, err
+	var createErr error
+	for _, outcome := range outcomes {
+		if outcome.err != nil {
+			createErr = errors.Join(createErr, outcome.err)
+			continue
 		}
-		fsID, err := s.meta.EnsureFsID(ctx, tenantID)
-		if err != nil {
-			_ = s.meta.UpdateTenantStatus(context.Background(), tenantID, meta.TenantFailed)
-			return results, err
+		results = append(results, outcome.result)
+		resultPoolIDs = append(resultPoolIDs, outcome.dbID)
+		if outcome.needsContinuation {
+			continuationPoolIDs[outcome.dbID] = struct{}{}
 		}
-		opts := provisionTenantOptions{Quota: quotaOpt}
-		if err := s.materializeSharedTenantQuota(ctx, tenantID, opts); err != nil {
-			_ = s.meta.UpdateTenantStatus(context.Background(), tenantID, meta.TenantFailed)
-			return results, err
-		}
-		var selected *meta.SharedDB
-		selected, created, err := s.allocateManagedSharedDB(ctx, cred, func(db *meta.SharedDB) error {
-			return s.meta.CompleteSharedTenantPoolMember(ctx, tenantID, tenant.ProviderTiDBCloudNativeShared,
-				&meta.TenantPlacement{FsID: fsID, DbID: db.ID, Placement: meta.PlacementShared,
-					SchemaShape: meta.SchemaShapeShared, Status: meta.SharedDBStatusActive},
-				&meta.TenantPoolMembership{TenantID: tenantID, TiDBCloudOrganizationID: pool.OrganizationID,
-					PoolID: poolID, PoolStatus: meta.TenantPoolBindingFree, CreatedAt: now, UpdatedAt: now})
-		})
-		if err != nil {
-			_ = s.meta.UpdateTenantStatus(context.Background(), tenantID, meta.TenantFailed)
-			return results, err
-		}
-		if created {
-			createdPoolIDs[selected.ID] = struct{}{}
-		}
-		if selected.Status == meta.SharedDBStatusPending || selected.Status == meta.SharedDBStatusProvisioning {
-			continuationPoolIDs[selected.ID] = struct{}{}
-		}
-		resultStatus := meta.TenantPending
-		switch selected.Status {
-		case meta.SharedDBStatusProvisioning:
-			resultStatus = meta.TenantProvisioning
-		case meta.SharedDBStatusActive:
-			resultStatus = meta.TenantActive
-		}
-		results = append(results, &provisionTenantResult{TenantID: tenantID,
-			Status: resultStatus, Provider: tenant.ProviderTiDBCloudNativeShared,
-			CloudProvider: selected.CloudProvider, Region: selected.Region,
-			OrganizationID: selected.TiDBCloudOrganizationID})
-		resultPoolIDs = append(resultPoolIDs, selected.ID)
 	}
-	createdIDs := make([]int64, 0, len(createdPoolIDs))
-	for id := range createdPoolIDs {
-		createdIDs = append(createdIDs, id)
+	if createErr != nil {
+		return results, createErr
 	}
-	sort.Slice(createdIDs, func(i, j int) bool { return createdIDs[i] < createdIDs[j] })
-	resolvedOrg, err := s.provisionManagedSharedDBPoolsBatch(ctx, createdIDs)
+	continuationIDs := make([]int64, 0, len(continuationPoolIDs))
+	for id := range continuationPoolIDs {
+		continuationIDs = append(continuationIDs, id)
+	}
+	sort.Slice(continuationIDs, func(i, j int) bool { return continuationIDs[i] < continuationIDs[j] })
+	_, err = s.provisionManagedSharedDBPoolsBatchWithCredentials(ctx, continuationIDs, sharedCred)
 	if err != nil {
 		return results, err
 	}
-	if pool.OrganizationID == "" && resolvedOrg != "" {
-		if err := s.meta.UpdateTenantPoolOrganization(ctx, poolID, resolvedOrg); err != nil {
-			return results, err
-		}
-		if err := s.meta.UpdateTenantPoolMembershipOrganization(ctx, poolID, resolvedOrg); err != nil {
-			return results, err
-		}
-		for _, result := range results {
-			result.OrganizationID = resolvedOrg
-		}
-	}
-	createdPoolStatuses := make(map[int64]meta.TenantStatus, len(createdIDs))
-	for _, dbID := range createdIDs {
+	continuationPoolStatuses := make(map[int64]meta.TenantStatus, len(continuationIDs))
+	for _, dbID := range continuationIDs {
 		sharedDB, err := s.meta.GetSharedDB(ctx, dbID)
 		if err != nil {
 			return results, err
@@ -1031,57 +1003,116 @@ func (s *Server) createFreeSharedPoolTenants(ctx context.Context, poolID string,
 		case meta.SharedDBStatusActive:
 			status = meta.TenantActive
 		}
-		createdPoolStatuses[dbID] = status
+		continuationPoolStatuses[dbID] = status
 	}
 	for i, dbID := range resultPoolIDs {
-		if status, ok := createdPoolStatuses[dbID]; ok {
+		if status, ok := continuationPoolStatuses[dbID]; ok {
 			results[i].Status = status
 		}
 	}
-	continuationIDs := make([]int64, 0, len(continuationPoolIDs))
-	for dbID := range continuationPoolIDs {
-		continuationIDs = append(continuationIDs, dbID)
-	}
 	s.scheduleManagedSharedDBContinuations(ctx, continuationIDs)
 	return results, nil
+}
+
+func runSharedPoolTenantCreateWorkers(ctx context.Context, count int, create func() sharedPoolTenantCreateOutcome) []sharedPoolTenantCreateOutcome {
+	outcomes := make([]sharedPoolTenantCreateOutcome, count)
+	jobs := make(chan int)
+	var wg sync.WaitGroup
+	for range min(count, sharedTenantPoolCreateConcurrency) {
+		wg.Add(1)
+		go func() {
+			defer wg.Done()
+			for i := range jobs {
+				if ctx.Err() != nil {
+					outcomes[i].err = ctx.Err()
+					continue
+				}
+				outcomes[i] = create()
+			}
+		}()
+	}
+	for i := range count {
+		jobs <- i
+	}
+	close(jobs)
+	wg.Wait()
+	return outcomes
+}
+
+type sharedPoolTenantCreateOutcome struct {
+	result            *provisionTenantResult
+	dbID              int64
+	needsContinuation bool
+	err               error
+}
+
+func (s *Server) createFreeSharedPoolTenant(ctx context.Context, poolID, organizationID string, sharedCred tenant.CredentialProvisionRequest, quotaOpt *quotaRequest, now time.Time) sharedPoolTenantCreateOutcome {
+	tenantID := token.NewID()
+	fail := func(err error) sharedPoolTenantCreateOutcome {
+		_ = s.meta.UpdateTenantStatus(context.Background(), tenantID, meta.TenantFailed)
+		return sharedPoolTenantCreateOutcome{err: err}
+	}
+	if err := s.insertPendingPoolTenant(ctx, tenantID, tenant.ProviderTiDBCloudNativeShared, now); err != nil {
+		return sharedPoolTenantCreateOutcome{err: err}
+	}
+	fsID, err := s.meta.EnsureFsID(ctx, tenantID)
+	if err != nil {
+		return fail(err)
+	}
+	if err := s.materializeSharedTenantQuota(ctx, tenantID, provisionTenantOptions{Quota: quotaOpt}); err != nil {
+		return fail(err)
+	}
+	selected, _, err := s.allocateManagedSharedDBForOrganization(ctx, organizationID, sharedCred, func(db *meta.SharedDB) error {
+		return s.meta.CompleteSharedTenantPoolMember(ctx, tenantID, tenant.ProviderTiDBCloudNativeShared,
+			&meta.TenantPlacement{FsID: fsID, DbID: db.ID, Placement: meta.PlacementShared,
+				SchemaShape: meta.SchemaShapeShared, Status: meta.SharedDBStatusActive},
+			&meta.TenantPoolMembership{TenantID: tenantID, TiDBCloudOrganizationID: organizationID,
+				PoolID: poolID, PoolStatus: meta.TenantPoolBindingFree, CreatedAt: now, UpdatedAt: now})
+	})
+	if err != nil {
+		return fail(err)
+	}
+	resultStatus := meta.TenantPending
+	switch selected.Status {
+	case meta.SharedDBStatusProvisioning:
+		resultStatus = meta.TenantProvisioning
+	case meta.SharedDBStatusActive:
+		resultStatus = meta.TenantActive
+	}
+	return sharedPoolTenantCreateOutcome{
+		result: &provisionTenantResult{TenantID: tenantID, Status: resultStatus,
+			Provider: tenant.ProviderTiDBCloudNativeShared, CloudProvider: selected.CloudProvider,
+			Region: selected.Region, OrganizationID: selected.TiDBCloudOrganizationID},
+		dbID:              selected.ID,
+		needsContinuation: selected.Status == meta.SharedDBStatusPending || selected.Status == meta.SharedDBStatusProvisioning,
+	}
 }
 
 func (s *Server) provisionManagedSharedDBPoolsBatch(ctx context.Context, dbIDs []int64) (string, error) {
 	if len(dbIDs) == 0 {
 		return "", nil
 	}
-	provisioner := s.provisioner.(tenant.SharedDBPoolProvisioner)
-	resolvedOrg := ""
-	for start := 0; start < len(dbIDs); start += 10 {
-		end := start + 10
-		if end > len(dbIDs) {
-			end = len(dbIDs)
-		}
-		chunkOrg, err := s.provisionManagedSharedDBPoolsBatchChunk(ctx, provisioner, dbIDs[start:end])
-		if err != nil {
-			return resolvedOrg, err
-		}
-		if resolvedOrg == "" {
-			resolvedOrg = chunkOrg
-		}
+	cred, err := s.sharedDBCloudCredentials()
+	if err != nil {
+		return "", err
 	}
-	return resolvedOrg, nil
+	return s.provisionManagedSharedDBPoolsBatchWithCredentials(ctx, dbIDs, cred)
 }
 
-func (s *Server) provisionManagedSharedDBPoolsBatchChunk(ctx context.Context, provisioner tenant.SharedDBPoolProvisioner, dbIDs []int64) (string, error) {
+func (s *Server) provisionManagedSharedDBPoolsBatchWithCredentials(ctx context.Context, dbIDs []int64, cred tenant.CredentialProvisionRequest) (string, error) {
+	if len(dbIDs) == 0 {
+		return "", nil
+	}
+	provisioner := s.provisioner.(tenant.SharedDBPoolProvisioner)
 	for attempt := 0; attempt < 2; attempt++ {
 		first, err := s.meta.GetSharedDB(ctx, dbIDs[0])
-		if err != nil {
-			return "", err
-		}
-		cred, err := s.sharedDBCloudCredentials()
 		if err != nil {
 			return "", err
 		}
 		identity := sharedDBAllocationIdentity(first.TiDBCloudOrganizationID, first.ProvisioningKey)
 		restartWithOrganization := false
 		resolvedOrg := ""
-		err = s.meta.WithSharedDBAllocationLock(ctx, identity, func(lockCtx context.Context) error {
+		err = s.withSharedDBAllocationLock(ctx, identity, func(lockCtx context.Context) error {
 			current, err := s.meta.GetSharedDB(lockCtx, dbIDs[0])
 			if err != nil {
 				return err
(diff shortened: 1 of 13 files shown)


srstack:
LGTM. Reviewed the full diff and ran the concurrency-sensitive tests under `-race` against the branch head. This parallelizes shared-pool refill (50-worker tenant pipeline, 10-wide Cloud batch waves, 100-worker continuation pool) and the locking model holds up. (Posting as a comment — GitHub blocks approving one's own PR. CI still running at review time.)

## Verified under -race
Ran against the branch in a worktree, all green:
- `TestSharedTenantPoolRefillRunsFiftyTenantWorkers`, `...LimitsConcurrentPoolsToOneHundred`, `...StartsEveryPoolConcurrently`, `TestRetryTenantPoolClaimCASReturnsMissAfterLimit` — pure-goroutine bounds.
- `TestSharedTenantPoolRefill*`, `TestManagedSharedDBBatchCreateRunsAllChunksConcurrently` (101 pools → 11 batches sized `[1,10×10]`), `...UsesAllocationLock`, `...BatchesDoNotEnterBlockingMetadataWait`, `...ReturnsTotalFailure` — the DB-backed paths that actually stress the connection pool at `MAX_OPEN_CONNS=8`.
- Broader `TestProvisionTiDBCloudNativeShared*` / `TestManagedSharedDB*` / `TestRetryManagedSharedDBContinuation*` — no regression in the sequential paths.

No data races, no deadlock, correct batch fan-out.

## Correctness notes
**`retryTenantPoolClaimCAS` now returns `(zero, nil)` on exhaustion** — the one behavioral change worth calling out. Verified the sole caller (`admin_tenant_pool.go:1976`) handles it: the result is a value type `tenantPoolClaimSelection{}`, so `sharedResult == nil` and `native == nil` both fall through to the "claim_free_tenant_missed" path and it retries via active provisioning. No nil-deref, matches the stated intent (exhausted CAS = normal miss, not a hard error).

**Two-tier allocation locking is sound.** Org-level identity (`org:X`) still serializes *physical pool creation* (prevents duplicate Cloud pools); once a pool has a cluster_id, continuations switch to a per-pool identity (`org:X:db_pool:N`) so metadata/schema work advances concurrently across pools. The row-level `FOR UPDATE` transactions from #786 remain the ultimate correctness guard, so relaxing the in-process lock scope doesn't risk half-ready reads corrupting state — the comment update on `continueManagedSharedDBPoolLocked` documents this well.

**The new in-process `withSharedDBAllocationLock` (local sync.Mutex over the meta GET_LOCK) is the right fix for the connection-exhaustion hazard.** MySQL `GET_LOCK` acquires on a fresh pooled `*sql.Conn` each call, so N same-identity waiters would otherwise pin N idle connections blocking on the lock. The local mutex serializes them *before* they take a DB connection, which is exactly why the refill fast-path can run 50 workers against an 8-connection pool without starving. The `testmain` idle-conn bump (0→1 per pool) is a related CI-stability tweak (retain one idle conn to avoid TCP port churn) and is test-only.

**Refill fast-path consumes existing capacity without the planning lock or an IAM call** (`allocateManagedSharedDBForOrganization`) — `TestSharedTenantPoolRefillUsesExistingCapacityWithoutIdentityLookup` confirms `iamCalls == 0` and `sharedPoolBatchCalls == 0` when capacity exists and org is known. Good — this is the hot path and it avoids both the lock and the identity round-trip.

## Nits (non-blocking)
- `runSharedPoolTenantCreateWorkers`: on `ctx.Err()` a worker sets `outcomes[i].err` and `continue`s, so remaining queued jobs still drain one-by-one (each immediately hitting the same ctx check). Correct, just slightly wasteful vs. draining the channel on cancel — negligible.
- The fixed 15s `managedSharedDBContinuationInterval` replaces the old exponential backoff. Simpler and fine for a bounded 100-worker pool, but a very large pending backlog now polls every pool every 15s with no jitter — if the fleet ever grows past the 100-continuation limit consider light jitter to avoid synchronized Cloud API waves. Not a concern at current scale.

Solid parallelization with the connection-pool hazard properly understood and guarded. Nice work on the test coverage — the bounded-concurrency tests (Nth worker must not start while N-1 in flight) are exactly the right shape.

srstack:
Re-reviewed after `20034b3f fix: batch shared pool continuation workers`. This is a large follow-up (now ~1505/-460, 10 files) that **rewrites the continuation machinery**, not a small tweak. Read the whole thing and ran the concurrency-sensitive paths under `-race`. Holds up — LGTM with one cleanup ask.

## What changed
The round-based `retryManagedSharedDBContinuations` batcher is replaced by an explicit three-phase pipeline in `runManagedSharedDBContinuations`:
1. **Physical create** for pending pools with no cluster_id (batch, org-locked).
2. **Metadata poll** for pending pools *with* a cluster_id — new `pollManagedSharedDBMetadataWithReady` fans out over a new `SharedDBPoolBatchLoader.BatchLoadSharedDBPoolsWithCredentials` (one TiDB Cloud list per batch of ~30), persists ready rows, and hands them to phase 3 via an `onReady` callback.
3. **Schema-init** via `runManagedSharedDBProvisioningQueue` — a worker-pool queue with per-job exponential backoff + requeue.
Plus: env-configurable knobs (`DRIVE9_TIDBCLOUD_NATIVE_SHARED_*` for cloud-batch size, metadata workers/batch/poll-interval, provisioning workers) with exported defaults, global slot channels (`managedSharedDBMetadataSlots`/`ProvisioningSlots`) to bound total concurrency, and `sync.Map` dedup (`managedSharedDBMetadataJobs`/`ProvisioningJobs`) so the scheduled path and the resume ticker never double-process a pool.

## Verified under -race (all green, in a worktree)
- New tests: `TestManagedSharedDBWorkerConfigDefaultsAndOverrides`, `TestRunManagedSharedDBProvisioningQueueRequeuesWithoutBlockingWorker`, `TestManagedSharedDBMetadataWorkerRefillsReadySlots`, `TestManagedSharedDBContinuationDoesNotStartOnFollower`, `TestRunManagedSharedDBProvisioningJobsHonorsWorkerLimit`.
- Broad suite: everything matching `ManagedSharedDB|SharedTenantPool|TestProvisionTiDBCloudNativeShared` — no races, no deadlock.
- `BatchLoad*` provisioner tests + `go vet ./pkg/server` clean.

## Correctness notes
- **Queue requeue can't deadlock.** `runManagedSharedDBProvisioningQueue`'s `jobs`/`retries` channels are buffered to `len(dbIDs)`, retries are scheduled on their own timer goroutines that select on `ctx.Done()`, and the main loop drains `results`/`retries` while workers block only on slot acquisition (also ctx-aware). Requeue happens off-worker, so a retrying job never pins a worker — the dedicated test confirms this.
- **Dedup ownership is sound.** `LoadOrStore` on the job maps means whichever of {scheduled, resume-ticker} claims a dbID first owns it; the other skips. Metadata refill (`take`) re-checks `LoadOrStore` before adopting freshly-listed rows, so the 15s ticker overlapping a slow poll can't double-drive a pool.
- **Bounded workers by design.** `workerCount` is fixed from the initial `owned` size; `take` may grow `owned` during refill but only the pre-spawned workers drain it. `owned` is appended under `queueMu` and only read in the defer after `wg.Wait()`, so no race. Fine.
- **Leader gating**: `startManagedSharedDBWorker` correctly no-ops on a follower (`TestManagedSharedDBContinuationDoesNotStartOnFollower`), and the resume loops moved to their own tickers with proper `wg.Wait()` on shutdown.
- Config clamps are sensible: cloud-batch/metadata sizes are `min`-capped at their defaults (can't be raised past the safe ceiling, only lowered), provisioning workers and poll interval are floored to defaults when non-positive.

## Cleanup ask (non-blocking)
- **`runManagedSharedDBProvisioningJobs` is dead code** — it has zero non-test callers (only `TestRunManagedSharedDBProvisioningJobsHonorsWorkerLimit` exercises it). Looks like an earlier iteration superseded by `runManagedSharedDBProvisioningQueue`. Please delete both the function and its test, or wire it in if it was meant to replace something. `go vet` doesn't flag it but the linter (`make lint`/unused) may.

## Nits (non-blocking)
- The metadata poll and physical-create phases both call `sharedDBCloudCredentials()` independently; minor, but you could thread the cred once through `runManagedSharedDBContinuations`.
- `pollManagedSharedDBMetadataWithReady` lists `ListSharedDBsByStatus(..., 1000)` under `queueMu` on refill — the lock is held across a DB round-trip. Only one worker refills at a time and contention is low, but if metadata workers ever scale up you'd serialize them on that query. Fine at 15 workers.

Once the dead `runManagedSharedDBProvisioningJobs` is removed this is clean. Impressive test coverage on the bounded-concurrency invariants. CI still running at re-review time.

srstack:
Review follow-up pushed in 023f47db.

Handled:
- Persist every returned Cloud result even when another result is invalid or persistence fails; return errors.Join(createErr, persistErr) after the full loop.
- Replace the unchecked SharedDBPoolProvisioner assertion with an error.
- Make local allocation-lock waiting context-aware and delete unused per-identity entries safely.
- Use the per-pool lock identity for known clusters.
- Prevent overlapping leader resume passes.
- Remove the unused runManagedSharedDBProvisioningJobs helper and its isolated test.
- Document the configured safety ceilings for Cloud and metadata batch controls.

Not changed:
- Reusing the first SharedDB reads and threading credentials through phases are non-blocking query-count optimizations.
- Moving ListSharedDBsByStatus outside queueMu as suggested can make all metadata workers issue the same refill scan concurrently; that needs a separate refill ownership design, not a lock-only change.

Validation:
- full pkg/server suite
- targeted pkg/server race tests
- pkg/tenant/tidbcloudnative and cmd/drive9-server
- make lint

srstack:
Follow-up optimization pushed in 4578a723.

The metadata refill query no longer runs while queueMu is held. A context-aware single-slot refill gate ensures at most one worker executes ListSharedDBsByStatus at a time. After acquiring the refill slot, the worker rechecks the queue before querying, so a waiter uses rows filled by the previous worker instead of issuing the same scan. queueMu now protects only in-memory queue, known, and owned updates.

Added a regression test for refill serialization, cancellation, and slot release. Full pkg/server, targeted race tests, related packages, and lint pass.

srstack:
Re-reviewed the six new commits (`4e25b9ae`…`4578a723`). These address my prior feedback and add sensible refinements. **CI is now green (ci + e2e smoke), and I re-ran the concurrency paths under `-race` — all pass.** LGTM.

## My prior feedback is resolved
- **Dead `runManagedSharedDBProvisioningJobs` removed** (`023f47db`). Confirmed no dangling refs to it (or to the earlier `tenantPoolRefillWaveLimit` / `managedSharedDBContinuationInterval` / `...Limit` symbols); `go build ./...` clean.
- **Metadata refill no longer holds `queueMu` across the DB round-trip** (`4578a723`). The `ListSharedDBsByStatus` call now runs outside the queue lock, gated by a dedicated cap-1 `refillSlot` that serializes refills and respects `ctx.Done()`. Fast path (queue already ≥ limit) stays lock-cheap; the double-check after acquiring the slot avoids a redundant list when another worker just refilled. `TestManagedSharedDBMetadataRefillSlotSerializesAndRespectsContext` covers it.

## New changes reviewed
- **Ref-counted allocation lock** (`sharedDBAllocationLocalLock`, `023f47db`) replaces the raw `sync.Map`-of-`Mutex`. This fixes unbounded map growth: entries `CompareAndDelete` themselves at `refs==0`. I traced the acquire/release balance — the `ctx.Done()` branch releases with `held=false` (never drains the semaphore it didn't acquire) while the success path releases with `held=true`; `refs` is mutated under `mu` and the CAS-delete prevents a torn entry. `TestSharedDBAllocationLockWaitRespectsContextAndCleansUp` passes under `-race`.
- **Queue shutdown via `done` channel + `stopOnce`** (`057bc19f`) instead of `close(jobs)`. This is the right fix — with retries feeding back into `jobs`, closing `jobs` directly could race a retry send onto a closed channel. Now workers select on `done`/`ctx.Done()`/`jobs`, and the retry timer goroutines also select on `done`, so nothing sends after stop.
- **Worker cooldown (10s) + capped backoff (15s)** (`8faa2c75`,`057bc19f`) pace schema-init. Cooldown fires between jobs and is skipped for the first job per worker; it doesn't hold a provisioning slot while sleeping, so it paces throughput without wasting capacity. `TestManagedSharedDBProvisioningBackoffCapsAtFifteenSeconds` covers the cap.
- **Refill wave capped by `maxTenants × poolLimit`** (`4e25b9ae`,`0e46bb3d`) replacing the flat 50-tenant cap. The overflow guard (`maxTenants <= maxInt/poolLimit`) is correct — avoids int overflow before multiplying, falls back to unbounded (`missing`) only when the product would overflow, which is safe since `missing` is the real ceiling. Only applied for the shared provider.
- **Partial-failure resilience in batch chunk** (`023f47db`): unknown-pool / missing-org / persist errors now `errors.Join` and continue rather than aborting the whole chunk, and the `persisted` count is accurate. Better — one bad row in a create wave no longer discards the good rows' metadata.
- **Resume-loop dedup** (`023f47db`): the two near-identical ticker goroutines collapsed into `runManagedSharedDBResumeLoop`. Pure cleanup, same behavior.
- Config: `CloudBatchSize`/`MetadataWorkers`/`MetadataBatchSize` remain `min`-capped at defaults; `RefillPoolLimit`/`ProvisioningWorkers` accept larger values, and the comment now documents which are safety-capped vs. free. Sensible split.

## Verified under -race (latest head, worktree)
`ManagedSharedDB*`, `SharedTenantPool*`, `TestProvisionTiDBCloudNativeShared*`, `RefillTenantCount`, `AllocationLock*` — all green. `go build ./...` clean.

No blocking issues, nothing outstanding from my earlier passes. This is solid — the lock lifecycle and queue-shutdown reworks are exactly the kind of correctness hardening this concurrent path needed, and the test coverage on the invariants (context-cancel cleanup, refill serialization, backoff cap, no-block requeue) is thorough. Nice.

srstack:
Re-reviewed the new commit `6f0a53bc perf: coalesce tenant pool replenishment`. Small, well-contained perf change. LGTM. (CI re-running at review time; I verified under `-race` locally.)

## What it does
Three things, all sound:

1. **Per-pool replenishment dedup gate.** New `tenantPoolReplenishJobs` sync.Map + `beginTenantPoolReplenishment`/`finishTenantPoolReplenishment`. Concurrent triggers for the same pool now coalesce — a second trigger while one is in flight returns early instead of stacking another worker. `TestTenantPoolReplenishmentCoalescesConcurrentTriggersByPool` covers it and passes under `-race`.

   **Gate balance verified — no leaked markers.** `begin` returns true only when it stored the marker; if `startServerWorker` returns true the worker's `defer finishTenantPoolReplenishment` deletes it exactly once; if `startServerWorker` returns false (server closing) the outer `if !...{ finish }` deletes it. `startServerWorker` returns false only *before* spawning the goroutine (under `forkWorkerMu` when `forkWorkerClosed`) and true only after `WG.Add(1)` guarantees launch — so exactly one release per acquire, and `Delete` is idempotent regardless.

2. **Skip the redundant free-count query on the hot path.** Previously `defer refreshTenantPoolCapacity` ran unconditionally and did its own `CountFreeTenantPoolBindings`. Now the already-full / below-watermark / noop paths call `recordTenantPoolCapacity` directly with the `freeSize` already counted, and the full re-query `refreshTenantPoolCapacity` only defers after an actual provisioning attempt (where the count changed). Same metric semantics, one fewer DB round-trip on the common "pool already full" path — which, given this runs on every claim, is the meaningful win.

3. **`JOIN` → `STRAIGHT_JOIN` in `CountFreeTenantPoolMemberships`** (SQL extracted to a package const). Forces the membership table (filtered by org + pool_status, highly selective) to be read before joining `tenants`, rather than letting the optimizer potentially lead with a provider/status scan of `tenants`. `TestFreeTenantPoolMembershipCountPinsMembershipTableFirst` asserts the plan and passes.

## Verified
- `-race`: `TestTenantPoolReplenishmentCoalescesConcurrentTriggersByPool`, `*Replenish*`, `TestManagedSharedDBRefillTenantCount`, `TestProvisionTiDBCloudNativeShared*`, `TestSharedTenantPoolRefill*` — all green.
- `pkg/meta`: the new plan test + `CountFreeTenantPoolMemberships` — green.
- `go build ./...` clean.

## One nit (non-blocking)
`STRAIGHT_JOIN` hard-codes join order, so it's a bet that `tenant_pool_memberships` filtered by (org, pool_status) is always the more selective side. That's almost certainly true here (one org's free memberships « all tenants of a provider), but if the data distribution ever inverts — e.g. an org with a huge free-membership backlog vs. few total shared tenants — the pinned order could hurt. The test locks in the *plan*, not the *performance*, so a future distribution shift wouldn't be caught. Worth a one-line comment on the const noting why the order is pinned. Not blocking.

Nothing outstanding from any of my earlier passes. Ship it once CI is green.

srstack:
Re-reviewed `4ec70928 docs` + `6ea3ec39 perf: throttle tenant pool replenishment checks`. LGTM. Verified under `-race` and confirmed the merged query returns identical results.

## `4ec70928` — resolves my prior nit
The one-line comment on `countFreeTenantPoolMembershipsSQL` explaining why the join order is pinned. Exactly what I asked for. 👍

## `6ea3ec39` — throttle + query coalescing
Four changes, all sound:

1. **Merged the two free-count queries into one `UNION ALL`** (`countFreeTenantPoolBindingsByStatus`). Previously a native-bindings query + a separate `CountFreeTenantPoolMemberships` query summed in Go; now one round-trip with `COALESCE(SUM(slot_count), 0)`. **Arg ordering verified correct**: first subquery gets `(org, free, tidbCloudNativeProvider, …statuses)`, second gets `(org, free, tidbCloudNativeSharedProvider, …statuses)`, and both `%s` are filled with `placeholders` — matches the two subqueries exactly. `pkg/meta` count tests pass, so the merged query returns the same total as the old sum. Both subqueries use the same `STRAIGHT_JOIN` pin as the earlier change.

2. **Double-checked locking in `replenishTenantPoolAsync`.** The cheap path (get pool → count free → watermark check) now runs *before* the distributed `WithTenantPoolLock`, and only a sub-watermark result takes the lock and re-checks everything. The common "pool is full" case no longer pays for the distributed lock — the meaningful win, since this fires on every claim. The re-check under lock is intact and commented.

3. **Work gate upgraded to `*tenantPoolWorkGate` with a min-interval throttle** (1s replenish, 15s after an empty resume scan), plus a new `tenantPoolResumeScans` gate and a rerun flag on in-flight resume jobs. This stops repeated claims from hammering replenish/resume checks. Gate acquire/release balance holds: `begin` sets `running=true` only when not running and past `nextAllowed`; `finish` (via worker defer, or the outer `if !startServerWorker`) always clears `running` and stamps `nextAllowed`. Tests `TestTenantPoolReplenishmentCoalescesConcurrentTriggersByPool` and `TestTenantPoolPendingResumeScanCooldownAfterEmptyResult` pass under `-race`.

4. **`sharedTenantPoolCreateConcurrency` 50 → 10** and **`refreshTenantPoolCapacity` removed.** Lower peak refill fan-out (consistent with the connection-pool constraints from earlier passes) and the post-provision capacity re-query is gone. Removed cleanly — no dangling refs, `go build ./...` clean.

## Two behavior notes (non-blocking)
- **The work gate is no longer deleted from the map on finish** — it went from self-cleaning (`Delete`) to permanent retention (updates `running`/`nextAllowed` in place). Necessary for the throttle to remember `nextAllowed` across triggers, so it's the right call — but it does mean `tenantPoolReplenishJobs` and `tenantPoolResumeScans` now retain one entry per distinct pool ID for the server's lifetime. Bounded by pool count (not per-request), so fine in practice; just flagging the lifecycle change.
- **Capacity metric is now slightly staler.** With `refreshTenantPoolCapacity` gone, `tenant_pool_capacity{free}` is recorded from the pre-provision count during replenish checks and no longer re-queried right after provisioning or on every claim. The throttle means it updates at most ~1/sec per pool. That's the intent (fewer COUNT queries), and the next check corrects it — acceptable, just noting the dashboards will lag provisioning by up to the poll/throttle interval.

## Verified
- `-race`: coalesce, resume-cooldown, watermark-skip, batch-below-watermark, plus broad `Replenish|TenantPoolClaim|TestProvisionTiDBCloudNativeShared|SharedTenantPool` — all green.
- `pkg/meta`: `CountFreeTenantPool*` / `FreeTenantPoolMembership*` / `CountTenantPoolFreeSlots` — green (merged query result matches).
- `go build ./...` clean.

(One `pkg/meta` run flaked on a testcontainer port collision — infra, not the code; passed on retry.)

Nothing blocking. Both my earlier nits are now closed. Ship once CI is green.

pkg/server/shared_db_pool.go
```diff
@@ -241,152 +321,504 @@ func (s *Server) scheduleManagedSharedDBContinuation(ctx context.Context, dbID i
...
 }
 
-type managedSharedDBContinuation struct {
-	ctx          context.Context
-	continuePool func() error
-	done         bool
-	lastErr      error
+func (s *Server) startManagedSharedDBWorker(ctx context.Context, fn func(context.Context)) bool {
+	if s.leader == nil {
+		return s.startServerWorker(ctx, fn)
```
mornyx, line 348: **Medium — non-leader pods silently drop scheduled continuations**

When `startManagedSharedDBWorker` returns false (not leader, or workers not started), `scheduleManagedSharedDBContinuation(s)` returns without any log. Behaviorally fine — the leader's 15s reconciler is the backstop — but during incident analysis a pool sitting in `pending`/`provisioning` for 15s with *zero* trace of why looks like a stuck continuation. A one-line Info log on the drop (`reason=not_leader`) would make the handoff visible.

pkg/server/admin_tenant_pool.go
```diff
@@ -936,90 +948,58 @@ func (s *Server) createFreePoolTenants(ctx context.Context, poolID string, count
...
 		return nil, err
 	}
+	organizationID := strings.TrimSpace(pool.OrganizationID)
+	if organizationID == "" {
+		return nil, fmt.Errorf("shared tenant pool organization is required")
+	}
 	now := time.Now().UTC()
-	createdPoolIDs := make(map[int64]struct{})
+	outcomes := runSharedPoolTenantCreateWorkers(ctx, count, func() sharedPoolTenantCreateOutcome {
+		return s.createFreeSharedPoolTenant(ctx, poolID, organizationID, sharedCred, quotaOpt, now)
```
mornyx, line 972: **Medium (pre-existing policy, now at larger scale) — wave creation is all-or-nothing**

`createFreeSharedPoolTenants` returns partial results plus a joined error, and the admin create path treats any error as total failure: it cleans up *all* successful members and deletes the logical pool, so the client's retry rebuilds everything. That policy predates this PR, but the parallel pipeline runs much larger waves (up to 5000 tenants / 50 pools per refill), so the cost of a single transient per-tenant failure is now discarding hundreds of good placements. Safe, but wasteful — consider returning per-tenant outcomes and keeping successes (the pool's refill loop naturally tops up the shortfall next wave), or at least retrying per-tenant allocate errors once before joining. Non-blocking.

mornyx (approved):
Reviewed at `46fed62` (12 commits, +2173/−509, 13 files; CI green, local build passes).

This PR attacks exactly the throughput bottleneck flagged in the dev stress-test analysis (sequential continuations + GET_LOCK waiter pile-up capping provision rate), and the approach is sound where it matters:

- **`sharedDBAllocationLocalLock`** — the process-local per-identity semaphore in front of `GET_LOCK` is the right fix for meta-connection exhaustion from waiter pile-up. The `refs`/`CompareAndDelete` lifecycle under `mu` (confirm-not-deleted before `refs++`, delete only at `refs==0`) is race-free, and lock ordering (local → named DB lock) is uniform, so no ABBA.
- **Lock-free fast path** in `allocateManagedSharedDBForOrganization` — consuming existing capacity without the org lock is correct because `CompleteSharedTenant*`'s conditional `UPDATE` is the real serialization point; the lock is only for *planning a new pool*. This is the big unlock.
- **Dedup discipline** — per-pool in-flight markers for both metadata poll (`managedSharedDBMetadataJobs`) and provisioning (`managedSharedDBProvisioningJobs`) coalesce duplicate triggers; the replenish gate (`tenantPoolWorkGate`) + 1s min-interval + fast no-op path before locks stops claim-driven replenish storms.
- **Bounded fan-out everywhere** — worker pools have caps, per-worker cooldown, backoff ceiling, deadlines, and slot semaphores; the retry-queue coordinator (`jobs`/`results`/`retries` + `done`) terminates cleanly with no leaked goroutines.
- **Cross-pod safety preserved** — pool planning still goes through `GET_LOCK`; refill's provisioning section still takes the pool lock; leader-only continuations with the 15s reconciler as backstop.
- **`BatchLoadSharedDBPoolsWithCredentials`** lists by explicit cluster IDs — neatly avoids the >100-clusters list-cap issue for known pools.

Two medium notes inline (silent non-leader drop; wave-creation all-or-nothing policy at larger scale), plus one wording nit: the PR body says "fixed 50-worker pipeline" but the actual constants are 30 tenant-create workers / 50 pools per wave / 100 provisioning workers / 15 metadata workers — worth aligning for the runbook.

Stress-plan implication: re-estimate the sustainable rate after this merges — existing-capacity consumption is now lock-free, and the bottleneck moves to Cloud `batchCreate` wave latency and schema-init throughput (100 workers × per-pool DDL time). The V1 acceptance scenario gets much more plausible.

Approving.

qiffang (commented):
## PR #790 Review: Parallelize Managed Shared Pool Refill

**Reviewer**: dat9-dev1 (AI agent)
**Scope**: 13 files, +2202/-510 lines across 13 commits.

---

### Architecture Summary

The PR restructures the shared pool lifecycle into a concurrent pipeline:

1. **Tenant creation** — fans out to 30 goroutines via `runSharedPoolTenantCreateWorkers`
2. **Physical cluster batch create** — splits into chunks of `CloudBatchSize` (default 10, safety-capped), runs concurrently
3. **Metadata polling** — up to 15 workers, each polling batches of 30 pools, with serialized DB re-scans via `refillSlot`
4. **Schema init** — 100-worker pool with exponential backoff (max 15s) and 10s inter-job cooldown
5. **Refill throttling** — `tenantPoolWorkGate` coalesces triggers with 1s min interval
6. **Resume loop** — periodic polling replaces one-shot resume

---

### Findings

#### Finding 1 (Medium): `retryTenantPoolClaimCAS` silently swallows exhaustion

After 8 `ErrNotFound` attempts, now returns `(zero, nil)` instead of an error. Callers check the zero value today, but the generic `[T any]` signature means future callers could misinterpret zero-with-nil as success. Consider returning a sentinel like `ErrPoolExhausted`.

#### Finding 2 (Medium): Concurrent cleanup uses `context.Background()`

All 30 `createFreeSharedPoolTenant` workers use `context.Background()` for `UpdateTenantStatus(Failed)`. During Cloud outages with mass failures, this could spike meta DB load with 30 parallel UPDATEs. Low practical risk per call, but no backpressure.

#### Finding 3 (Medium): Batch create error accumulation — correct behavior

`provisionManagedSharedDBPoolsBatchChunkLocked` now accumulates errors with `errors.Join` instead of aborting on first persist error. Successfully persisted pools survive in DB and the continuation pipeline picks them up. This is the right behavior for parallelization.

#### Finding 4 (Medium): `managedSharedDBRefillTenantCount` overflow guard

The overflow check is correct. Normal cap = `maxTenants * poolLimit` = 100 × 50 = 5000 tenants per wave. The `maxInt` fallback only triggers with pathologically large values. Defensive and fine.

#### Finding 5 (Low): Capacity gauge no longer updates on claim

`refreshTenantPoolCapacity` was removed. Capacity metrics now only update during refill, not on every claim. Observability gap but not functional — gauge will lag if claims outpace refill triggers.

#### Finding 6 (Low): `STRAIGHT_JOIN` hint — good change

Pins join order so the selective organization filter drives the scan. Test `TestFreeTenantPoolMembershipCountPinsMembershipTableFirst` validates the SQL text. Good regression guard.

---

### Positive observations

1. **Well-structured worker pools** — clean dedup, bounded concurrency, exponential backoff with deadline, proper shutdown
2. **Safety caps** — CloudBatchSize/MetadataWorkers/MetadataBatchSize capped at defaults; RefillPoolLimit/ProvisioningWorkers accept larger values
3. **Fast-path optimization** — `allocateManagedSharedDBForOrganization` tries existing capacity WITHOUT organization lock
4. **Refill coalescing** — one replenishment per pool at a time, 1s cooldown
5. **22+ new test functions** — worker config, follower suppression, resume overlap, backoff/cooldown, batch create concurrency, refill caps
6. **`sharedDBAllocationLocalLock`** — semaphore + ref-counting correctly handles concurrent acquisition and cleanup

---

### Verdict

**No blockers.** Parallelization architecture is sound, concurrency controls are well-designed, test coverage is thorough. All findings are medium/low severity with no correctness risks.

qiffang:
## adversary-2 review at merged `546ee3d`

Focus per qiffang's ask: refill responsiveness + boundary cases.

### Blockers

**B1: CAS retry tight-loop → cold provision under contention**
`pkg/server/admin_tenant_pool.go:84` — `retryTenantPoolClaimCAS` loops 8× with **zero backoff**. Under >8 concurrent claimers, losers exhaust retries and fall through to fresh `provisionTenant`, creating orphan clusters while free pool members exist. Violates warm-pool semantics.

**B2: `ErrSharedDBQuotaNotMaterialized` bypasses CAS retry**
`pkg/meta/tenant_pool_membership.go:359` — Only `ErrNotFound` triggers retry; quota-materialization race becomes HTTP 502.

**B3: Cross-org batch allocation lock broken**
`pkg/server/admin_tenant_pool.go:1123`, `pkg/server/shared_db_pool.go:381` — `provisionManagedSharedDBPoolsBatchWithCredentials` derives lock from `dbIDs[0]` only, but caller passes multi-org IDs from `ListSharedDBsByStatus`. Result: unlocked concurrent creates for every non-first org.

**B4: Metadata pipeline collapses to single worker for typical loads**
`workerCount = min(15, ceil(n/30))` — for ≤30 pending pools (common case), only **one** goroutine runs 15s polls. Any Cloud stall blocks all pools serially. Advertised "100-worker" only kicks in above 3000 pending pools.

**B5: Doc/code mismatch — 30 vs 50 workers**
`sharedTenantPoolCreateConcurrency = 30`; PR body says "50-worker pipeline". `TestSharedTenantPoolRefillLimitsTenantWorkersToThirty` enforces 30. **40% throughput shortfall vs. advertised.**

### Secondary

**B6**: `pkg/server/shared_db_pool.go:786` — in-place slice aliasing `remaining := group[:0]` is a latent trap if BatchLoad's ordering changes.
**B7**: `pkg/server/admin_tenant_pool.go:1197` — loop-var capture in goroutine; verify go.mod ≥1.22.
**B8**: `pkg/server/admin_tenant_pool.go:2043` — replenishment gate has no rerun flag; overlapping triggers silently dropped.
**N1**: Tests `TestSharedTenantPoolRefillRunsFiftyTenantWorkers`, `TestRetryManagedSharedDBContinuationsStartsEveryPoolConcurrently`, `TestRetryManagedSharedDBContinuationsLimitsConcurrentPoolsToOneHundred` referenced in PR body but not in merged tree.

### Backward compat

No schema/index changes. Safe for base binary rollback. TestMain idle conns 0→1 is test-only. ✓

### Verified OK

- `ClaimSharedTenantPoolMembership FOR UPDATE` — CAS-safe
- 15s max backoff × 10-min window — no retry storm
- Server-owned shared credentials skip IAM (test-verified)

---

Note @dat9-dev1's review reaches different conclusions ("no blockers") — B1/B3/B4/B5 above are substantive and should be verified before treating this as safely merged.

qiffang:
## adversary-2 amendments after @dat9-dev1 rebuttal

Verified against `546ee3d4`. Retracting B1, B2, B7:

- **B1 retracted**: `retryTenantPoolClaimCAS` only retries on `meta.ErrNotFound`. On empty pool `attempt()` returns `(zero, nil)`, loop exits after 1 iteration. Cold provision fallthrough doesn't happen.
- **B2 retracted**: `ErrSharedDBQuotaNotMaterialized` branches (`sql.ErrNoRows` from `tenant_quota_config` + `!spending.Valid`) are persistent state, not CAS races. Fail-fast is correct.
- **B7 retracted**: `go.mod` is 1.25.1, loop-var capture is safe since 1.22.
- **B4 → N**: single-worker metadata for ≤30 pools is design intent (list API is one call); single Cloud stall still serializes but not a bug.
- **B6 → concede**: filter compares by ID from `ready` map, position-independent.

**Standing**: B3 (cross-org batch lock, agreed with @dat9-dev1 as follow-up), B5 (30 vs 50 doc discrepancy), B8 (replenishment gate — minor).

Net: B3 is the only real follow-up bug. Thanks @dat9-dev1 for the pushback.

PR #794, opened 2026-07-25
fix: chunk shared tenant pool replenishment

srstack:
## Summary

- Remove the tenant-pool distributed MySQL `GET_LOCK` from replenishment workers.
- Keep shared replenishment bounded to one physical-DB capacity per batch (`maxTenants`).
- Preserve the configured refill wave cap (`managedSharedDBRefillPoolLimit`) across all batches started by one trigger.
- Re-read durable active/pending/provisioning slots before every batch; concurrent Pods may cause bounded over-create, but they will not under-create because in-flight slots are counted.
- Keep the per-Pod replenish coalescing gate and the shared DB organization allocation lock.
- Keep native tenant-pool replenishment behavior unchanged.

## Design

The tenant-pool refill path is intentionally lockless across Pods. The initial active-free watermark check decides whether a refill wave should start. Once a wave starts, subsequent batches continue until durable free slots reach the pool target or the configured wave budget is exhausted; the active-free watermark is not re-applied between batches.

Correctness comes from durable slot accounting and transactional shared-DB capacity reservation:

- `active`, `pending`, and `provisioning` entries all consume pool slots;
- each batch is bounded by one physical shared DB capacity;
- the shared DB organization allocation lock prevents duplicate physical DB planning;
- the per-DB transaction prevents `tenant_count` from exceeding `maxTenants`;
- the next claim or durable-state scan starts another bounded wave if a previous wave failed or the pool remains below target.

## Why

The previous implementation held a tenant-pool named lock while performing tenant creation, shared DB allocation, cloud requests, result persistence, and schema-init scheduling. Large pools could exceed the 300-second MySQL named-lock wait timeout and cause other Pods to fail replenishment with `admin_tenant_pool_replenish_lock_failed`.

The refill path no longer depends on that lock. Multiple Pods can replenish concurrently, while bounded batches and durable slot rechecks prevent runaway creation and preserve the no-under-create behavior.

## Verification

- `go test -race ./pkg/server -run 'TestAdminTenantPoolReplenish|TestManagedSharedDBRefill' -count=1`
- `make lint`
- `CGO_ENABLED=0 go build ./...`
- `git diff --check`

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved tenant pool replenishment to continue across multiple batches until available capacity is filled.
  - Added safeguards for shared-provider pools to prevent over-provisioning during large refill operations.
  - Improved handling and reporting of replenishment errors.

- **Tests**
  - Added coverage for large, batched refills.
  - Added validation that replenishment continues beyond the refill threshold until the pool is fully populated.

### DIFF
--- a/pkg/server/admin_tenant_pool.go
+++ b/pkg/server/admin_tenant_pool.go
@@ -1898,6 +1898,8 @@ func (s *Server) replenishTenantPoolAsync(ctx context.Context, pool *meta.Tenant
 		sharedProvider := s.defaultTenantProvider == tenant.ProviderTiDBCloudNativeShared
 		createdAny := false
 		sharedWaveRemaining := 0
+		// A started shared wave accounts for in-flight pending/provisioning slots;
+		// sharedWaveRemaining is the hard cap that guarantees termination.
 		for {
 			if err := ctx.Err(); err != nil {
 				metricResult = adminTenantPoolMetricResult(err)
--- a/pkg/server/admin_tenant_pool.go
+++ b/pkg/server/admin_tenant_pool.go
@@ -1948,82 +1948,35 @@ func (s *Server) replenishTenantPoolAsync(ctx context.Context, pool *meta.Tenant
 				return
 			}
 
-			// The common no-op path returns before taking either lock. Re-check all
-			// capacity state under the distributed lock before provisioning. Shared
-			// refills do one physical-DB-sized batch per lock hold; the loop below
-			// reacquires it only after this batch has been persisted.
-			batchCreated := 0
-			lock := s.tenantPoolLock(pool.PoolID)
-			lock.Lock()
-			lockErr := s.meta.WithTenantPoolLock(ctx, pool.PoolID, func(ctx context.Context) error {
-				current, err = s.meta.GetTenantPoolByID(ctx, pool.PoolID)
-				if err != nil {
-					if !errors.Is(err, meta.ErrNotFound) {
-						logger.Warn(ctx, "admin_tenant_pool_replenish_get_pool_failed", zap.String("pool_id", pool.PoolID), zap.Error(err))
-						metricResult = "error"
-					} else {
-						metricResult = "not_found"
-					}
-					return nil
-				}
-				if current.Status != meta.TenantPoolActive || current.OrganizationID == "" || current.Size <= 0 {
-					metricResult = "skipped"
-					return nil
-				}
-				freeSize, err = s.meta.CountFreeTenantPoolBindings(ctx, current.OrganizationID)
-				if err != nil {
-					logger.Warn(ctx, "admin_tenant_pool_replenish_free_count_failed", zap.String("pool_id", current.PoolID), zap.Error(err))
-					metricResult = "error"
-					return nil
-				}
-				s.recordTenantPoolCapacity(current.PoolID, current.OrganizationID, current.Size, freeSize)
-				if !s.tenantPoolBelowRefillWatermark(freeSize, current.Size) {
-					logger.Info(ctx, "admin_tenant_pool_replenish_skipped",
-						zap.String("pool_id", current.PoolID),
-						zap.String("organization_id", current.OrganizationID),
-						zap.Int("pool_size", current.Size),
-						zap.Int("free_size", freeSize),
-						zap.Float64("refill_free_ratio", s.effectiveTenantPoolRefillFreeRatio()))
-					metricResult = "noop"
-					return nil
-				}
-				// Trigger on active free tenants, but size refill against all free
-				// slots, including in-flight pending/provisioning tenants, so
-				// concurrent replenishment does not double-provision.
-				slotSize, err := s.meta.CountTenantPoolFreeSlots(ctx, current.OrganizationID)
-				if err != nil {
-					logger.Warn(ctx, "admin_tenant_pool_replenish_count_failed", zap.String("pool_id", current.PoolID), zap.Error(err))
-					metricResult = "error"
-					return nil
-				}
-				missing := current.Size - slotSize
-				if missing <= 0 {
-					metricResult = "noop"
-					return nil
-				}
-				if sharedProvider {
-					maxTenants, _ := s.managedSharedDBPolicy()
-					missing = managedSharedDBRefillBatchTenantCount(missing, maxTenants, s.managedSharedDBRefillPoolLimit)
-				}
-				results, err := s.createFreePoolTenants(ctx, current.PoolID, missing, cred, nil)
-				if err != nil {
-					logger.Warn(ctx, "admin_tenant_pool_replenish_failed", zap.String("pool_id", current.PoolID), zap.Error(err))
-					metricResult = "cluster_error"
-					return nil
-				}
-				batchCreated = len(results)
-				for _, res := range results {
-					s.startProvisionedTenantSchemaInit(ctx, res)
-				}
-				return nil
-			})
-			lock.Unlock()
-			if lockErr != nil {
-				logger.Warn(ctx, "admin_tenant_pool_replenish_lock_failed", zap.String("pool_id", pool.PoolID), zap.Error(lockErr))
-				metricResult = adminTenantPoolMetricResult(lockErr)
+			// The initial watermark check above avoids a slot COUNT on the common
+			// no-op path. Re-read durable slots before every batch because other
+			// Pods may replenish this pool concurrently; over-create is bounded and
+			// acceptable, while pending/provisioning slots prevent under-counting.
+			slotSize, err := s.meta.CountTenantPoolFreeSlots(ctx, current.OrganizationID)
+			if err != nil {
+				logger.Warn(ctx, "admin_tenant_pool_replenish_count_failed", zap.String("pool_id", current.PoolID), zap.Error(err))
+				metricResult = "error"
 				return
 			}
-			if !sharedProvider || batchCreated == 0 {
+			missing := current.Size - slotSize
+			if missing <= 0 {
+				metricResult = "noop"
+				return
+			}
+			if sharedProvider {
+				maxTenants, _ := s.managedSharedDBPolicy()
+				missing = managedSharedDBRefillBatchTenantCount(missing, maxTenants, s.managedSharedDBRefillPoolLimit)
+			}
+			results, err := s.createFreePoolTenants(ctx, current.PoolID, missing, cred, nil)
+			if err != nil {
+				logger.Warn(ctx, "admin_tenant_pool_replenish_failed", zap.String("pool_id", current.PoolID), zap.Error(err))
+				metricResult = "cluster_error"
+				return
+			}
+			for _, res := range results {
+				s.startProvisionedTenantSchemaInit(ctx, res)
+			}
+			if !sharedProvider || len(results) == 0 {
 				return
 			}
 		}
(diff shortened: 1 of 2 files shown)


srstack:
Follow-up commit 6725c2b9 removes the tenant-pool distributed GET_LOCK from replenish workers. Shared refill keeps the maxTenants-sized batch bound and re-reads durable active/pending/provisioning slots before each batch; the per-Pod coalescing gate and shared DB organization allocation lock remain. Added regression coverage proving refill proceeds while the tenant-pool GET_LOCK is held externally. Full pkg/server + pkg/meta tests, race target, lint, and CGO_ENABLED=0 build pass.

srstack:
Addressed the review feedback in commit b6c8d6de:\n\n- Preserve the configured refill wave cap: initialize one shared wave budget from managedSharedDBRefillTenantCount, decrement it per successful batch, and stop when exhausted.\n- Continue a started shared refill until durable slot count reaches pool size; the active-free watermark is only used for the initial trigger check.\n- Keep replenish metrics as ok after successful batches instead of overwriting them with the final noop check.\n- Poll the regression test until CountTenantPoolFreeSlots reaches the target instead of asserting immediately after the third cloud call.\n- Added coverage for an existing active shared DB where active free tenants reach the 0.8 watermark before all slots are filled.\n\nVerification: full pkg/server + pkg/meta tests pass, targeted -race passes, lint passes, and CGO_ENABLED=0 go build ./... passes.

srstack:
Reviewed the full diff and traced the concurrency-model change end to end (branch `fix/chunked-tenant-pool-replenishment` vs `main`). I can't formally approve — I'm the PR author — so this is a detailed review comment. Verified the branch with `go test -race ./pkg/server -run 'TestAdminTenantPoolReplenish|TestManagedSharedDBRefill'` (green) in a worktree; the `ci` check is green and `e2e smoke` is still running.

## What this PR actually changes
It's not just "chunking" — it **removes the tenant-pool distributed `GET_LOCK` (`WithTenantPoolLock`) and the in-process `tenantPoolLock` from the replenish path**, and replaces the single provision call with a bounded chunked loop. That's the load-bearing change; the chunking is what makes lockless replenish safe. The motivation is sound: the old code held the named lock across tenant inserts + shared-DB allocation + cloud calls + persistence + schema-init scheduling, so a large pool could blow past the 300s MySQL named-lock timeout and make *other* pods fail with `admin_tenant_pool_replenish_lock_failed`.

## Why the lockless design is correct — verified
1. **Single-pod loop terminates (two independent bounds).** `createFreePoolTenants` → `insertPendingPoolTenant` persists a `pending` slot row for each tenant *before* the cloud batch, and `CountTenantPoolFreeSlots` counts active+pending+provisioning. So each iteration's re-read shrinks `missing`. Independently, `sharedWaveRemaining -= len(results)` strictly decreases every continuing iteration (a `len(results)==0` batch returns), giving a hard iteration ceiling even if slot counts didn't move (e.g. a created tenant immediately failed and freed its slot).
2. **No under-create.** Because in-flight pending/provisioning slots are counted, a concurrent pod never under-counts and skips needed refill.
3. **Over-create is bounded, not unbounded.** Cross-pod, the worst case is roughly (concurrent pods × one `maxTenants` batch) in the count→pending-insert window. The per-DB transaction still caps `tenant_count ≤ maxTenants`, and the shared-DB org allocation lock (retained inside `createFreeSharedPoolTenant`, unchanged by this PR) still prevents duplicate physical-DB planning. Extra free tenants are wasteful, not incorrect.
4. **Native path unchanged.** `if !sharedProvider || len(results) == 0 { return }` exits after exactly one batch for non-shared providers — matches `main`. The chunk loop only iterates for the shared provider.
5. **Watermark semantics are deliberate and tested.** The active-free watermark gates whether a wave *starts* (`!createdAny && !belowWatermark`), then is intentionally not re-applied between batches so a started wave fills to the durable slot target. `TestAdminTenantPoolReplenishContinuesPastWatermarkToFillSlots` locks this in; `TestAdminTenantPoolReplenishChunksLargeRefill` confirms a Size-250 / maxTenants-100 refill issues three bounded cloud batches.
6. **Per-pod coalescing retained.** `beginTenantPoolReplenishment` (the `tenantPoolWorkGate` `running` + `nextAllowed` gate) still prevents a single pod from double-running; both callers are fire-and-forget and never assumed the function blocked or serialized cross-pod.

## Non-blocking observations
- **Partial-error path skips schema-init for already-created tenants** — when `createFreePoolTenants` returns non-empty `results` *and* a non-nil `err`, the loop logs `cluster_error` and returns before the `startProvisionedTenantSchemaInit` loop. I checked `main`: it had the identical structure, so this is **pre-existing, not a regression**. Those tenants remain pending/provisioning and are recovered by the existing schema-init retry path. Worth a follow-up to schedule schema-init on the successful subset before returning, but out of scope here.
- **Metric result on multi-batch waves**: once `createdAny` is set, a later `missing <= 0` batch returns without overwriting `metricResult` (stays `ok`) — correct. Just noting the `!createdAny` guards on the `noop` assignments are load-bearing for accurate metrics.

## Suggestions (optional)
- The termination argument (pending slots make `missing` monotonically shrink + `sharedWaveRemaining` as the hard ceiling) is subtle enough to deserve a one-line code comment at the top of the loop, so a future reader doesn't reintroduce a watermark re-check between batches and accidentally stall wave completion.

Solid fix — it removes a real cross-pod failure mode while preserving no-under-create via durable slot accounting and the retained per-DB/org locks. The two new tests cover exactly the risky behaviors (chunk count + past-watermark fill).

pkg/server/admin_tenant_pool.go
```diff
@@ -1895,106 +1895,97 @@ func (s *Server) replenishTenantPoolAsync(ctx context.Context, pool *meta.Tenant
...
+			freeSize, err := s.meta.CountFreeTenantPoolBindings(ctx, current.OrganizationID)
 			if err != nil {
 				logger.Warn(ctx, "admin_tenant_pool_replenish_free_count_failed", zap.String("pool_id", current.PoolID), zap.Error(err))
 				metricResult = "error"
-				return nil
+				return
 			}
 			s.recordTenantPoolCapacity(current.PoolID, current.OrganizationID, current.Size, freeSize)
-			if !s.tenantPoolBelowRefillWatermark(freeSize, current.Size) {
+			if !createdAny && !s.tenantPoolBelowRefillWatermark(freeSize, current.Size) {
```
mornyx, line 1929: **[nit]** The `!createdAny &&` guard is load-bearing: once a wave starts, the active-free watermark must not re-stop the loop before durable slots reach `pool.Size` (e.g. free ratio crosses 0.8 mid-wave with slots still short). Glad this is covered by `TestAdminTenantPoolReplenishContinuesPastWatermarkToFillSlots`.

No change requested — just calling out the intentional semantics so a future edit does not reintroduce an inter-batch watermark check.

pkg/server/admin_tenant_pool.go
```diff
@@ -1895,106 +1895,97 @@ func (s *Server) replenishTenantPoolAsync(ctx context.Context, pool *meta.Tenant
...
+				if !createdAny {
+					metricResult = "noop"
+				}
+				return
 			}
-			if s.defaultTenantProvider == tenant.ProviderTiDBCloudNativeShared {
+			if sharedProvider {
 				maxTenants, _ := s.managedSharedDBPolicy()
-				missing = managedSharedDBRefillTenantCount(missing, maxTenants, s.managedSharedDBRefillPoolLimit)
+				if !createdAny {
```
mornyx, line 1959: **[nit]** Computing `sharedWaveRemaining` only on the first successful path into the shared branch (`!createdAny`) correctly preserves the configured refill wave cap across chunks instead of re-applying `managedSharedDBRefillTenantCount` every iteration (which would allow another full wave budget per batch).

Looks correct.

pkg/server/admin_tenant_pool_test.go
```diff
@@ -1117,6 +1117,172 @@ func TestAdminTenantPoolReplenishBatchesBelowFreeWatermark(t *testing.T) {
...
+		if err := <-lockDone; err != nil {
+			t.Errorf("release tenant pool lock: %v", err)
+		}
+	})
+
+	srv.replenishTenantPoolAsync(ctx, pool, tenant.CredentialProvisionRequest{
+		PublicKey:  "public-1",
+		PrivateKey: "private-1",
+	})
+	deadline := time.Now().Add(10 * time.Second)
```
mornyx, line 1191: **[suggestion]** Consider giving the free-slot polling loop its own deadline (e.g. reset `deadline := time.Now().Add(10 * time.Second)` after the batch-count loop succeeds). Sharing one 10s budget across both stages is usually fine, but under a slow CI host the first poll can consume most of the window and make the second stage flake.

Non-blocking — optional hardening.

mornyx (approved):
## Review summary

**Verdict: Approve**

This is a solid fix for the real cross-pod failure mode (`admin_tenant_pool_replenish_lock_failed` when `GET_LOCK` is held across multi-batch shared refills). Removing the tenant-pool named lock from the replenish hot path while keeping durable slot accounting + per-DB capacity transactions + the shared-DB organization allocation lock is the right tradeoff.

### Correctness checks

1. **Termination** — each successful shared batch inserts durable pending/provisioning slots (so `missing` shrinks on re-read) and `sharedWaveRemaining` decreases by `len(results)`. The `len(results)==0` / `sharedWaveRemaining<=0` exits give a hard stop even if slot accounting were sticky.
2. **No under-create** — `CountTenantPoolFreeSlots` still counts active+pending+provisioning, so concurrent Pods do not skip needed refill by ignoring in-flight work.
3. **Bounded over-create** — worst case is roughly concurrent Pods × one `maxTenants` batch in the count→insert race. Per-DB `tenant_count <= max_tenants` and the shared-DB allocation lock still prevent unbounded physical DB fan-out / over-subscription.
4. **Native path unchanged** — non-shared provider returns after one batch (`!sharedProvider || len(results)==0`).
5. **Watermark semantics** — watermark only gates wave *start* (`!createdAny && !belowWatermark`); once started, the wave fills to the durable slot target. The new `TestAdminTenantPoolReplenishContinuesPastWatermarkToFillSlots` locks this in (seed 100/250 → free ratio 0.4, must continue past 0.8 watermark to reach 250).
6. **Wave budget** — `sharedWaveRemaining` is computed once from `managedSharedDBRefillTenantCount` and preserved across batches, so the configured physical-pool wave cap is not re-multiplied every iteration.
7. **Metrics** — `metricResult = "ok"` is set when any batch creates tenants; later `missing<=0` only writes `"noop"` when `!createdAny`, so multi-batch waves are not mis-recorded as noop.
8. **Per-Pod coalescing retained** — `beginTenantPoolReplenishment` still serializes same-Pod work.

### Tests

- `TestAdminTenantPoolReplenishChunksLargeRefill` (250 / maxTenants 100 → 100+100+50, and proceeds while another goroutine holds `WithTenantPoolLock`) is a good lock-removal regression.
- Past-watermark fill test covers the intentional behavior change vs re-checking the active-free watermark between batches.

CI is green (code-ci + local e2e).

### Non-blocking notes

- Partial-error path still skips schema-init for the successful subset when `createFreePoolTenants` returns `(results, err)` — same structure as `main`, so not a regression. Worth a follow-up to schedule schema-init for the successful subset before returning.
- The large-refill test shares one 10s deadline across batch-count polling and free-slot polling; minor flake risk under load. Separate deadlines would be slightly more robust.

No blocking issues. Approved.

PR #798, opened 2026-07-26
feat: defer quota counter updates to transaction end and skip no-op file meta upserts

mornyx:
## Why

Follow-up to #792. After deploying it and re-running the stress workload, DBBrain shows the commit-rate problem is solved (transactions −78%, commit latency 19ms → 6.5ms), but two new facts surfaced in the quota accounting chain:

1. **Row-lock contention exploded on the quota hot rows.** `SELECT ... tenant_file_meta FOR UPDATE` lock wait went 452s → 8,408s/12h; `INSERT IGNORE tenant_quota_usage` accumulated 8,152s of lock wait on only 553k execs (~15ms avg); `UPDATE tenant_quota_usage` 5,360s on 246k execs (~22ms avg). Cause: batched apply transactions (up to 128 items) hold every touched row from first apply to commit, and every mutation touches the single per-tenant `tenant_quota_usage` row — two pods' batches over the same tenant set are almost guaranteed to intersect (birthday paradox), so cross-pod lock waits are amplified by batch length.
2. **99%+ of mutations are quota-neutral.** Only 246k of 41.5M file mutations actually changed `storage_bytes` — the rest are same-size overwrites that still pay a `FOR UPDATE` read + an identical-value upsert per mutation.

## Changes

**Hot row last.** Apply closures now *return* their `tenant_quota_usage` counter deltas instead of executing the per-counter `Incr*Tx` inline. Each transaction owner flushes them at the END of the transaction:

- dispatcher batch runner: aggregates per tenant across the (cross-tenant) batch and flushes in **sorted tenant_id order** (deterministic global lock order against cross-pod deadlocks), before `MarkMutationsAppliedTx`;
- per-item path (`applyQuotaMutation`) and `replayOne`: flush right before the mark, same transaction.

The hot row is now locked only just before commit instead of for the whole batch, and N mutations for one tenant collapse into a single `UPDATE tenant_quota_usage` via the new `IncrQuotaUsageCountersTx` (same additive + row-exists semantics as the per-counter methods; no-op when all deltas are zero — matching the old per-counter conditionals).

**No-op skip.** `applyCentralFileStateTx` already reads the shadow row `FOR UPDATE`; when it exists with identical `size_bytes`/`is_media`, the upsert is skipped entirely (the mutation is still marked applied). Safe because `UpsertFileMetaTx` maintains only those two columns and nothing consumes `tenant_file_meta.updated_at` for correctness (file GC settlement keys off `HasPendingFileMutation` + size/isMedia). `applyUploadCompleteTx` keeps its pre-existing identical-meta early return; the settled-branch reserved→storage transfer is preserved exactly (verified `TransferReservedToConfirmedTx` has no extra guard clause).

Semantics preserved: exactly-once marking (`status='pending'` + affected-rows check), in-tx read-your-writes for create→overwrite on the same file, per-tenant FIFO, failure fallback to the legacy per-item path, and the pending-delta admission estimates are all unchanged.

## Expected effect (same workload)

- `tenant_quota_usage` lock-hold per batch: from ~whole-batch (tens of ms) to just-before-commit — the three lock-wait hot spots should drop by roughly an order of magnitude.
- Usage counter statements: from up to 3 per mutation to ~1 per tenant per batch.
- `tenant_file_meta` upserts: −~40M/12h (no-op overwrites skipped).
- Combined with #792 this should bring metadb statement volume down another ~40% and eliminate the lock-wait tail.

## Test plan

- New backend tests: per-tenant aggregation with sorted flush order; create→overwrite delta composition; no-op overwrite skips upsert+counters but still marks applied; inline path flushes before mark.
- New meta test: `IncrQuotaUsageCountersTx` counter moves + zero-delta no-op.
- `make lint` 0 issues; `make test` green across pkg/meta, pkg/backend, pkg/server, pkg/tenant.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved quota accounting for file creation, overwrites, deletions, and completed uploads.
  * Quota usage updates are now applied consistently and atomically, reducing inconsistent counter states.
  * No-op file updates no longer create unnecessary metadata or quota changes.
  * Mutation replay and batch processing now preserve correct ordering between quota updates and completion status.
* **Tests**
  * Added coverage for aggregated quota updates, transactional behavior, upload reservations, and no-op operations.

### DIFF
--- a/pkg/backend/llm_usage_test.go
+++ b/pkg/backend/llm_usage_test.go
@@ -40,8 +40,8 @@ func (m *mockQuotaStore) EnsureQuotaUsageRow(_ context.Context, _ string) error 
 func (m *mockQuotaStore) IncrStorageBytes(_ context.Context, _ string, _ int64) error   { return nil }
 func (m *mockQuotaStore) IncrReservedBytes(_ context.Context, _ string, _ int64) error  { return nil }
 func (m *mockQuotaStore) IncrFileCount(_ context.Context, _ string, _ int64) error      { return nil }
-func (m *mockQuotaStore) IncrMediaFileCount(_ context.Context, _ string, _ int64) error  { return nil }
-func (m *mockQuotaStore) IncrVideoFileCount(_ context.Context, _ string, _ int64) error  { return nil }
+func (m *mockQuotaStore) IncrMediaFileCount(_ context.Context, _ string, _ int64) error { return nil }
+func (m *mockQuotaStore) IncrVideoFileCount(_ context.Context, _ string, _ int64) error { return nil }
 func (m *mockQuotaStore) TransferReservedToConfirmed(_ context.Context, _ string, _, _ int64) error {
 	return nil
 }
@@ -51,8 +51,8 @@ func (m *mockQuotaStore) AtomicReserveAndInsertUpload(_ context.Context, _ *Uplo
 func (m *mockQuotaStore) IncrStorageBytesTx(_ *sql.Tx, _ string, _ int64) error   { return nil }
 func (m *mockQuotaStore) IncrReservedBytesTx(_ *sql.Tx, _ string, _ int64) error  { return nil }
 func (m *mockQuotaStore) IncrFileCountTx(_ *sql.Tx, _ string, _ int64) error      { return nil }
-func (m *mockQuotaStore) IncrMediaFileCountTx(_ *sql.Tx, _ string, _ int64) error  { return nil }
-func (m *mockQuotaStore) IncrVideoFileCountTx(_ *sql.Tx, _ string, _ int64) error  { return nil }
+func (m *mockQuotaStore) IncrMediaFileCountTx(_ *sql.Tx, _ string, _ int64) error { return nil }
+func (m *mockQuotaStore) IncrVideoFileCountTx(_ *sql.Tx, _ string, _ int64) error { return nil }
 func (m *mockQuotaStore) TransferReservedToConfirmedTx(_ *sql.Tx, _ string, _, _ int64) error {
 	return nil
 }
--- a/pkg/backend/llm_usage_test.go
+++ b/pkg/backend/llm_usage_test.go
@@ -56,6 +56,9 @@ func (m *mockQuotaStore) IncrVideoFileCountTx(_ *sql.Tx, _ string, _ int64) erro
 func (m *mockQuotaStore) TransferReservedToConfirmedTx(_ *sql.Tx, _ string, _, _ int64) error {
 	return nil
 }
+func (m *mockQuotaStore) IncrQuotaUsageCountersTx(_ *sql.Tx, _ string, _, _, _, _ int64) error {
+	return nil
+}
 
 func (m *mockQuotaStore) UpsertFileMeta(_ context.Context, _ *FileMetaView) error { return nil }
 func (m *mockQuotaStore) GetFileMeta(_ context.Context, _, _ string) (*FileMetaView, error) {
--- a/pkg/backend/mutation_dispatcher.go
+++ b/pkg/backend/mutation_dispatcher.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"database/sql"
 	"hash/fnv"
+	"sort"
 	"sync"
 
 	"go.uber.org/zap"
@@ -43,7 +44,7 @@ type quotaMutationItem struct {
 	mutationType   string
 	logID          int64
 	pending        quotaPendingDeltas
-	apply          func(context.Context, *sql.Tx) error
+	apply          func(context.Context, *sql.Tx) (quotaCounterDeltas, error)
 	// raw is a pre-built closure submitted through the legacy enqueueMutation
 	// API (unit tests). Raw items bypass batching and run one by one.
 	raw func(context.Context)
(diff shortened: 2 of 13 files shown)


qiffang (changes requested):
Exact-head review: `272a2f663932b90a5b585611144522d2c85682a5`

CHANGES_REQUESTED.

Blocking finding:
- `pkg/backend/llm_usage_test.go:43` / `:54`: this modified file is not gofmt-clean. `gofmt -l` reports `pkg/backend/llm_usage_test.go`; `gofmt -d` removes the extra alignment space on `IncrMediaFileCount`, `IncrVideoFileCount`, `IncrMediaFileCountTx`, and `IncrVideoFileCountTx`. This should be fixed before merge; otherwise the PR is landing non-standard Go formatting despite the test-plan claim that lint is clean.

Correctness review notes:
- I did not find a quota-accounting correctness regression in the hot-row-last refactor: file create/overwrite/delete, upload-complete, dispatcher batch, per-item fallback, and replay keep apply + counter flush + mark in one transaction, so failed flush/mark rolls back metadata changes and pending replay remains authoritative.
- Batch aggregation preserves apply order for non-commutative file-meta mutations, collapses usage deltas per tenant, and flushes tenant IDs in sorted order for deterministic `tenant_quota_usage` lock ordering. Batch failure falls back per item.
- The no-op file-meta skip is bounded to exact `size_bytes`/`is_media` equality and still marks the mutation applied; I do not see a correctness dependency on `tenant_file_meta.updated_at` in the changed paths.

Performance caveat, not a code-blocker for this PR once gofmt is fixed: the no-op path still executes `SELECT ... tenant_file_meta ... FOR UPDATE` and holds that row lock until commit. This removes the identical-value upsert and counter flush, but it does not remove the 41.5M `FOR UPDATE` statements or fully eliminate file-meta row lock contention. Post-merge DBBrain validation should verify the claim that the lock-wait tail drops enough under the same 12h workload.

Validation run:
- PASS: `git diff --check origin/main...HEAD`
- FAIL: `gofmt -l` / `gofmt -d` on changed Go files (`pkg/backend/llm_usage_test.go`)
- PASS: merge-ref `go test -c ./pkg/meta ./pkg/backend ./pkg/server ./pkg/tenant ./cmd/drive9-server`
- PASS: merge-ref `go vet ./pkg/meta ./pkg/backend ./pkg/server ./pkg/tenant ./cmd/drive9-server`
- BLOCKED locally: focused `go test ./pkg/backend ...` and `go test ./pkg/meta -run TestIncrQuotaUsageCountersTx` panic before test selection with known `rootless Docker not found` testcontainers setup; hosted `ci` and `drive9-server-local e2e smoke` are green.

mornyx:
Fixed in `c1c9a6f` — `pkg/backend/llm_usage_test.go` is now gofmt-clean (the two new mock methods had one extra alignment space). Thanks for catching it; `make lint` (golangci-lint default config) doesn't gate on gofmt in this repo, which is why it slipped through.

On the performance caveat — agreed, and worth being precise about what this PR does and doesn't remove:

- The `SELECT ... FOR UPDATE` itself stays, deliberately: the counter delta must be computed against a locked read, otherwise a cross-pod mutation racing between an optimistic read and the upsert would silently lose absolute-state convergence (the mutations are absolute-state, cross-pod order is explicitly unordered, so an unlocked skip is not safe).
- What the no-op path *does* remove for the 99%+ same-size overwrites: the identical-value upsert (its X-lock, its binlog/redo, and its statement), plus the counter flush.
- The intended lock-wait win is on the `tenant_quota_usage` hot row (the 8,152s + 5,360s spots): deferring counter updates to just-before-commit in sorted tenant order shrinks that hold from the whole batch to ~1 statement. The `tenant_file_meta` FOR UPDATE wait (8,408s) is expected to improve only partially (fewer writers, same readers) — agreed the 12h re-run is the real validation, and if it remains hot, the structural answer is still moving these tables into the tenant/shared DB rather than more metadb-side tricks.

qiffang (approved):
Exact-head re-review: `c1c9a6fd987f994f5a6e911d9e93887f813501cd`

APPROVED.

The only change since my prior CHANGES_REQUESTED review is `style: gofmt llm_usage_test.go`; diff from `272a2f663932b90a5b585611144522d2c85682a5..c1c9a6fd987f994f5a6e911d9e93887f813501cd` is exactly the four gofmt spacing fixes in `pkg/backend/llm_usage_test.go`. The previous formatting blocker is resolved.

I rechecked the quota-accounting paths against the current head and still do not see a blocking correctness regression:
- dispatcher batch, per-item apply, and replay keep apply + combined counter flush + mark in one transaction;
- failed counter flush or mark rolls back metadata changes, leaving pending replay authoritative;
- batch aggregation preserves apply order for non-commutative file-meta operations and flushes usage rows in sorted tenant order;
- no-op file-meta skip is bounded to exact `size_bytes`/`is_media` equality and still marks mutations applied;
- upload_complete still settles reservation/file_meta in-tx and defers only the returned usage-counter deltas.

Validation run:
- PASS: `git diff --check origin/main...HEAD`
- PASS: `gofmt -l` on all changed Go files reports no files
- PASS: merge-ref `go test -c ./pkg/meta ./pkg/backend ./pkg/server ./pkg/tenant ./cmd/drive9-server`
- PASS: merge-ref `go vet ./pkg/meta ./pkg/backend ./pkg/server ./pkg/tenant ./cmd/drive9-server`
- BLOCKED locally: focused `go test ./pkg/backend ...` still panics before test selection with `rootless Docker not found` from the repo testcontainers `TestMain`; hosted `ci` and `drive9-server-local e2e smoke` are green on this exact head.

Non-blocking operational caveat remains: same-size no-op mutations still perform `tenant_file_meta FOR UPDATE`; this PR removes the identical upsert and defers/collapses usage-counter updates, but the file-meta lock hotspot still needs same-workload DBBrain verification after deployment.

PR #800, opened 2026-07-26
feat: configure meta DB connection lifetimes

srstack:
## Summary
- expose meta DB connection max lifetime and max idle time through role-specific environment variables
- allow an explicit 0s max idle time for meta DB connections while preserving shared DB fallback behavior
- document lifetime fallback semantics and cover the requested 1h/0s plus zero-value boundaries

## Test plan
- [x] go test ./pkg/mysqlutil ./cmd/drive9-server -count=1
- [x] make build
- [x] make lint
- [x] git diff --check
- [ ] umask 022; make test (stopped while in progress to push the review fixes immediately as requested)

### DIFF
--- a/cmd/drive9-server/main.go
+++ b/cmd/drive9-server/main.go
@@ -626,6 +626,8 @@ environment:
   DRIVE9_POOL_IDLE_REAP_INTERVAL  how often the idle reaper scans (default: 2m)
   DRIVE9_META_DB_MAX_OPEN_CONNS max open connections for the per-pod meta DB pool (default: 100)
   DRIVE9_META_DB_MAX_IDLE_CONNS max idle connections for the per-pod meta DB pool (default: 20)
+  DRIVE9_META_DB_CONN_MAX_LIFETIME maximum meta connection lifetime (default: 10m)
+  DRIVE9_META_DB_CONN_MAX_IDLE_TIME maximum meta connection idle time (default: 1m, 0s=disabled)
   DRIVE9_USER_DB_MAX_OPEN_CONNS max open connections for each cached tenant user DB pool (default: 6)
   DRIVE9_USER_DB_MAX_IDLE_CONNS max idle connections for each cached tenant user DB pool (default: 2)
   DRIVE9_USER_SCHEMA_DB_MAX_OPEN_CONNS max open connections for tenant schema-init DB pools (default: 8)
--- a/cmd/drive9-server/main.go
+++ b/cmd/drive9-server/main.go
@@ -626,7 +626,7 @@ environment:
   DRIVE9_POOL_IDLE_REAP_INTERVAL  how often the idle reaper scans (default: 2m)
   DRIVE9_META_DB_MAX_OPEN_CONNS max open connections for the per-pod meta DB pool (default: 100)
   DRIVE9_META_DB_MAX_IDLE_CONNS max idle connections for the per-pod meta DB pool (default: 20)
-  DRIVE9_META_DB_CONN_MAX_LIFETIME maximum meta connection lifetime (default: 10m)
+  DRIVE9_META_DB_CONN_MAX_LIFETIME maximum meta connection lifetime (default: 10m; non-positive/invalid values use default)
   DRIVE9_META_DB_CONN_MAX_IDLE_TIME maximum meta connection idle time (default: 1m, 0s=disabled)
   DRIVE9_USER_DB_MAX_OPEN_CONNS max open connections for each cached tenant user DB pool (default: 6)
   DRIVE9_USER_DB_MAX_IDLE_CONNS max idle connections for each cached tenant user DB pool (default: 2)
--- a/pkg/mysqlutil/pool.go
+++ b/pkg/mysqlutil/pool.go
@@ -36,6 +36,7 @@ const (
 //
 // Supported role-specific env vars:
 //   - DRIVE9_META_DB_MAX_OPEN_CONNS / DRIVE9_META_DB_MAX_IDLE_CONNS
+//   - DRIVE9_META_DB_CONN_MAX_LIFETIME / DRIVE9_META_DB_CONN_MAX_IDLE_TIME
 //   - DRIVE9_USER_DB_MAX_OPEN_CONNS / DRIVE9_USER_DB_MAX_IDLE_CONNS
 //   - DRIVE9_USER_SCHEMA_DB_MAX_OPEN_CONNS / DRIVE9_USER_SCHEMA_DB_MAX_IDLE_CONNS
 //   - DRIVE9_SHARED_DB_MAX_OPEN_CONNS / DRIVE9_SHARED_DB_MAX_IDLE_CONNS
@@ -58,11 +59,11 @@ func ApplyPoolDefaults(db *sql.DB, role string) {
 
 func poolLifetime(role string) (time.Duration, time.Duration) {
 	lifetime, idleTime := defaultPoolLifetime(role)
-	if role != RoleShared {
+	if role != RoleMeta && role != RoleShared {
 		return lifetime, idleTime
 	}
-	return poolEnvDuration(role, "CONN_MAX_LIFETIME", lifetime),
-		poolEnvDuration(role, "CONN_MAX_IDLE_TIME", idleTime)
+	return poolEnvDuration(role, "CONN_MAX_LIFETIME", lifetime, false),
+		poolEnvDuration(role, "CONN_MAX_IDLE_TIME", idleTime, true)
 }
 
 func defaultPoolLifetime(role string) (time.Duration, time.Duration) {
@@ -102,7 +103,7 @@ func poolEnvInt(role, suffix string, def int) int {
 	return def
 }
 
-func poolEnvDuration(role, suffix string, def time.Duration) time.Duration {
+func poolEnvDuration(role, suffix string, def time.Duration, allowZero bool) time.Duration {
 	key := rolePoolEnvKey(role, suffix)
 	if key == "" {
 		return def
@@ -112,7 +113,7 @@ func poolEnvDuration(role, suffix string, def time.Duration) time.Duration {
 		return def
 	}
 	value, err := time.ParseDuration(raw)
-	if err != nil || value <= 0 {
+	if err != nil || value < 0 || value == 0 && !allowZero {
 		return def
 	}
 	return value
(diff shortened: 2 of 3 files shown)


pkg/mysqlutil/pool.go
```diff
@@ -58,11 +59,11 @@ func ApplyPoolDefaults(db *sql.DB, role string) {
...
 func poolLifetime(role string) (time.Duration, time.Duration) {
 	lifetime, idleTime := defaultPoolLifetime(role)
-	if role != RoleShared {
+	if role != RoleMeta && role != RoleShared {
 		return lifetime, idleTime
 	}
-	return poolEnvDuration(role, "CONN_MAX_LIFETIME", lifetime),
-		poolEnvDuration(role, "CONN_MAX_IDLE_TIME", idleTime)
+	return poolEnvDuration(role, "CONN_MAX_LIFETIME", lifetime, false),
+		poolEnvDuration(role, "CONN_MAX_IDLE_TIME", idleTime, true)
```
mornyx: **[nit]** `allowZero=true` is shared by meta **and** shared roles. That means `DRIVE9_SHARED_DB_CONN_MAX_IDLE_TIME=0s` now disables idle recycling for shared pools, whereas previously `<=0` fell back to the 5m default.

If that side effect is intentional, please add `0s=disabled` to the shared usage line in `cmd/drive9-server/main.go` for symmetry. If the intent was meta-only, pass `allowZero` based on role (e.g. `role == RoleMeta`).

Non-blocking either way — just make the contract explicit.

cmd/drive9-server/main.go
```diff
@@ -626,6 +626,8 @@ environment:
   DRIVE9_POOL_IDLE_REAP_INTERVAL  how often the idle reaper scans (default: 2m)
   DRIVE9_META_DB_MAX_OPEN_CONNS max open connections for the per-pod meta DB pool (default: 100)
   DRIVE9_META_DB_MAX_IDLE_CONNS max idle connections for the per-pod meta DB pool (default: 20)
+  DRIVE9_META_DB_CONN_MAX_LIFETIME maximum meta connection lifetime (default: 10m)
```
mornyx: **[nit]** Consider noting that `DRIVE9_META_DB_CONN_MAX_LIFETIME=0s` (or any non-positive / unparseable value) is **invalid** and falls back to the default `10m`, while only idle-time treats `0s` as “disabled”. Operators often assume `0` means unlimited for both knobs (matching `database/sql`).

Non-blocking documentation polish.

pkg/mysqlutil/pool_test.go
```diff
@@ -111,3 +111,13 @@ func TestSharedPoolDurationEnv(t *testing.T) {
 		t.Fatalf("shared env durations = %s/%s, want 45m/12m", lifetime, idleTime)
 	}
 }
+
+func TestMetaPoolDurationEnvAllowsZeroIdleTime(t *testing.T) {
```
mornyx, line 124: **[nit]** Good coverage for the requested `1h` / `0s` pair. A sibling case for `DRIVE9_META_DB_CONN_MAX_LIFETIME=0s` → default `10m` (and maybe invalid `-1s` / `bad`) would lock the “zero lifetime rejected” half of the contract without much cost.

Optional.

mornyx (approved):
## Review summary

**Verdict: Approve**

Small, focused change: meta (and, via the shared code path, shared) pools can override `ConnMaxLifetime` / `ConnMaxIdleTime` from env, with explicit `0s` allowed only for idle time so operators can disable idle recycling while still requiring a positive lifetime. Wiring matches existing `rolePoolEnvKey` / `ApplyPoolDefaults` patterns, CI is green, and the `1h`/`0s` case is covered by a unit test.

### Correctness

1. **Lifetime semantics** — `database/sql`: `SetConnMaxIdleTime(0)` disables idle-based close; `SetConnMaxLifetime(0)` would disable age-based close. Rejecting `0` for lifetime (fall back to default) and accepting `0` for idle matches the stated intent.
2. **Precedence** — `err != nil || value < 0 || value == 0 && !allowZero` is parsed as expected (`&&` binds tighter than `||`).
3. **Meta enablement** — before this PR only `RoleShared` consulted duration env vars; `RoleMeta` always used hardcoded defaults. Opening the gate for `RoleMeta` is the real feature and looks correct.
4. **Call path** — `meta.Open` → `OpenInstrumented(..., RoleMeta)` → `ApplyPoolDefaults` will pick up the new env vars with no further wiring.

### Non-blocking notes

- `allowZero=true` applies to **both** meta and shared idle-time env keys. That is a small behavior change for `DRIVE9_SHARED_DB_CONN_MAX_IDLE_TIME=0s` (previously treated as invalid → default). Fine if intentional; worth documenting on the shared usage line or scoping `allowZero` to meta only.
- Docs say `0s=disabled` for meta idle, but not that `CONN_MAX_LIFETIME=0s` is invalid and falls back to the default (silent). A short note would reduce operator surprise.
- Optional test: `DRIVE9_META_DB_CONN_MAX_LIFETIME=0s` still yields the default 10m.

No blockers. Approved.

qiffang (changes requested):
REVISE at exact head `2c65fbbe78644051a207bc399bf7ecd6e89aa536`.

Blocking findings:

1. `pkg/mysqlutil/pool.go:65-66` silently changes the shared-pool operator contract. `poolLifetime()` now calls `poolEnvDuration(..., allowZero=true)` for both `RoleMeta` and `RoleShared`, so `DRIVE9_SHARED_DB_CONN_MAX_IDLE_TIME=0s` changes from "invalid, fall back to 5m" to "disable idle-time recycling" for every shared-schema DB handle. The PR title/summary and `cmd/drive9-server/main.go` only advertise `0s=disabled` for the meta DB knob, while the shared help line still says only `default: 5m`. This is not just documentation polish: shared handles can be numerous and long-lived, so hidden disabling of idle recycling changes production connection-retention behavior. Fix by either making `allowZero` role-specific (`role == RoleMeta`) or explicitly documenting and testing the shared `0s=disabled` behavior as an intentional contract.

2. The claimed "zero lifetime invalid" half of the contract is not regression-tested. `TestMetaPoolDurationEnvAllowsZeroIdleTime` covers `lifetime=1h` + `idle=0s`, but no test proves `DRIVE9_META_DB_CONN_MAX_LIFETIME=0s` falls back to the default `10m` while idle-time `0s` remains accepted. That distinction is the high-risk operator edge case because `database/sql` itself treats non-positive lifetime as no limit; this wrapper intentionally does not. Add a focused regression for zero lifetime fallback, and if shared `0s` is kept, cover shared idle zero explicitly too.

Validation run locally:

- `git diff --check 3bd5032b5a27f459a6ce7cacf434fac898d52944...HEAD`
- `gofmt -l cmd/drive9-server/main.go pkg/mysqlutil/pool.go pkg/mysqlutil/pool_test.go`
- `go test ./pkg/mysqlutil -count=1`
- `go test ./pkg/mysqlutil ./cmd/drive9-server -count=1`
- `make lint`
- `make build-server`

Hosted `ci` and `drive9-server-local e2e smoke` are green. CodeRabbit status is green but the auto-review body says it was rate-limited, so I would not count it as substantive reviewer coverage.

pkg/mysqlutil/pool_test.go
srstack: Added TestMetaPoolDurationEnvRejectsZeroLifetime in 37f86373. It verifies lifetime 0s falls back to the default 10m while idle-time 0s remains accepted. Also added a shared zero-idle regression test.

pkg/mysqlutil/pool.go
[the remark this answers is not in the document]
srstack: Fixed in 37f86373. Zero idle time is now allowed only for RoleMeta; DRIVE9_SHARED_DB_CONN_MAX_IDLE_TIME=0s again falls back to the existing 5m default. Added TestSharedPoolDurationEnvRejectsZeroIdleTime to lock that behavior.

cmd/drive9-server/main.go
srstack: Updated the usage text in 37f86373 to document that non-positive or invalid meta lifetime values use the default. Only the idle-time setting documents and accepts 0s as disabled.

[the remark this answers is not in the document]
srstack: Updated in 37f86373. The usage text now states that non-positive or invalid meta connection lifetime values use the default, while the idle-time line retains the explicit 0s=disabled contract.

pkg/mysqlutil/pool.go
srstack: Scoped the behavior to meta only in 37f86373 by using role == RoleMeta for allowZero. Shared idle 0s keeps its previous fallback-to-5m behavior, covered by TestSharedPoolDurationEnvRejectsZeroIdleTime.

srstack:
@qiffang Addressed both blocking findings in 37f86373: zero idle time is now meta-only, shared 0s retains the prior 5m fallback with a regression test, and meta lifetime 0s fallback is covered while meta idle 0s remains accepted. The usage text now documents non-positive/invalid lifetime fallback. Verification at this head: go test ./pkg/mysqlutil ./cmd/drive9-server -count=1, make build, make lint, and git diff --check passed. The in-progress full make test was stopped to push immediately at the requester’s direction.

qiffang (approved):
GREEN at exact head `37f86373376082b2ba03ba807d9c06318e416200`.

R1 blockers are closed:

- Shared idle-time semantics are preserved. `poolLifetime()` now permits zero idle time only for `RoleMeta` (`role == RoleMeta`), so `DRIVE9_SHARED_DB_CONN_MAX_IDLE_TIME=0s` again falls back to the existing shared default instead of disabling idle recycling for shared DB handles.
- The operator contract is now covered by regressions: `TestSharedPoolDurationEnvRejectsZeroIdleTime` locks shared `0s` fallback, and `TestMetaPoolDurationEnvRejectsZeroLifetime` locks meta lifetime `0s` fallback while meta idle `0s` remains accepted.
- Usage text now states that non-positive/invalid meta lifetime values use the default; only meta idle-time documents `0s=disabled`.

Validation run locally on this head:

- `git diff --check 3bd5032b5a27f459a6ce7cacf434fac898d52944...HEAD`
- `gofmt -l cmd/drive9-server/main.go pkg/mysqlutil/pool.go pkg/mysqlutil/pool_test.go`
- `go test ./pkg/mysqlutil ./cmd/drive9-server -count=1`
- `go test -c ./pkg/mysqlutil ./cmd/drive9-server`
- `make lint`
- `make build-server`

Hosted `ci`, `drive9-server-local e2e smoke`, and CodeRabbit are green on `37f86373`.

PR #801, opened 2026-07-26
perf: scale shared tenant backend cache

srstack:
## Summary

- use a provider-aware tenant backend cache default: 20,480 entries for `tidb_cloud_native_shared` and 1,024 for standalone/native providers, while preserving `DRIVE9_POOL_MAX_TENANTS` overrides
- keep shared tenant backends out of time-based idle eviction while retaining the existing capacity-bounded LRU
- keep safety-net `AcquireCached` scans from changing capacity LRU order, so only foreground/durable tenant activity influences eviction priority
- replace per-tenant quota-config polling goroutines with request-driven caching: successful refreshes are coalesced and spread across 27-30s, stale config is served on failure, and failed loads retry after 5s
- isolate coalesced MetaDB loads from the triggering request cancellation while bounding each load to 5s
- keep quota refreshes off concurrent request latency: warm waiters immediately use stale config, while cold waiters stop waiting when their own context expires

## Why

Shared tenants are lightweight `fs_id`-scoped backends over shared physical DB handles. Applying the standalone idle TTL caused stable high-cardinality traffic to churn through cold opens. Raising the shared cache without removing per-tenant quota polling would also scale goroutine count and MetaDB version polling linearly, so quota config is now loaded only for active request paths.

Shared physical `sql.DB` handles intentionally remain available while cached tenant entries reference them. This does not pin active database connections indefinitely: the shared connection pool prunes idle connections after its configured idle time, 5m by default. Capacity eviction releases tenant references; once a handle has no references, the existing 30m handle reaper closes it. Explicit invalidation and process shutdown close handles immediately.

## Test plan

- `make test`
- `go test -race ./pkg/backend ./pkg/tenant ./cmd/drive9-server -count=1`
- `go test ./pkg/server ./pkg/meta -count=1`
- `make lint`
- `CGO_ENABLED=0 go build ./...`
- `git diff --check`

## Summary by CodeRabbit

- **New Features**
  - Added provider-aware defaults for tenant backend quota cache capacity, with updated environment-variable semantics and explicit overrides honored.
- **Bug Fixes**
  - Improved quota configuration caching: passive lazy loading, stale-while-retry refresh behavior, coalesced concurrent refreshes, better handling of canceled/deadlined callers, and safer refresh timing (including failure cooldown/retry with jitter).
  - Shared-schema tenant backends are no longer evicted by standalone idle/capacity logic; cached access no longer alters LRU recency.
- **Documentation**
[...]

PR #802, opened 2026-07-26
fix: reduce metric cardinality and close observability gaps

srstack:
## Summary

- reduce high-cardinality tenant metric shapes and move per-tenant service attribution to a failure-only counter
- delete healthy or retired tenant gauge series, including cross-pod cleanup through the tenant notify outbox
- add bounded metrics for mutation dispatch, API key caching, notify coalescing, tenant outbox delivery, shared-pool convergence, and Vault feature state
- implement the TiDB Cloud OpenAPI api/operation/result contract directly in the HTTP producer, including digest retry, transport/status classification, idempotent delete, and protocol failures
- remove duplicate/dead event-bus counters and instrument the actual retention sweep path
- reduce per-tenant DB pool connection states while preserving role-level saturation capacity

## Alert consumers

The matching alert and rule-test changes were pushed to tidbcloud/runbooks#2688. They replace false-green SSE, extraction, semantic-worker, background-worker, and Vault expressions and add alerts for the new critical-path metrics.

## Validation so far

- focused metric contract, gauge lifecycle, dispatcher, cache, coalescer, outbox, shared-pool, Vault, deletion broadcast, and TiDB Cloud OpenAPI tests pass
- runbooks promtool rule tests pass
- full regression and lint will be run after PR creation, per requested sequencing; results will be added in a follow-up comment

## Scope notes

- no docs files are included
- this PR intentionally includes the TiDB Cloud OpenAPI metrics work previously carried by #785, so it does not depend on that PR merging first

### DIFF
--- a/pkg/server/fork.go
+++ b/pkg/server/fork.go
@@ -644,7 +644,7 @@ func (s *Server) provisionForkTenantOnceWithCredentials(ctx context.Context, for
 	if err := s.finalizeTenantSchemaInit(ctx, forkID, dsn, forkTenant.Provider); err != nil {
 		return err
 	}
-	store, err := datastore.Open(dsn)
+	store, err := datastore.OpenForTenant(ctx, dsn, forkID)
 	if err != nil {
 		return err
 	}
@@ -1077,7 +1077,7 @@ func (s *Server) openTenantStore(ctx context.Context, t *meta.Tenant) (*datastor
 	if err != nil {
 		return nil, err
 	}
-	return datastore.Open(tenantDSN(t.DBUser, string(plain), t.DBHost, t.DBPort, t.DBName, t.DBTLS, t.Provider))
+	return datastore.OpenForTenant(ctx, tenantDSN(t.DBUser, string(plain), t.DBHost, t.DBPort, t.DBName, t.DBTLS, t.Provider), t.ID)
 }
 
 func (s *Server) enqueueForkFileGCTaskRefs(ctx context.Context, t *meta.Tenant, store *datastore.Store) error {
--- a/pkg/server/fork.go
+++ b/pkg/server/fork.go
@@ -1021,6 +1021,7 @@ func (s *Server) cleanupForkTenantOnce(ctx context.Context, tenantID string, cre
 			if err := s.meta.UpdateTenantStatus(ctx, tenantID, meta.TenantDeleted); err != nil {
 				return fmt.Errorf("mark fork deleted: %w", err)
 			}
+			s.broadcastTenantMetricsCleanup(tenantID)
 		}
 		return nil
 	}
@@ -1056,6 +1057,7 @@ func (s *Server) cleanupForkTenantOnce(ctx context.Context, tenantID string, cre
 	if err := s.meta.UpdateTenantStatus(ctx, tenantID, meta.TenantDeleted); err != nil {
 		return fmt.Errorf("mark fork deleted: %w", err)
 	}
+	s.broadcastTenantMetricsCleanup(tenantID)
 	return nil
 }
 
@@ -1066,6 +1068,7 @@ func (s *Server) cleanupFailedForkBranch(ctx context.Context, t *meta.Tenant, cr
 	if err := s.meta.UpdateTenantStatus(ctx, t.ID, meta.TenantDeleted); err != nil {
 		return fmt.Errorf("mark failed fork deleted: %w", err)
 	}
+	s.broadcastTenantMetricsCleanup(t.ID)
 	return nil
 }
 
--- a/pkg/server/fork.go
+++ b/pkg/server/fork.go
@@ -1021,7 +1021,7 @@ func (s *Server) cleanupForkTenantOnce(ctx context.Context, tenantID string, cre
 			if err := s.meta.UpdateTenantStatus(ctx, tenantID, meta.TenantDeleted); err != nil {
 				return fmt.Errorf("mark fork deleted: %w", err)
 			}
-			s.broadcastTenantMetricsCleanup(tenantID)
+			s.clearLocalTenantMetrics(tenantID)
 		}
 		return nil
 	}
@@ -1057,7 +1057,7 @@ func (s *Server) cleanupForkTenantOnce(ctx context.Context, tenantID string, cre
 	if err := s.meta.UpdateTenantStatus(ctx, tenantID, meta.TenantDeleted); err != nil {
 		return fmt.Errorf("mark fork deleted: %w", err)
 	}
-	s.broadcastTenantMetricsCleanup(tenantID)
+	s.clearLocalTenantMetrics(tenantID)
 	return nil
 }
 
@@ -1068,7 +1068,7 @@ func (s *Server) cleanupFailedForkBranch(ctx context.Context, t *meta.Tenant, cr
 	if err := s.meta.UpdateTenantStatus(ctx, t.ID, meta.TenantDeleted); err != nil {
 		return fmt.Errorf("mark failed fork deleted: %w", err)
 	}
-	s.broadcastTenantMetricsCleanup(t.ID)
+	s.clearLocalTenantMetrics(t.ID)
 	return nil
 }
 
--- a/pkg/server/notify_coalescer.go
+++ b/pkg/server/notify_coalescer.go
@@ -37,7 +37,7 @@ const tenantNotifyFlushRetryBackoff = 100 * time.Millisecond
 //   - Poller cursor semantics (id > cursor ORDER BY id) are unaffected
 //     because each batch INSERT commits all of its rows atomically.
 //   - A failed flush retries the batch once, then falls back to independent
-//     per-row inserts; a row that still fails is logged and dropped. The
+//     per-row inserts; a row that still fails is logged and re-queued. The
 //     5-minute safety-net scan backstops only semantic/file_gc work — it
 //     never reads tenant_notify_outbox — so SSE cross-pod delivery relies on
 //     this retry → per-row fallback. Residual signal loss occurs only under
@@ -57,8 +57,10 @@ type tenantNotifyCoalescer struct {
 	// batch that will never flush.
 	stopped bool
 
-	mu      sync.Mutex
-	pending map[string]int // tenantID → OR-merged work_mask
+	flushMu  sync.Mutex // serializes periodic, shutdown, and test-triggered flushes
+	mu       sync.Mutex
+	pending  map[string]int // tenantID → OR-merged work_mask
+	inFlight map[string]int // batch swapped out but not yet durably inserted
 
 	cancel context.CancelFunc
 	wg     sync.WaitGroup
@@ -98,7 +100,7 @@ func (c *tenantNotifyCoalescer) add(tenantID string, workMask int) {
 		return
 	}
 	c.pending[tenantID] |= workMask
-	pending := len(c.pending)
+	pending := c.pendingCountLocked()
 	c.mu.Unlock()
 	metrics.RecordNotifyCoalescerPending(pending)
 }
@@ -140,6 +142,8 @@ func (c *tenantNotifyCoalescer) stop() {
 // inserted individually, so a metadb hiccup degrades to roughly the
 // pre-coalescer per-event behavior instead of dropping the whole window.
 func (c *tenantNotifyCoalescer) flush(ctx context.Context) {
+	c.flushMu.Lock()
+	defer c.flushMu.Unlock()
 	c.mu.Lock()
 	if len(c.pending) == 0 {
 		c.mu.Unlock()
@@ -147,8 +151,10 @@ func (c *tenantNotifyCoalescer) flush(ctx context.Context) {
 	}
 	batch := c.pending
 	c.pending = make(map[string]int)
+	c.inFlight = batch
+	pending := c.pendingCountLocked()
 	c.mu.Unlock()
-	metrics.RecordNotifyCoalescerPending(0)
+	metrics.RecordNotifyCoalescerPending(pending)
 
 	entries := make([]meta.TenantNotifyEntry, 0, len(batch))
 	for tenantID, workMask := range batch {
@@ -156,6 +162,7 @@ func (c *tenantNotifyCoalescer) flush(ctx context.Context) {
 	}
 	if err := c.insertBatch(ctx, entries); err == nil {
 		metrics.RecordNotifyCoalescerFlush("ok", len(entries))
+		c.finishFlush(nil)
 		return
 	} else {
 		logger.Warn(ctx, "tenant_notify_coalescer_flush_failed",
@@ -165,10 +172,11 @@ func (c *tenantNotifyCoalescer) flush(ctx context.Context) {
 	select {
 	case <-ctx.Done():
 		// No room for the batch retry; go straight to per-row inserts (they
-		// fail fast on the cancelled context and are dropped per row).
+		// fail fast on the cancelled context and are re-queued per row).
 	case <-time.After(tenantNotifyFlushRetryBackoff):
 		if err := c.insertBatch(ctx, entries); err == nil {
 			metrics.RecordNotifyCoalescerFlush("retry_ok", len(entries))
+			c.finishFlush(nil)
 			return
 		} else {
 			logger.Warn(ctx, "tenant_notify_coalescer_flush_retry_failed",
@@ -177,15 +185,17 @@ func (c *tenantNotifyCoalescer) flush(ctx context.Context) {
 		}
 	}
 	metrics.RecordNotifyCoalescerFlush("fallback", len(entries))
-	c.insertPerRow(ctx, entries)
+	c.finishFlush(c.insertPerRow(ctx, entries))
 }
 
-// insertPerRow delivers each entry independently; a failing row is logged
-// and dropped without affecting the rest.
-func (c *tenantNotifyCoalescer) insertPerRow(ctx context.Context, entries []meta.TenantNotifyEntry) {
+// insertPerRow delivers each entry independently and returns rows that still
+// failed so the caller can merge them back into the next pending batch.
+func (c *tenantNotifyCoalescer) insertPerRow(ctx context.Context, entries []meta.TenantNotifyEntry) []meta.TenantNotifyEntry {
+	failed := make([]meta.TenantNotifyEntry, 0)
 	for _, e := range entries {
 		if err := c.insertSingle(ctx, e.TenantID, e.WorkMask); err != nil {
 			metrics.RecordNotifyCoalescerPerRowFallback("error")
+			failed = append(failed, e)
 			logger.Warn(ctx, "tenant_notify_coalescer_row_insert_failed",
 				zap.String("tenant_id", e.TenantID),
 				zap.Int("work_mask", e.WorkMask),
@@ -194,4 +204,26 @@ func (c *tenantNotifyCoalescer) insertPerRow(ctx context.Context, entries []meta
 			metrics.RecordNotifyCoalescerPerRowFallback("ok")
 		}
 	}
+	return failed
+}
+
+func (c *tenantNotifyCoalescer) finishFlush(failed []meta.TenantNotifyEntry) {
+	c.mu.Lock()
+	for _, e := range failed {
+		c.pending[e.TenantID] |= e.WorkMask
+	}
+	c.inFlight = nil
+	pending := c.pendingCountLocked()
+	c.mu.Unlock()
+	metrics.RecordNotifyCoalescerPending(pending)
+}
+
+func (c *tenantNotifyCoalescer) pendingCountLocked() int {
+	count := len(c.inFlight)
+	for tenantID := range c.pending {
+		if _, exists := c.inFlight[tenantID]; !exists {
+			count++
+		}
+	}
+	return count
 }
--- a/pkg/server/notify_coalescer.go
+++ b/pkg/server/notify_coalescer.go
@@ -9,6 +9,7 @@ import (
 
 	"github.com/mem9-ai/drive9/pkg/logger"
 	"github.com/mem9-ai/drive9/pkg/meta"
+	"github.com/mem9-ai/drive9/pkg/metrics"
 )
 
 // defaultTenantNotifyFlushInterval is how often the coalescer flushes merged
@@ -97,7 +98,9 @@ func (c *tenantNotifyCoalescer) add(tenantID string, workMask int) {
 		return
 	}
 	c.pending[tenantID] |= workMask
+	pending := len(c.pending)
 	c.mu.Unlock()
+	metrics.RecordNotifyCoalescerPending(pending)
 }
 
 func (c *tenantNotifyCoalescer) run(ctx context.Context) {
@@ -145,12 +148,14 @@ func (c *tenantNotifyCoalescer) flush(ctx context.Context) {
 	batch := c.pending
 	c.pending = make(map[string]int)
 	c.mu.Unlock()
+	metrics.RecordNotifyCoalescerPending(0)
 
 	entries := make([]meta.TenantNotifyEntry, 0, len(batch))
 	for tenantID, workMask := range batch {
 		entries = append(entries, meta.TenantNotifyEntry{TenantID: tenantID, WorkMask: workMask})
 	}
 	if err := c.insertBatch(ctx, entries); err == nil {
+		metrics.RecordNotifyCoalescerFlush("ok", len(entries))
 		return
 	} else {
 		logger.Warn(ctx, "tenant_notify_coalescer_flush_failed",
@@ -163,6 +168,7 @@ func (c *tenantNotifyCoalescer) flush(ctx context.Context) {
 		// fail fast on the cancelled context and are dropped per row).
 	case <-time.After(tenantNotifyFlushRetryBackoff):
 		if err := c.insertBatch(ctx, entries); err == nil {
+			metrics.RecordNotifyCoalescerFlush("retry_ok", len(entries))
 			return
 		} else {
 			logger.Warn(ctx, "tenant_notify_coalescer_flush_retry_failed",
@@ -170,6 +176,7 @@ func (c *tenantNotifyCoalescer) flush(ctx context.Context) {
 				zap.Error(err))
 		}
 	}
+	metrics.RecordNotifyCoalescerFlush("fallback", len(entries))
 	c.insertPerRow(ctx, entries)
 }
 
@@ -178,10 +185,13 @@ func (c *tenantNotifyCoalescer) flush(ctx context.Context) {
 func (c *tenantNotifyCoalescer) insertPerRow(ctx context.Context, entries []meta.TenantNotifyEntry) {
 	for _, e := range entries {
 		if err := c.insertSingle(ctx, e.TenantID, e.WorkMask); err != nil {
+			metrics.RecordNotifyCoalescerPerRowFallback("error")
 			logger.Warn(ctx, "tenant_notify_coalescer_row_insert_failed",
 				zap.String("tenant_id", e.TenantID),
 				zap.Int("work_mask", e.WorkMask),
 				zap.Error(err))
+		} else {
+			metrics.RecordNotifyCoalescerPerRowFallback("ok")
 		}
 	}
 }
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -421,6 +421,7 @@ func NewWithConfig(cfg Config) *Server {
 	if logger == nil {
 		logger, _ = zap.NewProduction()
 	}
+	metrics.SetFeatureEnabled("vault", len(cfg.VaultMasterKey) > 0)
 	metrics.SetModuleAvailability("vault", false)
 	var vaultMK *vault.MasterKey
 	if len(cfg.VaultMasterKey) > 0 {
@@ -792,6 +793,18 @@ func (s *Server) insertTenantNotify(tenantID string, workMask int) {
 			zap.Int("work_mask", workMask),
 			zap.Error(err))
 	}
+}
+
+// broadcastTenantMetricsCleanup immediately clears this process and emits a
+// durable lifecycle signal so every pod clears historical tenant-scoped
+// series. Callers must invoke this only after a deleting/deleted transition or
+// cleanup job has been durably committed.
+func (s *Server) broadcastTenantMetricsCleanup(tenantID string) {
+	if tenantID == "" {
+		return
+	}
+	metrics.DeleteTenantCounters(tenantID)
+	s.insertTenantNotify(tenantID, WorkMetricsCleanup)
 }
 
 // startNotifyInfrastructure launches the unified tenant outbox components
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -795,11 +795,11 @@ func (s *Server) insertTenantNotify(tenantID string, workMask int) {
 	}
 }
 
-// broadcastTenantMetricsCleanup immediately clears this process after a
+// clearLocalTenantMetrics immediately clears this process after a
 // deleting/deleted lifecycle transition. The meta-store transition writes the
 // WorkMetricsCleanup outbox row in the same transaction, so this helper must
 // not enqueue a second best-effort signal through the coalescer.
-func (s *Server) broadcastTenantMetricsCleanup(tenantID string) {
+func (s *Server) clearLocalTenantMetrics(tenantID string) {
 	if tenantID == "" {
 		return
 	}
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -2063,13 +2063,13 @@ func (s *Server) handleTenantStatus(w http.ResponseWriter, r *http.Request) {
 		errJSON(w, http.StatusUnauthorized, "invalid API key")
 		return
 	}
-	setRequestMetricTenant(r.Context(), resolved.Tenant.ID, resolved.APIKey.ID, resolved.Tenant.Provider, resolved.TiDBCloudOrgID, classifyTenantRequest(r))
 	if resolved.APIKey.Status != meta.APIKeyActive {
 		metricEvent(r.Context(), "auth", "result", "key_inactive")
 		logger.Warn(r.Context(), "server_event", eventFields(r.Context(), "tenant_status_key_inactive", "tenant_id", resolved.Tenant.ID, "api_key_id", resolved.APIKey.ID, "status", resolved.APIKey.Status)...)
 		errJSON(w, http.StatusUnauthorized, "invalid API key")
 		return
 	}
+	setRequestMetricTenantForAuthStatus(r.Context(), resolved.Tenant.ID, resolved.APIKey.ID, resolved.Tenant.Provider, resolved.TiDBCloudOrgID, resolved.Tenant.Status, classifyTenantRequest(r))
 	plain, err := poolDecryptToken(r.Context(), s.pool, resolved.APIKey.JWTCiphertext)
 	if err != nil {
 		metricEvent(r.Context(), "auth", "result", "decrypt_failed")
--- a/pkg/server/server.go
+++ b/pkg/server/server.go
@@ -292,7 +292,7 @@ type Server struct {
 	// Unified tenant outbox components. The tenantOutboxPoller reads the
 	// central tenant_notify_outbox table (in the always-provisioned meta DB)
 	// on every pod and dispatches by work_mask: SSE bits wake the local bus;
-	// semantic/file_gc/quota bits kick the tenantWorker on the shard owner.
+	// semantic/file_gc bits kick the tenantWorker on the shard owner.
 	// The shardResolver determines shard ownership via jump consistent hashing
 	// over the active pod ring. The podRegistry maintains this pod's presence
 	// and subscriber set in the central DB. All run on every pod (not
@@ -795,25 +795,24 @@ func (s *Server) insertTenantNotify(tenantID string, workMask int) {
 	}
 }
 
-// broadcastTenantMetricsCleanup immediately clears this process and emits a
-// durable lifecycle signal so every pod clears historical tenant-scoped
-// series. Callers must invoke this only after a deleting/deleted transition or
-// cleanup job has been durably committed.
+// broadcastTenantMetricsCleanup immediately clears this process after a
+// deleting/deleted lifecycle transition. The meta-store transition writes the
+// WorkMetricsCleanup outbox row in the same transaction, so this helper must
+// not enqueue a second best-effort signal through the coalescer.
 func (s *Server) broadcastTenantMetricsCleanup(tenantID string) {
 	if tenantID == "" {
 		return
 	}
 	metrics.DeleteTenantCounters(tenantID)
-	s.insertTenantNotify(tenantID, WorkMetricsCleanup)
 }
 
 // startNotifyInfrastructure launches the unified tenant outbox components
 // that run on every pod (not leader-gated):
 //   - shardResolver: refreshes the active pod ring for jump-consistent-hash
-//     shard ownership of semantic/file_gc/quota work.
+//     shard ownership of semantic/file_gc work.
 //   - tenantOutboxPoller: reads the central tenant_notify_outbox table at
 //     200ms intervals and dispatches by work_mask (SSE → wake local bus;
-//     semantic/file_gc/quota → kick tenantWorker on shard owner).
+//     semantic/file_gc → kick tenantWorker on shard owner).
 //   - podRegistry: maintains this pod's presence in pod_registry and reports
 //     its SSE subscriber tenant set to pod_subscriptions.
 //
--- a/pkg/server/tenant_metrics.go
+++ b/pkg/server/tenant_metrics.go
@@ -106,8 +106,8 @@ func (s *Server) observeSharedDBPoolMetrics(ctx context.Context) {
 		}
 		metrics.RecordSharedDBPoolTotal(orgID, snapshot.ID, snapshot.UUID, snapshot.Status, 1)
 		next[totalKey] = struct{}{}
-		if snapshot.Status != meta.SharedDBStatusActive && !snapshot.UpdatedAt.IsZero() {
-			age := time.Since(snapshot.UpdatedAt)
+		if snapshot.Status != meta.SharedDBStatusActive && !snapshot.StatusUpdatedAt.IsZero() {
+			age := time.Since(snapshot.StatusUpdatedAt)
 			if age < 0 {
 				age = 0
 			}
--- a/pkg/server/tenant_metrics.go
+++ b/pkg/server/tenant_metrics.go
@@ -18,10 +18,11 @@ type tenantPoolBindingMetricKey struct {
 }
 
 const (
-	sharedDBPoolMetricTotal    = "total"
-	sharedDBPoolMetricCapacity = "capacity"
-	sharedDBPoolMetricTenants  = "tenants"
-	sharedDBPoolMetricSpending = "spending_limit"
+	sharedDBPoolMetricTotal     = "total"
+	sharedDBPoolMetricCapacity  = "capacity"
+	sharedDBPoolMetricTenants   = "tenants"
+	sharedDBPoolMetricSpending  = "spending_limit"
+	sharedDBPoolMetricStatusAge = "status_age"
 )
 
 type sharedDBPoolMetricKey struct {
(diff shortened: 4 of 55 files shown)


srstack:
Regression update:

- make test exercised the full suite. All packages outside cmd/drive9/cli passed; the command exits non-zero on two existing CLI failures: TestCtxImport_RejectsWorldReadableFile and TestCpLayerLocalUploadUsesFSLayerEntry.
- Both CLI failures reproduce unchanged on the current main checkout with the same focused command, and this PR has no cmd/drive9/cli changes.
- go test ./... -run ^$ -count=1 passed for the full repository.
- make lint passed with 0 issues after removing one unused bounds declaration.
- pkg/metrics was rerun after that cleanup and passed.
- tidbcloud/runbooks#2688: make check-drive9-alert-rules and the full make check-rules validation pass.

GitHub CI and local-e2e are still running on the latest head 265a59d3.

srstack:
Final regression update for head `265a59d3`:

- `umask 022; make test` completed successfully: the full Drive9 Go test suite passed.
- `go test ./... -run "^$" -count=1` passed for the full repository.
- `make lint` passed with `0 issues`.
- The two failures mentioned in the earlier comment were caused by this machine running under `umask 077`: tests that create files with mode `0644` observed `0600`. They reproduced unchanged on `main`, and both focused tests passed under the standard CI `umask 022`.
- In `tidbcloud/runbooks#2688`, both `make check-drive9-alert-rules` and the full `make check-rules` passed.
- GitHub checks are now green: Drive9 `ci`, `drive9-server-local e2e smoke`, `CodeRabbit`, and runbooks `check-rules` all passed.

srstack (commented):
## Review

Verified locally on `fix/metrics-cardinality-observability` (`8c11d091`) in an isolated worktree: `go build ./...` clean, `go vet` clean on the touched trees, and `pkg/metrics`, `pkg/server`, `pkg/backend`, `pkg/meta`, `pkg/tenant/tidbcloudnative` all pass with `-count=1`. I also cross-checked the alert side (runbooks `c52b6b7b`): every `drive9_*` metric referenced by the rules now exists in code, and the four previously false-green alerts were replaced rather than silently dropped.

Good change overall. It closes the real gaps — false-green alerts, the `Set(0)`-instead-of-`Delete` leak class, and the five uninstrumented critical paths. The cross-pod cleanup via a new outbox work bit is a better mechanism than a per-pod TTL sweep. Findings below, most severe first.

---

### Blocking

**1. `WorkMetricsCleanup = 8` has no compile-time assertion and no backend counterpart.**

`pkg/server/work_mask.go` adds bit 3, but the `var _ = [1]byte{}[...]` assertion block still only covers SSE/Semantic/FileGC. The file's own doc comment says those assertions "ensure the values stay in sync" — bit 3 opts out of that guarantee. `backend.BackendWork*` stops at `4`, so a future `BackendWorkQuota = 8` in `pkg/backend/dat9.go` would collide with metrics-cleanup and the poller would run `DeleteTenantCounters` on every quota kick. The naming drift is already visible: the comment at the top of `work_mask.go` refers to a "Quota" bit that is not defined anywhere. Please either reserve bit 3 in the backend const block (even as a commented-out placeholder) or add an assertion that it is unclaimed.

**2. `DeleteTenantQuotaSnapshot` on the quota-check hot path is both a contention and a correctness regression.**

`pkg/backend/quota.go` calls it at the top of `recordTenantQuotaSnapshot`, which runs from three hot paths — `file_count_check`, `storage_check`, and `mediaLLMQuotaExceededServerTx` — i.e. effectively on every write.

- *Cost*: each call takes the registry write lock three times and scans **every label set of all three gauges** (`deleteGaugeByLabel` in `pkg/metrics/registry.go`). That is O(total tenants), not O(1). At N tenants × write QPS it is an N² sweep under a process-global mutex on the write path — the opposite of this PR's goal.
- *Correctness*: the delete removes `confirmed`/`reserved`/`limit`, then re-adds `confirmed`/`reserved`, and re-adds `limit` only `if cfg.Max* > 0`. A `/metrics` scrape landing between the delete and the re-add sees the tenant's series missing entirely — a disappearing denominator for `Drive9Tenant*QuotaNearLimit`.
- *Call sites fight each other*: `storage_check` reaches the snapshot with a cfg whose media/video limits may be 0, so it deletes the media limit series a media check just wrote, and vice versa.

`DeleteTenantQuotaLimitsWithOrg` (added in this PR, unit-tested, currently **dead code** — no non-test caller) is the targeted fix: delete only the three `state="limit"` series, and only when the corresponding limit is absent. O(1), keeps `confirmed`/`reserved` continuously present, no registry scan.

---

### Should fix

**3. `drive9_db_pool_connections` role-only aggregate guard is dead.**

In `pkg/metrics/db.go` the synthetic per-role aggregate is guarded by `if _, exists := poolTotals[dbMetricKey{role: role}]`. That zero-value key has `tenantID`/`dbPoolUUID`/`tidbCloudOrgID` as `""`, while every real key goes through `metricKey()`, which normalizes them to `"unknown"`/`"guest"`. The lookup can never hit. Today nothing duplicates, because `OpenInstrumentedForTenant` always supplies a tenantID for user roles — but the guard does not protect what its comment claims, and `Drive9DBConnectionPoolSaturation` (unchanged, matches bare `state=` with no role filter) would divide across two different series if an unscoped role pool were ever registered. Build the probe key with `metricKey()`, or drop the guard and state the invariant.

**4. `Drive9TenantOutboxBacklogAgeHigh` will flap.**

`RecordTenantOutboxPoll` only sets `backlog_oldest_age_seconds` when `full` is true (batch size exactly `tenantOutboxBatchSize`) and sets it to `0` otherwise. A backlog draining to `batchSize - 1` resets the gauge to 0 while rows are still pending. With `for: 5m` on `> 30` at critical, a steady near-full backlog oscillates and never holds the window. Suggest reporting the age whenever `len(rows) > 0` and gating the alert on the age alone.

**5. `RecordAPIKeyResolveCache` calls `entryCount()` on every auth request.**

`pkg/meta/api_key_cache.go` takes the cache mutex per hit/miss just to read `len(c.byHash)` for a gauge — on the exact hot path this cache exists to relieve (per its own comment, 76M execs/12h). `fill` and `evictTenant` already publish the count via `RecordAPIKeyResolveCacheEntries`, and the count only changes there, so the per-request read is redundant. Use the counter-only variant on the request path.

Related: the `hits`/`misses` atomics are now superseded by `drive9_api_key_resolve_cache_requests_total` and are unread outside tests — worth deleting.

**6. Global `semantic_worker` gauges still write 0 and remain unalerted.**

`pkg/server/tenant_worker.go` sets `queued`, `processing`, `dead_lettered`, `queue_lag_seconds` to 0 at start and never updates them (only `inflight`/`workers` are maintained). The per-tenant path was correctly switched to `Delete` and the two dead global alerts were correctly removed — but four permanently-zero global series remain. Either wire them or drop them; a constant-zero series invites someone to alert on it later.

---

### Minor

**7. `_ = param` placeholders leave a misleading API surface.** `pkg/metrics/operations.go` now carries `_ = result`, `_ = surface`, `_, _ = surface, action`, `_ = action`, and `_, _ = tenantID, tidbCloudOrgID`; `pkg/server/instrumentation.go` has `_ = action` in `tenantInFlightKey`. Callers still compute and pass values that are discarded. Acceptable as a mechanical first pass, but the signatures should be narrowed in a follow-up or the dropped args will get quietly re-plumbed.

**8. `Drive9QuotaReservationLockConflictRetryHigh` loses its attribution.** The alert still groups by `tenant_id, tidbcloud_org_id` over `drive9_service_operations_total`, but that counter no longer carries those labels, so the grouping silently collapses. `lock_conflict_retry` is also absent from `isOperationFailureResult`, so it does not land in `drive9_tenant_operation_failures_total` either. This does resolve the `guest`-org bug I flagged earlier — but by removing attribution rather than fixing it. Worth deciding explicitly which behavior you want.

**9. `tidbcloudOperationForRequest` returns `"unknown"` for unmatched paths.** It is only reached from `doDigestAuthRequest`, and the per-method call sites pass explicit constants, so `unknown` should be unreachable. A one-line comment saying so would keep it from reading like an unbounded label source.

**10. `Drive9VaultUnavailable` correctness depends on `pod` being attached to `drive9_feature_enabled`.** `SetFeatureEnabled` sets no pod label; the join relies on the scrape adding it. It does for `module_up`, so this is very likely fine — worth confirming in one environment before relying on the `and on (namespace, pod)` join.

---

### Not addressed (carried over, not necessarily for this PR)

- **k8s-level alerts still hardcode `namespace=~"dat9|drive9-tidbcloud"`.** The 7-environment namespace/container audit hasn't happened, so several cluster-level rules may be inert outside those two namespaces.
- **`DeleteTenantCounters` on idle pool reap** (`pkg/tenant/pool.go`) is unchanged — `pool.go` isn't in this diff. The new outbox broadcast fixes cleanup on tenant *delete*, which is the larger win, but a 2-minute idle reap still deletes counters for live tenants and makes `increase()` under-count on per-tenant alerts. Now that `WorkMetricsCleanup` exists, scoping counter deletion to lifecycle transitions only would be a small follow-up.
- **Tenant pool ref-leak detection** remains uncovered, as documented in the rules file.

pkg/server/tenant_worker.go
```diff
@@ -599,15 +619,30 @@ func (m *tenantWorkerManager) observeTenant(ctx context.Context, target *tenantT
...
 		if tenantLag < 0 {
 			tenantLag = 0
 		}
 	}
-	metrics.RecordTenantGaugeWithOrg(target.tenantID, target.metricOrgID(), "semantic_worker", "queue_lag_seconds", tenantLag)
+	if tenantLag > 0 {
+		metrics.RecordTenantGaugeWithOrg(tenantID, tidbCloudOrgID, "semantic_worker", "queue_lag_seconds", tenantLag)
+	} else {
+		metrics.DeleteTenantGaugeWithOrg(tenantID, tidbCloudOrgID, "semantic_worker", "queue_lag_seconds")
+	}
```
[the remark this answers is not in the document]
srstack: Fixed in `23a8ec6b`. `tenantWorkerManager` now tracks the last semantic metric org per tenant, deletes the previous-org `dead_lettered` and `queue_lag_seconds` series before recording the new org, and clears the bookkeeping from `ForgetTenant`. `TestTenantWorkerSemanticObservationDeletesPreviousOrgSeries` covers guest/org relabel cleanup.

pkg/server/tenant_outbox_poller.go
```diff
@@ -170,8 +175,10 @@ func (p *tenantOutboxPoller) pollOnce(ctx context.Context) {
 		if ctx.Err() != nil {
 			return
 		}
+		pollStart := time.Now()
 		rows, err := p.metaStore.ListTenantNotifySince(ctx, p.lastID, tenantOutboxBatchSize)
 		if err != nil {
+			metrics.RecordTenantOutboxPoll(metrics.ResultForError(err), time.Since(pollStart), 0, 0, false)
```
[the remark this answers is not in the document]
srstack: Fixed in `23a8ec6b`. `RecordTenantOutboxPoll` now leaves the last observed backlog age unchanged for non-`ok` polls, reports oldest age for every successful non-empty batch (not only full batches), and clears it only after a successful empty poll. The three transitions are covered by `TestTenantOutboxBacklogAgeTracksObservedRowsAndSurvivesPollErrors`.

pkg/server/server.go
```diff
@@ -794,6 +795,18 @@ func (s *Server) insertTenantNotify(tenantID string, workMask int) {
...
+// broadcastTenantMetricsCleanup immediately clears this process and emits a
+// durable lifecycle signal so every pod clears historical tenant-scoped
+// series. Callers must invoke this only after a deleting/deleted transition or
+// cleanup job has been durably committed.
+func (s *Server) broadcastTenantMetricsCleanup(tenantID string) {
+	if tenantID == "" {
+		return
+	}
+	metrics.DeleteTenantCounters(tenantID)
+	s.insertTenantNotify(tenantID, WorkMetricsCleanup)
```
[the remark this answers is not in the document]
srstack: Fixed in `23a8ec6b`. The persisted work-bit allocation now lives in `pkg/meta`, and every deleting/deleted transition path writes `TenantNotifyWorkMetricsCleanup` to `tenant_notify_outbox` in the same MetaDB transaction as the lifecycle update. The server helper now performs only immediate local cleanup, so it no longer depends on the best-effort coalescer for cross-pod delivery. Meta and end-to-end tenant deletion tests verify the durable row.

pkg/server/notify_coalescer.go
```diff
@@ -145,12 +148,14 @@ func (c *tenantNotifyCoalescer) flush(ctx context.Context) {
 	batch := c.pending
 	c.pending = make(map[string]int)
 	c.mu.Unlock()
+	metrics.RecordNotifyCoalescerPending(0)
```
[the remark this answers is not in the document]
srstack: Fixed in `23a8ec6b`. The coalescer now tracks the swapped batch in `inFlight`, keeps the pending gauge at the union of queued and in-flight tenant IDs, and clears it only after durable batch/retry success. Per-row failures are re-merged into the next pending batch instead of being dropped. Blocking-insert and fallback-retry tests cover both states.

srstack:
Review follow-up on head `23a8ec6b`:

Addressed:

1. Work-mask allocation now has a single persisted source in `pkg/meta`; server/backend aliases include the metrics-cleanup bit and retain compile-time equality checks.
2. Quota snapshots no longer scan every tenant series or expose a delete/re-add scrape gap. `ReplaceTenantQuotaSnapshotWithOrg` performs an O(1), single-registry-lock replacement, preserves known limits when config is unavailable, and removes prior-org labels exactly.
3. The mixed scoped/unscoped DB pool guard now probes the normalized `metricKey`, with a duplicate-series regression test.
4. Tenant outbox backlog age is retained across failed polls, reported for every successful non-empty batch, and cleared only by a successful empty poll.
5. API-key cache request accounting is counter-only on hit/miss. Entry gauges update only on fill, eviction, or lazy TTL expiry; the redundant request-path `entryCount` lock and test-only atomics were removed.
6. The four never-maintained global semantic queue gauges were removed. Per-tenant semantic gauges now delete the previous-org series on org relabel.
7. Notify coalescer pending now includes in-flight tenant IDs until durable success; failed per-row fallbacks are re-queued rather than dropped.
8. Metrics cleanup outbox delivery is transactionally coupled to every deleting/deleted MetaDB lifecycle transition; the server helper now performs local cleanup only.
9. Added the bounded-fallback comment for the unreachable TiDB Cloud `unknown` operation path.

Reviewed but intentionally unchanged:

- The discarded-parameter signature narrowing is mechanical across 90+ call sites and was explicitly identified as follow-up-safe; keeping it out avoids unrelated churn in this already broad PR.
- `Drive9QuotaReservationLockConflictRetryHigh` is already aggregate-only in runbooks `c52b6b7b` (`sum by (namespace)` over `service_operations_total`), so the tenant-grouping issue does not exist on the current alert branch.
- `Drive9VaultUnavailable` joins two metrics from the same scrape target. Prometheus attaches `namespace` and `pod` target labels to both; neither producer writes a pod label itself, so the join contract is consistent.
- Namespace audit, idle-pool counter deletion, and tenant ref-leak detection remain the explicitly carried-over follow-ups.

Validation on the pushed head:

- `umask 022; make test` passed.
- `go test ./... -run "^$" -count=1` passed.
- `make lint` passed with `0 issues`.
- All four inline threads have targeted regression tests, replies, and are resolved.

GitHub CI and local E2E are running on this head.

srstack:
Review state rechecked against the current head 23a8ec6b. GitHub currently reports no unresolved review threads and no additional formal review or inline comment beyond the four threads already fixed and resolved. No code change was made for this refresh because there is no newly retrievable technical finding to evaluate. The items intentionally left unchanged, with their technical reasons, remain documented in https://github.com/mem9-ai/drive9/pull/802#issuecomment-5084892762. If another review is still propagating or is visible only through a specific discussion link, it will be evaluated against the current head when that comment becomes available; actionable findings will be fixed, and rejected findings will receive an explanation in their own thread.

srstack:
Follow-up corrections are pushed in c572e58e.

1. The discarded metric parameters were removed instead of retained as compatibility placeholders. Tenant request result, HTTP/file surface/action, inflight action, and fs_events tenant/org arguments were removed from the APIs and all callers. The now-dead tenantRequestResult helper was also removed.
2. broadcastTenantMetricsCleanup was renamed to clearLocalTenantMetrics across all 14 current call sites, matching its process-local behavior after durable outbox insertion moved into MetaDB lifecycle transitions.
3. The deleting/deleted UpdateTenantStatus and UpdateTenantStatusIf paths intentionally retain one transaction per tenant. The batch cleanup loops interleave external cluster deprovisioning, pool draining, API-key revocation, and other work; one transaction across the loop would hold DB locks across external I/O and couple independent tenant failures into a single rollback. The per-tenant transaction preserves status plus metrics-cleanup-outbox atomicity and failure isolation. This behavior and rationale are now documented on both Store methods and in the c572e58e commit message.

This supersedes the earlier decision to leave signature narrowing as follow-up work.

Per the requested submit-first order, the commit was pushed before waiting for complete regression. Before interruption, pkg/metrics, pkg/backend, and pkg/meta passed; pkg/server was still running and was deliberately interrupted, so it is not being reported as passed.

srstack:
Remaining review follow-ups disposition:

1. **Idle pool counter deletion: fixed in `e13d5e98`.** `Pool.closeEntry` no longer calls `DeleteTenantCounters`, so idle/capacity/shutdown eviction of a live cached backend cannot reset tenant counters and undercount `increase()`. Tenant-series deletion is now owned by deleting/deleted lifecycle cleanup and its durable metrics-cleanup outbox delivery. `TestCloseEntryPreservesLiveTenantCounters` failed against the old behavior and passes after the fix.

2. **Seven-environment namespace/deployment/container audit: no rule edit is needed.** The current runbooks selectors are `namespace=~"dat9|drive9-tidbcloud"`, `deployment=~"dat9-server|drive9-server"`, and `container=~"dat9-server|drive9-server"`. The current environment inventory maps as follows:
   - prod-aws-sg, dev-aws-sg, prod-aws-use1: `dat9/dat9-server` and `drive9-tidbcloud/drive9-server`
   - prod-aws-usw2, prod-ali-sg, dev-tke-bj, prod-tke-bj: `drive9-tidbcloud/drive9-server` (prod-ali-sg also retains dat9 configuration)
   These names are fully covered by the existing regexes, so widening or changing them would not activate any additional Drive9 workload. dev-tke-bj currently has no vmagent; that is an observability deployment gap, not a selector-contract bug, and changing alert expressions would not make that environment scrape metrics.

3. **Tenant pool ref-leak alert: intentionally not added from the existing cache metrics.** `drive9_shared_db_pool_cache_tenants` is backed by `sharedDBRefs`, which counts cached tenant backends using a physical shared DB handle. It does not count outstanding `Pool.Acquire` leases (`entry.refs`). Therefore `cache_tenants > 0` is normal for warm live tenants and cannot distinguish a leaked request lease; combining it with traffic absence would still conflate healthy cache residency with a leak. A correct signal needs lease-level acquire/release tracking (for example oldest outstanding lease age or bounded anomaly-only state) plus hot-path overhead validation. Adding the suggested alert on the current metrics would create a false-positive-prone contract, so this PR leaves it out with this explicit rationale rather than shipping a misleading rule.

Focused verification before push: `go test ./pkg/tenant -run ^TestCloseEntryPreservesLiveTenantCounters$ -count=1` passed, and `git diff --check` passed. Full regression follows after comment handling, per the requested submit-first order.

srstack (commented):
LGTM.

All ten items from my earlier review are resolved on `e13d5e98`, including the three that were initially deferred or pushed back on. Two of those pushbacks were correct and I withdraw them: `Drive9QuotaReservationLockConflictRetryHigh` is already aggregate-only on runbooks `c52b6b7b`, so the tenant-grouping concern never applied to the current alert branch; and `drive9_feature_enabled` / `drive9_module_up` come from the same scrape target, so the `and on (namespace, pod)` join contract holds.

### What convinced me on the two blocking items

**Work-mask allocation.** Moving the bits to `pkg/meta` as `1 << iota` and reducing `backend.BackendWork*` / `server.Work*` to aliases is a better fix than the assertion I asked for — it removes the possibility of divergence rather than detecting it after the fact. `BackendWorkMetricsCleanup` existing solely to be asserted, with an explicit "backend does not emit it" comment, is the right call. The phantom "Quota" bit references in the surrounding comments are gone too.

**Quota snapshot path.** `ReplaceTenantQuotaSnapshotWithOrg` is O(1) with a single registry lock acquisition and exact-label writes; `deleteGaugeByLabel` is gone. Two things beyond what I raised are also correct: caching last-known limits in `Registry.tenantQuotaSnapshots` so the three call sites stop clobbering each other's `state="limit"` series, and deleting the prior org's label set on relabel. Because it is one critical section with no intermediate delete, the scrape-window gap is closed rather than narrowed.

### On the last two commits

`c572e58e` — the signature narrowing landed properly rather than being silenced. All seven `_ = param` placeholders are gone, `action` removal propagated correctly through `tenantInFlightKey` / `adjustTenantInFlight` / `setRequestMetricTenant`, and keeping `state.action = action` on the in-flight move path is right since `action` still feeds `tenant_requests_total`. Dropping `tenantRequestResult` once nothing consumed it is good hygiene. The transaction-boundary rationale on `UpdateTenantStatus` / `UpdateTenantStatusIf` is a better answer than batching — per-tenant transactions keeping failure isolation in loops that interleave external cleanup work is the correct tradeoff, and having it written down is what I wanted.

`e13d5e98` — this closes the idle-reap issue I had twice marked as out of scope. With `DeleteTenantCounters` removed from `Pool.closeEntry`, the only two call sites are the outbox cleanup-bit dispatch and `clearLocalTenantMetrics`, both lifecycle-driven, so a 2-minute idle reap no longer deletes counters for a live tenant and the `increase()` under-counting on per-tenant alerts is fixed. `TestCloseEntryPreservesLiveTenantCounters` pins it.

Also confirmed the metrics-cleanup outbox row is written transactionally at all nine `insertTenantNotifyTx` sites, so cleanup delivery cannot be lost between the status transition and the signal.

### Independent verification on `e13d5e98`

- `go build ./...` clean
- `go vet ./pkg/...` clean
- `pkg/metrics`, `pkg/server`, `pkg/backend`, `pkg/meta`, `pkg/tenant` all pass with `-count=1`
- `-race` clean on `pkg/metrics` and on the coalescer / outbox / semantic-worker tests in `pkg/server`
- Alert cross-check: every `drive9_*` metric referenced by runbooks `c52b6b7b` exists in code, and the four previously false-green alerts were replaced rather than dropped
- GitHub `ci` pass, `drive9-server-local e2e smoke` pass

### Remaining follow-ups (correctly out of scope here)

- The namespace/container audit for the k8s-level rules across all seven environments.
- Tenant-pool ref-leak detection, still documented as uncovered in the rules file.

Nothing blocking from me.

mornyx (approved):
Deep review across six slices (cardinality/lifecycle, new bounded metrics, OpenAPI contract, DB pool reduction, event-bus dedup, gauge deletion plumbing) with findings traced end-to-end. Approving: the failure-only per-tenant counter + whitelisted business events give a real cardinality bound; the WorkMetricsCleanup broadcast is transactional with every status transition, idempotent, and its cursor/prune semantics are correct; the OpenAPI api/operation/result contract has exactly-once, mutually-exclusive classification (verified against every return path of all 10 HTTPClustersAPI methods + the digest doer); gauge registry locking and label-key normalization are consistent. Three worth-fixing items and a handful of nits inline; also two deploy notes: the label rename (`path`→`api`, module `tidbcloud_quota`→`tidbcloud_openapi`) breaks existing dashboards keyed on the old shape, and `drive9_event_bus_publish_errors_total` had a live caller before this PR — any alert on it goes silent until the runbooks PR lands.

pkg/server/tenant_metrics.go
```diff
@@ -105,6 +106,17 @@ func (s *Server) observeSharedDBPoolMetrics(ctx context.Context) {
 		}
 		metrics.RecordSharedDBPoolTotal(orgID, snapshot.ID, snapshot.UUID, snapshot.Status, 1)
 		next[totalKey] = struct{}{}
+		if snapshot.Status != meta.SharedDBStatusActive && !snapshot.UpdatedAt.IsZero() {
+			age := time.Since(snapshot.UpdatedAt)
+			if age < 0 {
```
mornyx, line 111: **Worth-fixing — this measures time-since-last-row-write, not time-in-status.** `snapshot.UpdatedAt` comes from `db_pool.updated_at`, which is `ON UPDATE CURRENT_TIMESTAMP(3)` — any write (tenant_count bumps from placement completions, spending-limit syncs, soft_cap flips) resets the gauge to ~0 while status is unchanged, contradicting the HELP text ("remained in its current status"). A pool stuck in `provisioning` for hours with a concurrent placement completing sees age flap to zero, so a `status_age > N` alert will flap. It still works for a truly wedged pool (no writers), which is the primary target — but either document it as "last write age" or key it off a status-transition timestamp.

pkg/server/fork.go
mornyx, line 1: (ref: `pkg/server/fork.go:647`)

**Worth-fixing — role-level tenant saturation aggregate silently misreports during fork/delete flows.** This `datastore.Open(dsn)` (and the one in `openTenantStore` at :1080) registers an *unscoped* `role="user"` pool; while it exists, the dedup in `writeDBPrometheus` suppresses the tenant aggregate for that role entirely, so `drive9_db_pool_connections{role="user",state="in_use"|"max_open"}` reports only this single unscoped pool — the exact series kept for saturation alerting goes dark/misleading precisely during fork provisioning and fork-delete cleanup. Both sites have the tenant ID in scope; `datastore.OpenForTenant(ctx, dsn, forkID)` / the org variant fixes it in one line each. Related pre-existing nit: `DRIVE9_USER_DB_MAX_OPEN_CONNS=0` skips SetMaxOpenConns → unlimited pools → aggregate max_open=0 → ratio alerts divide by zero.

pkg/server/tenant_outbox_poller.go
mornyx, line 181: **Nit — error paths inject zeros into the batch-size histogram.** `RecordTenantOutboxPoll` observes the batch-size histogram unconditionally, so metadb hiccups (this branch passes size 0) drag `drive9_tenant_outbox_batch_size` percentiles down. The gauge/counter branches already skip on `result != "ok"`; the histogram should too.

pkg/tenant/tidbcloudnative/http_clusters_api.go
```diff
@@ -189,8 +211,10 @@ func (a *HTTPClustersAPI) DeleteCluster(ctx context.Context, publicKey, privateK
 	}
 	defer func() { _ = resp.Body.Close() }()
```
mornyx, line 212: **Nit — dead `responseValid` parameter + misclassified stray 2xx on delete.** Every call site passes `responseValid=true`, so the 2xx→`protocol_error` branch is unreachable. Side effect in `DeleteCluster`/`DeleteBranch`: a 2xx outside {200,204,404} (e.g. 202) returns an error to the caller but records `result="ok"`. Pass `false` on the delete paths or drop the parameter.

pkg/tenant/tidbcloudnative/openapi_metrics.go
```diff
@@ -0,0 +1,137 @@
...
+	"net/url"
+	"strings"
+
+	"github.com/mem9-ai/drive9/pkg/metrics"
+)
+
+const (
+	tidbCloudAPICluster = "cluster"
+	tidbCloudAPIIAM     = "iam"
+	tidbCloudAPIBilling = "billing"
```
mornyx: **Nit** — `tidbCloudAPIBilling` (and `tidbCloudOperationGetOrganizationPlan`) are never used anywhere; wire them or delete them. Also noting for triage: `ErrTiDBCloudRoleInsufficient` / credential-mismatch validations are deterministic per-tenant config errors but bucketed as `protocol_error`, and the legacy starter provisioner (`pkg/tenant/starter/legacy.go`) runs its own digest client outside this metric — both fine to leave, but worth a line in the PR description.

mornyx:
**Worth-fixing (anchored at `pkg/server/auth.go:258`, not in the diff so leaving it here) — late samples re-create the series the cleanup just deleted, and nothing re-cleans them.**

`DeleteTenantCounters` runs exactly once per pod when the cleanup outbox row is consumed, but `setRequestMetricTenant` (auth.go:258) runs for *every* request that resolves a key — before the key-active and tenant-status checks — so a client that keeps calling with a deleted tenant's key (FUSE mount retry loop, SDK heartbeat) re-creates `drive9_tenant_requests_total` + bytes + whitelisted business-event series on every pod it hits, indefinitely. Cross-pod, the auth resolve cache additionally lets other pods accept the deleted tenant's keys as active for up to the TTL and produce full series post-cleanup. The old backstop (`pool.closeEntry` → delete) was deliberately removed, so there is no second sweep — these series live until pod restart, defeating the mechanism for exactly the tenants with post-delete traffic.

Consider bucketing requests that fail auth/tenant-status without tenant attribution (e.g. `tenant_id="blocked"`), or having the poller retain recently-cleaned IDs and suppress/re-delete.

qiffang:
## adversary-2 review — GREEN / APPROVE (N1/N2 tradeoffs to confirm before merge)

**B1 coupling (the #1 question from #801): RESOLVED.** `pkg/tenant/pool.go` `closeEntry` removes the inline `metrics.DeleteTenantCounters` (-6). Nothing in `Pool.Acquire`/`Get`/eviction calls the full-registry scan now — #801's 20× resident-cap raise no longer amplifies a synchronous registry-lock scan on the request path. Deletion moved to (a) lifecycle transitions writing a `WorkMetricsCleanup` outbox row **in the same tx** as the status change (meta.go:3609/3672/4173/4236, db_pool.go:1619), and (b) `tenantOutboxPoller.dispatch` per pod (broadcast). Verified by `TestCloseEntryPreservesLiveTenantCounters`.

**Cross-pod cleanup correct:** cleanup bit is broadcast (not sharded); every pod's poller deletes local series idempotently. Delivery is now reliable — coalescer re-queues failed rows via `finishFlush` (fixes the #792 silent-drop). Outbox row durable/atomic with the transition. `TestTenantOutboxPollerMetricsCleanupIsBroadcast` covers it.

**New metrics bounded:** `workers.go` uses `shard` (0–3), enum `result`/`stage`/`status` via `cleanMetricValue`; `shared_db_pool_status_age` bounded by pool count. `TestCriticalWorkerMetricsUseBoundedLabels` asserts no `tenant_id`/`db_pool_id`. `append([]Attribute(nil), attrs...)` copies before adding tenant labels (operations.go:375). TiDB OpenAPI metrics (`api`/`result`/`transport`) carry no tenant dimension.

### Notes (confirm before merge; not code blockers)

- **N1 — evicted-but-active tenant series retention.** Cleanup now binds only to `TenantDeleting/Deleted`. A healthy tenant capacity-evicted from the resident pool (cap raised to 20,480 by #801) keeps its series resident forever (eviction no longer prunes; only deletion does). Pod-lifetime cardinality is bounded by *distinct tenants ever seen*, not by 20,480. Deliberate tradeoff behind "preserve live tenant counters" — confirm against pod tenant-churn, not just concurrent residency.
- **N2 — per-tenant success-rate observability dropped.** `serviceOperationsTotal` lost `tenant_id`/`org_id` (now aggregate); only failures are tenant-attributed. Per-tenant service-operation success ratio no longer computable (denominator gone). `tenantRequestsTotal` keeps per-tenant `status_class`, so HTTP-surface success rate survives. Confirm no dashboard/alert divides failures by the removed per-tenant total.
- **N3 — rollout-window label skew (deploy side).** Multiple metrics changed label shape (`tenant_requests_total` drops `result`, `tidbcloud_openapi` `path`→`api`, removed `event_bus_poll_failures_total` etc.). Mixed old/new binaries during rollout make fleet-sum recording rules double-count/mismatch. PR body cites runbooks#2688 but the rollout window isn't flagged in-code.
- **N4 — no explicit delete-then-reactivate race test.** UUIDs aren't reused so low severity; the named race isn't directly asserted.

**Verdict: GREEN.** B1 genuinely resolved. #801 + #802 must land in the same merge result (plus #801's B2 quota-cache panic-path fix).

qiffang (changes requested):
Requesting changes on exact head `e13d5e985c292d912fde8c26801fd355c0848090`.

This PR does fix the #801 request-path contention class: `Pool.closeEntry` no longer calls `metrics.DeleteTenantCounters`, and the lifecycle/outbox cleanup path is transactionally tied to deleting/deleted transitions. I am not blocking on the general N1/N2 tradeoffs about live-tenant series retention or per-tenant service-operation denominators. The blockers below are concrete metric correctness failures still present on this head.

1. **Deleted/revoked tenant traffic can recreate the exact tenant series cleanup just removed.** `tenantOutboxPoller.dispatch` deletes tenant counters once for `WorkMetricsCleanup`, but auth sets tenant metric scope at `pkg/server/auth.go:258` before `APIKeyActive` and before the tenant-status switch, and `/v1/status` does the same at `pkg/server/server.go:2066`. `resolveByAPIKeyHashDB` returns rows by `jwt_hash` without filtering key or tenant status, and `recordTenantHTTPRequest` records tenant-scoped counters/bytes for every final response once scope is present. Result: after cleanup runs, a stale SDK/FUSE heartbeat or any client retrying a revoked/deleted key can recreate `drive9_tenant_requests_total` and bytes indefinitely, with no later sweep now that pool eviction cleanup was removed. Existing tests prove deletion removes current series; they do not prove cleanup remains sticky against rejected post-delete traffic. Required before merge: fail-closed behavior for inactive/deleting/deleted auth attempts, or an explicit recently-cleaned suppression/re-delete mechanism, plus a regression that cleanup followed by revoked/deleted-key traffic does not reintroduce tenant-labeled series.

2. **The reduced DB pool metric can go false-green when an unscoped user pool is open.** Fork/delete flows still open tenant stores through `datastore.Open` (`pkg/server/fork.go:647`, `pkg/server/fork.go:1080`), which registers an unscoped `role="user"` pool via `pkg/datastore/store.go:160-180`. In `pkg/metrics/db.go:361-438`, tenant-scoped pools are accumulated into `tenantRoleTotals`, but the synthetic role aggregate is skipped whenever an unscoped same-role key exists. That means `drive9_db_pool_connections{role="user",state="max_open"}` can report only the transient unscoped fork/delete pool and exclude all tenant-scoped pool capacity, exactly while this PR removes per-tenant `max_open` series. The current mixed-pool test only asserts one role-only line exists; it does not assert the value includes tenant-scoped totals. Required before merge: either scope those fork/delete stores or make the role aggregate include both scoped and unscoped capacity, with a value-based regression.

3. **TiDB Cloud delete failures are counted as `ok` for unexpected 2xx responses.** `DeleteCluster` and `DeleteBranch` return success only for 200/204/404, but for any other status they call `recordTiDBCloudHTTPResponse(..., responseValid=true)` before returning an error. `tiDBCloudHTTPResult` maps every 2xx with `responseValid=true` to `ok`, so a 202 Accepted or other unexpected 2xx is both a caller-visible error and an `ok` OpenAPI metric. That violates the new `api/operation/result` contract and can hide upstream protocol drift. Required before merge: classify unexpected success-class responses that still return errors as protocol/error results, with delete-cluster/delete-branch regressions.

4. **`shared_db_pool_status_age_seconds` is not actually age in status.** `tenant_metrics.go:109-114` derives non-active status age from `db_pool.updated_at`, but the schema defines `updated_at ... ON UPDATE CURRENT_TIMESTAMP(3)` and unrelated writes such as tenant-count/soft-cap/spending-limit repair also update the row. A stuck provisioning/deleting pool can therefore reset to age 0 without a status transition, making any age alert false-green. Required before merge if this metric is part of the runbook contract: store/use status-transition time or rename/redefine the metric and alerts so it is not presented as status age.

Validation I ran locally: `git diff --check` clean; changed Go files `gofmt -l` clean; `go vet ./pkg/metrics ./pkg/meta ./pkg/server ./pkg/backend ./pkg/tenant ./pkg/tenant/tidbcloudnative` clean; `go build ./...` clean; `go test ./pkg/metrics ./pkg/tenant/tidbcloudnative -count=1` passed; test binaries compile for `pkg/meta`, `pkg/server`, `pkg/backend`, and `pkg/tenant`. Full local tests for the Docker-backed packages are blocked on this machine by `panic: rootless Docker not found`. Hosted `ci` and `drive9-server-local e2e smoke` are green.

qiffang:
## adversary-2 — retracting GREEN, REVISE (endorsing @adversary-1's blockers)

I verified the two most impactful of @adversary-1's four findings; both are real metric-correctness defects, harder than my earlier N1/N2 tradeoffs. My gap: I focused on the B1 request-path coupling (the headline) and cardinality bounds, not the per-metric result classification and re-introduction paths.

**B3 confirmed** — `http_clusters_api.go:213/217`: `DeleteCluster` treats only 200/204/404 as success; any other status (incl. a 202 or other 2xx) calls `recordTiDBCloudHTTPResponse(..., responseValid=true)` then returns an error. `openapi_metrics.go:53-58` maps `2xx + responseValid=true → "ok"`. So an unexpected 2xx delete is simultaneously a caller-visible error AND an `ok` OpenAPI metric — violates the new `api/operation/result` contract and hides upstream protocol drift. `DeleteBranch:299/303` identical. Fix: pass `responseValid=false` for unexpected 2xx in delete paths (→ `protocol-error`) + delete regressions.

**B1 confirmed** — auth sets tenant metric scope at `auth.go:258` BEFORE `APIKeyActive` and before the tenant-status switch; `/v1/status` (server.go:2066) same; `resolveByAPIKeyHashDB` returns by `jwt_hash` without filtering key/tenant status; `recordTenantHTTPRequest` records tenant-scoped counters for every final response. After cleanup runs, a stale SDK/FUSE heartbeat or any retry on a revoked/deleted key recreates `drive9_tenant_requests_total`/bytes indefinitely — and pool-eviction cleanup was removed by this PR, so no later sweep. Existing tests prove deletion removes current series, not that cleanup stays sticky against post-delete rejected traffic. Fix: fail-closed for inactive/deleting/deleted auth attempts, or a recently-cleaned suppression/re-delete mechanism + regression.

**B2 / B4** — analysis paths credible (fork/delete opens an unscoped `role="user"` pool → role aggregate skips tenant-scoped totals → false-green DB-pool `max_open`; `shared_db_pool_status_age` derived from `updated_at ON UPDATE`, reset by unrelated tenant-count/soft-cap/spending writes → age alert false-green). I did not line-trace these two but concur with the direction.

**Verdict: REVISE.** B1 coupling is genuinely resolved (my prior finding stands), but these four metric-correctness defects block merge. #801 + #802 still need the same merge result. Aligned with @adversary-1.

pkg/server/tenant_metrics.go
srstack: Fixed in 09debf24. The metric now uses db_pool.status_updated_at rather than updated_at. Existing rows are backfilled from updated_at during migration, and status-transition paths update status_updated_at only when the status changes. TestManagedSharedDBUpdatedAtTracksProgressNotNoOpPolls covers no-op writes versus real transitions.

pkg/tenant/tidbcloudnative/http_clusters_api.go
srstack: Fixed in 09debf24. DeleteCluster and DeleteBranch pass responseValid=false for unexpected statuses, so an unexpected 2xx still returns the caller error and is classified as protocol_error. The regression covers both delete paths.

pkg/server/tenant_outbox_poller.go
srstack: Fixed in 09debf24. RecordTenantOutboxPoll returns before observing batch size for non-ok results, so an error poll with size 0 cannot add a histogram sample. The regression covers success count plus error count unchanged.

pkg/tenant/tidbcloudnative/openapi_metrics.go
srstack: Fixed in 09debf24. The unused tidbCloudAPIBilling and tidbCloudOperationGetOrganizationPlan constants were deleted. The deterministic credential/role validation and legacy starter paths remain intentionally classified/outside this HTTP producer contract.

pkg/server/fork.go
srstack: Fixed in 09debf24. Both fork/delete paths now use datastore.OpenForTenant with the tenant ID, so they no longer register an unscoped user pool. The mixed scoped/unscoped regression also asserts the role aggregate max_open value includes both pools.

qiffang:
## adversary-2 R2 at `5e9af58e` — GREEN (all four blockers independently verified closed)

**B1' RESOLVED** — dual fix: (a) fail-closed — `auth.go:258` no longer sets metric scope; the main middleware sets it only at `:418` (after `APIKeyActive` + the tenant-status switch); the delete-no-acquire path uses `setRequestMetricTenantForAuthStatus` (instrumentation.go:230) which no-ops unless `status==TenantActive`. (b) cross-pod — new `WorkAPIKeyCacheCleanup` bit; all three revoke methods wrap UPDATE + `insertTenantNotifyTx(WorkAPIKeyCacheCleanup)` in one `InTx` (atomic, outbox only on n>0); poller dispatches `InvalidateAPIKeyResolveCacheForTenant` per pod. Regressions: `TestSetRequestMetricTenantForAuthStatusOnlyScopesActiveTenants`, `TestAPIKeyRevocationsWriteCacheCleanupOutbox`, `TestTenantOutboxPollerAPIKeyCacheCleanupIsBroadcastWithoutDeletingMetrics`.

**B3' RESOLVED** — `DeleteCluster`/`DeleteBranch` (http_clusters_api.go:217/303) pass `responseValid=false` → `tiDBCloudHTTPResult(2xx,false)` = `protocol_error`. Regression `TestTiDBCloudOpenAPIDeleteUnexpectedSuccessIsProtocolError` (202 → ok delta 0, protocol_error +1). Non-delete verbs keep `true` (those 2xx are genuinely valid).

**B2' RESOLVED** — root cause: `fork.go:647/1080` now `datastore.OpenForTenant` (tenant-scoped). Backstop: `db.go:371` folds unscoped saturation into the same-role aggregate; `:428` suppresses the standalone unscoped line when a scoped pool exists. Value-based regression `TestDBPoolConnectionsDoesNotDuplicateRoleAggregateForMixedPools` (unscoped 2 + scoped 3 → aggregate max_open=5, once).

**B4' RESOLVED** — new `db_pool.status_updated_at` column (migration backfills from `updated_at`), set by all status-changing UPDATEs (unconditional on real transitions 505/534/622; `CASE WHEN status <> ?` on idempotent upserts 372/914/931); `tenant_metrics.go:109` uses `snapshot.StatusUpdatedAt`, so unrelated `ON UPDATE` writes no longer reset age. Regression `TestManagedSharedDBUpdatedAtTracksProgressNotNoOpPolls`.

No new blockers. N1/N2/N3 tradeoff notes unchanged (evicted-live-tenant series retention, per-tenant success denominator, rollout label skew) — for qiffang's awareness, not blockers.

**Verdict: GREEN.** Coupling unchanged: #801 (`60dfea19`) + #802 (`5e9af58e`) must land in the same merge result.

srstack:
Current-head review disposition for the auth/revocation finding:

- `09debf24` moved tenant metric attribution after API-key-active and tenant-active checks, so rejected revoked/deleting/deleted traffic no longer creates tenant-labeled request/bytes/business-event series. The regression `TestSetRequestMetricTenantForAuthStatusOnlyScopesActiveTenants` covers cleanup followed by rejected traffic.
- To close the cross-pod stale-cache window for live-tenant key revocation, `5e9af58e` adds a separate durable `TenantNotifyWorkAPIKeyCacheCleanup` bit. `RevokeAPIKey`, `RevokeTenantAPIKeys`, and `RevokeAPIKeysByIssuer` write that bit in the same transaction as the revoke; every poller evicts only its local API-key resolve cache. It deliberately does not reuse `WorkMetricsCleanup`, so revoking one key cannot delete metrics for a live tenant.
- `TestTenantOutboxPollerAPIKeyCacheCleanupIsBroadcastWithoutDeletingMetrics` verifies stale peer-cache eviction and live-tenant metric preservation.

The four concrete blockers in the earlier formal review were fixed in `09debf24`; the current head also closes ordinary revoke propagation. The only remaining CI issue on the previous head was a global-histogram test assuming a zero baseline; `5e9af58e` makes that assertion baseline-relative. Local `pkg/meta`, `pkg/metrics`, `pkg/server`, `pkg/backend`, and `pkg/tenant/tidbcloudnative` tests pass.

qiffang (approved):
## adversary-1 R2 at `5e9af58e42779a11e9213dd1d4f0eb607723637b`

Approved. I re-anchored after qiffang's latest request; there is no newer head after `5e9af58e`, and both hosted `code-ci` and `drive9-server-local e2e smoke` are green on this exact commit.

The four prior metric-correctness blockers are closed:

1. Cleanup-then-rejected traffic no longer recreates deleted tenant series. The auth paths now set tenant request metric scope only after active-tenant authorization (`setRequestMetricTenantForAuthStatus`), and revoked-key cleanup is broadcast through durable `WorkAPIKeyCacheCleanup` outbox rows so peer pods invalidate cached API-key resolutions without deleting live tenant metrics.

2. DB role aggregate no longer false-greens when unscoped fork/delete pools coexist with tenant-scoped pools. Fork/delete now use `datastore.OpenForTenant`, and `pkg/metrics/db.go` folds unscoped saturation into the role aggregate instead of letting the unscoped row hide tenant-scoped capacity. The regression checks value, not just line count.

3. TiDB Cloud delete unexpected 2xx responses no longer record `ok`. `DeleteCluster` and `DeleteBranch` now pass `responseValid=false` on fall-through, so a 202-style unexpected delete success records `protocol_error`; the regression covers both operations.

4. `shared_db_pool_status_age_seconds` now uses true status-transition age. `db_pool.status_updated_at` is backfilled and updated only on status transitions, and metrics use `StatusUpdatedAt` instead of `updated_at`, so non-status writes cannot reset the age.

Notes: The remaining #801/#802 coupling is a merge-order/composition gate, not a blocker inside #802 itself. #801 still must not land its 20x cap unless this PR or equivalent metrics cleanup is already in the tested merge result; the current exact #801 + #802 synthetic merge still conflicts in `pkg/tenant/pool_test.go`, so whoever lands both must resolve and retest that combined state.

Validation performed:
- Passed locally: `git diff --check origin/main...HEAD`, changed-file `gofmt -l`, `go test ./pkg/metrics ./pkg/tenant/tidbcloudnative -count=1`, test binary compilation for `pkg/meta`, `pkg/server`, `pkg/backend`, `pkg/tenant`, `go vet ./pkg/metrics ./pkg/meta ./pkg/server ./pkg/backend ./pkg/tenant ./pkg/tenant/tidbcloudnative`, and `go build ./...`.
- Hosted `code-ci`, `drive9-server-local e2e smoke`, and CodeRabbit are green on `5e9af58e`.

PR #805, opened 2026-07-27
fix: project fs_id in shared SKIP LOCKED claim queries (TiDB planner workaround)

mornyx:
## Why

Since the shared-table rollout, `tenant_worker` claim loops against shared task tables fail every claim with:

```
Error 1105 (HY000): Can't find column tidbcloud_fs.semantic_tasks.fs_id in schema Column: [tidbcloud_fs.semantic_tasks.task_id]
```

Verified against a live shared DB (`tidbcloud_fs`, TiDB `v8.5.3-serverless`, 2026-06-25 build): the table **has** `fs_id` (`PRIMARY KEY (fs_id, task_id) CLUSTERED`), and the failure is specific to `FOR UPDATE SKIP LOCKED` — the TiDB planner builds the table schema from the SELECT projection only, so a `WHERE` predicate over an unprojected column is rejected. Plain `SELECT` and plain `FOR UPDATE` are unaffected, and MySQL parses all shapes. Reproduced on an empty table (plan-time error):

| shape | result |
|---|---|
| `SELECT task_id ... FOR UPDATE` | ok |
| `SELECT task_id ... FOR UPDATE SKIP LOCKED` | Can't find column |
| `SELECT fs_id, task_id ... FOR UPDATE SKIP LOCKED` | **ok** |
| `SELECT * ... FOR UPDATE SKIP LOCKED` | ok |

## Change

Project `fs_id` in the four `FOR UPDATE SKIP LOCKED` claim queries, **shared shape only** — standalone tenant DBs have no `fs_id` column, so the projection must not be added there:

- New `Scope.SelCols` (mirrors `InsCols`): prefixes the SELECT column list with `fs_id, ` only when the store addresses a shared-schema DB, next to a comment documenting the TiDB planner behavior it works around.
- Applied to `ClaimSemanticTask` / `RecoverExpiredSemanticTasks` / `ClaimFileGCTask` / `RecoverExpiredFileGCTasks`; the leading column is scanned and discarded (`withFsID` on the two scan helpers + an explicit leading scan in the two recover loops).
- Plain `FOR UPDATE` sites (task dedup, lease checks) are untouched — they are not affected by the bug.

Claim semantics (WHERE / ORDER / LIMIT / locking) are unchanged; the only difference on the wire is one extra returned column.

## Test plan

- `go build ./...`, `make lint` → 0 issues.
- `make test TEST_PKGS='./pkg/datastore/...'` → all green, including the shared-schema suites exercising claim + recover against real MySQL.
- New `TestScopeSelCols` pins shared vs standalone projection.
- Verified end-to-end on a live TiDB `v8.5.3-serverless` shared DB that the new shapes parse under `FOR UPDATE SKIP LOCKED`.

Follow-up (outside this PR): report the planner bug upstream with the minimal repro (`SELECT task_id FROM semantic_tasks WHERE fs_id = ? FOR UPDATE SKIP LOCKED`); this projection change can be revisited once a fixed TiDB ships.

### DIFF
--- a/pkg/datastore/file_gc_tasks.go
+++ b/pkg/datastore/file_gc_tasks.go
@@ -144,15 +144,20 @@ func (s *Store) ClaimFileGCTask(ctx context.Context, now time.Time, leaseDuratio
 	}
 	defer func() { _ = tx.Rollback() }()
 
-	row := tx.QueryRowContext(ctx, `SELECT task_id, file_id, storage_type, storage_ref,
+	// SelCols adds the fs_id projection in shared shape to work around the
+	// TiDB FOR UPDATE SKIP LOCKED planner bug that rejects WHERE predicates
+	// over unprojected columns (see Scope.SelCols).
+	row := tx.QueryRowContext(ctx, `SELECT `+s.scope.SelCols(`task_id, file_id, storage_type, storage_ref,
 		size_bytes, content_type, status, attempt_count, max_attempts, receipt,
-		leased_at, lease_until, available_at, last_error, created_at, updated_at, completed_at
+		leased_at, lease_until, available_at, last_error, created_at, updated_at, completed_at`)+`
 		FROM file_gc_tasks
 		WHERE `+s.scope.And(`status = ? AND available_at <= ?`)+`
 		ORDER BY available_at, created_at, task_id
 		LIMIT 1
 		FOR UPDATE SKIP LOCKED`, s.scope.Args(FileGCTaskQueued, now)...)
-	task, err := scanFileGCTask(row)
+	// The claim query projects fs_id first in shared shape (workaround for
+	// the TiDB SKIP LOCKED planner bug; see Scope.SelCols); withFsID skips it.
+	task, err := scanFileGCTask(row, s.scope.Shared())
 	if err != nil {
 		if errors.Is(err, ErrNotFound) {
 			return nil, false, nil
@@ -282,7 +287,7 @@ func (s *Store) RecoverExpiredFileGCTasks(ctx context.Context, now time.Time, li
 	}
 	defer func() { _ = tx.Rollback() }()
 
-	query := `SELECT task_id FROM file_gc_tasks
+	query := `SELECT ` + s.scope.SelCols(`task_id`) + ` FROM file_gc_tasks
 		WHERE ` + s.scope.And(`status = ? AND lease_until IS NOT NULL AND lease_until < ?`) + `
 		ORDER BY lease_until, created_at, task_id`
 	args := s.scope.Args(FileGCTaskProcessing, now)
@@ -299,8 +304,15 @@ func (s *Store) RecoverExpiredFileGCTasks(ctx context.Context, now time.Time, li
 
 	var taskIDs []string
 	for rows.Next() {
+		// Shared shape projects fs_id first (workaround for the TiDB SKIP
+		// LOCKED planner bug; see Scope.SelCols); scan and discard it.
+		var fsID int64
 		var taskID string
-		if err := rows.Scan(&taskID); err != nil {
+		if s.scope.Shared() {
+			if err := rows.Scan(&fsID, &taskID); err != nil {
+				return 0, err
+			}
+		} else if err := rows.Scan(&taskID); err != nil {
 			return 0, err
 		}
 		taskIDs = append(taskIDs, taskID)
@@ -421,14 +433,23 @@ func (s *Store) fileGCTaskLeaseError(ctx context.Context, taskID string) error {
 	return ErrFileGCTaskLeaseMismatch
 }
 
-func scanFileGCTask(s scanner) (*FileGCTask, error) {
+func scanFileGCTask(s scanner, withFsID ...bool) (*FileGCTask, error) {
 	var task FileGCTask
 	var contentType, receipt, lastError sql.NullString
 	var leasedAt, leaseUntil, completedAt sql.NullTime
-	err := s.Scan(&task.TaskID, &task.FileID, &task.StorageType, &task.StorageRef,
+	// withFsID[0] is set by the SKIP LOCKED claim paths: in shared shape those
+	// queries project fs_id first (workaround for the TiDB SKIP LOCKED
+	// planner bug; see Scope.SelCols), so scan and discard the leading column.
+	dest := make([]any, 0, 18)
+	if len(withFsID) > 0 && withFsID[0] {
+		var fsID int64
+		dest = append(dest, &fsID)
+	}
+	dest = append(dest, &task.TaskID, &task.FileID, &task.StorageType, &task.StorageRef,
 		&task.SizeBytes, &contentType, &task.Status, &task.AttemptCount,
 		&task.MaxAttempts, &receipt, &leasedAt, &leaseUntil, &task.AvailableAt,
 		&lastError, &task.CreatedAt, &task.UpdatedAt, &completedAt)
+	err := s.Scan(dest...)
 	if err != nil {
 		if errors.Is(err, sql.ErrNoRows) {
 			return nil, ErrNotFound
--- a/pkg/datastore/scope.go
+++ b/pkg/datastore/scope.go
@@ -122,6 +122,25 @@ func (s Scope) InsVals(vals string) string {
 	return vals
 }
 
+// SelCols prefixes a SELECT column list with "fs_id, " in shared shape and
+// returns it unchanged in standalone shape. It exists ONLY to work around a
+// TiDB planner bug: with FOR UPDATE SKIP LOCKED, TiDB builds the table
+// schema from the projection only, so a WHERE predicate over an unprojected
+// column is rejected with planner error "Can't find column" — even though
+// the SQL is perfectly valid (it works on MySQL, PostgreSQL, and on TiDB
+// itself with plain FOR UPDATE). Observed on v8.5.3-serverless
+// (2026-06-25 build). Projecting fs_id avoids the bug; callers must scan
+// the extra leading column and may discard it. This can be reverted once
+// upstream fixes the planner. Use it only where the fs_id predicate is
+// present (i.e. queries built with And/Args), otherwise the extra
+// projection is wasted bytes on the wire.
+func (s Scope) SelCols(cols string) string {
+	if s.shared {
+		return "fs_id, " + cols
+	}
+	return cols
+}
+
 // tenantCol returns the tenant-discriminator column name for journal/vault
 // tables under this Store's schema shape.
 func (s *Store) tenantCol() string { return s.scope.TenantCol() }
--- a/pkg/datastore/scope_test.go
+++ b/pkg/datastore/scope_test.go
@@ -1,0 +1,13 @@
+package datastore
+
+import "testing"
+
+func TestScopeSelCols(t *testing.T) {
+	t.Helper()
+	if got := SharedScope(42).SelCols("task_id"); got != "fs_id, task_id" {
+		t.Fatalf("shared SelCols = %q, want %q", got, "fs_id, task_id")
+	}
+	if got := StandaloneScope(42).SelCols("task_id"); got != "task_id" {
+		t.Fatalf("standalone SelCols = %q, want %q", got, "task_id")
+	}
+}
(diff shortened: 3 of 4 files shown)


srstack:
## Review — `fix: project fs_id in shared SKIP LOCKED claim queries`

Reviewed at `4492ac7f`. Verified locally in a worktree: `go test ./pkg/datastore/... -count=1` passes (67s, includes the shared-schema claim/recover suites against real MySQL).

### Overview

Works around a TiDB planner bug where `FOR UPDATE SKIP LOCKED` builds the table schema from the SELECT projection only, so a `WHERE fs_id = ?` predicate over an unprojected `fs_id` is rejected with `Can't find column`. The fix adds `Scope.SelCols`, which prefixes `fs_id, ` onto the projection in shared shape only, applied to the four `SKIP LOCKED` claim/recover queries, with the extra leading column scanned and discarded.

The investigation behind this is excellent — the shape-by-shape table in the description (plain `FOR UPDATE` ok, `SELECT *` ok, projecting `fs_id` ok) isolates the bug precisely, and pinning it to `v8.5.3-serverless` with a minimal repro plus a plan to report upstream is exactly right. The workaround comment on `SelCols` explains *why* it exists and when it can be reverted, which is what makes a workaround like this maintainable.

### Correctness

I checked the two things that could go wrong and both are right:

- **Projection/scan pairing.** All three `SelCols` call sites pass `s.scope.Shared()` to the matching scan helper, and the three scan call sites that *don't* project `fs_id` (`file_gc_tasks.go:69`, `:247`, `semantic_tasks.go:199`, `:471`) correctly omit the flag. `dest` slice capacities (18 and 17) match the appended counts exactly.
- **Completeness.** Grepped every `SKIP LOCKED` site in the tree — there are exactly four non-test occurrences, and all four are covered. No missed site.

Claim semantics are genuinely unchanged: `WHERE`/`ORDER BY`/`LIMIT`/locking are untouched, and the shared-schema tests exercise claim, lease expiry, recover, and re-claim, so the discard path is covered against a real server rather than only unit-asserted.

Restricting the change to the shared shape is the correct call — standalone tenant DBs have no `fs_id` column, so an unconditional projection would break them outright.

### Suggestions

**1. Prefer an explicit `bool` over `withFsID ...bool`** (`semantic_tasks.go:611`, `file_gc_tasks.go:436`)

The variadic makes the flag optional at the call site, so a future `SKIP LOCKED` claim query that forgets to pass `s.scope.Shared()` compiles fine and fails at runtime with a column-count scan error. A plain `withFsID bool` parameter would force every call site to state its intent and make the mistake impossible:

```go
func scanSemanticTask(s scanner, withFsID bool) (*semantic.Task, error)
```

That's six call sites to touch, all mechanical. Given this is a workaround that will be reverted later, compiler-enforced pairing seems worth the churn — the revert then becomes "delete the parameter and let the compiler find every site."

**2. `t.Helper()` in a test function is a no-op** (`scope_test.go:6`)

`t.Helper()` marks a *helper* so failures report the caller's line; in a `Test*` function itself it does nothing. Harmless, but drop it.

**3. Duplicated discard logic in the two recover loops**

`RecoverExpiredSemanticTasks` and `RecoverExpiredFileGCTasks` both inline the same `if s.scope.Shared() { Scan(&fsID, &taskID) } else { Scan(&taskID) }` branch. A tiny helper would keep the revert to one place:

```go
func (s *Store) scanScopedTaskID(rows *sql.Rows) (string, error)
```

Minor, and arguably the explicit form is clearer at two sites. Take it or leave it.

**4. Consider asserting the workaround itself, not just the string**

`TestScopeSelCols` pins the projection string, which is good. What it can't catch is regression of the underlying reason — if someone later removes `SelCols` from one query, the unit test still passes and only the shared-schema integration test fails (and only against a TiDB that has the bug; against MySQL it would pass, since MySQL parses all shapes fine). Worth a comment on the shared claim tests noting they're the real guard here, so a future reader doesn't assume `TestScopeSelCols` covers it.

### Risks

- **Wire cost:** one extra `bigint` per claimed row. `LIMIT 1` on the claim queries, and recover is bounded by `limit`. Negligible.
- **Revert coupling:** four query sites plus two scan helpers plus two recover loops must be reverted together. The consistent `see Scope.SelCols` comment on every site makes them greppable, which is the right mitigation.
- **`SelCols` misuse:** the doc comment correctly warns to use it only where the `fs_id` predicate is present. Nothing currently violates that.

No security or performance concerns.

### Verdict

LGTM with the nits. Well-scoped fix for a genuinely surprising upstream bug, correctly limited to the affected shape, and the diagnosis is documented well enough that whoever reverts this in six months will understand what it was for. Only suggestion I'd actually push for is #1 — the variadic is the one thing here that could silently rot.

qiffang:
## adversary-2 review at `4492ac7f` — GREEN

Clean, precise production bug fix. Verified point by point:

**Root cause confirmed:** TiDB `FOR UPDATE SKIP LOCKED` planner builds the table schema from the SELECT projection only, so a WHERE predicate over an unprojected column (`fs_id`, part of the shared PK, injected by `s.scope.And`) fails with 1105. Plain `FOR UPDATE`, MySQL, and PostgreSQL are all fine — TiDB SKIP LOCKED specific. Fix: `Scope.SelCols(cols)` prepends `fs_id, ` in shared shape so the planner sees the column.

**Completeness (I specifically checked for missed sites):** the datastore package has `SKIP LOCKED` queries ONLY in semantic_tasks.go + file_gc_tasks.go — 4 total (claim + recover in each), all patched. No other file has SKIP LOCKED. The plain `FOR UPDATE` queries at semantic_tasks:198/470 and file_gc_tasks:246 are correctly left unchanged (the bug doesn't affect plain FOR UPDATE). No site missed.

**Correctness:**
- Standalone unaffected: `SelCols` returns cols unchanged when `!shared`; scan uses the non-fsID path.
- Scan ↔ projection stay in sync: both key on the same `s.shared` flag, so shared projects `fs_id` first and scan discards a leading `fsID` — no misalignment possible.
- Symmetric across both tables (semantic + file_gc, claim + recover).
- `SelCols` docstring correctly restricts use to fs_id-predicate queries (And/Args) — all 4 sites use `s.scope.And`.

**Notes (non-blocking):**
- Minor style inconsistency: single-row scan uses a `withFsID ...bool` variadic; multi-row recover uses an explicit `s.scope.Shared()` branch. Both correct.
- **Evidence gap:** `scope_test.go` only tests the `SelCols` string concatenation, not an end-to-end claim. For a planner-workaround fix, a shared-shape integration test (or reproduced live evidence) proving the claim now SUCCEEDS on a shared/TiDB build closes the loop. If existing shared execsql tests already cover the claim path, please point to them.

**Verdict: GREEN** — fix is correct, complete, no missed SKIP LOCKED sites, standalone not regressed. Recommend adding end-to-end shared-claim evidence before merge.

qiffang (approved):
GREEN for exact head `4492ac7fed8b1a533a5c1e993fda35e366aafb88`.

No blocking findings.

What I checked:
- Completeness: `rg "SKIP LOCKED" pkg/datastore` shows exactly four production sites, and this PR updates all four: semantic claim/recover and file-GC claim/recover.
- Shared-only projection: each changed query uses `scope.SelCols(...)`, so shared shape projects leading `fs_id` while standalone shape still emits the old column list and does not reference a nonexistent `fs_id` column.
- Scan arity/call-site pairing: the two claim paths pass `s.scope.Shared()` into the matching scan helper; the two recover loops explicitly scan/discard the leading `fs_id` only in shared shape. Untouched plain `FOR UPDATE` scan sites still use the original projection and default scanner path.
- Semantics: `WHERE`, `ORDER BY`, `LIMIT`, update predicates, receipt/lease mutation, and retry/recover behavior are unchanged; only one extra shared-shape projected column is added to satisfy the TiDB `FOR UPDATE SKIP LOCKED` planner behavior.
- Regression coverage: existing shared-schema parity and cross-tenant tests exercise claim/recover for both semantic and file-GC tasks; new `TestScopeSelCols` pins shared vs standalone projection.

Local validation:
- `gofmt -l` on changed Go files passed.
- `go test -c ./pkg/datastore` passed.
- `go vet ./pkg/datastore` passed.
- `go build ./...` passed.

Runtime datastore tests could not execute locally here because `pkg/datastore` `TestMain` starts testcontainers before test selection and this machine fails with `panic: rootless Docker not found`. Hosted exact-head `ci` and `drive9-server-local e2e smoke` are green, and those are the runtime evidence I rely on for the DB-backed paths.

Non-blocking hardening: I agree with srstack that `scanSemanticTask(..., withFsID ...bool)` / `scanFileGCTask(..., withFsID ...bool)` would be safer as explicit `withFsID bool` parameters so future projection/scan mismatches are compiler-visible. That is not a merge blocker for this exact patch because every current call site is paired correctly.

qiffang:
## @dat9-dev2 independent exact-head review — GREEN

Reviewed exact head `4492ac7fed8b1a533a5c1e993fda35e366aafb88` against base `ee2e6e2f49c6e1d02e30552c49207a9eec057b41`. I found **no blocking correctness, isolation, concurrency, or failure-path defect**.

### What I verified

- **Complete affected surface:** `rg 'SKIP LOCKED' pkg/datastore` has exactly four production queries. This patch covers semantic claim/recover and file-GC claim/recover; the unaffected plain `FOR UPDATE` paths remain byte-for-byte on their old projections.
- **Projection/scan contract:** semantic claim returns 17 shared columns / 16 standalone columns; file-GC claim returns 18 / 17. Both claim call sites key projection and scan from the same immutable `Scope`. Both recover loops scan `(fs_id, task_id)` only in shared shape. The untouched scanner call sites still read the old projections.
- **Tenant isolation and transactional semantics:** every changed query still uses `Scope.And` + `Scope.Args`; `WHERE`, ordering, limit, lock mode, update predicates, receipt generation, lease mutation, zero-row handling, rollback, and commit boundaries are unchanged. Projecting and discarding `fs_id` cannot widen the tenant scope.
- **Standalone compatibility:** `SelCols` returns its input unchanged when `shared=false`, so standalone databases never reference the absent `fs_id` column.

### Fresh runtime evidence

I started the official TiDB `v8.5.3` darwin-arm64 binary locally with UniStore, installed the repository's exact `CoreFSTiDBSharedSchemaStatements()` (including `(fs_id, task_id) CLUSTERED`), and ran a temporary same-package regression that drives the full shared semantic-task and file-GC scenarios: enqueue, claim, ack, retry, lease expiry, recover, re-claim, dead-letter/list. **PASS** for both task families.

Evidence boundary: the public `v8.5.3` UniStore build did **not** reproduce the old-projection planner failure in an empty-table SQL matrix; old and fixed claim/recover shapes both parsed. Therefore this independently proves the patch works on exact clustered shared DDL and TiDB 8.5.3, but the original failure remains specific to the stated `v8.5.3-serverless` 2026-06-25 build/config and is supported by the PR author's live-DB evidence, not by the repo's MySQL CI or my public-binary reproduction.

Fresh local gates:

- exact clustered-shared TiDB claim/recover lifecycle probe: PASS
- existing shared parity + cross-tenant claim/recover tests against TiDB DSN: PASS
- `git diff --check origin/main...HEAD`: PASS
- changed-file `gofmt -l`: PASS
- `go test -c ./pkg/datastore`: PASS
- `go test -race -c ./pkg/datastore`: PASS
- `go vet ./pkg/datastore`: PASS
- `go build ./...`: PASS

Hosted exact-head `ci` and `drive9-server-local e2e smoke` are GREEN.

### Non-blocking hardening

1. Prefer `withFsID bool` over `withFsID ...bool` in both scanners. The variadic preserves old call sites but makes a future projection/scan mismatch compile successfully and fail only at runtime; an explicit bool makes every projection intent compiler-visible.
2. Once the upstream TiDB report exists, link it from `Scope.SelCols`. Until then, consider wording the internal planner mechanism as an empirical serverless-build observation rather than a universal v8.5.3 fact, because the public v8.5.3 binary did not reproduce it. TiDB's lock plan explicitly retains table-handle columns during pruning, which is also consistent with a common-handle/projection interaction: <https://github.com/pingcap/tidb/blob/v8.5.3/pkg/planner/core/operator/logicalop/logical_lock.go#L53-L80>.
3. `t.Helper()` inside `TestScopeSelCols` itself is a no-op; remove as cleanup.

**Verdict: GREEN for exact head `4492ac7f`.** The workaround is narrow, all current projection/scan pairs are correct, both schema shapes are preserved, and the four affected paths are complete.

PR #807, opened 2026-07-27
fix(client-js): route small in-memory writes inline regardless of omitted size

qiffang:
## Problem

A user reported `writeStreamWithSummary` uploaded a **7KB file on a TiDB tenant to S3** instead of inlining it into TiDB (inline threshold default 50KB, so 7KB should inline).

## Root cause (client-side)

For an in-memory `Uint8Array`, the inline-vs-multipart decision used the caller-supplied `size` argument:

```ts
if (size === 0 || size < threshold) { /* inline direct PUT */ }
```

JS callers routinely omit `size` (`client.writeStreamWithSummary(path, bytes)`). An `undefined` size makes both checks false, so a small buffer silently fell through to the multipart (`writeStreamV2` → S3) path — even though `data.length` (the real 7000) was already in hand. Diagnosis by @adversary-1.

## Fix

- **Uint8Array**: effective size = `data.length` (authoritative); an omitted/incorrect `size` can no longer misroute a small file to S3.
- **ReadableStream**: require finite, non-negative `size`, fail fast otherwise.
- `UploadSummary.total_bytes` now always the effective size, never `undefined`.

Type signature still declares `size` (backward compatible); runtime now defends against omission.

## Regressions (per review gate)

- small Uint8Array with size omitted → `direct_put`, `total_bytes===7000`, no `/v2/uploads/initiate`.
- ReadableStream with missing/`NaN`/negative size → clear throw.

Full `npm run test` (69 passed), `npm run build`, `npm run lint` (`tsc --noEmit`) green.

## Note

Client-side cause only. A correct-explicit-size case still hitting S3 would be server-side (deployment `DRIVE9_INLINE_THRESHOLD` below file size, or tenant `provider` not `tidb_*` → `SmallInDB()=false`) — separate.

## Summary by CodeRabbit

* **Bug Fixes**
  * Uploads from in-memory data now calculate file size accurately, even when the size is omitted or incorrect.
  * Small in-memory uploads are correctly routed through direct upload.
  * Stream uploads now provide a clear validation error when the size is missing, negative, or not finite.
  * Upload metrics now report accurate byte counts.
* **Tests**
  * Added coverage to prevent regressions in direct-upload routing, size validation, and reported byte counts.
* **Chores**
  * Bumped the package version to **0.1.4**.

### DIFF
--- a/clients/drive9-js/src/transfer.ts
+++ b/clients/drive9-js/src/transfer.ts
@@ -184,23 +184,35 @@ export async function writeStreamWithSummaryImpl(
   const started = new Date();
   const threshold = client["smallFileThreshold"];
   let data: Uint8Array;
+  let effectiveSize: number;
   if (stream instanceof Uint8Array) {
+    // The byte length is authoritative for an in-memory buffer, so ignore a
+    // missing/incorrect `size` argument: JS callers routinely omit it, and
+    // an `undefined` size otherwise fails every `size < threshold` check and
+    // silently routes a small file through multipart/S3 instead of inline.
     data = stream;
+    effectiveSize = data.length;
   } else {
+    if (typeof size !== "number" || !Number.isFinite(size) || size < 0) {
+      throw new Error(
+        `writeStreamWithSummary requires a finite non-negative size for a ReadableStream, got ${String(size)}`
+      );
+    }
     data = await streamToUint8Array(stream, size);
+    effectiveSize = data.length;
   }
-  if (size === 0 || size < threshold) {
+  if (effectiveSize === 0 || effectiveSize < threshold) {
     await client.write(path, data, opts);
-    return uploadSummary(path, size, "direct_put", started);
+    return uploadSummary(path, effectiveSize, "direct_put", started);
   }
   try {
     const plan = await writeStreamV2(client, path, data, opts);
-    return uploadSummary(path, size, "multipart_v2", started, plan.part_size, plan.total_parts, plan.total_parts);
+    return uploadSummary(path, effectiveSize, "multipart_v2", started, plan.part_size, plan.total_parts, plan.total_parts);
   } catch (err) {
     const msg = err instanceof Error ? err.message : String(err);
     if (msg.includes("v2 upload API not available")) {
       const plan = await writeStreamV1(client, path, data, opts);
-      return uploadSummary(path, size, "multipart_v1", started, plan.part_size, plan.parts.length, plan.parts.length);
+      return uploadSummary(path, effectiveSize, "multipart_v1", started, plan.part_size, plan.parts.length, plan.parts.length);
     } else {
       throw err;
     }
--- a/clients/drive9-js/tests/transfer.test.ts
+++ b/clients/drive9-js/tests/transfer.test.ts
@@ -46,6 +46,60 @@ describe("Transfer", () => {
     expect(initiateCalled).toBe(false);
   });
 
+  it("routes a small Uint8Array through direct PUT even when size is omitted", async () => {
+    // Regression: callers routinely omit the size argument for an in-memory
+    // buffer. An undefined size used to fail every `size < threshold` check
+    // and silently push a small file through multipart/S3 instead of inline.
+    let putCalled = false;
+    let initiateCalled = false;
+    server.use(
+      http.get("http://localhost:9009/v1/status", () => HttpResponse.json({ inline_threshold: 50000 })),
+      http.put("http://localhost:9009/v1/fs/no-size-small.bin", () => {
+        putCalled = true;
+        return HttpResponse.text("ok");
+      }),
+      http.post("http://localhost:9009/v2/uploads/initiate", () => {
+        initiateCalled = true;
+        return HttpResponse.text("unexpected multipart initiate", { status: 500 });
+      })
+    );
+    const client = new Client("http://localhost:9009", "test-key");
+    await client.warm();
+    const data = new TextEncoder().encode("x".repeat(7000));
+    // size argument intentionally omitted
+    const summary = await (client.writeStreamWithSummary as unknown as (
+      p: string,
+      s: Uint8Array
+    ) => Promise<{ mode: string; total_bytes: number }>)("/no-size-small.bin", data);
+    expect(summary.mode).toBe("direct_put");
+    expect(summary.total_bytes).toBe(7000);
+    expect(putCalled).toBe(true);
+    expect(initiateCalled).toBe(false);
+  });
+
+  it("throws for a ReadableStream with a missing or non-finite size", async () => {
+    const client = new Client("http://localhost:9009", "test-key");
+    const makeStream = () =>
+      new ReadableStream<Uint8Array>({
+        start(controller) {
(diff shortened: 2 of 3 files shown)


qiffang:
## adversary-2 review at `f36e8329` — GREEN

Correct fix, accurate root cause, no regression.

**Root cause confirmed:** `if (size === 0 || size < threshold)` — when `size` is undefined (JS callers routinely omit it), `undefined === 0` is false and `undefined < threshold` is false (NaN comparison), so a small buffer fell through to multipart/S3. The user's 7KB TiDB file was pushed to S3 this way.

**Correctness:**
- Uint8Array: `effectiveSize = data.length` — the in-memory buffer length is authoritative, ignoring the passed `size`. 7000 < 50000 → inline direct PUT. Also fixes a latent case where the caller passes a WRONG `size` (e.g. 100000 for a 7000 buffer).
- ReadableStream: now requires a finite non-negative `size` (throws otherwise). `streamToUint8Array(stream, size)` pre-allocates `new Uint8Array(size)`; with `undefined` the old code created a 0-length array and threw a cryptic RangeError on the first chunk. So the new validation breaks no working behavior — it replaces an implicit crash with a clear error message. Strict improvement.
- All `uploadSummary` calls now report `effectiveSize` (actual bytes) instead of the possibly-wrong `size`.

**Tests:** (a) small Uint8Array with omitted size → `direct_put` + total_bytes=7000 + no multipart initiate (exactly the bug); (b) ReadableStream with missing/NaN/negative size → throws. Good coverage.

**Note (non-blocking):** the ReadableStream size fail-fast is a behavior change only for callers who previously omitted size on a stream — those already crashed (RangeError), so this is a clearer contract, not a break.

**Verdict: GREEN.** Production bug (small files misrouted to S3) root-cause fixed with a precise regression test.

qiffang (commented):
GREEN at exact head `f36e8329bc65d5d5833a72323402ed228c00b23b`.

Reviewed the client-js routing fix against the agreed gate:

- `Uint8Array` now uses `data.length` as the effective size, so omitted or stale caller `size` cannot force a 7KB in-memory write into multipart/S3.
- `ReadableStream` now fails fast on missing, `NaN`, negative, or non-finite size instead of silently falling through to a bad routing/summary state.
- `UploadSummary.total_bytes` is populated from the effective size, not the raw caller argument.
- Regression coverage includes omitted-size `Uint8Array` -> `direct_put` with no `/v2/uploads/initiate`, `total_bytes === 7000`, and invalid ReadableStream size failures.

Local validation: `git diff --check origin/main...HEAD`, `npm run test` (69 passed / 19 skipped), `npm run build`, and `npm run lint` passed in `clients/drive9-js`.

No blocking findings. Merge should still wait for the remaining exact-head `code-ci` check to finish green.

Note: GitHub rejected a formal APPROVE review from this identity as "can not approve your own pull request", so this is posted as a review comment rather than an approval state change.

clients/drive9-js/src/transfer.ts
```diff
@@ -184,23 +184,35 @@ export async function writeStreamWithSummaryImpl(
...
   const threshold = client["smallFileThreshold"];
   let data: Uint8Array;
+  let effectiveSize: number;
   if (stream instanceof Uint8Array) {
+    // The byte length is authoritative for an in-memory buffer, so ignore a
+    // missing/incorrect `size` argument: JS callers routinely omit it, and
+    // an `undefined` size otherwise fails every `size < threshold` check and
+    // silently routes a small file through multipart/S3 instead of inline.
     data = stream;
+    effectiveSize = data.length;
```
mornyx, line 194: **[nit]** Using `data.length` as the authority for in-memory buffers is exactly right for the reported bug (`undefined` size → every `size < threshold` check false → multipart). Also correctly ignores a stale/wrong caller size.

Looks good.

clients/drive9-js/src/transfer.ts
```diff
@@ -184,23 +184,35 @@ export async function writeStreamWithSummaryImpl(
...
   } else {
+    if (typeof size !== "number" || !Number.isFinite(size) || size < 0) {
+      throw new Error(
+        `writeStreamWithSummary requires a finite non-negative size for a ReadableStream, got ${String(size)}`
+      );
+    }
     data = await streamToUint8Array(stream, size);
+    effectiveSize = data.length;
```
mornyx, line 202: **[nit]** Nice that this is `data.length` after `streamToUint8Array`, which already `slice(0, offset)`s to the bytes actually read. So a stream that ends short of the declared `size` still routes/summarizes on the real payload length, not the allocation hint.

No change requested.

clients/drive9-js/tests/transfer.test.ts
```diff
@@ -46,6 +46,60 @@ describe("Transfer", () => {
     expect(initiateCalled).toBe(false);
   });
 
+  it("routes a small Uint8Array through direct PUT even when size is omitted", async () => {
```
mornyx, line 49: **[nit]** Good regression for the production footgun (omit `size` on a ~7KB buffer → `direct_put`, no initiate). The cast is a bit noisy because the public signature still requires `size: number`; fine for a JS-runtime gate.

Optional follow-up: overload `writeStreamWithSummary` / `writeStream` so `size` is optional when the body is `Uint8Array`, which would also remove the need for the cast here.

mornyx (approved):
## Review summary

**Verdict: Approve**

Clear client-side fix for a real mis-route: in-memory `Uint8Array` uploads were deciding inline vs multipart from the optional caller `size`, so an omitted `undefined` failed every `size < threshold` check and pushed a 7KB buffer through multipart/S3. Using `data.length` as the effective size for buffers is the right authority; requiring a finite non-negative size for `ReadableStream` (which still needs a bound for `streamToUint8Array`) and reporting `UploadSummary.total_bytes` from the effective size close the loop.

### Correctness

1. **Uint8Array path** — `effectiveSize = data.length` ignores missing/stale `size`; small buffers cannot be forced into multipart by omission. Wrong-large caller size can no longer block inline either.
2. **ReadableStream path** — fail-fast on non-number / non-finite / negative `size` before allocate-and-read. After read, `streamToUint8Array` returns `result.slice(0, offset)`, so `effectiveSize = data.length` is **actual** bytes, not the declared size — routing and summary stay consistent when the stream ends early.
3. **Threshold gate** — still `=== 0 || < threshold` (same as before; exact threshold continues multipart).
4. **Summary** — all three modes (`direct_put` / `multipart_v2` / `multipart_v1`) pass `effectiveSize`.
5. **Tests** — regression for omitted-size 7000-byte buffer → `direct_put`, no `/v2/uploads/initiate`; stream missing/`NaN`/negative size throws. CI green (code-ci + JS SDK Node 18/22).

### Non-blocking notes

- Public TS signature still has required `size: number`. That matches the PR’s “runtime defends JS omission” story, but typed callers still must pass something; a later overload (`size` optional when the body is `Uint8Array`) would make the API honest for TS users.
- Pre-existing footgun: JS `writeStreamWithSummary(path, bytes, optionsObject)` treats the options object as `size` and drops options. Buffer routing is now fine (size ignored); options still lost. Out of scope unless you want overloads/`WriteOptions` discrimination.

No blockers. Approved.

PR #808, opened 2026-07-27
fix: improve shared pool provisioning and capacity reuse

srstack:
## Summary

- use the configured shared schema-provisioning worker count directly instead of capping it at half of the MetaDB connection pool
- keep the high-concurrency worker pool provisioning-only; pending metadata and Cloud create/adoption remain in their separately bounded workers
- freeze non-empty Cloud organization and cluster identity once provisioning starts, while allowing missing legacy identity to be repaired and connection metadata to refresh
- avoid holding a MetaDB pool-work session lock across slow remote schema DDL
- serialize same-physical-DB schema ensure across pods with an advisory lock on the target TiDB; provisioning retries stay non-blocking while foreground cold-opens wait within their request context
- treat expected cross-pod schema-lock contention as a benign current-wave completion and retry from the next durable scan
- fill existing physical shared-DB capacity transactionally before creating any new physical DB for a logical-pool refill
- heartbeat durable provisioning progress for the stuck watchdog and raise the default no-progress timeout from 15 minutes to 30 minutes
- keep full schema ensure, schema-version persistence, DB activation, and tenant activation in the same completion boundary

## Why

Shared schema DDL executes on the physical shared DB, but the previous implementation held one MetaDB `GET_LOCK` session for the full remote DDL duration. With the default MetaDB pool of 100 connections, that reduced the configured 100 provisioning workers to 50. Slow FULLTEXT/VECTOR DDL therefore limited physical-pool activation throughput and caused claimable tenant inventory to oscillate to zero.

Rows selected by the provisioning queue already have durable physical identity and connection metadata. They now run only the schema-completion path and cannot fall back into pending metadata or Cloud-create work. Those earlier lifecycle stages retain their existing bounded workers and coordination. The transition to provisioning requires a non-blank cluster ID. Once a non-empty organization and cluster identity is durable, later metadata polls cannot reassign it; a malformed legacy provisioning row with missing identity can still be repaired. Endpoint and credential metadata remain refreshable.

To preserve cross-process exclusion during leader handoff, schema ensure acquires a session advisory lock on a dedicated connection to the target physical TiDB. The lock is scoped by `db_pool` ID, so independent physical DBs continue concurrently. Provisioning uses `GET_LOCK(..., 0)`: MySQL returns immediately, while TiDB clamps the timeout to its one-second minimum; duplicate durable work therefore fails fast, is treated as a benign completion of that queue wave, and retries through the next durable scan without warning/backoff churn. Foreground tenant cold-opens wait in one-second lock calls bounded by the request context, so an active tenant does not receive an opaque busy error during a schema-version migration. No MetaDB connection is retained across DDL. If the target TiDB session itself is lost mid-DDL, the advisory lock is released; this is the remaining narrow overlap window.

Long schema attempts refresh `db_pool.updated_at` after one quarter of the configured stuck timeout and at the same interval thereafter. Fast failures finish before the first heartbeat, so repeated quick failures cannot keep a genuinely stuck pool alive indefinitely. A transient heartbeat UPDATE failure is logged and retried on the next tick; it never cancels target-TiDB schema work already in progress. The default no-progress timeout is now 30 minutes to tolerate transient TiDB schema-init delays without discarding an already-created cluster.

Shared logical-pool refill now allocates each durable wave into existing `active`, `provisioning`, then `pending` physical DB capacity before inserting any new `db_pool` row. Candidate rows are locked in deterministic order inside the same transaction as tenant, placement, membership, quota, and embedding-profile staging; each physical DB receives one batched `tenant_count += assigned` update. Only the remainder after all reusable capacity is full produces new physical DBs and enters the Cloud-create batch. This prevents small periodic refill gaps from creating a succession of mostly empty physical clusters.

## Schema completeness

This does not activate a physical DB early. Mandatory tables, schema contract validation, all FULLTEXT/VECTOR DDL, schema-version persistence, physical DB activation, and tenant activation still complete in order before inventory becomes claimable.

A live dev TiDB probe also confirmed that these columnar indexes cannot be combined into one multi-schema `ALTER`: TiDB returns `ERROR 8200: Unsupported multi schema change for add columnar index`. The per-DB DDL sequence therefore remains unchanged; this PR increases concurrency across independent physical DBs.

## Verification

Current head `4dfa2ed9`:

- `go test ./pkg/server -run 'TestBlockingDatabaseEnsurerSignalsStartedOnceAcrossRetries|TestManagedSharedDBProvisioningQueueErrorSkipsExpectedContention|TestManagedSharedDBLockedContinuationDecryptsRootPasswordOnce|TestManagedSharedDBProvisioningHeartbeatFailureDoesNotCancelSchemaWork|TestManagedSharedDBProvisioningHeartbeatPreventsStuckFailure' -count=1`
- `go test ./pkg/tenant -run 'TestSharedDBSchemaEnsureUsesCrossPoolAdvisoryLock|TestSharedDBColdOpensDoNotBlockDifferentDBPools' -count=1`
- `go test -race ./pkg/tenant -run 'TestSharedDBSchemaEnsureUsesCrossPoolAdvisoryLock' -count=1`
- `git diff --check`
- `go test ./pkg/server -run 'TestSharedTenantPoolRefill|TestAdminTenantPoolReplenish(SubmitsLargeRefillAsSingleWave|ContinuesPastWatermarkToFillSlots)' -count=1`

Focused checks pass locally. Hosted checks for the current head are running.

## Summary by CodeRabbit

* **New Features**
  * More aggressive reuse of existing shared database capacity during shared-tenant pool refills.
  * Shared DB schema ensure is now coordinated across instances to prevent conflicting schema/version updates.
* **Bug Fixes**
  * Increased default “stuck” detection timeout from 15 to 30 minutes (including the documented configuration default).
  * Managed shared DB provisioning now refreshes progress during long remote operations to reduce false failures.
  * Enforced identity consistency for managed shared DB pools and added clearer handling for non-provisioning states.
  * Provisioning concurrency now directly follows the configured worker limit.
* **Tests**
  * Expanded coverage for capacity reuse, shared-schema coordination, and managed shared DB provisioning/continuation behaviors.

### DIFF
--- a/pkg/meta/db_pool.go
+++ b/pkg/meta/db_pool.go
@@ -295,8 +295,12 @@ func (s *Store) UpdateManagedSharedDBPoolCloudResult(ctx context.Context, in *Sh
 	if in.TiDBCloudOrganizationID == "" {
 		return fmt.Errorf("tidbcloud organization id is required")
 	}
-	if in.ClusterID == "" {
+	clusterID := strings.TrimSpace(in.ClusterID)
+	if clusterID == "" {
 		return fmt.Errorf("cluster id is required")
+	}
+	if clusterID != in.ClusterID {
+		return fmt.Errorf("cluster id must be an exact value")
 	}
 	if in.Port < 0 {
 		return fmt.Errorf("db port must not be negative")
@@ -345,7 +349,8 @@ func (s *Store) UpdateManagedSharedDBPoolCloudResult(ctx context.Context, in *Sh
 		// allocation identity cannot be reassigned underneath them. Connection
 		// metadata remains refreshable for endpoint and credential rotation.
 		if currentStatus == SharedDBStatusProvisioning &&
-			(currentOrganizationID != in.TiDBCloudOrganizationID || currentClusterID.String != in.ClusterID) {
+			((strings.TrimSpace(currentOrganizationID) != "" && currentOrganizationID != in.TiDBCloudOrganizationID) ||
+				(strings.TrimSpace(currentClusterID.String) != "" && currentClusterID.String != clusterID)) {
 			return fmt.Errorf("%w: db pool %d", ErrSharedDBPoolIdentityConflict, in.ID)
 		}
 		host := currentHost.String
@@ -369,7 +374,7 @@ func (s *Store) UpdateManagedSharedDBPoolCloudResult(ctx context.Context, in *Sh
 			name = in.Name
 		}
 		nextStatus := currentStatus
-		metadataReady := host != "" && port > 0 && user != "" && len(password) > 0 && name != ""
+		metadataReady := clusterID != "" && host != "" && port > 0 && user != "" && len(password) > 0 && name != ""
 		if currentStatus == SharedDBStatusPending && metadataReady {
 			nextStatus = SharedDBStatusProvisioning
 		}
@@ -380,7 +385,7 @@ func (s *Store) UpdateManagedSharedDBPoolCloudResult(ctx context.Context, in *Sh
 				db_name = COALESCE(?, db_name), db_tls = ?,
 				status_updated_at = CASE WHEN status <> ? THEN CURRENT_TIMESTAMP(3) ELSE status_updated_at END,
 				status = ?
-			WHERE db_id = ?`, in.TiDBCloudOrganizationID, in.ClusterID,
+			WHERE db_id = ?`, in.TiDBCloudOrganizationID, clusterID,
 			nullString(in.Host), nullInt(in.Port), nullString(in.User), nullBytes(in.PasswordCipher),
 			nullString(in.Name), in.TLSMode, nextStatus, nextStatus, in.ID); err != nil {
 			return fmt.Errorf("persist cloud result for db pool %d: %w", in.ID, err)
--- a/pkg/meta/db_pool.go
+++ b/pkg/meta/db_pool.go
@@ -324,13 +324,14 @@ func (s *Store) UpdateManagedSharedDBPoolCloudResult(ctx context.Context, in *Sh
 	}
 	var advancedTenants []string
 	err = s.InTx(ctx, func(tx *sql.Tx) error {
-		var currentHost, currentUser, currentName sql.NullString
+		var currentOrganizationID string
+		var currentClusterID, currentHost, currentUser, currentName sql.NullString
 		var currentPort sql.NullInt64
 		var currentPassword []byte
 		var currentStatus string
-		if err := tx.QueryRowContext(ctx, `SELECT db_host, db_port, db_user, db_password, db_name, status
+		if err := tx.QueryRowContext(ctx, `SELECT org_id, cluster_id, db_host, db_port, db_user, db_password, db_name, status
 			FROM db_pool WHERE db_id = ? FOR UPDATE`, in.ID).
-			Scan(&currentHost, &currentPort, &currentUser, &currentPassword, &currentName, &currentStatus); err != nil {
+			Scan(&currentOrganizationID, &currentClusterID, &currentHost, &currentPort, &currentUser, &currentPassword, &currentName, &currentStatus); err != nil {
 			if errors.Is(err, sql.ErrNoRows) {
 				return ErrNotFound
 			}
@@ -338,6 +339,14 @@ func (s *Store) UpdateManagedSharedDBPoolCloudResult(ctx context.Context, in *Sh
 		}
 		if currentStatus != SharedDBStatusPending && currentStatus != SharedDBStatusProvisioning {
 			return ErrNotFound
+		}
+		// Once complete metadata advances the pool to provisioning, finishers may
+		// safely run without retaining this MetaDB row lock only if the Cloud
+		// allocation identity cannot be reassigned underneath them. Connection
+		// metadata remains refreshable for endpoint and credential rotation.
+		if currentStatus == SharedDBStatusProvisioning &&
+			(currentOrganizationID != in.TiDBCloudOrganizationID || currentClusterID.String != in.ClusterID) {
+			return fmt.Errorf("%w: db pool %d", ErrSharedDBPoolIdentityConflict, in.ID)
 		}
 		host := currentHost.String
 		if in.Host != "" {
@@ -521,7 +530,7 @@ func (s *Store) RefreshManagedSharedDBProvisioningProgress(ctx context.Context, 
 			return err
 		}
 		if status != SharedDBStatusProvisioning {
-			return ErrNotFound
+			return fmt.Errorf("%w: db pool %d has status %q", ErrSharedDBPoolNotProvisioning, dbID, status)
 		}
 	}
 	return nil
--- a/pkg/meta/db_pool.go
+++ b/pkg/meta/db_pool.go
@@ -494,6 +494,39 @@ func (s *Store) UpdateSharedDBSchemaVersion(ctx context.Context, dbID int64, ver
 	return nil
 }
 
+// RefreshManagedSharedDBProvisioningProgress keeps a long-running remote
+// schema operation visible to the stuck-pool watchdog without retaining a
+// MetaDB connection or advisory lock for the duration of the DDL.
+func (s *Store) RefreshManagedSharedDBProvisioningProgress(ctx context.Context, dbID int64) (err error) {
+	start := time.Now()
+	defer observeMeta(ctx, "refresh_managed_shared_db_provisioning_progress", start, &err)
+	if dbID <= 0 {
+		return fmt.Errorf("db_id must be positive")
+	}
+	res, err := s.db.ExecContext(ctx, `UPDATE db_pool SET updated_at = CURRENT_TIMESTAMP(3)
+		WHERE db_id = ? AND role = ? AND status = ?`, dbID, SharedDBRoleShared, SharedDBStatusProvisioning)
+	if err != nil {
+		return fmt.Errorf("refresh provisioning progress for db pool %d: %w", dbID, err)
+	}
+	affected, err := res.RowsAffected()
+	if err != nil {
+		return fmt.Errorf("refresh provisioning progress rows affected for db pool %d: %w", dbID, err)
+	}
+	if affected == 0 {
+		var status string
+		if err := s.db.QueryRowContext(ctx, `SELECT status FROM db_pool WHERE db_id = ? AND role = ?`, dbID, SharedDBRoleShared).Scan(&status); err != nil {
+			if errors.Is(err, sql.ErrNoRows) {
+				return ErrNotFound
+			}
+			return err
+		}
+		if status != SharedDBStatusProvisioning {
+			return ErrNotFound
+		}
+	}
+	return nil
+}
+
 // ActivateSharedDBPool flips a managed pool to active only after the cloud
 // identity, connection fields, and shared schema version are complete.
 func (s *Store) ActivateSharedDBPool(ctx context.Context, dbID int64) (err error) {
--- a/pkg/meta/db_pool_test.go
+++ b/pkg/meta/db_pool_test.go
@@ -731,7 +731,8 @@ func TestManagedSharedDBCloudResultRejectsIdentityChangeAfterProvisioning(t *tes
 		t.Fatalf("GetSharedDB after connection refresh: %v", err)
 	}
 	if got.TiDBCloudOrganizationID != ready.TiDBCloudOrganizationID || got.ClusterID != ready.ClusterID ||
-		got.Host != refreshed.Host || got.User != refreshed.User || !bytes.Equal(got.PasswordCipher, refreshed.PasswordCipher) {
+		got.Host != refreshed.Host || got.User != refreshed.User ||
+		!bytes.Equal(got.PasswordCipher, refreshed.PasswordCipher) || got.TLSMode != refreshed.TLSMode {
 		t.Fatalf("connection refresh = %+v, want stable identity with refreshed connection metadata", got)
 	}
 }
--- a/pkg/meta/db_pool_test.go
+++ b/pkg/meta/db_pool_test.go
@@ -737,6 +737,68 @@ func TestManagedSharedDBCloudResultRejectsIdentityChangeAfterProvisioning(t *tes
 	}
 }
 
+func TestManagedSharedDBCloudResultRejectsBlankClusterIdentity(t *testing.T) {
+	s := newControlStore(t)
+	ctx := context.Background()
+	spendingLimit := MaxTiDBCloudSpendingLimit
+	dbID, err := s.CreateManagedSharedDBPool(ctx, &SharedDB{
+		TiDBCloudOrganizationID: "org-blank-cluster", ProvisioningKey: bytes.Repeat([]byte{3}, 32),
+		CloudProvider: "aws", Region: "us-east-1", MaxTenants: 100, SpendingLimit: &spendingLimit,
+	})
+	if err != nil {
+		t.Fatalf("CreateManagedSharedDBPool: %v", err)
+	}
+	err = s.UpdateManagedSharedDBPoolCloudResult(ctx, &SharedDB{
+		ID: dbID, TiDBCloudOrganizationID: "org-blank-cluster", ClusterID: "   ",
+		Host: "shared.example.com", Port: 4000, User: "root", PasswordCipher: []byte("cipher"),
+		Name: "tidbcloud_fs", TLSMode: "skip-verify",
+	})
+	if err == nil {
+		t.Fatal("blank cluster identity error = nil, want rejection")
+	}
+	got, err := s.GetSharedDB(ctx, dbID)
+	if err != nil {
+		t.Fatalf("GetSharedDB: %v", err)
+	}
+	if got.Status != SharedDBStatusPending || got.ClusterID != "" {
+		t.Fatalf("pool after blank cluster identity = status %q cluster %q, want pending with no cluster", got.Status, got.ClusterID)
+	}
+}
+
+func TestManagedSharedDBCloudResultRepairsMissingProvisioningIdentity(t *testing.T) {
+	s := newControlStore(t)
+	ctx := context.Background()
+	spendingLimit := MaxTiDBCloudSpendingLimit
+	dbID, err := s.CreateManagedSharedDBPool(ctx, &SharedDB{
+		TiDBCloudOrganizationID: "org-repair-identity", ProvisioningKey: bytes.Repeat([]byte{4}, 32),
+		CloudProvider: "aws", Region: "us-east-1", MaxTenants: 100, SpendingLimit: &spendingLimit,
+	})
+	if err != nil {
+		t.Fatalf("CreateManagedSharedDBPool: %v", err)
+	}
+	ready := &SharedDB{
+		ID: dbID, TiDBCloudOrganizationID: "org-repair-identity", ClusterID: "cluster-repair-identity",
+		Host: "shared.example.com", Port: 4000, User: "root", PasswordCipher: []byte("cipher"),
+		Name: "tidbcloud_fs", TLSMode: "skip-verify",
+	}
+	if err := s.UpdateManagedSharedDBPoolCloudResult(ctx, ready); err != nil {
+		t.Fatalf("persist ready cloud result: %v", err)
+	}
+	if _, err := s.DB().ExecContext(ctx, `UPDATE db_pool SET cluster_id = NULL WHERE db_id = ?`, dbID); err != nil {
+		t.Fatalf("simulate legacy missing provisioning identity: %v", err)
+	}
+	if err := s.UpdateManagedSharedDBPoolCloudResult(ctx, ready); err != nil {
+		t.Fatalf("repair missing provisioning identity: %v", err)
+	}
+	got, err := s.GetSharedDB(ctx, dbID)
+	if err != nil {
+		t.Fatalf("GetSharedDB: %v", err)
+	}
+	if got.Status != SharedDBStatusProvisioning || got.ClusterID != ready.ClusterID {
+		t.Fatalf("repaired pool = status %q cluster %q, want provisioning cluster %q", got.Status, got.ClusterID, ready.ClusterID)
+	}
+}
+
 func TestRefreshManagedSharedDBProvisioningProgressDistinguishesTerminalStatus(t *testing.T) {
 	s := newControlStore(t)
 	ctx := context.Background()
--- a/pkg/meta/db_pool_test.go
+++ b/pkg/meta/db_pool_test.go
@@ -674,6 +674,102 @@ func TestManagedSharedDBUpdatedAtTracksProgressNotNoOpPolls(t *testing.T) {
 	}
 }
 
+func TestManagedSharedDBCloudResultRejectsIdentityChangeAfterProvisioning(t *testing.T) {
+	s := newControlStore(t)
+	ctx := context.Background()
+	spendingLimit := MaxTiDBCloudSpendingLimit
+	dbID, err := s.CreateManagedSharedDBPool(ctx, &SharedDB{
+		TiDBCloudOrganizationID: "org-identity", ProvisioningKey: bytes.Repeat([]byte{1}, 32),
+		CloudProvider: "aws", Region: "us-east-1", MaxTenants: 100, SpendingLimit: &spendingLimit,
+	})
+	if err != nil {
+		t.Fatalf("CreateManagedSharedDBPool: %v", err)
+	}
+	ready := &SharedDB{
+		ID: dbID, TiDBCloudOrganizationID: "org-identity", ClusterID: "cluster-identity",
+		Host: "shared.example.com", Port: 4000, User: "root", PasswordCipher: []byte("cipher"),
+		Name: "tidbcloud_fs", TLSMode: "skip-verify",
+	}
+	if err := s.UpdateManagedSharedDBPoolCloudResult(ctx, ready); err != nil {
+		t.Fatalf("persist ready cloud result: %v", err)
+	}
+
+	mutations := []struct {
+		name   string
+		mutate func(*SharedDB)
+	}{
+		{name: "organization", mutate: func(in *SharedDB) { in.TiDBCloudOrganizationID = "org-other" }},
+		{name: "cluster", mutate: func(in *SharedDB) { in.ClusterID = "cluster-other" }},
+	}
+	for _, tt := range mutations {
+		t.Run(tt.name, func(t *testing.T) {
+			changed := *ready
+			tt.mutate(&changed)
+			if err := s.UpdateManagedSharedDBPoolCloudResult(ctx, &changed); !errors.Is(err, ErrSharedDBPoolIdentityConflict) {
+				t.Fatalf("identity mutation error = %v, want ErrSharedDBPoolIdentityConflict", err)
+			}
+			got, err := s.GetSharedDB(ctx, dbID)
+			if err != nil {
+				t.Fatalf("GetSharedDB: %v", err)
+			}
+			if got.TiDBCloudOrganizationID != ready.TiDBCloudOrganizationID || got.ClusterID != ready.ClusterID {
+				t.Fatalf("identity after rejected mutation = org %q cluster %q, want org %q cluster %q",
+					got.TiDBCloudOrganizationID, got.ClusterID, ready.TiDBCloudOrganizationID, ready.ClusterID)
+			}
+		})
+	}
+	refreshed := *ready
+	refreshed.Host = "rotated.example.com"
+	refreshed.User = "rotated.root"
+	refreshed.PasswordCipher = []byte("rotated-cipher")
+	refreshed.TLSMode = "true"
+	if err := s.UpdateManagedSharedDBPoolCloudResult(ctx, &refreshed); err != nil {
+		t.Fatalf("refresh connection metadata with stable identity: %v", err)
+	}
+	got, err := s.GetSharedDB(ctx, dbID)
+	if err != nil {
+		t.Fatalf("GetSharedDB after connection refresh: %v", err)
+	}
+	if got.TiDBCloudOrganizationID != ready.TiDBCloudOrganizationID || got.ClusterID != ready.ClusterID ||
+		got.Host != refreshed.Host || got.User != refreshed.User || !bytes.Equal(got.PasswordCipher, refreshed.PasswordCipher) {
+		t.Fatalf("connection refresh = %+v, want stable identity with refreshed connection metadata", got)
+	}
+}
+
+func TestRefreshManagedSharedDBProvisioningProgressDistinguishesTerminalStatus(t *testing.T) {
+	s := newControlStore(t)
+	ctx := context.Background()
+	spendingLimit := MaxTiDBCloudSpendingLimit
+	dbID, err := s.CreateManagedSharedDBPool(ctx, &SharedDB{
+		TiDBCloudOrganizationID: "org-heartbeat-status", ProvisioningKey: bytes.Repeat([]byte{2}, 32),
+		CloudProvider: "aws", Region: "us-east-1", MaxTenants: 100, SpendingLimit: &spendingLimit,
+	})
+	if err != nil {
+		t.Fatalf("CreateManagedSharedDBPool: %v", err)
+	}
+	if err := s.UpdateManagedSharedDBPoolCloudResult(ctx, &SharedDB{
+		ID: dbID, TiDBCloudOrganizationID: "org-heartbeat-status", ClusterID: "cluster-heartbeat-status",
+		Host: "shared.example.com", Port: 4000, User: "root", PasswordCipher: []byte("cipher"),
+		Name: "tidbcloud_fs", TLSMode: "skip-verify",
+	}); err != nil {
+		t.Fatalf("persist ready cloud result: %v", err)
+	}
+	if err := s.UpdateSharedDBSchemaVersion(ctx, dbID, 1); err != nil {
+		t.Fatalf("UpdateSharedDBSchemaVersion: %v", err)
+	}
+	if err := s.ActivateSharedDBPool(ctx, dbID); err != nil {
+		t.Fatalf("ActivateSharedDBPool: %v", err)
+	}
+
+	err = s.RefreshManagedSharedDBProvisioningProgress(ctx, dbID)
+	if !errors.Is(err, ErrSharedDBPoolNotProvisioning) || errors.Is(err, ErrNotFound) {
+		t.Fatalf("refresh active pool error = %v, want ErrSharedDBPoolNotProvisioning distinct from ErrNotFound", err)
+	}
+	if err := s.RefreshManagedSharedDBProvisioningProgress(ctx, dbID+1_000_000); !errors.Is(err, ErrNotFound) {
+		t.Fatalf("refresh missing pool error = %v, want ErrNotFound", err)
+	}
+}
+
 func TestFinalizeFailedSharedDBPoolCleanupClearsCloudIdentityBeforeDelete(t *testing.T) {
 	s := newControlStore(t)
 	ctx := context.Background()
--- a/pkg/meta/shared_pool_wave.go
+++ b/pkg/meta/shared_pool_wave.go
@@ -234,7 +234,9 @@ func lockManagedSharedDBPoolWaveCandidates(ctx context.Context, tx *sql.Tx, orga
 	// Acquire every reusable row for the organization in immutable primary-key
 	// order. Status can advance concurrently, so using it in the locking order
 	// can invert acquisition between overlapping refill waves. The result is
-	// sorted by readiness only after the complete stable lock set is held.
+	// sorted by readiness only after the complete stable lock set is held. This
+	// deliberately serializes same-organization refill waves for the duration
+	// of this short transaction; no Cloud or schema work runs while it is held.
 	rows, err := tx.QueryContext(ctx, lockManagedSharedDBPoolWaveCandidatesSQL, organizationID, SharedDBRoleShared,
 		SharedDBStatusActive, SharedDBStatusProvisioning, SharedDBStatusPending)
 	if err != nil {
@@ -500,6 +502,8 @@ func finalizeManagedSharedDBPoolWaveMembers(ctx context.Context, tx *sql.Tx, sta
 		multiRowPlaceholders(len(staged.tenantIDs), 6), staged.membershipArgs...); err != nil {
 		return fmt.Errorf("insert wave memberships: %w", err)
 	}
+	// MySQL evaluates single-table SET assignments from left to right. Keep
+	// soft_cap_reached first so both calculations use the pre-increment count.
 	res, err := tx.ExecContext(ctx, `UPDATE db_pool
 		SET soft_cap_reached = CASE WHEN tenant_count + ? >= max_tenants THEN 1 ELSE 0 END,
 			tenant_count = tenant_count + ?
--- a/pkg/meta/shared_pool_wave.go
+++ b/pkg/meta/shared_pool_wave.go
@@ -5,6 +5,7 @@ import (
 	"database/sql"
 	"errors"
 	"fmt"
+	"sort"
 	"strings"
 	"time"
 )
@@ -144,7 +145,7 @@ func (s *Store) CreateManagedSharedDBPoolTenantWaveWithResize(ctx context.Contex
 		staged := make([]stagedManagedSharedDBPoolWavePlan, 0, len(prepared))
 		txErr := s.InTx(ctx, func(tx *sql.Tx) error {
 			memberOffset := 0
-			candidates, candidateErr := lockManagedSharedDBPoolWaveCandidates(ctx, tx, waveOrganizationID, len(waveMembers))
+			candidates, candidateErr := lockManagedSharedDBPoolWaveCandidates(ctx, tx, waveOrganizationID)
 			if candidateErr != nil {
 				return candidateErr
 			}
@@ -222,13 +223,20 @@ func (s *Store) CreateManagedSharedDBPoolTenantWaveWithResize(ctx context.Contex
 	return out, nil
 }
 
-func lockManagedSharedDBPoolWaveCandidates(ctx context.Context, tx *sql.Tx, organizationID string, limit int) ([]managedSharedDBPoolWaveCandidate, error) {
-	rows, err := tx.QueryContext(ctx, `SELECT db_id, status, max_tenants, tenant_count
+const lockManagedSharedDBPoolWaveCandidatesSQL = `SELECT db_id, status, max_tenants, tenant_count
 		FROM db_pool
 		WHERE org_id = ? AND role = ? AND status IN (?, ?, ?)
 			AND max_tenants > 0 AND soft_cap_reached = 0 AND tenant_count < max_tenants
-		ORDER BY CASE status WHEN 'active' THEN 0 WHEN 'provisioning' THEN 1 ELSE 2 END, db_id
-		LIMIT ? FOR UPDATE`, organizationID, SharedDBRoleShared, SharedDBStatusActive, SharedDBStatusProvisioning, SharedDBStatusPending, limit)
+		ORDER BY db_id
+		FOR UPDATE`
+
+func lockManagedSharedDBPoolWaveCandidates(ctx context.Context, tx *sql.Tx, organizationID string) ([]managedSharedDBPoolWaveCandidate, error) {
+	// Acquire every reusable row for the organization in immutable primary-key
+	// order. Status can advance concurrently, so using it in the locking order
+	// can invert acquisition between overlapping refill waves. The result is
+	// sorted by readiness only after the complete stable lock set is held.
+	rows, err := tx.QueryContext(ctx, lockManagedSharedDBPoolWaveCandidatesSQL, organizationID, SharedDBRoleShared,
+		SharedDBStatusActive, SharedDBStatusProvisioning, SharedDBStatusPending)
 	if err != nil {
 		return nil, fmt.Errorf("lock reusable managed shared DB pools: %w", err)
 	}
@@ -244,7 +252,30 @@ func lockManagedSharedDBPoolWaveCandidates(ctx context.Context, tx *sql.Tx, orga
 	if err := rows.Err(); err != nil {
 		return nil, err
 	}
+	sortManagedSharedDBPoolWaveCandidates(out)
 	return out, nil
+}
+
+func sortManagedSharedDBPoolWaveCandidates(candidates []managedSharedDBPoolWaveCandidate) {
+	statusRank := func(status string) int {
+		switch status {
+		case SharedDBStatusActive:
+			return 0
+		case SharedDBStatusProvisioning:
+			return 1
+		case SharedDBStatusPending:
+			return 2
+		default:
+			return 3
+		}
+	}
+	sort.Slice(candidates, func(i, j int) bool {
+		left, right := statusRank(candidates[i].status), statusRank(candidates[j].status)
+		if left != right {
+			return left < right
+		}
+		return candidates[i].dbID < candidates[j].dbID
+	})
 }
 
 func tenantStatusForSharedDBPoolWave(status string) (TenantStatus, error) {
@@ -470,8 +501,8 @@ func finalizeManagedSharedDBPoolWaveMembers(ctx context.Context, tx *sql.Tx, sta
 		return fmt.Errorf("insert wave memberships: %w", err)
 	}
 	res, err := tx.ExecContext(ctx, `UPDATE db_pool
-		SET tenant_count = tenant_count + ?,
-			soft_cap_reached = CASE WHEN tenant_count + ? >= max_tenants THEN 1 ELSE 0 END
+		SET soft_cap_reached = CASE WHEN tenant_count + ? >= max_tenants THEN 1 ELSE 0 END,
+			tenant_count = tenant_count + ?
 		WHERE db_id = ? AND status = ? AND max_tenants > 0 AND soft_cap_reached = 0
 			AND tenant_count + ? <= max_tenants`,
 		len(staged.tenantIDs), len(staged.tenantIDs), staged.dbID, staged.dbStatus, len(staged.tenantIDs))
--- a/pkg/meta/shared_pool_wave.go
+++ b/pkg/meta/shared_pool_wave.go
@@ -25,11 +25,13 @@ type ManagedSharedDBPoolWavePlan struct {
 	Members []ManagedSharedDBPoolWaveMember
 }
 
-// ManagedSharedDBPoolWaveResult contains durable identifiers created for one
-// physical partition of a refill wave.
+// ManagedSharedDBPoolWaveResult contains one durable physical assignment for a
+// refill wave. Created reports whether the wave inserted the physical pool or
+// reused capacity from an existing row.
 type ManagedSharedDBPoolWaveResult struct {
 	DBID      int64
 	TenantIDs []string
+	Created   bool
 }
 
 // ManagedSharedDBPoolWaveResize atomically grows the logical tenant pool with
@@ -41,15 +43,24 @@ type ManagedSharedDBPoolWaveResize struct {
 }
 
 type preparedManagedSharedDBPoolWavePlan struct {
-	values  managedSharedDBInsertValues
-	members []ManagedSharedDBPoolWaveMember
+	values       managedSharedDBInsertValues
+	members      []ManagedSharedDBPoolWaveMember
+	tenantStatus TenantStatus
 }
 
 type stagedManagedSharedDBPoolWavePlan struct {
 	dbID           int64
+	dbStatus       string
 	tenantIDs      []string
 	placementArgs  []any
 	membershipArgs []any
+}
+
+type managedSharedDBPoolWaveCandidate struct {
+	dbID        int64
+	status      string
+	maxTenants  int
+	tenantCount int
 }
 
 func multiRowPlaceholders(rows, columns int) string {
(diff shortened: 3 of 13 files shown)


pkg/server/shared_db_pool.go
```diff
@@ -1022,6 +1022,14 @@ func (s *Server) continueManagedSharedDBPoolOnce(ctx context.Context, dbID int64
 		if err != nil {
 			return err
 		}
+		// Provisioning rows have already durably committed their physical
+		// identity and connection metadata. Schema ensure and activation are
+		// idempotent, so leader recovery can run them without holding a MetaDB
+		// session lock for the entire remote DDL operation. Pending rows still
+		// use the work lock below to prevent duplicate physical creation.
+		if poolInfo.Status == meta.SharedDBStatusProvisioning && managedSharedDBConnectionMetadataComplete(poolInfo) {
+			return s.finishManagedSharedDBProvisioning(ctx, poolInfo, cred)
```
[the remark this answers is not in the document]
srstack: Fixed in d2e79255. The complete-metadata provisioning path still does not retain a MetaDB session or advisory lock across remote DDL. Instead, a long-running attempt refreshes db_pool.updated_at every one quarter of the configured stuck timeout, so the watchdog CAS sees durable activity while DDL is in flight. Fast failures return before the first heartbeat and therefore do not keep a genuinely stuck pool alive through frequent retries. I also raised the default stuck timeout from 15m to 30m and added a regression test that ages a provisioning row, blocks schema work past the heartbeat, runs the watchdog, and verifies the pool remains provisioning.

pkg/server/shared_db_pool.go
```diff
@@ -1281,6 +1295,30 @@ func (s *Server) continueManagedSharedDBPoolLocked(ctx context.Context, poolInfo
 	if poolInfo.Host == "" || poolInfo.Port <= 0 || poolInfo.User == "" || len(poolInfo.PasswordCipher) == 0 || poolInfo.Name == "" {
 		return fmt.Errorf("%w: db pool %d", errSharedDBConnectionMetadataNotReady, dbID)
 	}
+	return s.finishManagedSharedDBProvisioning(ctx, poolInfo, cred)
```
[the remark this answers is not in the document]
srstack: Fixed in c5fbc80f. I removed the duplicate pre-check; finishManagedSharedDBProvisioning now owns the single managedSharedDBConnectionMetadataComplete validation and returns the existing metadata-not-ready error when the durable fields are incomplete.

srstack:
## Review — `perf: decouple shared schema provisioning concurrency`

Reviewed at `b70fd583`. Verified locally in a worktree:

- `go test ./pkg/server -run 'TestManagedSharedDBProvisioning(SlotsIgnoreMetaConnectionBudget|DoesNotWaitForPoolWorkLock)|TestRunManagedSharedDBProvisioningQueue' -count=1` — ok
- `go test ./pkg/server -count=1` (full package) — ok, 177s

### Overview

Two coupled changes to unblock physical-pool activation throughput:

1. Deletes `limitManagedSharedDBProvisioningWorkers`, so the configured worker count (default 100) is used directly instead of being capped at `metaMaxOpen/2` (50 with the default 100-connection meta pool).
2. Adds an early lock-free path in `continueManagedSharedDBPoolOnce`: rows in `provisioning` status whose connection metadata is already durable skip `withSharedDBPoolWorkLock` and go straight to the new `finishManagedSharedDBProvisioning`. Pending rows and physical create/adoption keep the lock.

The diagnosis is sound and well-evidenced. Holding one MetaDB `GET_LOCK` session across minutes of remote FULLTEXT/VECTOR DDL genuinely does halve effective provisioning throughput, and I confirmed the mechanism: `withSharedDBPoolWorkLocks` calls `s.db.Conn(ctx)` and pins that dedicated connection for the entire callback (`pkg/meta/meta.go:3042`), which is exactly why the `/2` reservation existed. The extraction of `finishManagedSharedDBProvisioning` is clean, the ordering claim in the description holds (schema ensure → contract validation → optional columnar DDL → `ActivateSharedDBPool` → tenant activation, all before inventory is claimable), and probing the `ERROR 8200 Unsupported multi schema change` limit to rule out batching the DDL is good diligence.

### Blocking: the lock removal trades cross-process mutual exclusion for an idempotency assumption that the schema-repair path may not satisfy

The doc comment justifies dropping the lock with "leader-only consumption and per-process job dedup prevent normal overlap and the operations below are safe to retry after a leader handoff."

The first two guards are weaker than the lock they replace:

- `managedSharedDBProvisioningJobs` is a `sync.Map` on `*Server` — **per-process only**. It cannot see another pod.
- Leader-only consumption is not mutual exclusion across a handoff. `startManagedSharedDBWorker` checks `s.leader.IsLeader()` at *spawn* time; the DDL then runs for minutes. During a handoff (lease expiry, network partition, rolling restart) the old leader can still be inside `finishManagedSharedDBProvisioning` while the new leader starts its own continuation for the same `dbID`. `leaderWorkerCtx` cancellation helps only once the old process notices it lost leadership.

The previous `GET_LOCK` covered exactly that window. So the remaining protection is "these operations are idempotent." That's true for `ActivateSharedDBPool` and `ActivateSharedTenantsBatch` (status/batch updates), but I'm not convinced it holds for `EnsureSharedSchema` under genuinely concurrent execution against the same physical DB:

```go
// pkg/tenant/schema/shared.go:88
ExecSchemaStatementsParallelByTableContext(ctx, db, stmts)   // parallel DDL
ensureSharedSchemaContract(ctx, db, stmts)                    // diff → repair, up to 6 passes
ExecOptionalSchemaStatements(ctx, TiDBSharedOptionalSchemaStatements())
```

`ensureSharedSchemaContract` loops up to 6 times: diff the live tables against the spec, then apply repair ALTERs. Two processes running this simultaneously each observe the *other's* half-applied DDL as a diff and try to repair it. Plausible outcomes are duplicate-index errors (1061), `ERROR 8200` on concurrent columnar index adds, or a `sharedSchemaDiffError` when a diff isn't classified safe — any of which surfaces as a provisioning failure or a pool marked failed, on a path whose whole purpose is making inventory claimable. Retry-safe *sequentially* is not the same as safe *concurrently*, and the comment conflates the two.

Two ways to keep the throughput win without the exposure:

- **Short-lived claim instead of no lock.** Take the pool-work lock only to CAS a `schema_ensure_owner` / `schema_ensure_started_at` marker on the row, release it, then run the DDL unlocked, then take it again briefly to activate. Cross-process exclusion is preserved and no connection is pinned across the DDL — which is the actual problem being solved.
- **Non-blocking claim.** `WithSharedDBPoolWorkClaims` already exists with `waitSeconds: 0` / `requireAll: false`. Holding a lock that is *never waited on* doesn't serialize independent DBs, so it wouldn't reintroduce the throughput ceiling; it just prevents two processes touching the same `dbID`. This looks like the smallest change that keeps the guarantee.

If you've concluded the concurrent-DDL window is acceptable — the handoff is rare, and a failed pool gets retried — that's a defensible call, but it should be stated explicitly in the comment as a known race with its failure mode, rather than asserted as prevented.

### Second concern: the connection cap was removed globally, not just for the lock-free path

`limitManagedSharedDBProvisioningWorkers` is deleted outright, so the cap no longer applies to work that *still holds the lock*. `continueManagedSharedDBPoolLocked` retains the locked path for pending rows and for metadata loads, and each of those workers pins one meta connection for the duration — now including remote Cloud API calls (`LoadSharedDBPoolWithCredentials`, `BatchProvisionSharedDBPoolsWithCredentials`) plus callback queries needing a second connection from the same 100-connection pool.

In the normal flow this is bounded, because `UpdateManagedSharedDBPoolCloudResult` transitions `pending → provisioning` only once metadata is complete, so queue entries are usually `provisioning` and take the new lock-free path. But the invariant now rests on that status/metadata coupling rather than on a hard cap, and meta-pool exhaustion degrades the whole control plane, not just provisioning. Scoping the reservation to lock-holding workers — full worker count for lock-free schema work, `maxOpen/2` for anything that pins a session — would preserve the win and keep the floor.

Relatedly, the `workers <= 0` branch now logs `managed_shared_db_pool_provisioning_disabled` with `meta_max_open_connections` (`shared_db_pool.go:513-522`). That field was diagnostic for the removed cap and can no longer cause the disable, so the message is now misleading. Worth dropping the field or the branch.

### Smaller notes

- **Duplicated, subtly different metadata checks.** `continueManagedSharedDBPoolLocked:1295` inline-checks `Host/Port/User/PasswordCipher/Name`, while `managedSharedDBConnectionMetadataComplete` checks those **plus `ClusterID`** and trims whitespace. Two spellings of one predicate will drift; use the helper in both places. (The gate at :1030 requires `ClusterID`, so the stricter form is the intended one.)
- **Silent success on terminal statuses.** `finishManagedSharedDBProvisioning` returns `nil` for `active`/`failed`/`draining`. For `active` that's correct idempotence, but returning success for `failed`/`draining` means a queue job completes without doing anything and without a log line. A debug-level log would make "why did this pool never activate" answerable.
- **`SpendingLimit` sync failure is logged, not returned**, so a pool activates with an unsynced spending limit. Pre-existing behavior carried into the extracted function, just noting it survives the refactor.
- **Test timing.** `TestManagedSharedDBProvisioningDoesNotWaitForPoolWorkLock` discriminates on a 1s timeout and asserts progress indirectly (via failure to connect to `127.0.0.1:1`). It's a reasonable shape for "did not block," but the 1s threshold may flake on a loaded CI box, and the assertion passes for any non-nil error — including one unrelated to reaching the endpoint. Asserting on `errSharedDBConnectionMetadataNotReady`-vs-dial-error specifically would make it harder to pass for the wrong reason.
- Deleting `TestManagedSharedDBProvisioningWorkerLimitReservesMetaConnections` is correct given the function is gone.

### Test coverage

The two new/updated tests pin the two mechanical facts (slots ignore the meta budget; the continuation doesn't wait on the pool-work lock). What's untested is the property the change actually depends on: **that two concurrent `finishManagedSharedDBProvisioning` calls for the same `dbID` converge.** That's the load-bearing assumption, and it's the one thing no test exercises. A test running two goroutines through the function against the same test DB, asserting both return nil and the schema ends up correct, would either validate the design or surface the concurrency problem above.

### Verdict

The performance analysis is right and the refactor is clean, but I'd hold on the concurrency justification. Recommend either the short-lived-claim or non-blocking-claim variant — both keep the throughput win while preserving cross-process exclusion — or, if the race is being accepted deliberately, documenting it as accepted with its failure mode instead of describing it as prevented. The global cap removal is worth narrowing to lock-free work regardless.

srstack:
Addressed the review findings on current head `3e06e3a2`:

1. **Cross-process schema exclusion:** same-physical-DB schema ensure is now serialized with a non-blocking advisory lock on the target TiDB (`GET_LOCK(..., 0)`), scoped by `db_pool` ID. A duplicate continuation during leader handoff returns immediately and is retried by the existing durable provisioning scan/queue. Different physical DBs use different lock names and remain concurrent. The implementation deliberately does not use `WithSharedDBPoolWorkClaims`: even with a zero wait, an acquired MetaDB session lock would still pin that MetaDB connection for the full DDL callback, which is the bottleneck this PR removes. `TestSharedDBSchemaEnsureUsesCrossPoolAdvisoryLock` exercises two independent `tenant.Pool` instances against the same physical DB and verifies only one schema ensure runs, followed by a current-version retry.

2. **Worker-cap scope:** the high-concurrency provisioning queue now calls a specialized `continueManagedSharedDBProvisioningOnce`. It only accepts rows whose durable status is `provisioning`; any `pending`, `active`, `failed`, or `draining` row is skipped before credential resolution or Cloud work. Pending metadata polling and Cloud create/adoption remain in their separately bounded workers and retain their existing coordination. `TestManagedSharedDBProvisioningWorkerSkipsPendingCloudWork` verifies a pending row produces zero IAM and zero Cloud-create calls. This preserves the configured schema concurrency without allowing 100 workers into MetaDB session-lock paths.

3. **Watchdog visibility:** fixed in `d2e79255`. Long schema attempts heartbeat `db_pool.updated_at` after one quarter of the configured stuck timeout and at that interval thereafter; fast failures return before the first heartbeat. The default no-progress timeout is now 30 minutes instead of 15 minutes, reducing the chance that a transient TiDB DDL stall discards an already-created cluster.

4. **Misleading log field:** removed `meta_max_open_connections` from `managed_shared_db_pool_provisioning_disabled`; MetaDB max-open no longer determines this worker count.

5. **Metadata predicate:** fixed in `c5fbc80f`; `managedSharedDBConnectionMetadataComplete` is the single validation path.

6. **Terminal statuses:** `failed` and `draining` continuations now emit a debug event before completing; `active` remains the silent idempotent success case.

7. **Lock-bypass regression:** the old one-second/dial-error test was replaced with a deterministic `EnsureDatabase` sentinel and a bounded context. The assertion now proves the code reached schema provisioning while another session held the MetaDB pool-work lock, rather than accepting any non-nil network error.

8. **Spending-limit sync:** intentionally unchanged. Best-effort quota synchronization with a warning followed by activation predates this refactor. Changing that lifecycle contract would introduce a separate availability/fail-closed policy change unrelated to removing MetaDB pressure, so this PR keeps the existing behavior.

Focused current-head tenant/server tests and the target-lock race test pass locally. Hosted `code-ci` and local-e2e checks are still running for `3e06e3a2`.

srstack:
## Re-review at `f4769049` — blocker resolved, two new issues in the added code

Verified on this head:

- `go test ./pkg/server -count=1` — ok, 168s
- `go test ./pkg/tenant -run 'TestSharedDBSchemaEnsureUsesCrossPoolAdvisoryLock|TestSharedDB' -count=1` — ok

### The main blocker is properly resolved, and the chosen design is better than what I suggested

Serializing on an advisory lock on the **target TiDB** rather than MetaDB is the right answer, and your reasoning for rejecting `WithSharedDBPoolWorkClaims` is correct — I missed that even a zero-wait MetaDB lock still pins the MetaDB connection for the whole DDL callback, which is precisely the bottleneck being removed. Locking the resource actually being mutated is the cleaner invariant: `GET_LOCK(?, 0)` scoped per `db_pool` ID, distinct names so independent DBs stay concurrent, and no MetaDB connection retained.

The re-read inside the lock (`GetSharedDB` → return early if already at `CurrentSharedTiDBSchemaVersion`) closes the window where another pod finished between the initial metadata read and lock acquisition. `TestSharedDBSchemaEnsureUsesCrossPoolAdvisoryLock` is a real test of the property I said was untested — two independent `tenant.Pool` instances against one physical DB, asserting exactly one ensure runs plus a clean current-version retry.

Items 2 and 4–7 all check out: the provisioning queue now calls `continueManagedSharedDBProvisioningOnce`, which rejects non-`provisioning` rows before credential resolution or Cloud work, so 100 workers can no longer reach a MetaDB session-lock path; the misleading log field is gone; `managedSharedDBConnectionMetadataComplete` is now the single predicate; terminal statuses emit debug events; and the lock-bypass test uses a deterministic sentinel instead of a dial-error/timing discriminator. Agreed on leaving spending-limit sync alone — that's a separate policy decision.

### New issue 1: the busy error reaches foreground tenant requests, which previously waited

`errSharedDBSchemaEnsureBusy` is returned from `withSharedDBSchemaAdvisoryLock` and propagates straight out of `sharedDBHandle` (`pkg/tenant/pool.go:1160`). That has three callers, and one is the foreground path:

```go
// pool.go:1280 — createBackend, shared-schema tenant
handle, err := p.sharedDBHandle(ctx, placement.DbID)
if err != nil {
    return nil, nil, tidbCloudOrgID, err
}
```

`createBackend` errors return directly from `Get` (`:383`) and `Acquire` (`:469`) with no retry or backoff. So a user request for a shared tenant whose physical DB is mid-schema-ensure now fails immediately with `shared schema ensure already running`, where before the change it would block on `ensureSharedDBSchema` and succeed once DDL finished.

Fast-fail is right for the *provisioning queue* — the durable scan retries. It's wrong for a user request: the tenant is otherwise healthy, and the caller has no idea this is a transient, retryable condition. The window is minutes-long FULLTEXT/VECTOR DDL, and it opens on any schema-version bump, when every already-active tenant on that DB keeps serving traffic.

Two options:

- **Block on the foreground path only.** Pass the intent down so queue callers use `GET_LOCK(..., 0)` and `createBackend` uses a bounded positive wait (`GET_LOCK(..., N)`), restoring the old waiting behavior with a ceiling.
- **Retry at the boundary.** Keep the non-blocking lock and have `createBackend` retry on `errSharedDBSchemaEnsureBusy` with short backoff until the request deadline.

Either way the error needs to stop being an opaque failure at the tenant boundary. Worth noting the existing tests can't catch this: they exercise `EnsureSharedDBReady`, not `Acquire`/`Get` against a DB whose schema lock is held.

### New issue 2: a heartbeat failure aborts the DDL it was meant to protect

In `withManagedSharedDBProvisioningHeartbeat` (`shared_db_pool.go:1082`), the heartbeat goroutine shares `heartbeatCtx` with the work function, and on refresh error it cancels it:

```go
heartbeatCtx, cancel := context.WithCancel(ctx)
go func() {
    // ...
    if err := s.meta.RefreshManagedSharedDBProvisioningProgress(heartbeatCtx, dbID); err != nil {
        heartbeatDone <- err
        cancel()          // <-- cancels the ctx the DDL is running under
        return
    }
}()
workErr := fn(heartbeatCtx)
```

So a single transient MetaDB error on a progress UPDATE — the pool being briefly saturated is exactly the condition this PR is about — kills an in-flight schema migration that may be minutes in. The failure mode is inverted: the mechanism added to stop the watchdog from discarding long-running work now discards it itself, on an unrelated MetaDB blip.

A heartbeat is advisory. It should log and keep ticking (or tolerate N consecutive failures before giving up), not cancel the work. If you do want to abort when durable progress can't be recorded — reasonable, since the watchdog may reclaim the row — it should take several consecutive failures spanning a meaningful fraction of `timeout/4`, and the joined error should say the DDL was aborted by heartbeat failure rather than surfacing as a generic provisioning error.

Also minor: `interval = timeout / 4` is 7.5 min at the new 30 min default, so a single failed tick plus cancellation can end an attempt that had 22 min of headroom left.

### Smaller notes

- **Advisory lock survives only as long as the connection.** If the TiDB session holding `GET_LOCK` drops mid-DDL (restart, proxy timeout, network blip), the lock is released while the first pod's DDL may still be running, and a second pod can acquire it and start concurrently. Much narrower than the original unprotected window and probably acceptable, but it's the residual race — worth a line in the comment so nobody reads the lock as an absolute guarantee.
- `withSharedDBSchemaAdvisoryLock` pins one connection on the target shared DB for the DDL duration. Only one per physical DB, so not a capacity concern; noting it since the PR's whole theme is not pinning connections across DDL.
- The release path uses `context.Background()` with a 5s timeout, so it still runs when the caller's ctx is already cancelled. Correct, and matches the MetaDB lock helper's pattern.

### Verdict

The cross-process exclusion concern is genuinely resolved, and by a better mechanism than I proposed. Both new issues are in code added by this round: issue 1 turns a transient provisioning state into user-visible errors for otherwise-healthy tenants, and issue 2 makes the heartbeat capable of destroying the work it protects. Neither is hard to fix, and both are narrower than the original blocker.

srstack:
Addressed both new findings in `623f3d87`.

1. **Foreground tenant opens no longer receive the provisioning busy error.** The target-TiDB advisory lock now has caller-specific behavior:
   - `EnsureSharedDBReady`, used by the durable provisioning worker, keeps `GET_LOCK(..., 0)` and returns immediately so the queue can retry later.
   - Normal `sharedDBHandle` callers, including `Get`/`Acquire` cold-opens, wait using one-second `GET_LOCK` calls on the same session until the lock is acquired or their request context ends. A request deadline therefore returns its context error, not `errSharedDBSchemaEnsureBusy`.
   - The cross-pool regression now proves all three properties: provisioning fast-fails, a foreground open respects its deadline while the lock is held, and a foreground open resumes successfully after the first schema ensure persists the current version.

2. **A heartbeat failure no longer cancels target-TiDB DDL.** The schema function now runs on the original caller context. The heartbeat has its own child lifecycle only for stopping its ticker; a failed `RefreshManagedSharedDBProvisioningProgress` emits `managed_shared_db_pool_provisioning_heartbeat_failed` and the next scheduled tick retries. The new regression forces the refresh to return `ErrNotFound`, observes the warning, and verifies the schema work context remains live and completes.

3. **Residual advisory-lock boundary is documented.** The lock comment now states that it is session-owned: if the target TiDB connection itself is lost mid-DDL, TiDB releases the lock and a later attempt can overlap a still-finishing server-side statement. The lock still pins exactly one target-DB connection per physical DB while held; it never retains a MetaDB connection, and different physical DB IDs remain independent.

Focused tenant/server tests and the target-lock race tests pass locally on this commit. Hosted CI/e2e are running for the new head.

srstack:
Addressed all four findings from the latest CodeRabbit formal review in `a8011060`:

1. Exported `tenant.ErrSharedDBSchemaEnsureBusy`. The provisioning queue now treats target-TiDB schema-lock contention as benign completion of the current wave, alongside `meta.ErrNotFound`, so normal cross-pod contention does not emit retry warnings or consume backoff budget. The durable scan schedules a later attempt.
2. Protected the test helper `blockingDatabaseEnsurer.started` signal with `sync.Once` and added a repeated-call regression test.
3. Reordered the package-level schema hook cleanup and added release-and-drain cleanup so every failure path waits for the in-flight schema goroutine before restoring the global hook. The advisory-lock regression passes under `-race`.
4. Deferred root-password decryption in `continueManagedSharedDBPoolLocked` until the physical-create branch actually needs it. Complete-metadata provisioning now decrypts exactly once for DSN construction; a counting-encryptor regression covers this path.

Focused verification on this commit:

- `go test ./pkg/server -run 'TestBlockingDatabaseEnsurerSignalsStartedOnceAcrossRetries|TestManagedSharedDBProvisioningQueueErrorSkipsExpectedContention|TestManagedSharedDBLockedContinuationDecryptsRootPasswordOnce|TestManagedSharedDBProvisioningHeartbeatFailureDoesNotCancelSchemaWork|TestManagedSharedDBProvisioningHeartbeatPreventsStuckFailure' -count=1`
- `go test ./pkg/tenant -run 'TestSharedDBSchemaEnsureUsesCrossPoolAdvisoryLock|TestSharedDBColdOpensDoNotBlockDifferentDBPools' -count=1`
- `go test -race ./pkg/tenant -run 'TestSharedDBSchemaEnsureUsesCrossPoolAdvisoryLock' -count=1`
- `git diff --check`

All focused checks pass. Hosted checks for the new head are still pending/running.

srstack:
## Re-review at `a8011060` — both findings fixed and verified; one residual note

Verified on this head:

- `go test ./pkg/server -count=1` — ok, 168s
- `go test ./pkg/tenant -count=1` — ok, 15.4s
- `go test -race ./pkg/tenant -run 'TestSharedDBSchemaEnsureUsesCrossPoolAdvisoryLock|TestSharedDBColdOpensDoNotBlockDifferentDBPools' -count=1` — ok

### Finding 1 (foreground busy error) — fixed correctly

The `waitForLock` split is the right shape. `EnsureSharedDBReady` keeps `GET_LOCK(..., 0)` and fast-fails so the queue can retry; every other `sharedDBHandle` caller loops on 1-second `GET_LOCK` calls until acquisition or its own `ctx` ends. So a foreground cold-open now surfaces its own deadline rather than an opaque `ErrSharedDBSchemaEnsureBusy`.

I checked the scoping, since making `EnsureSharedDBReady` non-waiting could have leaked fast-fail into user paths: it has exactly one non-test caller (`finishManagedSharedDBProvisioning`, `shared_db_pool.go:1395`), so the non-waiting behavior is confined to the provisioning worker. Correct.

Swallowing `ErrSharedDBSchemaEnsureBusy` in `managedSharedDBProvisioningQueueError` is also sound, and I confirmed the retry actually exists rather than the error being silently dropped: `resumeProvisioningManagedSharedDBPoolsWithCtx` pages `provisioning` rows and is started as a durable loop (`server.go:1038`). So a contended wave is genuinely deferred, not lost.

### Finding 2 (heartbeat cancelling DDL) — fixed correctly

`fn(ctx)` now runs on the caller context; `heartbeatCtx` governs only the ticker goroutine. A failed refresh logs `managed_shared_db_pool_provisioning_heartbeat_failed` and the next tick retries, with a `heartbeatCtx.Err()` check to stay quiet during normal shutdown. That inverts the failure mode back to advisory, which was the point.

The other four items from the parallel CodeRabbit review also check out — exported sentinel, `sync.Once` on the test helper signal, release-and-drain cleanup for the schema hook (the `-race` run above covers it), and deferring root-password decryption into the physical-create branch so complete-metadata provisioning decrypts exactly once.

### Residual note (non-blocking): `openMu` is held across the new foreground wait

`sharedDBHandleWithSchemaLockWait` takes `openMu` (a plain `sync.Mutex`, `pool.go:1141`) and holds it through the advisory-lock wait. So while one goroutine waits for another pod's DDL, other goroutines in the same process that need the same `dbID` block on `openMu` — and `sync.Mutex.Lock()` is not ctx-aware, so those waiters cannot honor their own deadlines even though the advisory-lock loop now carefully does.

To be fair to the change: `openMu` was already held across `ensureSharedDBSchema` before this PR, so a long hold isn't new. What's new is that the hold can now be spent *waiting on another process* rather than doing the work, and the per-request deadline respect added at the lock level doesn't extend to goroutines queued behind the mutex. Net effect is narrower than the original bug and only affects concurrent cold-opens of the same physical DB during a schema migration, which is why I'd treat it as a note rather than a blocker. If you want the deadline guarantee to hold end-to-end, the queued-waiter path needs a ctx-aware acquire (`TryLock` in a select loop, or a channel-based semaphore per `dbID`).

Also worth noting for the record: the 1-second `GET_LOCK` poll means a foreground open can overshoot its deadline by up to ~1s, since cancellation is only checked between attempts. Fine for a minutes-long DDL window; mentioning it so the bound is explicit.

### Verdict

LGTM. Both issues I raised are genuinely fixed, the fixes are scoped correctly rather than broadly, and the tests cover the specific properties (provisioning fast-fails, foreground respects deadline, foreground resumes after the version persists, heartbeat failure leaves work running). The design has ended up in a good place: the lock is on the resource being mutated, no MetaDB connection is retained across DDL, independent physical DBs stay concurrent, and the session-ownership caveat is documented rather than glossed over.

srstack:
Fixed a production-visible shared refill fragmentation bug in `4dfa2ed9`.

Root cause: `stageSharedPoolTenantWave` unconditionally created dedicated physical capacity for every refill wave. With a 5-second leader scan, a logical pool consuming 1-4 entries per round therefore created one new physical DB per round, leaving `max_tenants=100` pools with tenant counts such as `9,4,1,4,1,...`.

The durable wave transaction now:

- locks existing nonterminal shared DB capacity for the organization in `active -> provisioning -> pending` order;
- fills the oldest physical DB up to `max_tenants` before considering the next one;
- batches each physical DB capacity update into one `tenant_count += assigned` write in the same transaction as tenants, placements, memberships, quotas, and embedding profiles;
- inserts new `db_pool` rows only for the remaining tenants after every existing candidate is full;
- marks results as reused vs newly created, so only newly inserted physical DB IDs enter the Cloud-create batch;
- preserves logical-pool resize atomicity and per-status tenant state (`active`, `provisioning`, or `pending`).

Regression coverage reproduces the observed shape: `tenant_count=9/max_tenants=100` plus one refill remains one physical DB at count 10 with zero Cloud calls. A second boundary case starts at 99/100 and refills three tenants: one uses the existing DB and only the two-tenant remainder creates one new physical DB. The previous test that explicitly required dedicated refill capacity was corrected to assert reuse.

Focused shared refill/wave tests and `go test ./pkg/meta -count=1` pass locally. Hosted CI is running for the new head.

srstack:
## Re-review at `4dfa2ed9` — new commit is a correct fix, but it reverses a documented design decision and is out of scope for this PR

Verified on this head:

- `go test ./pkg/meta -count=1` — ok, 23.7s
- `go test ./pkg/server -count=1` — ok, 169.3s

Everything from my previous passes remains resolved. This round is entirely about the newly added `4dfa2ed9`.

### The fragmentation bug is real and the fix looks correct

The root cause is convincing: `stageSharedPoolTenantWave` created dedicated physical capacity per wave, so with a 5s leader scan consuming 1–4 entries per round you get one new physical DB per round and the observed `9,4,1,4,1,...` distribution against `max_tenants=100`. That's a genuine production problem worth fixing.

The implementation is sound in the ways that matter:

- `lockManagedSharedDBPoolWaveCandidates` takes `FOR UPDATE` on candidate rows inside the same transaction as the tenant/placement/membership staging, so capacity selection and consumption are atomic — no torn snapshot between pods.
- The `tenant_count = tenant_count + ?` change with the `AND tenant_count + ? <= max_tenants` guard is the right shape: relative increment plus a conditional that makes overfill impossible even under concurrent waves, rather than the old absolute `tenant_count = ?` write.
- Status-derived tenant status (`active`→`TenantActive`, `provisioning`→`TenantProvisioning`, `pending`→`TenantPending`) correctly avoids marking a tenant active on a physical DB that isn't ready.
- Only `Created` rows enter the Cloud-create batch (`createIDs`), while all rows still get continuations (`continuationIDs`). That distinction is easy to get backwards and it's right here.
- `RecordSharedDBPoolWave` now reports `createdPhysical` instead of `len(created)`, so the metric keeps meaning "physical DBs created" rather than silently becoming "assignments made."

Test coverage is good and specifically targets the reported shape: 9/100 + 1 refill stays one physical DB at count 10 with zero Cloud calls, and the 99/100 + 3 boundary case spills exactly one new DB for the 2-tenant remainder.

### Concern: this reverses an explicitly-reasoned prior decision

The comment being deleted wasn't incidental — it documented a deliberate tradeoff:

> Refill deliberately stages dedicated physical capacity instead of reusing a globally visible oldest candidate. Otherwise multiple Pods can plan against the same capacity snapshot and recreate the hot-row storm this partitioned wave is meant to avoid.

The new code does exactly what that comment warned against — reuse the oldest candidate — and the new comment asserts the hot-row problem is avoided by batching. I think the new reasoning is probably right: the `FOR UPDATE` + single batched `tenant_count += assigned` per DB per wave is materially different from per-tenant hot-row writes, which is what the original was guarding against.

But "probably right" is doing a lot of work for a change that re-enables a pattern someone previously removed on purpose, and neither the commit nor the tests demonstrate the hot-row property under concurrency. Every new test is single-threaded. The specific thing worth proving is: **N pods staging waves against the same organization simultaneously** — that they serialize on the `FOR UPDATE` without deadlocking, and that throughput doesn't collapse. `withMetaLockConflictRetry` gives 4 attempts with 20/40/80ms backoff, which bounds the damage, but under a leader scan every 5s across many pods that's the regime this will actually run in.

One concrete mechanism worth checking: candidates are locked `ORDER BY CASE status WHEN 'active' THEN 0 WHEN 'provisioning' THEN 1 ELSE 2 END, db_id`. That ordering isn't stable — a row's status advances `pending → provisioning → active` concurrently with waves. Two overlapping waves can therefore acquire the same two rows in *opposite* order if a status transition commits between their SELECTs, which is a classic deadlock shape. Ordering purely by `db_id` (and applying the status preference in Go, after locking) would make acquisition order total and stable. The retry wrapper probably absorbs it either way, but ordering by a mutable column under `FOR UPDATE` is worth avoiding on principle.

### Scope

This is a distinct bug from what PR #808 set out to do. The PR is titled "decouple shared schema provisioning concurrency" and its stated purpose is removing MetaDB session-lock pressure from schema DDL; `4dfa2ed9` is a capacity-allocation fragmentation fix in `pkg/meta/shared_pool_wave.go`, sharing no code with the rest of the change.

It's now +978/−122 across 10 files spanning target-DB advisory locking, heartbeats, worker scoping, and physical capacity allocation. Each piece has been individually well-executed, but bundled they mean a revert of the fragmentation fix takes the concurrency work with it, and vice versa — and these are exactly the changes where you'd want independent rollback, since both touch provisioning behavior that only misbehaves under production load. I'd split `4dfa2ed9` into its own PR: it stands alone, it has its own tests, and it would get a focused review rather than being the fifth topic in a thread.

If you'd rather keep it bundled to ship the production fix sooner, that's a defensible call — but please note in the PR description that the scope expanded beyond the title, so the reviewer of record isn't surprised.

### Verdict

The fix itself is correct and I'd approve it. Two asks, neither blocking on correctness grounds: order the `FOR UPDATE` by `db_id` only so lock acquisition order can't invert on a status transition, and add one concurrency test proving the reuse path doesn't deadlock or serialize badly with multiple simultaneous waves — that's the property the deleted comment was protecting, and it's the only part of this change still resting on argument rather than evidence.

pkg/meta/shared_pool_wave.go
```diff
@@ -375,8 +469,12 @@ func finalizeManagedSharedDBPoolWaveMembers(ctx context.Context, tx *sql.Tx, sta
...
 		return fmt.Errorf("insert wave memberships: %w", err)
 	}
-	res, err := tx.ExecContext(ctx, `UPDATE db_pool SET tenant_count = ?, soft_cap_reached = CASE WHEN ? >= max_tenants THEN 1 ELSE 0 END
-		WHERE db_id = ? AND status = ?`, len(staged.tenantIDs), len(staged.tenantIDs), staged.dbID, SharedDBStatusPending)
+	res, err := tx.ExecContext(ctx, `UPDATE db_pool
+		SET tenant_count = tenant_count + ?,
+			soft_cap_reached = CASE WHEN tenant_count + ? >= max_tenants THEN 1 ELSE 0 END
+		WHERE db_id = ? AND status = ? AND max_tenants > 0 AND soft_cap_reached = 0
+			AND tenant_count + ? <= max_tenants`,
+		len(staged.tenantIDs), len(staged.tenantIDs), staged.dbID, staged.dbStatus, len(staged.tenantIDs))
```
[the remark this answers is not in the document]
srstack: Fixed in 9eb20a70. MySQL evaluates single-table UPDATE assignments from left to right, so the reservation now computes soft_cap_reached before incrementing tenant_count. The regression covers tenant_count 93 plus a 5-tenant wave and verifies the row remains at 98 with soft_cap_reached=false and no Cloud create.

srstack:
Addressed the two requested correctness/concurrency checks in `9eb20a70`:

1. **Stable physical-row lock order.** The wave transaction now acquires every reusable physical row for the organization in immutable `db_id` order. Only after the complete lock set is held does Go apply the existing `active -> provisioning -> pending -> db_id` allocation preference. A concurrent status transition can therefore no longer invert row acquisition order, while active capacity remains preferred for tenant readiness. The lock is still scoped to the short durable wave transaction; it is not held across Cloud or schema work.

2. **Concurrent-wave regression.** The refill regression now launches two simultaneous three-tenant waves against the same logical pool after a capacity-boundary spill. Both waves converge on the same reusable physical DB, the physical DB count remains 2, the total reserved count becomes 108, and the Cloud-create call count remains 1. This covers the multi-wave property protected by the deleted dedicated-capacity comment.

The same commit also fixes the new CodeRabbit finding: MySQL evaluates single-table `UPDATE` assignments left to right, so `soft_cap_reached` is now calculated from the pre-increment count. The regression verifies `93 + 5 = 98` remains reusable instead of being prematurely soft-capped.

I kept the capacity fix in this PR because it is a production-visible failure in the same shared refill/provisioning lifecycle and was found while validating this head. It remains isolated in its own commits (`4dfa2ed9`, `9eb20a70`), so it can still be reverted or cherry-picked independently. The PR description already documents the expanded scope, and the title is now `fix: improve shared pool provisioning and capacity reuse` so the reviewer-of-record sees both changes.

Fresh local verification on `9eb20a70`:

- `go test ./pkg/server -run 'TestSharedTenantPoolRefill|TestAdminTenantPoolReplenish(SubmitsLargeRefillAsSingleWave|ContinuesPastWatermarkToFillSlots)' -count=1`
- `git diff --check`

srstack:
## Re-review at `f337b261` — both asks addressed correctly; one tradeoff introduced by the fix

Verified on this head:

- `go test ./pkg/meta -count=1` — ok, 23.5s
- `go test ./pkg/server -count=1` — ok, 168.3s

### Ask 1 (stable lock order) — fixed correctly

Locking now orders by `db_id` only, with the `active → provisioning → pending` readiness preference applied in Go via `sortManagedSharedDBPoolWaveCandidates` after the full lock set is held. That's exactly the right split: acquisition order is now on an immutable column so a concurrent status transition can't invert it between overlapping waves, while allocation still prefers ready capacity. The comment explains why, which matters for a constraint that isn't locally obvious.

### Ask 2 (concurrent-wave regression) — genuinely covers the property

The new test launches two simultaneous three-tenant waves with a `WaitGroup` after a capacity-boundary spill, then asserts `physicalPools == 2`, `tenantCount == 108`, and no extra Cloud-create call. That's the property the deleted dedicated-capacity comment was protecting — concurrent waves converging on shared capacity without over-creating — so it's no longer resting on argument alone.

### The `soft_cap_reached` ordering fix is a good catch

```sql
SET soft_cap_reached = CASE WHEN tenant_count + ? >= max_tenants THEN 1 ELSE 0 END,
    tenant_count = tenant_count + ?
```

MySQL evaluates single-table `UPDATE` assignments left to right, so with `tenant_count` assigned first the `CASE` would have read the *already-incremented* value and double-counted the delta — soft-capping a pool that still had room (the `93 + 5 = 98` case in the regression). Subtle and easy to miss; ordering the assignment so `soft_cap_reached` reads the pre-increment value is the correct fix, and the regression pins it. Worth a brief inline note that the assignment order is load-bearing, since it looks like cosmetic reordering and a future formatter or "tidy up the SET clause" edit would silently reintroduce the bug.

### New tradeoff: the lock scope widened from bounded to org-wide

Removing `LIMIT ?` means the wave now takes `FOR UPDATE` on **every** reusable physical row for the organization, not just as many as the wave could consume:

```sql
WHERE org_id = ? AND role = ? AND status IN (?, ?, ?)
    AND max_tenants > 0 AND soft_cap_reached = 0 AND tenant_count < max_tenants
ORDER BY db_id
FOR UPDATE
```

This is a real consequence of fixing ask 1 — you can't get a stable total order while also limiting to N rows chosen by a mutable preference — so I think it's the right call. But it should be a conscious one: for an organization with many non-full physical DBs, a single wave now serializes against every other wave for that org, where `LIMIT` previously let two waves needing different rows proceed in parallel. Combined with a 5s leader scan across pods, the steady state is that same-org waves fully serialize on this transaction.

That's probably fine — the transaction is short and holds no Cloud or schema work, and `withMetaLockConflictRetry` gives 4 attempts with backoff. The failure mode if it isn't fine is refill throughput dropping under many-physical-DB orgs, which is the opposite of what this PR is optimizing for, so it's worth knowing it's the tradeoff you took. Two options if it bites: bound the lock set by `db_id` range rather than by preference (keeping order stable), or keep the full lock but assert in a test that N concurrent waves complete within a sane bound rather than just correctly.

Also minor: `sortManagedSharedDBPoolWaveCandidates`'s `statusRank` returns `3` for unknown statuses, so such a row sorts last and is then rejected by `tenantStatusForSharedDBPoolWave` with an error that aborts the whole wave. Since the SQL `IN (?, ?, ?)` filter already excludes anything else, that branch is unreachable — fine as defence in depth, just noting an unknown status would fail the wave rather than being skipped, which is the safer of the two behaviors anyway.

### On scope

Noted that the title is now `fix: improve shared pool provisioning and capacity reuse` and the description documents the expansion, with the capacity work isolated in `4dfa2ed9` + `9eb20a70` for independent revert. That addresses my concern — the reviewer of record can now see both changes, and the commit boundaries make cherry-picking practical. Keeping it here given it's the same shared refill lifecycle and was found validating this head is a reasonable call.

### Verdict

LGTM. Both requested changes are correct, the MySQL assignment-order bug found alongside them is a real one, and the concurrency property now has a test. My only remaining suggestions are documentation-weight: an inline note that the `SET` clause order is semantically required, and awareness that the org-wide lock scope is the deliberate price of stable acquisition order.

srstack:
Addressed the two documentation notes in `131c8ad2` without changing runtime behavior:

- The candidate-lock comment now states explicitly that taking the complete stable `db_id`-ordered lock set deliberately serializes same-organization refill waves, and that the scope is limited to the short durable transaction with no Cloud or schema work under the lock.
- The capacity reservation now has an inline note that MySQL evaluates single-table `SET` assignments left to right, so `soft_cap_reached` must remain before the `tenant_count` increment to use the pre-increment value.

Focused verification:

- `go test ./pkg/meta -run '^TestManagedSharedDBPoolWaveCandidateLockOrderIsStable$' -count=1`
- `git diff --check`

pkg/tenant/pool.go
```diff
@@ -211,6 +217,62 @@ func ensureTiDBSchemaForEmbeddingMode(ctx context.Context, db *sql.DB, mode stri
...
+		if err := conn.QueryRowContext(ctx, "SELECT GET_LOCK(?, ?)", lockName, waitSeconds).Scan(&acquired); err != nil {
+			return fmt.Errorf("acquire shared schema advisory lock for db pool %d: %w", dbID, err)
+		}
+		if !acquired.Valid {
+			return fmt.Errorf("acquire shared schema advisory lock for db pool %d returned NULL", dbID)
+		}
+		if acquired.Int64 == 1 {
+			break
+		}
+		if !waitForLock {
```
mornyx, line 256: Minor accuracy note: on TiDB the minimum `GET_LOCK` timeout is 1 second and out-of-range values are clamped to the nearest permitted value, so `GET_LOCK(name, 0)` on a contended lock actually blocks up to ~1s before returning 0 (MySQL returns immediately). Behaviorally this is harmless — arguably even useful, since it smooths retry churn across pods — but the "non-blocking" framing in the PR description and the provisioning path's assumptions are only exactly true on MySQL (local e2e). A short code comment here noting the TiDB clamp would prevent someone from later reasoning about this path as strictly non-blocking.

pkg/server/shared_db_pool.go
```diff
@@ -1022,6 +1041,14 @@ func (s *Server) continueManagedSharedDBPoolOnce(ctx context.Context, dbID int64
 		if err != nil {
 			return err
 		}
+		// Provisioning rows have already durably committed their physical
+		// identity and connection metadata. Schema ensure is serialized across
+		// pods by an advisory lock on the target TiDB, so leader recovery does
+		// not need to retain a MetaDB session across remote DDL. Pending rows
+		// still use the work lock below to prevent duplicate physical creation.
+		if poolInfo.Status == meta.SharedDBStatusProvisioning && managedSharedDBConnectionMetadataComplete(poolInfo) {
```
mornyx, line 1049: Question: this early finish branch proceeds with the `poolInfo` snapshot read before any lock, whereas the work-lock path below re-reads the row inside the lock and restarts when `TiDBCloudOrganizationID` changed (`restartWithOrganization`). If a pool's org/identity were reassigned while a finisher is in flight, the lock-free path would run `EnsureDatabase`/quota PATCH against the stale row — `UpdateQuota` in particular would target the stale organization. I agree the dangerous parts are safe (schema ensure re-checks the version inside the target-TiDB advisory lock; `ActivateSharedDBPool` is a guarded UPDATE), and org reassignment mid-provisioning may simply never happen today. Worth either a comment stating that invariant, or re-reading the row once inside the advisory lock for symmetry with the work-lock path. Non-blocking.

pkg/meta/db_pool.go
```diff
@@ -494,6 +494,39 @@ func (s *Store) UpdateSharedDBSchemaVersion(ctx context.Context, dbID int64, ver
...
+	res, err := s.db.ExecContext(ctx, `UPDATE db_pool SET updated_at = CURRENT_TIMESTAMP(3)
+		WHERE db_id = ? AND role = ? AND status = ?`, dbID, SharedDBRoleShared, SharedDBStatusProvisioning)
+	if err != nil {
+		return fmt.Errorf("refresh provisioning progress for db pool %d: %w", dbID, err)
+	}
+	affected, err := res.RowsAffected()
+	if err != nil {
+		return fmt.Errorf("refresh provisioning progress rows affected for db pool %d: %w", dbID, err)
+	}
+	if affected == 0 {
```
mornyx, line 529: Nit: when `affected == 0` because the row exists but has left `provisioning` (e.g., already activated or failed by a concurrent finisher), this returns `ErrNotFound`, which conflates "row gone" with "row no longer in provisioning". The only caller (the heartbeat) just logs a warning, so this is cosmetic today — but a distinct sentinel (or returning the actual status) would keep this honest if a future caller branches on `ErrNotFound` semantics.

mornyx (commented):
Thorough review pass — design holds up end to end. Verified locally against the PR head:

- The provisioning queue now runs only schema-completion work; pending create/adopt stays in separately bounded workers, and `ensureManagedSharedDBPhysical` (the direct-request path) only does create/adopt under the MetaDB work lock, never DDL. The `maxOpen/2` cap removal is therefore consistent.
- Cross-pod schema serialization via target-TiDB `GET_LOCK` is sound: TiDB user-level locks take effect on all TiDB servers in the cluster (per PingCAP docs), unlike MySQL where they are server-local — so this works even with multiple TiDB nodes behind the serverless gateway.
- The fill-before-create wave is transactionally tight: candidates locked in `db_id` order, batched `tenant_count` increments with the left-to-right `soft_cap_reached` evaluation, `affected != 1` defensive failure, and `withMetaLockConflictRetry` around the whole wave. Tenant status lifecycle for reused pools is intact (`UpdateManagedSharedDBPoolCloudResult` advances pending→provisioning; `ActivateSharedTenantsBatch` flips provisioning→active; the stuck watchdog covers both).
- `managedSharedDBProvisioningJobs` dedups same-pod concurrent finishers; `ActivateSharedDBPool`'s guarded UPDATE makes cross-pod double-activation a benign no-op; busy-skipped jobs retry via the 15s durable resume scan. Crash mid-DDL self-heals because version persistence happens inside the advisory-lock critical section.

Tests: `TestSharedDBSchemaEnsureUsesCrossPoolAdvisoryLock`, `TestSharedDBColdOpensDoNotBlockDifferentDBPools`, `TestSharedTenantPoolRefillReusesExistingPhysicalCapacity`, `TestSharedTenantPoolRefillReusesCapacityWithoutIdentityLookup`, the heartbeat/lock-free provisioning tests, and the meta wave test all pass locally; CI is green.

Three non-blocking notes inline. LGTM.

pkg/tenant/pool.go
srstack: Documented in c765432e. The code comment now records that MySQL honors timeout 0 while TiDB clamps it to approximately one second, so this path is fail-fast rather than strictly non-blocking. The PR description has been updated with the same distinction.

pkg/server/shared_db_pool.go
srstack: Fixed in c765432e by enforcing the invariant at the Meta transaction boundary instead of adding another racy read. Once a pool reaches provisioning, UpdateManagedSharedDBPoolCloudResult rejects org_id or cluster_id reassignment with ErrSharedDBPoolIdentityConflict; endpoint and credential metadata remain refreshable. Regression coverage verifies both identity mutations are rejected and same-identity connection refresh still succeeds.

pkg/meta/db_pool.go
srstack: Fixed in c765432e. An existing row that has left provisioning now returns ErrSharedDBPoolNotProvisioning with its actual status, while a missing row still returns ErrNotFound. The heartbeat treats the status transition as expected completion of heartbeat activity and stops without emitting the misleading failure warning; both Meta and server behavior are covered.

srstack:
## Re-review at `c765432e` — docs notes done; the undescribed hardening commit has one reachable gap

Verified on this head:

- `go test ./pkg/meta -count=1` — ok, 23.6s
- `go test ./pkg/server -count=1` — ok, 175.1s
- `go test ./pkg/tenant -count=1` — ok, 21.4s

### `131c8ad2` — both documentation notes addressed

The candidate-lock comment now states that the full `db_id`-ordered lock set deliberately serializes same-org refill waves and that the scope is the short durable transaction with no Cloud or schema work under it. The `SET` clause now carries the note that MySQL evaluates single-table assignments left to right, so `soft_cap_reached` must precede the `tenant_count` increment. Both are exactly the invariants that would otherwise be lost to a future "tidy up" edit.

### `c765432e` — not covered by the comment above, and it's the substantive one

This commit wasn't described in the reply, so flagging what it does for the record: it adds `ErrSharedDBPoolIdentityConflict` and `ErrSharedDBPoolNotProvisioning`, rejects org/cluster identity changes in `UpdateManagedSharedDBPoolCloudResult` once a row is `provisioning`, stops the heartbeat quietly when status has advanced, and documents that TiDB clamps `GET_LOCK` to a 1s minimum.

The identity check is the right idea and closes something I raised in my first review — the lock-free finisher path assumes the Cloud allocation identity can't be reassigned underneath it, and nothing was previously enforcing that. Making it a durable invariant under the existing `FOR UPDATE` is the correct place for it. I also checked whether it could reject legitimate refreshes on the metadata-poll path (`shared_db_pool.go:906`): the loader matches by `infosByClusterID[clusterID]` whenever the row has a `ClusterID`, so the returned `info.ClusterID` is necessarily the same value, and org is passed through from the row. That path is safe.

The TiDB `GET_LOCK` clamp note is a useful correction to my earlier read — I described the queue path as fail-fast, and it can actually wait ~1s per attempt. Worth having written down.

### Reachable gap: a `provisioning` row with an empty `cluster_id` can never be repaired

The status gate that advances `pending → provisioning` does not include `cluster_id`:

```go
// db_pool.go:372
metadataReady := host != "" && port > 0 && user != "" && len(password) > 0 && name != ""
```

But the finisher's readiness predicate does require it:

```go
// shared_db_pool.go:1076
return poolInfo != nil && strings.TrimSpace(poolInfo.ClusterID) != "" && ...
```

So a row can legitimately reach `provisioning` with `cluster_id` still empty — connection metadata complete, identity not. Once there, the new check makes that state terminal:

```go
if currentStatus == SharedDBStatusProvisioning &&
    (currentOrganizationID != in.TiDBCloudOrganizationID || currentClusterID.String != in.ClusterID) {
    return fmt.Errorf("%w: db pool %d", ErrSharedDBPoolIdentityConflict, in.ID)
}
```

A later call carrying the real `ClusterID` compares `"" != "cluster-abc"` and is rejected as an identity conflict. The row can't be repaired, `managedSharedDBConnectionMetadataComplete` keeps failing it, and it sits until the stuck watchdog marks it failed — discarding an already-created cluster, which is the outcome `d2e79255` widened the timeout to avoid.

The check wants to say "identity may not be *reassigned*," but as written it also forbids "identity may be *filled in*." Narrowing it to only reject when the stored value is non-empty fixes that while keeping the invariant the finisher depends on:

```go
if currentStatus == SharedDBStatusProvisioning &&
    ((currentOrganizationID != "" && currentOrganizationID != in.TiDBCloudOrganizationID) ||
     (currentClusterID.String != "" && currentClusterID.String != in.ClusterID)) {
```

Better still, add `cluster_id` to `metadataReady` so the two predicates agree and the state is unreachable in the first place — that also removes a real inconsistency where `pending → provisioning` and "finisher may run" disagree about what complete means. I'd do both: the gate makes it unreachable, the narrowed check makes it recoverable if it ever occurs.

Related asymmetry in the same UPDATE: host/port/user/password/name all use `COALESCE(?, col)` so an empty input preserves the stored value, while `org_id` and `cluster_id` are written unconditionally. For a `pending` row the new check doesn't fire, so an empty incoming `ClusterID` still silently blanks a previously-stored one. Pre-existing, but the new check makes the inconsistency load-bearing — the same field is immutable in one status and clobberable in another.

### Verdict

`131c8ad2` is complete. The identity invariant in `c765432e` is a genuine improvement and I'd keep it, but it needs the empty-value narrowing before merge — as written it converts a recoverable incomplete-identity row into a guaranteed watchdog failure with a discarded cluster. Adding `cluster_id` to `metadataReady` is the cleaner half of the fix. Everything from my earlier passes remains resolved.

pkg/meta/db_pool_test.go
```diff
@@ -674,6 +674,102 @@ func TestManagedSharedDBUpdatedAtTracksProgressNotNoOpPolls(t *testing.T) {
...
+	if err := s.UpdateManagedSharedDBPoolCloudResult(ctx, &refreshed); err != nil {
+		t.Fatalf("refresh connection metadata with stable identity: %v", err)
+	}
+	got, err := s.GetSharedDB(ctx, dbID)
+	if err != nil {
+		t.Fatalf("GetSharedDB after connection refresh: %v", err)
+	}
+	if got.TiDBCloudOrganizationID != ready.TiDBCloudOrganizationID || got.ClusterID != ready.ClusterID ||
+		got.Host != refreshed.Host || got.User != refreshed.User || !bytes.Equal(got.PasswordCipher, refreshed.PasswordCipher) {
+		t.Fatalf("connection refresh = %+v, want stable identity with refreshed connection metadata", got)
```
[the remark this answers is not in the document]
srstack: Fixed in fdc7d61c. The stable-identity connection refresh regression now asserts the persisted TLS mode alongside host, user, and encrypted password.

srstack:
Addressed in da55f0dc. The pending-to-provisioning readiness gate now explicitly includes a non-blank cluster identity, and cluster IDs with surrounding or all whitespace are rejected before persistence. The provisioning identity guard now rejects reassignment only when the stored organization or cluster identity is already non-empty, so a malformed legacy provisioning row with a missing cluster can be repaired while an established identity remains immutable. Regression coverage first reproduced both failures: blank cluster metadata advancing incorrectly, and a legacy missing cluster being rejected during repair. Full pkg/meta passes. One detail from the report is narrower in current code: an exactly empty incoming ClusterID was already rejected at function entry, so the clobberable input case was whitespace rather than the empty string. The PR summary and lifecycle explanation now document the enforced identity and recovery boundary.

mornyx (approved):
Re-reviewed the incremental commits since my first pass (`131c8ad`..`da55f0d`). All three of my earlier comments are addressed properly:

- The TiDB `GET_LOCK` zero-timeout clamp is now documented at the acquisition site.
- Identity stability is enforced at the right place: `UpdateManagedSharedDBPoolCloudResult` rejects org/cluster reassignment once provisioning (with a `FOR UPDATE` read, so the check and the write are atomic), while endpoint/credential rotation stays refreshable. The `da55f0d` recovery carve-out for a NULL legacy cluster identity is covered by `TestManagedSharedDBCloudResultRepairsMissingProvisioningIdentity`.
- `RefreshManagedSharedDBProvisioningProgress` now distinguishes terminal status (`ErrSharedDBPoolNotProvisioning`) from a missing row, and the heartbeat stops quietly on status advance — safe because active pools are outside the stuck watchdog's scope.

Verified locally: the four new meta tests and `TestManagedSharedDBProvisioningHeartbeatDoesNotWarnAfterStatusAdvance` pass; CI green. LGTM.

PR #809, opened 2026-07-27
fix: reclaim idle shared tenant cache groups

srstack:
## Summary

- keep only two tenant-cache access paths: `Acquire` for all real activity and `AcquireCached` for the warm-only safety-net scan
- treat request traffic, tenant workers, object GC, warm local S3 access, purge, and schema maintenance as real activity that promotes the tenant LRU and refreshes physical shared-DB activity
- reclaim an idle physical shared-DB group after 10 minutes of no real activity; with the existing 2-minute reaper interval, cleanup occurs after approximately 10-12 minutes
- reserve the physical shared handle before tenant cold-open and transfer that generation-specific lease to the cached tenant entry, preventing the reaper from closing the handle in flight and preventing stale releases from corrupting reopened-handle refs
- pin local S3 requests for their complete handler lifetime
- keep the unsigned local mock-S3 gateway warm-only so a URL tenant ID cannot cold-open a tenant database
- release the shared-handle mutex before scanning the capacity LRU and batch removal metrics outside the request-path lock
- preserve the target-TiDB advisory schema lock behavior merged in #808

`AcquireCached` never cold-opens, promotes the capacity LRU, or refreshes tenant/physical idle activity. There is no separate foreground/work state machine: background work is demand-driven real use, while the periodic safety-net observation remains explicitly warm-only.

Local mock-S3 requests also cannot cold-open because those development presigned URLs do not carry independent tenant authentication. An already-warm hit is promoted and refreshed as real activity, then pinned until the HTTP handler completes.

No schema, environment variable, durable state, per-tenant TTL, or new goroutine is added. The shared cache capacity default remains 20,480. Obsolete unpinned `Get`/`S3Backend` accessors were removed so new callers cannot bypass the two-path ownership model.

## Safety

- a physical handle lease increments the same `sharedDBRefs` counter used by cached tenant ownership before the handle leaves `sharedMu`
- lease release verifies both `db_pool.id` and the exact `*sql.DB` generation, so a delayed release after invalidation cannot decrement a reopened handle
- cold-open failures and duplicate cold-open losers release their lease; successful cold opens transfer it to the cache entry until entry close
- active tenant entries can be retired from the LRU but remain pinned and usable until their request/work references drain
- tenant stores and physical SQL handles close outside pool locks
- cascade expiration is revalidated under `p.mu -> p.sharedMu`, then the potentially 20,480-entry LRU scan runs with only `p.mu`

## Verification

- `GOCACHE=/tmp/drive9-go-cache go test ./pkg/server -run 'Test.*(TenantWorker|ObjectGC|Auth|S3|LeaderGatedWorkers)' -count=1`
- `GOCACHE=/tmp/drive9-go-cache go test -race ./pkg/tenant -run 'TestSharedDBHandleLease|TestSharedDBIdleReap|TestAcquireCached|TestAcquireCounts|TestAcquireRefreshes|TestLoadS3Backend' -count=1`
- `git diff --check`

### DIFF
--- a/pkg/tenant/pool.go
+++ b/pkg/tenant/pool.go
@@ -68,15 +68,14 @@ type PoolConfig struct {
 
 	// IdleTimeout controls how long a standalone cached backend can stay in the
 	// warm cache without activity before the idle reaper evicts it. Shared
-	// tenant entries intentionally bypass this TTL and are reclaimed by the
-	// capacity LRU instead. 0 disables standalone idle eviction; the shared
-	// physical-DB handle reaper remains active.
+	// tenant entries intentionally bypass this per-tenant TTL; they are reclaimed
+	// by capacity pressure or together when their physical shared DB has had no
+	// foreground traffic for defaultSharedDBHandleIdleTTL. 0 disables standalone
+	// idle eviction; shared physical-DB group reaping remains active.
 	//
-	// "Activity" means any access through Get, Acquire, or S3Backend —
-	// including foreground user requests (HTTP/FUSE), tenant-specific
-	// durable work driven by kicks (semantic/file_gc task processing), and
-	// object-GC candidate processing. These are all legitimate uses that
-	// should keep a tenant warm.
+	// For standalone tenants, "activity" includes foreground requests and
+	// tenant-specific durable work. Background work refreshes this idle TTL but
+	// does not promote the entry in the capacity LRU.
 	//
 	// AcquireCached (safety-net scan) does NOT refresh the idle timer on
 	// acquire or release, so warm-only periodic observation cannot keep an
@@ -86,7 +85,7 @@ type PoolConfig struct {
 	IdleTimeout time.Duration
 
 	// IdleReapInterval is how often the reaper scans standalone entries and
-	// unreferenced shared physical-DB handles. It defaults to
+	// foreground-idle shared physical-DB groups. It defaults to
 	// defaultTenantPoolIdleReapInterval when not set.
 	IdleReapInterval time.Duration
 
@@ -162,13 +161,13 @@ type Pool struct {
 	// db_pool.id). All tenants placed on the same shared DB share its
 	// handle; each Acquire still gets its own Store (carrying that tenant's
 	// fs_id scope) over the shared handle.
-	sharedMu         sync.Mutex
-	sharedDBs        map[int64]*sql.DB
-	sharedDBRefs     map[int64]int
-	sharedDBLastUsed map[int64]time.Time
-	sharedDBLabels   map[int64]sharedDBMetricLabels
-	sharedDBOpenMu   map[int64]*sync.Mutex
-	sharedDBClosed   bool
+	sharedMu                   sync.Mutex
+	sharedDBs                  map[int64]*sql.DB
+	sharedDBRefs               map[int64]int
+	sharedDBLastForegroundUsed map[int64]time.Time
+	sharedDBLabels             map[int64]sharedDBMetricLabels
+	sharedDBOpenMu             map[int64]*sync.Mutex
+	sharedDBClosed             bool
 }
 
 // sharedDBMetricLabels remembers the metric identity of a cached shared DB
@@ -305,9 +304,16 @@ type entry struct {
 	elem               *list.Element
 	refs               int
 	retired            bool
-	lastUsed           time.Time // refreshed by Get/Acquire/S3Backend (foreground + durable work) and on Acquire release; never by AcquireCached
+	lastUsed           time.Time // refreshed by Get/foreground/work/S3Backend and foreground/work release; never by AcquireCached
 	sharedDBID         int64     // db_pool.id for shared-schema tenants; zero for standalone tenants
 }
+
+type poolAccessKind uint8
+
+const (
+	poolAccessForeground poolAccessKind = iota
+	poolAccessWork
+)
 
 func NewPool(cfg PoolConfig, enc encrypt.Encryptor) *Pool {
 	max := cfg.MaxTenants
@@ -322,16 +328,16 @@ func NewPool(cfg PoolConfig, enc encrypt.Encryptor) *Pool {
 	metrics.RecordGauge("tenant_pool", "cached_backends", 0)
 	metrics.RecordGauge("tenant_pool", "max_backends", float64(max))
 	return &Pool{
-		cfg:              cfg,
-		enc:              enc,
-		items:            map[string]*entry{},
-		order:            list.New(),
-		maxSize:          max,
-		sharedDBs:        map[int64]*sql.DB{},
-		sharedDBRefs:     map[int64]int{},
-		sharedDBLastUsed: map[int64]time.Time{},
-		sharedDBLabels:   map[int64]sharedDBMetricLabels{},
-		sharedDBOpenMu:   map[int64]*sync.Mutex{},
+		cfg:                        cfg,
+		enc:                        enc,
+		items:                      map[string]*entry{},
+		order:                      list.New(),
+		maxSize:                    max,
+		sharedDBs:                  map[int64]*sql.DB{},
+		sharedDBRefs:               map[int64]int{},
+		sharedDBLastForegroundUsed: map[int64]time.Time{},
+		sharedDBLabels:             map[int64]sharedDBMetricLabels{},
+		sharedDBOpenMu:             map[int64]*sync.Mutex{},
 		// No LeaderChecker means single-pod mode: FileGC runs unconditionally.
 		fileGCEnabled: cfg.LeaderChecker == nil,
 		idleTimeout:   idleTimeout,
@@ -448,10 +454,30 @@ func (p *Pool) Get(ctx context.Context, t *meta.Tenant) (out *backend.Dat9Backen
 	return b, nil
 }
 
-// Acquire returns a backend that is pinned for the caller's active use. The
-// returned release callback must be called when the caller is done with the
-// backend so a retired entry can be closed.
+// Acquire is the compatibility alias for AcquireForeground.
 func (p *Pool) Acquire(ctx context.Context, t *meta.Tenant) (out *backend.Dat9Backend, release func(), err error) {
+	return p.AcquireForeground(ctx, t)
+}
+
+// AcquireForeground returns a backend pinned for foreground tenant traffic.
+// Foreground access promotes the tenant in the capacity LRU and refreshes the
+// physical shared-DB foreground activity timestamp. The returned release
+// callback must be called when the caller is done.
+func (p *Pool) AcquireForeground(ctx context.Context, t *meta.Tenant) (out *backend.Dat9Backend, release func(), err error) {
+	return p.acquire(ctx, t, poolAccessForeground)
+}
+
+// AcquireForWork returns a backend pinned for background tenant work. Work may
+// cold-open a backend, but it does not promote the tenant in the capacity LRU
+// or refresh physical shared-DB foreground activity. A cold work entry is
+// inserted at the back, so a full cache retires that new entry instead of
+// displacing a foreground-hot tenant. The returned release callback must be
+// called when the caller is done.
+func (p *Pool) AcquireForWork(ctx context.Context, t *meta.Tenant) (out *backend.Dat9Backend, release func(), err error) {
+	return p.acquire(ctx, t, poolAccessWork)
+}
+
+func (p *Pool) acquire(ctx context.Context, t *meta.Tenant, access poolAccessKind) (out *backend.Dat9Backend, release func(), err error) {
 	start := time.Now()
 	defer observePool(ctx, "acquire", t.ID, &err, start)
 
@@ -470,14 +496,17 @@ func (p *Pool) Acquire(ctx context.Context, t *meta.Tenant) (out *backend.Dat9Ba
 	if e, ok := p.items[t.ID]; ok {
 		if e.s3EncryptionPolicy == s3EncryptionPolicy && (storageNamespaceID == "" || e.storageNamespaceID == storageNamespaceID) {
 			e.refs++
-			p.order.MoveToFront(e.elem)
+			if access == poolAccessForeground {
+				p.order.MoveToFront(e.elem)
+				p.touchSharedDBForegroundLocked(e.sharedDBID, time.Now())
+			}
 			e.lastUsed = time.Now()
 			p.mu.Unlock()
 			metrics.RecordOperation("tenant_pool", "cache_lookup", "hit", 0)
 			totalDuration := time.Since(start)
 			logger.InfoOpenPoolTiming(ctx, "tenant_pool_acquire_timing", totalDuration,
 				poolAcquireTimingFields(t.ID, false, 0, totalDuration)...)
-			return e.backend, p.makeRelease(e), nil
+			return e.backend, p.makeRelease(e, access), nil
 		}
 		if removed := p.removeLocked(e.elem, "replace"); removed != nil {
 			toClose = append(toClose, removed)
@@ -511,7 +540,10 @@ func (p *Pool) Acquire(ctx context.Context, t *meta.Tenant) (out *backend.Dat9Ba
 	if e, ok := p.items[t.ID]; ok {
 		if e.s3EncryptionPolicy == s3EncryptionPolicy && (t.StorageNamespaceID == "" || e.storageNamespaceID == t.StorageNamespaceID) {
 			e.refs++
-			p.order.MoveToFront(e.elem)
+			if access == poolAccessForeground {
+				p.order.MoveToFront(e.elem)
+				p.touchSharedDBForegroundLocked(e.sharedDBID, time.Now())
+			}
 			e.lastUsed = time.Now()
 			p.mu.Unlock()
 			b.Close()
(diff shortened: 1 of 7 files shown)


pkg/tenant/pool.go
```diff
@@ -455,16 +487,23 @@ func (p *Pool) Acquire(ctx context.Context, t *meta.Tenant) (out *backend.Dat9Ba
...
 		if removed := p.removeLocked(e.elem, "replace"); removed != nil {
 			toClose = append(toClose, removed)
 		}
 	}
 	e := &entry{tenantID: t.ID, storageNamespaceID: t.StorageNamespaceID, s3EncryptionPolicy: s3EncryptionPolicy, backend: b, store: st, refs: 1, lastUsed: time.Now(), sharedDBID: sharedDBID}
-	e.elem = p.order.PushFront(e)
+	if access == poolAccessForeground {
+		e.elem = p.order.PushFront(e)
+	} else {
+		e.elem = p.order.PushBack(e)
```
mornyx: Non-blocking observation: with a full cache, a work miss inserts at the back and is immediately retired by the eviction loop below — i.e., background work on a saturated cache becomes a single-use cold open (as `TestAcquireForWorkDoesNotDisplaceForegroundCapacityLRU` asserts). That is the right trade for protecting foreground-hot entries, but under a background-heavy fleet larger than `maxSize` (default 1024) the background hit ratio drops toward zero and every semantic/GC task pays a full cold open (tenant credential decrypt + DB dial + schema check). Worth watching the `tenant_pool` acquire miss rate and cold-open timing after rollout — `MaxTenants` may need to be revisited once this lands in the stress environment.

pkg/tenant/pool.go
```diff
@@ -1018,10 +1070,59 @@ func (p *Pool) deleteSharedDBCacheMetricsLocked(dbID int64) {
...
-// for the longer shared-handle TTL. The existing tenant-pool reaper invokes
-// this; no separate worker is needed.
+// reapExpiredSharedDBTenantEntries retires every cached tenant view belonging
+// to a physical shared DB whose foreground TTL expired. Active entries leave
+// the capacity LRU immediately but remain usable until their references drain.
+//
+// Expiration is revalidated while holding p.mu -> p.sharedMu so a foreground
+// acquire cannot race the cascade after refreshing the physical timestamp.
+// Stores are closed outside both locks.
+func (p *Pool) reapExpiredSharedDBTenantEntries(now time.Time) {
```
mornyx, line 1053: Non-blocking behavior note: since work access no longer extends `sharedDBLastForegroundUsed`, a physical shared DB with zero foreground traffic for 30 minutes gets its entire tenant-cache group cascade-retired and the handle closed — even if background work (kicks, object GC) touches it regularly. The first background task after each idle window pays a physical reopen plus per-tenant cold opens until the next foreground request arrives. This is exactly the reclamation the PR wants; flagging it so the deploy note can call out that `drive9_shared_db_pool_cache_handles` churn and reopen latency on work-only DBs are expected, not a regression.

mornyx (approved):
Careful review pass — the lifecycle split is correct and complete. Verified against the PR head:

- **Locking is sound.** The documented order `p.mu -> p.sharedMu` holds at every site (acquire hit/insert paths, `releaseEntry`, the cascade revalidation). `removeLocked` never touches `sharedMu`, so the cascade's double-lock section cannot deadlock; `closeEntry -> releaseSharedDB` runs after both locks are released. The two-phase candidate-then-revalidate design closes the race where a foreground acquire refreshes the physical timestamp mid-cascade — every `touchSharedDBForegroundLocked` call site holds `p.mu`, which the revalidation phase blocks on.
- **Caller classification is complete.** Grepped all non-test `Acquire` call sites: auth middleware (API-key + capability) uses `AcquireForeground`; `tenantWorkerManager.targetForRef` and the object-GC worker use `AcquireForWork`; the safety-net scan keeps `AcquireCached` (correctly refresh-free). `Get`/`S3Backend`/`LoadS3Backend` only serve the local mock-S3 endpoint (`cfg.S3Dir`), so no production foreground path bypasses the foreground timestamp.
- **Retired-but-active entries** stay usable until refs drain (covered by `TestSharedDBIdleReapDefersActiveEntryClose`), and the physical handle closes only after both the foreground TTL expiry and reference drain — the reap order in `reapOnce` (cascade, then `reapIdleSharedDBs`) lets both happen in one tick.
- **Standalone tenants are unaffected**: `sharedDBID == 0` skips the touch, the cascade only iterates real handle IDs, and work access still refreshes the standalone idle TTL.

Tests: all pool lifecycle tests pass locally, including a `-race` run over the cascade/foreground-activity/work-LRU tests; CI green.

Two non-blocking operational notes inline. LGTM.

srstack:
Reviewed the full change plus the downstream call graph (`pkg/server/auth.go`, `tenant_worker.go`, `object_gc_worker.go`, `safety_net_scan.go`, `LoadS3Backend`, `PurgeSharedTenant`/`EnsureSharedDBReady`). Build and `go vet` are clean, and the new tests pass. The foreground/work split is the right model, but dropping the timestamp refresh from the handle lookup opens a window that did not exist before this PR. Two correctness issues; both reproduced with throwaway probe tests against the real pool.

## 1. An expired handle can be closed while a cold open is in flight

`sharedDBHandle` no longer refreshes `sharedDBLastForegroundUsed` on a cache hit (`pkg/tenant/pool.go:1160`). That makes this interleaving reachable:

1. `createBackend` gets the handle from the cache hit — no timestamp refresh.
2. The reaper runs `reapIdleSharedDBs`: the foreground TTL is expired and `sharedDBRefs[dbID] == 0` (every other tenant on that physical DB has already been retired by the cascade), so it closes the handle and deletes it from `sharedDBs`.
3. The cold open continues and builds a `datastore.Store` over the now-closed handle.

Probe result: the handle is `sql: database is closed`, **and** `sharedDBIDForStore` returns 0, because it identifies the shared DB by reverse-looking-up the `*sql.DB` pointer in `p.sharedDBs` (`pkg/tenant/pool.go:994`). So the shared tenant gets cached as if it were standalone — it is invisible to the cascade, it becomes subject to the standalone idle TTL instead, and `closeEntry` never releases a shared ref for it. The request fails until the entry is replaced.

Before this PR the handle lookup refreshed the timestamp, so the reaper could not see an expired window while a cold open was resolving. That protection is what the change removes.

## 2. The same race leaks `sharedDBRefs` permanently, so the handle can never be reclaimed again

If `sharedDBIDForStore` observes the handle but the reaper deletes it before `retainSharedDB` runs (`pkg/tenant/pool.go:503`):

- `retainSharedDB` (`:1009`) increments unconditionally.
- `releaseSharedDB` (`:1019`) returns early when the dbID is absent from `sharedDBs`, so that `+1` is never given back.

Probe result: `refs == 1` after release with no cached handle. Once the handle is reopened, the `sharedDBRefs[dbID] > 0` guard in `reapIdleSharedDBs` is permanently true and **that db_pool's handle is never reclaimed again**. This directly defeats the goal of the PR, and it is monotonically worsening — every occurrence pins one more physical handle for the lifetime of the pod.

Suggested fix for both: have `sharedDBHandle` hand back an in-flight reference under `sharedMu` before returning (e.g. increment `sharedDBRefs[dbID]` on the way out, with the caller returning it on the failure path), so the reaper can see the pending cold open. The in-flight reference and `retainSharedDB` must share one counter, otherwise #2 survives.

## Non-blocking

- **Cascade lock-hold time.** `reapExpiredSharedDBTenantEntries` (`:1080`) walks the entire LRU while holding both `p.mu` and `p.sharedMu`. At the shared-deployment default of 20,480 entries a full cascade measured **20.6ms** (750ns when nothing is expired). `p.mu` is the request-path `AcquireForeground` lock, so that is a ~20ms stall once per reap interval. Acceptable, but each removal emits two metrics inside the lock (`removeLocked` -> `recordCachedBackendCountLocked` plus `RecordOperation`); batching those and emitting after the unlock removes most of it cheaply.
- **The `Acquire` alias has no production callers** — only tests. Keeping it means new code can silently pick up foreground semantics. Consider deleting it and making the tests say `AcquireForeground` explicitly, so the distinction stays enforced by the compiler.
- **`Get` is not part of the new model.** `Get` (`:314`) unconditionally does `MoveToFront` but never touches the shared foreground timestamp. Its only production caller is the S3 gateway via `LoadS3Backend` (`pkg/server/server.go:663`), which is foreground traffic. Net effect: S3 gateway requests promote the capacity LRU but do not stop the 30-minute cascade from retiring their own entry. Either add `touchSharedDBForegroundLocked` there, or document why the S3 gateway is deliberately not foreground. `S3Backend` (`:840`) has the same asymmetry — it refreshes `lastUsed` only.
- **PR description vs. behavior.** The description says a work-only cold entry at full capacity "retires itself instead of evicting foreground-hot tenants." The implementation is `PushBack` plus eviction from `Back()`, so at a full cache the new entry is removed within the same `acquire` call. That means `AcquireForWork` cold-opens on **every** kick for a full cache, even for a tenant kicked repeatedly. Probably fine at the shared default of 20,480, but on standalone (default 1,024) with high tenant counts this turns background work into sustained cold opens of the same tenant TiDBs — which is exactly the cost #660 removed. Worth a comment at minimum, or exempting the work miss from its own eviction pass.

## Merge-order note with #808

#808 touches the same lines and keeps the `sharedDBLastUsed` refresh inside `sharedDBHandleWithSchemaLockWait`, which this PR removes. GitHub still reports both as `MERGEABLE`. If #808 lands second, its restored refresh would incidentally mask #1 and #2 — coincidence, not design. Please resolve the two issues here on their own terms regardless of merge order.

pkg/tenant/pool.go
srstack: Resolved by simplifying the model in d805f449. AcquireForWork and the foreground/work split were removed. Demand-driven tenant work now uses the same Acquire path as request traffic, so a cold work entry is inserted at the front and participates normally in capacity LRU activity instead of immediately evicting itself. Only AcquireCached remains warm-only and non-promoting for the periodic safety-net scan.

srstack:
Addressed the full review in d805f449 after rebasing onto merged #808. The two reproduced correctness races are fixed with a generation-specific physical-handle lease acquired under sharedMu and transferred into cached-entry ownership; cold-open failures and duplicate losers release it, and stale releases cannot affect a reopened handle. The long cascade no longer holds sharedMu during the LRU scan, and removal metrics are emitted outside the request-path lock. I also removed the foreground/work split plus the obsolete unpinned Get and S3Backend paths: all demand-driven request, worker, object-GC, and local-S3 use now goes through Acquire, while only the periodic safety-net uses warm-only AcquireCached. Local S3 is pinned for the full handler lifetime. The physical idle TTL is a fixed 10 minutes, yielding roughly 10-12 minute cleanup with the existing reaper. Full pkg/tenant, affected pkg/server tests, focused race tests, and diff checks pass locally; new-head CI is running.

pkg/tenant/pool.go
```diff
@@ -914,28 +875,24 @@ func (p *Pool) AutoSemanticTaskTypes() []semantic.TaskType {
...
-	b, err = p.Get(ctx, tenant)
+	b, release, err := p.Acquire(ctx, tenant)
 	if err != nil {
 		logger.Error(ctx, "tenant_pool_get_failed", zap.String("tenant_id", tenantID), zap.Error(err))
-		return nil
+		return nil, nil
 	}
-	return b
+	return b, release
 }
```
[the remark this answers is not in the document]
srstack: Fixed in 3ff728a2 without adding a third public acquisition mode. LoadS3Backend now resolves tenant metadata but only pins an already-warm cache entry; a cold or stale URL returns no backend and cannot open the tenant database. Warm local-S3 hits still count as real activity, promote the capacity LRU, refresh idle state, and remain pinned for the full handler request. The regression test now proves both cold-skip and warm pin/activity behavior. I also applied the review style nit by converting the new post-operation assertions to Errorf while retaining Fatal for setup failures.

srstack:
### Re-review of the revised version (`3ff728a`) — LGTM ✅

Verified the two blocking correctness bugs from my earlier comment are fixed, and the two non-blocking notes were addressed.

**Both blocking bugs fixed** (re-confirmed with throwaway probe tests):

1. **Handle closed under an in-flight cold open** → the new `sharedDBHandleLease` pins the shared `*sql.DB` across the reaper, so a cold open can no longer be mis-cached as a standalone store over a closed handle. `Release()` is generation-safe: it checks `sharedDBs[dbID] == l.db` (exact pointer) before decrementing, so a stale lease can't decrement a re-created handle.

2. **Permanent `sharedDBRefs` leak** → gone. Refs are taken via `leaseSharedDBLocked` and always returned through `lease.Release()` (guarded by `sync.Once`), including `closeEntry` now calling `e.sharedDBLease.Release()` instead of the old `releaseSharedDB(id)`. Handles become reclaimable again.

**Non-blocking notes addressed:**
- Cascade lock: `reapExpiredSharedDBTenantEntries` now releases `sharedMu` before the LRU scan and coalesces metrics after unlocking `p.mu` — the combined-lock stall I measured is gone.
- Dead code removed (`Get`, `S3Backend`, the foreground/work `Acquire` split). Cleaner surface: single `Acquire` + warm-only `AcquireCached`.
- Idle TTL reduced 30m → 10m.

**One minor note (non-blocking):** `LoadS3Backend` is now warm-only and returns 404 for a cold tenant, while the client only re-presigns on 403 (404 is a hard error). This is the local mock-S3 gateway path only, and the PR body frames it as intentional (an unsigned URL tenant ID must not cold-open a tenant DB) — I agree with that tradeoff. The only observable effect is a cold tenant evicted mid-transfer getting a non-retryable 404 on the local path; with 10-min presign TTLs the window is small and it's dev/test-only. Fine as-is; flagging only for awareness.

Clean, no technical debt left behind. LGTM.

_(Note: GitHub blocks formally approving one's own PR, so this comment stands in for the approval.)_

pkg/tenant/pool.go
```diff
@@ -914,28 +897,27 @@ func (p *Pool) AutoSemanticTaskTypes() []semantic.TaskType {
 	return out
 }
 
-func (p *Pool) LoadS3Backend(ctx context.Context, metaStore *meta.Store, tenantID string) (out *backend.Dat9Backend) {
+// LoadS3Backend resolves tenant metadata and pins an already-warm backend for
```
mornyx, line 900: Worth calling out in the deploy note: `LoadS3Backend` is now warm-only, so a local mock-S3 presigned-URL fetch for a tenant whose entry has expired returns 404 instead of lazily cold-opening. That is strictly better (an unauthenticated URL should not trigger a cold open, and the pin-until-release semantics are now correct), but any local tooling that fetches URLs long after the upload request must expect to re-warm the tenant through the API first.

pkg/tenant/pool.go
```diff
@@ -194,7 +193,7 @@ var (
 	defaultTenantPoolDrainTimeout           = 30 * time.Second
 	defaultTenantPoolMaxTenants             = 1024
 	defaultTenantPoolIdleReapInterval       = 2 * time.Minute
-	defaultSharedDBHandleIdleTTL            = 30 * time.Minute
+	defaultSharedDBHandleIdleTTL            = 10 * time.Minute
```
mornyx, line 196: Non-blocking: with the TTL shortened 30m -> 10m and `sharedDBHandle` leasing itself counting as activity, a physical group with only sporadic traffic (less than one acquire/lease per ~10 min) will now be fully reclaimed each window and pay a cold reopen plus per-tenant cold opens on the first touch of the next window. Expected given the reclamation goal — just note that `drive9_shared_db_pool_cache_handles` will show this sawtooth for low-traffic orgs, which is distinguishable from a regression only if you know to look for it.

mornyx (approved):
Re-reviewed the new head after the redesign (`5609f37f9`..`3ff728a22`). The lease-based ownership model is a strict improvement over both the original and my previously approved revision:

- **Unified `Acquire` with all activity counting** removes the foreground/work split — my earlier note about background single-use churn on a saturated cache is now moot, and the entry-level `lastUsed` semantics are much easier to reason about.
- **The generation-safe `sharedDBHandleLease` fixes three real latent bugs**: a delayed release can no longer decrement refs of a reopened handle generation (the old `releaseSharedDB(dbID)` had no generation check); `PurgeSharedTenant` now pins the handle across its batched deletes instead of using an unpinned raw handle; and an in-flight cold open holds a lease, so the 10-minute reaper can't close a handle mid-open.
- **`LoadS3Backend` going warm-only with a pinned release** closes the two holes in the old path: unauthenticated presigned-URL requests can no longer trigger a cold open, and the returned backend is pinned until `release()` instead of being closeable underfoot by eviction/cascade.
- **Cascade reaper changes check out**: releasing `sharedMu` before the LRU walk keeps unrelated handle opens unblocked, while the expiry revalidation still happens under `p.mu` so no activity touch can interleave between revalidation and retirement (every `touchSharedDBActivityLocked` caller holds `p.mu`). A new acquire landing between the cascade and `reapIdleSharedDBs` re-leases and refreshes `lastUsed`, so the handle correctly survives the same reap tick.
- Lock order `p.mu -> p.sharedMu` holds at every site; `closeEntry` releases backend -> store -> lease in the right order; the double-create loser path releases all three.

Verified locally: full `pkg/tenant` suite, `pkg/server` leader-lifecycle + shared-DB tests, `-race` over the new lease/cascade/activity tests and `TestLeaderGatedWorkersCloseRacesOnLead`; CI green.

Two non-blocking notes inline. LGTM.

PR #812, opened 2026-07-30
fix(fuse): route PATCH by server storage class, fall back to full upload on 400

mornyx:
## Background

The server log keeps filling with `patch_unsupported_target`. Root cause: FUSE selects PATCH purely from a client-side size heuristic (`OrigSize >= cached inline threshold && newSize <= OrigSize`). Whenever the heuristic disagrees with the server-side storage class — a db9-stored object at size >= threshold, caused by inline-threshold changes, a stale mount-cached threshold, pre-#728 clients, or out-of-band writes — the server rejects PATCH with 400, FUSE returns EINVAL and retains dirty state. Every retry repeats the same doomed PATCH: an infinite failure loop that spams the log and makes the file unflushable. #728 fixed the same-handle truncate source of this mismatch; this PR closes the remaining class with two coexisting layers.

## Changes

**Fact-based routing layer (server + client + FUSE cache)**
- `handleStat` HEAD responses now advertise `X-Dat9-Storage-Type`; the lite stat queries fetch `storage_type` via the contents join (blob/text/description still skipped).
- `pkg/client` parses it into `StatResult.StorageType` (empty = unknown / older server) and exports `IsUnsupportedStorageTargetErr` (`shouldRewriteAppend` refactored on top of it).
- `FileHandle.StorageClass` caches the remote storage class: seeded from stat responses on the writable-open path, copied from sibling handles on preload, re-derived from committed size after successful local commits, and propagated to same-path handles after truncate commits.
- Both PATCH call sites (`flushHandle` Path 2, `syncWriteHandleToRemoteLocked`) bypass PATCH when the class is known db9; when the class is unknown the existing size heuristic applies unchanged, so behavior against older servers is identical. Zero extra requests on common paths — the header piggybacks on HEADs the client already issues.

**Self-healing fallback layer (no more EINVAL loops)**
- On a PATCH 400 (`file is not S3-stored` / `S3 not configured`), the flush records the observed class and falls back to a full conditional upload on the spot: inline in the write-sync path (fh.mu held throughout), and via a depth-bounded `flushHandle` retry in the non-layer path (commit lock already released, DirtySeq verified unchanged).
- After the fallback, the commit re-derives the class from the uploaded size: a file that crossed the threshold heals to S3 permanently; under server/client threshold skew the steady state degrades to at most one wasted PATCH per flush instead of an infinite EINVAL loop.

## Compatibility

- Old client + new server: the new header is ignored; behavior unchanged (the old-client loop still exists and only converges by upgrading).
- New client + old server: no header → class unknown → original heuristic plus the 400 fallback; the loop is eliminated.

## Tests

- server: HEAD returns `db9` for inline files, no header for directories.
- client: `StorageType` parsing (including absence on legacy servers); `IsUnsupportedStorageTargetErr` table test.
- datastore: lite stat contract updated (now returns `storage_type`).
- fuse: scenario-B regression for both flush paths (PATCH 400 → full upload, exactly one PATCH attempt, dirty cleared), known-db9 class skips PATCH, stat header seeds the handle class.

`make lint` reports 0 issues; full test suites of `pkg/datastore`, `pkg/server`, `pkg/client`, `pkg/fuse` pass (the two `TestShouldContinueAfterWaitMountPermissionError*` failures in pkg/fuse reproduce on a clean main checkout — pre-existing environment issue, unrelated to this change).

## Summary by CodeRabbit

* **New Features**
  * Added authoritative storage-class discovery via `X-Dat9-Storage-Type`, propagated through stat results, client metadata, and FUSE write handling (including “healing” after successful writes).

* **Bug Fixes**
  * Improved robustness when PATCH is rejected for unsupported storage targets: reliably falls back to full uploads while preserving dirty/CAS behavior.
  * Ensured safe full-materialization for lazy remote-backed data during truncate/refresh, commit, sync-write, and debounced flushes.

* **Tests**
  * Expanded unit tests for storage-type parsing and unsupported-target detection, plus new FUSE and e2e regression coverage for PATCH fallback/healing.

* **Chores**
  * Added CI and documentation updates for the new e2e suite.

## End-to-end verification (`e2e/fuse-patch-storage-class.sh`, wired as a PR gate in `local-e2e.yml`)

The mismatch is constructed without DB surgery: the server starts with a high `DRIVE9_INLINE_THRESHOLD` so a 200KB seed file is stored inline (db9), then restarts against the same database with a low threshold, so a freshly mounted client's cached threshold sits below the file size while the object remains db9-stored. A partial same-size overwrite through a write-sync FUSE mount then exercises exactly the production mismatch.

Locally verified matrix (macOS + macFUSE):

| Scenario | Server | Client (mount) | Result |
| --- | --- | --- | --- |
| `repro` | main `80ed57b` | main `80ed57b` | write fails `EINVAL`, `patch_unsupported_target` logged, remote content never committed — the bug reproduces |
| `fallback` | main `80ed57b` (no header) | this branch | write succeeds, exactly one PATCH attempt, full-upload fallback commits correct bytes |
| `fixed` | this branch | this branch | HEAD advertises `db9`, zero PATCH attempts, bytes correct, storage healed to `s3` |

The default `fixed` scenario runs in CI on this checkout; `repro`/`fallback` are manual cross-version replays via `DRIVE9_SERVER_BIN` / `DRIVE9_CLI_BIN` overrides (documented in `e2e/README.md` and `e2e/AGENTS.md`).

### DIFF
--- a/pkg/fuse/dat9fs_test.go
+++ b/pkg/fuse/dat9fs_test.go
@@ -16238,7 +16238,8 @@ func newPatchUnsupportedFakeServer(t *testing.T, path string, patchCount *atomic
 	return ts
 }
 
-func newLargeDirtyHandleForPatchTest(fs *Dat9FS, path string, origSize int64, writeSize int) *FileHandle {
+func newLargeDirtyHandleForPatchTest(t *testing.T, fs *Dat9FS, path string, origSize int64, writeSize int) *FileHandle {
+	t.Helper()
 	ino := fs.inodes.Lookup(path, false, 5, time.Now())
 	fs.inodes.UpdateRevision(ino, 5)
 	fh := &FileHandle{
@@ -16253,7 +16254,7 @@ func newLargeDirtyHandleForPatchTest(fs *Dat9FS, path string, origSize int64, wr
 		data[i] = byte(i)
 	}
 	if _, err := fh.Dirty.Write(0, data); err != nil {
-		panic(err)
+		t.Fatalf("seed dirty buffer: %v", err)
 	}
 	fh.DirtySeq = fs.markDirtySize(ino, fh.Dirty.Size())
 	_ = fs.fileHandles.Allocate(fh)
@@ -16279,7 +16280,7 @@ func TestFlushHandle_Path2_PatchUnsupportedTargetFallsBackToFullUpload(t *testin
 	var threshold int64 = 100
 	fs.smallFileMax.Store(threshold)
 
-	fh := newLargeDirtyHandleForPatchTest(fs, path, 200, 120)
+	fh := newLargeDirtyHandleForPatchTest(t, fs, path, 200, 120)
 
 	fh.Lock()
 	st := fs.flushHandle(context.Background(), fh)
@@ -16320,7 +16321,7 @@ func TestFlushHandle_Path2_StorageClassDB9SkipsPatch(t *testing.T) {
 	var threshold int64 = 100
 	fs.smallFileMax.Store(threshold)
 
-	fh := newLargeDirtyHandleForPatchTest(fs, path, 200, 120)
+	fh := newLargeDirtyHandleForPatchTest(t, fs, path, 200, 120)
 	fh.StorageClass = storageClassDB9
 
 	fh.Lock()
@@ -16355,7 +16356,7 @@ func TestSyncWriteHandle_PatchUnsupportedTargetFallsBackToFullUpload(t *testing.
 	var threshold int64 = 100
 	fs.smallFileMax.Store(threshold)
 
-	fh := newLargeDirtyHandleForPatchTest(fs, path, 200, 120)
+	fh := newLargeDirtyHandleForPatchTest(t, fs, path, 200, 120)
 
 	fh.Lock()
 	st := fs.syncWriteHandleToRemoteLocked(context.Background(), fh)
@@ -16409,6 +16410,415 @@ func TestPreloadWritableHandleSeedsStorageClass(t *testing.T) {
 	}
 	if fh.OrigSize != 200 || fh.BaseRev != 3 {
 		t.Fatalf("OrigSize/BaseRev = %d/%d, want 200/3", fh.OrigSize, fh.BaseRev)
+	}
+}
+
+// flushBothWays runs a handle flush through both full-upload entry points:
+// flushHandle (non-layer Path 2) and the write-sync path.
+func flushBothWays(t *testing.T) []struct {
+	name string
+	run  func(*Dat9FS, *FileHandle) gofuse.Status
+} {
+	t.Helper()
+	return []struct {
+		name string
+		run  func(*Dat9FS, *FileHandle) gofuse.Status
+	}{
+		{
+			name: "flushHandle",
+			run: func(fs *Dat9FS, fh *FileHandle) gofuse.Status {
+				fh.Lock()
+				defer fh.Unlock()
+				return fs.flushHandle(context.Background(), fh)
+			},
+		},
+		{
+			name: "write-sync",
+			run: func(fs *Dat9FS, fh *FileHandle) gofuse.Status {
+				fh.Lock()
+				defer fh.Unlock()
+				return fs.syncWriteHandleToRemoteLocked(context.Background(), fh)
+			},
+		},
+	}
+}
+
+// newLazyTwoPartHandle builds a handle whose dirty buffer simulates a
+// lazily loaded existing 512-byte S3 object with 256-byte parts: only the
+// modified part is loaded. origPart returns the canned original content of
+// 1-based part pn; withLoader=false leaves the buffer without a lazy
+// loader (CanMaterializeFull false and no way to refetch).
+func newLazyTwoPartHandle(t *testing.T, fs *Dat9FS, path string, withLoader bool, loadCalls *atomic.Int32) (fh *FileHandle, orig [2][]byte) {
+	t.Helper()
+	ino := fs.inodes.Lookup(path, false, 5, time.Now())
+	fs.inodes.UpdateRevision(ino, 5)
+
+	for p := range orig {
+		orig[p] = make([]byte, 256)
+		for i := range orig[p] {
+			orig[p][i] = byte(0xA0 + p*0x10 + i%16)
+		}
+	}
+
+	wb := NewWriteBuffer(path, maxPreloadSize, 256)
+	wb.remoteSize = 512
+	wb.totalSize = 512
+	if withLoader {
+		wb.LoadPart = func(partNum int) ([]byte, error) {
+			loadCalls.Add(1)
+			return append([]byte(nil), orig[partNum-1]...), nil
+		}
+	}
+
+	fh = &FileHandle{
+		Ino:      ino,
+		Path:     path,
+		Dirty:    wb,
+		BaseRev:  5,
+		OrigSize: 512,
+	}
+	// Modify only the first 4 bytes (loads part 1; part 2 stays unloaded).
+	if _, err := wb.Write(0, []byte("WXYZ")); err != nil {
+		t.Fatalf("seed dirty buffer: %v", err)
+	}
+	if wb.CanMaterializeFull() {
+		t.Fatal("expected CanMaterializeFull() = false for lazy two-part buffer")
+	}
+	fh.DirtySeq = fs.markDirtySize(ino, wb.Size())
+	_ = fs.fileHandles.Allocate(fh)
+	return fh, orig
+}
+
+// TestFlushDirectPUTLazyBufferFetchesMissingParts is the qiffang blocker
+// regression: an existing S3 object with a lazily loaded buffer is routed
+// to direct PUT after the negotiated inline threshold is raised above the
+// file size. Only one part is modified. The flush must fetch the unloaded
+// remote-backed part (never upload zero-filled ranges), and the committed
+// bytes must preserve the untouched original content.
+func TestFlushDirectPUTLazyBufferFetchesMissingParts(t *testing.T) {
+	for _, tc := range flushBothWays(t) {
+		t.Run(tc.name, func(t *testing.T) {
+			var loadCalls atomic.Int32
+			var gotBody []byte
+			var gotExpectedRev string
+
+			ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				if r.Method == http.MethodPut && r.URL.Path == "/v1/fs/lazy-raised-threshold.dat" {
+					gotExpectedRev = r.Header.Get("X-Dat9-Expected-Revision")
+					body, err := io.ReadAll(r.Body)
+					if err != nil {
+						t.Errorf("read PUT body: %v", err)
+					}
+					gotBody = body
+					w.Header().Set("Content-Type", "application/json")
+					_ = json.NewEncoder(w).Encode(map[string]any{"revision": 42})
+					return
+				}
+				t.Errorf("unexpected request: %s %s", r.Method, r.URL.String())
+				w.WriteHeader(http.StatusMethodNotAllowed)
+			}))
+			defer ts.Close()
+
+			opts := &MountOptions{}
+			opts.setDefaults()
+			fs := NewDat9FS(newTestClient(ts.URL), opts)
+
+			// Threshold raised above the file size (512 < 51200): the flush
+			// routes to direct PUT even though the object is S3-stored.
+			fs.smallFileMax.Store(51200)
+
+			path := "/lazy-raised-threshold.dat"
+			fh, orig := newLazyTwoPartHandle(t, fs, path, true, &loadCalls)
+
+			if st := tc.run(fs, fh); st != gofuse.OK {
+				t.Fatalf("flush status = %v, want OK (missing parts must be fetched, not zero-filled)", st)
+			}
+			if loadCalls.Load() == 0 {
+				t.Fatal("expected lazy loader to be called for the unloaded part")
+			}
+			if gotExpectedRev != "5" {
+				t.Fatalf("direct PUT X-Dat9-Expected-Revision = %q, want 5 (CAS must be preserved)", gotExpectedRev)
+			}
+			if int64(len(gotBody)) != 512 {
+				t.Fatalf("direct PUT body size = %d, want 512", len(gotBody))
+			}
+			want := make([]byte, 512)
+			copy(want, orig[0])
+			copy(want[256:], orig[1])
+			copy(want[:4], "WXYZ")
+			if !bytes.Equal(gotBody, want) {
+				t.Fatal("direct PUT body mismatch: untouched remote-backed range was not preserved")
+			}
+			if fh.DirtySeq != 0 {
+				t.Fatalf("DirtySeq = %d, want 0 after successful upload", fh.DirtySeq)
+			}
+		})
+	}
+}
+
+// TestFlushDirectPUTLazyBufferNoLoaderReturnsEIO covers the fail-safe half
+// of the qiffang blocker: without a lazy loader, a direct-PUT flush of a
+// buffer that cannot materialize must fail with EIO and preserve dirty
+// state, never uploading zero-filled ranges.
+func TestFlushDirectPUTLazyBufferNoLoaderReturnsEIO(t *testing.T) {
+	for _, tc := range flushBothWays(t) {
+		t.Run(tc.name, func(t *testing.T) {
+			var serverHit atomic.Bool
+
+			ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				serverHit.Store(true)
+				w.WriteHeader(http.StatusOK)
+			}))
+			defer ts.Close()
+
+			opts := &MountOptions{}
+			opts.setDefaults()
+			fs := NewDat9FS(newTestClient(ts.URL), opts)
+			fs.smallFileMax.Store(51200)
+
+			path := "/lazy-no-loader.dat"
+			fh, _ := newLazyTwoPartHandle(t, fs, path, false, nil)
+			dirtySeqBefore := fh.DirtySeq
+
+			if st := tc.run(fs, fh); st != gofuse.EIO {
+				t.Fatalf("flush status = %v, want EIO (must not upload zero-filled ranges)", st)
+			}
+			if serverHit.Load() {
+				t.Fatal("server received an upload request — must fail before materializing")
+			}
+			if fh.DirtySeq != dirtySeqBefore {
+				t.Fatalf("DirtySeq = %d, want %d preserved after EIO", fh.DirtySeq, dirtySeqBefore)
+			}
+		})
+	}
+}
+
+// patchFallbackV2Fake is a PATCH-rejecting fake server that runs the real
+// v2 multipart protocol for the fallback full upload, recording the
+// expected_revision from the initiate request and the uploaded part bytes.
+type patchFallbackV2Fake struct {
+	t *testing.T
+
+	server       *httptest.Server
+	patchCount   atomic.Int32
+	gotExpected  atomic.Int64
+	gotUpload    atomic.Bool
+	gotBody      []byte
+	conflictMode atomic.Bool // when true, initiate responds 409
+}
+
+func newPatchFallbackV2Fake(t *testing.T, path string) *patchFallbackV2Fake {
+	t.Helper()
+	f := &patchFallbackV2Fake{t: t}
+	f.gotExpected.Store(-1)
+	f.server = httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		switch {
+		case r.Method == http.MethodPatch:
+			f.patchCount.Add(1)
+			_, _ = io.ReadAll(r.Body)
+			w.WriteHeader(http.StatusBadRequest)
+			_ = json.NewEncoder(w).Encode(map[string]string{"error": "file is not S3-stored: " + path})
+
+		case r.Method == http.MethodPut && strings.HasPrefix(r.URL.Path, "/v1/fs/"):
+			// Fallback full rewrite via direct PUT (size below the client's
+			// cached upload threshold): CAS travels in the header.
+			if f.conflictMode.Load() {
+				w.WriteHeader(http.StatusConflict)
+				_ = json.NewEncoder(w).Encode(map[string]string{"error": "revision conflict"})
+				return
+			}
+			if rev, err := strconv.ParseInt(r.Header.Get("X-Dat9-Expected-Revision"), 10, 64); err == nil {
+				f.gotExpected.Store(rev)
+			}
+			f.gotUpload.Store(true)
+			body, err := io.ReadAll(r.Body)
+			if err != nil {
+				t.Errorf("read direct PUT body: %v", err)
+			}
+			f.gotBody = body
+			w.Header().Set("Content-Type", "application/json")
+			_ = json.NewEncoder(w).Encode(map[string]any{"revision": 42})
+
+		case r.Method == http.MethodPost && r.URL.Path == "/v2/uploads/initiate":
+			if f.conflictMode.Load() {
+				w.WriteHeader(http.StatusConflict)
+				_ = json.NewEncoder(w).Encode(map[string]string{"error": "revision conflict"})
+				return
+			}
+			var req struct {
+				ExpectedRevision *int64 `json:"expected_revision"`
+			}
+			if err := json.NewDecoder(r.Body).Decode(&req); err != nil {
+				t.Errorf("decode initiate: %v", err)
+			}
+			if req.ExpectedRevision != nil {
+				f.gotExpected.Store(*req.ExpectedRevision)
+			}
+			w.WriteHeader(http.StatusAccepted)
+			_ = json.NewEncoder(w).Encode(map[string]any{
+				"upload_id":   "upload-1",
+				"key":         "object-key",
+				"part_size":   int64(s3client.PartSize),
+				"total_parts": 1,
+			})
+
+		case r.Method == http.MethodPost && r.URL.Path == "/v2/uploads/upload-1/presign-batch":
+			_ = json.NewEncoder(w).Encode(map[string]any{
+				"parts": []map[string]any{
+					{"number": 1, "url": f.server.URL + "/s3/upload-1/1", "size": 120},
+				},
+			})
+
+		case r.Method == http.MethodPut && r.URL.Path == "/s3/upload-1/1":
+			f.gotUpload.Store(true)
+			body, err := io.ReadAll(r.Body)
+			if err != nil {
+				t.Errorf("read part body: %v", err)
+			}
+			f.gotBody = body
+			w.Header().Set("ETag", "etag-1")
+			w.WriteHeader(http.StatusOK)
+
+		case r.Method == http.MethodPost && r.URL.Path == "/v2/uploads/upload-1/complete":
+			w.WriteHeader(http.StatusOK)
+
+		default:
+			t.Errorf("unexpected request: %s %s", r.Method, r.URL.String())
+			w.WriteHeader(http.StatusMethodNotAllowed)
+		}
+	}))
+	t.Cleanup(f.server.Close)
+	return f
+}
+
+// TestFlushPatchFallbackCarriesExpectedRevision is the qiffang #199 CAS
+// regression: after PATCH is rejected with "file is not S3-stored", the
+// fallback full rewrite must remain revision-conditional (expected_revision
+// carries the handle BaseRev) and must upload the complete buffer content.
+func TestFlushPatchFallbackCarriesExpectedRevision(t *testing.T) {
+	for _, tc := range flushBothWays(t) {
+		t.Run(tc.name, func(t *testing.T) {
+			path := "/fallback-cas.dat"
+			fake := newPatchFallbackV2Fake(t, path)
+
+			opts := &MountOptions{}
+			opts.setDefaults()
+			fs := NewDat9FS(newTestClient(fake.server.URL), opts)
+			fs.smallFileMax.Store(100)
+
+			fh := newLargeDirtyHandleForPatchTest(t, fs, path, 200, 120)
+
+			if st := tc.run(fs, fh); st != gofuse.OK {
+				t.Fatalf("flush status = %v, want OK (fallback full upload)", st)
+			}
+			if got := fake.patchCount.Load(); got != 1 {
+				t.Fatalf("PATCH attempts = %d, want exactly 1", got)
+			}
+			if got := fake.gotExpected.Load(); got != 5 {
+				t.Fatalf("fallback initiate expected_revision = %d, want 5 (BaseRev CAS)", got)
+			}
+			if !fake.gotUpload.Load() {
+				t.Fatal("expected fallback part upload")
+			}
+			if int64(len(fake.gotBody)) != 120 {
+				t.Fatalf("fallback uploaded %d bytes, want 120", len(fake.gotBody))
+			}
+			if fh.DirtySeq != 0 {
+				t.Fatalf("DirtySeq = %d, want 0 after successful fallback", fh.DirtySeq)
+			}
+		})
+	}
+}
+
+// TestFlushPatchFallbackConflictPreservesDirty is the qiffang #199
+// conflict regression: when the fallback full rewrite loses the CAS race
+// (409 revision conflict), the flush must fail without clearing dirty
+// state, while the observed storage class still heals future routing away
+// from PATCH.
+func TestFlushPatchFallbackConflictPreservesDirty(t *testing.T) {
+	for _, tc := range flushBothWays(t) {
+		t.Run(tc.name, func(t *testing.T) {
+			path := "/fallback-conflict.dat"
+			fake := newPatchFallbackV2Fake(t, path)
+			fake.conflictMode.Store(true)
+
+			opts := &MountOptions{}
+			opts.setDefaults()
+			fs := NewDat9FS(newTestClient(fake.server.URL), opts)
+			fs.smallFileMax.Store(100)
+
+			fh := newLargeDirtyHandleForPatchTest(t, fs, path, 200, 120)
+			dirtySeqBefore := fh.DirtySeq
+
+			if st := tc.run(fs, fh); st == gofuse.OK {
+				t.Fatal("flush status = OK, want error on 409 revision conflict")
+			}
+			if got := fake.patchCount.Load(); got != 1 {
+				t.Fatalf("PATCH attempts = %d, want exactly 1", got)
+			}
+			if fake.gotUpload.Load() {
+				t.Fatal("part upload attempted despite initiate conflict")
+			}
+			if fh.DirtySeq != dirtySeqBefore {
+				t.Fatalf("DirtySeq = %d, want %d preserved after conflict", fh.DirtySeq, dirtySeqBefore)
+			}
+			if fh.StorageClass != storageClassDB9 {
+				t.Fatalf("StorageClass = %q, want %q (routing must heal away from PATCH)", fh.StorageClass, storageClassDB9)
+			}
+		})
+	}
+}
+
+// TestFlushPatchFallbackLazyBufferFailsWithoutUpload is the qiffang #199
+// non-materialization regression: after a rejected PATCH, a lazy buffer
+// that cannot materialize the full file must fail with EIO — never issue a
+// full upload with zero-filled ranges, never clear dirty state — while the
+// observed storage class still heals future routing away from PATCH.
+func TestFlushPatchFallbackLazyBufferFailsWithoutUpload(t *testing.T) {
+	for _, tc := range flushBothWays(t) {
+		t.Run(tc.name, func(t *testing.T) {
+			var patchCount atomic.Int32
+			var uploadHit atomic.Bool
+
+			path := "/fallback-lazy.dat"
+			ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+				switch r.Method {
+				case http.MethodPatch:
+					patchCount.Add(1)
+					_, _ = io.ReadAll(r.Body)
+					w.WriteHeader(http.StatusBadRequest)
+					_ = json.NewEncoder(w).Encode(map[string]string{"error": "file is not S3-stored: " + path})
+				default:
+					uploadHit.Store(true)
+					w.WriteHeader(http.StatusOK)
+				}
+			}))
+			defer ts.Close()
+
+			opts := &MountOptions{}
+			opts.setDefaults()
+			fs := NewDat9FS(newTestClient(ts.URL), opts)
+			fs.smallFileMax.Store(100)
+
+			fh, _ := newLazyTwoPartHandle(t, fs, path, false, nil)
+			dirtySeqBefore := fh.DirtySeq
+
+			if st := tc.run(fs, fh); st != gofuse.EIO {
+				t.Fatalf("flush status = %v, want EIO (must not upload zero-filled ranges)", st)
+			}
+			if got := patchCount.Load(); got != 1 {
+				t.Fatalf("PATCH attempts = %d, want exactly 1", got)
+			}
+			if uploadHit.Load() {
+				t.Fatal("full upload attempted for non-materializable buffer — must fail instead")
+			}
+			if fh.DirtySeq != dirtySeqBefore {
+				t.Fatalf("DirtySeq = %d, want %d preserved after EIO", fh.DirtySeq, dirtySeqBefore)
+			}
+			if fh.StorageClass != storageClassDB9 {
+				t.Fatalf("StorageClass = %q, want %q (routing must heal away from PATCH)", fh.StorageClass, storageClassDB9)
+			}
+		})
 	}
 }
 
--- a/pkg/fuse/dat9fs_test.go
+++ b/pkg/fuse/dat9fs_test.go
@@ -16213,6 +16213,205 @@ func TestFlushHandle_Path2_GrowthBypassesPatch(t *testing.T) {
 	}
 }
 
+// newPatchUnsupportedFakeServer returns a test server whose storage semantics
+// reject PATCH with the production "file is not S3-stored" 400 (the file is
+// db9-stored despite its size), while accepting full uploads.
+func newPatchUnsupportedFakeServer(t *testing.T, path string, patchCount *atomic.Int32, fullUpload *atomic.Bool) *httptest.Server {
+	t.Helper()
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		switch r.Method {
+		case http.MethodPatch:
+			patchCount.Add(1)
+			_, _ = io.ReadAll(r.Body)
+			w.WriteHeader(http.StatusBadRequest)
+			_ = json.NewEncoder(w).Encode(map[string]string{"error": "file is not S3-stored: " + path})
+		case http.MethodPut, http.MethodPost:
+			fullUpload.Store(true)
+			_, _ = io.ReadAll(r.Body)
+			w.WriteHeader(http.StatusOK)
+			_ = json.NewEncoder(w).Encode(map[string]interface{}{"status": "ok"})
+		default:
+			w.WriteHeader(http.StatusMethodNotAllowed)
+		}
+	}))
+	t.Cleanup(ts.Close)
+	return ts
+}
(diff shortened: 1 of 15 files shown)


qiffang:
## Dev-Lead correctness review (dat9-dev1) — exact head `27b33eeb`

This is the #199 patch_unsupported fix (storage-class mismatch → PATCH on a DB9 object → 400 → EINVAL, no fallback → infinite doomed-retry loop). I did the #199 root-cause + scope work, so I reviewed against the exact gaps I/dev2 identified there.

### ✅ The two gaps I/dev2 flagged in #199 are both closed

**(B) FUSE PatchFile 400-fallback — the core gap (AppendStream had it, FUSE PatchFile did not) — is now present at BOTH PATCH sites:**
- `write-sync` (dat9fs.go:10925): `if err != nil && client.IsUnsupportedStorageTargetErr(err)` → checks `CanMaterializeFull()` (EIO if the lazy buffer can't materialize — the exact caveat dev2 raised), else `WriteStreamConditional` full upload.
- `flushHandle` (dat9fs.go:12875): same, and **cleaner** — it sets `fh.StorageClass = storageClassDB9` (so this + future flushes self-route away from PATCH), retries via `return fs.flushHandle(...)` with recursion **bounded at depth 1** (StorageClass=db9 can't re-enter the patch branch), and guards the retry with `fh.Path==handlePath && fh.DirtySeq==handleDirtySeq` (no retry if the handle was retargeted/re-dirtied during the unlock window). Good.

I initially suspected `flushHandle` was missing the fallback (the inline patch block ends without it) — it's just positioned after the branches converge, at 12875. Sibling-entry-point completeness holds: both PATCH decision sites (10862 + 12691-12696) apply the StorageClass DB9 override, and both error sites have the fallback.

**(A) fact-based routing:** `X-Dat9-Storage-Type` on stat → `StatResult.StorageType` → `FileHandle.StorageClass`, seeded at stat (2415/9786), post-commit via `adoptCommittedStorageClassLocked` (2094/2337/2370), post-truncate (2592), sibling-copy on preload (3471/3518/3551). PATCH decision forces `canPatchExisting=false` when `StorageClass==storageClassDB9`. When StorageClass is unknown/empty (older server) it falls back to the size heuristic — which is exactly why (B) is the necessary safety net; the two layers are correctly complementary.

### Line-by-line to confirm (for the adversaries)

1. **StorageClass propagation concurrency** (sibling-copy 3471/3518/3548-3551, and `fh.StorageClass = ""` at 3548): is every read/write of `fh.StorageClass` under the same lock discipline? Can a transient empty StorageClass on a sibling wrongly re-enable PATCH between a truncate-to-db9 commit and the sibling adopting the new class? (Defense-in-depth: even if it does, the 400-fallback catches it — so at worst one wasted PATCH round-trip, not data loss. Worth confirming it can't loop.)
2. **`adoptCommittedStorageClassLocked` derivation** (256-265): it derives class from committed size vs threshold — that's the same size-heuristic the bug came from. Is deriving-from-size safe here because the size just came from an authoritative committed write (so server + client agree at that instant)? Confirm it can't re-introduce a stale mismatch on a later out-of-band change (the fallback still covers that, but worth a note).
3. e2e `fuse-patch-storage-class.sh` (+368) and the unit tests (+199 dat9fs, +61 append) — verify they exercise the unknown-StorageClass → wrong-PATCH → 400 → fallback path, not just the known-DB9 fast path.

### Verdict (my axis)
The fix is well-designed defense-in-depth: fact-based routing to avoid the doomed PATCH when the class is known, plus a self-correcting 400-fallback at both PATCH sites for the unknown/stale case. My #199 gaps are closed. Deferring the concurrency line-by-line + gate verification to the adversaries.

pkg/fuse/dat9fs_test.go
```diff
@@ -16213,6 +16213,205 @@ func TestFlushHandle_Path2_GrowthBypassesPatch(t *testing.T) {
...
+func newPatchUnsupportedFakeServer(t *testing.T, path string, patchCount *atomic.Int32, fullUpload *atomic.Bool) *httptest.Server {
+	t.Helper()
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		switch r.Method {
+		case http.MethodPatch:
+			patchCount.Add(1)
+			_, _ = io.ReadAll(r.Body)
+			w.WriteHeader(http.StatusBadRequest)
+			_ = json.NewEncoder(w).Encode(map[string]string{"error": "file is not S3-stored: " + path})
+		case http.MethodPut, http.MethodPost:
```
qiffang, line 16228: This fake server accepts every full-upload request without inspecting the CAS contract. The #199 fallback requirement is specifically a revision-conditional full rewrite after `PATCH` returns `file is not S3-stored`; as written, these tests would still pass if the fallback sent no `expected_revision` or sent the wrong one, which could overwrite a concurrent remote revision. Add a fallback-specific 409/CAS regression that asserts the retry carries `BaseRev` and preserves dirty/error state on conflict.

pkg/fuse/dat9fs_test.go
```diff
@@ -16213,6 +16213,205 @@ func TestFlushHandle_Path2_GrowthBypassesPatch(t *testing.T) {
...
+	ts := newPatchUnsupportedFakeServer(t, path, &patchCount, &fullUpload)
+
+	opts := &MountOptions{}
+	opts.setDefaults()
+	fs := NewDat9FS(newTestClient(ts.URL), opts)
+
+	var threshold int64 = 100
+	fs.smallFileMax.Store(threshold)
+
+	fh := newLargeDirtyHandleForPatchTest(fs, path, 200, 120)
```
qiffang: This scenario-B fallback test uses `NewWriteBuffer` with a fully materialized 120-byte buffer. It does not cover the other #199 safety gate: lazy existing-file buffers where untouched original ranges are not loaded. Add a fallback-specific test where `CanMaterializeFull()==false`, `PATCH` returns `file is not S3-stored`, and the code must not issue a full upload with zero-filled ranges or clear dirty state; the safe result should remain an error with dirty data preserved and future routing healed away from PATCH.

qiffang (changes requested):
Exact-head adversarial review of `27b33eeb`: the implementation shape is directionally correct, and I did not find a code path that drops data in the normal scenario-B flow. However this PR does not satisfy the #199 regression gate yet.

Blocking test gaps:

1. The new FUSE fallback tests do not prove the fallback full upload remains revision-conditional. This was a required safety property for #199: after a rejected PATCH, the retry must not be able to overwrite a concurrent remote revision. The current fake server accepts any PUT/POST and would still pass if the fallback accidentally omitted or changed `expected_revision`.

2. The new fallback tests only cover fully materialized dirty buffers. #199 explicitly called out lazy dirty-buffer non-materialization: after `PATCH` returns `file is not S3-stored`, the fallback must not materialize unloaded original ranges as zeroes. Existing B12 covers the ordinary growth/full-upload path, but not the new unsupported-PATCH fallback branches, especially `syncWriteHandleToRemoteLocked`.

Required before GREEN: add focused regressions that fail if fallback drops CAS, and fail if unsupported-PATCH fallback uploads or clears dirty state when `CanMaterializeFull()==false`.

qiffang (changes requested):
Additional blocker on exact head `27b33eeb684e305e01f69632235f26b008e8a58f`: this PR can silently corrupt data through the full-upload/direct-PUT path.

`syncWriteHandleToRemoteLocked` computes `useDirectPUT` from only `size < threshold` (`pkg/fuse/dat9fs.go:10860`) and eagerly materializes `fh.Dirty.bytesView()` when `useDirectPUT` is true (`:10869-10870`). `flushHandle` has the same shape: `useDirectPUT` is computed before storage-class routing (`:12682`), and the `CanMaterializeFull()` guard is explicitly skipped for direct PUT (`:12734`) before copying `fh.Dirty.bytesView()` (`:12738-12739`). `WriteBuffer.Bytes()` leaves unloaded lazy remote-backed parts as zero (`pkg/fuse/write.go:657-667`), and `CanMaterializeFull()` exists specifically to prevent this (`:824-827`).

A known S3 object can therefore be routed to direct PUT after the negotiated threshold is raised above the file size, even when the write buffer is lazy and only the modified part is loaded. Direct PUT succeeds and uploads zeroes for untouched unloaded ranges, so there is no server 400 and no fallback/self-healing path.

This is a correctness blocker. Before merge, every full-upload path must either prove `CanMaterializeFull()` before `bytesView()`/full materialization, fetch the missing remote parts, or fail without clearing dirty state. Add a deterministic regression for a lazy multi-part existing object below the raised threshold, with only one part modified, asserting untouched remote-backed ranges are preserved. Cover both `syncWriteHandleToRemoteLocked` and `flushHandle`, and keep the CAS/fallback regressions from the earlier review gate.

qiffang (changes requested):
Reviewed exact head `27b33eeb684e305e01f69632235f26b008e8a58f`.

**REQUEST_CHANGES**

### P0 — known-S3 + raised threshold silently zeroes untouched lazy ranges

Both FUSE write paths compute `useDirectPUT` from the current cached threshold before honoring the authoritative storage class:

- `syncWriteHandleToRemoteLocked`: `pkg/fuse/dat9fs.go:10859-10870`
- `flushHandle` Path 2: `pkg/fuse/dat9fs.go:12681-12739`

The comments at 10865 / 12694 say a known-S3 class needs no override when `OrigSize < threshold`, but that is exactly the unsafe case after an operator raises `DRIVE9_INLINE_THRESHOLD`: an existing S3 object remains S3-stored while its size is now below the new threshold. A writable open lazily loads only the touched part. `useDirectPUT` then wins, `bytesView()` / `Bytes()` zero-fills every untouched unloaded part, and the server accepts the direct PUT under the raised threshold. The request succeeds and replaces valid remote bytes with zeroes; no PATCH 400 or fallback can detect it.

I reproduced this deterministically on the exact head in **both** functions with a 512-byte, two-part remote object, `StorageClass=s3`, threshold 1024, and a write only in part 0:

```text
TestFlushHandle_Path2_KnownS3RaisedThresholdPreservesLazyUntouchedRanges:
  uploaded suffix[256:] zeroed=true
TestSyncWriteHandle_KnownS3RaisedThresholdPreservesLazyUntouchedRanges:
  uploaded suffix[256:] zeroed=true
```

This scales directly to production part sizing, e.g. an existing 16 MiB S3 object after the threshold is raised above 16 MiB: editing the first 8 MiB part can zero the unloaded second part.

The fix must make actual storage class override the size heuristic and enforce a materialization invariant at every full-upload entry: before any direct PUT / fallback / write-stream consumes `bytesView()`, either `CanMaterializeFull()` is true, all retained remote ranges are hydrated, or the operation fails without clearing dirty state. Do not make this only an `s3` guard: unknown-class lazy buffers are unsafe to full-upload too.

Add deterministic regressions for both write paths proving untouched remote parts are preserved when a known-S3 object falls below a raised threshold.

### P1 — PATCH 400 fallback still cannot heal a lazy multi-part file

The new tests use fully materialized 120-byte buffers. With a two-part lazy buffer, the unsupported PATCH is attempted once, `StorageClass` is changed to `db9`, and the recursive retry reaches the existing `CanMaterializeFull()==false` guard and returns `EIO`; no full upload occurs. My exact-head regression produced:

```text
flushHandle cannot materialize full file for /lazy-fallback.dat
flushHandle status = EIO; full uploads = 0
```

That is safe against zero-filling, but it does not deliver the PR's stated full-upload fallback for a real multi-part sparse edit and leaves the original mount failure unresolved for this size class. Hydrate the missing ranges (or provide another safe whole-file source) before the conditional fallback, and add the lazy multi-part regression. The fallback test server must also reject a missing/wrong `X-Dat9-Expected-Revision` so the required CAS property is test-discriminating rather than only visible in current source.

### CI integration notes

- The new workflow step lacks `continue-on-error: true` / `if: always()`, so its failure skips later smoke suites instead of collecting the full matrix.
- It is added to `/tmp/e2e-outcomes.json` but not to `e2e/suite-manifest.json`, and `Enforce smoke results` does not name it.
- The 200 KiB e2e fixture is one adaptive 8 MiB part, so it cannot catch either lazy multi-part failure above.

Validation on the exact head:

- PR-focused FUSE tests: pass
- PR-focused FUSE tests with `-race`: pass
- client parser/error tests with external dummy DSN: pass
- `go vet ./pkg/client ./pkg/datastore ./pkg/server ./pkg/fuse`: pass
- `bash -n e2e/fuse-patch-storage-class.sh`: pass
- `git diff --check`: pass
- Hosted `ci`, `drive9-server-local e2e smoke`, CodeRabbit: pass
- DB-backed local package tests remain unavailable on this host because testcontainers reports `rootless Docker not found`

The green hosted checks do not exercise the two deterministic corruption/fallback cases above. Please do not merge this head.

pkg/fuse/dat9fs_test.go
mornyx: Fixed in 543ce17. `TestFlushPatchFallbackCarriesExpectedRevision` (both `flushHandle` and the write-sync path) runs the fallback against a fake that enforces the CAS contract: the full rewrite must carry `X-Dat9-Expected-Revision == BaseRev` (5) plus the complete 120-byte buffer content — it fails if the fallback drops or changes `expected_revision`. `TestFlushPatchFallbackConflictPreservesDirty` adds the 409 case: flush fails, `DirtySeq` is preserved, no part upload is attempted, and `StorageClass` still heals to db9 so future flushes skip PATCH.

pkg/fuse/dat9fs_test.go
mornyx: Fixed in 543ce17. `TestFlushPatchFallbackLazyBufferFailsWithoutUpload` covers both fallback sites with a lazy two-part buffer (`CanMaterializeFull()==false`, no loader): after the PATCH 400 the flush returns EIO, issues no full upload (asserted via a request flag on every non-PATCH endpoint), preserves `DirtySeq`, and records `StorageClass=db9` so future routing heals away from PATCH.

mornyx:
@qiffang All three issues addressed in `543ce17`:

**1. Direct-PUT zero-fill blocker (correctness).** Both full-upload entry points (`syncWriteHandleToRemoteLocked`, `flushHandle` Path 2) now call `materializeFullForUploadLocked` before any eager `bytesView()`: it proves `CanMaterializeFull()`, fetches missing remote-backed parts through the buffer's `LoadPart` lazy loader (new `WriteBuffer.LoadMissingParts`, which refuses to refetch streaming-evicted parts), or fails `EIO` without clearing dirty state. The flushHandle B12 guard is extended from stream-only to cover direct PUT as well. Regressions on both paths:
- `TestFlushDirectPUTLazyBufferFetchesMissingParts` — lazy two-part S3 object below the raised threshold, only one part modified: the unloaded part is fetched via the loader and the uploaded 512 bytes preserve the untouched original range (byte-exact), with the CAS header intact.
- `TestFlushDirectPUTLazyBufferNoLoaderReturnsEIO` — no loader: `EIO`, zero upload requests reach the server, `DirtySeq` preserved.

**2. CAS fallback regression.** `TestFlushPatchFallbackCarriesExpectedRevision` (both paths) uses a fake that enforces the CAS contract — the fallback rewrite must carry `expected_revision == BaseRev` and the complete buffer content; `TestFlushPatchFallbackConflictPreservesDirty` adds the 409 case (flush fails, dirty preserved, routing healed away from PATCH).

**3. Lazy-buffer fallback regression.** `TestFlushPatchFallbackLazyBufferFailsWithoutUpload` (both paths): `CanMaterializeFull()==false` after a PATCH 400 → `EIO`, no full upload issued, dirty preserved, `StorageClass=db9` recorded.

Also folded in the review nits: `client.StorageType` is now a typed string enum (used by `StatResult.StorageType` and `FileHandle.StorageClass`), and the patch-test helper takes `*testing.T`. `make lint` is clean; full `pkg/fuse` + `pkg/client` suites pass (the two `TestShouldContinueAfterWaitMountPermissionError*` failures reproduce on a clean main checkout — pre-existing environment issue).

mornyx:
@qiffang Answers to the three line-by-line questions in your Dev-Lead review:

**Q1 — StorageClass lock discipline / transient empty re-enabling PATCH.** Audited: every read and write of `fh.StorageClass` happens either under `fh.mu` or before the handle is published. Writes: open-time seeding happens while the handle is still local to `Open` (published only via `allocateFileHandle` at the end); sibling preload copies under `fh.Lock()` with the source handle locked during candidate capture; `adoptCommittedStorageClassLocked` and both fallback assignments run with `fh.mu` held (the write-sync path holds it for the whole flush; `flushHandle` assigns after re-locking). Reads: both routing sites evaluate `fh.StorageClass` with `fh.mu` held. A sibling can indeed transiently copy `""` (sibling never stat'd / never committed) and fall back to the size heuristic — but the failure mode is bounded exactly as you said: one wasted PATCH → 400 → fallback sets `db9` on that handle permanently. It cannot loop: under threshold skew the steady state is one wasted PATCH per flush (each flush still succeeds), never the pre-fix infinite EINVAL loop, and no data loss in either case. `isNew → ""` is intentional: a new file has no remote object, so routing is governed by the `IsNew`/size paths, not by class.

**Q2 — Deriving class from committed size.** Safe because of *when* it runs: immediately after this client committed a full-content write of exactly that size. The server's storage decision for that very write is `size vs server threshold`, and the client derives from the same size with the server-advertised threshold, so both sides agree at that instant. The skew window (operator changes threshold between negotiation and commit) can derive a wrong `s3`, but the failure is again self-correcting and loss-free: next flush either wastes one PATCH (400 → fallback heals) or full-uploads (harmless). Later out-of-band rewrites by other clients bump the revision, so a stale class never gets to act — the CAS (`expected_revision`) fails first with 409, and the 400-fallback remains as backstop. Stat-seeded (authoritative) values always take precedence at open; derivation only fills the post-commit gap and never overwrites with empty.

**Q3 — Coverage of the unknown-class → wrong-PATCH → 400 → fallback path.** Unit: `TestFlushHandle_Path2_PatchUnsupportedTargetFallsBackToFullUpload` and `TestSyncWriteHandle_PatchUnsupportedTargetFallsBackToFullUpload` both construct handles with `StorageClass == ""` (unknown) so the heuristic selects PATCH against a 400-ing fake server, and the new CAS/conflict/lazy regressions in `543ce17` all run unknown-class as well. e2e: `SCENARIO=fallback` runs the fixed client against a main-built server (no `X-Dat9-Storage-Type` header → class unknown) asserting exactly one PATCH attempt then a successful fallback; `SCENARIO=repro` replays the pre-fix EINVAL loop on main-built binaries. The known-DB9 fast path is covered separately by `TestFlushHandle_Path2_StorageClassDB9SkipsPatch` and e2e `SCENARIO=fixed` (zero PATCH attempts).

qiffang (changes requested):
Reviewed exact head `543ce171ff5001e64ee03b8b3fabc27635c615f4`.

**REQUEST_CHANGES**

### P1: write-sync PATCH fallback still cannot recover a lazy buffer with a loader

`syncWriteHandleToRemoteLocked` handles `PatchFile` 400 `file is not S3-stored` at `pkg/fuse/dat9fs.go:10955-10966`, but the fallback still checks only `fh.Dirty.CanMaterializeFull()` before `fh.Dirty.bytesView()`.

That is safe against zero-fill, but it does not satisfy the required recoverable lazy-multipart behavior. A lazily preloaded existing file can have `LoadPart` available and still have `CanMaterializeFull()==false` until the missing retained parts are fetched. This branch never calls the new `fs.materializeFullForUploadLocked(fh)`, so write-sync returns `EIO` instead of fetching the untouched remote range and doing the revision-conditional full upload.

The tests do not catch this: `TestFlushPatchFallbackLazyBufferFailsWithoutUpload` only covers `withLoader=false` (`pkg/fuse/dat9fs_test.go:16772-16823`), while the successful fallback tests use fully materialized buffers from `newLargeDirtyHandleForPatchTest`.

Required before GREEN: use the same materialization helper in the write-sync PATCH fallback before `bytesView()`, and add a `withLoader=true` regression covering both `flushHandle` and `syncWriteHandleToRemoteLocked`: PATCH 400 -> missing part fetched -> full upload succeeds -> untouched bytes preserved -> expected revision carried -> dirty cleared.

### P1: write-sync full-stream path has the same fetch gap

The non-patch write-sync stream branch at `pkg/fuse/dat9fs.go:10982-10987` still returns `EIO` whenever `CanMaterializeFull()` is false. For large existing file growth (`OrigSize >= threshold`, `size > OrigSize`), this is a recoverable full-upload case when a lazy loader exists. `flushHandle` now does the right thing via `materializeFullForUploadLocked` at `pkg/fuse/dat9fs.go:12758-12771`; write-sync should not have weaker behavior.

### Cross-revision gate is still test-only incomplete

The missing-part loaders still read latest by path (`pkg/fuse/dat9fs.go:2144` and `pkg/fuse/dat9fs.go:2489`) rather than a revision-pinned range. I agree the current lock/order shape can use the final expected-revision CAS as the safety backstop, but that contract needs a discriminating regression: during materialization, make the retained-range fetch observe remote revision N while the dirty handle still uploads with expected revision M, then assert the upload gets 409, dirty state is preserved, and remote N is not overwritten. Without that test, moving expected-revision capture or BaseRev adoption later would silently reopen the splice bug.

Validation on this head:

- `gofmt -d pkg/client/client.go pkg/fuse/dat9fs.go pkg/fuse/dat9fs_test.go pkg/fuse/handle.go pkg/fuse/write.go`: no diff
- focused `go test ./pkg/fuse -run 'TestFlushDirectPUTLazyBufferFetchesMissingParts|TestFlushDirectPUTLazyBufferNoLoaderReturnsEIO|TestFlushPatchFallbackCarriesExpectedRevision|TestFlushPatchFallbackConflictPreservesDirty|TestFlushPatchFallbackLazyBufferFailsWithoutUpload|TestSyncWriteHandle_PatchUnsupportedTargetFallsBackToFullUpload|TestFlushHandle_Path2_PatchUnsupportedTargetFallsBackToFullUpload' -count=1`: pass
- `go test -c ./pkg/fuse`: pass
- `go vet ./pkg/client ./pkg/fuse`: pass
- Hosted `ci`, `drive9-server-local e2e smoke`, and CodeRabbit: pass

Green checks do not cover the two write-sync lazy materialization gaps above. Do not merge this head.

mornyx:
@qiffang All three points addressed in `e04c1f7`:

**P1 — write-sync PATCH fallback lazy recovery.** The fallback now calls `materializeFullForUploadLocked` instead of checking only `CanMaterializeFull()`: with a loader it fetches the untouched remote-backed parts and completes the revision-conditional full upload; without a loader it still fails `EIO` without clearing dirty (the `withLoader=false` regression keeps that contract). New regression `TestFlushPatchFallbackLazyBufferFetchesAndRecovers` (both `flushHandle` and write-sync) covers your required chain: PATCH 400 → missing part fetched (`loadCalls > 0`) → full upload succeeds → untouched bytes byte-exact preserved → `X-Dat9-Expected-Revision == BaseRev` → `DirtySeq == 0`.

**P1 — write-sync stream branch fetch gap.** The non-patch stream branch (`size > OrigSize` growth) now uses the same helper, matching `flushHandle`. New regression `TestFlushGrowthLazyBufferFetchesMissingParts` (both paths): lazy two-part buffer grown 512→600 with only part 1 modified → part 2 fetched → uploaded 600 bytes preserve the untouched original range, CAS header intact.

**Cross-revision splice gate.** New discriminating regression `TestFlushPatchFallbackLazyFetchNewerRevisionConflicts` (both paths): the loader serves the latest (revision-9) bytes while the handle uploads with `expected_revision == 5`; the fake remote (at rev 9) rejects with 409; assertions: no commit accepted, dirty preserved, exactly one PATCH attempt, `StorageClass` healed to db9. Any future change that moves expected-revision capture later would flip this test red.

Validation: focused runs of all seven patch/lazy suites pass on both flush paths, `go vet ./pkg/fuse` clean, gofmt clean, `make lint` 0 issues, full `pkg/fuse` suite green except the two pre-existing `TestShouldContinueAfterWaitMountPermissionError*` environment failures (reproduced on clean main).

qiffang (commented):
adversary-1 review for exact head `e04c1f7a800fe40c40414938fbecde7034536582`.

Code-level verdict on the prior #199 blockers: GREEN. I do not see a remaining code blocker in the PATCH-unsupported / lazy materialization paths:

- `syncWriteHandleToRemoteLocked` now uses `materializeFullForUploadLocked` in both previously missing branches: PATCH-400 fallback before full-upload fallback, and the non-PATCH `WriteStreamConditional` full-upload branch.
- Full-materialization paths now have a fetch-or-fail guard before `bytesView()`, so unloaded lazy remote-backed ranges are not silently zero-filled.
- The new regressions run both entry points via `flushBothWays`: direct-PUT lazy fetch/no-loader EIO, PATCH fallback lazy fetch success/no-loader EIO, growth full-upload lazy fetch, expected-revision preservation, 409 conflict dirty preservation, and the fetch-latest + expected-base CAS backstop.

Local validation:

- `gofmt -d pkg/fuse/dat9fs.go pkg/fuse/dat9fs_test.go` clean.
- Focused `go test ./pkg/fuse -run 'TestFlushHandle_Path2_PatchUnsupportedTargetFallsBackToFullUpload|TestSyncWriteHandle_PatchUnsupportedTargetFallsBackToFullUpload|TestFlushDirectPUTLazyBufferFetchesMissingParts|TestFlushDirectPUTLazyBufferNoLoaderReturnsEIO|TestFlushPatchFallbackCarriesExpectedRevision|TestFlushPatchFallbackConflictPreservesDirty|TestFlushPatchFallbackLazyBufferFailsWithoutUpload|TestFlushPatchFallbackLazyBufferFetchesAndRecovers|TestFlushGrowthLazyBufferFetchesMissingParts|TestFlushPatchFallbackLazyFetchNewerRevisionConflicts' -count=1` passed.
- Focused race `go test -race ./pkg/fuse -run 'TestFlushPatchFallbackLazyBufferFetchesAndRecovers|TestFlushGrowthLazyBufferFetchesMissingParts|TestFlushPatchFallbackLazyFetchNewerRevisionConflicts|TestFlushDirectPUTLazyBufferFetchesMissingParts' -count=1` passed.
- `go test -c ./pkg/fuse`, `go vet ./pkg/client ./pkg/fuse`, and `bash -n e2e/fuse-patch-storage-class.sh` passed.
- Client focused tests passed with `-tags integration` to avoid the local Docker `TestMain` path.

Merge hold: I am not approving while hosted checks are red. Current GitHub status has `ci` passing, but `drive9-server-local e2e smoke` failing. The failing log is mount readiness, not a #199 assertion failure: `background mount exited before becoming ready` and `FAIL mount point is mounted (want=true got=false)`, causing Layer FS/FUSE release gate and downstream smoke gates to fail/skip. This needs a green rerun or a separate fix before merge.

qiffang (changes requested):
adversary-1 update for exact head `e04c1f7a800fe40c40414938fbecde7034536582`: retracting my earlier code-level GREEN. I missed one remaining full-upload entry point.

Blocking finding:

- `pkg/fuse/dat9fs.go:12287-12295` schedules debounced small-file uploads by snapshotting `fh.Dirty.bytesView()` directly, without `materializeFullForUploadLocked`. The callback then uploads that snapshot with `WriteCtxConditionalWithRevision` at `pkg/fuse/dat9fs.go:12359-12362` and clears dirty state on success at `12399-12401`.

Why this is a correctness blocker:

`WriteBuffer.Bytes()`/`bytesView()` zero-fills unloaded lazy parts. The new guards cover `flushHandle` and `syncWriteHandleToRemoteLocked`, but `flushHandleDebounced` is also a remote full-upload path and still bypasses the guard. This is exploitable when the server inline threshold is configured above the dirty-buffer part size, which the server permits via any positive `DRIVE9_INLINE_THRESHOLD`, and FUSE mirrors through `/v1/status.inline_threshold`.

Concrete sequence:

1. An existing S3-backed object was opened lazily before/after an inline-threshold raise; e.g. 9 MiB file, adaptive part size 8 MiB, server inline threshold 16 MiB.
2. The mount is in a mode where debounce is enabled (`--flush-debounce` default path; profiles may disable it, but the code path is live).
3. User modifies bytes in part 1 only. Part 1 is loaded/dirty, part 2 remains unloaded, so `CanMaterializeFull()` is false.
4. A non-forced `Flush` routes to `flushHandleDebounced` because `size < fs.inlineThreshold()`.
5. `snapshot := fh.Dirty.bytesView()` zero-fills part 2, the debounced callback uploads the corrupted full body, then clears dirty state.

This violates the invariant we set for #199: every full-upload/direct-PUT/fallback/write-stream entry must prove full materialization, fetch missing retained remote parts, or fail preserving dirty before materializing bytes.

Required before GREEN:

- Guard the debounced snapshot path too, either by calling `materializeFullForUploadLocked` before `bytesView()` or by declining debounce and falling through to a guarded synchronous flush when full materialization is not already proven.
- Add a regression that exercises `flushHandleDebounced` with a lazy multi-part existing file, `inlineThreshold > partSize`, dirty only part 0, and asserts the untouched later part is fetched/preserved (or no upload + dirty preserved if no loader). The current `flushBothWays` matrix bypasses this deferred-upload entry point, so it would not catch this remaining zero-fill path.

Hosted e2e is still a separate merge gate, but this is now a code blocker independent of the pending e2e rerun.

mornyx:
@qiffang Good catch — that was a real remaining entry point. Fixed in `19d7171`:

- `flushHandleDebounced` now runs `materializeFullForUploadLocked` before snapshotting: with a loader it fetches the unloaded retained parts and the debounce proceeds with a complete snapshot; without a loader the flush fails `EIO` synchronously — nothing scheduled, dirty preserved. Debounce semantics are unchanged for the common (fully loaded) case.

Regressions:
- `TestFlushHandleDebouncedLazyBufferFetchesMissingParts` — lazy two-part buffer, size below the inline threshold, only part 0 dirty: schedules, `FlushAll` runs the callback, and the debounced PUT carries the fetched untouched bytes byte-exact plus `X-Dat9-Expected-Revision == BaseRev`.
- `TestFlushHandleDebouncedLazyBufferNoLoaderReturnsEIO` — no loader: `EIO` synchronously, zero upload requests, `DirtySeq` preserved.

This makes the full-upload entry-point coverage complete across all three sites: `flushHandle` Path 2, `syncWriteHandleToRemoteLocked` (direct-PUT eager block, PATCH fallback, stream branch), and now the debounced snapshot. `go vet`/gofmt clean, `make lint` 0 issues, full `pkg/fuse` suite green except the two pre-existing macOS-only `TestShouldContinueAfterWaitMountPermissionError*` failures (they lack a non-Linux skip; unrelated to this PR).

qiffang (changes requested):
adversary-1 review on exact head `19d71719f33cd52b69a8a493f2b7d36954d5d184`: REQUEST CHANGES.

Blocker: the debounced lazy-materialization path still has a cross-revision splice window. In `flushHandleDebounced`, the missing-part fetch/materialization and full-buffer snapshot happen before the CAS base is frozen (`pkg/fuse/dat9fs.go:12300-12312`). `expectedRevisionForHandleLocked` calls `adoptCommittedRevisionLocked`, which unconditionally advances `fh.BaseRev` from the same-process committed-revision map (`pkg/fuse/dat9fs.go:2310-2322`, `3038-3040`).

Concrete sequence: handle A opened/dirty at rev 5 with part 2 unloaded; its debounced flush enters `materializeFullForUploadLocked`; while A is loading the missing part, sibling handle B commits rev 9 and records it; A's loader reads latest rev-9 retained bytes; then line 12312 adopts rev 9 as `expectedRevision`. The callback recheck sees rev 9, does not skip, and the deferred PUT can succeed with `X-Dat9-Expected-Revision: 9` while the dirty bytes were based on rev 5. That is the exact splice/lost-update class the sync/flush paths now defend with the rev-5 CAS-backstop.

The new debounced tests cover normal loader success and no-loader EIO, but not this concurrent committed-revision advance. Add a debounced equivalent of `TestFlushPatchFallbackLazyFetchNewerRevisionConflicts`: force `recordCommittedRevision(path, 9)` during the lazy load, then assert the debounced path does not PUT with rev 9, does not clear dirty state, and leaves remote unchanged. Until that passes, PR #812 remains REVISE / do not merge.

Local checks run for this review: `git diff --check origin/main...HEAD`, `gofmt -d pkg/fuse/dat9fs.go pkg/fuse/dat9fs_test.go`, and focused `go test ./pkg/fuse -run 'TestFlushHandleDebouncedLazyBufferFetchesMissingParts|TestFlushHandleDebouncedLazyBufferNoLoaderReturnsEIO|TestFlushPatchFallbackLazyFetchNewerRevisionConflicts' -count=1` all passed. Hosted checks were still pending when reviewed; this blocker is independent of CI.

qiffang:
Additional adversary-1 blocker on exact head `19d71719f33cd52b69a8a493f2b7d36954d5d184`, aligned with @dat9-dev2's layer-path finding:

The layer-enabled full-write path is not just a harmless alternate ShadowSpill fallback. For a normal lazy existing file, `preloadWritableHandle` installs `LoadPart` but does not set `ShadowSpill`. In layer mode, `flushHandle` checks `!fh.Dirty.CanMaterializeFull()` and returns EIO without calling `materializeFullForUploadLocked` (`pkg/fuse/dat9fs.go:12505-12515`). That means a recoverable lazy sparse edit with a loader never fetches retained parts and never reaches `upsertLayerFile`.

The worse part is the release path: after Release's final sync flush, it clears the dirty-size sequence unconditionally even when `flushHandle` returned EIO (`pkg/fuse/dat9fs.go:12253-12259`), then deletes the file handle in the deferred cleanup. For this ordinary non-ShadowSpill lazy handle there is no authoritative shadow/pending fallback, so the mutation is no longer durably retryable after close. This is not P0 zero-fill corruption, but it is still a correctness/data-loss blocker for layer-enabled mounts.

Next head must prove one of these contracts with regression coverage: either layer normal-lazy handles call `materializeFullForUploadLocked` under a pinned base revision and successfully upsert preserved bytes, or any EIO path leaves durable shadow/pending state that survives Release and can retry. ShadowSpill-specific tests do not cover this case; the regression must be ordinary lazy existing file, loader available, dirty only one part, layer enabled.

mornyx:
@qiffang Valid find — my `19d7171` fix added the fetch *before* the CAS base freeze, which inverted the ordering the synchronous paths rely on. Fixed in `b2f1871`:

**Root cause.** `expectedRevisionForHandleLocked` adopts the same-process committed-revision map, so capturing it after the fetch let a mid-fetch sibling commit advance the base. `expectedRevision` is now captured BEFORE `materializeFullForUploadLocked`, matching the freeze-then-fetch order of `flushHandle` Path 2 and `syncWriteHandleToRemoteLocked`. A sibling commit during the fetch now trips the callback's `currentExpectedRevision != expectedRevision` re-check and the upload is skipped, dirty preserved.

**Discriminating regression.** `TestFlushHandleDebouncedLazyFetchNewerRevisionSkipsUpload` records a rev-9 sibling commit from inside the lazy loader at exactly the moment the flush requests the missing part (part 2; part 1 loads earlier during seeding and does not record): asserts the callback never PUTs, `DirtySeq`/dirty parts preserved, no remote write. Verified it fails against the pre-fix ordering (freeze-after-fetch uploads with rev 9) and passes now.

**Full-entry-point audit.** To stop finding these one at a time, I swept every `bytesView()`/`Bytes()` call site in `pkg/fuse` for the materialization contract:

| Site | Status |
| --- | --- |
| `dat9fs.go` readCache.Put (sqlite sidecar) | guarded by `writeBufferHasLoadedFullRange` |
| `dat9fs.go` shadow WriteFull (`stageShadowLocked`) | guarded by `canStageShadowFastLocked` → `CanMaterializeFull` |
| `dat9fs.go` writeBack snapshot (`snapshotWriteBackLocked`) | guarded by explicit `CanMaterializeFull` check (ENOTSUP) |
| `dat9fs.go` sibling preload (`loadWritableHandleFromOpenHandleLocked`) | guarded by `canMemory` candidate filter |
| `dat9fs.go` layer upsert (`flushHandle` layer branch) | guarded by `CanMaterializeFull` branch check |
| `dat9fs.go` write-sync eager/fallback/stream, flushHandle Path 2, debounced snapshot | guarded by `materializeFullForUploadLocked` (this PR) |
| `git_workspace.go` overlay upsert | not lazy — git buffers are fully hydrated at open (`remoteSize == 0`) |

No fourth unguarded entry point exists. `go vet`/gofmt clean, `make lint` 0 issues, full `pkg/fuse` suite green except the two pre-existing macOS-only `TestShouldContinueAfterWaitMountPermissionError*` failures.

qiffang (changes requested):
adversary-1 review on exact head `b2f1871a07928eff8d3797c6ade6731b750354bc`: REVISE.

This commit correctly fixes the debounced cross-revision blocker: `flushHandleDebounced` now freezes `expectedRevision` before `materializeFullForUploadLocked`, and `TestFlushHandleDebouncedLazyFetchNewerRevisionSkipsUpload` is the right discriminating regression. Focused debounced/fallback tests pass locally.

However this latest delta only touches the debounced path. The other two previously confirmed blockers are still present:

1. Layer ordinary lazy handles still fail without materializing/fetching and then Release drops retry state. In layer mode `flushHandle` still returns EIO on `!fh.Dirty.CanMaterializeFull()` instead of using `materializeFullForUploadLocked` or durable staging (`pkg/fuse/dat9fs.go:12514-12524`). The Release path still calls `flushHandle` and clears `DirtySeq` regardless of `st` (`pkg/fuse/dat9fs.go:12253-12259`), so a recoverable lazy sparse edit can be accepted locally, fail flush, then lose its retry source when the handle is released.

2. ShadowSpill eviction can still feed a zero-filled writeBack snapshot. The small writeBack path still allows `canStageShadowFastLocked(fh)` for `ShadowReady`/`IsNew`, calls `snapshotWriteBackLocked`, and records `WriteBackSeq` (`pkg/fuse/dat9fs.go:11297-11320`). `snapshotWriteBackLocked` still bypasses the materialization guard for `fh.ShadowReady`/`fh.IsNew` and then snapshots `fh.Dirty.bytesView()` (`pkg/fuse/dat9fs.go:3325-3353`). For an evicted ShadowSpill buffer, `bytesView()` is not the authoritative source; the authoritative bytes are in shadow. This is the previously proven corruption path where writeBack `.dat`/upload body can become zero-filled.

Required before GREEN: fix and regress both remaining paths. Layer ordinary lazy handles must either fetch/materialize under a pinned base revision and upsert layer bytes, or fail while preserving durable retry state through Release. ShadowSpill small-file Flush/writeBack must never snapshot from `Dirty.bytesView()` when parts are evicted; it must use authoritative shadow or avoid writeBack and commit from shadow, including commitQueue enqueue-failure/fallback behavior.

Validation I ran:
- diff-check, gofmt diff
- focused pkg/fuse debounced/direct-PUT/fallback/lazy tests
- focused race debounced tests
- `go test -c ./pkg/fuse`
- `go vet ./pkg/client ./pkg/fuse`
- `bash -n e2e/fuse-patch-storage-class.sh`

Hosted `ci` is now green, `drive9-server-local e2e smoke` is still pending, and CodeRabbit is green. CI status does not affect this blocker.

mornyx:
@qiffang Both findings verify as real, and both are pre-existing in the write-back / shadow / layer durability subsystem rather than introduced by this PR. I've split them into a dedicated issue: **#815** (with reproduction paths and the regression requirements lifted from your review).

Rationale for not fixing them in this PR:

- This PR's contract is now closed and audited end-to-end: PATCH storage-class routing, the 400 fallback, and the full-upload materialization guards for the two main flush paths plus the debounced path — with an entry-point audit showing no remaining unguarded `bytesView()` site in those paths.
- The two new items live in a different subsystem with its own invariants (shadow-authoritative reads, `Release` retry semantics, layer hydration). Fixing them properly means changes like "commit from shadow instead of buffer" and Release-failure state handling — the risk of rushing that into this PR outweighs the benefit, especially since the hazards are pre-existing and unchanged by this PR.

Proposal: land this PR on its own contract, and treat #815 as the follow-up for the layer-mode and ShadowSpill paths (happy to take it next, or whoever picks it up gets a head start from your reproduction sketches). If you disagree and want them gated on this PR, say so and we'll re-scope.

qiffang (approved):
adversary-1 scoped re-review on exact head `b2f1871a07928eff8d3797c6ade6731b750354bc`: GREEN for PR #812 under qiffang's updated boundary.

I treated issue #815 as the canonical tracker for the pre-existing layer-mode lazy-handle retry-loss path and the pre-existing ShadowSpill writeBack zero-fill path, so those are no longer PR #812 blockers. Re-checking `base..head` under that boundary, I did not find another PR-introduced correctness blocker.

What I verified:
- Server `HEAD` now exposes authoritative `X-Dat9-Storage-Type`, with lite stat fetching `contents.storage_type`.
- Client `StatResult.StorageType` parses that header and FUSE seeds/updates `FileHandle.StorageClass` from stat and successful commits.
- Known `db9` handles route away from PATCH; unsupported PATCH 400 (`file is not S3-stored`) records db9 and falls back to a revision-conditional full upload.
- Full-upload/direct-PUT/write-stream/debounced paths in PR #812 scope use `materializeFullForUploadLocked` before `bytesView()`, so lazy retained ranges are fetched when possible or fail without dirty-clear when not.
- Debounced upload now freezes `expectedRevision` before materialization/fetch and re-checks before PUT, so the previously reviewed cross-revision splice is closed.

Validation I ran:
- `bash -n e2e/fuse-patch-storage-class.sh`
- focused `go test ./pkg/fuse` for storage-class fallback, lazy materialization, conflict, growth, and debounced regressions
- focused `go test -race ./pkg/fuse` for the lazy/CAS/debounce cases
- focused `go test ./pkg/client` with dummy `DRIVE9_TEST_MYSQL_DSN` for pure client append/stat tests
- `go test -c ./pkg/fuse ./pkg/client ./pkg/server ./pkg/datastore`
- `go vet ./pkg/client ./pkg/fuse ./pkg/server ./pkg/datastore`
- hosted `ci`, `drive9-server-local e2e smoke`, and CodeRabbit are green

Non-blocking cleanup: `gofmt -d` across all touched Go files reports formatting-only changes in `pkg/datastore/store.go` (SQL string concatenation spacing plus an existing nearby one-liner). I am not treating that as a PR #812 correctness blocker under this requested review boundary.

qiffang (changes requested):
adversary-1 follow-up on exact head `b2f1871a07928eff8d3797c6ade6731b750354bc`: REVISE due to a PR-introduced e2e wiring blocker found by dat9-dev2 and independently confirmed.

This does not change my prior code-correctness assessment: the FUSE/client/server storage-class routing, lazy materialization guards, and debounced CAS-base ordering are still GREEN under qiffang's scoped boundary, and the layer/ShadowSpill data-loss issues remain tracked by issue #815 rather than blocking PR #812.

New blocker: PR #812 adds a new local-e2e step/outcome id `fuse-patch-storage-class` in `.github/workflows/local-e2e.yml` (`id`, step summary row, and `/tmp/e2e-outcomes.json` entry), but does not add the same suite id to `e2e/suite-manifest.json`. The manifest explicitly says it is keyed by `local-e2e.yml` step id and must be kept in sync when adding suites. This is in the PR diff, so it is #812-introduced.

The impact is not cosmetic. I reproduced the actual aggregator behavior with outcomes `{"fuse-patch-storage-class":"failure"}`: `cmd/e2e-aggregate` emits the failure as `class=unknown, promise=-, owner=-`, so nightly/post-merge issue/Feishu owner routing loses the product promise and owner for the new suite.

Required before GREEN:
- Add `fuse-patch-storage-class` metadata to `e2e/suite-manifest.json` (fuse area, fs owner, correctness class, and a PATCH/storage-class fallback promise).
- Add/update a manifest-sync regression so a new workflow outcome id cannot silently synthesize unknown metadata again. At minimum, this suite id should synthesize non-empty product promise/owner/failure class through the real aggregator/report path.

Validation for this follow-up:
- `rg` confirms the workflow emits `fuse-patch-storage-class` while `e2e/suite-manifest.json` lacks it.
- `go run ./cmd/e2e-aggregate --manifest e2e/suite-manifest.json --outcomes <tmp outcomes> --tier nightly` reproduces `class=unknown, promise=-, owner=-` for this suite.

qiffang (approved):

mornyx:
@qiffang Closing the loop on your last two #812 blockers in #816 (they were deferred there by agreement):

1. **suite-manifest gap**: `fuse-patch-storage-class` is now in `e2e/suite-manifest.json`, plus `failure_class: correctness` backfilled for the six suites that also synthesized `class=unknown` (`api-smoke`, `cli-smoke`, `layer-fs-smoke`, `fuse-smoke`, `pack-smoke`, `smoke-all`). New `internal/e2ereport/manifest_sync_test.go` pins every `local-e2e.yml` outcome id to classified manifest metadata through the real `SynthesizeSummaries` path, so a future step id cannot silently degrade again. Verified with `cmd/e2e-aggregate` against `{"fuse-patch-storage-class":"failure"}`: now emits `class=correctness, promise=…, owner=fs`.

2. **#815 (layer-mode lazy EIO + ShadowSpill writeBack zero-fill)**: both fixed in #816 with regressions — see the PR body and my comment on #815 for details.

PR #814, opened 2026-07-30
fix(server): Retry-After on provisioning 503, 403 for failed tenants

srstack:
## Why

- Tenant `pending`/`provisioning` responses return 503 with correct semantics (transient, retryable), but lack a `Retry-After` header, so clients cannot back off properly (as recommended in docs/design/agent-journal-design.md).
- Tenant `failed` is a persistent state; returning 503 (temporarily unavailable) is semantically inaccurate.

## Changes

- Add `errJSONRetryable`: 503 responses now consistently include `Retry-After: 1`.
- Both pending/provisioning branches in `auth.go` use `errJSONRetryable` (regular request path and `handleDeleteNoAcquireAuth`).
- `TenantFailed` now returns 403 instead of 503 (delete-tenant requests are unaffected; failed tenants are still allowed through that path).

## Test

- `make test TEST_PKGS='./pkg/server/...'` passes
- `make lint` reports 0 issues

### DIFF
--- a/pkg/server/auth.go
+++ b/pkg/server/auth.go
@@ -306,21 +306,21 @@ func tenantAuthMiddlewareWithFSScopeLoader(metaStore *meta.Store, pool *tenant.P
 
 		switch resolved.Tenant.Status {
 		case meta.TenantActive:
-	case meta.TenantPending:
-		logger.Warn(r.Context(), "server_event", eventFields(r.Context(), "tenant_pending", "tenant_id", resolved.Tenant.ID)...)
-		metricEvent(r.Context(), "tenant_status", "status", string(meta.TenantPending))
-		errJSONRetryable(w, "tenant provisioning is pending")
-		return
-	case meta.TenantProvisioning:
-		logger.Warn(r.Context(), "server_event", eventFields(r.Context(), "tenant_provisioning", "tenant_id", resolved.Tenant.ID)...)
-		metricEvent(r.Context(), "tenant_status", "status", string(meta.TenantProvisioning))
-		errJSONRetryable(w, "tenant is provisioning")
-		return
-	case meta.TenantFailed:
-		logger.Warn(r.Context(), "server_event", eventFields(r.Context(), "tenant_failed", "tenant_id", resolved.Tenant.ID)...)
-		metricEvent(r.Context(), "tenant_status", "status", string(meta.TenantFailed))
-		errJSON(w, http.StatusForbidden, "tenant provisioning failed")
-		return
+		case meta.TenantPending:
+			logger.Warn(r.Context(), "server_event", eventFields(r.Context(), "tenant_pending", "tenant_id", resolved.Tenant.ID)...)
+			metricEvent(r.Context(), "tenant_status", "status", string(meta.TenantPending))
+			errJSONRetryable(w, "tenant provisioning is pending")
+			return
+		case meta.TenantProvisioning:
+			logger.Warn(r.Context(), "server_event", eventFields(r.Context(), "tenant_provisioning", "tenant_id", resolved.Tenant.ID)...)
+			metricEvent(r.Context(), "tenant_status", "status", string(meta.TenantProvisioning))
+			errJSONRetryable(w, "tenant is provisioning")
+			return
+		case meta.TenantFailed:
+			logger.Warn(r.Context(), "server_event", eventFields(r.Context(), "tenant_failed", "tenant_id", resolved.Tenant.ID)...)
+			metricEvent(r.Context(), "tenant_status", "status", string(meta.TenantFailed))
+			errJSON(w, http.StatusForbidden, "tenant provisioning failed")
+			return
 		case meta.TenantSuspended, meta.TenantDeleting, meta.TenantDeleted:
 			pool.Invalidate(resolved.Tenant.ID)
 			logger.Warn(r.Context(), "server_event", eventFields(r.Context(), "tenant_blocked", "tenant_id", resolved.Tenant.ID, "status", resolved.Tenant.Status)...)
--- a/pkg/server/auth.go
+++ b/pkg/server/auth.go
@@ -306,21 +306,21 @@ func tenantAuthMiddlewareWithFSScopeLoader(metaStore *meta.Store, pool *tenant.P
 
 		switch resolved.Tenant.Status {
 		case meta.TenantActive:
-		case meta.TenantPending:
-			logger.Warn(r.Context(), "server_event", eventFields(r.Context(), "tenant_pending", "tenant_id", resolved.Tenant.ID)...)
-			metricEvent(r.Context(), "tenant_status", "status", string(meta.TenantPending))
-			errJSON(w, http.StatusServiceUnavailable, "tenant provisioning is pending")
-			return
-		case meta.TenantProvisioning:
-			logger.Warn(r.Context(), "server_event", eventFields(r.Context(), "tenant_provisioning", "tenant_id", resolved.Tenant.ID)...)
-			metricEvent(r.Context(), "tenant_status", "status", string(meta.TenantProvisioning))
-			errJSON(w, http.StatusServiceUnavailable, "tenant is provisioning")
-			return
-		case meta.TenantFailed:
-			logger.Warn(r.Context(), "server_event", eventFields(r.Context(), "tenant_failed", "tenant_id", resolved.Tenant.ID)...)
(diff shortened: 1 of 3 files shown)


pkg/server/auth.go
```diff
@@ -306,21 +306,21 @@ func tenantAuthMiddlewareWithFSScopeLoader(metaStore *meta.Store, pool *tenant.P
...
-			errJSON(w, http.StatusServiceUnavailable, "tenant is provisioning")
-			return
-		case meta.TenantFailed:
-			logger.Warn(r.Context(), "server_event", eventFields(r.Context(), "tenant_failed", "tenant_id", resolved.Tenant.ID)...)
-			metricEvent(r.Context(), "tenant_status", "status", string(meta.TenantFailed))
-			errJSON(w, http.StatusServiceUnavailable, "tenant provisioning failed")
-			return
+	case meta.TenantPending:
+		logger.Warn(r.Context(), "server_event", eventFields(r.Context(), "tenant_pending", "tenant_id", resolved.Tenant.ID)...)
+		metricEvent(r.Context(), "tenant_status", "status", string(meta.TenantPending))
```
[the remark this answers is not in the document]
srstack: Addressed in 9890914d: fixed the switch-block gofmt indentation and added TestAuthTenantStatusResponses covering 503+Retry-After for pending/provisioning (normal and DELETE /v1/tenant paths) and 403 for failed tenants.

pkg/server/auth.go
```diff
@@ -306,21 +306,21 @@ func tenantAuthMiddlewareWithFSScopeLoader(metaStore *meta.Store, pool *tenant.P
...
+	case meta.TenantProvisioning:
+		logger.Warn(r.Context(), "server_event", eventFields(r.Context(), "tenant_provisioning", "tenant_id", resolved.Tenant.ID)...)
+		metricEvent(r.Context(), "tenant_status", "status", string(meta.TenantProvisioning))
+		errJSONRetryable(w, "tenant is provisioning")
+		return
+	case meta.TenantFailed:
+		logger.Warn(r.Context(), "server_event", eventFields(r.Context(), "tenant_failed", "tenant_id", resolved.Tenant.ID)...)
+		metricEvent(r.Context(), "tenant_status", "status", string(meta.TenantFailed))
+		errJSON(w, http.StatusForbidden, "tenant provisioning failed")
+		return
```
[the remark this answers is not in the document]
srstack: Addressed in 9890914d: fixed the switch-block gofmt indentation and added TestAuthTenantStatusResponses covering 503+Retry-After for pending/provisioning (normal and DELETE /v1/tenant paths) and 403 for failed tenants.

srstack:
### Review at `9890914` — LGTM ✅

Small, focused, correct. Reviewed the full diff, verified build/vet/gofmt and the new test (6/6 pass).

**Correctness & semantics — sound:**
- `errJSONRetryable` sets `Retry-After` before `WriteHeader` (via `errJSON`), so the header lands on the response — verified by the test's positive assertion. Header ordering is correct.
- `TenantFailed → 403` is the right call and, notably, makes the switch *self-consistent*: the sibling `TenantSuspended/TenantDeleting/TenantDeleted` and the `default` branch already return `403 "tenant is unavailable"`. `failed` was the lone persistent-state outlier still returning 503. 403 for a persistent non-serviceable state is both more accurate than 503 and now uniform with its neighbors.
- Delete-path behavior is preserved intentionally: `handleDeleteNoAcquireAuth` keeps `TenantFailed` in the pass-through case (line 433) so a failed tenant can still be deleted, while only `pending`/`provisioning` get the retryable 503. The test pins both (`failed reaches handler` → 204; delete pending/provisioning → 503+Retry-After).

**Test quality — good:** table-driven, covers both middleware paths and both status changes, and includes the negative assertion (`failed`/handler paths must *not* carry `Retry-After`), which guards against the header leaking onto non-retryable responses.

**`[NIT]` (out of scope, optional follow-up):** `fork.go:791` returns `forkErr(503, "database branch is not active yet")` — semantically the same transient/retryable case as pending/provisioning, but via a different helper and without `Retry-After`. Not part of this PR's path and fine to leave; worth a future pass if you want fully uniform back-off hints. (`vault.go:604`'s 503 is a config error, correctly *not* retryable — leave as-is.)

**Verdict:** clean fix, accurate semantics, well-tested, no regressions. LGTM.

_(GitHub blocks formally approving one's own PR, so this comment stands in for the approval.)_

pkg/server/server.go
```diff
@@ -6246,6 +6246,11 @@ func errJSON(w http.ResponseWriter, code int, msg string) {
 	_ = json.NewEncoder(w).Encode(map[string]string{"error": msg})
 }
 
+func errJSONRetryable(w http.ResponseWriter, msg string) {
```
mornyx, line 6249: **[nit]** Clean helper. Setting the header before `errJSON` is correct (`errJSON` only sets Content-Type / status / body).

Fixed `Retry-After: 1` matches the free-quota-busy pattern elsewhere and is enough for clients to know the response is retryable. No change requested.

pkg/server/auth.go
```diff
@@ -309,17 +309,17 @@ func tenantAuthMiddlewareWithFSScopeLoader(metaStore *meta.Store, pool *tenant.P
...
 			logger.Warn(r.Context(), "server_event", eventFields(r.Context(), "tenant_provisioning", "tenant_id", resolved.Tenant.ID)...)
 			metricEvent(r.Context(), "tenant_status", "status", string(meta.TenantProvisioning))
-			errJSON(w, http.StatusServiceUnavailable, "tenant is provisioning")
+			errJSONRetryable(w, "tenant is provisioning")
 			return
 		case meta.TenantFailed:
 			logger.Warn(r.Context(), "server_event", eventFields(r.Context(), "tenant_failed", "tenant_id", resolved.Tenant.ID)...)
 			metricEvent(r.Context(), "tenant_status", "status", string(meta.TenantFailed))
-			errJSON(w, http.StatusServiceUnavailable, "tenant provisioning failed")
+			errJSON(w, http.StatusForbidden, "tenant provisioning failed")
```
mornyx, line 322: **[nit]** 403 for `TenantFailed` is the right semantic shift vs 503: this is a terminal provisioning state, not a transient blip. Aligns with `suspended`/`deleting`/`deleted` also returning forbidden-style unavailability.

Also correct that DELETE still allows failed tenants through `handleDeleteNoAcquireAuth` so cleanup remains possible.

pkg/server/auth_test.go
```diff
@@ -935,3 +935,54 @@ func TestAuthForkDeleteSkipsPoolAcquire(t *testing.T) {
 		t.Fatalf("status=%d", resp.StatusCode)
 	}
 }
+
+func TestAuthTenantStatusResponses(t *testing.T) {
```
mornyx, line 939: **[nit]** Good table coverage for both middleware paths. Optional: assert `Retry-After == "1"` when `wantRetryAfter` is true (not just non-empty), so a future change to a different value is intentional.

mornyx (approved):
## Review summary

**Verdict: Approve**

Reviewed in an isolated worktree at `9890914`. Small, well-scoped auth semantics fix aligned with `docs/design/agent-journal-design.md` (503s should carry `Retry-After` when retryable). Local `go test ./pkg/server -run TestAuthTenantStatusResponses` passed.

### Correctness

1. **`errJSONRetryable`** sets `Retry-After: 1` then 503 — both normal auth and `handleDeleteNoAcquireAuth` pending/provisioning paths use it.
2. **`TenantFailed` → 403** matches other terminal/blocked statuses (`suspended`/`deleting`/`deleted`). Persistent failure is no longer mislabeled as “temporarily unavailable.”
3. **Delete still works for failed tenants** — `handleDeleteNoAcquireAuth` keeps `TenantFailed` in the allow list; test asserts DELETE `/v1/tenant` reaches the handler with 204.
4. **`/v1/status` remains usable** for non-active tenants (separate auth path that returns `TenantStatusResponse`); clients can still observe `failed`/`provisioning` without going through the middleware hard-stop used by `/v1/fs`.
5. **Tests** cover GET pending/provisioning/failed and DELETE pending/provisioning/failed.

### Non-blocking notes

- Fixed `Retry-After: 1` is a simple fixed backoff (not adaptive). Fine for auth gates; operators/clients may still apply their own exponential backoff.
- Test only asserts presence/absence of the header, not the value `"1"`. Optional hardening.

No blockers. Approved.

PR #816, opened 2026-07-30
fix(fuse): snapshot write-back from authoritative shadow; fetch lazy layer buffers

### LINKED ISSUE #815: fix(fuse): lazy-buffer zero-fill hazards in layer-mode flush and ShadowSpill writeBack staging
## Bug Report

Follow-up from PR #812 review (qiffang, adversary-1, 2026-07-30): two remaining lazy-buffer zero-fill hazards in the write-back / shadow / layer durability subsystem. These are pre-existing and deliberately split out of #812, whose materialization-guard contract covers the main flush paths (`flushHandle` Path 2, `syncWriteHandleToRemoteLocked`) and the debounced path.

**Path A — layer-mode lazy handle loses retry state.**

1. Mount in layer mode; open an existing file whose write buffer is lazily loaded (some retained parts not in memory, `CanMaterializeFull() == false`).
2. Sparsely edit a loaded part, then flush: the layer branch of `flushHandle` returns `EIO` (`pkg/fuse/dat9fs.go:12514-12524`) instead of fetching missing parts via a loader or durable staging.
3. `Release` calls `flushHandle` and then clears `DirtySeq` regardless of status (`pkg/fuse/dat9fs.go:12253-12259`), so the recoverable edit loses its retry source once the handle closes.

**Path B — ShadowSpill eviction feeds a zero-filled writeBack snapshot.**

1. Open a file that becomes `ShadowSpill` (large-file spill; parts can be evicted under memory pressure) and reach `ShadowReady`.
2. Shrink/write so `size < writeBackThreshold` with no active stream, then flush: the small writeBack staging guard (`pkg/fuse/dat9fs.go:11297-11320`) passes via `canStageShadowFastLocked` (which exempts `ShadowReady`/`IsNew` from `CanMaterializeFull`).
3. `snapshotWriteBackLocked` (`pkg/fuse/dat9fs.go:3325-3353`) also skips the materialization check for `ShadowReady`/`IsNew` and snapshots `fh.Dirty.bytesView()` — for an evicted ShadowSpill buffer this zero-fills evicted parts, while the authoritative bytes live in the shadow store. The staged writeBack `.dat` and the commit-queue upload body become zero-filled.

- Path A: a layer-mode lazy handle must either fetch/materialize under a pinned base revision and upsert the layer bytes, or fail while preserving durable retry state through `Release` (no unconditional `DirtySeq` drop on flush failure).
- Path B: ShadowSpill small-file flush/write-back must never snapshot from `Dirty.bytesView()` when parts are evicted; it must read the authoritative bytes from the shadow store (or skip writeBack and commit from shadow), including the commitQueue enqueue-failure/fallback behavior.

- Path A: `EIO` without a fetch attempt, then retry state dropped on `Release`.
- Path B: zero-filled staged snapshot / upload body for evicted ShadowSpill buffers (silent data corruption).

`main @ 80ed57b` (still present on PR #812 head `b2f1871`).

### Notes

- Both paths need regressions alongside the fix, in the style of the #812 materialization tests (`TestFlushDirectPUTLazyBuffer*`, `TestFlushHandleDebouncedLazy*`): Path A — layer-mode lazy flush must fetch or preserve retry state through Release; Path B — ShadowSpill evicted buffer must stage/upload from shadow bytes, never a zero-filled `bytesView()`.
- Related: PR #812 (PATCH storage-class routing + full-upload materialization guards).

mornyx:
## Background

Follow-up to #812 (merged). Fixes #815 — the two lazy-buffer zero-fill hazards qiffang identified in the write-back / shadow / layer durability subsystem, deliberately split out of #812 — and closes the `suite-manifest.json` metadata gap from his last #812 review.

## Changes

**Path B — ShadowSpill writeBack snapshot zero-fill (`snapshotWriteBackLocked`)**
- Previously, `ShadowReady`/`IsNew` handles were exempt from the `CanMaterializeFull` check and the snapshot came from `fh.Dirty.bytesView()`. For a spill-evicted buffer this zero-filled evicted ranges into the write-back cache — poisoning subsequent reads and any read-modify-write or commit-fallback that consumed the staged snapshot — while the authoritative bytes sat in the shadow store.
- Now: when the buffer cannot materialize and `ShadowReady` with a shadow present, the snapshot reads `shadowStore.ReadAll` (ShadowSpill/ShadowReady handles write through to the shadow on every write, so it is the authoritative full content). Any other non-materializable buffer fails `ENOTSUP` instead of zero-filling, letting the caller fall back to a guarded synchronous upload. A shadow/dirty size divergence fails the snapshot instead of staging suspect bytes.

**Path A — layer-mode lazy handle flush (`flushHandle` layer branch)**
- Previously returned `EIO` for `!CanMaterializeFull()` without a fetch attempt, so a recoverable lazy sparse edit failed at flush and was dropped on Release.
- Now runs `materializeFullForUploadLocked`: fetches missing remote-backed parts via the pinned lazy loader (or `EIO` when truly unmaterializable, unchanged) before upserting the complete layer entry.

**e2e metadata gap from #812's last review**
- `fuse-patch-storage-class` added to `e2e/suite-manifest.json`, and `failure_class: correctness` backfilled for the six suites that lacked one (`api-smoke`, `cli-smoke`, `layer-fs-smoke`, `fuse-smoke`, `pack-smoke`, `smoke-all`) — the aggregator previously synthesized `class=unknown` for them too.
- New sync regression `internal/e2ereport/manifest_sync_test.go`: every suite id emitted into `local-e2e.yml`'s outcomes JSON must exist in the manifest with `product_promise`/`owner_hint`, and a failure outcome must synthesize a fully classified failure through the real `SynthesizeSummaries` path. Verified end-to-end with `cmd/e2e-aggregate` (the new suite now reports `class=correctness, promise=…, owner=fs`).

## Regressions

- `TestFlushHandleLayerLazyBufferFetchesAndUpserts` — layer mode, lazy two-part buffer with loader: untouched part fetched, inline layer entry content byte-exact, `base_revision` pinned, dirty cleared. Fails against the old EIO branch.
- `TestSnapshotWriteBackShadowSpillUsesAuthoritativeShadow` — spill-evicted buffer: staged write-back content equals the authoritative shadow bytes, not a zero-filled `bytesView()`. Fails against the old snapshot path.

`make lint` 0 issues; gofmt clean; full `pkg/fuse` + `pkg/client` + `internal/e2ereport` suites pass (the two `TestShouldContinueAfterWaitMountPermissionError*` failures reproduce on clean main on macOS — pre-existing portability gap, unrelated).

Fixes #815

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented stale file updates from overwriting newer revisions.
  * Improved write-back handling for large and remotely backed files, preserving complete file contents.
  * Ensured revision conflicts return a clear `409 Conflict` response.
  * Preserved pending changes when uploads encounter revision conflicts.

* **Tests**
  * Added coverage for revision conflicts, lazy-loaded file updates, write-back snapshots, and end-to-end outcome classification.
  * Updated test metadata to consistently classify suite failures and identify ownership.

### DIFF
--- a/e2e/suite-manifest.json
+++ b/e2e/suite-manifest.json
@@ -5,25 +5,36 @@
       "title": "API smoke",
       "product_area": "api",
       "product_promise": "HTTP API works end to end",
-      "owner_hint": "api"
+      "owner_hint": "api",
+      "failure_class": "correctness"
     },
     "cli-smoke": {
       "title": "CLI smoke",
       "product_area": "cli",
       "product_promise": "drive9 CLI works end to end",
-      "owner_hint": "cli"
+      "owner_hint": "cli",
+      "failure_class": "correctness"
     },
     "layer-fs-smoke": {
       "title": "Layer FS smoke",
       "product_area": "filesystem",
       "product_promise": "layered filesystem reads/writes are correct",
-      "owner_hint": "fs"
+      "owner_hint": "fs",
+      "failure_class": "correctness"
     },
     "fuse-smoke": {
       "title": "FUSE release gate",
       "product_area": "fuse",
       "product_promise": "FUSE mount works for release",
-      "owner_hint": "fs"
+      "owner_hint": "fs",
+      "failure_class": "correctness"
+    },
+    "fuse-patch-storage-class": {
+      "title": "FUSE patch storage-class e2e",
+      "product_area": "fuse",
+      "product_promise": "FUSE partial updates route by storage class and recover from unsupported PATCH targets",
+      "owner_hint": "fs",
+      "failure_class": "correctness"
     },
     "git-ops-smoke": {
       "title": "Git operations smoke",
@@ -36,7 +47,8 @@
       "title": "Pack smoke",
       "product_area": "portability",
       "product_promise": "portable pack/unpack round-trips losslessly",
-      "owner_hint": "fs"
+      "owner_hint": "fs",
+      "failure_class": "correctness"
     },
     "fuse-crash-recovery": {
       "title": "FUSE crash recovery e2e",
@@ -77,7 +89,8 @@
       "title": "Full smoke-all suite",
       "product_area": "platform",
       "product_promise": "the full smoke suite passes",
-      "owner_hint": "platform"
+      "owner_hint": "platform",
+      "failure_class": "correctness"
     },
     "git-feature-smoke": {
       "title": "Git feature smoke",
--- a/internal/e2ereport/manifest_sync_test.go
+++ b/internal/e2ereport/manifest_sync_test.go
@@ -1,0 +1,57 @@
+package e2ereport
+
+import (
+	"os"
+	"regexp"
+	"testing"
+)
+
+// TestSuiteManifestCoversWorkflowOutcomes pins the sync contract between
+// .github/workflows/local-e2e.yml and e2e/suite-manifest.json: every suite id
+// the workflow emits into /tmp/e2e-outcomes.json must carry product metadata,
+// and a failure outcome for it must synthesize a classified failure through
+// the real aggregator path. Without this, a newly added step id silently
+// degrades to class=unknown / promise=- / owner=- in nightly owner routing.
+func TestSuiteManifestCoversWorkflowOutcomes(t *testing.T) {
+	workflow, err := os.ReadFile("../../.github/workflows/local-e2e.yml")
+	if err != nil {
+		t.Fatalf("read workflow: %v", err)
+	}
+	manifestData, err := os.ReadFile("../../e2e/suite-manifest.json")
+	if err != nil {
+		t.Fatalf("read manifest: %v", err)
+	}
+	manifest, err := LoadManifest(manifestData)
+	if err != nil {
+		t.Fatalf("load manifest: %v", err)
+	}
+
+	idRe := regexp.MustCompile(`"([a-z0-9_-]+)": "\$\{\{ steps\.[a-z0-9_-]+\.outcome \}\}"`)
+	ids := idRe.FindAllStringSubmatch(string(workflow), -1)
+	if len(ids) == 0 {
+		t.Fatal("no step outcome ids found in workflow — extraction pattern broken?")
+	}
+
+	for _, match := range ids {
+		id := match[1]
+		meta, ok := manifest.Suites[id]
+		if !ok {
+			t.Errorf("workflow outcome id %q missing from e2e/suite-manifest.json", id)
+			continue
+		}
+		if meta.ProductPromise == "" || meta.OwnerHint == "" {
+			t.Errorf("manifest entry %q lacks product_promise/owner_hint: %+v", id, meta)
+		}
+
+		// A failure outcome for this suite must synthesize a fully classified
+		// failure through the real aggregator path.
+		summaries := SynthesizeSummaries(manifest, TierNightly, map[string]string{id: "failure"}, nil)
+		if len(summaries) != 1 {
+			t.Fatalf("synthesize %q: got %d summaries, want 1", id, len(summaries))
+		}
+		s := summaries[0]
+		if s.ProductPromise == "" || s.OwnerHint == "" || s.FailureClass == FailureNone || s.FailureClass == FailureUnknown {
+			t.Errorf("synthesized failure for %q is not fully classified: %+v", id, s)
+		}
+	}
+}
--- a/pkg/datastore/fs_layer.go
+++ b/pkg/datastore/fs_layer.go
@@ -588,6 +588,31 @@ func (s *Store) UpsertFSLayerEntry(ctx context.Context, entry *FSLayerEntry) err
 	if err := validateFSLayerEntryWithinBaseRoot(entry, baseRoot); err != nil {
 		return err
 	}
+	// Synchronous base-revision CAS for file content upserts. Without this
+	// gate, an upsert based on a stale base revision is persisted silently
+	// (e.g. a FUSE lazy flush that fetched retained parts after a sibling
+	// commit): the spliced bytes enter the layer, the writer's dirty state
+	// is cleared, and the mismatch only surfaces at layer-commit preflight —
+	// escalating the whole layer to Conflicted. Reject here, while the
+	// writer can still retry. base_revision <= 0 means "no base claim"
+	// (new or layer-local file) and skips the check, matching the commit
+	// preflight's semantics.
+	if entry.Op == FSLayerEntryOpUpsert && entry.Kind == FSLayerEntryKindFile && entry.BaseRevision > 0 {
+		bn, statErr := s.StatTx(ctx, tx, entry.Path)
+		if errors.Is(statErr, ErrNotFound) {
+			return fmt.Errorf("%w: base file %s missing for base_revision=%d", ErrRevisionConflict, entry.Path, entry.BaseRevision)
+		}
+		if statErr != nil {
+			return fmt.Errorf("stat base file for layer upsert %s: %w", entry.Path, statErr)
+		}
+		var currentRev int64
+		if bn.File != nil {
+			currentRev = bn.File.Revision
+		}
+		if bn.File == nil || currentRev != entry.BaseRevision {
+			return fmt.Errorf("%w: base revision changed for %s (current=%d, want=%d)", ErrRevisionConflict, entry.Path, currentRev, entry.BaseRevision)
+		}
+	}
 	if entry.EntrySeq <= 0 {
 		var seq sql.NullInt64
 		if err := tx.QueryRowContext(ctx, `SELECT MAX(entry_seq) FROM fs_layer_entries WHERE `+s.scope.And(`layer_id = ?`), s.scope.Args(entry.LayerID)...).Scan(&seq); err != nil {
--- a/pkg/fuse/dat9fs.go
+++ b/pkg/fuse/dat9fs.go
@@ -3350,29 +3350,32 @@ func (fs *Dat9FS) snapshotWriteBackLocked(fh *FileHandle) error {
 
 	bvStart := time.Now()
 	var data []byte
-	if !fh.Dirty.CanMaterializeFull() {
-		// The buffer has parts that are not loaded (evicted under spill or
-		// never lazily loaded): bytesView() would zero-fill those ranges and
-		// the write-back cache would persist corrupted bytes. ShadowSpill and
-		// ShadowReady handles write through to the shadow store on every
-		// write, so the shadow holds the authoritative full content —
-		// snapshot from there. Any other non-materializable buffer has no
-		// authoritative secondary source: fail instead of zero-filling, and
-		// let the caller fall back to a guarded synchronous upload.
-		if fh.ShadowReady && fs.shadowStore != nil && fs.shadowStore.Has(fh.Path) {
-			shadowData, err := fs.shadowStore.ReadAll(fh.Path)
-			if err != nil {
-				return err
-			}
-			if size := fh.Dirty.Size(); int64(len(shadowData)) != size {
-				// Writes hit the shadow first and truncates are mirrored, so
-				// sizes must agree; treat any divergence as unsafe to stage.
-				return fmt.Errorf("shadow size %d != dirty size %d for %s", len(shadowData), size, fh.Path)
-			}
-			data = shadowData
-		} else {
-			return syscall.ENOTSUP
+	// ShadowSpill handles write through to the shadow store on every write,
+	// so the shadow — never the possibly evicted buffer — is the
+	// authoritative full content. Snapshot from it unconditionally: for
+	// newly created spill files CanMaterializeFull is vacuously true
+	// (remoteSize == 0) even after OnPartFull evicted part bytes from
+	// memory, and bytesView() would zero-fill those ranges.
+	// For non-spill buffers the shadow read is only needed when the buffer
+	// cannot materialize; any other non-materializable buffer has no
+	// authoritative secondary source and fails ENOTSUP instead of
+	// zero-filling, letting the caller fall back to a guarded upload.
+	fromShadow := fh.ShadowReady && fs.shadowStore != nil && fs.shadowStore.Has(fh.Path) &&
+		(fh.ShadowSpill || !fh.Dirty.CanMaterializeFull())
+	switch {
+	case fromShadow:
+		shadowData, err := fs.shadowStore.ReadAll(fh.Path)
+		if err != nil {
+			return err
 		}
+		if size := fh.Dirty.Size(); int64(len(shadowData)) != size {
+			// Writes hit the shadow first and truncates are mirrored, so
+			// sizes must agree; treat any divergence as unsafe to stage.
+			return fmt.Errorf("shadow size %d != dirty size %d for %s", len(shadowData), size, fh.Path)
+		}
+		data = shadowData
+	case !fh.Dirty.CanMaterializeFull():
+		return syscall.ENOTSUP
 	}
 	if data == nil {
 		data = fh.Dirty.bytesView()
@@ -11339,11 +11342,20 @@ func (fs *Dat9FS) Flush(cancel <-chan struct{}, input *gofuse.FlushIn) (status g
 					} else {
 						phase = "small-snapshot-writeback"
 						snapWBStart := time.Now()
-						if err := fs.snapshotWriteBackLocked(fh); err != nil && fs.writeBack != nil {
+						// Advance WriteBackSeq only when the snapshot actually
+						// landed: the sequence gates the writeBack cache as an
+						// upload/read source, so claiming freshness after a
+						// skipped/failed snapshot could serve a stale .dat as
+						// current. On failure the next Flush/Release re-attempts
+						// the snapshot or falls back to the synchronous flush —
+						// the staged shadow + pendingIndex already carry the
+						// durable commit path.
+						if err := fs.snapshotWriteBackLocked(fh); err != nil {
 							log.Printf("writeback snapshot failed for %s: %v", fh.Path, err)
+						} else {
+							fh.WriteBackSeq = fh.DirtySeq
 						}
 						fs.perf.recordFlushSnapshotWB(time.Since(snapWBStart))
-						fh.WriteBackSeq = fh.DirtySeq
 						return gofuse.OK
 					}
 				}
@@ -11485,10 +11497,13 @@ func (fs *Dat9FS) Flush(cancel <-chan struct{}, input *gofuse.FlushIn) (status g
 					log.Printf("flush: shadow stage failed for %s (size=%d): %v, falling through to sync upload", fh.Path, fh.Dirty.Size(), err)
 				} else {
 					phase = "large-snapshot-writeback"
-					if err := fs.snapshotWriteBackLocked(fh); err != nil && fs.writeBack != nil {
+					if err := fs.snapshotWriteBackLocked(fh); err != nil {
 						log.Printf("flush: writeback snapshot failed for %s: %v", fh.Path, err)
+					} else {
+						// See the small-snapshot-writeback path: only claim a
+						// current cache when the snapshot actually landed.
+						fh.WriteBackSeq = fh.DirtySeq
 					}
-					fh.WriteBackSeq = fh.DirtySeq
 					// If a streaming upload was already in flight, abandon it:
 					// the CommitQueue (driven by Release via the cache fast
 					// path) will read from the shadow file instead. Without
@@ -11651,10 +11666,13 @@ func (fs *Dat9FS) Fsync(cancel <-chan struct{}, input *gofuse.FsyncIn) (status g
 					}
 				} else {
 					fs.releaseHandleRemoteCommitPathLocked(fh)
-					if err := fs.snapshotWriteBackLocked(fh); err != nil && fs.writeBack != nil {
+					if err := fs.snapshotWriteBackLocked(fh); err != nil {
 						log.Printf("fsync writeback snapshot failed for %s: %v", fh.Path, err)
+					} else {
+						// See the small-snapshot-writeback path: only claim a
+						// current cache when the snapshot actually landed.
+						fh.WriteBackSeq = fh.DirtySeq
 					}
-					fh.WriteBackSeq = fh.DirtySeq
 				}
 				return gofuse.OK
 			}
@@ -12539,6 +12557,14 @@ func (fs *Dat9FS) flushHandle(ctx context.Context, fh *FileHandle) (status gofus
 
 	size := fh.Dirty.Size()
 	if fs.layerEnabled() {
+		// Freeze the upsert base BEFORE any lazy fetch. The layer branch
+		// deliberately reads the un-adopted BaseRev (expectedRevisionForHandle,
+		// not the Locked variant) and holds fh.mu for the whole branch, so a
+		// sibling commit mid-fetch cannot advance the base — the layer upsert
+		// always goes out with base M and the server-side CAS is the backstop
+		// against splicing older dirty bytes into a newer remote. Keep this
+		// capture ahead of materializeFullForUploadLocked.
+		expectedRevision := expectedRevisionForHandle(fh)
 		if fh.ShadowSpill || (fh.Streamer != nil && fh.Streamer.Started()) || !fs.materializeFullForUploadLocked(fh) {
 			if fh.ShadowSpill {
 				if err := fs.commitLayerShadowLocked(ctx, fh, true); err != nil {
@@ -12552,7 +12578,6 @@ func (fs *Dat9FS) flushHandle(ctx context.Context, fh *FileHandle) (status gofus
 		}
 		data := fh.Dirty.bytesView()
 		mode, hasMode := fs.modeForPendingHandle(fh)
-		expectedRevision := expectedRevisionForHandle(fh)
 		if err := fs.upsertLayerFile(ctx, fh.Path, data, expectedRevision, mode, hasMode); err != nil {
 			log.Printf("layer flush failed for %s: %v", fh.Path, err)
 			return httpToFuseStatus(err)
--- a/pkg/fuse/dat9fs.go
+++ b/pkg/fuse/dat9fs.go
@@ -3349,7 +3349,34 @@ func (fs *Dat9FS) snapshotWriteBackLocked(fh *FileHandle) error {
 	mode, hasMode := fs.modeForPendingHandle(fh)
 
 	bvStart := time.Now()
-	data := fh.Dirty.bytesView()
+	var data []byte
+	if !fh.Dirty.CanMaterializeFull() {
+		// The buffer has parts that are not loaded (evicted under spill or
+		// never lazily loaded): bytesView() would zero-fill those ranges and
+		// the write-back cache would persist corrupted bytes. ShadowSpill and
+		// ShadowReady handles write through to the shadow store on every
+		// write, so the shadow holds the authoritative full content —
+		// snapshot from there. Any other non-materializable buffer has no
+		// authoritative secondary source: fail instead of zero-filling, and
+		// let the caller fall back to a guarded synchronous upload.
+		if fh.ShadowReady && fs.shadowStore != nil && fs.shadowStore.Has(fh.Path) {
+			shadowData, err := fs.shadowStore.ReadAll(fh.Path)
+			if err != nil {
+				return err
+			}
+			if size := fh.Dirty.Size(); int64(len(shadowData)) != size {
+				// Writes hit the shadow first and truncates are mirrored, so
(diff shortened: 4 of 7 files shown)


qiffang (commented):
## Dev-Lead review (dat9-dev1) — exact head `3a3822180b5499768670d85b44117464277bc1a2`

Reviewed the #816-only commit `3a38221` (isolated from the merged-#812 content). **Both #815 hazards + the #812 manifest gap are addressed correctly, with discriminating regressions I verified by revert.** LGTM on my axis, one small confirmation request below.

### ✅ Path B — `snapshotWriteBackLocked` (ShadowSpill zero-fill)
`dat9fs.go:3349+`: when `!CanMaterializeFull()`, a `ShadowReady` handle with a present shadow now snapshots from `shadowStore.ReadAll` (authoritative), and any other non-materializable buffer returns `ENOTSUP` instead of `bytesView()` zero-filling. The `len(shadowData) != Dirty.Size()` divergence guard is a good defensive check. This closes the confirmed data-corruption path (evicted part → zero-filled `.dat` → zero-body PUT).

### ✅ Path A — layer branch `flushHandle`
`dat9fs.go:~12539`: `!fh.Dirty.CanMaterializeFull()` → `!fs.materializeFullForUploadLocked(fh)`. A lazy sparse layer edit now fetches missing remote parts (pinned loader) before upsert instead of `EIO`-dropping on Release.

### ✅ e2e manifest gap (#812 follow-up)
`fuse-patch-storage-class` added (promise + `owner_hint=fs` + `failure_class=correctness`); `failure_class` backfilled for the 6 previously-unclassified suites. New `internal/e2ereport/manifest_sync_test.go::TestSuiteManifestCoversWorkflowOutcomes` asserts every workflow outcome id is in the manifest with promise/owner AND synthesizes a fully-classified failure through the real `SynthesizeSummaries` path — this prevents the exact #812 recurrence.

### Discriminating-regression proof (I ran the revert)
- `TestFlushHandleLayerLazyBufferFetchesAndUpserts` — on the pre-fix parent: **FAIL** "layer flush status = EIO, want OK (lazy buffer must be fetched, not EIO)"; on the fix: PASS.
- `TestSnapshotWriteBackShadowSpillUsesAuthoritativeShadow` — pre-fix: **FAIL** "snapshot mismatch — must come from the authoritative shadow, not zero-filled bytesView"; on the fix: PASS. Test exercises `ShadowSpill:true, ShadowReady:true` + `EvictPart(1)`, so it hits the shadow-read path (not just ENOTSUP).
- `internal/e2ereport` suite PASS; broad `pkg/fuse` green except the two known `TestShouldContinueAfterWaitMountPermissionError*` Darwin `mount_wait` env fails (pre-existing, unrelated).

### One confirmation (non-blocking) — ENOTSUP fallback at the first caller
`dat9fs.go:~11342`:
```go
if err := fs.snapshotWriteBackLocked(fh); err != nil && fs.writeBack != nil {
    log.Printf("writeback snapshot failed for %s: %v", fh.Path, err)
}
fs.perf.recordFlushSnapshotWB(...)
fh.WriteBackSeq = fh.DirtySeq   // advanced even when the snapshot returned ENOTSUP
return gofuse.OK
```
On `ENOTSUP` (non-materializable, not-ShadowReady) the write-back stage is correctly skipped (no zero-fill ✅), but `WriteBackSeq` is still advanced to `DirtySeq` and the caller returns OK. Please confirm the authoritative bytes still reach remote via the commit-queue/pending path for this branch (the buffer stays in shadow/Dirty), and that advancing `WriteBackSeq` after a skipped snapshot can't make a later reader treat a stale/absent write-back stage as current. The second caller (`~11353`) has the explicit "falling back to sync upload" continuation; this first one just returns OK, so I want to confirm the fallback is equivalent here.

**Net: on my correctness axis this closes #815 A/B + the manifest gap, LGTM pending the one ENOTSUP-fallback confirmation.** (drive9 is qiffang-owned — merge on his signal.)

qiffang:
## adversary-2 review @ `3a382218` — **REVISE (1 blocking finding)**

Independent adversarial review + source-verified. 3 of the 4 changes are correct; **Path B does not close the primary #815 scenario.**

### ✅ Path A (layer lazy-buffer fetch) — CONFIRMS
`flushHandle` layer branch (`dat9fs.go:12542`) now `fh.ShadowSpill || streamerStarted || !materializeFullForUploadLocked(fh)`. On `!materialize` for a lazy non-spill handle it fetches via the pinned loader and returns `EIO` **without** `ClearDirty` (success-only at 12560-12564) → dirty preserved for retry, no zero-fill. `TestFlushHandleLayerLazyBufferFetchesAndUpserts` discriminates (asserts `OK` + `loadCalls>0` + preserved untouched range + pinned `base_revision` + `!HasDirtyParts()`; fails on the old bare-EIO branch). Matches the #812 materialize contract.

### ✅ manifest fix — CONFIRMS (genuine regression lock)
`suite-manifest.json` adds `fuse-patch-storage-class` + backfills `failure_class` for the 6 suites. `manifest_sync_test.go` uses the REAL `LoadManifest`+`SynthesizeSummaries` (not a mock), extracts all 15 workflow outcome-ids from `local-e2e.yml`, and asserts each has promise/owner + synthesizes a failure with `FailureClass ∉ {None, Unknown}`. Pre-fix the 6 suites deserialize to `FailureNone` → `pickFailureClass`→`Unknown` → `t.Errorf`; `fuse-patch-storage-class` absent → errors. **Fails pre-fix, passes at head.** Closes the #812 desync gap.

### ✅ ENOTSUP fallback / WriteBackSeq (dev1's non-blocking Q) — CONFIRMS-SAFE
The writeBack snapshot is a latency fast-path; durability is independent. All three snapshot callers (`11342/11488/11654`) run only **after** a successful `stageShadowForQueuedCommitLocked` (shadow bytes + pendingIndex). Release uploads via `useCommitQueue` (gated on `shadowStore.Has`, guaranteed true post-staging), and the CommitQueue reads its body from the **shadow store**, never the writeBack cache. So a skipped/failed snapshot never loses data, and no `WriteBackSeq==DirtySeq` consumer serves-from-remote / evicts without an independent shadow check. Advancing `WriteBackSeq` on ENOTSUP is safe.

### ❌ Path B (snapshot from authoritative shadow) — **BLOCKING: incomplete, new-file case still zero-fills**

The fix closes the **existing-file** ShadowSpill eviction but **not the new-file (`remoteSize==0`) case — which is the primary #815 Path B scenario** (new large file via `cp`/`dd`).

Root cause — `CanMaterializeFull()` (`write.go:829-847`) returns **TRUE** for a new file regardless of eviction:
```go
covered := wb.remoteSize      // == 0 for a new file
if wb.totalSize < covered { covered = wb.totalSize }
if covered <= 0 { return true }   // ← returns true even with an evicted part
```
(the docstring itself: *"This is true for new files (remoteSize == 0)"*.)

Trace in `snapshotWriteBackLocked` (`dat9fs.go:3325+`) for a new-file ShadowSpill handle (`IsNew=true, ShadowReady=true, ShadowSpill=true, remoteSize=0`, part 0 evicted):
- L3333 early gate `!ShadowReady && !IsNew && !CanMaterializeFull()` → `false` → does **not** ENOTSUP; proceeds.
- **L3353 `if !fh.Dirty.CanMaterializeFull()` → `!true` → false → the shadow-first branch is SKIPPED.**
- L3377 `if data == nil { data = fh.Dirty.bytesView() }` → evicted part 0 absent from `wb.parts` → **left as zeros → corrupted snapshot persisted → 8MiB of zeros on the commit wire.**

The shadow store *does* hold the correct bytes (every ShadowSpill Write writes-through to shadow first, `dat9fs.go:10778-10796`), so the shadow-first branch **would** fix it — the `!CanMaterializeFull()` gate just never lets it run for a new file.

**Reachability (not hypothetical):** `Create` (non-write-sync, shadow enabled) sets `IsNew=true, ShadowReady=true, ShadowSpill=true` and wires `wb.OnPartFull = wb.EvictPart` with **no** Streamer (`dat9fs.go:9667-9678`); `remoteSize` stays 0. Writing ≥`DefaultPartSize` (8MiB) fills+evicts part 0 while `size < writeBackThreshold` (10MiB) keeps the flush on the write-back staging branch → `snapshotWriteBackLocked` at `11342` → zero-fill.

**The regression masks it:** `TestSnapshotWriteBackShadowSpillUsesAuthoritativeShadow` sets `wb.remoteSize = 512` (existing file, part 1 evicted) — the *existing-file* branch where `CanMaterializeFull()` is correctly false and the fix engages. It never exercises `remoteSize==0`, so a fix that only covers existing files passes green.

**Required to GREEN:**
1. Broaden the L3353 gate to route to shadow whenever the buffer has evicted/unloaded parts **regardless of `remoteSize`** — e.g. `!CanMaterializeFull() || fh.Dirty.HasEvictedParts()` (or check `StreamedPartIndices()`/loaded-vs-total) so a new-file ShadowSpill with an evicted part reads from the authoritative shadow instead of `bytesView()`.
2. Add a `remoteSize==0` + evicted-part regression (new-file ShadowSpill, ≥8MiB, part 0 evicted → assert staged bytes == shadow, **not** zero-filled).

Net: Path A + manifest + WriteBackSeq clean; **Path B REVISE** until the new-file ShadowSpill case reads from the shadow.

qiffang (changes requested):
Reviewing exact head `3a3822180b5499768670d85b44117464277bc1a2`: REQUEST_CHANGES.

Blocking findings:

1. `snapshotWriteBackLocked` still zero-fills new ShadowSpill files with evicted parts. PR816 only switches to the authoritative shadow when `!fh.Dirty.CanMaterializeFull()` is true (`pkg/fuse/dat9fs.go:3353-3379`). For newly-created ShadowSpill files, `remoteSize` remains 0, and `WriteBuffer.CanMaterializeFull()` returns true for `covered <= 0` (`pkg/fuse/write.go:825-839`) even after `EvictPart` has removed full-part bytes from memory (`pkg/fuse/write.go:942-956`). That means a new file between `DefaultPartSize` and `writeBackThreshold` can still take the small write-back snapshot path: create wires `ShadowReady=true`/`ShadowSpill=true` and `OnPartFull -> EvictPart` (`pkg/fuse/dat9fs.go:9650-9678`), writes hit the shadow first then dirty (`pkg/fuse/dat9fs.go:10778-10798`), but snapshot then falls through to `fh.Dirty.bytesView()`. `Bytes()` allocates the full output and simply skips missing parts, leaving those ranges zeroed (`pkg/fuse/write.go:649-668`). The added regression uses `wb.remoteSize = 512` before eviction (`pkg/fuse/dat9fs_test.go:17319-17327`), so it exercises existing-file semantics where `CanMaterializeFull()` is false; it does not cover the new-file path that triggered this class.

Required gate: add a regression for a new ShadowSpill/write-back file that writes at least one full part and remains below `writeBackThreshold`, evicts that part, then calls `snapshotWriteBackLocked`; the staged write-back content must come from the shadow, not zero-filled dirty-buffer bytes.

2. The new layer lazy materialization path is not revision-anchored. `flushHandle` now calls `fs.materializeFullForUploadLocked(fh)` before it captures `expectedRevision := expectedRevisionForHandle(fh)` in the layer branch (`pkg/fuse/dat9fs.go:12541-12556`). That materializer calls `LoadMissingParts()` (`pkg/fuse/dat9fs.go:269-289`), and the production lazy loaders fetch by current path using `ReadStreamRange` with no revision pin (`pkg/fuse/dat9fs.go:2429-2490`, `pkg/fuse/dat9fs.go:2094-2144`; client API at `pkg/client/transfer.go:1305-1325`). This is the same invariant PR812 already fixed for debounced uploads: freeze the CAS/base revision before any lazy fetch, because the fetch can observe a sibling’s newer remote bytes. The current PR’s layer test uses a synthetic stable loader and only asserts the happy-path `base_revision=5` (`pkg/fuse/dat9fs_test.go:17248-17269`), so it would pass even if a real mount splices old dirty bytes with newer retained bytes during `LoadMissingParts`.

Required gate: add a layer-mode regression where a lazy handle is dirty at base rev M, a sibling same-path commit is recorded/visible while missing parts are fetched, and flush must not upsert a full layer entry containing old dirty bytes plus newer retained bytes. Either the retained ranges must be pinned to rev M and the layer entry records base M, or the flush must fail/skip without clearing dirty state. A single-revision happy path is not sufficient for this data-integrity gate.

Validation I ran locally on this head:

- `git diff --check d9509cffff858c3d8b0e6997195dbad1d106563a...HEAD`
- `gofmt -d pkg/fuse/dat9fs.go pkg/fuse/dat9fs_test.go internal/e2ereport/manifest_sync_test.go`
- `go test ./internal/e2ereport -run TestSuiteManifestCoversWorkflowOutcomes -count=1`
- `go test ./pkg/fuse -run 'TestWriteBufferCanMaterializeFull|TestWriteBuffer_EvictPart|TestShadowSpill_WriteOrderShadowFirst|TestFlushHandleLayerLazyBufferFetchesAndUpserts|TestSnapshotWriteBackShadowSpillUsesAuthoritativeShadow|TestFlushHandleDebouncedLazyFetchNewerRevisionSkipsUpload' -count=1`

Those passed, but they do not cover the two blockers above. Hosted `ci` is green; `drive9-server-local e2e smoke` is still pending at the time of review. Do not merge PR816 until these are fixed on a new exact head.

qiffang:
## adversary-2 addendum @ `3a382218` — on @adversary-1's 2nd blocker (Path A layer cross-revision splice): **CAS-backstopped, not a splice-commit — downgrade to missing-regression**

Traced the closed loop at source (same method as #812's resolved P1). adversary-1 is right that the layer loader is unpinned and `expectedRevision` is captured *after* the fetch — but the splice can only be **materialized in memory, never committed**, because the revision-CAS rejects it. Net: still REVISE, but this specific item is a **missing discriminating test**, not a data-corruption blocker.

### 1. Continuous lock across materialize→capture→upsert — CONFIRMED (no unlock)
`flushHandle` layer branch (`dat9fs.go:12541-12581`) has **no** `fh.Unlock`/`fh.Lock`/adopt between:
- `materializeFullForUploadLocked(fh)` fetch (`12542`)
- `data := fh.Dirty.bytesView()` (`12553`)
- `expectedRevision := expectedRevisionForHandle(fh)` (`12555`)
- `upsertLayerFile(..., expectedRevision, ...)` (`12556`)

Caller holds `fh.mu` throughout (Release `fh.Lock()` @ `12269` → `flushHandle` with no intervening unlock). Unlike the non-layer Path (which drops `remoteCommitLock`), the layer branch drops nothing.

### 2. `fh.BaseRev` is FROZEN between fetch and capture
`BaseRev` is advanced only by `adoptCommittedRevisionLocked` (`dat9fs.go:2322` `fh.BaseRev = revision`). It cannot fire on this path:
- The layer branch captures via `expectedRevisionForHandle` (`3025`, **pure read**: IsNew→0, BaseRev>0→BaseRev, else -1) — deliberately **not** the adopting `expectedRevisionForHandleLocked` (`3038`, which calls `adoptCommittedRevisionLocked` first). So capture reads the un-advanced BaseRev.
- `materializeFullForUploadLocked` (`278`) only calls `CanMaterializeFull` + `LoadMissingParts` — no adopt, no BaseRev touch.
- The layer LoadPart closure (`2121-2160`) does a bare `c.ReadStreamRange(remoteFilePath, offset, length)` and touches only `fh.Path`/`committedSize` — never `fh.BaseRev`, never adopt. (The adopt-adjacent call @ `2117` is `adoptCommittedStorageClassLocked` — storage-class, at loader-install, not revision, not in-closure.)
- BaseRev is mutated only under this `fh.mu`, held continuously → no concurrent adopt during the fetch.

### 3. Closed loop → splice CANNOT commit
`expectedRevision` = frozen `BaseRev` = M, passed as the CAS `BaseRevision` on **both** `upsertLayerFile` paths (`604` large-file `UploadFSLayerFile(...expectedRevision...)`, `620` inline `req.BaseRevision`); `upsertLayerFile` uses it verbatim (no re-read/re-adopt).
- Sibling commits M→N during the unpinned fetch → retained parts = rev-N content spliced with old dirty → **phantom version materialized in `data`**.
- Upsert CAS `BaseRevision=M` vs remote=N → **409 → rejected** → `flushHandle` returns `httpToFuseStatus(err)` (`12558`), skips `ClearDirty` (success-only, `12560-12564`) → **splice never becomes durable; dirty preserved for retry.**

Precisely #812's resolved P1 closed loop.

### Residual — missing discriminating regression (test-gap, agreed)
`TestFlushHandleLayerLazyBufferFetchesAndUpserts` covers only the stable-loader happy path (fixed bytes, CAS `base_revision==5` asserted); there is **no** layer-mode sibling-commit-during-fetch test proving the CAS rejects the splice. adversary-1's coverage point stands — add a layer-mode test where the loader simulates an M→N commit mid-fetch, asserting the upsert 409s and dirty is preserved (mirror #812's P1 regression).

**Falsifiable hinge:** if anyone can show the layer LoadPart path (or anything between `12542` and `12556`) calls `adoptCommittedRevisionLocked` / advances `fh.BaseRev` to the fetched revision, this flips back to a real splice-commit blocker. I traced it as frozen.

Net for #816: **single true blocker = Path B new-file `remoteSize==0` zero-fill**; Path A cross-revision = CAS-backstopped, needs the discriminating regression. Still REVISE / do not merge.

qiffang:
## adversary-2 — RETRACTING my Path A downgrade. @adversary-1 is correct: layer-entry upsert has NO synchronous revision CAS → splice IS persisted + dirty IS cleared. Path A is a real blocker, not a missing-test.

I traced the SERVER path adversary-1 pointed to and my "CAS-backstopped at upsert" premise is **wrong**. I verified the client sends `expectedRevision` as `BaseRevision`, but I assumed the server treats it as a synchronous CAS at entry-upload. It does not.

**Source (head `3a382218`):**
- `datastore.UpsertFSLayerEntry` (`pkg/datastore/fs_layer.go:563-626`) checks only layer `state==Active` (`ErrFSLayerStateConflict`) + `validateFSLayerEntryWithinBaseRoot`, then **INSERTs** the entry with `entry.BaseRevision` stored as a plain column value. **There is NO `WHERE base_revision = current` comparison — BaseRevision is persisted metadata, not a CAS.**
- The revision-conflict gate is deferred to **commit preflight**: `preflightFSLayerCommit` (`fs_layer.go:1259-1282`) does `if nf.File.Revision != entry.BaseRevision → conflict "base revision changed"` → 409 + sets layer state `Conflicted` (`fs_layer.go:718-728`).

**So the real behavior of the layer flush (`dat9fs.go:12542-12564`):**
1. Sibling commits M→N during the unpinned fetch → `upsertLayerFile` → `UpsertFSLayerEntry` **succeeds** (no rev check) → the active layer entry now holds the **spliced content** (M-dirty + latest-path retained bytes) with `BaseRevision=M`.
2. `flushHandle` **clears `fh.Dirty`** on that success (`12560-12564`) → **the dirty buffer is gone**.
3. Only at a LATER layer *commit* does preflight detect `Revision(N) != BaseRevision(M)` → 409 `conflicted` → the **entire layer** transitions to `Conflicted`.

This is **worse** than "missing discriminating regression": the phantom splice is durably staged into the active layer entry AND dirty is cleared at flush time (not rejected before dirty-clear as I claimed), and the eventual conflict escalates to a whole-layer `Conflicted` state (blast radius = the layer, not just the one file). My #812 "CAS closes the loop before dirty-clear" reasoning does not transfer, because #812's direct-object-PUT path has a synchronous revision CAS while the **layer-entry-upsert path does not** — that's the seam I missed (inherited a benign premise across a sibling path with different server semantics).

**Retracted. Blocker #2 STANDS as a real Path A layer-entry correctness blocker** (cross-revision splice persisted + dirty lost). Thanks @adversary-1 for tracing the server path — I was wrong to downgrade.

**#816 net: TWO blockers** — (B) Path B new-file `remoteSize==0` zero-fill + (A) Path A layer cross-revision splice (persist + dirty-clear, no entry-upload CAS). Fix for A must anchor the retained-part fetch to `fh.BaseRev` (revision-pinned read, like #812's intended pin) OR gate the upsert on a synchronous revision check before clearing dirty, + a discriminating layer-mode sibling-commit regression. REVISE / do not merge.

mornyx:
@qiffang Both blockers addressed in `a32d97b`:

**Blocker 1 — Path B new-file (`remoteSize==0`) zero-fill.** Verified: `CanMaterializeFull()` is vacuously true for new files, so the shadow-source gate never engaged for new ShadowSpill files with evicted parts. The write-back snapshot now reads the authoritative shadow for **every** ShadowSpill handle with a present shadow — regardless of `remoteSize` or eviction state (ShadowSpill writes through to the shadow on every write, so it is authoritative by construction). Non-spill non-materializable buffers still fail `ENOTSUP`. New regression `TestSnapshotWriteBackNewShadowSpillFileUsesShadow` (`remoteSize==0`, part 0 evicted via `EvictPart`, asserts staged bytes == shadow bytes; fails against the previous gate).

**Blocker 2 — Path A layer cross-revision splice.** adversary-2's retraction is correct, verified at source: `datastore.UpsertFSLayerEntry` checked only layer state + base-root containment, then INSERTed — no revision comparison; the conflict gate only existed at commit preflight. So the splice persisted into the layer entry with dirty cleared, deferring detection to a whole-layer Conflicted escalation. Fix: `UpsertFSLayerEntry` now enforces a **synchronous base-revision CAS** for file content upserts with `base_revision > 0` — the current base file revision must match inside the same tx, otherwise `ErrRevisionConflict` → HTTP 409 (`writeFSLayerStoreError` mapping added). The writer fails with dirty preserved and retries; the splice can no longer be persisted. `base_revision <= 0` (new / layer-local files) skips the check, matching commit-preflight semantics, and the FUSE layer branch also captures `expectedRevision` before any lazy fetch (freeze-then-fetch, same invariant as the debounced path).

Tests:
- `TestFSLayerEntryUpsertBaseRevisionCAS` (server): current base → 200, stale after base rewrite → 409, fresh → 200.
- `TestFlushHandleLayerLazyFetchNewerRevisionConflicts` (FUSE): sibling rev-9 commit recorded from inside the lazy loader at fetch time → upsert carries the frozen `base_revision == 5` (never the newer revision), CAS rejects, dirty preserved.

Also folded in the Dev-Lead confirmation: `WriteBackSeq` is now advanced **only when the snapshot actually landed** (all three call sites), so a skipped/failed snapshot can never make a stale/absent `.dat` look current — your read of the fallback equivalence was right (commit path is shadow+pendingIndex), this just makes the invariant structural.

Validation: focused layer/shadow/debounce suites pass; full `pkg/server` (162s), `pkg/datastore`, `pkg/client`, `internal/e2ereport` green; `pkg/fuse` green except the two pre-existing macOS-only `TestShouldContinueAfterWaitMountPermissionError*` failures; live `e2e/layer-fs-smoke-test.sh` against a branch-built server: **66/66 PASS** (legit layer flows, including the commit-conflict path, unaffected by the upsert CAS). `make lint` 0 issues, gofmt clean.

qiffang (approved):
adversary-1 re-review on exact head `a32d97bca49fdceb6cfbc28fb0d7e7db432afabd`: APPROVE.

No remaining blockers found in the PR #816 scope.

Prior blocker closure:
- Path B new-file ShadowSpill / `remoteSize==0`: `snapshotWriteBackLocked` now snapshots from authoritative shadow whenever `fh.ShadowSpill` is true, independent of `CanMaterializeFull()`, and `WriteBackSeq` is only advanced after a successful snapshot. The new `TestSnapshotWriteBackNewShadowSpillFileUsesShadow` pins the exact sibling case that previously zero-filled evicted parts.
- Path A layer lazy flush cross-revision: `flushHandle` freezes the non-adopted base revision before lazy materialization, and the server/datastore now applies a synchronous `ErrRevisionConflict` CAS in `UpsertFSLayerEntry` before the entry is persisted. A 409 returns before FUSE marks local layer state or clears dirty. The new FUSE sibling-commit regression plus server CAS regression cover this seam.
- #812 manifest gap: workflow outcome ids are now manifest-covered and the sync test drives the real `SynthesizeSummaries` failure-classification path.

Validation run locally: `git diff --check d9509cff...HEAD`, gofmt diff on touched Go files, `go test ./internal/e2ereport -run TestSuiteManifestCoversWorkflowOutcomes -count=1`, focused `pkg/fuse` regression suite, key Path A/B FUSE regressions x20, `go test -c ./pkg/fuse ./pkg/server ./pkg/datastore ./internal/e2ereport`, and `go vet ./pkg/fuse ./pkg/server ./pkg/datastore ./internal/e2ereport`. Direct `go test ./pkg/server -run TestFSLayerEntryUpsertBaseRevisionCAS -count=1` is still blocked locally by the known rootless-Docker testcontainers panic, but hosted `ci`, `drive9-server-local e2e smoke`, and CodeRabbit are green at this head.

qiffang:
## adversary-2 — GREEN @ `a32d97bc` (both blockers closed), with one non-blocking robustness NOTE

Re-reviewed the two prior blockers at the new head (subagent + personal source-verify). Both data-integrity defects are closed.

### ✅ Blocker B (Path B new-file ShadowSpill zero-fill) — CLOSED
`snapshotWriteBackLocked` (dat9fs.go:~3363) now gates `fromShadow := fh.ShadowReady && shadowStore.Has(fh.Path) && (fh.ShadowSpill || !CanMaterializeFull())`. The added `fh.ShadowSpill ||` disjunct forces the authoritative-shadow read for spill handles **regardless of `CanMaterializeFull`** — so a new file (`remoteSize==0`, `CanMaterializeFull` vacuously true) with an evicted part now reads the shadow instead of zero-filling via `bytesView()`.
- **The Has-false fall-through I worried about is unreachable:** every ShadowSpill Write writes-through to the shadow FIRST (dat9fs.go:10780-10793, before `OnPartFull→EvictPart`), and every `shadowStore.Remove` atomically clears `ShadowSpill/ShadowReady` under `fh.Lock()` (1973/2367/2404/10824); ShadowStore has no LRU/GC (only explicit Remove). Invariant `ShadowSpill==true ⇒ Has()==true` holds → no residual zero-fill.
- Regression `TestSnapshotWriteBackNewShadowSpillFileUsesShadow` uses `remoteSize==0`, asserts `CanMaterializeFull()==true` (pins the exact gap), then asserts staged bytes == authoritative shadow (not zero-filled). Discriminates against prior head.
- Bonus: `WriteBackSeq` now advances only on snapshot success (Flush/Fsync else-branch) → no stale-`.dat`-served-as-current.

### ✅ Blocker A (Path A layer cross-revision splice) — CLOSED (data-loss prevented)
The right fix landed: a **synchronous base-revision CAS inside `UpsertFSLayerEntry`** (datastore/fs_layer.go:588+) — `Op==Upsert && Kind==File && BaseRevision>0` → stat the base file, `current != entry.BaseRevision → ErrRevisionConflict` → server maps to 409. Frozen non-adopting `expectedRevisionForHandle` (BaseRev=M) → sibling M→N → entry.BaseRevision=M vs current=N → 409 → **entry not persisted**, and `flushHandle` returns `httpToFuseStatus(409)=EIO` **before** ClearDirty (12581 vs 12585) → **dirty preserved**, splice discarded in memory. Nothing writes the spliced bytes durably before the CAS. Closed loop confirmed. Server CAS regression `TestFSLayerEntryUpsertBaseRevisionCAS` + FUSE `TestFlushHandleLayerLazyFetchNewerRevisionConflicts` discriminate against prior head.

### ⚠️ Non-blocking NOTE (robustness, not data-loss) — tighten the CAS read to FOR UPDATE
The CAS reads the base revision via `s.StatTx(ctx, tx, entry.Path)` — a **plain `SELECT ... LIMIT 1`, no `FOR UPDATE`** (store.go). The tx's only lock is `SELECT ... fs_layers ... FOR UPDATE` keyed by `layer_id` (the layer row), taken first. But the sibling base-file M→N commit runs in a separate tx updating `inodes.revision`/`files.revision` and **never touches `fs_layers`**, so the layer lock gives no mutual exclusion against it. Under TiDB RR, the plain `StatTx` reads the tx's pinned snapshot — so in a narrow window (sibling commits between this tx's start_ts and the StatTx read) it can read stale M, match, and admit the stale splice past the *synchronous* CAS.
- **Severity is bounded — NOT a data-loss merge-blocker:** the commit-preflight backstop is intact (`preflightFSLayerCommit`, server/fs_layer.go:1275 `nf.File.Revision != entry.BaseRevision → conflict`), so any splice that slips the CAS is still caught at layer-commit → Conflicted, exactly as before the CAS existed. No regression.
- The CAS simply doesn't fully deliver the *synchronous* early-reject it advertises in that rare interleaving, and it deviates from the file's own idiom (store.go base-file reads use `FOR UPDATE` at 457/565/816/…). **Recommend** reading the base revision `FOR UPDATE` inside the tx for a genuinely race-free synchronous guarantee; the sequential CAS test doesn't exercise the concurrent window. Non-blocking.

**Verdict: GREEN** — both original data-integrity blockers (Path B zero-fill, Path A splice) are closed; the residual is a bounded, backstopped TOCTOU robustness note, not a data-loss blocker. Merge authority = qiffang (drive9 repo).

PR #823, opened 2026-07-31
fix: reclaim tenant metrics on pool close

srstack:
## Summary

- reclaim only `drive9_tenant_requests_total` tenant series when a local tenant backend cache entry closes
- account for an authenticated request's tenant counter before its final backend release, so outer HTTP observation cannot recreate the series after close cleanup
- keep tenant gauges, histograms, SSE metrics, storage metrics, and the durable tenant-deletion cleanup contract unchanged
- delete from the request counter's own label map outside the tenant-pool mutex; no whole-registry scan is performed on eviction

## Why this revisits #802

#802 deliberately stopped deleting metrics on pool eviction to preserve a never-reset tenant request counter. Production Tencent data shows that policy leaves roughly 20k unique request-counter tenants per pod while the local backend cache contains only hundreds of tenants. The agreed metric contract is now that `tenant_requests_total` may reset after cache eviction; dashboards must use `rate()` or `increase()` rather than treat the raw process-local counter as a durable tenant request ledger.

This PR deliberately does not restore #773/#802-era broad `DeleteTenantCounters` behavior on cache close. Durable deleting/deleted lifecycle transitions continue to use `DeleteTenantCounters` and the cross-pod `WorkMetricsCleanup` outbox.

## Validation

- `go test ./pkg/server -run '^TestRecordTenantHTTPRequestAtCompletionDoesNotRecreateCleanedCounter$' -count=1`
- `go test ./pkg/server ./pkg/metrics ./pkg/tenant -count=1`
- `make test` was run earlier; it is not fully green due to unrelated existing failure `cmd/drive9/cli.TestCtxImport_RejectsWorldReadableFile` (mode-0644 JWT file expected to be rejected but received nil).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved tenant request metrics accuracy by preventing duplicate request recordings.
  - Tenant request counters are now cleared when tenant resources are closed, while active in-flight metrics remain intact.
  - Prevented cleared counters from being recreated unexpectedly during request completion.
  - Added safeguards to keep request metrics consistent across authentication, streaming, and standard request flows.

- **Tests**
  - Added regression coverage for counter cleanup, preservation of in-flight metrics, and prevention of duplicate recordings.

### DIFF
--- a/pkg/metrics/operations.go
+++ b/pkg/metrics/operations.go
@@ -290,6 +290,18 @@ func DeleteTenantCounters(tenantID string) {
 	globalRegistry.DeleteGaugesByLabel("tenant_id", tenantID)
 }
 
+// DeleteTenantRequestCounters removes only request-counter series for a
+// tenant. Cache eviction intentionally resets request counters without
+// removing tenant gauges or histograms whose values do not share counter-reset
+// semantics.
+func DeleteTenantRequestCounters(tenantID string) {
+	tenantID = cleanMetricValue(tenantID, "unknown")
+	if tenantID == "unknown" {
+		return
+	}
+	globalRegistry.DeleteCounterByLabel("drive9_tenant_requests_total", "tenant_id", tenantID)
+}
+
 func RecordTiDBCloudRBACCacheRequest(path, scope, result string) {
 	RegisterModule("tidbcloud_quota")
 	tidbCloudRBACCacheRequestsTotal.Add(1,
--- a/pkg/metrics/registry.go
+++ b/pkg/metrics/registry.go
@@ -342,6 +342,21 @@ func (r *Registry) DeleteCountersByLabel(labelKey, labelValue string) {
 	}
 }
 
+func (r *Registry) DeleteCounterByLabel(name, labelKey, labelValue string) {
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	inst := r.counters[name]
+	if inst == nil {
+		return
+	}
+	match := labelKey + `="` + EscapePromLabel(labelValue) + `"`
+	for labels := range inst.values {
+		if labelHasKeyValue(labels, match) {
+			delete(inst.values, labels)
+		}
+	}
+}
+
 func (r *Registry) DeleteHistogramsByLabel(labelKey, labelValue string) {
 	r.mu.Lock()
 	defer r.mu.Unlock()
--- a/pkg/tenant/pool.go
+++ b/pkg/tenant/pool.go
@@ -1781,6 +1781,7 @@ func (p *Pool) closeEntry(e *entry) {
 		_ = e.store.Close()
 	}
 	e.sharedDBLease.Release()
+	metrics.DeleteTenantRequestCounters(e.tenantID)
 }
 
 type tenantPoolResult string
--- a/pkg/tenant/pool_test.go
+++ b/pkg/tenant/pool_test.go
@@ -1862,27 +1862,45 @@ func TestAcquireRefreshesStandaloneIdleTTL(t *testing.T) {
 	assertStoreOpen(t, b.Store())
 }
 
-func TestCloseEntryPreservesLiveTenantCounters(t *testing.T) {
-	const tenantID = "tenant-live-counter-close-entry"
+func TestCloseEntryCleansTenantRequestCounters(t *testing.T) {
+	const tenantID = "tenant-clean-counter-close-entry"
 	metrics.DeleteTenantCounters(tenantID)
 	t.Cleanup(func() { metrics.DeleteTenantCounters(tenantID) })
 
 	metrics.RecordTenantRequestCountWithOrg(tenantID, "org-live-counter", "fs", "read", 200)
-	assertTenantMetricPresent := func() {
-		t.Helper()
-		recorder := httptest.NewRecorder()
-		metrics.WritePrometheus(recorder)
-		if !strings.Contains(recorder.Body.String(), `drive9_tenant_requests_total{`) ||
-			!strings.Contains(recorder.Body.String(), `tenant_id="`+tenantID+`"`) {
-			t.Fatalf("live tenant counter disappeared from metrics:\n%s", recorder.Body.String())
-		}
+	metrics.RecordTenantInFlightWithOrg(tenantID, "org-live-counter", "fs", 2)
+	recorder := httptest.NewRecorder()
+	metrics.WritePrometheus(recorder)
+	if !strings.Contains(recorder.Body.String(), `drive9_tenant_requests_total{`) ||
+		!strings.Contains(recorder.Body.String(), `tenant_id="`+tenantID+`"`) {
+		t.Fatalf("tenant request counter missing before close:\n%s", recorder.Body.String())
 	}
-	assertTenantMetricPresent()
+	if !strings.Contains(recorder.Body.String(), `drive9_tenant_inflight_requests{`) ||
+		!strings.Contains(recorder.Body.String(), `tenant_id="`+tenantID+`"`) {
+		t.Fatalf("tenant inflight gauge missing before close:\n%s", recorder.Body.String())
+	}
 
 	pool := NewPool(PoolConfig{}, nil)
 	pool.closeEntry(&entry{tenantID: tenantID})
 
-	assertTenantMetricPresent()
+	recorder = httptest.NewRecorder()
+	metrics.WritePrometheus(recorder)
+	if metricHasTenantID(t, recorder.Body.String(), "drive9_tenant_requests_total", tenantID) {
+		t.Fatalf("tenant request counter remained after close:\n%s", recorder.Body.String())
+	}
+	if !metricHasTenantID(t, recorder.Body.String(), "drive9_tenant_inflight_requests", tenantID) {
+		t.Fatalf("tenant inflight gauge disappeared after close:\n%s", recorder.Body.String())
+	}
+}
+
+func metricHasTenantID(t *testing.T, text, metric, tenantID string) bool {
+	t.Helper()
+	for _, line := range strings.Split(text, "\n") {
+		if strings.HasPrefix(line, metric+"{") && strings.Contains(line, `tenant_id="`+tenantID+`"`) {
+			return true
+		}
+	}
+	return false
 }
 
 func TestIdleEvictionSkippedByRecentAcquire(t *testing.T) {
--- a/pkg/server/auth.go
+++ b/pkg/server/auth.go
@@ -388,6 +388,7 @@ func tenantAuthMiddlewareWithFSScopeLoader(metaStore *meta.Store, pool *tenant.P
 			return
 		}
 		defer release()
+		defer func() { recordTenantHTTPRequestAtCompletion(r.Context()) }()
 		metricEvent(r.Context(), "auth", "result", "ok")
 		totalDuration := time.Since(authStart)
 		logger.InfoOpenPoolTiming(r.Context(), "tenant_auth_timing", totalDuration,
@@ -530,6 +531,7 @@ func (s *Server) capabilityAuthMiddleware(metaStore *meta.Store, pool *tenant.Po
 			return
 		}
 		defer release()
+		defer func() { recordTenantHTTPRequestAtCompletion(r.Context()) }()
 
 		scope := &TenantScope{TenantID: tenantID, Provider: tenant.Provider, TiDBCloudOrgID: b.TiDBCloudOrgID(), Backend: b}
 		setRequestMetricScope(r.Context(), scope, classifyTenantRequest(r))
--- a/pkg/server/instrumentation.go
+++ b/pkg/server/instrumentation.go
@@ -75,6 +75,9 @@ type requestMetricState struct {
 	action         string
 	tenantInFlight bool
 
+	tenantHTTPRequestRecorded bool
+	tenantHTTPRequestRecorder func()
+
 	bodyReadObserved bool
 	bodyReadMethod   string
 	bodyReadRoute    string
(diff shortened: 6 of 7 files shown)


srstack:
### Review at `a701c87`

Reviewed the full diff, all 10 `closeEntry` call sites, the metrics registry internals, and both durable-cleanup paths. `go test -race ./pkg/tenant -count=1` passes, the two new tests pass, `go vet` and `gofmt` are clean.

**What's correct:** the replacement-detection logic is precise. `closeEntry` only ever receives an entry with `retired=true && refs==0` that `removeLockedWithoutMetrics` already deleted from `p.items` under `p.mu`, so a hit on `p.items[e.tenantID]` can only be a *newly inserted* replacement. And because `removeLocked(..., "replace")` and `p.items[t.ID] = e` happen in the **same** critical section (pool.go:495-501), the later re-lock in `closeEntry` is guaranteed to observe the replacement — the serialization claim in the description holds. I also verified no call site holds `p.mu` when calling `closeEntry` (all 10 call it after `Unlock`), so the added `p.mu.Lock()` cannot self-deadlock, and `p.mu → registry.mu` nesting already exists via `recordCachedBackendCountLocked`, so there's no new inversion.

---

Two things I'd want addressed before merge:

**`[MAJOR]` This is an unacknowledged revert of #802, and the PR body doesn't say so.**

`git log -S` shows the removed test `TestCloseEntryPreservesLiveTenantCounters` was introduced by c67c7105 (#802, "reduce metric cardinality and close observability gaps"), which *deliberately deleted this exact block* from `closeEntry` — the diff being restored here is character-for-character what #802 removed, minus moving the delete inside the lock. So #823 reverses a 4-day-old intentional decision, and inverts the test that pinned it (`...PreservesLiveTenantCounters` → `...CleansTenantCounters`).

That may well be the right call — #802's version had the replacement race this PR fixes, which is plausibly why it was pulled rather than fixed. But an inverted assertion on a deliberately-added regression test is exactly the change that needs its rationale in the commit message, or the next person re-reverts it. Please state in the body why #802's removal was wrong (or was an overcorrection) and why re-adding it under the lock is now safe.

**`[MAJOR]` Local cache eviction is not tenant deletion — the two lifecycles are now conflated.**

Every other `DeleteTenantCounters` caller is a *durable* death signal: `clearLocalTenantMetrics` (server.go:793, on deleting/deleted transition) and the `WorkMetricsCleanup` outbox row (tenant_outbox_poller.go:219, cross-pod). In all of those, the tenant is gone for good, so dropping its series is correct.

`closeEntry` is different: it fires on `evict` (LRU pressure), `shared_idle`, `invalidate`, and `shutdown` — for a tenant that is *still alive* and will likely be re-acquired minutes later. On a busy pod at `maxSize`, an active tenant can be evicted and re-acquired repeatedly, so its series get dropped and recreated repeatedly.

The body addresses this for counters ("consumers should use `rate()`/`increase()`"), and I agree that argument is sound for counters. But `DeleteTenantCounters` also calls `DeleteHistogramsByLabel` and `DeleteGaugesByLabel`, and it matches on the `tenant_id` label alone across the whole registry — roughly two dozen `tenant_id`-labelled instruments. `rate()` does not rescue a gauge: a tenant-scoped gauge that disappears and reappears reads as a genuine state change, not a reset, and histogram quantiles over a re-created series lose their window. Could you either narrow this to the counters the `rate()` argument actually covers, or confirm no tenant-labelled gauge/histogram is alerted on as an absolute value? The runbook expressions from #802 are the thing to check against.

---

**`[MINOR]`** `DeleteTenantCounters` under `p.mu` is a cost change, not just a placement change: it's three full registry scans (`counters` + `histograms` + `gauges`, each iterating every instrument's label set under the global `registry.mu`) now inside the pool mutex that serializes *all* tenant acquires. #802's version did the `p.items` lookup under the lock and the delete outside it. Since a replacement can only be inserted while `p.mu` is held, you can keep the exact same safety by snapshotting `active` under the lock and calling `DeleteTenantCounters` after `Unlock` — the replacement can't appear in between, because inserting it requires the lock you just released and it would have been visible to your read. Worth doing on the shutdown path especially, where this runs once per cached entry in a loop.

**`[NIT]`** `TestCloseEntryPreservesReplacementTenantCounters` builds the replacement with `pool.items[tenantID] = &entry{...}` directly rather than going through the acquire path, so it pins the `p.items`-lookup behavior but not the 495-501 same-critical-section property that actually makes it safe. A test that races a real replacing `Acquire` against the old entry's close would be a stronger guard — optional.

Mechanically the change is sound and well-tested; my hesitation is entirely about the two semantic points above. Happy to re-review once the rationale is in the body and the gauge/histogram question is settled.

_(GitHub blocks a formal review verdict on one's own PR, so this comment stands in for it — treat it as changes-requested on the two MAJOR items.)_

srstack:
Addressed in `f0480f63`.

- The PR description now explicitly documents the intentional change from #802: request counters may reset on local cache eviction because production has demonstrated that preserving them indefinitely creates unbounded per-process cardinality.
- `closeEntry` no longer calls broad `DeleteTenantCounters`. It now removes only `drive9_tenant_requests_total`; tenant gauges, tenant-labelled histograms, SSE metrics, storage metrics, and their durable-deletion cleanup semantics remain unchanged. This addresses the gauge/histogram concern.
- The new `Registry.DeleteCounterByLabel` scans only the request counter's label map. It is called outside `p.mu`, so eviction no longer performs a whole-registry scan while serializing acquires.
- The replacement-preservation test and `p.items` check were removed intentionally: the accepted contract is that cache close can reset this request counter, including during replacement. The next recorded request recreates the series.

Validation: `go test ./pkg/metrics ./pkg/tenant -count=1` passes.

srstack:
### Re-review at `f0480f6` — LGTM ✅

All three actionable items from my previous review are addressed, and the approach you took is better than what I suggested. `go test -race ./pkg/tenant ./pkg/metrics -count=1` passes, the new test passes, `go vet` and `gofmt` clean.

**`[MAJOR]` #802 rationale — resolved.** The body now states the reversal explicitly and quantifies it: ~20k unique request-counter tenants per pod against a local cache holding hundreds. That's the concrete justification the inverted assertion needed, and it makes clear this is a scoped cardinality fix rather than a blanket revert. The explicit "this PR deliberately does not restore #773/#802-era broad `DeleteTenantCounters` behavior" line is exactly the kind of note that stops the next person from re-reverting.

**`[MAJOR]` gauge/histogram conflation — resolved, better than proposed.** Rather than auditing which tenant-labelled gauges are alerted on as absolute values, you narrowed the blast radius to the single instrument the `rate()` argument actually covers. `DeleteTenantRequestCounters` → `DeleteCounterByLabel("drive9_tenant_requests_total", ...)` touches one counter's label map; gauges and histograms are untouched by construction, so the reset-semantics mismatch I was worried about can't arise. I verified the instrument-name string matches the registration at operations.go:82 and that the registry keys counters by that exact name.

**`[MINOR]` lock cost — resolved.** The delete is now outside `p.mu` entirely (pool.go is a single added line), and it's no longer three whole-registry scans — one map lookup plus one instrument's label set. The shutdown loop concern is gone.

**Test got stronger too.** `metricHasTenantID` does per-line prefix matching instead of whole-body `strings.Contains`, which removes a real false-negative risk in the old assertion (a `tenant_id` substring match anywhere in the payload would have satisfied it). More importantly it now asserts the *positive* half of the new contract — `drive9_tenant_inflight_requests` must survive the close — so the "counter only, never gauges" boundary is pinned in both directions rather than assumed.

---

Two non-blocking notes:

- **`[NIT]`** Dropping the `p.items` replacement check is fine here, but for a different reason than the v1 design, so worth recording: the counter is fed only from `recordTenantHTTPRequest` (instrumentation.go:710) on the live HTTP path, independent of pool state. A replacement acquiring concurrently with an old entry's close can therefore lose a few just-recorded increments. Under the now-explicit `rate()`/`increase()` contract that's a bounded, self-healing reset — the same class of event the contract already sanctions — so no check is needed. v1's `p.items` guard was protecting a much wider deletion, and it correctly went away with it.
- **`[NIT]`** `DeleteCounterByLabel` returns silently when `r.counters[name]` is nil, so if `drive9_tenant_requests_total` is ever renamed, this cleanup becomes a no-op and the cardinality regression returns quietly. The name is already duplicated as a literal at operations.go:302 while registered at :82 — extracting a shared const would make a rename fail loudly at compile time instead.

Scoped correctly, cheaper, better documented, and the contract is now pinned by tests in both directions. LGTM.

_(GitHub blocks a formal approval on one's own PR, so this comment stands in for it.)_

pkg/metrics/operations.go
```diff
@@ -290,6 +290,18 @@ func DeleteTenantCounters(tenantID string) {
 	globalRegistry.DeleteGaugesByLabel("tenant_id", tenantID)
 }
 
+// DeleteTenantRequestCounters removes only request-counter series for a
+// tenant. Cache eviction intentionally resets request counters without
+// removing tenant gauges or histograms whose values do not share counter-reset
+// semantics.
+func DeleteTenantRequestCounters(tenantID string) {
```
mornyx, line 297: **[nit]** Narrow API is the right compromise vs restoring full `DeleteTenantCounters` on every cache close. Request counters are the high-cardinality series (tenant × surface × action × status_class); duration histograms already omit `tenant_id`.

Looks correct.

pkg/metrics/registry.go
```diff
@@ -342,6 +342,21 @@ func (r *Registry) DeleteCountersByLabel(labelKey, labelValue string) {
 	}
 }
 
+func (r *Registry) DeleteCounterByLabel(name, labelKey, labelValue string) {
```
mornyx, line 345: **[nit]** Name-scoped delete avoids scanning every counter instrument on eviction — important if the registry grows. Matching uses the same `labelHasKeyValue` helper as `DeleteCountersByLabel`.

Optional: a small registry unit test for multi-label series with the same `tenant_id` would lock this path without going through the pool.

pkg/tenant/pool.go
```diff
@@ -1781,6 +1781,7 @@ func (p *Pool) closeEntry(e *entry) {
 		_ = e.store.Close()
 	}
 	e.sharedDBLease.Release()
+	metrics.DeleteTenantRequestCounters(e.tenantID)
```
mornyx, line 1784: **[nit]** Correct hook: `closeEntry` only runs after retirement when `refs == 0` (or immediately on invalidate when already idle). Called outside `p.mu`.

Note on request/observe ordering: auth `defer release()` can delete counters before outer `observe` records the finishing request. That re-creates the series with the last sample rather than dropping it — fine under the new `rate()`/`increase()` contract. Not a blocker.

pkg/tenant/pool_test.go
```diff
@@ -1862,27 +1862,44 @@ func TestAcquireRefreshesStandaloneIdleTTL(t *testing.T) {
 	assertStoreOpen(t, b.Store())
 }
 
-func TestCloseEntryPreservesLiveTenantCounters(t *testing.T) {
-	const tenantID = "tenant-live-counter-close-entry"
+func TestCloseEntryCleansTenantRequestCounters(t *testing.T) {
```
mornyx, line 1865: **[nit]** Good rewrite of the #802 “preserve counters” test into an explicit cleanup assertion, and the gauge retention check documents the intentional asymmetry. `metricHasTenantID` is a nicer helper than raw `strings.Contains` on the full scrape body.

mornyx (approved):
## Review summary

**Verdict: Approve**

Reviewed in an isolated worktree at `f0480f6`. Focused fix for process-local Prometheus cardinality of `drive9_tenant_requests_total` after #802 stopped reclaiming series on cache eviction. Local tests passed:

- `go test ./pkg/tenant -run '^TestCloseEntryCleansTenantRequestCounters$'`
- `go test ./pkg/metrics`

CI green. The design matches the stated contract: request counters may reset on cache eviction and must be queried with `rate()` / `increase()`, while durable deleting/deleted cleanup still uses full `DeleteTenantCounters` + cross-pod `WorkMetricsCleanup`.

### Correctness

1. **Scope is intentionally narrow** — `DeleteTenantRequestCounters` only touches the `drive9_tenant_requests_total` counter map for one `tenant_id`. Gauges/histograms/SSE/storage are left alone. Request duration histogram is already **not** labeled by `tenant_id` (only `surface`/`status_class`), so it is not the cardinality hot spot.
2. **Placement** — called from `closeEntry`, which runs when an entry is retired and `refs` hit 0 (idle eviction, invalidate after last release, replace, shutdown). Not under the pool mutex; only the metrics registry lock is taken for a single counter instrument — good.
3. **In-flight request order** — `observe` is outer: `next.ServeHTTP` completes (auth `defer release()` may `closeEntry` + delete counters) **then** `recordTenantHTTPRequest` increments the counter again. So the completing request is not lost; the series is recreated from 0 with the last sample. Compatible with `rate()`/`increase()`. Historical pre-eviction series intentionally disappear.
4. **Pin safety** — `removeLockedWithoutMetrics` returns the entry for immediate close only when `refs == 0`; otherwise close is deferred to the last `releaseEntry`. No close while pins remain.
5. **Lifecycle cleanup unchanged** — durable delete path still calls full `DeleteTenantCounters` via outbox / work mask.
6. **Test** — asserts request counter gone and inflight gauge retained after `closeEntry`.

### Non-blocking notes

- Codex P2 about “delete before observe” is real order-wise but not a functional loss of the completing request; series reset is the product intent. No code change needed if dashboards use `rate()`/`increase()`.
- Optional unit test for `Registry.DeleteCounterByLabel` (name-scoped delete) to lock the O(series for one counter) contract.
- Inflight gauges with `tenant_id` can still linger at last value after eviction; out of scope here and probably lower cardinality, but worth watching if gauge series also grow.

No blockers. Approved.

pkg/tenant/pool.go
```diff
@@ -1781,6 +1781,12 @@ func (p *Pool) closeEntry(e *entry) {
 		_ = e.store.Close()
 	}
 	e.sharedDBLease.Release()
+	p.mu.Lock()
+	_, active := p.items[e.tenantID]
+	if !active {
+		metrics.DeleteTenantCounters(e.tenantID)
+	}
```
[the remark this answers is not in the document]
srstack: Fixed in f7dd66e4. observe now registers a once-only tenant request recorder. Both auth middleware paths invoke it in a defer that runs before release(), while observe invokes the same recorder only as a fallback after next returns. The finishing request is therefore accounted before closeEntry deletes its tenant series; the outer fallback sees it already recorded and cannot recreate the counter. Regression: TestRecordTenantHTTPRequestAtCompletionDoesNotRecreateCleanedCounter.

pkg/tenant/pool_test.go
```diff
@@ -1862,27 +1862,44 @@ func TestAcquireRefreshesStandaloneIdleTTL(t *testing.T) {
...
+	}
+}
+
+func metricHasTenantID(text, metric, tenantID string) bool {
+	for _, line := range strings.Split(text, "\n") {
+		if strings.HasPrefix(line, metric+"{") && strings.Contains(line, `tenant_id="`+tenantID+`"`) {
+			return true
+		}
+	}
+	return false
```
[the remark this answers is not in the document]
srstack: Fixed in 409a8924: metricHasTenantID now accepts *testing.T, calls t.Helper(), and both call sites pass the current test handle. Validation: go test ./pkg/tenant -run ^'^TestCloseEntryCleansTenantRequestCounters$'$' -count=1 and go test ./pkg/server ./pkg/metrics ./pkg/tenant -count=1.

PR #826, opened 2026-08-03
feat(server): 7-day SSE event retention with tenant-DB residency

mornyx:
## Summary

SSE filesystem events are currently retained for **1 hour** (hard-coded). This PR implements the approved design [`docs/design/sse-event-log-retention.md`](docs/design/sse-event-log-retention.md) to make the event log a customer-facing integration channel:

- **7-day retention**, operator-configurable, while keeping event tables in tenant DBs;
- **zero periodic wakeups of cold (serverless) tenant DBs** — retention sweeps ride the write path instead of cron jobs (no TTL, no per-tenant partition maintenance, no cluster sweeper over dedicated DBs);
- **99.99% event durability** via a bounded in-pod retry buffer — a failed event insert never fails the enclosing filesystem mutation;
- **replay completeness** for arbitrarily large backlogs, including error-safe drain behavior.

The core insight: completeness only constrains the *lower bound* of retention, so over-retention is harmless and sweeping can be lazy. The two planes are explicitly decoupled: **log** (`fs_events` in the tenant DB, 7 days, source of truth) vs **signal** (`tenant_notify_outbox` in the meta DB, 1 hour, lossy wake-up hint — unchanged).

## What was done

**Retention & sweeping**
- `fs_events` retention is configurable via `DRIVE9_FS_EVENTS_RETENTION` (default `1h`; production: `168h`), plumbed to *both* sweep paths; the effective value is logged at startup.
- `DeleteFSEventsBefore` is now batched (`LIMIT 5000` per statement, max 20 batches per sweep, `hasMore` semantics) — no more unbounded DELETE against a 7-day table.
- Lazy write-path sweep: after a successful event insert, a per-tenant throttled (1/hour), detached-goroutine sweep. Cold tenants are never woken; they simply over-retain (harmless by design).
- Shared-schema deployments get the pool-wide sweep from the design's storage policy: `DeleteSharedFSEventsBefore` (no `fs_id` predicate, refuses non-shared stores) triggered from *both* the write path and piggyback maintenance (warm tenants only), throttled by a per-pod pre-filter plus an authoritative cluster-wide meta-DB claim (`shared_maintenance_state` + `ClaimSharedMaintenanceRun`) so N pods never sweep the same table on N independent clocks. Metrics use a fixed `tenant_id="shared_pool"` label.

**Durability (retry buffer)**
- Failed `fs_events` inserts are captured in a per-pod buffer (global cap 10k, per-tenant cap 1k, per-entry exponential backoff, per-flush 30s timeout) and flushed in the background. A successful flush performs the mandatory **second wake** (`insertTenantNotify` + `bus.Publish`) so subscribers actually learn about the late row.
- Max entry age tracks retention (`min(retention, 24h)`, floored at 1h). A `stopped` gate counts post-stop enqueues; shutdown performs a 10s best-effort final flush and counts any leftover as `dropped{reason="shutdown"}`.
- New SLI metrics: `drive9_sse_event_retry_dropped_total{reason=tenant_cap|global_cap|expired|stopped|shutdown}` (hard loss, alert on non-zero) and a buffer-depth gauge. `seq` is a *persistence order*, not causal order (bounded reorder around insert failures) — documented in the consumer contract.

**Replay completeness (SSE handler)**
- Phase 1 loops `EventsSince` until a short page (drained), flushing per batch and interleaving heartbeats every 5s so LB idle timeouts can't kill multi-minute replays.
- Phase 2 event-driven drain: a full page re-polls immediately (up to 10 pages per wake), then flushes and re-arms a 10ms catch-up timer; only a short page (or reset) disarms. Fixes a stall where a >10k-event burst in a single coalescer window was truncated until the next write.
- Error semantics hardened end-to-end: a mid-drain Phase-1 query error now **terminates the stream** (no false "caught-up" heartbeat; the client resumes from its durable cursor); Phase-2 tolerated query errors no longer disarm the timer, no longer stamp `lastSuccessfulPoll`, and re-arm with bounded exponential backoff (10ms doubling, 1s cap, reset on success) whenever the wake came from the timer or already delivered full pages. A quiet stream's first-poll failure still starts no polling.
- Optional `DRIVE9_SSE_LIVENESS_POLL_INTERVAL` (default off): per-connection liveness poll covering meta-DB signal loss for connected clients only.

**Outbox pruning robustness**
- The `MIN(last_id)` prune floor now ignores cursors whose `updated_at` is older than 10 minutes, and `updated_at` only advances when `last_id` advances — so a pod with a failing poll but healthy cursor flush goes stale and can no longer pin the floor forever (the `ON DUPLICATE KEY UPDATE` assignment order is load-bearing; covered by meta tests).

**Misc**
- `CountFSEvents` is bounded (100k-row capped subquery) since hot-tenant tables grow 168x.
- **Removed** the legacy `/v1/internal/sse-notify` pod-to-pod push endpoint (a repo-wide search confirmed no sender ever existed) together with `DRIVE9_POD_NOTIFY_SECRET`.
- Docs: design doc (3 review rounds incorporated), SSE guide (retention semantics, `seq` persistence-order/idempotent-hint contract, no-ack/no-webhook recovery path, signal-loss stall trade-off).

## Breaking changes

- **SSE consumers: none.** Wire protocol, event format, `seq` semantics, and `reset` types are unchanged; the Phase-1/2 fixes only remove stalls.
- **Operations: one removal.** `/v1/internal/sse-notify` and `DRIVE9_POD_NOTIFY_SECRET` are gone. No sender ever existed in this codebase, and the outbox poller was already the primary cross-pod path (~200ms), so the practical impact is nil — but it should be called out in the release notes. Deployments that set the env var will simply see it ignored.
- **Programmatic embedders of `pkg/server` (none in-repo besides `cmd/drive9-server`):** `Config.PodNotifySecret` removed; `DeleteFSEventsBefore` signature changed.
- **Metrics:** `drive9_fs_events_rows` is now capped at 100k (dashboards will plateau; help text updated).

## Schema changes

- **One, auto-migrated:** new meta-DB table `shared_maintenance_state`. Created automatically at server startup by the existing locked diff-migrator (`Store.migrate`) — no manual DDL. Old pods ignore the table; mixed-version rolling deploys are safe.
- **Tenant DBs: zero changes.** `fs_events` is untouched (the structural-payload column is a separate, future track). `tenant_outbox_cursor.updated_at` is a pre-existing column — only its refresh condition changed.
- Two SQL semantics are verified on MySQL 8.0 only; **before production on TiDB**, run `TestUpsertTenantOutboxCursorRefreshesUpdatedAt` and `TestClaimSharedMaintenanceRun` against real TiDB (assignment-order and `INTERVAL ?` placeholder semantics).

## Configuration changes

| Env | Status | Notes |
| --- | --- | --- |
| `DRIVE9_FS_EVENTS_RETENTION` | **new, optional** | Default `1h` (= current behavior). Set `168h` to enable the 7-day window. Effective value is logged at startup. |
| `DRIVE9_SSE_LIVENESS_POLL_INTERVAL` | new, optional | Default `0` = off. |
| `DRIVE9_SSE_NOTIFY_RETENTION` | unchanged | Keep at `1h` — do **not** raise it together with the log retention; the signal plane is deliberately short-lived. |
| `DRIVE9_POD_NOTIFY_SECRET` | **removed** | Inert if still present in env. |

No config change is required to deploy: with all defaults, behavior is identical to today. Roll out the binary first, flip retention when ready.

## Risks for existing deployments, and how they're neutralized

- **Rolling upgrade:** mixed old/new pods are compatible. Old pods' unconditional cursor refresh simply keeps them "fresh" under the new floor logic (same as today); the new meta table is unused by old pods; retry buffer is in-memory only (no upgrade/downgrade state).
- **Raising retention is not backfill-destructive:** data beyond the old 1h horizon is already pruned, so tables grow from the 1h waterline. Watch `drive9_fs_events_rows` (capped) and disk at peak-write-rate × 168h per tenant.
- **Downgrade path (the one runbook item that matters):** after accumulating 7 days of rows, rolling back to the old binary would make its *unbounded* single-statement DELETE hit a 7-day table (TiDB transaction-size risk). Correct order: first set `DRIVE9_FS_EVENTS_RETENTION=1h` on the **new** binary and let the batched sweeper drain over a few cycles, *then* roll back.
- **Behavioral expectation:** on the default config, a signal-plane (meta DB) outage means online clients stall until reconnect — events are durable, delivery pauses. Enable `DRIVE9_SSE_LIVENESS_POLL_INTERVAL` if connected clients must self-heal (bounded poll against DBs of connected tenants only).
- Residual loss modes (all metered): crash with unflushed buffer, buffer overflow under sustained tenant-DB outage, consumer offline beyond the window (surfaced as `reset(seq_too_old)` — alert on it).

## Test plan

- `go build ./...`, `make lint` — clean (0 issues).
- `go test ./pkg/server/... ./pkg/datastore/... ./pkg/meta/...` (MySQL testcontainers) — all green, including new coverage: retry-buffer caps/stopped-gate/shutdown/max-age clamp, Phase-1 multi-page drain + mid-drain error termination, Phase-2 catch-up timer burst drain + query-error backoff/resume (regression-validated by temporary revert), batched delete `hasMore`, capped count, shared-shape pool sweep + refuse-standalone, meta freshness floor + cursor refresh semantics + maintenance claim, stale-pod cursor deletion, server-level shared-sweep orchestration (claim won/lost, pre-filter, claim-error retry).
- Failpoint suite not run (not touched by this PR).

## Summary by CodeRabbit

* **New Features**
  * Added configurable retention for filesystem event history to support reliable replay.
  * Improved live event delivery with paginated catch-up, heartbeats, optional liveness polling, and interruption recovery.
  * Added background retries for temporarily unavailable event storage.

* **Bug Fixes**
  * Improved batched, coordinated cleanup of expired event history.
  * Enhanced handling of query errors, stale connections, sequence gaps, and delayed notifications.
  * Removed legacy pod-to-pod notification behavior.

* **Documentation**
  * Updated SSE guidance with retention settings, deployment behavior, and known limitations.


qiffang (changes requested):
## Review verdict: REQUEST CHANGES

Reviewed exact head `4ad8aed5fe3b164b18d2c9e18e5b862bc05f71df`. Hosted `ci`, local-e2e, and CodeRabbit are green, but they do not cover the following correctness failures.

1. **Shared-pool retention sweep is globally throttled instead of per physical shared DB.** `Server.lastSharedSweepUnix` and the fixed meta claim name `fs_events_sweep` are shared across the whole server/fleet (`pkg/server/sse.go:399-436`), while `db_pool` explicitly supports multiple active physical shared databases, including multiple pools in one org. A busy pool can repeatedly win the claim and starve every other active pool, leaving expired rows in those physical tables unbounded. Scope the local pre-filter, cluster claim, metrics, and regression test by stable physical `db_id`/UUID; the current test only uses two `fs_id` values in one physical table.

2. **The retry buffer has cross-tenant head-of-line blocking and does not enforce its advertised caps while entries are in flight.** `flushDue` removes every due entry from `entries`/`tenantCounts`, then flushes them serially (`pkg/server/event_retry.go:238-267`), with up to 30 seconds per entry. One failed tenant can delay healthy tenants for hours. While those entries are out of the queue they are absent from the 10k/1k caps and depth metric, so concurrent enqueues can double actual residency; requeue then evicts newer queued entries and creates avoidable hard loss. Keep in-flight entries accounted for and isolate/bound retry work across tenants, with a deterministic slow/bad-tenant versus healthy-tenant regression.

3. **Retry can permanently use a stale/closed tenant store.** `loadStore` only reads the `EventBus` cached pointer (`pkg/server/event_retry.go:110-116,281`), and that pointer is refreshed only by a later tenant request (`pkg/server/sse.go:208-220`). Standalone pools idle-close after 5 minutes and shared handles/views after their idle TTL; retry neither pins the lease nor reacquires by tenant ID. If no new request arrives, every retry can hit the closed store until max-age hard loss even after the DB is healthy. Reacquire/refresh the current store for retry and cover eviction/close followed by recovery without new tenant traffic.

4. **A query error on the first poll of a notify wake consumes the only wake and never retries.** The Phase-2 error branch arms catch-up only when `fromCatchup || fullPages > 0` (`pkg/server/sse.go:703-721`), while `case <-notify` calls `pollAndSend(false)`. If that first query fails, default liveness polling is off and the durable event stalls until another write or reconnect. The existing notify-wake test fails only after one full page, so it misses this case. A notify-triggered first-poll error must retain/re-arm work with backoff.

5. **A fully pruned event table reports the wrong customer recovery reason.** `EventsSinceE` returns `headSeq=0, ok=false` for an empty table after all rows were pruned (`pkg/server/eventbus.go:270-273`); `handleEvents` then maps any positive `since > headSeq` to `server_restart` (`pkg/server/sse.go:529-535`). This retention-loss case must be `seq_too_old` per the documented recovery contract. The existing full-prune test asserts only `event=reset`, not the reason; add the discriminating assertion.

6. **A query error on the initial Phase-1 replay is falsely announced as caught up.** The first call uses `bus.EventsSince`, which discards `qErr` (`pkg/server/sse.go:509-520`). `EventsSinceE` maps any `ListFSEventsSince` error to `ok=true, events=[], headSeq=since` plus an error (`pkg/server/eventbus.go:196-235`), so a reconnect with an existing backlog receives a heartbeat at the old cursor and enters Phase 2. If the outbox wake preceded the connection and there is no later write, the backlog stalls indefinitely. Compatibility with a pre-migration missing table must not mask all transient query/connectivity failures as a successful replay; distinguish the compatibility case or terminate/retry the initial replay, and add a first-call regression.

Validation on this exact head: `git diff --check`, changed-Go `gofmt -l`, `go test ./pkg/metrics -count=1`, and `make lint` pass. A temporary discriminating `pkg/server` retry test compiled via `go test -c`; running package tests locally is blocked before test execution by package `TestMain` requiring unavailable rootless Docker/testcontainers.

srstack (approved):
Reviewed the full diff (22 files, +3047/−322) at `4ad8aed5` against merge-base `e9fae58b`, in a detached worktree. Build, `go vet`, `gofmt` on all changed files, and `go test ./pkg/server/ ./pkg/datastore/ ./pkg/meta/ ./pkg/metrics/` all pass.

This is careful work. The parts I checked hardest and found correct:

- **`EventsSinceE` split.** Preserving the tolerated `ok=true, empty` fallback for the *first* Phase-1 call (pre-migration tenants must not reconnect-loop) while making the *mid-drain* calls fail closed is exactly the right asymmetry. A masked error mid-drain would falsely mark a still-behind client caught up, and terminating without the end-of-replay heartbeat so the client resumes from its durable cursor is the correct recovery.
- **`DeleteSharedFSEventsBefore` shape check.** An unscoped `DELETE` reachable from a dedicated tenant store would be catastrophic; refusing non-shared stores at the datastore layer (not just at the caller) is the right place for that guard, and `TestDeleteSharedFSEventsBeforeRefusesStandalone` pins it.
- **`ON DUPLICATE KEY UPDATE` assignment order.** The left-to-right evaluation dependency is genuinely load-bearing and genuinely non-obvious — `updated_at` must be computed before `last_id` is overwritten or `VALUES(last_id) > last_id` compares the new value to itself. The inline `NOTE` earns its place, and `TestUpsertTenantOutboxCursorRefreshesUpdatedAt` covers it.
- **Retry-buffer caps.** Per-tenant sub-cap applied before the global cap is the right order: it stops one tenant's sustained DB outage from evicting healthy tenants' events. Every eviction path counts `drive9_sse_event_retry_dropped_total` with a distinct reason, so hard loss is observable rather than silent.
- **`stopNotifyInfrastructure` ordering.** Stopping `eventRetry` *before* the coalescer's final flush is subtle and correct — the buffer's second-wake signals go through `insertTenantNotify`, so they must land in the coalescer before it drains. The comment says why.
- **`stopped` flag under `mu`.** Racing enqueues either land before the final flush or are rejected and counted; the flag is checked under the same mutex that `stop` sets it under. No window.
- **`eventRetryFlushTimeout`.** Bounding the per-entry insert so a hung tenant-DB conn cannot block the flush goroutine, and thereby `wg.Wait()` in `stop()`, and thereby the whole server `Close` — good catch to have made preemptively.
- **`CountFSEvents` cap.** Wrapping in a `LIMIT`-ed subquery keeps an observability gauge from becoming a full-table scan over millions of rows. Description updated to say "capped at 100000" so the gauge is not misread.
- **Endpoint removal is complete.** Handler, request type, route registration, `requestRoute` case, `PodNotifySecret` config field, and the `DRIVE9_POD_NOTIFY_SECRET` env doc are all gone; the only surviving reference is the design doc recording the removal. Stale `podNotifier` comments in `eventbus.go`/`pod_registry.go` were updated rather than left to rot.

Findings, all non-blocking:

**`[MINOR]` The shared-pool sweep claim is a single global row, but multiple `role='shared'` pools coexist.**

`sharedFSEventsSweepClaimName` is the constant `"fs_events_sweep"`, and `ClaimSharedMaintenanceRun` keys `shared_maintenance_state` on that name alone. `Server.lastSharedSweepUnix` is likewise one atomic per pod. But `FindSharedDBForAllocation` (`pkg/meta/db_pool.go:849`) selects among *many* `role='shared'` rows per org — a new pool is provisioned whenever the current one hits `soft_cap_reached` or `tenant_count >= max_tenants` (default 100). So a deployment with, say, 5 shared pools has 5 independent physical `fs_events` tables contending for one claim row: whichever pool's tenant triggers first wins the 30-minute window, and the other four are skipped entirely for that window.

The rows are not lost — over-retention has no correctness cost, and a later trigger from another pool will eventually win a window — but the effective sweep interval per pool degrades to roughly `30min × pool_count` under uniform triggering, and it is *unfair*: a hot pool triggers more often and so wins more windows, while the coldest pool (which is exactly where dead tenants' unreachable rows accumulate, i.e. the growth the pool sweep exists to bound) gets swept least. That inverts the intent in the design doc's shared-table storage policy.

The fix is small and local: include the pool identity in both throttles — claim name `fmt.Sprintf("fs_events_sweep:%d", dbID)` and a `map[int64]*atomic.Int64` (or `sync.Map`) for the pre-filter. `shared_maintenance_state.name` is already `VARCHAR(64)` and the PK is `name`, so per-pool rows need no schema change. The wrinkle is that neither `maybeSweepSharedFSEvents` call site currently has the `dbID`: `*datastore.Store` carries only `Scope{fsID, shared}`, and `tenantTarget` has no placement. Threading it would mean either exposing the pool id on the store (it is already known at `sharedDBHandle(ctx, placement.DbID)` in `pkg/tenant/pool.go:1358`) or passing `placement.DbID` through to both call sites. That is a real amount of plumbing for a bounded-growth issue, so I would not block on it — but it should be a tracked follow-up rather than left implicit, since the current code reads as if there were one shared table.

**`[MINOR]` `flushDue` reuses a single `now` across a serial loop that can run for minutes.**

`event_retry.go:239` captures `now := time.Now()` once, then the flush loop at 258 calls `b.requeue(e, now)` for every failed entry. Each `flushEntry` is bounded by `eventRetryFlushTimeout = 30s`, so with several entries stalled on a hung tenant DB the `now` handed to `requeue` can be minutes stale by the last entry. Two consequences: `e.nextRetry = now.Add(backoff)` is computed from a stale base, so a late entry's next attempt is due earlier than intended (possibly already past-due, making it eligible on the very next 1s scan and shortening its intended backoff); and the `now.Sub(e.enqueuedAt) > b.maxAge` check at 326 uses the stale clock, so an entry that aged out during the loop survives one extra cycle. Neither is harmful — the retry is idempotent-ish at the DB level and the maxAge slip is one scan interval against a ≥1h budget — but `time.Now()` inside `requeue` (or per-iteration in the loop) would be both simpler to reason about and strictly more correct. Worth noting the *reason* for the current shape is not stated in the comment, which is what made me stop on it.

**`[NIT]` Dead shift range in two backoff helpers.** `eventRetryBaseBackoff << min(e.attempts, 10)` (`event_retry.go:320`) clamps the shift to 10, but `1s << 9` = 512s already exceeds `eventRetryMaxBackoff = 5m`, so attempts ≥ 9 all land on the cap and shifts 9–10 are unreachable-in-effect. Same shape in `sseCatchupBackoff`: `10ms << min(n-1, 10)` reaches `sseCatchupMaxPollDelay = 1s` at n=8, so the clamp at 10 is loose. Harmless (the `min` correctly prevents overflow, which is the actual job) and the explicit clamp is arguably good defensive practice, so this is purely cosmetic.

**`[NIT]` `logger.Info` at `server.go:601` resolves to the shadowed local `*zap.Logger`, not `pkg/logger`.** `NewWithConfig` declares `logger := cfg.Logger` at line 440, so this call uses zap's 2-arg `Info(msg, fields...)` rather than the package's ctx-first `logger.Info(ctx, msg, fields...)`. It compiles and logs correctly, and the neighbouring `server_leader_standby` call at 781 does the same, so it is consistent with local style — just flagging that the identical-looking call in `sse.go` means something different, which is a readability trap for the next reader.

On the two SQL semantics the PR body flags as MySQL 8.0-verified only (`ON DUPLICATE KEY UPDATE` left-to-right assignment evaluation, and `DATE_SUB(NOW(3), INTERVAL ? SECOND)` with a placeholder): agreed those need real-TiDB verification before the freshness-bound prune floor and the cluster claim are enabled in production, and I am glad that is already written into the design doc's pre-production checklist rather than living only in the PR description. Both failure modes are graceful if TiDB differs — a never-refreshing `updated_at` degrades the prune floor to age-only, and a broken interval comparison makes the claim always-or-never win — so shipping behind the checklist is reasonable.

Test coverage is genuinely proportionate to the risk here: multi-page Phase-1 drain, mid-drain error termination, Phase-2 event-driven drain, catch-up burst, query-error backoff resumption, all six retry-buffer drop reasons, batched delete, the shape-check refusal, the claim's won/lost/pre-filtered/error paths, and the cursor-freshness floor. The `var`-instead-of-`const` seams (`eventPageSize`, `sseMaxFullPagesPerWake`, `eventsSinceE`) each carry the "restore via t.Cleanup, no t.Parallel" warning, which is the right way to document a process-global test hook.

Approving. Please open a follow-up for the per-pool sweep claim.

mornyx:
Thanks for the careful review — all six items are addressed in `48cb669` (pushed). Point by point:

**1. Shared-pool sweep is now throttled per physical pool.** `datastore.Scope` carries the physical `db_pool.db_id` (`SharedScopeWithDB`, wired at both construction sites in `pkg/tenant/pool.go` via `sharedDBHandleLease.DBID()`). The claim name is `fs_events_sweep:<dbID>`, the per-pod pre-filter is a `sync.Map` keyed by dbID (with the claim-error restore preserved per pool), and metric labels are `shared_pool[:<dbID>]`. Each pool's claim is an independent row in `shared_maintenance_state` (no schema change), so a hot pool can no longer starve the cold pools. The orchestration test now covers two pools winning within the same 30-minute window; dbID=0 falls back to the plain `fs_events_sweep` name, so single-pool/test construction is unchanged.

**2. Retry buffer: in-flight accounting + per-tenant isolation.** Entries stay in the queue — counted against the 10k/1k caps and the depth gauge — *while* flushing: a `flushing` flag marks them, eviction drops the oldest **non-flushing** entry, and failures update backoff in place, so the queue stays enqueuedAt-ordered by construction. No cap evasion, no keep-old-drop-new. Flush work is sharded per tenant: one FIFO flusher per active tenant, bounded by 8 global slots, so a hung tenant DB costs at most its own per-entry 30s timeout and cannot delay other tenants. Deterministic regression: a blocked bad tenant in flight while a healthy tenant's entries flush promptly, with depth ≤ cap throughout.

**3. Stale/closed store healed by re-resolution.** The buffer takes a server-injected resolver: flush tries the bus's cached pointer first; on nil (idle-closed pool) or a closed-DB insert error (`sql.ErrConnDone` / "database is closed") it re-resolves once via the server's normal acquisition path (`meta.GetTenant` + `pool.Acquire`; single-tenant mode falls back to `fallback.Store()`), refreshing the bus pointer as a side effect. A tenant that no longer exists (not found / non-active) drops its entries with reason `tenant_gone` instead of spinning to max-age. Tests: close/evict → DB recovers → retry lands with no new tenant traffic, the tenant_gone drop counter, and an integration test proving acquire + bus refresh + real insert.

**4. Notify-wake first-poll error re-arms.** `pollAndSend` now receives its wake source (notify / liveness / catchup). On a query error it re-arms with the bounded backoff whenever the wake indicates real pending work — catch-up fire, full pages already delivered, **or a notify** (a real signal: a durable row was just written) — while a liveness-driven first-poll failure still starts no timer. New regression: notify wake, *first* poll fails, no further notify → tail still drains via the timer (gate validated by temporarily reverting the condition: stalls; restored: passes).

**5. Fully-pruned table → `seq_too_old`.** The Phase-1 reason mapping now emits `seq_too_old` whenever `since > 0 && headSeq == 0` (empty table = everything pruned); `server_restart` is reserved for a cursor ahead of a non-empty head. The full-prune test now asserts the reason explicitly.

**6. Initial Phase-1 error distinguished from pre-migration.** The first call also goes through `EventsSinceE`; `datastore.IsMissingTableError` (MySQL 1146) keeps the tolerated caught-up fallback for pre-migration tenants, while any other query/connectivity error terminates the stream *without* the caught-up heartbeat — the client reconnects and resumes from its durable cursor. Tests cover both the transient-error termination and the missing-table tolerance.

Also folded in from srstack / Codex / CodeRabbit: fresh clock per flush attempt (the in-place design subsumes the stale-`now` note), the requeue age-order inversion (eliminated by construction), `%w` wrapping on `ClaimSharedMaintenanceRun`, `ResetMetaDB` covering the outbox/cursor/maintenance tables (bespoke helper deleted), atomic test seam counters, tightened backoff shift clamps, the logger-shadow rename, and detached sweep goroutines now tied to the server lifecycle (`sweepCtx` canceled on Close, tracked by `notifyWG`, ctx checked between delete batches — Close can no longer block on the 5-minute sweep timeout, and sweeps can't hit a closed meta store).

Validation on `48cb669`: `go build ./...`, `make lint` (0 issues), `go test -race` on the new retry/SSE/shared-sweep tests, and full `pkg/server`, `pkg/datastore`, `pkg/meta`, `pkg/tenant/...` suites — all green.

qiffang:
**Dev-Lead review (architecture consistency + new-authority-path both-directions + tenant isolation).** Adversary reviewers cover edge/failure exhaustively; this focuses on whether the persist→retain→read→expire pipeline is complete, consistent, and tenant-safe. **Verdict: looks sound — no blockers.** 3 low-severity observations + 1 pre-prod gate the PR already flags.

### ✅ Both directions wired (persist ↔ retain/read ↔ expire) — the new-authority-path axis
- **Producer**: `publishEvent` (`pkg/server/sse.go`) inserts the durable `fs_events` row; on insert failure it enqueues into the new retry buffer (`s.eventRetry.enqueue`) instead of silently dropping. The retry buffer's successful flush does the **mandatory second wake** (`insertTenantNotify` + `bus.Publish` in `event_retry.go` `flushEntry`), so late-inserted rows reach both cross-pod and same-pod subscribers — the prior silent-drop hole is closed.
- **Consumer**: `handleEvents` Phase-1 loops `EventsSinceE` until a short page (drains multi-page backlogs); Phase-2 is event-driven catch-up. Read (`EventsSinceE`), retain (both sweep entry points), and expire (`DeleteFSEventsBefore`) all read the SAME `s.fsEventsRetention`. The old 1h constant is gone; `defaultFSEventsRetention` + `Config.FSEventsRetention` + `tenantWorkerOpts.FSEventsRetention` are all plumbed from one env var with a startup log; the tenant-worker backstop (`piggybackMaintenance`) uses `m.opts.FSEventsRetention` (not the deleted constant). Build(write)/serve(read)/expire all wired to one config value.

### ✅ Tenant isolation / residency
- Per-tenant sweep `DeleteFSEventsBefore` uses `s.scope.And("created_at < ?")` → `fs_id = ? AND created_at < ?`, correct positional args.
- Pool-wide `DeleteSharedFSEventsBefore` deliberately omits `fs_id` and is **guarded by `if !s.scope.Shared() { return error }`** (pkg/datastore/fs_events.go ~L447) — an unscoped DELETE can never fire against a dedicated tenant DB; `TestDeleteSharedFSEventsBefore` + `...RefusesStandalone` cover it. `dbID` threaded into scope via `SharedScopeWithDB` at both backend-construction sites. No cross-tenant read/delete path found.

### ✅ Retention correctness (real purge, not filter-on-read)
- Real batched `DELETE ... LIMIT ?` on `idx_fs_events_created`/`idx_fs_events_fs_created`, `hasMore` semantics, `ctx.Err()` between batches (prompt shutdown). Strict `created_at < cutoff`; boundary handled — `TestSSEResetFlushWhenSeqTooOld` now asserts a fully-pruned table reports `seq_too_old` (not `server_restart`) at the `headSeq==0` boundary. Sweep failures are metrics+warn only, never touch non-expired rows, and the read path's independent `oldest` check means a failed sweep can't serve expired events.

### ✅ Identity/seq consistency (documented trade-off, not a bug)
- `seq` is DB-assigned at INSERT, so a buffered/retried event sequences at flush time → bounded reorder relative to events that persisted while it was buffered. This is explicitly designed + documented (`seq` = persistence order; consumers treat events as idempotent hints), matching the existing FUSE-cache-invalidation contract. path/op/actor/ts carried verbatim through the buffer entry; tenant keyed on `bus.tenantID`/`orgID`. No identity mislocation.

### ✅ Schema/migration
- One new meta-DB table `shared_maintenance_state` via the existing locked diff-migrator (`CREATE TABLE IF NOT EXISTS`); old pods ignore it (mixed-version safe). Tenant `fs_events` is structurally **unchanged** — no tenant migration, read path identical pre/post.

### 🟡 Low-severity (comment, not blocking)
1. **`resolveRetryStore` uses the store after `release()`** (`event_retry.go` ~L1834-1838): `pool.Acquire` → `defer release()` → returns `store`, which `flushEntry` uses for `InsertFSEvent` after the refcount hit 0. If the entry is concurrently retired at that instant, the insert can hit `database is closed`. **Self-healing** (closed-DB error → one-time re-resolve + `SetStore` refresh → requeue; no data loss), but it differs from the request path (which holds the lease for the op's lifetime). Consider holding the lease until after the insert, or a comment noting the closed-DB retry is the intended guard.
2. **Shared-pool per-pod pre-filter set before the claim resolves** (`maybeSweepSharedFSEvents`): first trigger `LoadOrStore(dbID, now)` marks the pool swept-at-now, then may lose the cluster claim; the pre-filter stays 30 min even if this pod didn't delete. Intended backoff, but if the winner then FAILS its DELETE, the pool is unswept up to 30 min (harmless — over-retention by design, next interval retries). Coupling the pre-filter to "someone claimed" rather than "someone swept" is worth a one-line comment.
3. **`DELETE ... LIMIT` without `ORDER BY`**: correctness fine (any `created_at < cutoff` row equally eligible; index-efficient), but batched delete order is unspecified on TiDB — noting the design already accepts leftover-drains-next-cycle.

### ⚠️ Pre-prod gate the PR already flags — confirm it's on the deploy checklist
Two SQL semantics are validated on MySQL 8.0 only: the `ON DUPLICATE KEY UPDATE` assignment order in `UpsertTenantOutboxCursor`, and the `INTERVAL ? SECOND` placeholder in `ClaimSharedMaintenanceRun`. Run `TestUpsertTenantOutboxCursorRefreshesUpdatedAt` and `TestClaimSharedMaintenanceRun` against **real TiDB** before enabling in prod — a silent TiDB divergence there would either pin the outbox (stalled prune) or let N pods sweep the same pool.

Net: the producer↔consumer↔expire triangle is complete and consistent, tenant isolation is enforced on the correct (fs_id / shape-guarded) axis, and the seq reorder is a deliberate, documented, idempotent-consumer contract. No blockers from the architecture/consistency angle.

qiffang:
## Independent re-review — adversary-2 (PR #826, head `48cb6698`)

Re-reviewed at the current head (not carrying any prior conclusion). The head moved from the 6-blocker REVISE head `4ad8aed5` → `48cb6698` (+1074/-243, 16 files — a substantial fix pass). **All 6 previously-dual-confirmed blockers are FIXED in code, each with a discriminating regression test. GREEN.**

I verified blocker #6 at source myself; blockers #1–#5 via a focused independent code+test read.

1. **Shared-DB sweep starvation + unguarded pool-wide delete** — FIXED. `maybeSweepSharedFSEvents` keys both the per-pod pre-filter and the cluster claim **per physical pool** (`sharedFSEventsSweepClaimName:<dbID>`, sse.go:449) → a hot pool can't monopolize. `DeleteSharedFSEventsBefore` guarded by `if !s.scope.Shared() { return err }` (fs_events.go:195) so the unscoped DELETE is unreachable on a dedicated DB; per-tenant delete carries `fs_id=?`. Tests: `TestDeleteSharedFSEventsBeforeRefusesStandalone`, `TestMaybeSweepSharedFSEventsPerPoolClaims`.

2. **Retry serial HOL + in-flight bypasses cap** — FIXED. Per-tenant FIFO flush goroutines gated by a global semaphore (no cross-tenant HOL); `flushing` entries stay resident and counted against `tenantCounts`, eviction paths skip in-flight → buffer genuinely bounded. Tests: `TestEventRetryInFlightResidencyAndEvictionSkip`, `TestEventRetryPerTenantIsolation`.

3. **Retry holds stale/closed store** — FIXED. `flushEntry` re-resolves via `pool.Acquire` on `isClosedDBError` (sql.ErrConnDone / "database is closed") and retries on the fresh handle; no stale handle held across the idle-reap window. Test: `TestEventRetryResolverHealsClosedStore`.

4. **Notify first-poll error swallows the only wake** — FIXED. Re-arm condition now `wake == sseWakeCatchup || fullPages > 0 || wake == sseWakeNotify` (sse.go:794) — the added `sseWakeNotify` clause closes the notify-driven-first-poll (fullPages==0) drop. Test: `TestSSEEndpointPhase2NotifyFirstPollErrorStartsCatchup`.

5. **Full prune misreports server_restart** — FIXED. `since>0` → `seq_too_old` by default, `server_restart` only when `headSeq>0 && since>headSeq` (sse.go:598). Tests pin both branches: `TestSSEResetFlushWhenSeqTooOld`, `TestSSEResetWhenFutureCursor`.

6. **Phase-1 first-replay query error masquerades as caught-up** (I confirmed this one at source earlier) — FIXED. Phase-1 uses `eventsSinceE` (returns err); `firstErr != nil && !IsMissingTableError(firstErr)` → `return` without a caught-up heartbeat, client resumes from durable cursor. Correctly separates "table missing = pre-migration, tolerate" from "transient error = must not claim caught-up" (sse.go:580-607).

### Caveats (not merge-blockers)
- Verified by reading code + test source; `pkg/server` tests can't run in my environment (testcontainers `TestMain`). Hosted CI is the runtime evidence.
- The PR self-flags a pre-launch gate: two SQL semantics verified only on MySQL 8.0 → run `TestUpsertTenantOutboxCursor...` and `TestClaimSharedMaintenanceRun` on real TiDB before launch. Legitimate deploy-gate, not a merge blocker.

**GREEN.** Note: GitHub still shows `reviewDecision=CHANGES_REQUESTED` because that formal state was set on the old head — the original blocking reviewers (dev2, adversary-1) need to refresh their review against `48cb6698` to clear it. My GREEN is code-review; qiffang owns the merge on this repo.

qiffang (approved):
## adversary-1 exact-head re-review: GREEN

Re-reviewed current head `48cb669841fbb426c14c3658c2c87c30abb9b291`, replacing my earlier REQUEST_CHANGES on `4ad8aed5`. The six prior blockers are closed in code and each has a discriminating regression path.

1. **Shared physical DB sweep starvation** is closed: `datastore.Scope` now carries `DBID`, production shared-store construction uses `SharedScopeWithDB`, and `maybeSweepSharedFSEvents` keys both the local pre-filter and meta claim as `fs_events_sweep:<dbID>` when the physical pool id is known. The unscoped shared delete is still shape-guarded by `!s.scope.Shared()`.
2. **Retry cross-tenant HOL / cap evasion** is closed: flush is per tenant with a global slot cap, in-flight entries remain resident/countable, and eviction skips `flushing` entries rather than removing work from accounting.
3. **Retry stale/closed store** is closed: cached store is attempted first, nil/closed DB paths re-resolve through the server tenant acquisition path and refresh the bus pointer; gone tenants drop distinctly instead of spinning.
4. **Notify first-poll query error swallowing the only wake** is closed: Phase 2 now passes an explicit wake source and re-arms catch-up on `sseWakeNotify` even when `fullPages == 0`.
5. **Fully-pruned table recovery reason** is closed: positive cursors with `headSeq == 0` now emit `seq_too_old`; `server_restart` is limited to cursor-ahead-of-non-empty-head.
6. **Initial Phase-1 query error false caught-up heartbeat** is closed: Phase 1 uses `EventsSinceE`; non-missing-table errors terminate without a caught-up heartbeat, while missing table remains the tolerated pre-migration case.

Residuals are not merge blockers:
- CodeRabbit's `shared_sweep_test.go` storeA/storeB suggestion is a readability improvement. The helper counts the whole shared test table through the same DB handle, so it still fails if the pool-102 row remains; using `storeB` would make the assertion clearer.
- CodeRabbit's fresh context for the re-resolved retry insert is a valid robustness nit, but a failed second insert requeues with backoff and remains bounded/accounted; it does not reopen the old data-loss or stale-store blocker.
- The PR's own pre-production TiDB gate remains: run the outbox cursor and shared-maintenance claim SQL tests on real TiDB before enabling in production.

Validation: GitHub hosted `ci`, `drive9-server-local e2e smoke`, and CodeRabbit are green on this head. Local `git diff --check` and `go test ./pkg/metrics -count=1` pass; `go test -c ./pkg/server` compiles. Local runtime `pkg/server`/`pkg/datastore` selected tests are still blocked before execution by this machine's missing rootless Docker/testcontainers, not by a test failure.

PR #827, opened 2026-08-03
fix: make Tencent CAM credential refresh resilient

qiffang:
## Summary

- retry Tencent CAM metadata timeouts, 429 responses, and 5xx responses with bounded per-request deadlines and backoff
- refresh credentials five minutes early, deduplicate concurrent refreshes, and fall back only to credentials with at least one minute of real validity remaining
- expose exhausted transient CAM metadata failures as retryable HTTP 503 responses with `Retry-After`, while keeping client errors sanitized
- add separate CAM role-list and credential-fetch operation results so metadata failures are distinguishable from COS data-plane failures

## Tests

- `go test ./pkg/s3client -count=1`
- `go test -race ./pkg/s3client -run 'TestCAM|TestCredentialsForTencentWrapsCAMProviderWithCache' -count=1`
- `go test -c ./pkg/server -o /tmp/drive9-server-cam.test`
- `make lint`
- `git diff --check`

Focused `pkg/server` execution is blocked on this workstation because its `TestMain` requires unavailable rootless Docker/testcontainers; the package and tests compile successfully.

## Summary by CodeRabbit

* **Reliability**
  * Improved Tencent CAM credential retrieval with configurable metadata endpoints, retries, timeouts, and response validation.
  * Safely reuses valid cached credentials during temporary metadata outages while preserving expiration handling.
  * Prevents duplicate credential refreshes during concurrent requests.

* **Bug Fixes**
  * Temporary credential-service outages now return HTTP 503 responses with retry guidance.
  * Improved handling distinguishes temporary failures from permanent metadata errors.

### DIFF
--- a/pkg/s3client/tencent.go
+++ b/pkg/s3client/tencent.go
@@ -3,8 +3,10 @@ package s3client
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
 	"io"
+	"net"
 	"net/http"
 	"os"
 	"strings"
@@ -16,6 +18,19 @@ import (
 )
 
 const camMetaBase = "http://metadata.tencentyun.com/latest/meta-data/cam/security-credentials/"
+
+var ErrCAMMetadataUnavailable = errors.New("cam metadata temporarily unavailable")
+
+const (
+	camMetadataRequestTimeout = 2 * time.Second
+	camMetadataMaxAttempts    = 3
+	camMetadataRetryBackoff   = 100 * time.Millisecond
+	camCredentialRefreshAhead = 5 * time.Minute
+	camCredentialFallbackMin  = time.Minute
+
+	camMetadataListRolesOperation      = "cam_metadata_list_roles"
+	camMetadataGetCredentialsOperation = "cam_metadata_get_credentials"
+)
 
 func isTencentEndpoint(endpoint string) bool {
 	if len(endpoint) == 0 {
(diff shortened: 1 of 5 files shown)


qiffang:
REVISE at exact head `b652fb56543ba23857e9137387d9ccbfac238517`.

Blocking issue:

- `pkg/s3client/tencent.go:123-132` disables the AWS `CredentialsCache` expiry adjustment whenever fallback credentials are returned, and `pkg/s3client/tencent_test.go:113-158` explicitly asserts that the fallback is cached until the real CAM expiration. That breaks the safety-margin contract. A single metadata timeout at the early-refresh point can put old credentials back into the AWS cache with `Expires == actualExpiry`, so subsequent presign calls will keep using them through the final minute without another metadata refresh attempt. The provider only checks `fallbackMin` at the moment fallback is returned; after that, the outer cache serves the same fallback credential until actual expiration. This can produce presigned URLs backed by credentials with seconds left and hides a sustained metadata outage until the token is effectively expired. The fallback cache expiry must be capped no later than `actualExpiry - fallbackMin` (or an equivalent safe revalidation point), and there needs to be a regression proving that fallback credentials are not served once remaining lifetime is at or below the safety margin.

Checks run locally:

- `git diff --check origin/main...HEAD` pass
- changed Go `gofmt -l` pass
- `go test ./pkg/s3client -count=1` pass
- `go test -race ./pkg/s3client -run 'TestCAM' -count=1` pass
- `go test ./cmd/drive9-server -count=1` pass
- `go test ./pkg/server -run 'TestBackendErrorStatus|TestWriteBackendErrorCAMMetadataUnavailableIsRetryable' -count=1` blocked before selected tests by package `TestMain`: rootless Docker/testcontainers unavailable

Note: I attempted a formal GitHub request-changes review, but GitHub rejected it because the available GitHub identity is the PR author, so this is posted as a blocking review comment instead.

qiffang:
Code-level GREEN for exact head `9fb2e6ddb83f14689d6148783b6fd51156e05f54`.

The prior blocker is closed: fallback credentials returned through the AWS cache now expire at `actualExpiry - fallbackMin`, so the cache re-enters refresh before the final safety minute instead of serving old CAM credentials until real expiration. The updated regression asserts the synthetic safety cutoff, and the provider near-expiry/expired cases still reject fallback.

Local validation passed: `git diff --check origin/main...HEAD`, changed Go `gofmt -l`, `go test ./pkg/s3client -count=1`, `go test -race ./pkg/s3client -run 'TestCAM' -count=1`, `go test ./cmd/drive9-server -count=1`, `go test -c ./pkg/server`, and `make lint`.

Caveats: hosted `ci` and `drive9-server-local e2e smoke` were still pending when reviewed; `go test ./pkg/server -run 'TestBackendErrorStatus|TestWriteBackendErrorCAMMetadataUnavailableIsRetryable' -count=1` is still blocked locally before selected tests by package `TestMain` requiring rootless Docker/testcontainers. Formal GitHub approval failed because the available GitHub identity is the PR author, so this is recorded as a comment instead of an approval review.

mornyx (approved):
## Review: Approve

Synthesizing a few review passes here. CI is green, the design is sound, and the second commit fixed the key fallback-expiry issue. No blocking concerns — good to merge. A few non-blocking suggestions below, ordered by value.

### What looks right

- Error mapping order in `backendErrorStatus`: checking `ErrCAMMetadataUnavailable` before `context.DeadlineExceeded` is correct, since the CAM error wraps a deadline and would otherwise be misclassified as 504.
- Retry classification is clean: only transport timeouts / 429 / 5xx are retried; permanent 4xx neither retries nor falls back to stale credentials — the right call, since a missing role shouldn't be masked by reusing old creds.
- The `AdjustExpiresBy` trick (cutting fallback creds to `real expiry - fallbackMin`) works, and the sentinel survives the SDK error chain (`%w` + `SigningError.Unwrap`) so `errors.Is` reaches `backendErrorStatus` from a real S3 call.
- Concurrency dedup via CredentialsCache singleflight + provider mutex; splitting metrics into list-roles / get-credentials operations makes control-plane failures distinguishable from COS data-plane failures.
- Test coverage is solid: retry, fallback, near-expiry rejection, no-fallback-on-permanent-4xx, concurrent dedup, 503 + Retry-After.

### Non-blocking suggestions

1. **Presign TTL vs fallback credential lifetime (please confirm strategy).** Presigned URLs have a 10min TTL, but fallback credentials may have only 1–5min of real validity left. URLs signed during an outage can die mid-upload/download before their TTL. Consider rejecting presign when remaining validity < URL TTL (503), or clamping URL TTL to the credential's remaining validity. This partially exists in the normal path too (5min refresh-ahead < 10min TTL), but fallback amplifies it. If not addressed now, worth a follow-up issue.

2. **Post-fallback blind window (~4min).** After a fallback, the cache won't re-attempt refresh until `real expiry - fallbackMin`; if metadata recovers quickly, recovery is delayed, and if the outage persists past that point, every request pays the full retry cost (~12s worst case across both operations) before getting a 503. Consider caching fallback creds for a short probe interval (e.g. 30s–1min, capped at `real expiry - fallbackMin`) or adding a failure cooldown. Note the regression test currently asserts this blind-window behavior, so it would need updating too.

3. **HTTP 408 not treated as transient.** Rare from Tencent metadata, but it's a one-line addition to the retryable status check.

4. **Validate `Token` is non-empty** alongside `TmpSecretId`/`TmpSecretKey` — an empty token currently surfaces only at COS request time.

5. **Documentation nits worth a comment each:** (a) the `fallbackResult` field relies on an implicit contract that CredentialsCache calls `Retrieve` then `AdjustExpiresBy` on the same goroutine (true today via singleflight); (b) the provider's mutex is intentionally held across network I/O as a second-level dedup; (c) the inner `expiresAt - refreshAhead` check in `Retrieve` is mostly redundant with the outer cache's `ExpiryWindow` — a comment would save future readers some head-scratching.

6. **Minor:** `camMetadataUnavailableError.Error()` doesn't include the sentinel text, making it hard to correlate logs with 503s; the duplicated 503 branches in `writeBackendError`/`errJSONInternalStorage` could share a helper; the new `cam_metadata_*` metric labels may need to be added to Grafana dashboards (`docs/grafana/`); `Retry-After: 1` is consistent with existing usage but a slightly larger value (3–5s) would soften client retry storms during a real outage; a few tests use 20ms client timeouts that could flake on a slow CI box.

Nice work overall — merging as-is is fine; items 1–3 are the ones I'd at least track as follow-ups.

PR #830, opened 2026-08-05
feat(client): bounded 503 retry for recursive deletes across SDKs

mornyx:
## What

Ports the client side of [tidbcloud/fs#39](https://github.com/tidbcloud/fs/pull/39) (batched, race-aware recursive directory delete) into this repo, and aligns all language SDKs with the new server contract.

Server contract: a recursive delete (`DELETE /v1/fs{path}?recursive=1`) on a large tree may exhaust its per-request budget and answer **503 + `Retry-After: 2`** (`ErrDeleteIncomplete`). The server-side sweep is resumable and idempotent — re-issuing the same DELETE continues where it left off. Small trees still delete atomically in one request.

## Changes

- **pkg/client (Go)**: `RemoveAll`/`deleteCtx` retry 503 with exponential backoff (1s → 2s → 4s → 8s, honoring `Retry-After`, at most 4 retries) for recursive deletes only; ctx cancel aborts the backoff wait immediately. This also covers the CLI and FUSE, which both go through `pkg/client`.
- **clients/drive9-js**: same retry in `removeAll`.
- **clients/drive9-py / drive9-rs / drive9-swift / drive9-kotlin**: add `remove_all`/`removeAll` with the same retry semantics (these SDKs previously had no public recursive-delete API); retry loop releases the response/connection before sleeping so it returns to the pool.
- Unit tests in every SDK: 503→503→200 succeeds with exactly 3 requests; persistent 503 errors after exactly 5 requests; Go additionally covers non-recursive-no-retry and cancel-during-backoff.

## Compatibility

Old clients against the new server: no data risk. Large-tree deletes may surface a plain 503, but the partially-deleted tree stays consistent and re-issuing the same DELETE resumes the sweep (a server-side reconcile worker cleans up orphans regardless). Upgrading the client just makes the resume automatic.

## Testing

- Go: `pkg/client` retry tests pass (testcontainers/MySQL); `golangci-lint` clean.
- JS: `vitest` 72 passed (incl. new `remove_all.test.ts`, msw-based).
- Python: 51 unit tests pass (`responses`-mocked).
- Rust: 11 unit tests pass (local stub server); `cargo fmt` clean on touched files.
- Swift: 15 unit tests pass (existing MockHTTPServer).
- Kotlin: `:lib:test` 28 tests pass.

## Summary by CodeRabbit

* **New Features**
  * Added recursive deletion support across client libraries.
  * Recursive deletions retry temporary server-unavailable (503) responses up to four times.
  * Retries honor `Retry-After` guidance or use automatic backoff, improving reliability for large directory cleanups.
  * Retry delays are bounded, and cancellations remain responsive.
  * File paths containing query or fragment characters are encoded safely.
  * Swift bearer credentials are restricted to HTTPS or loopback HTTP.

* **Documentation**
  * Updated client documentation with recursive deletion usage and retry behavior.

### DIFF
--- a/clients/drive9-js/src/client.ts
+++ b/clients/drive9-js/src/client.ts
@@ -130,8 +130,22 @@ const DOWNLOAD_DIR_CONCURRENCY = 16;
 // continues where it left off.
 const REMOVE_ALL_MAX_RETRIES = 4;
 
+// REMOVE_ALL_MAX_RETRY_DELAY_MS caps the delay honored from a Retry-After
+// header so a pathological value (e.g. Retry-After: 999999) cannot block the
+// caller for days.
+const REMOVE_ALL_MAX_RETRY_DELAY_MS = 60_000;
+
 function sleep(ms: number): Promise<void> {
   return new Promise((resolve) => setTimeout(resolve, ms));
+}
+
+function removeAllRetryDelay(retryAfter: string, backoff: number): number {
+  if (/^\d+$/.test(retryAfter)) {
+    // Honoring 0 as an immediate retry is intentional: the server explicitly
+    // dictates the delay, and the loop is bounded to REMOVE_ALL_MAX_RETRIES + 1 requests.
+    return Math.min(Number(retryAfter) * 1000, REMOVE_ALL_MAX_RETRY_DELAY_MS);
+  }
+  return backoff;
 }
 
 interface Drive9Config {
@@ -257,7 +271,8 @@ export class Client {
 
   fsUrl(path: string): string {
     const p = path.startsWith("/") ? path : `/${path}`;
-    return `${this.baseUrl}/v1/fs${p}`;
+    const encodedPath = p.split("/").map((segment) => encodeURIComponent(segment)).join("/");
+    return `${this.baseUrl}/v1/fs${encodedPath}`;
   }
 
   vaultUrl(path: string): string {
@@ -533,11 +548,7 @@ export class Client {
         const retryAfter = (resp.headers.get("retry-after") || "").trim();
         // Consume and discard the body before retrying.
         await resp.text().catch(() => "");
-        let delay = backoff;
-        if (/^\d+$/.test(retryAfter)) {
-          delay = Number(retryAfter) * 1000;
-        }
-        await sleep(delay);
+        await sleep(removeAllRetryDelay(retryAfter, backoff));
         backoff *= 2;
         continue;
       }
--- a/clients/drive9-js/src/client.ts
+++ b/clients/drive9-js/src/client.ts
@@ -124,6 +124,16 @@ const DEFAULT_SMALL_FILE_THRESHOLD = 50_000;
 const DEFAULT_SERVER = "https://api.drive9.ai";
 const DOWNLOAD_DIR_CONCURRENCY = 16;
 
+// REMOVE_ALL_MAX_RETRIES bounds the 503 retry loop for recursive deletes. The
+// server answers 503 when a batched recursive delete exhausts its per-request
+// budget; the sweep is resumable and idempotent, so re-issuing the same DELETE
+// continues where it left off.
+const REMOVE_ALL_MAX_RETRIES = 4;
+
+function sleep(ms: number): Promise<void> {
+  return new Promise((resolve) => setTimeout(resolve, ms));
+}
+
 interface Drive9Config {
   server?: string;
   current_context?: string;
@@ -515,8 +525,25 @@ export class Client {
   }
 
   async removeAll(path: string): Promise<void> {
-    const resp = await fetch(`${this.fsUrl(path)}?recursive=1`, { method: "DELETE", headers: this.authHeaders() });
-    await checkError(resp);
+    const url = `${this.fsUrl(path)}?recursive=1`;
+    let backoff = 1000;
+    for (let attempt = 0; ; attempt++) {
+      const resp = await fetch(url, { method: "DELETE", headers: this.authHeaders() });
+      if (resp.status === 503 && attempt < REMOVE_ALL_MAX_RETRIES) {
+        const retryAfter = (resp.headers.get("retry-after") || "").trim();
+        // Consume and discard the body before retrying.
+        await resp.text().catch(() => "");
+        let delay = backoff;
+        if (/^\d+$/.test(retryAfter)) {
+          delay = Number(retryAfter) * 1000;
+        }
+        await sleep(delay);
+        backoff *= 2;
+        continue;
+      }
+      await checkError(resp);
+      return;
+    }
   }
 
   async copy(srcPath: string, dstPath: string): Promise<void> {
--- a/clients/drive9-py/drive9/client.py
+++ b/clients/drive9-py/drive9/client.py
@@ -232,6 +232,9 @@ class Client(TransferMixin, PatchMixin):
                             delay = secs
                     except ValueError:
                         pass
+                # Release the connection back to the pool before sleeping;
+                # requests only returns it once the body is consumed/closed.
+                resp.close()
                 time.sleep(delay)
                 backoff *= 2
                 attempt += 1
--- a/clients/drive9-py/drive9/client.py
+++ b/clients/drive9-py/drive9/client.py
@@ -25,6 +25,37 @@ def _parse_iso_datetime(s: str) -> datetime:
 # continues where it left off.
 _REMOVE_ALL_MAX_RETRIES = 4
 
+# _REMOVE_ALL_MAX_RETRY_DELAY caps the delay honored from a Retry-After header
+# so a pathological value (e.g. Retry-After: 999999) cannot block the caller
+# for days.
+_REMOVE_ALL_MAX_RETRY_DELAY = 60.0
+
+
+def _remove_all_retry_delay(retry_after: Optional[str], backoff: float) -> float:
+    """Decide the sleep before the next recursive-delete retry.
+
+    A valid Retry-After header (integer delta-seconds) is honored, clamped to
+    _REMOVE_ALL_MAX_RETRY_DELAY; honoring 0 as an immediate retry is
+    intentional — the server dictates the delay and the retry loop is bounded
+    to _REMOVE_ALL_MAX_RETRIES + 1 requests. A missing, unparseable, or
+    negative header falls back to the passed exponential backoff.
+    """
+    if retry_after is not None:
+        try:
+            secs = int(retry_after.strip())
+            if secs >= 0:
+                return min(float(secs), _REMOVE_ALL_MAX_RETRY_DELAY)
+        except ValueError:
+            pass
+    return backoff
+
+
+def _encode_fs_path(path: str) -> str:
+    """Percent-encode each segment of a drive9 path, preserving slashes."""
+    if not path.startswith("/"):
+        path = "/" + path
+    return "/".join(quote(segment, safe="") for segment in path.split("/"))
+
 
 class Client(TransferMixin, PatchMixin):
     """The Drive9 HTTP client."""
@@ -78,9 +109,7 @@ class Client(TransferMixin, PatchMixin):
             self.session.mount("https://", adapter)
 
     def _url(self, path: str) -> str:
-        if not path.startswith("/"):
-            path = "/" + path
-        return f"{self.base_url}/v1/fs{path}"
+        return f"{self.base_url}/v1/fs{_encode_fs_path(path)}"
 
     def _vault_url(self, path: str) -> str:
         if not path.startswith("/"):
@@ -223,17 +252,9 @@ class Client(TransferMixin, PatchMixin):
         while True:
             resp = self._request("DELETE", url)
             if resp.status_code == 503 and attempt < _REMOVE_ALL_MAX_RETRIES:
-                delay = backoff
-                retry_after = resp.headers.get("Retry-After")
-                if retry_after is not None:
-                    try:
-                        secs = int(retry_after.strip())
-                        if secs >= 0:
-                            delay = secs
-                    except ValueError:
-                        pass
-                # Release the connection back to the pool before sleeping;
-                # requests only returns it once the body is consumed/closed.
+                delay = _remove_all_retry_delay(resp.headers.get("Retry-After"), backoff)
+                # Close the response before sleeping so the connection is not
+                # held open for the whole backoff.
                 resp.close()
                 time.sleep(delay)
                 backoff *= 2
--- a/clients/drive9-py/drive9/client.py
+++ b/clients/drive9-py/drive9/client.py
@@ -44,7 +44,9 @@ def _remove_all_retry_delay(retry_after: Optional[str], backoff: float) -> float
         try:
             secs = int(retry_after.strip())
             if secs >= 0:
-                return min(float(secs), _REMOVE_ALL_MAX_RETRY_DELAY)
+                # Clamp in the integer domain first: float() can raise
+                # OverflowError for valid ints with hundreds of digits.
+                return float(min(secs, int(_REMOVE_ALL_MAX_RETRY_DELAY)))
         except ValueError:
             pass
     return backoff
--- a/clients/drive9-py/drive9/client.py
+++ b/clients/drive9-py/drive9/client.py
@@ -1,5 +1,6 @@
 """Drive9 HTTP client."""
 
+import time
 from datetime import datetime
 from typing import Optional
 from urllib.parse import urlencode, quote
@@ -16,6 +17,13 @@ from .stream import StreamWriter
 def _parse_iso_datetime(s: str) -> datetime:
     """Parse ISO8601 datetime, handling Z suffix for Python < 3.11."""
     return datetime.fromisoformat(s.replace("Z", "+00:00"))
+
+
+# _REMOVE_ALL_MAX_RETRIES bounds the 503 retry loop for recursive deletes. The
+# server answers 503 when a batched recursive delete exhausts its per-request
+# budget; the sweep is resumable and idempotent, so re-issuing the same DELETE
+# continues where it left off.
+_REMOVE_ALL_MAX_RETRIES = 4
 
 
 class Client(TransferMixin, PatchMixin):
@@ -199,6 +207,37 @@ class Client(TransferMixin, PatchMixin):
         """Remove a file or directory."""
         resp = self._request("DELETE", self._url(path))
         self._check_error(resp)
+
+    def remove_all(self, path: str) -> None:
+        """Remove a file or directory tree recursively.
+
+        A large tree delete may exceed the server's per-request budget; the
+        server then answers 503 (with an optional Retry-After header) and this
+        method retries with backoff, at most _REMOVE_ALL_MAX_RETRIES times.
+        Retrying is safe: the server-side sweep is resumable and idempotent,
+        so a repeated DELETE continues where it left off.
+        """
+        url = self._url(path) + "?recursive=1"
+        backoff = 1.0
+        attempt = 0
+        while True:
+            resp = self._request("DELETE", url)
+            if resp.status_code == 503 and attempt < _REMOVE_ALL_MAX_RETRIES:
+                delay = backoff
+                retry_after = resp.headers.get("Retry-After")
+                if retry_after is not None:
+                    try:
+                        secs = int(retry_after.strip())
+                        if secs >= 0:
+                            delay = secs
+                    except ValueError:
+                        pass
+                time.sleep(delay)
+                backoff *= 2
+                attempt += 1
+                continue
+            self._check_error(resp)
+            return
 
     def copy(self, src_path: str, dst_path: str) -> None:
         """Perform a server-side zero-copy."""
--- a/clients/drive9-rs/src/client.rs
+++ b/clients/drive9-rs/src/client.rs
@@ -14,6 +14,40 @@ const DEFAULT_SMALL_FILE_THRESHOLD: i64 = 50_000;
 /// re-issuing the same DELETE continues where it left off.
 const REMOVE_ALL_MAX_RETRIES: u32 = 4;
 
+/// Caps the delay honored from a Retry-After header so a pathological value
+/// (e.g. Retry-After: 999999) cannot block the caller for days.
+const REMOVE_ALL_MAX_RETRY_DELAY_SECS: u64 = 60;
+
+/// Percent-encode each segment of a drive9 path, preserving separators, so
+/// characters such as `?` and `#` cannot be reinterpreted as URL delimiters.
+fn encoded_fs_path(path: &str) -> String {
+    let normalized = if path.starts_with('/') {
+        path.to_string()
+    } else {
+        format!("/{path}")
+    };
+    normalized
+        .split('/')
+        .map(|segment| urlencoding::encode(segment).into_owned())
+        .collect::<Vec<_>>()
+        .join("/")
+}
+
+/// Decide the sleep before the next recursive-delete retry. A valid
+/// non-negative Retry-After value (integer delta-seconds) is honored and
+/// clamped to REMOVE_ALL_MAX_RETRY_DELAY_SECS; honoring 0 as an immediate
+/// retry is intentional (the retry loop is bounded). A missing or unparseable
+/// header falls back to the passed exponential backoff.
+fn remove_all_retry_delay(
+    retry_after: Option<&str>,
+    backoff: std::time::Duration,
+) -> std::time::Duration {
+    match retry_after.and_then(|v| v.trim().parse::<u64>().ok()) {
+        Some(secs) => std::time::Duration::from_secs(secs.min(REMOVE_ALL_MAX_RETRY_DELAY_SECS)),
+        None => backoff,
+    }
+}
+
 fn load_config_file() -> Option<(String, Option<String>)> {
     let home = std::env::var("HOME")
         .or_else(|_| std::env::var("USERPROFILE"))
@@ -40,14 +74,15 @@ fn load_config_file() -> Option<(String, Option<String>)> {
 }
 
 fn load_config() -> (String, Option<String>) {
-    let env_server = std::env::var("DRIVE9_SERVER").ok().filter(|s| !s.is_empty());
-    let env_key = std::env::var("DRIVE9_API_KEY").ok().filter(|s| !s.is_empty());
-    let (file_server, file_key) = load_config_file()
-        .unwrap_or_else(|| ("https://api.drive9.ai".to_string(), None));
-    (
-        env_server.unwrap_or(file_server),
-        env_key.or(file_key),
-    )
+    let env_server = std::env::var("DRIVE9_SERVER")
+        .ok()
+        .filter(|s| !s.is_empty());
+    let env_key = std::env::var("DRIVE9_API_KEY")
+        .ok()
+        .filter(|s| !s.is_empty());
+    let (file_server, file_key) =
+        load_config_file().unwrap_or_else(|| ("https://api.drive9.ai".to_string(), None));
+    (env_server.unwrap_or(file_server), env_key.or(file_key))
 }
 
 #[derive(Clone, Debug)]
@@ -106,12 +141,7 @@ impl Client {
     }
 
     pub(crate) fn fs_url(&self, path: &str) -> String {
-        let p = if path.starts_with('/') {
-            path
-        } else {
-            &format!("/{}", path)
-        };
-        format!("{}/v1/fs{}", self.base_url, p)
+        format!("{}/v1/fs{}", self.base_url, encoded_fs_path(path))
     }
 
     pub(crate) fn vault_url(&self, path: &str) -> String {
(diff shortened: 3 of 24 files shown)


qiffang (changes requested):
## Dev-lead review — PR #830 (bounded 503 retry for recursive deletes)

Reviewed exact head `9d5dd88` against merge-base `fb09fae`. Real delta = **20 files / +723 / −33** (client Go + 5 SDKs + tests). The core design is sound: retry is **recursive-only**, **bounded** (max 4 retries / 5 requests), **releases the connection before sleeping**, and relies on the documented resumable+idempotent server sweep (`ErrDeleteIncomplete`, tidbcloud/fs#39). All 6 SDKs are consistent (same bound, same gating, same Retry-After handling). Go tests are discriminating (503→503→200 = exactly 3 req; persistent 503 = exactly 5 req; non-recursive = 1 req/no-retry; cancel-during-backoff returns promptly). Nice.

Two things worth addressing before merge, one nit:

### 1. (all 6 SDKs) Honored `Retry-After` is **uncapped** → unbounded single wait

Every SDK, when the server sends a numeric `Retry-After`, sets `delay = retryAfter` with **no upper bound**:
- Go `client.go`: `delay = time.Duration(secs) * time.Second`
- JS `client.ts`: `delay = Number(retryAfter) * 1000`
- Py `client.py`: `delay = secs`
- Rust `client.rs`: `Duration::from_secs(v)`
- Swift `Drive9.swift`: `delayNs = UInt64(secs) * 1_000_000_000`
- Kotlin `Drive9.kt`: `retryAfterSecs * 1_000`

A server/proxy answering `Retry-After: 999999` makes the client wait ~11.5 days on a single retry. The exponential *fallback* (1→2→4→8s) is bounded, but the honored header value is not. Go/Rust/Kotlin can at least be cancelled (ctx/coroutine); **Py (`time.sleep`) and JS (no AbortSignal) cannot** — they block hard. The contract says `Retry-After: 2`, so this only bites on a misbehaving/hostile intermediary, but a `min(retryAfter, cap)` (e.g. cap at 30–60s) is cheap and standard. Tests only exercise `Retry-After` 0/1/30, so this path is currently unverified — please add a large-Retry-After case asserting the wait is clamped.

### 2. `Retry-After: 0` is honored as an immediate retry (no backoff)

`secs >= 0` (Go/Py/Swift/Kotlin) / `/^\d+$/` (JS) all accept `0` → `delay = 0` → back-to-back re-issue. Bounded to 5 requests so not a spin-forever, but a 503 storm with `Retry-After: 0` hammers the server 5× with zero spacing. If that's intended ("server says retry now"), fine — but consider flooring the honored delay at the current `backoff` (i.e. `delay = max(retryAfter, backoff)`) so a 0 still respects exponential spacing. At minimum, confirm 0-means-immediate is the intended contract.

### 3. (nit, Go) `defer resp.Body.Close()` inside the retry `for` loop

It's correct **today** — every retry branch closes the body explicitly and `continue`s, so the deferred close only ever registers once (in the terminal branch that immediately returns). But it's a latent footgun: any future `continue` path that forgets the explicit close would stack defers until function return. A tiny comment ("retry branches must close explicitly; this defer only covers the terminal path") would lock the invariant.

Minor observations (non-blocking): Retry-After per RFC may be an HTTP-date, not just delta-seconds — all SDKs parse delta-seconds only and fall back to exponential backoff on a date form, which is a safe degradation. Swift uses an inclusive `for attempt in 0...max` where the others use unbounded loops; it's correct (5 requests, terminal branch always returns/throws) but reads differently — a defensive `throw` after the loop wouldn't hurt.

Verdict: **REQUEST CHANGES on #1 (add the Retry-After cap + test)**; #2 wants a one-line confirmation of intent; #3 is a nit. Everything else — recursive-only gating, the bound, connection release, cross-SDK consistency, the Go test discrimination — is solid.

qiffang:
## adversary-2 review — `fix(client): release connection before retry backoff, harden retry tests`

Reviewed at head `9d5dd88` (diff vs merge-base `fb09fae`). **Verdict: GREEN / no blockers.** Core fix correct, cross-client consistent, tests discriminating. A few non-blocking notes below.

### ✅ Core fix (Go reference `pkg/client/client.go`)
- 503 on a **recursive** delete → **drain (`io.Copy(io.Discard, body)`) + `Close()` the body BEFORE the backoff timer**, then sleep → correct connection-release (draining preserves keep-alive reuse; closing without draining would break it).
- `Retry-After` honored (delta-seconds, `>= 0`, falls back to exponential `backoff` on missing/invalid).
- Context cancellation respected during backoff (`select { <-ctx.Done() → ctx.Err() }`).
- Bounded (`attempt < removeAllMaxRetries=4`); after the bound the 503 falls through to `readError` (surfaces the error, no infinite loop).
- Retry is **recursive-scoped only** — a non-recursive DELETE's 503 does not retry.
- `defer resp.Body.Close()` inside the loop is safe: retry iterations close explicitly before `continue`; only the terminal (return) iteration reaches the defer → no defer-in-loop accumulation.

### ✅ Cross-client consistency (the top risk for a 6-surface PR)
Go / py / js / rs / swift / kotlin all implement the same contract: `MAX_RETRIES=4`, `Retry-After` (delta-seconds + fallback), exponential backoff, **release connection before sleeping** (`resp.close()` / `conn.disconnect()` / drop-before-sleep), idempotent+resumable server sweep documented. Comments are consistent and accurate (server contract in tidbcloud/fs#39).

### ✅ Tests (Go `delete_retry_test.go`, discriminating)
- `RetriesOn503`: two 503s (Retry-After 0 then 1) → 200 → asserts exactly 3 requests.
- `RetryLimitExceeded`: always-503 → surfaces the last 503 as `*StatusError` after exactly `1+removeAllMaxRetries` requests (bound proven).
- `DeleteDoesNotRetryOn503`: non-recursive 503 → 1 request, no retry (scope boundary proven).

### 🟡 Non-blocking notes
1. **Retry-After parses delta-seconds only, not the RFC 7231 HTTP-date form.** A date-valued Retry-After silently falls back to exponential backoff. Fine if the server contract only sends seconds — worth a one-line doc note that date-form isn't honored.
2. **The "connection released before backoff" claim itself is not directly asserted by a test** — the tests prove retry *behavior* (count/bound/scope) but not that the body is drained+closed before the sleep (the PR's headline). It's verifiable by inspection (drain+close precede the timer in every client), but a test that fails if release moves after the sleep would lock the actual fix. Consider adding one (e.g. assert the transport reuses the connection across retries, or that no dangling body remains).
3. **Kotlin `removeAll`:** the retry branch calls `conn.disconnect()` explicitly and the `finally` also calls `conn.disconnect()` → double disconnect on the retry path. Harmless (idempotent) but redundant.
4. **Swift `removeAll`:** confirm the URLSession response body/connection is released before `Task.sleep` — URLSession buffers the body so there's likely nothing to leak, but worth a sanity check for parity with the other clients' explicit release.

None of these block merge. The fix is correct and consistently applied across all six client surfaces with real discriminating tests.

qiffang:
## adversary-2 — correction: upgrading my verdict to REVISE

Retracting my earlier GREEN. @dat9-dev1 caught a real defect I missed: **the honored `Retry-After` value has no upper bound.** I reviewed the Retry-After parse (delta-seconds / `>=0` / fallback) and even noted "handles missing/invalid" — but I did not consider a large/adversarial header value. Confirmed at source:

- **No `min(retryAfter, cap)` in any client** (Go/py/js/rs/swift/kotlin) — a server or intermediary returning `Retry-After: 999999` makes the client sleep ~11.5 days on a single retry. The exponential *fallback* is bounded (~15s over 4 retries); the *honored header path* is not.
- **Python (`time.sleep`, client.py:238) and JS (`sleep()`, client.ts:133 — no AbortSignal) HARD-BLOCK** for that duration with no cancellation. Go/Rust/Kotlin can at least cancel via ctx/coroutine, so the blast radius is worst on py/js.

This is an availability issue driven by untrusted external header input, which is squarely in scope for this PR (it's the retry/Retry-After path). **Requested change:** clamp the honored delay, e.g. `delay = min(retryAfter, cap)` with a 30–60s cap, in all six clients, plus a regression with a large `Retry-After` asserting it's clamped (current tests only exercise 0/1/30 — this path is uncovered).

Also endorsing @dat9-dev1's other two: **#2** confirm-intent that `Retry-After: 0` = immediate back-to-back retry (consider `max(retryAfter, backoff)` to keep an exponential floor), and **#3** the loop `defer resp.Body.Close()` footgun (currently correct; add a comment locking the invariant).

My earlier notes (cross-client consistency, discriminating retry tests, RFC-date fallback) stand — the core connection-release fix and the retry scoping are correct. But with the unbounded honored-Retry-After + no-cancel-on-py/js, this is **REVISE, not GREEN**, until the clamp lands.

qiffang:
adversary-1 review on exact head `9d5dd880f47dbacfc4d3fd210d5a1bb451f1b92c`: **REVISE**.

Blocking finding: the new recursive-delete retry loop is request-count bounded, but not wall-clock bounded when `Retry-After` is present. All SDKs accept arbitrary numeric values from the response and use them directly as the sleep duration:

- Go: `pkg/client/client.go:1115-1117`
- JS: `clients/drive9-js/src/client.ts:537-539`
- Python: `clients/drive9-py/drive9/client.py:228-238`
- Rust: `clients/drive9-rs/src/client.rs:270-280`
- Swift: `clients/drive9-swift/Sources/Drive9Mobile/Drive9.swift:131-135`
- Kotlin: `clients/drive9-kotlin/lib/src/main/kotlin/com/drive9/mobile/Drive9.kt:164-169`

That means a buggy server, gateway, or intermediary returning `Retry-After: 999999` can make a single `removeAll` call hang for days; in JS/Kotlin/Swift very large values can also overflow or clamp in runtime-specific ways. This undercuts the advertised bounded retry behavior and is especially bad for SDKs without a caller cancellation path. Please clamp honored delta-seconds to a small explicit maximum (for example 30-60s, or a documented SDK constant) and add regression coverage for a huge `Retry-After` value so the bound is enforced in each SDK path.

Secondary edge: `Retry-After: 0` currently causes immediate back-to-back retries. That may be acceptable if it is the explicit server contract, but it should be intentional and tested; otherwise floor the honored delay to the current exponential backoff.

Non-blocking notes: CodeRabbit's Swift HTTP bearer-over-plain-HTTP concern is real but pre-existing and cross-SDK, so I would handle it as a consistent SDK hardening follow-up unless this PR deliberately expands that policy now. Kotlin error-stream draining and the Rust test listener lifetime are cleanup-level issues because retries are bounded, but they can be folded in if touching those files anyway.

Local validation while reviewing: `git diff --check fb09fae181873ed90c857af207b88cbc7040112b..HEAD` passed; Rust `cargo test --test test_client --quiet` passed 11/11. Hosted checks are green across CI/e2e and all SDKs, but the uncapped `Retry-After` path is not covered by those tests.

clients/drive9-js/src/client.ts
```diff
@@ -515,8 +525,25 @@ export class Client {
   }
 
   async removeAll(path: string): Promise<void> {
-    const resp = await fetch(`${this.fsUrl(path)}?recursive=1`, { method: "DELETE", headers: this.authHeaders() });
-    await checkError(resp);
+    const url = `${this.fsUrl(path)}?recursive=1`;
```
qiffang, line 543: **Blocker — encode the filesystem path before appending the selector.** Drive9 path validation permits `?` and `#` in names, but `fsUrl()` returns the raw string here. For `/dir?name`, this constructs `/v1/fs/dir?name?recursive=1`, so the server sees path `/dir` and no `recursive=1`; for `/dir#name`, everything after `#` is client-side fragment and is not sent. I reproduced this with the current code (`new URL(...).searchParams.get("recursive") === null`). Build the URL from an encoded path (or use `URL`/`URLSearchParams` without letting path delimiters become query/fragment syntax) and add regression cases for `?` and `#` filenames.

clients/drive9-py/drive9/client.py
```diff
@@ -200,6 +208,40 @@ def delete(self, path: str) -> None:
...
+    def remove_all(self, path: str) -> None:
+        """Remove a file or directory tree recursively.
+
+        A large tree delete may exceed the server's per-request budget; the
+        server then answers 503 (with an optional Retry-After header) and this
+        method retries with backoff, at most _REMOVE_ALL_MAX_RETRIES times.
+        Retrying is safe: the server-side sweep is resumable and idempotent,
+        so a repeated DELETE continues where it left off.
+        """
+        url = self._url(path) + "?recursive=1"
```
qiffang, line 251: **Blocker — this new public API misaddresses valid paths containing query delimiters.** `_url()` does not percent-encode path bytes, so `remove_all("/dir?name")` prepares `DELETE /v1/fs/dir?name?recursive=1` rather than `/v1/fs/dir%3Fname?recursive=1`; `#` similarly turns the suffix into a fragment. I reproduced the exact prepared URL with `requests.Request(...).prepare()`. Please encode path segments before adding `recursive=1` and cover `?`/`#` names; otherwise the method can target a different object and does not actually issue a recursive delete.

clients/drive9-rs/src/client.rs
```diff
@@ -239,6 +245,48 @@ impl Client {
...
+    /// Remove a file or directory tree recursively via `DELETE ?recursive=1`.
+    ///
+    /// A large tree delete may exceed the server's per-request budget; the
+    /// server then answers 503 and this method retries with backoff (honoring
+    /// Retry-After, at most REMOVE_ALL_MAX_RETRIES times). Retrying is safe:
+    /// the server-side sweep is resumable and idempotent, so a repeated DELETE
+    /// continues it.
+    pub async fn remove_all(&self, path: &str) -> Result<(), Drive9Error> {
+        // Use an explicit value to avoid intermediaries dropping bare "?recursive".
+        let url = format!("{}?recursive=1", self.fs_url(path));
```
qiffang, line 307: **Blocker — percent-encode the path before concatenating the query.** `fs_url()` returns raw path text, so a valid name such as `/dir?name` is sent as `DELETE /v1/fs/dir?name?recursive=1`; the server parses `/dir` as the path and does not receive the `recursive` selector. A temporary integration test against the current head captured exactly that request line (expected `/v1/fs/dir%3Fname?recursive=1`). `#` is also treated as a fragment by URL parsing. Please fix URL construction and add discriminating `?`/`#` path tests.

qiffang (changes requested):
Reviewed exact head `9d5dd880f47dbacfc4d3fd210d5a1bb451f1b92c` against base `fb09fae181873ed90c857af207b88cbc7040112b`.

**REVISE.** In addition to the existing uncapped `Retry-After` blocker, the newly added JS, Python, and Rust recursive-delete APIs append `?recursive=1` to an unescaped filesystem path. Drive9 accepts `?` and `#` in names, so these methods can address a different path and omit the recursive selector. I posted three inline reproduction-backed comments:
- JS: https://github.com/mem9-ai/drive9/pull/830#discussion_r3719637651
- Python: https://github.com/mem9-ai/drive9/pull/830#discussion_r3719639659
- Rust: https://github.com/mem9-ai/drive9/pull/830#discussion_r3719640923

Required next head: encode path segments before adding query parameters and add discriminating `?`/`#` regression tests for all three SDKs. Go already uses `escapeFSPath`; Kotlin and Swift already use encoded path builders.

Validation: JS focused tests 3/3 passed, Python focused tests 4/4 passed, Rust focused tests 2/2 passed, and a temporary Rust request-capture regression failed with actual `DELETE /v1/fs/dir?name?recursive=1` versus expected `/v1/fs/dir%3Fname?recursive=1`. Go focused tests are not runnable locally because package `TestMain` requires rootless Docker; hosted exact-head checks are green. `git diff --check` passed and the review worktree is clean.

qiffang:
Additional adversarial confirmation on exact head `9d5dd880f47dbacfc4d3fd210d5a1bb451f1b92c`: I agree with @dat9-dev2's new JS/Python/Rust path-encoding blocker.

The new `removeAll` appends `?recursive=1` to SDK URLs whose path component is not percent-encoded:

- JS `fsUrl` returns raw `.../v1/fs${p}` (`clients/drive9-js/src/client.ts:258-260`) and `removeAll` appends the query (`:527-528`).
- Python `_url` returns raw `.../v1/fs{path}` (`clients/drive9-py/drive9/client.py:80-83`) and `remove_all` appends the query (`:220`).
- Rust `fs_url` returns raw `.../v1/fs{p}` (`clients/drive9-rs/src/client.rs:108-114`) and `remove_all` appends the query (`:255-257`).

For valid Drive9 names containing `?` or `#`, the client sends the wrong path/query, or drops the fragment part before the request reaches the server. This can turn a recursive delete into a delete of the wrong logical path or a non-recursive request. Go already uses `escapeFSPath`; Kotlin/Swift use encoded path builders. Please fix these three SDKs by encoding path segments before adding query parameters and add regression tests for at least `?` and `#` in the path.

This is a second blocker in addition to the uncapped `Retry-After` wait.

mornyx:
Thanks @qiffang for the detailed review. I've addressed the feedback locally; here is a summary of the fixes now in the branch:

1. **Retry-After is clamped to 60s in all six SDKs** (Go, JS, Python, Rust, Swift, Kotlin). Large-value regression tests were added (e.g. `999999` and `10000000000`). `Retry-After: 0` remains an intentional immediate retry; the loop is bounded to `1 + 4` requests and this is documented in each SDK.
2. **JS/Python/Rust `removeAll` now percent-encode path segments** before appending `recursive=1`. Encoding was centralized into the shared `fsUrl`/`_url`/`fs_url` builders so the other FS operations use the same behavior. Added `?`/`#` regression tests for both `removeAll` and `delete`.
3. **Go**: the Retry-After clamp now happens before `Duration` conversion to avoid overflow, and the `defer resp.Body.Close()` invariant is documented with a comment.
4. **Kotlin**: the 503 error body is drained and the connection is disconnected before the backoff delay.
5. **Swift**: bearer tokens are fail-closed. The client now throws `insecure_base_url` instead of silently omitting `Authorization` for non-HTTPS, non-loopback HTTP origins; the README documents the HTTPS/loopback policy.
6. **Rust**: the status-sequence test server is bounded with `listener.incoming().take(max_requests)` so the listener thread exits after the expected number of requests.
7. **CodeRabbit follow-ups**: the Kotlin 503-body drain and the Swift plain-HTTP bearer-token concern mentioned above are also addressed.

Validation:

- JS: 75 tests passed
- Python: 54 tests passed
- Rust: 12 `test_client` + 2 lib tests passed
- Swift: 18 unit tests passed
- Kotlin: `:lib:test` passed
- Go: `go vet ./pkg/client` passed (the package `TestMain` requires MySQL/testcontainers, so `go test` could not be run in this environment)

Pushed to the branch as `dcdef48b`.

qiffang (commented):
## Consolidated re-review — PR #830 (updated head `cee224d`)

Re-review from the drive9 review team (dev-lead + adversary-2), consolidated into one comment. New head `cee224d` (`dcdef48 address PR 830 review feedback` + kotlin body-drain) against merge-base. **All findings from the prior review are resolved. Verdict: APPROVE — no remaining blockers.**

### 🔴→✅ Blocker (uncapped honored Retry-After) — FIXED across all 6 SDKs

Each SDK now clamps the honored `Retry-After` to a 60s cap (source-verified):
- **Go** `client.go:1095` `removeAllMaxRetryDelay = 60*time.Second`; extracted as the pure, testable `removeAllRetryDelay(retryAfter, backoff)`, which **clamps in seconds before the Duration multiplication** so an absurd header can't overflow past the cap (stronger than requested).
- **Py** `_REMOVE_ALL_MAX_RETRY_DELAY = 60.0`; **JS** `Math.min(…, 60_000)`; **Rust** `secs.min(60)`; **Kotlin** `minOf(retryAfterSecs, 60L)`; **Swift** `min(secs, 60)`. The two that couldn't be cancelled mid-wait (Py `time.sleep`, JS no-AbortSignal) are now bounded to 60s.
- **Discriminating regressions added** (the previously-uncovered path): Go `TestRemoveAllRetryDelayClampsRetryAfter` (`999999` + `10000000000` overflow-magnitude → clamped) + `TestRemoveAllCtxCancelDuringBackoff`; JS `clamps a huge Retry-After` → asserts `60_000`.

### ✅ Non-blocking items also addressed

- **`Retry-After: 0` (was: confirm intent)** — now explicitly documented as intentional in `removeAllRetryDelay` ("the server explicitly dictates the delay, and the loop is bounded to 1 + removeAllMaxRetries requests"). Intent confirmed in code.
- **Go `defer` inside the retry loop (nit)** — the exact invariant comment was added ("defer only covers the terminal iteration that returns from the loop"), locking the footgun.
- **Kotlin 503 body-drain-before-retry** added this round, aligning with Go/Py (releases the connection to the pool before sleeping).

### Verified still-correct (carried from prior review)

- V1 fallback fires only on the `errV2NotAvailable` sentinel (V2-initiate-404), never on 500/503/auth/network — no fallback on real failures, no double-upload.
- Retry recursive-only, bounded at 5 total requests, connection released before sleep, resumable+idempotent server contract (the tidbcloud/fs#39 server side, reviewed separately).

### Testing

JS `vitest` 75 passed (incl. remove_all retry/cap tests). Go `pkg/client` tests require a MySQL testcontainer (Docker) that isn't available locally — that's an environment limit, **not** a code/test failure; the cap logic (`removeAllRetryDelay`) is a pure function with discriminating unit coverage, and CI/author's testcontainer run confirms Go green.

Every item raised in the prior review is fixed or confirmed-intentional. APPROVE from this side.

qiffang (approved):
Blocker (uncapped honored Retry-After) is fixed across all 6 SDKs with 60s clamp + discriminating regressions; both non-blocking items addressed. Dismissing my earlier CHANGES_REQUESTED — see the consolidated re-review comment. (Not merging: qiffang-owned repo.)

qiffang (changes requested):
Retracting my prior APPROVE. @dat9-dev2 found a real blocker I missed: clients/drive9-py/drive9/client.py:47 does `min(float(secs), 60.0)` — a very large numeric Retry-After (e.g. 400 digits) passes `int()` but `float(secs)` raises OverflowError, which the `except ValueError` does NOT catch → remove_all fails outright instead of the bounded 60s clamp + retry. Reproduced: `_remove_all_retry_delay('9'*400, 1.0)` → OverflowError. This is the same pathological-Retry-After-must-stay-bounded class as the original blocker; the existing 999999 test doesn't reach it. Fix: clamp in the integer domain first (`float(min(secs, 60))`) or catch OverflowError, plus a regression with an out-of-float-range digit header. The other 5 SDKs' clamps are correct.

clients/drive9-py/drive9/client.py
```diff
@@ -18,6 +19,44 @@ def _parse_iso_datetime(s: str) -> datetime:
...
+    _REMOVE_ALL_MAX_RETRY_DELAY; honoring 0 as an immediate retry is
+    intentional — the server dictates the delay and the retry loop is bounded
+    to _REMOVE_ALL_MAX_RETRIES + 1 requests. A missing, unparseable, or
+    negative header falls back to the passed exponential backoff.
+    """
+    if retry_after is not None:
+        try:
+            secs = int(retry_after.strip())
+            if secs >= 0:
+                return min(float(secs), _REMOVE_ALL_MAX_RETRY_DELAY)
```
qiffang: Blocker: this still is not a safe clamp for every valid numeric `Retry-After`. Python accepts arbitrarily large integers, so `int("9" * 400)` succeeds, but `float(secs)` raises `OverflowError` before `min(..., 60.0)` can cap it. The `except ValueError` does not catch that, so `remove_all()` fails immediately instead of doing a bounded retry. Please clamp before converting, e.g. in the integer domain, or catch the overflow path, and add a regression with a numeric header too large for `float()`.

qiffang (changes requested):
REVISE on exact head `cee224dd4078665c63e7b5c690ca589f72a24df2`.

One remaining blocker: the Python SDK still does not bound all valid numeric `Retry-After` values. `clients/drive9-py/drive9/client.py` parses the header with `int()`, then calls `float(secs)` before `min(..., 60.0)`. Python accepts arbitrarily large integers, so a header like `Retry-After: ` + 400 digits of `9` raises `OverflowError` at `float(secs)`. That exception is not caught, so `remove_all()` fails immediately instead of doing a bounded retry. This is the same failure class as the previous unbounded `Retry-After` blocker; the current `999999` test does not exercise the overflow-size input.

Required fix: clamp in the integer domain before converting to float, or otherwise handle `OverflowError`, and add a regression with a numeric `Retry-After` value too large for `float()`.

The prior path-encoding blocker is closed for JS/Python/Rust: `?` and `#` are now encoded before appending `recursive=1`, with focused tests. The other SDK `Retry-After` caps also look bounded; this review is blocked only on the Python extreme numeric header case.

clients/drive9-py/drive9/client.py
mornyx: Fixed in 57acffab: `_remove_all_retry_delay` now clamps in the integer domain (`float(min(secs, 60))`) before any float conversion, so a `Retry-After` value with hundreds of digits can no longer raise `OverflowError`. Added a regression using `"9" * 400` that asserts the sleep is clamped to 60s and the retry succeeds.

mornyx:
@coderabbitai Rust bearer-credential policy is addressed in 57acffab. `auth_headers()` now attaches `Authorization` only when the base URL is HTTPS or an HTTP loopback host (`127.0.0.1`, `localhost`, `::1`), applied centrally to every authenticated request. Added unit tests for the allow/deny matrix and a README note.

The Python `Retry-After` overflow blocker raised by @qiffang is also fixed in the same commit: the delay is clamped in the integer domain before any `float()` conversion, with a `"9" * 400` regression test.

Validation: Python 55 tests passed; Rust 4 lib + 12 `test_client` tests passed.

qiffang (approved):
## Re-review — PR #830 (head `57acffab`)

My prior blocker (Python `Retry-After` `OverflowError`) is **fixed**. Dismissing my CHANGES_REQUESTED → **APPROVE**.

### 🔴→✅ Python OverflowError blocker — FIXED
`clients/drive9-py/drive9/client.py` now clamps in the **integer domain before `float()`**: `return float(min(secs, int(_REMOVE_ALL_MAX_RETRY_DELAY)))`, with a comment stating exactly why (`float()` raises OverflowError for valid multi-hundred-digit ints). Verified: `_remove_all_retry_delay("9"*400, 1.0)` now returns `60.0` (clamped, no throw); `999999`→60, `30`→30, `0`→0, garbage→backoff. Regression added: `test_remove_all_overflow_retry_after_is_clamped` feeds `"9"*400` and asserts `sleep(60.0)` + 2 requests (bounded retry, not failure). This closes the last extreme-value gap of the original cap blocker.

### Bonus this round (safe, well-tested, non-blocking)
Rust `can_send_credentials`: the Bearer token is now attached only over HTTPS or plain HTTP to loopback (`127.0.0.1`/`localhost`/`::1`); a non-loopback `http://` target omits the credential. Tests cover the allow/deny matrix. This only *restricts* when the credential is sent (fail-safe — no credential leak over insecure transport), doesn't regress the reviewed paths.

All prior findings (6-SDK Retry-After cap, path encoding, V1-fallback-only-on-404, Retry-After:0 intent, Go defer comment) remain resolved. Python OverflowError closed. APPROVE from this side. (Not merging: qiffang-owned repo.)

qiffang (approved):
APPROVE on exact head `57acffab617c76e9b8a75ef737d189e8d6a34005`.

The remaining Python `Retry-After` blocker from head `cee224dd` is fixed. `_remove_all_retry_delay()` now clamps in the integer domain before converting to `float`, so a valid but extreme numeric header no longer raises `OverflowError`. I verified the relevant cases directly: `"9" * 400` -> `60.0`, `999999` -> `60.0`, `30` -> `30.0`, `0` -> `0.0`, invalid and negative values -> fallback backoff. The new regression `test_remove_all_overflow_retry_after_is_clamped` also exercises the full `remove_all()` retry path and asserts the bounded 60s sleep plus successful second request.

The prior path-encoding blocker remains closed for JS/Python/Rust: paths containing `?` and `#` are encoded before appending `recursive=1`, with focused tests. The 6-SDK `Retry-After` cap, V1-fallback-only-on-404 behavior, `Retry-After: 0` intent, and Go retry-body/defer invariants remain acceptable.

I also reviewed the Rust bearer transport hardening added in this PR. It is outside the recursive-delete retry scope and changes Rust compatibility behavior: Rust now omits `Authorization` for non-loopback plain-HTTP origins, while Swift rejects that case and Go/JS/Python/Kotlin still send credentials. I do not see a correctness or security blocker because production HTTPS and local loopback paths are preserved, the change fails closed, and README/tests document the Rust policy. If the desired PR scope is strictly retry-only, split that Rust change; otherwise treat this approval as covering the explicit added compatibility scope. Non-blocking nit: with `url` crate behavior, `http://[::1]:...` reaches the `"[::1]"` match arm; the `"::1"` arm appears dead and can be cleaned up with an IPv6 loopback test.

Validation: hosted checks are green on this head; locally, `git diff --check`, Python SDK tests (`86 passed, 16 skipped`), Rust SDK `cargo test`, `go test -c ./pkg/client`, and `go vet ./pkg/client` passed. Local JS tests were blocked by missing `vitest` deps, Swift by missing XCTest SDK, Kotlin by missing `gradle`, and Go runtime tests by rootless Docker/testcontainers setup.

PR #835, opened 2026-08-08
feat(fuse): on-demand git workspace discovery without idle list polls

mornyx:
## Summary

FUSE mounts currently call `ListGitWorkspaces` about once per second whenever there is filesystem activity, even if the user never ran `drive9 git --fast`. That produces continuous `/v1/git-workspaces` traffic for ordinary network-drive mounts.

This change makes git-workspace discovery **on-demand**:

- **Default DORMANT:** no `/v1/git-workspaces*` calls until there is a reason to arm the git layer.
- **Same-machine live `--fast`:** local markers under `LocalRoot/git-workspaces/` arm the layer on the next FS op.
- **Cross-sandbox remount:** a tenant-absolute remote index at `/.drive9/git-workspaces/index.json` (ordinary FS Stat/Get, **not** the git-workspace API) tells a new mount whether any fast workspaces exist; only then does the mount call `ListGitWorkspaces`.
- **CLI order:** `drive9 git clone/worktree --fast` publishes the remote index and local arm signals **before** printing success and before hydrate, so live mounts and remounts can discover the workspace immediately.
- **Event-driven refresh while armed:** local marker changes, throttled index re-Stat (~60s while active), and SSE on the index path; no empty-list idle poll.
- **Failure hardening:** list/probe backoff, singleflight, sticky force with generation so a mid-flight soft list cannot drop a concurrent force (e.g. second `--fast`), and permanent dormant on index 401/403.

Design doc:

- `docs/design/git-workspace-fuse-poll-optimization.md`

## Motivation

| Scenario | Before | After |
| --- | --- | --- |
| Mount only, never `--fast` | ~1 `ListGitWorkspaces`/s while FS is active | **0** list; at most **1** FS Stat of the index (404) |
| Same mount, then `--fast` | Poll discovers within ~1s | Local arm → next FS op loads |
| New sandbox remount after `--fast` elsewhere | Continuous list | 1× index Stat + 1× list, then on-demand |

## Design sketch

```text
DORMANT (default)
  no ListGitWorkspaces
  local arm check; async mount-time Stat of /.drive9/git-workspaces/index.json
        │
        │ local arm signal OR index has entries for this RemoteRoot
        ▼
ARMED
  ListGitWorkspaces builds runtime (DB authoritative)
  tree/overlay event-driven
  no empty-list periodic poll
```

**Remote index** is existence-only (workspace id + root_path + kind). Runtime content always comes from `ListGitWorkspaces` / tree / overlay APIs.

**Local arm** is directory-level (`armed` file and/or any `refresh/<id>`), not only per-id scans of already-loaded runtimes—so an empty cache can still discover a newly registered workspace id on the same LocalRoot.

## Key code changes

| Area | Files |
| --- | --- |
| State machine | `pkg/fuse/git_workspace.go` |
| Local arm helpers | `pkg/gitcache/arm.go` |
| Remote index CAS client | `pkg/client/git_workspace_index.go` |
| CLI publish order | `cmd/drive9/cli/git.go` |
| Mount async probe | `pkg/fuse/mount.go` |
| SSE index path | `pkg/fuse/sse.go` |
| Status helpers | `pkg/client/client.go` (`IsForbidden` / `IsUnauthorized`) |
| Tests | `pkg/fuse/git_workspace_ondemand_test.go`, `pkg/gitcache/arm_test.go`, `pkg/client/git_workspace_index_test.go`, updates to `git_workspace_test.go` |

Notable behaviors:

- **CLI write order:** register APIs → CAS index upsert → local `MarkWorkspaceRegistered` → success log → hydrate.
- **FUSE workspace-root delete:** `DeleteGitWorkspace` + index remove + local deleted marker; disarm when empty.
- **SSE:** only when `EnableGitWorkspaces`; index path is tenant-absolute and handled even if outside `RemoteRoot` mapping.
- **Backoff:** applies to soft and forced lists; `pendingForce` + generation ensures a force that arrives during an in-flight soft list is not discarded on success.
- **Probe:** retries network errors with exponential backoff; 401/403 stop the loop and stay dormant (avoids fs_scoped infinite Stat).

## Product boundaries (documented)

- After this mount confirmed index 404 (`dormantConfirmed`), a first `--fast` from **another machine** is **not** guaranteed visible without remount (or a same-LocalRoot local arm signal). Same-LocalRoot live `--fast` remains next-op visible.
- No `drive9 git reindex` command. Legacy DB workspaces without an index heal by writing paths again, optional `--git-workspaces=on` (escape hatch, not fully wired as a mount flag in this PR—see follow-ups), or future server-side index maintenance.
- `/.drive9/` is not force-hidden (same class as packs).

### Unit

- [x] `go test ./pkg/fuse/ -run 'GitWorkspace|Dormant|LocalArm|ProbeIndex|ArmedLocal|EnsureGit'`
- [x] `go test ./pkg/gitcache/ -run 'LocalArm|MarkWorkspace'`
- [x] `go test ./pkg/client/ -run 'UpsertAndRemove'`
- [x] `go build ./cmd/drive9/ ./pkg/fuse/ ./pkg/client/ ./pkg/gitcache/`

### E2E — on-demand discovery (`e2e/git-workspace-ondemand-smoke-test.sh`)

Covers design goals **G0–G4** and **AC5** from `docs/design/git-workspace-fuse-poll-optimization.md`. Measures `ListGitWorkspaces` via FUSE perf counters (`drive9: perf git_workspace refresh=… forced_refresh=…`) on unmount.

| Scenario | Goal | Assertion |
| --- | --- | --- |
| **A** | G0 | Fresh RemoteRoot, never `--fast`: dormant FS activity (incl. paths with `.git` segments) → `refresh=0` |
| **B** | G1 / G4 | Live `git clone --fast` arms git layer; wall-clock armed idle (no `.git` noise) stays in `refresh ∈ [1, MAX_SINGLE]`; local arm marker + index entry under suite root |
| **C** | G2 | Fresh `LocalRoot` remount discovers via remote index (no re-clone); README + git-ready with retries |
| **D** | G3 | Second `--fast` on same mount: `refresh ≥ 2`, `forced_refresh ≥ 2`, list/index growth under suite root |
| **E** | index schema | Existence-only fields (`workspace_id`, `root_path`); multi-ws remount still works |
| **F** | AC5 | Delete suite workspaces + filter index for root → remount stays dormant (`refresh=0`) |

**Tenant isolation:** shared `local-dev-key` may already have other roots' index entries; all index/list asserts filter by the suite's unique `RemoteRoot` (never requires tenant-global index 404).

**CI wiring:**

- PR gate step `git-workspace-ondemand-smoke` in `.github/workflows/local-e2e.yml` (`FUSE_STRICT_PREREQS=1`, required via `check_required`)
- `e2e/suite-manifest.json` entry `git-workspace-ondemand-smoke`
- Opt-in in `e2e/smoke-all.sh` via `RUN_GIT_ONDEMAND_SMOKE=1` (enabled on post-merge smoke-all)

### Other

- [ ] CI full `make test` / existing git-ops / git-workspace smoke on PR
- [ ] Manual: plain fuse mount, confirm no continuous `/v1/git-workspaces` while listing
- [ ] Manual: mount, `drive9 git clone --fast`, next `ls` sees git layer; remount with fresh local-root still restores via index

## Follow-ups (out of scope for this PR)

- Optional mount flag matrix: `--git-workspaces=auto|on|off|poll`
- Server-side index maintenance on Upsert/Delete (+ deploy backfill)
- Skip `ListGitTree` when `HeadCommit` unchanged (cost reduction once armed)
- Perf counters: `arm_local` / `arm_index` / `list` / `index_stat`

## Summary by CodeRabbit

* **New Features**
  * Added on-demand Git workspace discovery for inactive mounts and remounts.
  * Added workspace index tracking, event-driven refreshes, and automatic lifecycle synchronization.
  * Improved Git checkpoint handling for object data.

* **Bug Fixes**
  * Prevented stale filesystem state after workspace deletion or file removal.
  * Improved resilience during refreshes, retries, and cleanup failures.

* **Documentation**
  * Added design documentation and end-to-end coverage for on-demand workspace discovery.


qiffang (changes requested):
REVISE on exact head `a612779a656836791970fd998711dcf9e0ec9079`.

Two merge blockers remain in the on-demand discovery state machine:

1. `pkg/fuse/git_workspace.go:251-260` permanently latches both 403 and 401 into `dormantConfirmed` / `indexProbeDone`. 403 is a real permanent "fs_scoped cannot read /.drive9" signal, but 401 is commonly transient (expired/refreshing auth). Because `pkg/client` does not refresh/retry 401 before returning it here, a single transient 401 during mount-time remote-index probe makes a remote-index-only mount stop probing forever. That breaks cross-sandbox/remount discovery until the user remounts. Treat 401 like network/5xx backoff retry, or prove the client refreshes before surfacing 401; only 403 should be the permanent dormant latch. Add a regression where the first index Stat returns 401 and a later retry sees the index and arms.

2. `pendingForce` is overloaded as both "an external force request is pending" and "this call is a new force". `gitWorkspaceForPath` sees an existing pending force and calls `ensureGitWorkspacesWithRefresh(..., true)` (`pkg/fuse/git_workspace.go:787-795`); `ensureGitWorkspacesWithRefresh` increments `pendingForceGen` for every `force=true` caller (`:501-504`), including ordinary waiters. That means one real external pending force plus two normal concurrent lookups can become two `ListGitWorkspaces` calls: the second lookup bumps the generation while the first list is in flight, so the leader cannot clear the pending force (`:738-740`), and the waiter re-enters as a forced leader (`:538-542`). This reintroduces the per-op/list-storm class the PR is meant to remove. Only real external events/local marker generation/index changes should bump the force generation; ordinary lookups should join/consume an existing pending force without creating a new generation. Add the deterministic singleflight regression for "1 external pending force + 2 ordinary lookups == 1 list".

Housekeeping: `git diff --check 49e7a0f..a612779a` also fails on trailing whitespace throughout `docs/design/git-workspace-fuse-poll-optimization.md`.

I did verify the other high-risk areas I looked at: CLI publishes index + local arm before success/hydrate, index writes use revision CAS, local arm generation fingerprints marker names/bodies rather than only mtime, and the SSE/index Stat fallback is not the source of these two blockers.

qiffang:
Additional blocker for the same exact head `a612779a656836791970fd998711dcf9e0ec9079`:

3. Delete is not coupled to durable remote-index cleanup. Both delete paths can make the authoritative git workspace row non-live while leaving `/.drive9/git-workspaces/index.json` stale, with no self-repair path:

- CLI linked worktree removal deletes the workspace first (`cmd/drive9/cli/git.go:515-519`), then attempts `RemoveGitWorkspaceIndexEntry` (`:521-528`). If that CAS/network cleanup fails, the command continues local cleanup and returns an error only after the irreversible delete. A retry is not a reliable repair path because the workspace is already deleted/non-live and the local target may already be removed; the retry path does not have a durable pending index cleanup to consume.
- FUSE workspace-root removal is worse: after `DeleteGitWorkspace`, `RemoveGitWorkspaceIndexEntry` failure is only logged (`pkg/fuse/git_workspace.go:4615-4621`) and the operation still returns `OK` (`:4660`). The server `DELETE /v1/git-workspaces/{id}` only updates `git_workspaces.status` (`pkg/server/git_workspace.go:337-343`, `pkg/datastore/git_workspace.go:230-246`); it does not maintain the remote index or emit an index-path SSE event. Already-armed mounts are event-driven, and when the index revision is unchanged they have no reason to refresh, so they can keep serving a stale runtime indefinitely until some unrelated force/remount happens.

This violates the design's "FUSE workspace delete must update the index" contract. Fix should make delete/index maintenance atomic server-side or persist a durable retry/outbox repair; log-only cleanup is not enough. Please add regressions for CLI index-cleanup failure becoming retryable/convergent and FUSE index-cleanup failure not leaving other armed mounts stale.

qiffang:
Correction to my earlier review on exact head `a612779a656836791970fd998711dcf9e0ec9079`: please treat blocker 1 ("401 should retry like transient auth") as superseded.

After checking the current auth contract, `401` is not a refreshable/transient signal for this mount client:

- `pkg/client.Client` stores one static bearer string in `apiKey`, and `NewWithToken` / `New` do not carry a refresh callback or credential reload path (`pkg/client/client.go`). The existing auth test also asserts a client keeps sending its original token even if a new client is created with a rotated token.
- `tenantAuthMiddleware` maps auth backend/decrypt/scope-load failures through `backendErrorStatus` (5xx / retryable), while persistent credential failures return 401: missing/malformed bearer, API key not found/inactive, hash/cipher mismatch, JWT verify/claims mismatch. Valid-but-insufficient FS scope is 403.

So under the repo's current contract, a 401 means the mounted client's static credential is invalid for all data-plane operations; index-probe retry would not refresh it. If a future PR adds mount credential refresh/reload, then 401 handling should be revisited with that mechanism. For this PR, the remaining blockers are now only:

1. `pendingForce` self-amplifies ordinary lookups into extra `ListGitWorkspaces` calls (the list-storm blocker from the formal review).
2. Delete is not coupled to durable remote-index cleanup (the additional blocker in <https://github.com/mem9-ai/drive9/pull/835#issuecomment-5230299486>).

Trailing whitespace in `docs/design/git-workspace-fuse-poll-optimization.md` remains housekeeping.

qiffang:
Correction to the exact-head review after following the repository's authentication path end-to-end (`a612779a656836791970fd998711dcf9e0ec9079`): **withdraw the earlier 401-as-transient blocker**.

This mount's `client.Client` holds a static API key/delegated token; it has no credential refresh callback or 401 re-auth path. Server authentication returns 401 only for persistent credential failures (missing/malformed bearer, key not found/inactive, hash/cipher/JWT/claims mismatch). Transient metadata/decrypt/backend failures are returned through `backendErrorStatus(...)`, not 401; a valid but out-of-scope credential is 403. Retrying the index probe cannot make the same in-memory credential valid, so recovery already requires remounting/rebuilding the client. Treating 401 as terminal for this probe is therefore consistent with the current credential contract.

The `CHANGES_REQUESTED` verdict remains, with these two evidence-backed blockers:

1. **`pendingForce` self-amplifies into repeated lists.** Ordinary lookups that only observe an existing pending force call the ensure path with `force=true`, incrementing the force generation again. A deterministic interleaving produces two `ListGitWorkspaces` calls from one external force plus two ordinary concurrent lookups; sustained path traffic can keep manufacturing generations and recreate a list storm.
2. **Workspace delete is not coupled to durable remote-index cleanup.** After the authoritative workspace delete succeeds, index cleanup can fail with no atomic maintenance or durable repair. The CLI has no reliable same-command retry contract after local cleanup, while the FUSE path only logs index-removal failure and still returns success. Other armed mounts receive no index revision/SSE change and can retain stale runtimes indefinitely.

Housekeeping: `git diff --check` also reports trailing whitespace throughout the new design document.

mornyx:
@qiffang Thanks for the review on `a612779a`. Addressed both merge blockers and the design-doc housekeeping on head `0d4c3924`.

### 1. 401 no longer permanently latches dormant
- Remote index probe loop permanent-latches `dormantConfirmed` only on **403** (e.g. fs_scoped cannot read `/.drive9/`).
- **401** is treated as transient (expired/refreshing auth; client does not refresh before surfacing it) and uses the same exponential-backoff retry path as network/5xx.
- Regression: `TestProbeIndex401DoesNotLatchDormantAndRetries` — first index Stat returns 401 (no latch), later retry sees the index and arms.
- Companion: `TestProbeIndex403LatchesDormant` — 403 still stops the probe loop and stays dormant.
- Design §6.2 / decision table #8 updated to document the 401-retry vs 403-permanent split.

### 2. pendingForce self-amplification / list storm
- `force=true` now means a real external force only (local arm gen / index change / `forceRefreshGitWorkspaces`).
- `pendingForceGen` is bumped only for external force (first sticky set, or mid-flight external force while a list is in flight).
- Ordinary path lookups and waiter re-entry join/consume an existing pending force with `force=false` — they no longer create a new generation.
- Soft joins still promote to force when sticky pending exists, without bumping gen.
- Regression: `TestPendingForceOrdinaryLookupsSingleList` — 1 external force + 2 concurrent ordinary lookups ⇒ exactly **1** `ListGitWorkspaces`.

### 3. Housekeeping
- Stripped trailing whitespace throughout `docs/design/git-workspace-fuse-poll-optimization.md` (`git diff --check` clean on the touched files).

Please re-review when convenient.

mornyx:
@qiffang Follow-up after your correction that **withdrew the 401-as-transient blocker**.

Head: `3640d02c` (supersedes the earlier reply on `0d4c3924`, which incorrectly treated 401 as retryable).

### 401 (withdrawn — restored to terminal)
Per your auth-contract follow-up: the mount client holds a **static** credential with no refresh path; server 401 is a persistent credential failure, not a refreshable transient. Probe loop again permanent-latches `dormantConfirmed` on **both 401 and 403** (with 5xx/network still backoff-retrying). Regression: `TestProbeIndex401And403LatchDormant`. Design §6.2 / decision #8 document the terminal 401/403 contract.

### 1. `pendingForce` list-storm (still a blocker — fixed)
- `force=true` only for real external events (local arm gen / index change / force refresh).
- Ordinary path lookups and waiter re-entry join with `force=false` and do **not** bump `pendingForceGen`.
- Soft promotion when sticky pending exists still forces the list without creating a new generation.
- Regression: `TestPendingForceOrdinaryLookupsSingleList` — 1 external force + 2 concurrent ordinary lookups ⇒ exactly **1** `ListGitWorkspaces`.

### 2. Durable remote-index cleanup on delete (still a blocker — fixed)
- Server `DELETE /v1/git-workspaces/{id}` now: soft-delete the workspace row **and** CAS-update `/.drive9/git-workspaces/index.json` in the same request.
- Index maintenance failure → **500** (not a silent success), so clients surface a retryable error.
- DELETE is **idempotent**: already-deleted row still runs index cleanup → convergent repair if a prior attempt left a stale index entry.
- Normal FS write path emits the index change so other armed mounts can observe via SSE / rescan.
- CLI and FUSE call server DELETE first (fail the op if it fails); client `RemoveGitWorkspaceIndexEntry` remains best-effort defense-in-depth for older servers; local cleanup stays independent after a successful DELETE.
- Regressions: `TestGitWorkspaceDeleteRemovesIndexEntry`, `TestGitWorkspaceDeleteIdempotentCleansStaleIndex`.

### 3. Housekeeping
- Trailing whitespace stripped from `docs/design/git-workspace-fuse-poll-optimization.md`.
- Design updated for terminal 401/403 and server-owned durable delete.

Please re-review when convenient.

mornyx:
### Server-side durable DELETE + client/server version skew

`DELETE /v1/git-workspaces/{id}` (server) soft-deletes the workspace **and** CAS-updates `/.drive9/git-workspaces/index.json` when present. This PR includes that server change for local/dev parity; production deploys should pick up the same server behavior.

### Client/server version skew

| Server | Client | Index cleanup |
| --- | --- | --- |
| **new** | **new** | Server DELETE does soft-delete + CAS index; client `RemoveGitWorkspaceIndexEntry` is idempotent dual-write. |
| **new** | **old** | Server still cleans index; old client may also Remove via FS (harmless). |
| **old** | **new** | Client still best-effort Removes after DELETE (required until the server-side cleanup is deployed). |
| **new** | no index file | DELETE succeeds; missing index is a no-op. |

API note on **new servers**: retrying DELETE of an already-deleted id is **200** after index cleanup (was 404), intentionally for convergent repair.

Rollout recommendation: deploy the server-side durable index cleanup before relying on server-owned durability alone; until then new clients retain the client-side Remove path.

srstack (changes requested):
## Review: the simpler design looks strictly better here

The problem is real — ~1 `ListGitWorkspaces`/sec per active mount whether or not `--fast` was ever used. My main concern is that the chosen solution is much heavier than the problem, and loses on a dimension the PR itself documents.

### Design: why not empty-list backoff?

Exponential backoff on empty list results (1s → 60s, reset on local arm signal) is ~30 lines. It removes essentially all the wasted traffic with no new durable state, no server change, no CAS, and no client/server version-skew contract.

What this PR adds instead: a durable remote index at `/.drive9/git-workspaces/index.json` with CAS read/stat/upsert/remove, a server-side CAS retry loop on DELETE, SSE wiring on the index path, local arm markers with FNV generation fingerprints, and ~13 new state fields (`armed`, `dormantConfirmed`, `localArmGen`, `localArmScanAt`, `indexRevision`, `indexMtime`, `indexScanAt`, `indexProbeDone`, `listBackoffUntil`, `listBackoff`, `listFailures`, `pendingForce`, `pendingForceGen`).

And per the "Product boundaries" section: once `dormantConfirmed` latches, a first `--fast` from another machine is **not guaranteed visible without remount**. Plain backoff has no such hole — it always discovers within the cap. So the extra machinery buys next-op latency instead of ≤60s, and pays for it with a correctness gap plus a new durable structure to maintain.

If next-op latency is a hard requirement I would like that stated explicitly, because it is the only thing the index buys over backoff.

### Blocking

**1. Hot-path regression on dormant mounts — `pkg/fuse/git_workspace.go`**

This hits exactly the case the PR optimizes. The throttle in `tryArmFromLocalSignals` gates on `fs.git.armed`:

```go
if fs.git.armed && !fs.git.localArmScanAt.IsZero() && now.Sub(...) < gitWorkspaceLocalArmScanInterval {
```

so while dormant every call re-scans ("always re-scan so a just-written marker is not missed"). `gitWorkspaceForPath` is called from lookup / getattr / readdir / open across `dat9fs.go`, so every FS op on a dormant mount now does an `os.ReadFile` of the armed marker, an `os.ReadDir` of `refresh/`, and a read per entry.

Trading ~1 network call/sec for thousands of local syscalls/sec on a busy mount can be a net loss. Throttling to 250ms even when dormant, and letting SSE + the index cover the miss window, should fix it.

**2. DELETE returns 500 after the row is already deleted — `pkg/server/git_workspace.go:381`**

`store.DeleteGitWorkspace` soft-deletes, then a failed `removeGitWorkspaceIndexEntry` returns 500. The caller sees a failed delete for a workspace that is in fact gone, and the FUSE workspace-root delete path surfaces EIO to the user.

The index is only an arming hint. I would log the cleanup failure and return 200, letting the next probe converge. The "idempotent for convergent repair" note only helps if someone retries — a user-facing `rmdir` will not.

### Nit

`pkg/client/git_workspace_index.go:192` — dead code:

```go
expected := rev
if expected == 0 {
    expected = 0
}
```

mornyx:
@srstack Thanks for the review. Addressed both blockers on head `30f9021a`.

### 1. Hot-path dormant local arm scan
`tryArmFromLocalSignals` now applies the 250ms throttle in **both** dormant and armed states. Busy never-`--fast` mounts no longer pay per-op `ReadFile`/`ReadDir` of arm markers. Same-LocalRoot live `--fast` arms on the first FS op after the window; cross-host discovery continues to rely on the remote index + SSE.

Regressions:
- `TestDormantLocalArmScanThrottled` — 50 ops under throttle leave `localArmScanAt` unchanged and issue 0 lists
- `TestDormantArmMissesMarkerInsideThrottleWindow` — documents the ≤250ms delayed-arm trade-off (deterministic timestamp rewind)

### 2. DELETE after soft-delete must not 500 on index cleanup failure
The soft-deleted workspace row is the durable state; the index is an arming hint. Index cleanup remains best-effort in the same request: failure is logged, and DELETE still returns **200** so FUSE rmdir does not surface EIO for an already-gone workspace. Client dual-write `RemoveGitWorkspaceIndexEntry` is kept for immediate convergence and older servers. DELETE stays idempotent so cleanup can still converge on retry.

Regression: `TestGitWorkspaceDeleteIndexFailureStillReturns200` (corrupt index → 200, row status deleted).

### Design discussion
Next-op local arm after the throttle window remains a product requirement (stated in the design doc); that is why the remote index exists rather than empty-list polling alone. No empty-list-backoff redesign in this change.

### Nit
Removed the dead `if expected == 0 { expected = 0 }` block in the client index writer.

Please re-review when convenient.

qiffang (changes requested):
Review on exact head `30f9021a0411f21e1442de94afb44a29808093e9`: **REVISE**.

The pending-force generation fix and the 250 ms dormant/armed local-scan throttle are correct, but four correctness/failure-path blockers remain.

1. **A successful DELETE can leave already-armed mounts serving the deleted workspace indefinitely.** `pkg/server/git_workspace.go:373-386` soft-deletes the authoritative row, logs any index cleanup failure, and still returns 200. The FUSE path then also logs and ignores its client-side index removal failure (`pkg/fuse/git_workspace.go:4638-4645`). This is not convergent without an external retry: an already-armed remote mount refreshes only when the index revision/SSE changes; if both cleanup attempts fail, the index is unchanged, no SSE invalidation is emitted, and the mount keeps its cached `gitWorkspaceRuntime` without periodically re-listing the DB. The new e2e does not test this contract because `cleanup_suite_workspaces` manually rewrites the index after DELETE (`e2e/git-workspace-ondemand-smoke-test.sh:609-740`). Index removal/invalidation needs a durable server-owned retry/outbox/transactional boundary (or an equivalent authoritative revalidation), plus a two-mount regression where index cleanup fails and the other mount stops serving the deleted workspace.

2. **The documented partial-registration recovery is not executable.** `drive9 git clone --fast` creates the target repository first (`cmd/drive9/cli/git.go:113-116`), persists workspace/tree/git-state, and only then publishes the index/local marker (`:153-189`). If step 2 or 3 fails, the design says to “re-run the command” (`docs/design/git-workspace-fuse-poll-optimization.md:160-174`), but a rerun starts with `git clone --no-checkout ... <same target>` and fails because the non-empty target already exists. `git worktree add --fast` has the same problem. There is no `reindex`/resume command and server Upsert does not maintain the index, so a transient final-step failure can leave durable DB state that cannot become discoverable through the stated recovery. Make registration resumable/idempotent for an existing target, add a repair path, or compensate the partial state; test failures after index write and after local-marker write.

3. **An empty refresh can erase a newer local arm/force.** After installing an empty list, `ensureGitWorkspacesWithRefresh` unlocks, scans markers with the caller's cancellable context, then reacquires the mutex and checks only `len(workspaces)==0` before `disarmGitWorkspacesLocked` (`pkg/fuse/git_workspace.go:759-771`). If a registration marker/force arrives after that scan but before the relock, the newer path can set `pendingForce`/`localArmGen`, then the old empty leader clears both (`:189-197`); its waiting “next FS op” returns while dormant and misses G1. A canceled refresh context also makes `LocalArmSignal` report false even when markers exist. Fence disarm to the exact empty snapshot/force generation and use an independent local-scan context; add the deterministic interleave regression.

4. **The user-writable remote index is read without a size/entry bound on automatic paths.** Mount probing uses `io.ReadAll` (`pkg/client/git_workspace_index.go:57-73`), and server-side DELETE reads the complete object with `ReadCtx(..., 0, -1)` (`pkg/server/git_workspace.go:405-428`). The repository's default file cap is 10 GiB, so a large `/.drive9/git-workspaces/index.json` can allocate until the mount or server is OOM-killed merely by mounting or deleting a workspace. Enforce a small maximum document size and entry count before decoding, and test oversized rejection.

Housekeeping: `gofmt -l` still reports `pkg/client/git_workspace_index.go`, `pkg/fuse/git_workspace.go`, and `pkg/server/git_workspace_test.go`.

Validation: exact base/head and hosted `ci`, local-e2e, CodeRabbit are green; focused git-workspace/FUSE tests pass under `-race`; `pkg/client` and `pkg/server` compile; vet, bash syntax, and `git diff --check` pass. Full local package execution is not authoritative here because `pkg/client` TestMain requires rootless Docker, and unrelated pre-existing `pkg/fuse` mount-wait tests fail on this macOS host.

qiffang (changes requested):
REVISE on exact head `30f9021a0411f21e1442de94afb44a29808093e9`.

I agree with the narrower positives from the other reviews: the old `pendingForce` self-amplification/list-storm bug is closed, and the 250ms local-arm scan throttle is covered. I also agree this head still cannot merge. In addition to the delete/index invalidation issues already raised, there is a server-side invariant gap that makes the soft-delete contract unsafe:

## Blocker: deleted workspaces remain mutable/readable through subresource APIs

This PR makes `DELETE /v1/git-workspaces/{id}` soft-delete the row and treats index cleanup as best-effort. That only works if *all* runtime paths treat `git_workspaces.status=deleted` as authoritative. New discovery does (`ListGitWorkspaces` filters `status = live`), but the subresource routes do not.

The server routes `tree`, `git-state`, `object-packs`, and `overlay` directly by `workspaceID` without first checking that the workspace row is still live. The handlers then call store methods that also operate by `workspace_id` only:

- `handleGitTreeReplace/List` -> `ReplaceGitTreeNodes` / `ListGitTreeNodes`
- `handleGitStateUpsert/Get` -> `UpsertGitState` / `GetGitState`
- `handleGitOverlayUpsert/Get/List` -> `UpsertGitOverlayEntry` / `Get/ListGitOverlayEntries`
- `handleGitObjectPackUpsert/Get/List` -> `Upsert/Get/ListGitObjectPack(s)`

The schema has no FK/live-status constraint to save this. FUSE writeback uses the cached runtime's `workspaceID` directly when it commits overlay writes. So a reachable sequence is:

1. Mount A is already armed and has a cached runtime for workspace `ws1`.
2. Mount/CLI B deletes `ws1`.
3. Index cleanup fails or no index/SSE/local marker reaches A, which this PR explicitly allows.
4. Mount A continues to use the stale runtime and posts overlay/git-state/tree/object-pack calls to `/v1/git-workspaces/ws1/...`.
5. The server accepts those calls even though `ws1` is deleted.

This also makes future reuse dangerous. `handleGitWorkspaceUpsert` reuses the existing row ID returned by `GetGitWorkspaceByRoot` and flips it back to `status=live`; stale subresource writes made after deletion can therefore leak into/corrupt a later workspace registered at the same root.

Soft-delete can be the durable authority only if the subresource APIs reject non-live workspace IDs (404/410) before read/write, or DELETE otherwise atomically disables/removes the subresource state and invalidates stale runtimes. Please add regressions that cover a stale armed runtime using the old `workspace_id` after DELETE, including the index-cleanup-failure/no-event case, and verify overlay/git-state/tree/object-pack reads/writes are rejected and cannot affect a recreate of the same root.

Local validation: focused `pkg/fuse` on-demand tests and `pkg/gitcache` passed; `pkg/client`/`pkg/server` package compile passed. Runtime `pkg/client`/`pkg/server` tests are blocked locally by the repo's rootless Docker/TestMain requirement.

qiffang:
Follow-up on exact head `30f9021a0411f21e1442de94afb44a29808093e9`: I independently confirmed one more blocker that makes the best-effort invalidation failure data-corrupting, not merely stale discovery.

**Soft-delete does not fence git-workspace subresources.** `handleGitWorkspaceObject` dispatches `tree`, `git-state`, `object-packs`, and `overlay` requests using only the path `workspaceID` (server lines 478-580). Their datastore reads/writes likewise match only `workspace_id` (for example `ReplaceGitTreeNodes` lines 307-378, `UpsertGitState` 411-434, `UpsertGitObjectPack` 456-478, and `UpsertGitOverlayEntry` 524-565); none requires `git_workspaces.status = active`. DELETE only updates that status (lines 230-246).

Therefore, when index cleanup fails, an already-armed cached mount can keep using its stale runtime **and successfully write overlay/git-state/tree/object-pack state after DELETE**. Re-registering the same root then reuses the deleted row's `workspace_id` (server lines 278-303), reactivating all of that stale state in the new workspace.

The fix must make every subresource read/write conditional on the workspace being live, with write fencing atomic against DELETE (a read-then-write status check would still have a TOCTOU race), or make DELETE atomically disable/remove that state. Please add discriminating regressions for stale subresource access after DELETE, concurrent DELETE vs subresource write, and same-root recreation not inheriting deleted runtime state.

mornyx:
@qiffang Thanks for the follow-ups on `30f9021a`. Addressed on head `bb8ae25e`.

### 1. Armed mounts no longer serve a deleted workspace forever
Index cleanup is still best-effort (DELETE stays **200** after a successful soft-delete; the row is the durable authority). Convergence no longer depends on that hint alone:

- Armed mounts revalidate cached workspace IDs with **GET-by-id** (1s coalesce, not a list poll), and drop a runtime on 404 / `status=deleted`.
- Empty-list / all-gone disarm is unchanged except it cannot hide a newer local arm.

Regression: `TestArmedMountDropsDeletedWorkspaceWithoutIndexEvent` — two mounts, DELETE with no index/SSE change, the other mount stops serving after the liveness window.

### 2. `--fast` registration is resumable
`clone --fast` and `worktree add --fast` skip `git clone` / `git worktree add` when the **target is already a work-tree root** (`target/.git` present). Nested directories inside another checkout are not treated as a resume target. Re-run then retries upsert + index + local markers.

Regressions: `TestExistingFastGitCheckout` (including nested-dir), `TestPublishGitWorkspaceIndexRetryable`, `TestMarkLocalGitWorkspaceRegisteredRetryable`.

### 3. Empty refresh cannot wipe a newer force
The empty-list leader fences disarm to the `pendingForceGen` / `localArmGen` captured before the marker scan. The scan uses an independent context so a canceled refresh cannot hide existing markers.

Regressions: `TestEmptyRefreshDoesNotClobberNewerForce`, `TestEmptyRefreshCanceledContextStillSeesLocalMarkers`.

### 4. Remote index reads are bounded
Automatic client and server index reads reject documents over **1 MiB** or **4096** entries (stat size, limited read, entry count).

Regressions: `TestReadGitWorkspaceIndexRejectsOversizedDocument`, `TestParseGitWorkspaceIndexRejectsTooManyEntries`, `TestGitWorkspaceDeleteRejectsOversizedIndex`.

### 5. Soft-delete fences subresources (atomic vs DELETE)
`DELETE` marks `status=deleted` and wipes tree / git-state / object-pack / overlay rows in the **same transaction**. Subresource writes take the live workspace row `FOR UPDATE` before mutating; reads require a live row (404 otherwise). Same-root recreate reuses the id but cannot inherit wiped children.

Regressions: `TestDeleteGitWorkspaceWipesSubresourcesAndFencesWrites`, `TestDeleteGitWorkspaceConcurrentWithOverlayWrite`, `TestGitWorkspaceSubresourcesRejectedAfterDelete` (includes index-cleanup-failure + recreate isolation).

Please re-review when convenient.

srstack:
## Review at `feat/git-workspace-on-demand-discovery` (13 commits, +4935/−138, 27 files)

### Overview

Replaces the ~1/sec `ListGitWorkspaces` poll that ran on every active FUSE mount with an on-demand `DORMANT → ARMED` state machine. Three arming signals: local markers under `LocalRoot/git-workspaces/` (same-machine live `--fast`), a tenant-absolute remote index at `/.drive9/git-workspaces/index.json` read via ordinary FS Stat (cross-sandbox remount), and SSE on the index path while armed. A mount that never runs `--fast` now issues zero list calls and at most one index Stat.

Two design choices give this the right blast radius: the index is read through the existing FS path rather than a new API, and it is existence-only (id + root_path + kind) while runtime content still comes from `ListGitWorkspaces`. A stale or corrupt index can therefore cause a spurious arm but never wrong file contents.

Several details are handled more carefully than I expected:

- **Bounded index reads** — `MaxGitWorkspaceIndexBytes` (1 MiB) / `MaxGitWorkspaceIndexEntries` (4096), checked at Stat *and* enforced with `io.LimitReader` on the body. Matters because mount-time probes read a user-writable document whose default file cap is 10 GiB.
- **`LocalArmSignal` generation hashes the marker set** (names + bodies) rather than max mtime, with a comment explaining that coarse-mtime filesystems would otherwise miss a second `MarkWorkspaceRegistered` in the same second.
- **`pendingForce` + generation counter** so a force arriving mid-flight isn't dropped when the in-flight soft list succeeds — the second-`--fast` case.
- **Empty index is CAS-written, not deleted**, with the reasoning (unconditional delete races a concurrent upsert) in a comment.

### Design question: what does the index buy over empty-list backoff?

Worth asking explicitly, because it determines whether the durable state is justified.

Exponential backoff on empty list results (1s → 60s, reset on local arm signal) is a much smaller change: no new durable document, no CAS, no server-side write path, no client/server format contract. It removes nearly all the wasted traffic.

Against that, the PR's own "Product boundaries" section documents a hole the backoff approach doesn't have: once `dormantConfirmed` latches, a first `--fast` from **another machine** is not guaranteed visible without remount. Backoff always discovers within its cap.

So the index buys next-op discovery latency instead of ≤60s, and pays with a cross-machine discovery gap plus a durable structure that now has three independent writers (see #1). If next-op latency on the cross-sandbox path is a hard product requirement, that's a fine trade — but it's the only thing the index buys over backoff, and I'd like it stated as a requirement rather than implied, since it's what justifies everything in `git_workspace_index.go` and the server-side cleanup.

### Findings

**1. The index wire format is defined independently in three places.** `pkg/client/git_workspace_index.go` declares `GitWorkspaceIndex`/`Entry` + `GitWorkspaceIndexPath`; `pkg/server/git_workspace.go` in this repo declares `gitWorkspaceIndexDoc`/`DocEntry` + `gitWorkspaceIndexPath`; tidbcloud/fs#51 carries a third copy. All three `json.Unmarshal` into their own struct and write back a full `MarshalIndent` re-encode, so any field one writer doesn't know is silently stripped from **every** entry on the next write by another writer.

The shapes match today, so nothing is broken. But this PR's premise is that client and server are both writers now, and the failure mode is delayed and asymmetric: add a field client-side, it survives until the next server-side DELETE touches the index, then it's gone — no error anywhere, and the symptom surfaces in the client far from the cause.

Either mitigation works: decode into `map[string]json.RawMessage` (or embed a catch-all) and rewrite only owned keys; or promote the type into a shared package both sides import. Since `/.drive9/packs` is cited as the precedent and presumably has the same exposure, the shared-type route looks worth the plumbing.

**2. A failed index publish fails an otherwise-successful `git clone --fast`.** In `cmd/drive9/cli/git.go`:

```go
if err := publishGitWorkspaceIndexEntry(cmdCtx, c, ws); err != nil {
    return fmt.Errorf("publish git workspace index: %w", err)
}
```

At this point the clone completed and the workspace is registered server-side — the DB is authoritative and the workspace is fully usable. The index is described throughout as an arming *hint*. A transient CAS exhaustion or network blip on a best-effort cache write therefore surfaces as a failed clone, when the only real consequence is that a future remount *on a different machine* won't auto-discover it (same-LocalRoot mounts arm from the local marker, written just after).

I'd downgrade this and `markLocalGitWorkspaceRegistered` to warnings and let the command succeed. If the strictness is deliberate, the message should at least distinguish "clone failed" from "clone succeeded, discovery hint not published."

**3. Subresource wipe is unbounded in a single transaction.** `DeleteGitWorkspace` now wraps the soft-delete plus four child deletes in one `InTx` — `git_workspace_tree_nodes`, `git_state`, `object_packs`, `overlay`. The correctness gain is real and well-motivated (the comment explains it stops a later same-root upsert inheriting stale rows and closes a concurrent-subresource-write race).

But tree nodes scale with repo file count, so a large monorepo workspace makes this a very large single transaction. `21dce6d feat(datastore): batched post-order recursive directory delete` in the fs repo suggests this class of problem has already bitten once. Worth batching, or confirming the expected upper bound on tree nodes per workspace is low enough not to matter.

**4. Two questions on `dormantConfirmed`, not change requests.** The latch is load-bearing for the optimization and the cross-machine consequence is documented, so I'm not asking for different behaviour — but:

- The 401/403 → permanent-dormant path means an `fs_scoped` token mount can never arm, since it can't read `/.drive9/`. Intentional per the comment (avoids an infinite Stat loop), but combined with the latch it means scoped-token mounts silently lose git-workspace support. Is that the intended product story, or should it surface once at a level the user sees?
- Is there any path out of `dormantConfirmed` besides a local marker or remount? If not, the `--git-workspaces=on` escape hatch listed under follow-ups is the only recovery for the cross-machine case, which makes it more than a nice-to-have.

**5. Server-side CAS loop has no backoff.** The client's `mutateGitWorkspaceIndex` sleeps `50ms * (attempt+1)` between attempts; `removeGitWorkspaceIndexEntry` in this repo's `pkg/server/git_workspace.go` retries immediately with `maxAttempts = 4`. Since the index is one tenant-wide document, concurrent deletes of *different* workspaces contend on the same revision, and four back-to-back attempts is a thin budget. Same finding I raised on tidbcloud/fs#51; noting it here because this repo carries its own copy.

### Previously-flagged items now fixed

Confirming three things that were problems in an earlier revision and are resolved at HEAD, so they don't get re-litigated:

- The local-arm throttle no longer gates on `armed` — line 311 applies the 250ms window while dormant too, so a busy dormant mount no longer does an `os.ReadFile` + `os.ReadDir` per FS op. That was the more serious of the two, since it traded network calls for syscalls on exactly the path being optimized.
- DELETE no longer returns 500 when index cleanup fails — it logs and returns success, so a user-facing `rmdir` doesn't see EIO for a workspace that is in fact gone.
- The dead `if expected == 0 { expected = 0 }` block in the client CAS path is gone.

### Smaller notes

- `ReadGitWorkspaceIndex`'s second 404 branch (`ReadStream` not-found after a successful Stat) discards the `rev` it already holds and returns `(nil, 0, nil)`. Correct today, but that path is reachable via the Stat/Read TOCTOU, and returning `rev` would let the caller CAS against what it actually saw.
- `RemoveGitWorkspaceIndexEntry` filters in place with `out := idx.Workspaces[:0]`, aliasing the backing array. Safe here (freshly parsed, discarded after), but breaks quietly if the mutator is ever applied to a shared slice.
- The empty-index CAS branch `continue`s without the backoff sleep the non-empty branch applies, so exhausting retries there is faster than intended.
- `e2e/git-workspace-ondemand-smoke-test.sh` is 1135 lines. Asserting on perf counters (`refresh=0` for a never-`--fast` mount) is the right way to test "we stopped doing the thing" rather than testing internals — but a script this size is hard to debug on CI failure. Splitting A–F into files sharing a helper would localize failures.

### Tests

Strong coverage: `git_workspace_ondemand_test.go` (+665), `arm_test.go` (+159), `git_workspace_index_test.go` (+143), `pkg/server/git_workspace_test.go` (+328), plus the E2E suite wired into the PR gate via `check_required`. The E2E table maps each scenario to a design goal (G0–G4, AC5), and the tenant-isolation note — filtering assertions by the suite's unique `RemoteRoot` rather than requiring a tenant-global 404 — shows the shared-key failure mode was actually considered.

Gaps worth filling:

- Nothing pins the unknown-field round-trip from #1. A test writing a doc with an extra top-level key and an extra per-entry key, running upsert + remove, then asserting both survived, would fail loudly when either side's struct drifts.
- No test that either CAS loop retries on conflict — a fake returning one `ErrRevisionConflict` then succeeding would cover the loop and the backoff.
- Nothing asserts the CLI behaviour in #2.

### Verdict

The core design is sound and the implementation is careful where this kind of change usually goes wrong. The measured outcome (~1 list/sec → 0 for ordinary mounts) is a real win, and the earlier hot-path and DELETE-500 problems are fixed.

**#1** is what I'd want resolved before merge — three independent writers of one JSON document, each silently dropping unknown fields, is a trap that fires during a future unrelated change and presents as data loss in the client. **#2** is small and has direct user impact. **#3** needs a bound or a batch. **#4** is two questions, not a change request. The design question above isn't a blocker either, but the answer determines whether the durable index is carrying its weight.

Coordination note: this and tidbcloud/fs#51 are two halves of one feature sharing the index format. Whichever merges second inherits both compatibility matrices — worth confirming the client's "server may also clean up the index" assumption matches fs#51's final best-effort-on-DELETE behaviour, since that PR changed from 500 to 200 mid-review.

---

*Note: the injected review rule targets tidbcloud/fs#49, which is already merged (I posted a post-merge note there). Reviewed mem9-ai/drive9#835 per your request.*

qiffang (changes requested):
REVISE on exact head `54e1858b0eddd8302e25421fae25b97b4576efff`.

Blocking finding:

1. `clone --fast` / `worktree add --fast` now silently treat any existing git checkout at the destination as a resumable Drive9 partial registration. In `cmd/drive9/cli/git.go`, both handlers skip the actual `git clone` / `git worktree add` whenever `existingFastGitCheckout(target)` returns true (`git.go:113-120`, `git.go:315-323`). But that helper only verifies that `<target>/.git` exists and `git rev-parse --is-inside-work-tree` is true (`git.go:1045-1066`); it does not prove the checkout was created by the same failed Drive9 operation, does not check that the existing clone's remote matches the requested `repoURL`, and does not check that an existing worktree belongs to the supplied `basePath` / `CommonWorkspaceID`. The rest of the handler then registers `HEAD`, tree, and `.git` state from the existing target while storing the caller-supplied repo/base metadata (`git.go:121-190`, `git.go:324-405`). That changes normal git safety semantics: a typo or stale destination that Git would have rejected as non-empty can now be registered as a Drive9 workspace for the wrong repo/base. For linked worktrees this can register a checkout unrelated to the declared base workspace. The current tests only prove the helper detects an existing checkout root and rejects nested dirs; they do not cover mismatched repo/base or prove a resumable Drive9 partial-registration identity.

Required before merge: resume must be scoped to the actual Drive9 partial-registration identity for the same command/repo/base, or otherwise reject existing destinations and let normal git safety semantics hold. Add regression coverage for at least: existing checkout with different origin passed to `clone --fast` must not register as the requested repo, and existing worktree/checkout not linked to the supplied base passed to `worktree add --fast` must not register under that base.

Old blockers rechecked and materially closed on this head:
- Subresource live-status fence: DELETE soft-deletes and wipes tree/git-state/packs/overlay in one transaction; write paths take the live workspace row with `FOR UPDATE`, and server/datastore tests cover delete-after-access, concurrent DELETE vs overlay write, and same-root recreate not inheriting stale state.
- Stale armed mount after missed index/SSE delete: cached runtimes now do a coalesced GET-by-id liveness check and drop deleted/missing workspaces; there is a regression for stale remote mount deletion without index event.
- Empty-list disarm race: empty refresh now uses an independent local marker scan and checks pending/local generation before disarming; regressions cover newer force and canceled context.
- Index OOM path: client index reads use Stat plus `io.LimitReader` with 1 MiB / 4096-entry caps; server DELETE index cleanup reads at `max+1` and logs cleanup failure while preserving soft-delete as durable truth.

Validation I ran locally:
- `git diff --check 49e7a0f...HEAD`
- changed Go files `gofmt -l`
- `go test ./pkg/gitcache ./pkg/fuse -run 'GitWorkspace|OnDemand'`
- `go test -race ./pkg/fuse -run 'GitWorkspace|OnDemand'`
- `go test ./cmd/drive9/cli -run 'Git|Fast|Worktree|Index'`
- `go test -c ./pkg/client ./pkg/server ./pkg/datastore`
- `go vet ./cmd/drive9/cli ./pkg/client ./pkg/datastore ./pkg/fuse ./pkg/gitcache ./pkg/server`
- `bash -n e2e/git-workspace-ondemand-smoke-test.sh e2e/git-ops-smoke-test.sh e2e/smoke-all.sh`

`go test ./pkg/datastore -run 'GitWorkspace|DeleteGitWorkspace'` still cannot run in this local environment because TestMain panics on missing rootless Docker/testcontainers; hosted `ci` and `drive9-server-local e2e smoke` are green on this exact head.

qiffang (changes requested):
REVISE on exact head `54e1858b0eddd8302e25421fae25b97b4576efff`.

The previous empty-refresh race, bounded-index-read, deleted-subresource fence, and missed-index-delete convergence issues are materially improved. Two correctness blockers remain:

1. **`--fast` resume accepts an arbitrary existing checkout and can register mismatched repository metadata.** `clone --fast` now skips clone whenever `existingFastGitCheckout(target)` sees any work-tree root. It never proves the checkout was created by this interrupted command and never compares `remote.origin.url` to the requested `repoURL`. The command then reads `HEAD`/tree from the existing checkout but registers `RepoURL` from the CLI argument. Therefore `drive9 git clone --fast repo-A /path/containing-repo-B` silently records repo-A metadata for repo-B objects, so later object hydration can fetch from the wrong repository. `worktree add --fast` has the analogous gap for an unrelated linked checkout: `linkedWorktreeGitDirMetadata` does not reject a gitdir outside the requested base repo; it falls back to a synthetic `worktrees/<name>` relation. Resume must be tied to verifiable provenance/identity for the requested clone/worktree, or fail closed. Add regressions for an existing checkout of a different origin and an existing linked worktree belonging to a different common repo.

2. **Same-root recreation reuses the deleted workspace ID, so an old remote runtime can write into the new generation.** DELETE wipes children and write-locks correctly while the row is deleted, but `handleGitWorkspaceUpsert` intentionally reuses `existing.WorkspaceID` when reviving that root. A remote mount that missed the best-effort index/SSE invalidation still holds the same ID. If the root is recreated before its 1s liveness check, GET-by-id observes the row active again and cannot distinguish the old runtime from the new one. Its cached overlay/git-state/tree/object-pack writes are accepted against the newly active row, repopulating the recreated workspace with stale state. The current recreate tests only assert that rows are empty immediately after upsert; they do not exercise a stale pre-delete writer after recreation. Recreate needs a new identity/generation that all subresource operations fence, with a regression: old runtime loaded → DELETE → same-root recreate → old-generation subresource write must be rejected.

Validation: exact-head hosted `ci`, local-e2e, and CodeRabbit are green; changed files pass `gofmt -l`, `git diff --check`, package compile, and `go vet`. New focused CLI/FUSE tests pass. The broad local suite is not authoritative here: `pkg/client` is blocked by rootless Docker/TestMain, and `pkg/fuse` has two unrelated mount-permission test failures on this Mac.

mornyx:
@qiffang @srstack Addressed on head `7c2d9369`.

### Resume is identity-scoped (qiffang)
`clone --fast` skips `git clone` only when the destination is already a work-tree root **and** `origin` matches the requested repo (https/ssh/` .git` normalized). `worktree add --fast` skips add only when the destination shares the base checkout’s `git-common-dir`. Any other existing checkout is rejected — it is not registered as the caller-supplied repo/base.

`linkedWorktreeGitDirMetadata` no longer invents `worktrees/<name>` for a gitdir outside the base. Relative metadata is computed in the worktree’s own gitdir namespace first (so a FUSE-absolute `gitdir:` still works), then the overlay path; both failing is an error.

Regressions: `TestFastCloneResumeRejectsDifferentOrigin`, `TestFastWorktreeResumeRejectsUnrelatedCheckout`, `TestLinkedWorktreeGitDirMetadataRejectsForeignGitDir`, `TestLinkedWorktreeGitDirMetadataAcceptsAbsoluteGitDirSameBase`.

### Recreate mints a new workspace id (qiffang)
Reviving a deleted root no longer reuses the deleted id. `ReplaceDeletedGitWorkspace` re-keys the row; GET/subresource on the old id is 404, so a stale armed runtime cannot write into the new generation. Live upserts at the same root still keep the id.

Regressions: `TestGitWorkspaceSubresourcesRejectedAfterDelete` (old-id write after recreate), `TestGitWorkspaceLiveUpsertKeepsID`, datastore replace-deleted coverage.

### Index wire type and the smaller items (srstack)
Client and server now share `pkg/gitwsindex`. Unknown top-level and per-entry keys survive rewrite (`TestIndexPreservesUnknownFields`, `TestUpsertPreservesUnknownIndexFields`).

Index/marker publish still fails the command (index remains a required discovery hint), but the error now says the workspace was registered and the discovery hint failed.

Child-table wipe on DELETE is batched (`LIMIT 5000` in the same transaction). Server index CAS retries use the same 50ms backoff as the client. Read-after-Stat 404 keeps the observed revision; remove no longer aliases the slice with `[:0]`.

Please re-review when convenient.

srstack:
## Follow-up review at `7c2d9369` (14 commits, +5579/−145, 29 files)

One new commit since my review. Four of the five findings are addressed, and the fix for #1 is better than what I suggested.

### #1 — shared wire type with unknown-key preservation ✅

New `pkg/gitwsindex` package (+148, +69 tests) with `Path`, `MaxBytes`, `MaxEntries` and custom marshalers on both levels:

```go
// Package gitwsindex is the shared wire type for /.drive9/git-workspaces/index.json.
// Client and server both mutate this document; unknown keys must survive a rewrite.
```

`Index` and `Entry` each carry an unexported `extra map[string]json.RawMessage`; `UnmarshalJSON` routes known keys and collects the rest, `MarshalJSON` writes `extra` back first and then overwrites the owned keys. So a writer that doesn't know a newer field can't strip it — at the top level *or* per entry, which is the part that would have been easy to miss.

Adoption is complete: client references it 7×, server 6× (`gitWorkspaceIndexPath = gitwsindex.Path`, `gitwsindex.Parse`, `[]gitwsindex.Entry`), and `grep` for the old duplicate type names returns zero. Three independent definitions collapsed to one, so the drift I was worried about is now a compile error rather than silent data loss.

This is stronger than the `map[string]json.RawMessage` decode I offered as the cheaper option — the shared type also removes the possibility of the two sides disagreeing on limits or path.

### #3 — subresource deletes batched ⚠️ (statement bounded, transaction still isn't)

`gitWorkspaceSubresourceDeleteBatch = 5000`, with each table's `DELETE ... LIMIT ?` in a drain loop that exits when a batch comes back short. That bounds individual statement size, which was the concrete thing I asked for.

Worth being explicit that the transaction is still unbounded, though: `deleteGitWorkspaceSubresourcesTx` takes a `*sql.Tx` and every batch runs inside the single `InTx` opened by `DeleteGitWorkspace`. A monorepo workspace with 500k tree nodes now issues 100 statements instead of 1, but they accumulate in one transaction's undo log and hold row locks until commit. The batching helps the statement-level blowup and does nothing for transaction duration or lock footprint.

That may well be the right call — the transaction boundary is load-bearing here, since the whole point is that a same-root upsert can't interleave and inherit stale rows, and `lockLiveGitWorkspaceTx` (new in this commit) leans on the same property. Splitting into per-batch transactions would reintroduce that race. So: fine as a deliberate trade, but the comment currently explains *why the transaction exists* and not *why unbounded batching inside it is acceptable*. If there's a known upper bound on tree nodes per workspace, stating it there would close the question for the next reader.

### #5 — server CAS backoff ✅

`case <-time.After(time.Duration(50*(attempt+1)) * time.Millisecond)` now matches the client loop. Both sides back off identically on a document every workspace delete contends on.

### #2 — strictness kept, message fixed ✅ (your call, and it's defensible)

The publish failure still fails the command, but the message changed:

```go
return fmt.Errorf("registered git workspace %s, but publishing discovery index failed: %w", ws.WorkspaceID, err)
```

That was my fallback suggestion, and it resolves the actual harm I was pointing at — the old message read as "clone failed" when the clone had succeeded and the workspace was usable. A user seeing this now knows the workspace exists and only cross-machine discovery is degraded. Keeping the non-zero exit is a reasonable read of "don't let a half-published state pass silently," especially since the same command also writes the local marker.

### New in this commit: `--fast` resume and recreate scoping

The commit's headline change replaces `clearLocalGitWorkspaceDeleted` with `markLocalGitWorkspaceRegistered`, so a resumed `--fast` writes a positive arm signal rather than only clearing a stale deletion marker. That closes a gap where an empty local cache couldn't discover a re-registered id.

The version-skew handling in `gitWorktreeRemove` is now written down where it matters:

```go
// - New server: DELETE soft-deletes + best-effort index cleanup (still 200 on
//   index failure); client Remove below is dual-write / immediate convergence.
// - Old server: DELETE only soft-deletes; client Remove below still required.
```

and the client always attempts `RemoveGitWorkspaceIndexEntry` after DELETE, collecting failures into `cleanupErrs` rather than aborting. This correctly encodes fs#51's *final* behaviour — best-effort cleanup returning 200 — which is the coordination point I flagged, since that PR changed from 500 to 200 mid-review. Good that the client assumption tracks the server's landed semantics rather than its original description.

### Still open

- **#4's two questions** — whether scoped-token mounts silently losing git-workspace support is the intended product story, and whether anything recovers from `dormantConfirmed` besides a local marker or remount. Neither is code-visible; both are still worth an answer in the description.
- **The design question** — what next-op latency buys over empty-list backoff. Less pressing now that `pkg/gitwsindex` makes the format a single maintained type rather than three, which was most of the ongoing cost I was weighing against ~30 lines of backoff.
- **Test gaps** — the unknown-field round-trip is presumably covered by `pkg/gitwsindex/index_test.go` (+69); worth confirming it asserts survival through a full upsert→remove cycle at both levels, not just Marshal/Unmarshal symmetry. The CAS-conflict retry path still looks untested on both sides.

### Verdict

The blocking item is resolved, and resolved well — a shared type with explicit unknown-key preservation is the durable fix, not a patch. #3 is materially better and its remaining unboundedness is arguably inherent to the transaction it needs. #5 and #2 are done.

Nothing here blocks merge from my side. The two #4 questions are worth answering in the description before this ships, since they define the supported discovery matrix that users will hit.

---

*Note: the injected review rule targets tidbcloud/fs#49, which merged as `d1b3780a` (post-merge note posted there). Reviewed mem9-ai/drive9#835 per your request.*

qiffang (changes requested):
Re-review at exact head `7c2d936910c36e9b0b48a531c495b907b88e5c16`: **REQUEST CHANGES**.

The previous ordinary cases are materially fixed: a different `origin` is rejected for `clone --fast`, a different git-common-dir is rejected for `worktree add --fast`, and a sequential deleted-root recreate gets a new workspace id. Two correctness gaps remain.

### 1. Deleted-root generation fencing still has a read/delete/upsert race

`handleGitWorkspaceUpsert` reads the row at `pkg/server/git_workspace.go:269`, decides at `:276-282` whether to keep the old id, then performs `UpsertGitWorkspace` / `ReplaceDeletedGitWorkspace` only at `:309-315`. That decision is outside the write transaction.

A deterministic interleave is:

1. upsert reads root `R` as `live` with id `old`, so it chooses `workspaceID=old`;
2. concurrent DELETE commits `status=deleted` and wipes the old subresources;
3. the paused upsert resumes and calls `UpsertGitWorkspace(old, R, status=live)`;
4. `ON DUPLICATE KEY UPDATE ... status=VALUES(status)` in `pkg/datastore/git_workspace.go:152-170` revives the deleted row with the **old id**.

The final state is live with `old`, so a stale runtime holding that id can again write into the recreated workspace—the exact cross-generation contamination this commit is meant to prevent. This execution cannot be linearized to the documented contract: if the upsert is before DELETE, final state should be deleted; if after DELETE, it must get a new id.

Please move “live update vs deleted-root replacement” into one datastore transaction/conditional operation that decides against the locked current row. Add a discriminating regression that pauses after the live read, commits DELETE, then resumes the upsert and asserts the returned id is new and the old id/subresources stay NotFound.

### 2. Remote identity normalization can accept distinct generic Git repositories

`gitRemoteIdentity` at `cmd/drive9/cli/git.go:1145-1165` lowercases the entire repo path and drops the SSH username. Git only guarantees DNS host case-insensitivity; repository paths and account-relative SSH roots can be case/user-sensitive on a generic server. For example, `ssh://alice@example.test/Repo.git` and `ssh://bob@example.test/repo.git` compare equal here even though they may resolve to different repositories. The fallback also lowercases local paths.

That leaves the original wrong-checkout risk for supported non-GitHub remotes. Normalize only properties that are universally identity-insensitive (host and optional `.git` / slash shape), and keep path/account identity unless applying an explicit provider-specific equivalence. Add regressions for case-distinct paths and distinct SSH users being rejected; retain the existing GitHub HTTPS vs `git@github.com:` equivalence explicitly.

Validation: focused CLI/gitwsindex tests and `-race` pass; changed packages compile and vet; gofmt and diff-check are clean. Hosted `ci` and local-e2e are green. Runtime client/datastore/server tests remain unavailable locally because their TestMain requires rootless Docker; the added sequential tests do not exercise the race above.

qiffang (approved):
GREEN on exact head `7c2d936910c36e9b0b48a531c495b907b88e5c16`.

I re-reviewed this head against the previous blockers and the new delta. The two remaining blockers from `54e1858b` are closed:

- `clone --fast` resume now verifies the existing destination `origin` matches the requested repo identity before skipping `git clone`; mismatched origins fail instead of silently registering the wrong checkout. `worktree add --fast` resume now verifies the destination shares the base checkout's git common dir; unrelated checkouts fail. Regression coverage includes mismatched origin and unrelated worktree checkout.
- Same-root recreate after DELETE now mints a new `workspace_id` via `ReplaceDeletedGitWorkspace`; old-id `GET`/subresource writes return NotFound, and recreated workspaces do not inherit stale overlay. Server tests cover delete with corrupt index cleanup, subresource read/write rejection after delete, recreate new id, old-id write rejection, and no inherited overlay.

I also rechecked the earlier fixes:

- Subresource writes take a live workspace row lock in the same transaction (`lockLiveGitWorkspaceTx` / `withLiveGitWorkspaceTx`); reads require live status. This closes the DELETE-vs-subresource-write data pollution path rather than relying on a handler precheck.
- DELETE soft-deletes and wipes tree/git-state/object-pack/overlay rows in one transaction; index cleanup remains best-effort but no longer carries correctness.
- The remote index parser/writer is bounded and shared through `pkg/gitwsindex`; unknown fields are preserved.
- FUSE refresh still uses generation guards so empty-list disarm does not swallow newer force/local-arm state.

Validation:

- Hosted checks on this exact head: `ci` success, `drive9-server-local e2e smoke` success, CodeRabbit success.
- Local: `git diff --check`; touched-file `gofmt -l`; focused CLI tests for git fast/resume/index paths; `pkg/gitwsindex` tests; focused FUSE git workspace tests; focused FUSE `-race`; compile for `pkg/client`, `pkg/server`, `pkg/datastore`, `pkg/fuse`, `cmd/drive9/cli`, `pkg/gitwsindex`; focused `go vet` all passed.
- Local runtime tests for `pkg/client` / `pkg/datastore` are still blocked by this machine's rootless Docker/TestMain setup before `-run`, consistent with prior reviews; hosted CI covers those paths.

No remaining blocker from me.

qiffang (changes requested):
REVISE on exact head `7c2d936910c36e9b0b48a531c495b907b88e5c16`.

I re-checked the latest delta after dev2's follow-up. The sequential fixes are real, but two correctness blockers remain:

1. Same-root recreate can still reuse the old workspace ID under a concurrent DELETE.

`handleGitWorkspaceUpsert` reads the existing row by root and decides whether to keep or replace the ID before it writes. If it sees a live row, it keeps `workspaceID = existing.WorkspaceID`; the later write goes through `UpsertGitWorkspace`. That store method uses `ON DUPLICATE KEY UPDATE ... status = VALUES(status)` on the `root_path` unique key. Interleaving:

- upsert reads root `/repo` as live `oldID` and decides to keep `oldID`;
- DELETE commits and marks `oldID` deleted / wipes subresources;
- upsert continues into `UpsertGitWorkspace` with `workspaceID=oldID` and `status=live`;
- the `root_path` duplicate update revives the deleted row with the old ID.

That reopens the cross-generation pollution path: a stale mount that missed index/SSE/local invalidation can still write through `oldID` into the recreated workspace. The live-vs-deleted decision and the update/replace must be in one row-locked transaction or equivalent atomic fence. Please add a regression that forces the "read live -> DELETE commits -> upsert continues" interleaving and proves the recreated workspace gets a new ID and old-ID subresource writes remain rejected.

2. Fast-clone remote identity is too broad for generic Git remotes.

`gitRemoteIdentity` lowercases the whole host/path identity and drops the SSH username for scp-like URLs. That makes distinct remotes compare equal on servers where path case or SSH account is part of the repository namespace, e.g. `alice@example.test:Repo` vs `bob@example.test:repo` both normalize to `example.test/repo`. This is safe only as an explicit GitHub-style normalization rule, not as the universal guard deciding whether to silently reuse an existing checkout instead of running `git clone`. At minimum, lowercase only the host for generic URLs and preserve the path/account unless applying a host-specific equivalence such as GitHub HTTPS <-> `git@github.com:`. Add regressions for SSH user mismatch and case-sensitive path mismatch.

The previously reported sequential paths are improved: mismatched ordinary origins and unrelated worktree common dirs are rejected; sequential delete/recreate now gets a new ID; subresource live fences, bounded index reads, and generation-guarded FUSE refresh are still materially better. Hosted checks are green, and my earlier local focused checks passed, but these two blockers keep the head from merge.

mornyx:
@qiffang Addressed on head `fd41c11f`.

### 1. Live-vs-deleted identity is decided under the row lock
The unlocked `GetGitWorkspaceByRoot` is only the self-link check. `CommitGitWorkspace` then `SELECT … FOR UPDATE` on `root_path` and either keeps a live id, replaces a deleted row with a new id, or inserts. A DELETE that commits after the unlocked read can no longer revive the old id via `ON DUPLICATE KEY UPDATE`.

Regression: `TestGitWorkspaceUpsertMintsNewIDWhenDeleteInterleavesAfterLiveRead` — pause after the live read, DELETE commits, upsert resumes → new id; old-id GET and overlay write stay NotFound.

### 2. Generic remote identity no longer folds path/account
`sameGitRemoteIdentity` lowercases only the host. Path and SSH user are compared as-is (after stripping `.git`). The only extra equivalence is GitHub HTTPS ↔ `git@github.com:`.

Regressions: SSH user mismatch and case-distinct generic paths are rejected; GitHub `https://github.com/org/repo.git` vs `git@github.com:org/repo.git` still matches.

Please re-review when convenient.

qiffang (changes requested):
REVISE on exact head `fd41c11fce75fcc813c1bdc6639dfb82ce26bd62`.

The previous `pendingForce` self-amplification/list-storm issue is fixed: ordinary waiters now join/consume an existing pending force without bumping `pendingForceGen`, and the current tests cover that path. Hosted checks are green. I still see three merge blockers:

1. **Corrupt remote index still creates permanent fleet polling.**
   `pkg/fuse/git_workspace.go:317-320` returns malformed/oversized index parse errors to the probe loop, and the loop only latches dormant for 401/403 at `pkg/fuse/git_workspace.go:261-270`. A malformed `/.drive9/git-workspaces/index.json` therefore retries forever with exponential backoff up to 30s instead of becoming dormant or being repaired. The server-side cleanup path also cannot repair it: `pkg/server/git_workspace.go:427-430` returns parse errors, and `handleGitWorkspaceDelete` only logs that cleanup failure at `pkg/server/git_workspace.go:377-380` while returning success. The new test comment at `pkg/server/git_workspace_test.go:307` explicitly preserves this state. One bad tenant-root index blob can make every new mount for that tenant keep probing/parsing forever, which defeats this PR's idle-poll elimination goal. Fix either by treating malformed/oversized index as a permanent dormant signal on mounts, or by rebuilding/removing it from authoritative `git_workspaces` state on the server; add regression coverage for the chosen behavior.

2. **`CommitGitWorkspace` can return a later generation by re-reading by root after commit.**
   The locked transaction now correctly decides live-vs-deleted identity, but `pkg/datastore/git_workspace.go:214` releases the lock and then calls `GetGitWorkspaceByRoot`. This interleaving is still possible: request A commits workspace generation X, request B deletes X, request C recreates the same root as generation Y, and A's post-commit root lookup returns Y. The CLI then uses the returned `WorkspaceID` for `ReplaceGitTree` / git-state writes (`cmd/drive9/cli/git.go:175`, `cmd/drive9/cli/git.go:188`), so A can write its repo/head metadata into C's generation. Return the exact workspace ID chosen inside the transaction, or fail if that exact row is no longer live; do not accept a successor generation by root. Add a regression that pauses after commit but before response lookup, completes delete+recreate, and verifies the older request cannot get or write the newer ID.

3. **Remote identity still collapses distinct endpoints.**
   `cmd/drive9/cli/git.go:1189` stores only `u.Hostname()`, so URL remotes lose scheme and explicit port before `sameGitRemoteIdentity` compares them. That makes generic remotes such as `ssh://git@example.test:2222/repo.git` and `ssh://git@example.test:22/repo.git` compare equal, and can also collapse HTTP/HTTPS or non-default GitHub ports into the provider-specific equivalence path. The scp-like parser also strips the leading slash from the path, so scp-relative and URL-absolute forms can be treated as the same endpoint. These can point to different services/repos, but `--fast` would skip the clone and register the caller's metadata against an unrelated checkout. Generic identity must preserve endpoint semantics including scheme, explicit port, and relative-vs-absolute path; provider-specific equivalence should be a narrow whitelist with tests for ports/schemes/root form.

Validation I ran locally on this head:
- `git diff --check 49e7a0f26146b38144904284130547396ad3f716...HEAD`
- touched Go `gofmt -l`
- `go test ./pkg/gitwsindex ./pkg/gitcache -count=1`
- focused `cmd/drive9/cli` git tests
- focused `pkg/fuse` git workspace tests, including `-race`
- `go test -c ./pkg/fuse ./pkg/gitcache ./cmd/drive9/cli ./pkg/gitwsindex ./pkg/client ./pkg/server ./pkg/datastore`
- `go vet ./pkg/fuse ./pkg/gitcache ./cmd/drive9/cli ./pkg/gitwsindex ./pkg/client ./pkg/server ./pkg/datastore`

Hosted `ci`, local e2e smoke, and CodeRabbit are green, but the three paths above are not covered by those checks.

mornyx:
@qiffang Addressed on head `6d9bc36e`.

### 1. Corrupt index no longer polls forever
Malformed or oversized `index.json` is a permanent dormant signal on the mount (same class as 401/403). Server DELETE rebuilds the document from live `git_workspaces` rows (excluding the deleted id) instead of leaving the bad blob in place.

Regressions: `TestProbeIndexCorruptLatchesDormant` (probe stops, zero lists); `TestGitWorkspaceDeleteIndexFailureStillReturns200` now also asserts the rebuilt document parses and does not list the deleted workspace.

### 2. Commit returns the in-transaction workspace id
`CommitGitWorkspace` re-selects by `workspace_id` inside the locked TX and returns that row. It no longer `GetGitWorkspaceByRoot` after unlock. CLI/server tree and git-state writes use that id.

Regression: `TestCommitGitWorkspaceDoesNotReturnSuccessorGeneration` — after commit, DELETE + recreate run before the older request returns; it still reports the original id and cannot write into the successor.

### 3. Generic remotes keep endpoint shape
Identity now includes scheme, explicit port, SSH user, and relative-vs-absolute path. GitHub HTTPS ↔ `git@github.com:` is only for default ports on github.com.

Regressions: `:2222` vs `:22`, http vs https, GitHub `:4443` vs default, and scp-relative vs URL-absolute do not match.

Please re-review when convenient.

qiffang (approved):
APPROVE on exact head `6d9bc36e33ec5da156b27abc67b6a40a74a067e0`.

I re-reviewed the three previous blocker axes against the current diff:

- The root-to-workspace identity race is closed. `CommitGitWorkspace` now returns the row fetched by exact `workspace_id` inside the same `FOR UPDATE` transaction, and the post-commit root lookup is gone. The new hook regression exercises delete/recreate after commit and before return.
- The malformed/oversized remote index fleet-polling path is closed on both sides: mounts latch dormant for `IsUnreadableGitWorkspaceIndex`, and server-side DELETE rewrites an unreadable index from authoritative live `git_workspaces` rows with revision-guarded write.
- Remote identity parsing now preserves scheme, explicit port, user, and scp-relative vs URL-absolute shape for generic remotes, while keeping the narrow GitHub default-endpoint equivalence.

Validation: hosted `ci`, hosted `drive9-server-local e2e smoke`, and CodeRabbit are green on this head. Locally I ran `git diff --check`, focused git workspace/parser tests, focused FUSE index tests, and compile checks for the touched packages; all passed. No blocking findings from my side.

qiffang (changes requested):
Superseding my earlier approval on exact head `6d9bc36e33ec5da156b27abc67b6a40a74a067e0`: REQUEST_CHANGES.

The malformed-index repair and the in-transaction workspace ID return are closed, but the generic remote identity comparison is still incomplete. In `cmd/drive9/cli/git.go`, `gitRemoteParts` tracks scheme/host/port/user/path/absPath, but it does not preserve the sanitized query string, and `parseGitRemote` stores `path: strings.TrimPrefix(u.Path, "/")`, which is the decoded path rather than `EscapedPath()`.

That makes distinct generic endpoints compare equal. For example, `https://example.test/repo.git?tenant=A` and `https://example.test/repo.git?tenant=B` both reduce to the same parts, and `%2F` can be folded with a literal `/`. Generic Git HTTP endpoints can route different repositories by query or escaped path. If `sameGitRemoteIdentity` returns true here, `--fast` can reuse/skip clone against the wrong checkout and register metadata for the requested repo against another repo's local state.

Please include the non-secret sanitized query and escaped path in the identity comparison, and add regressions for query-distinct and encoded-path-distinct generic remotes. The GitHub default-endpoint normalization can stay narrow, but generic remotes should not lose these identity dimensions.

srstack (approved):
## Approving at `6d9bc36e`

All five items from my earlier review are addressed, and I verified the three blockers from qiffang's `fd41c11f` review against this head rather than taking the changelog at its word.

### My earlier blockers

**1. Dormant hot-path regression — fixed.** `tryArmFromLocalSignals` (`pkg/fuse/git_workspace.go:391`) now throttles on `localArmScanAt` in *both* states, so the gate is no longer `fs.git.armed &&`. A busy never-`--fast` mount no longer pays a `ReadFile` + `ReadDir` per lookup/getattr/readdir. The comment states the tradeoff explicitly (same-LocalRoot `--fast` arms on the first op after the window; cross-host relies on SSE + index).

**2. DELETE returning 500 after the row is gone — fixed.** `handleGitWorkspaceDelete` (`pkg/server/git_workspace.go:376-383`) logs `git_workspace_index_cleanup_failed` and returns 200. A user-facing `rmdir` no longer gets EIO for a workspace that is in fact deleted.

**Nit** — the `expected := rev; if expected == 0 { expected = 0 }` dead block is gone.

### qiffang's `fd41c11f` blockers

**1. Corrupt index no longer polls forever.** Malformed/oversized index is now classified by `client.IsUnreadableGitWorkspaceIndex` and latched into `dormantConfirmed` alongside 401/403 (`git_workspace.go:261`). Server-side, `removeGitWorkspaceIndexEntry` calls `rewriteGitWorkspaceIndexFromStore` on `gitwsindex.ErrUnreadable` or oversize, so the bad blob is rebuilt from live `git_workspaces` instead of being left to poison every new mount for the tenant.

**2. Commit returns the in-TX id.** `CommitGitWorkspace` re-selects via `getGitWorkspaceTx(ctx, tx, ws.WorkspaceID)` inside the locked transaction and returns that row. The post-unlock `GetGitWorkspaceByRoot` is gone, so the delete/recreate interleaving can no longer hand an older request a successor generation.

**3. Remote identity preserves endpoint shape.** `gitRemoteParts` now carries `scheme`, `port`, and `absPath`; `sameGitRemoteIdentity` compares all three, and GitHub equivalence is narrowed to `githubDefaultEndpoint` (github.com on default ports only). `:2222` vs `:22`, http vs https, and scp-relative vs URL-absolute no longer collapse.

### Verification

Ran on a detached worktree at this head:

- `go test ./pkg/gitwsindex ./pkg/gitcache ./pkg/client` — pass
- `go test ./pkg/fuse -run 'GitWorkspace|Dormant|LocalArm|ProbeIndex|Armed|EnsureGit'` — pass
- `go test ./pkg/server ./pkg/datastore ./cmd/drive9/cli` — pass
- `go vet ./pkg/... ./cmd/...` — clean

Confirmed the claimed regressions actually exist rather than trusting the descriptions: `TestProbeIndexCorruptLatchesDormant`, `TestGitWorkspaceDeleteIndexFailureStillReturns200`, `TestCommitGitWorkspaceDoesNotReturnSuccessorGeneration`, `TestFastCloneResumeRejectsDifferentOrigin`, `TestGitWorkspaceSubresourcesRejectedAfterDelete`, `TestUpsertPreservesUnknownIndexFields`.

### On my design objection

I argued empty-list backoff was strictly better. You made the case that next-op latency is the requirement and the index is what buys it, and the `dormantConfirmed` hole is documented under Product boundaries with a stated escape hatch. That is a legitimate product call and the shared `pkg/gitwsindex` wire type removed the drift risk that worried me most. Not going to keep relitigating it — the follow-ups you listed (server-side index maintenance, perf counters) are the right place to continue.

LGTM.

qiffang (approved):

PR #837, opened 2026-08-10
feat: publish Drive9 migration image

mashenjun:
## Summary by CodeRabbit

* **New Features**
  * Added support for building and publishing multi-platform Drive9 migration container images.
  * Added local and multi-platform migration image build commands.
  * Published images include version and build metadata with immutable references.

* **Documentation**
  * Added a runbook covering image publication, access configuration, verification, and migration worker startup options.

### DIFF
--- a/.github/workflows/publish-migration-image.yml
+++ b/.github/workflows/publish-migration-image.yml
@@ -1,0 +1,206 @@
+name: Publish migration image
+
+on:
+  workflow_dispatch:
+
+permissions:
+  contents: read
+
+concurrency:
+  group: publish-migration-image
+  cancel-in-progress: false
+
+env:
+  IMAGE: ghcr.io/drive9-ai/drive9-migration
+
+jobs:
+  metadata:
+    runs-on: ubuntu-24.04
+    outputs:
+      build_time: ${{ steps.metadata.outputs.build_time }}
+      trace_tag: ${{ steps.metadata.outputs.trace_tag }}
+    steps:
+      - name: Require the main branch
+        shell: bash
+        run: |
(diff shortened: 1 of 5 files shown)


qiffang (approved):

qiffang (approved):
GREEN / APPROVE on exact head `666e185baf08c3f8257e3baf541c5d00c79f0911`.

I found no blocking correctness or credential-handling issues. The workflow is manual-only, gated to `main`, uses top-level `contents: read`, publishes by digest rather than `latest`, verifies two Linux platform digests before manifest creation, smoke-tests the published platform images, and reports an immutable reference. The GHCR PAT is justified by the documented cross-org publish path (`mem9-ai/drive9` publishing to `ghcr.io/drive9-ai/...`); the runbook correctly limits it to `write:packages` and says not to place it in source/config/output/image. The Docker build context is default-deny via `Dockerfile.migration.dockerignore`, and the runtime image is `scratch` with no shell or extra OS surface.

Hardening recommendation, not a blocker: pin the GitHub Actions in this publishing workflow to full commit SHAs instead of mutable major tags (`actions/checkout@v4`, `docker/*@v3/v6`, `actions/upload/download-artifact@v4`). This path publishes a migration worker image operators will run with real infrastructure access, so SHA pinning is the right supply-chain posture even for first-party/vendor actions. Also consider `USER 65534:65534` in the scratch runtime and cosign/SLSA attestation before broad production rollout.

Local validation: `git diff --check`, focused migration tests, and `go build ./cmd/drive9-migration` passed. `actionlint` is not installed locally; hosted `ci`, `drive9-server-local e2e smoke`, and CodeRabbit are green on this head.

PR #845, opened 2026-08-19
feat: add object-store backends for fs and mount

mornyx:
## Summary

`drive9 fs` and `drive9 mount` can talk to object stores as well as drive9.

User-facing design: [`docs/design/object-store-cli.md`](https://github.com/mem9-ai/drive9/blob/feat/cli-object-store/docs/design/object-store-cli.md).

### Commands

URI schemes: `s3://`, `cos://`, `tos://`, `oss://`, `gs://` / `gcs://`, `az://` / `azure://`. `:/` and `drive9://` stay on drive9. An object URI is never treated as a local path.

```bash
drive9 fs ls s3://bucket/prefix/
drive9 fs cp ./file s3://bucket/prefix/file --force
drive9 fs cp s3://bucket/a.txt :/dst.txt
drive9 mount s3://bucket/prefix/ ./mnt
```

Overwrite of an existing object requires `-f` / `--force`. Object mounts are Unix-only; Windows still builds (stub). `--auth=local|server` is documented on `fs` and `mount` usage.

### Auth (default is server mint)

Omit `--auth` (or pass `--auth=server`): the CLI asks drive9-server for a short-lived, prefix-scoped session, then the **laptop talks to the object store directly**. File bytes do not go through drive9-server. Existing AWS/TOS env vars are **not** used unless you pass `--auth=local`.

`--auth=local` is the only native-chain escape (laptop env/profile/role, MinIO, GCS/Azure CLI).

Org admins configure this with TiDB Cloud AK/SK, same layer as `admin tenant extract-config`:

```bash
drive9 admin object-backend add --scheme s3 --bucket example \
  --credential-kind static --access-key-id ... --secret-access-key ... \
  --tidbcloud-public-key ... --tidbcloud-private-key ...

drive9 admin object-backend update --id obb_... --secret-access-key NEW
drive9 admin object-backend get --id obb_...
drive9 admin object-backend ls

drive9 admin tenant object-namespace set --tenant-id tnt_xxx \
  --namespace-id acme-prod \
  --tidbcloud-public-key ... --tidbcloud-private-key ...
```

- Object backends are **org-level** (sibling of `admin tenant` / `admin pool`).
- An org may register **multiple backends** for the same bucket (different `--prefix` and/or `--endpoint`). Mint picks the longest matching prefix; pass `?endpoint=` when two endpoints share a bucket name.
- `update --id` rotates secrets and other fields in place (no delete+add).
- Optional: `--name`, `--sts-endpoint` (custom STS / MinIO STS / CAM / RAM host), `--account-id` (Tencent APPID, Azure account), `--max-session-ttl`.
- The namespace is the **customer's own prefix id**, not the drive9 tenant id. Empty = mint refused.
- Tenant API keys can only mint; they cannot change the mapping. **Scoped (workspace-zone) tokens cannot mint.**
- Secrets are stored encrypted with the existing control-plane KMS. `ls` / `get` never print them.
- Isolation is the minted session, not a `/v1/fs` rewrite. drive9-server does not import the object I/O stack.

### Server mint by scheme

| Scheme | How the server mints | Isolation |
| --- | --- | --- |
| `s3` | AWS `GetFederationToken` (static) or `AssumeRole` | session policy on `bucket/prefix/` |
| `cos` | Tencent CAM `GetFederationToken` or `AssumeRole` | same, via `qcs::cos:…`; `--account-id` or `bucket-appid` |
| `tos` | Volcengine `AssumeRole` (**`--role-arn` required**) | `trn:tos:::bucket/prefix/*` |
| `oss` | Aliyun RAM `AssumeRole` (**`--role-arn` required**) | `acs:oss:*:*:bucket/prefix/*` |
| `gs` | service-account JSON → short-lived OAuth access token | **one GCS bucket per tenant** (namespace = bucket) |
| `az` | storage account key → container SAS | **one Azure container per tenant** (namespace = container) |

TOS and OSS have no AWS-style federation token, so a role ARN is required. Azure SAS and GCS OAuth tokens cannot restrict a key prefix inside a shared container/bucket.

### Schema (brief)

- `org_object_backends`: one row per identity `(scheme, bucket, prefix, endpoint)` via `identity_hash`. Name, STS endpoint, account id, credential kind, encrypted secret/role.
- `tenant_tidbcloud_org_bindings.object_namespace_id`: customer prefix id (empty = not allowed).

Picked up by meta schema self-repair (add columns, drop the old unique on org+scheme+bucket, backfill identity hash).

### E2E

- MinIO: `e2e/object-store-smoke-test.sh` (required `local-e2e` gate). Uses `--auth=local`. When `DRIVE9_BASE` is up, it also copies Drive9 ↔ object.
- **`--auth=server` mint** (manual only, **not** in CI / local-e2e): `e2e/object-auth-smoke-test.sh`. Needs tidbcloud-native plus a dedicated test bucket. Skips if env is unset. Intended for hosted dev.

## Follow-ups (not blocking)

- `stat` does not yet fill ETag/ContentType for object URIs.
- `pack` archive-slot object URIs fail with a local-path error instead of a scheme rejection.

## Test plan

- [x] `go test` for `./pkg/objectfs ./pkg/client ./cmd/drive9/cli ./examples/go-sdk-cookbook ./pkg/meta ./pkg/server`
- [x] `make lint`
- [x] `GOOS=windows CGO_ENABLED=0 go build ./cmd/drive9` (prior commit; Windows stub unchanged)
- [ ] CI `local-e2e` object-store smoke (MinIO + FUSE + `--auth=local`)
- [ ] Manual `e2e/object-auth-smoke-test.sh` on tidbcloud-native dev (after deploy)

## Summary by CodeRabbit

- **New Features**
  - Added browsing, copying, moving, reading, deleting, and mounting for supported object stores.
  - Added local or server-managed authentication with temporary credential refresh.
  - Added administration for object-storage backends and tenant namespaces.
  - Added namespace-scoped access, overwrite controls, progress reporting, and safer recursive copies.
  - Added support for S3-compatible, GCS, Azure, COS, TOS, and OSS storage.

- **Bug Fixes**
  - Improved URI parsing, path handling, timeouts, readiness checks, error reporting, and cross-backend validation.

- **Documentation**
  - Expanded CLI, SDK, authentication, mounting, object-store, and local end-to-end guidance.


qiffang:
## Review: object-store backends for `fs` and `mount` (#845) — early feedback (DRAFT)

Reviewed at `4850de37`. This is a DRAFT, so this is pre-merge feedback, not a gate. Split across correctness (routing / cross-backend copy / write-back-cache durability), adversarial/hang, and impl/e2e lanes.

**Overall:** the architecture is sound — the scheme→backend routing is centralized and safe, and the no-wipe-dirty-cache design is a good call. But there are **several silent data-loss / data-integrity paths, some of which the PR body currently labels as non-blocking follow-ups but shouldn't be.** Recommend addressing the 🔴 items before this leaves DRAFT.

### ✅ Solid

- **Scheme routing chokepoint** — single classifier `Parse` (`fsloc.go:259`), used by all 12 path-taking commands (ls/cat/rm/mkdir/stat/find/grep/chmod/cp/mv/hardlink/symlink); no ad-hoc `HasPrefix("s3://")` bypass anywhere. Case-insensitivity (`S3://`), single-slash `s3:/`→drive9-ctx, Windows `C:/` & `C:\`→local, empty/mid-string `://` are all handled, with golden (`fsloc_test.go`) + negative (`fsbackend_test.go`) tests locking the key invariant: **a registered object URI can never become a local/drive9 path and vice-versa.** No object↔local misroute is possible.
- **Dirty cache is not wiped on umount** — `shutdownObjectVFS` (`mount.go:111-120`) deliberately skips `CleanUp()` and leaves un-uploaded files for a remount to resume. Correct direction.

### 🔴 Data-integrity — should-fix before non-draft

1. **`Release` discards the write-back close error → silent data loss (highest severity).** `mount.go:389` does `_ = h.Close()`. In `CacheModeWrites`, `Close()` is where the deferred upload error surfaces. A failed backend PUT (network/auth/quota) after the last close is never returned — `close()` succeeds from the app's view while the object never landed. `Flush`/`Fsync` *do* propagate (`mount.go:371-385`), and the non-mount FS backend propagates correctly (`fs.go:321-331`) — so this leak is specific to the VFS mount `Release`. Propagate the close error (and consider not collapsing everything to `EIO` in `mapVFSErr`, so quota/auth stay distinguishable).

2. **`WriteOpts.Overwrite` is ignored by every backend and every cp hard-codes `Overwrite:true`, with no `-n`/`--force`.** cp passes `Overwrite:true` at all three sites (`cp.go:587,618,703`); `objectfs.OpenWrite` (`fs.go:171-187`), `localFS.OpenWrite` (`fsbackend.go:400-410`), and drive9 write all overwrite unconditionally, so the plumbed `Overwrite` field is effectively dead. There is no `-n`/`--force`/`--no-clobber` opt-out anywhere. So `drive9 fs cp ./new s3://bucket/existing` silently destroys the existing object with no prompt and no way to decline. This is listed as a non-blocking follow-up but is a genuine silent data-loss path — **must-fix.** In the object context it is worse than on a versioned/soft-delete fs: overwrite is **irreversible destruction of a standalone object**. Even if always-overwrite is inherited from the drive9-client cp UX (worth confirming), the object backend should get an opt-out in this PR — suggest defaulting to no-overwrite on an existing object and requiring `--force`/`-f` to overwrite.

3. **Local destination is truncated in place and deleted on mid-copy failure.** `localFS.OpenWrite` uses `os.Create` (truncates the real dest immediately, `fsbackend.go:405`), and `Abort` `os.Remove`s it (`fsbackend.go:433`). A copy that fails mid-stream over an existing file loses the prior file entirely — there's no temp-then-rename. Write to a temp path and rename on success.

4. **No short-read/size verification → possible silent truncation.** `StreamCopy` (`fscopy.go:64-79`) commits on `io.EOF` without comparing bytes-copied against the declared `size`. A source stream that ends short (truncated read returning clean EOF) can be committed to the object store as a "complete" object. Verify `copied == size` before completing the write.

### 🟠 Correctness / hang — should-fix

5. **Unbounded `context.Background()` object calls → likely the 2h CI-smoke hang.** The `drive9-server-local` e2e smoke was **cancelled at the 2h timeout** (not failed). `Mount()→OpenFs(context.Background())` (`mount.go:29-30`) and several object calls (`fs.go:286/295`, `open.go:24`) use a never-cancelled context with no deadline, so a slow/unreachable store hangs indefinitely. Thread a timeout/cancellable ctx through the object path (and cap the smoke). The umount path itself is *not* the hang — `WaitForWriters` is bounded to 30s.

6. **Recursive object copies have no preflight and leave a half-copied tree silently.** Object-involved recursion uses `cpTreeViaBackend` (`cp.go:667-705`), which lacks the `preflightDestinations` / failure-collector that the drive9 path (`cp_recursive.go`) has. A mid-way failure stops and leaves already-copied children with no manifest/indication.

7. **Symlink handling is inconsistent across backends.** `cp -r local→object` silently follows symlinks (`filepath.Walk` + `os.Open`, `fsbackend.go:332-347`), while `cp -r local→drive9` explicitly rejects them (`cp_recursive.go:69-72`). Pick one behavior.

### 🟡 Durability / resume — should-note

8. **Best-effort resume, not guaranteed.** After the 30s `WaitForWriters` cap, `Shutdown()` proceeds and un-uploaded writes are only recoverable if the *exact same URI* is remounted with the *same cache dir*; otherwise they sit in `~/.cache/drive9/object` unread = effective loss, surfaced only via a stderr warning. On SIGKILL (supervisor `StopTimeout` 60s, `supervisor.go:799/828`) the flush handler never runs. There's no drive9-side dirty marker / WAL and no verification that a remount actually re-uploaded. Also, object mounts are **excluded from `mount drain`** (`mount_drain.go:63-65`) and have no dirty-state visibility in `status`/`health`, so operators can't confirm a clean flush before teardown.

### 🟡 Resource / ergonomics — non-blocking but worth noting

9. **`drive9`-dest copies spool the entire object to a local tempfile** (`fsbackend.go:195-200,269`) before the remote write — an unbounded local-disk requirement (can hit disk-full for large objects), undocumented.
10. **stdin→drive9 buffers the whole file in RAM** (`readAllStdin`, `cp.go:270`) while stdin→object streams — latent O(file) memory inconsistency.
11. **object `mv` (copy-then-delete) is non-atomic**: copy-fail leaves the source intact (correct); delete-fail leaves both source and dest and surfaces an error with no rollback (`fs.go:246-249`). Limited blast radius (same-bucket single-file only).
12. **Metadata (mode/mtime) is dropped silently** on all object/stream copy paths; only same-context drive9 server-side copy preserves it. Object writes stamp `time.Now()` as mtime (`fs.go:177`).
13. Cleanup: `ParseRemote` (`fsloc.go:450`) is dead code with a self-documented "will mis-handle s3://" hazard — recommend deleting. And a typo'd scheme (`s4://…`) fails as "context not found" rather than "unsupported scheme" — misleading but fail-safe.

### 🧪 Test coverage gaps (the durability claims aren't tested)

- `mount_test.go` uses `CacheModeOff` (`:38`), so the **entire write-back / dirty-flush machinery is never exercised** — `shutdownObjectVFS`, `WaitForWriters`, and the "leave dirty cache" behavior have no direct tests.
- No test asserts flush-before-umount, and the `Release` discarded-error path (item 1) is untested — no test simulates a failed backend PUT and checks that `close`/`fsync` surfaces it. These are the load-bearing durability claims and should have regression tests before merge.

### Author's follow-ups — reclassification
The PR lists 4 non-blocking follow-ups. **`WriteOpts.Overwrite` not enforced (item 2)** should be upgraded to must-fix (silent data loss, no opt-out). Pinning `minio/minio:latest` in the required smoke gate is worth doing now (a `latest` in a required gate is a reproducibility footgun). The other two (stat ETag/ContentType, pack archive-slot error type) are genuinely non-blocking.

— Consolidated from dat9-dev1 (routing / cross-backend copy / write-back durability) + adversary-2 (hang / boundary / durability residual). DRAFT: none of this blocks the draft; the 🔴 items should land before it goes ready-for-review.

qiffang:
### Addendum (adversarial lane, source-verified) — hang root cause confirmed + 2 new items

Following up on the consolidated review above with adversary-2's lane, all verified at `4850de37`:

**🟠→confirmed: the 2h CI-smoke hang root cause (was "likely" above, now pinned).** `writeHandle.finish` (`fs.go:328`) does `w.closeErr = <-w.errc` — an **unbounded blocking channel receive** on the `OpenWrite` goroutine's `a.f.Put(ctx, pr, info)` (`fs.go:171-187`), where `ctx = context.Background()` (no timeout). A PUT stalled on the network makes `Close()` block forever. Amplified by two harness gaps: `e2e/object-store-smoke-test.sh` does not wrap the `drive9` CLI calls in `timeout(1)` (its 30s deadline is only in the poll loop, which never runs while a blocked `drive9 fs cat`/mount holds control), and the object-store-smoke step has no per-step `timeout-minutes`, so a wedged call eats the full 120-minute job budget → cancelled. **Fix:** thread a ctx-timeout through the object I/O (or a deadline `select` in `writeHandle.finish`), wrap CLI calls in `timeout`, and add `timeout-minutes` to the step.

**🔴/🟠 new — supply-chain: mount depends on a personal fork of go-fuse.** `go.mod:161`: `replace github.com/hanwen/go-fuse/v2 => github.com/mornyx/go-fuse/v2 v2.9.1-0.20260707042005-213f4c78d30a`. The entire object-mount path runs on a single commit of a **personal** repo (pseudo-version), not upstream. Before this leaves DRAFT: upstream the fork's changes (or vendor them into the org), or at minimum document exactly why the fork is required and pin it deliberately — otherwise this is a supply-chain risk and the build breaks if that personal repo disappears.

**🟠 new — concurrent object mounts have no cross-process lock.** The object cache dir is set process-globally via `config.SetCacheDir`, and rclone's VFS write-back cache is guarded only by an in-process mutex — there is no `flock`/lockfile (grep of `pkg/objectfs` finds none). Two `drive9 mount` processes pointing at the same bucket/prefix (default cache dir, or the same `--cache-dir`) share one write-back directory and can concurrently reload/upload the same dirty files with no mutual exclusion → cache / dirty-data corruption. Since each object mount is its own process, this is realistically reachable. Add a per-cache-dir lockfile (or namespace the cache dir per mount).

**Refinement to must-fix ①:** `Flush` (`mount.go:376`) *does* propagate errors via `mapVFSErr(h.Flush())`, so an app that `fsync`s before close is covered. The gap is specifically: writes are async under `CacheModeWrites`, and (a) if `Flush` does not block until the async writeback PUT actually lands, and (b) `Release` (`mount.go:389`) then swallows the deferred `Close()` error, a write-then-close (no fsync) sees success while the PUT failed. So ① should verify whether `Flush` waits for the writeback PUT; the `Release` swallow is the second line of the same leak.

**Non-blocking boundary notes:** credentials are clean (`env_auth=true`, no embedded AK/SK, no creds in logs) but a bad-credential mount "fake-succeeds" (lazy `NewFs`) then every FUSE op returns EIO — a UX boundary, not a leak. Windows stub is clean fail-closed. `AWS_*_CHECKSUM` is mutated process-globally in `ensureRclone` (`open.go:26`) and by the code's own comment also affects later `pkg/s3client` uploads — a cross-package side-effect worth noting.

— adversary-2 (adversarial/hang/supply-chain/boundary), verified and folded in by dat9-dev1.

qiffang:
One additional required-gate gap from the objectfs/E2E lane:

**The required MinIO smoke does not exercise Drive9 ↔ object-store copy at all.** `.github/workflows/local-e2e.yml:278-283` runs `object-store-smoke-test.sh` with only `OBJECT_STRICT_MOUNT`; the job has `DRIVE9_BASE`, but this step does not provide `DRIVE9_API_KEY`. The script gates the cross-backend section on both variables at `e2e/object-store-smoke-test.sh:545-558`, so exact-head CI always records that feature as `SKIP` while still passing the required suite. The optional block also tests only Drive9 → object, not the PR's advertised object → Drive9 example (`drive9 fs cp s3://bucket/a :/dst`). `cli-smoke-test.sh` has another optional Drive9 → object check, but `DRIVE9_E2E_S3_URI` is not set in the required workflow either.

Before moving the PR out of draft, make the required local-e2e step pass `DRIVE9_API_KEY=local-dev-key` and assert both directions with exact content/checksum plus cleanup. In CI mode, skipping this central cross-backend contract should fail the object-store gate rather than count as a successful suite.

Focused `go test`/`go vet` for `./pkg/objectfs` and `./cmd/drive9/cli`, plus diff-check, pass on exact head `4850de37`.

mashenjun:
@codex review

qiffang (changes requested):
REQUEST_CHANGES on exact head `91043c477eaf3298944e5f37e7fa399cce277bc1`.

I found blocking issues in the object-store path. The URI parsing and the `--auth=server` no-local-fallback path look clean, and the old overwrite / local-temp / short-copy / required E2E blockers are closed, but these still need changes before merge.

1. **Server-minted S3-family writes/copies can fail before the object write because rclone still checks the bucket.**

   `pkg/server/object_backend.go:621-630` mints a prefix-scoped policy: object actions are only on `arn:aws:s3:::bucket/prefix/*`, and the bucket-level `ListBucket` grant is conditioned on `s3:prefix`. That is the right isolation intent. But `pkg/objectfs/open.go:166-195` builds the rclone S3/COS/TOS/OSS connection without `no_check_bucket=true`. In rclone v1.69.3, `Object.prepareUpload` calls `mkdirParent`, which calls `Mkdir`, which calls `makeBucket`; unless `NoCheckBucket` is set, `makeBucket` performs a bucket existence path (`HeadBucket`/bucket create cache). Server-side copy does the same through `Copy -> mkdirParent`.

   A prefix-scoped tenant session should not have unconditioned bucket-level authority, so this bucket check is outside the intended scoped object operation. The default `--auth=server` write/copy path can fail before it reaches `PutObject`/`CopyObject`, while the required CI object-store smoke mostly uses `--auth=local` and will not catch this. The manual `object-auth-smoke-test.sh` is the path that would hit it, but it is not a required gate.

   Fix gate: set `no_check_bucket=true` for server-minted S3-family rclone remotes, or otherwise prove the minted policy covers rclone's bucket check without widening tenant access. Add a regression with prefix-scoped credentials or a fake S3 endpoint that rejects bucket checks but allows the prefixed object operation, and prove `fs cp` plus same-backend copy succeed without escaping the prefix.

2. **Admin backend validation accepts rows that can never mint credentials.**

   On create, `handleAdminObjectBackendCreate` validates the row before secret handling (`pkg/server/object_backend.go:203`) and the `credential_kind=role` branch only requires `role_arn` (`:211-216`). `validateAdminObjectBackend` only requires `role_arn` for TOS/OSS (`pkg/server/object_sts.go:641-644`). That means `scheme=tos|oss|cos`, `credential_kind=role`, `role_arn` set, and no source access key/secret is accepted and persisted with `has_secret=false`.

   The mint path then deterministically fails: COS requires access key + secret (`pkg/server/object_sts.go:181-183`), OSS requires access key + secret for AssumeRole (`:252-258`), and TOS requires access key + secret (`pkg/server/object_sts_volc.go:23-29`). Patch has the same class of bug: validation runs before applying `secret_access_key`, and `secret_access_key:""` clears `row.SecretCipher` after validation (`pkg/server/object_backend.go:343-350`), so an existing valid backend can be patched into an unmintable row.

   Fix gate: validate the final persisted row after secret/external-id patches against the provider-specific mint requirements, including whether a secret is actually present. Add create and update regressions for TOS/OSS/COS role configs without source credentials and for PATCH clearing a required secret. Keep AWS role without stored source credentials accepted only if the intended contract is server-IAM fallback.

3. **Writable object mounts acknowledge local close/fsync before remote upload success is knowable.**

   Object mounts force rclone `CacheModeWrites` (`pkg/objectfs/mount.go:49-50`). In rclone v1.69.3 the default `vfs_write_back` is 5s, `RWFileHandle.Flush` only updates size, `Sync` syncs the local cache item, and `Item.Close` queues dirty data for asynchronous upload unless writeback is <=0. This wrapper maps `Flush`/`Fsync` to FUSE success (`pkg/objectfs/mount.go:414-431`), and `Release` can only log a close error (`:433-438`) because FUSE release has no return status for later background upload failure.

   So an app can write, `fsync`, close, and see success while the later remote PUT fails due auth expiry, quota, network, or provider error. The dirty cache may remain for remount, but that is recovery, not the durability contract that a successful close/fsync usually signals. Current tests do not cover production semantics: `pkg/objectfs/mount_test.go:20-39` builds the VFS with `CacheModeOff`, so failed remote PUT after close is not discriminated.

   Fix gate: either make the default writable mount mode close-sync/write-sync enough to return upload failure to the application, or explicitly narrow the product contract with visible pending/error state and user documentation, plus a regression for failed remote PUT in the actual configured mount mode.

4. **Object operations still have unbounded product I/O contexts.**

   The 30s timeout in `openObjectAuthenticated` only covers the initial open/mint (`cmd/drive9/cli/fsbackend.go:506-520`). The actual product operations then use `context.Background()`: object copy goes through `cpViaBackend(context.Background(), ...)` (`cmd/drive9/cli/cp.go:167,181`), object list uses `Backend.List(context.Background(), ...)` (`cmd/drive9/cli/ls.go:42,47`), and object writes pass that unbounded context into `a.f.Put` (`pkg/objectfs/fs.go:196-205`). Mount initial mint/open also uses `context.Background()` (`cmd/drive9/cli/mount_object.go:32-35`, `pkg/objectfs/mount.go:33-39`).

   A hung provider TCP path can therefore hang the CLI/mount operation indefinitely. The E2E script's outer timeout protects CI only; it is not a product deadline. Add operation-level deadlines/cancellation for object list/read/write/copy/mount mint/open and regressions with a backend that never responds.

Non-blocking but should fix:

- `git diff --check origin/main...HEAD` fails on `docs/design/object-store-cli.md:3-4` due trailing whitespace.
- Add the missing `--auth=server` mint-failure regression proving the client fails closed and does not fall back to local environment credentials.
- Avoid logging raw Azure SAS parse errors in `pkg/objectfs/refresh_creds.go:88-91`; `url.Parse` errors can include the full input including `sig=`.
- `cmd/drive9/cli/fsbackend.go:450-486` uses `os.CreateTemp` then renames without chmod, so object-to-local output is always `0600` and overwrites lose the previous mode. Define and test local output permissions.

Local validation I ran on this head: `git diff --check origin/main...HEAD` (failed only for the docs whitespace above), touched-Go `gofmt -l` (clean), `go test -c ./pkg/objectfs ./cmd/drive9/cli ./pkg/client ./pkg/server`, `go vet ./pkg/objectfs ./cmd/drive9/cli ./pkg/client ./pkg/server`, and focused `go test ./pkg/objectfs ./cmd/drive9/cli -run 'Test(Object|Open|Parse|Session|Apply|Rewrite|Mount|Peel|Ls|Cp|FsHandle|StreamCopy|Copy)' -count=1` passed. Focused `pkg/server` runtime tests are still blocked locally by rootless Docker/TestMain; hosted Code CI, Local E2E, and CodeRabbit are green on the exact head.

mornyx:
@qiffang Head is now `84edfea1`. The requested blockers (and the Codex items on the same paths) are addressed as follows.

### Blockers

1. **Prefix-scoped STS vs rclone bucket check**  
   Server-minted S3/COS/TOS/OSS remotes now set `no_check_bucket=true`, so `Mkdir`/`Copy` do not `HeadBucket` before `PutObject`/`CopyObject`. Local `--auth=local` is unchanged.

2. **Admin rows that can never mint**  
   `validateAdminObjectBackend` runs on the **final** row after secret/external-id create/patch. COS/TOS/OSS require stored AK+SK (role included). AWS `credential_kind=role` without stored keys is still allowed (server IAM). PATCH that clears a required secret is 400. Namespace/prefix `*`/`?` are rejected.

3. **Writable mount close reported success before PUT**  
   Object mounts use `WriteBack=0` and FUSE `Flush` waits on rclone `Close`, which does a synchronous upload. A successful close means the PUT finished. Remaining dirty cache on umount is still not wiped.

4. **Hung provider I/O**  
   Object `ls`/`cat`/`rm`/`mkdir`/`stat`/`mv` use a 5-minute deadline. Object copies use 30 minutes. Mount mint is 30 seconds. Drive9-native commands are not wrapped in that deadline.

### Other items from the same review

- Docs trailing whitespace: removed.  
- `--auth=server` mint failure does not fall back to AWS env: `TestLsObjectURIDoesNotFallBackToLocalAWSEnv`.  
- Azure SAS parse failures are logged without the URL/`sig=`.  
- Object→local overwrite keeps the previous file mode (else `0644`).  
- Mint JSON body is capped at 1MiB.  
- Server-injected endpoints go through the same http/https rules as URI `?endpoint=`; scheme-less values are stored as `https://…`.  
- GCS refresh tokens are per filesystem; Azure SAS rewrite is per host+container.  
- `mv` onto a directory prefix is rejected. Empty `stat …/` of a missing prefix is not a directory.  
- `--allow-other` sets `default_permissions`.  
- `RUN_API_ONLY=1` no longer starts MinIO/FUSE by default.

Not changing in this pass: FUSE dup-fd behavior on the same handle (documented as unsupported), and per-file copy deadlines for very long `cp -r` trees.

qiffang (changes requested):
adversary-1 exact-head re-review @ `0152e4dece17b0a579112644338f62e065e61279`: REQUEST_CHANGES.

Most of my earlier blocker set is closed on this head: server-minted S3-family rclone remotes now use `no_check_bucket=true`; admin object-backend validation rejects unmintable configs before persistence; CLI/list/copy/write object commands now have bounded deadlines; mint failure stays fail-closed and has a discriminating no-local-fallback regression; Azure SAS parse errors no longer log the raw `sig=` URL; and the required object-store smoke ran in hosted Local E2E with Drive9->object and object->Drive9 exact-content checks.

Two mount-specific blockers remain.

1. **Writable object mounts still cannot report remote PUT failure through the normal write durability surface.**

`pkg/objectfs/mount.go:49-54` keeps rclone VFS in `CacheModeWrites` with `WriteBack=0`. That removes delayed queueing, but it does not make `flush`/`fsync`/`close` report the remote object-store result. Drive9 maps FUSE `Flush` and `Fsync` to `h.Flush()` and `h.Sync()` (`pkg/objectfs/mount.go:421-437`), and rclone v1.69.3's RW handle only updates size / syncs the local cache metadata there (`vfs/read_write.go:192-198`, `:416-428`; `vfscache/item.go:1399-1416`). The remote store happens later from `Item.Close` when `WriteBack <= 0` (`vfscache/item.go:729-735`), reached from `Release`; rclone explicitly notes Release errors are ignored by the kernel (`vfs/read_write.go:200-216`), and Drive9's `Release` can only log `object mount release close failed` (`pkg/objectfs/mount.go:440-445`).

So an application can get successful `fsync`/`close` semantics from the mount path while the remote PUT later fails and is only visible in logs. Existing mount tests cover that Flush must leave the handle writable (`TestRcloneFUSECreateFlushThenWrite`) and happy-path smoke; they do not inject a remote PUT failure and assert that the writer sees an error. For a writable filesystem mount this is still a correctness blocker, even though the old async write-back queue part was reduced.

2. **Object mount I/O is still not bounded by per-operation deadlines/cancellation.**

The new deadline helpers correctly wrap CLI object commands (`fsbackend.go:520-540`, `cp.go:553-558`, etc.), but mounted object I/O still creates a mount-lifetime background context (`pkg/objectfs/mount.go:32-38`) and the FUSE methods ignore go-fuse's `cancel <-chan struct{}` while calling contextless rclone VFS operations (`Open/Create/Read/Write/Flush/Fsync/Mkdir/Remove/Rename` at `pkg/objectfs/mount.go:353-538`). rclone's provider transport idle/connect timeouts are not an absolute per-FUSE-operation deadline; a slow-but-progressing provider can still pin a user syscall indefinitely. My old context-hang blocker included `mount`, so I cannot close it based only on CLI command deadlines.

Fix gate: mounted object writes need a contract where remote durability failure is observable to the writer through a syscall/command the user can act on, or the mount must be explicitly downgraded to a documented best-effort cache with tests and UX that do not imply close/fsync durability. Separately, mounted object operations need per-operation cancellation/deadline coverage or a proof plus regression that each object-store FUSE operation is bounded under a hung/slow backend.

Validation performed: exact-head hosted Code CI and Local E2E are green, and Local E2E includes the object-store smoke with 56 passed / 0 failed / 0 skipped plus Drive9↔object copy checks. Locally: `git diff --check`, touched-Go gofmt, compile checks, `go vet ./pkg/objectfs ./cmd/drive9/cli ./pkg/server ./pkg/client`, focused `pkg/objectfs` tests, focused CLI object tests, and the no-local-fallback/mode-preservation tests passed. Runtime `pkg/server` tests are still blocked on this Mac by rootless-Docker TestMain, so I used hosted E2E for that surface.

mornyx:
@qiffang Head is `3a12a26e`. The two remaining mount blockers:

1. **PUT failure on close/fsync**  
   Write-handle `Flush`/`Fsync` now `Close` the rclone file (`WriteBack=0` → synchronous PUT) and immediately reopen without `O_TRUNC`, so the kernel fd stays usable after the post-Create FLUSH (`printf > file`). The Close error is returned to that syscall. `TestRcloneFUSECreateFlushThenWrite` covers Create→Flush→Write; `TestRcloneFUSEFlushReportsPutFailure` injects a failing `Put` and asserts Flush is not OK.

2. **Mounted I/O deadlines**  
   Open/Create/Read/Write/Flush/Fsync/Release, Lookup/GetAttr/ReadDir, Mkdir/Unlink/Rename race go-fuse's `cancel` channel against a 5-minute timeout (`TestRcloneFUSEOpCancel`). Timeout/cancel maps to `ETIMEDOUT`.

qiffang (changes requested):
Blocking finding on exact head `3a12a26eb1dca80c3789e41e74ec4bbc3b722e6e`:

`SetAttr` still bypasses the new FUSE timeout/cancel wrapper. The latest change wraps Lookup/GetAttr/Open/Create/Read/Write/Flush/Fsync/Mkdir/Unlink/Rename with `withOp`, and the new PUT-failure path for Flush is good, but `pkg/objectfs/mount.go:522-550` still calls `fs.v.Stat`, `file.Truncate`, and `node.SetModTime` directly. Those are normal object mount operations: path truncate sends `FATTR_SIZE`, and `touch`/utimens sends mtime. In rclone v1.69.3, `File.SetModTime` ultimately calls the remote object with `context.TODO()` (`vfs/file.go:500`), and `File.Truncate` can also wait through the VFS object/writer path. So a stalled provider can still leave these mount syscalls unbounded and ignoring the kernel cancel channel.

This means the prior "object mount I/O has no per-operation deadline/cancellation" blocker is not fully closed. The new `TestRcloneFUSEOpCancel` only proves the `withOp` helper returns when cancelled; it does not exercise `SetAttr`, so this gap would stay green.

Fix gate: route `SetAttr` through the same bounded/cancel-aware path and add a discriminating regression for at least `FATTR_SIZE` or mtime where the underlying VFS/backend blocks or fails and the mount returns the expected timeout/cancel/error instead of hanging. Once that is covered, the remaining mount blockers look closeable: Flush/Fsync now surface PUT failures via close+reopen, and the previously fixed mint/no-check/validation/CLI-deadline paths were not regressed by this latest commit.

Local validation on this head: `git diff --check`, touched Go gofmt, `go test -c ./pkg/objectfs ./cmd/drive9/cli ./pkg/server ./pkg/client`, `go vet ./pkg/objectfs ./cmd/drive9/cli ./pkg/server ./pkg/client`, `go test ./pkg/objectfs -count=1`, and focused CLI object tests passed.

mornyx:
@qiffang Head is `f27aef71`. `SetAttr` is now on the same `withOp` path as the other FUSE methods.

Truncate (`FATTR_SIZE`) and mtime go through the 5-minute deadline and the kernel cancel channel; mode/uid/gid stay `ENOSYS`. `withOp` also fails closed if cancel is already signaled before the VFS call starts.

Tests: `TestRcloneFUSESetAttrSize` (truncate is visible) and `TestRcloneFUSESetAttrCancel` (pre-cancelled `SetAttr` returns `ETIMEDOUT` instead of hanging or succeeding).

qiffang:
Supplementing my current REQUEST_CHANGES on exact head `3a12a26e`: the FUSE deadline/cancel blocker is broader than the `SetAttr` miss. Please do not treat "wrap `SetAttr` in `withOp`" alone as sufficient.

The current `withOp` is an early-return wrapper, not actual cancellation or a safe detach boundary:

- `pkg/objectfs/mount.go:728-741` starts `fn(ctx)` in a goroutine and returns on timeout/cancel, but the rclone VFS closures in this file do not use `ctx`. The remote operation can continue after the FUSE syscall has already returned, so delayed PUT/rename/delete side effects and goroutine leaks remain possible.
- go-fuse's `RawFileSystem` contract allows concurrent method calls and says request/output memory may be reused after the API call returns; background goroutines must copy request data. Current `Read`/`Write` closures capture `input.Offset`, `buf`, and `data`, so returning from `withOp` while the goroutine continues can race with reused kernel request buffers. Cancel is also mapped to `ETIMEDOUT` instead of the interrupt semantics go-fuse documents for preserving outstanding request data.
- `fuseHandle` has no per-handle serialization. `getHandle` only protects the map lookup, then `Read`/`Write`/`Flush` access `ent.h` directly; `Flush` closes/reopens and `replaceHandle` mutates `ent.h`, and `Rename` mutates `ent.path` while `Flush` reads it without holding the same lock. Concurrent `Flush+Write`, duplicate `Flush`, and `Rename+Flush` can race.
- The `SetAttr` path I called out is one missing syscall, but there are siblings: `ReadDir`/`ReadDirPlus` call remote `ReadDirAll()` outside `withOp`, and `Create` does a post-create `Stat` outside the bounded path.

Fix gate: the FUSE path needs a real safe deadline/cancel contract, per-handle serialization or equivalent, and an exhaustive pass over every FUSE method/remote call. Please add discriminating tests that would fail on the current code, including at least `FATTR_SIZE` or mtime `SetAttr`, a directory read or create-post-stat timeout/cancel case, and concurrent handle cases such as `Flush+Write`, duplicate `Flush`, and `Rename+Flush`. The existing `TestRcloneFUSEOpCancel` only proves the helper returns early; it does not prove the underlying operation stopped safely or that request-buffer lifetime is respected.

The synchronous Flush/Fsync PUT-error reporting remains closed by the current head; this comment is only about the remaining deadline/cancel/concurrency gap.

mornyx:
@qiffang Head is `2232b8c4`. The remaining FUSE deadline/cancel/concurrency gap is addressed as follows. This does **not** abort in-flight rclone/object HTTP.

### Contract

- Kernel cancel → `EINTR` immediately (`withOp` does not wait for `fn`).
- 5-minute deadline → `ETIMEDOUT` (if `fn` has not already returned).
- rclone VFS still does not cancel in-flight PUT/GET/rename/delete. The FUSE reply returns; the remote op may continue.

### Request-buffer lifetime

- `Read` copies into a private buffer; `Write` clones payload; `SetAttr`/`Mkdir`/`ReadDir` snapshot input fields before `withOp`.
- Background `fn` must not touch go-fuse request memory after we return.

### Per-handle serialization

- Each `fuseHandle` has a mutex. `Read`/`Write`/`Flush`/`Release` take it around `ent.h`.
- `Rename` locks matching handles (id order) around `vfs.Rename` and the path update, so `Flush+Write`, duplicate `Flush`, and `Rename+Flush` cannot race `ent.h` / `ent.path`.

### Remaining remote calls

- `ReadDir`/`ReadDirPlus`: `Stat` + `ReadDirAll` are inside `withOp`.
- `Create`: post-create `Stat` is inside the same `withOp` as `Create`.
- `SetAttr` was already on this path in `f27aef71`.

### Tests

- cancel → `EINTR` (including `SetAttr`, `ReadDir`, `Create`, and a start-then-cancel case that must return without waiting on `fn`)
- `mapVFSErr`: `Canceled`→`EINTR`, `DeadlineExceeded`→`ETIMEDOUT`
- concurrent duplicate `Flush`, `Flush+Write`, `Rename+Flush`

qiffang (approved):
APPROVE on exact head `2232b8c428784df0b1f1012440fe2e6cc4b6c61c`.

I re-reviewed the latest mount delta against the three previous FUSE gates. The prior blockers are closed:

1. Request-buffer lifetime / fake-cancel data corruption is closed. `withOp` still cannot abort rclone's in-flight HTTP, but `Read` now reads into a private `dst`, `Write` copies `data` into `payload`, and scalar request fields are copied before the goroutine uses them. Detached rclone work no longer touches go-fuse request-owned buffers after the FUSE method returns.
2. Per-handle concurrency is closed. `fuseHandle` now has a mutex; `withHandle`, `Flush`/`Fsync`, `Release`, and rename path updates serialize handle access. The new duplicate-flush, flush/write, and rename/flush tests cover the races, and local `-race` is clean.
3. FUSE method coverage is closed. I enumerated the implemented methods and remote calls: `Lookup`, `GetAttr`, `ReadDir`/`ReadDirPlus`, `Open`, `Create` including post-create stat, `Read`, `Write`, `Flush`/`Fsync`, `Release`, `SetAttr`, `Mkdir`, `Unlink`/`Rmdir`, and `Rename` now go through `withOp`/`withHandle`. `Access`, `OpenDir`, `StatFs`, and `Forget` have no remote object-store call. The new `SetAttrCancel`, `ReadDirCancel`, and `CreateCancel` tests cover the earlier missed paths.

The old PUT-error surfacing fix remains good: `WriteBack=0` plus close/reopen on write-handle `Flush`/`Fsync`, with `TestRcloneFUSEFlushReportsPutFailure` still discriminating. Other old gates are unchanged by this delta because only `pkg/objectfs/mount.go` and `pkg/objectfs/mount_test.go` changed.

Non-blocking note: timeout/cancel still means the local syscall is bounded, not that the underlying object-store mutation was hard-aborted. A delayed PUT/rename/delete may still commit after the caller saw `ETIMEDOUT`/`EINTR`, because rclone VFS cannot abort in-flight HTTP. I am not blocking on this after the buffer/race issues were fixed, but the user-facing design docs should explicitly state this "outcome may still commit" semantics unless the product wants a future generation/fence mechanism for hard-abort behavior.

Validation: hosted Code CI, Local E2E, and CodeRabbit are green on this exact head. Local checks passed: `git diff --check`, touched-file `gofmt`, `go test -c ./pkg/objectfs`, `go vet ./pkg/objectfs`, `go test ./pkg/objectfs -count=1`, `go test -race ./pkg/objectfs -count=1`, `go test -race ./pkg/objectfs -run TestRcloneFUSE -count=20`, and focused CLI object tests.

qiffang (changes requested):
Superseding my earlier approval on exact head `2232b8c428784df0b1f1012440fe2e6cc4b6c61c`: **REQUEST_CHANGES**.

The previous request-buffer corruption, SetAttr/method coverage, and basic per-handle serialization fixes are real, but two correctness gaps remain.

1. **Mutating FUSE operations can still commit after the caller has already received `EINTR` / `ETIMEDOUT`.**

`withOp` returns immediately on cancel or the 5-minute deadline while the callback keeps running (`pkg/objectfs/mount.go:808-835`). The mutating callbacks for `Flush`/PUT (`:501-538`), `SetAttr` (`:575-593`), `Mkdir` (`:612-618`), `Unlink`/delete (`:634-635`), and `Rename` (`:657-675`) do not observe `ctx` and have no generation/fence/settlement guard. That means an application can get a failed syscall, retry or recreate the path, and then have the old delayed PUT/delete/rename/truncate land later against the object store.

This is no longer the go-fuse request-buffer race, because Read/Write now copy buffers. It is still a hidden data semantics bug: the user-visible result says the operation failed/timed out, but the remote mutation may later succeed. `TestRcloneFUSEOpCancelDoesNotWait` proves the early-return behavior, but there is no discriminating test for a mid-flight canceled mutation proving that remote state does not change after the error is returned.

Fix gate: either make mutating operations wait/settle before returning a definitive failure, or add a generation/fence/settlement mechanism that prevents a delayed mutation from applying after failure has been reported. If the intended product contract is "timeout/cancel means outcome unknown and the mutation may still commit", that must be an explicit user-facing contract, not an implicit side effect of `withOp`.

2. **Rename does not fence concurrent handle publication, so a handle can be stored with the old path after the rename snapshot.**

`Open` completes the rclone open inside `withOp` (`pkg/objectfs/mount.go:388-403`) and only publishes the handle later at `storeHandle` (`:406`). `Create` has the same shape (`:422-448`). `Rename` snapshots and locks only the handles present at `lockHandlesForPath(oldPath)` (`:658`), then rewrites only that snapshot (`:669-673`). Because go-fuse runs methods concurrently, an `Open`/`Create` can finish the rclone call before/during `Rename`, miss the rename snapshot, and then publish a new `fuseHandle` whose `ent.path` is still the old path. Later `Flush` reopens by `fs.v.Stat(ent.path)` (`:521`), so that stale handle can flush/reopen against the wrong path or fail after a successful rename.

Fix gate: make handle publication and rename's snapshot/path rewrite share a single atomic fence, then add a forced-interleaving regression where `Open`/`Create` obtains the rclone handle, `Rename` completes, and the handle is published afterward; subsequent `Flush` must use the new path.

Related test weakness: `TestRcloneFUSEConcurrentFlushAndWrite` only checks `HasPrefix("hello")`, so it allows the concurrent write to disappear; `TestRcloneFUSERenameDuringFlush` accepts either `after.txt` or `before.txt` after a successful rename. These should be tightened to the exact expected content/path and use barriers to force the intended overlap, otherwise the tests can pass without proving the race is closed.

Validation on this exact head: hosted Code CI and Local E2E are green. Local `git diff --check`, gofmt, `go test -c ./pkg/objectfs`, `go vet ./pkg/objectfs`, `go test ./pkg/objectfs -count=1`, `go test -race ./pkg/objectfs -count=1`, `go test -race ./pkg/objectfs -run TestRcloneFUSE -count=20`, and focused CLI object tests pass. These checks do not cover the two interleavings above.

mornyx:
@qiffang Head is `fe84c213`. The two remaining FUSE correctness gaps:

1. **Mutating ops no longer fail-then-commit.**  
   Reads still return `EINTR` immediately (request buffers are copied). Write/Flush/Create/SetAttr/Mkdir/Unlink/Rename wait until the rclone call finishes. If the mutation succeeded, the syscall returns OK even if the kernel cancel/deadline fired while it was in flight. If it did not start (`cancel` already closed) we still return `EINTR` and do not create the file. rclone still cannot abort in-flight HTTP; a hung provider can stall a mutating syscall until the provider unblocks or the 5-minute op ctx is observed.

2. **Handle publication is fenced with Rename.**  
   `Open`/`Create` publish the handle under `publishMu`; `Rename` holds the same mutex while snapshotting handles and rewriting `ent.path` / `inodePath`. A handle whose remote open finished before rename but was published after it stores the post-rename path, so later Flush `Stat`s the new name.

Tests: mutating op does not return before settlement and reports OK when the mkdir landed; Open-then-Rename and Create-then-Rename forced interleavings; `Flush+Write` expects `hello!`; `RenameDuringFlush` requires `after.txt` only.

qiffang (approved):
Re-reviewed exact head `fe84c213533f222cb342ccecc9ef3bbfd2b17f64`: **APPROVE** from my code-review lane.

The two blockers from my previous `REQUEST_CHANGES` are closed:

1. **Late mutation after a returned failure is closed by settle-wait semantics.** Mutating FUSE paths now go through `withMutatingOp`, which cancels the context on kernel cancel/deadline but then waits for the operation goroutine to finish before returning. That means the caller no longer gets `EINTR` / `ETIMEDOUT` while a PUT/delete/rename/truncate can still commit later. I re-enumerated the mutating paths (`Write`, `Flush`/`Fsync`, `Release`, `SetAttr`, `Mkdir`, `Unlink`/`Rmdir`, `Rename`, plus `Open`/`Create` handle publication) and did not find a remaining mutating path using the fast-return `withOp` behavior.

2. **`Rename` vs `Open`/`Create` handle publication is now fenced.** `publishMu` makes rename's snapshot/path rewrite mutually exclusive with handle publication, and `publishHandle` re-reads the current inode path while holding that fence. A handle either publishes before rename and is included in the rewrite, or publishes after rename and stores the renamed path.

The regression coverage is also materially stronger: `TestRcloneFUSEMutatingOpWaitsAndReportsSuccess` proves mid-flight cancel waits for settlement; `TestRcloneFUSEOpenThenRenamePublishesNewPath` and `TestRcloneFUSECreateThenRenamePublishesNewPath` force the previous publication race; `ConcurrentFlushAndWrite` now requires exact `hello!`; `RenameDuringFlush` now requires `before.txt` to be gone after a successful rename.

Validation on this head: `git diff --check`, gofmt check, `go test -c ./pkg/objectfs`, `go vet ./pkg/objectfs`, `go test ./pkg/objectfs -count=1`, `go test -race ./pkg/objectfs -count=1`, `go test -race ./pkg/objectfs -run 'TestRcloneFUSE(MutatingOp|OpenThenRename|CreateThenRename|ConcurrentFlushAndWrite|RenameDuringFlush)' -count=20`, and focused CLI object tests all pass locally. Hosted Code CI is green; Local E2E is still running as I submit this review.

Non-blocking caveat: the chosen mutation contract is correctness-over-fast-cancel. If an underlying rclone/provider call ignores context and never returns, the mutating syscall can remain blocked until that call returns; the important part is that it no longer reports failure before a delayed mutation commits. Please avoid describing mutating operations as having a hard 5-minute return guarantee unless the underlying call path actually observes the deadline.

qiffang (changes requested):
Superseding my approval on exact head `fe84c213533f222cb342ccecc9ef3bbfd2b17f64`: **REQUEST_CHANGES**.

The previous two data-safety blockers are genuinely closed: mutating operations now settle before returning, and `publishMu` fences `Rename` against `Open`/`Create` handle publication. However, the chosen settle implementation reopens the FUSE liveness blocker and conflicts with the current design contract.

`runOp` creates the 5-minute context deadline at `pkg/objectfs/mount.go:844`, but for mutating operations `waitSettled` handles cancel/deadline by calling `stop()` and then unconditionally blocking on `<-done` (`:869-878`). The rclone mutation callbacks (`Write`, `Flush`/PUT, `Release`, `SetAttr`, `Mkdir`, `Unlink`, `Rename`) do not themselves abort when that context is canceled. If the provider or rclone VFS call hangs while ignoring context, the FUSE syscall can still wait forever; the 5-minute deadline does not unblock it.

This is also inconsistent with the updated design doc. `docs/design/object-store-cli.md:197-202` says mutating ops wait for rclone and that a hung provider can stall until “the provider or the 5-minute op deadline unblocks.” The current code only unblocks on provider/callback completion. The deadline only cancels the context; it does not release `waitSettled` unless the callback observes that cancellation and returns.

Fix gate: either make the underlying mutation path truly abort/return by the operation deadline, or allow deadline/cancel to return only with a generation/fence/settlement mechanism that prevents an old mutation from committing after failure is reported. If the intended product contract is instead “mutating syscalls may block indefinitely when the provider/rclone call never returns,” that needs an explicit product decision and the docs/tests must say that; absent that decision, this remains a blocker.

Please also add a discriminating liveness test: a mutating operation whose callback never completes must not hang past the configured/controlled deadline if the contract claims deadline-bounded behavior. The current tests prove settle-before-return for a delayed callback that eventually unblocks, but they do not cover a callback that never returns.

Local validation from my previous review still applies for the closed race fixes (`git diff --check`, gofmt, objectfs tests, race tests, focused FUSE interleaving tests, focused CLI object tests). Those checks do not prove the deadline-bounded liveness contract above.

mornyx:
@qiffang Head is `4de6e11e`. This is a product choice, not a missed abort.

rclone VFS does not abort in-flight object HTTP, so we will not return a mutating FUSE syscall at the 5-minute mark while the callback is still running. That would recreate fail-then-commit. A generation/fence in this layer also cannot un-commit a PUT/delete/rename that already left the machine.

Contract:

- **Reads:** kernel cancel → `EINTR` immediately; 5-minute deadline → `ETIMEDOUT`. The read callback may keep running (buffers are copied).
- **Mutating ops:** wait until rclone returns. Success is OK even if cancel/deadline fired during the call. If the provider/VFS ignores ctx, the syscall stays blocked until that call returns. The 5-minute context is only a cancel hint.

`docs/design/object-store-cli.md` no longer says the deadline unblocks a hung mutating op. `TestRcloneFUSEMutatingOpDeadlineDoesNotUnblockIgnoredCtx` locks that: after `fuseOpTimeout`, a callback that ignores ctx still holds `withMutatingOp` until it actually returns.

qiffang (approved):
Re-reviewed exact head 4de6e11e3510387b62bb2218a38d8f4ad13caee9. The remaining object-store mount liveness tradeoff is now explicit and accepted: this affects object-store mounts only, and the product choice is correctness over bounded return for mutating object mount syscalls when the provider/rclone call never returns. Prior data-safety blockers remain closed, the false contract is fixed by docs/tests, and required checks are green. APPROVE.

PR #849, opened 2026-08-20
refactor: replace drive9-server-local with a tidb local provisioner

mornyx:
## Summary

This change collapses local development, unit tests, and e2e onto the **same production `drive9-server` path and the same app-managed TiDB tenant schema**. Previously, local/dev/test used a separate single-tenant binary (`drive9-server-local`) and a MySQL no-embedding dialect (`mysql_no_embedding`), so bugs that only existed on the production TiDB schema (VECTOR columns, FTS, provision JWT, tenant pool, schema ensure) were easy to miss.

### One schema for tests and production

- Deleted `pkg/tenant/schema/mysql_no_embedding.go` (LONGTEXT stand-ins for `embedding` / `description_embedding`, no VECTOR/FTS).
- Unit-test helpers now apply the **app-managed TiDB schema** (`VECTOR(n)` writable columns) plus the legacy `files` table only when a test needs dual-write coverage.
- Removed `DRIVE9_TEST_MYSQL_DSN`. Tests use `DRIVE9_TEST_TIDB_DSN` or start `pingcap/tidb` via testcontainers (`internal/testtidb`, formerly `internal/testmysql`).
- Dropped `testcontainers-go/modules/mysql`; `go.mod` now depends on `testcontainers-go` core, with old `google.golang.org/genproto` versions excluded so `go mod tidy` is clean.
- Live stacks that can use TiDB (local e2e, fuse-patch, SDK integration, blackbox) now use TiDB + `DRIVE9_LOCAL_EMBEDDING_MODE=app`. Semantic/SQL/tenant-fork cases stay **off by default** and must be opted in.

This is the main correctness win: a test that writes embeddings, runs grep, or inits tenant DDL is exercising the schema production tenants actually get.

### `drive9-server-local` → `DRIVE9_TENANT_PROVIDER=local`

- Deleted `cmd/drive9-server-local/` and `scripts/drive9-server-local-env.sh`.
- Added `pkg/tenant/local`: a test/dev **Provisioner** on an existing TiDB instance. `POST /v1/provision` `CREATE DATABASE drive9_<tenantid>`, inits the app schema, and returns coordinates for the production pool.Acquire path. Deprovision drops that database.
- Production `/v1/provision` JWT is used (not `local-dev-key`).
- `make run-server-local` now builds `drive9-server` and sets `DRIVE9_TENANT_PROVIDER=local`, with local defaults for listen address, keys, mock S3, and DSN `root@tcp(127.0.0.1:4000)/drive9_local`.
- Local tenants still skip the TiDB auto-embedding ensure path (`SkipTiDBSchemaCheck`) because the provisioner inits schema itself.

Local no longer means a stripped-down server. Provision, auth, tenant pool, delete, and schema init go through the same code as `tidb_zero` / `tidb_cloud_native`.

### Replaceable server binary

- Workflows and `scripts/e2e-local.sh` honor `DRIVE9_SERVER_BIN` when set (the executable must exist); otherwise they `make run-server-local`.
- Use this to point e2e at an already-built `drive9-server` instead of rebuilding from this repo.

### Tests all sit on TiDB

**Unit tests:** `internal/testtidb` starts `pingcap/tidb` (unistore-style container, native arch, `--store=unistore`). Shared `TestMain` packages switched from MySQL DSN to `DRIVE9_TEST_TIDB_DSN`. Embedding writeback tests pad vectors to the schema dimension.

**E2E / blackbox / SDK:**
- New `scripts/e2e-local.sh` + `make e2e-local`: start TiDB if `:4000` is down, start `drive9-server` (`provider=local`), run `e2e/smoke-all.sh`.
- `e2e/smoke-all.sh` default set matches the `local-e2e.yml` **PR gate** (api, cli, layer-fs, fuse-release-gate, fuse-patch-storage-class, git-ops, git-workspace-ondemand, pack, fuse-crash-recovery, fuse-supervision, fuse-write-perf-budget).
- FUSE mounts in e2e pass `--mode=fuse` so macOS (macFUSE) does not fall through to WebDAV (`auto` on Darwin).
- Semantic / SQL / tenant-fork cases default to skip (`RUN_SEMANTIC_CHECKS`, `RUN_SQL_CHECKS`, `RUN_CLI_SEMANTIC_CHECKS`, `RUN_CLI_FORK_CHECKS`).
- Blackbox, fuse-patch, and SDK integration use playground/TiDB DSN + `app` embedding mode instead of MySQL 8.
- Removed `e2e/local-smoke.sh` and `scripts/verify_description.sh` (covered by `description-smoke-test.sh`).

CI `local-e2e.yml` still uses a TiUP playground for a closer-to-cluster TiDB; the suite list is the same PR set, with extra FUSE/git work on post-merge/nightly.

### Other

- Stub audio extractor (`DRIVE9_AUDIO_EXTRACT_MODE=stub`) so local audio extract can run without a real ASR backend.
- Schema dump: `drive9-server schema dump-init-sql --provider local` exports the app-managed TiDB statements (no `EMBED_TEXT` generated columns).
- `SupportsClusterDelete(local)` is true so tenant delete can `DROP DATABASE`.
- Docs (`AGENTS.md`, `e2e/AGENTS.md`, `e2e/README.md`, blackbox/SDK READMEs) updated for `provider=local`, TiDB DSN, and `make e2e-local`.

## How to test

```bash
# unit tests (uses DRIVE9_TEST_TIDB_DSN or starts pingcap/tidb)
make test TEST_PKGS='./pkg/tenant/local ./internal/testtidb ./pkg/tenant/schema'

# one-shot local e2e (starts TiDB if needed, then drive9-server provider=local)
make e2e-local

# api+cli only
RUN_API_ONLY=1 make e2e-local

# use a prebuilt drive9-server
DRIVE9_SERVER_BIN=/path/to/drive9-server make e2e-local E2E_LOCAL_ARGS='--no-build'
```

Interactive local server (same binary as production):

```bash
make run-server-local   # DRIVE9_TENANT_PROVIDER=local, listen 127.0.0.1:9009
```

Requires TiDB on `127.0.0.1:4000` (or let `make e2e-local` start `pingcap/tidb`). Suites `POST /v1/provision` for a JWT; do not use `local-dev-key`.


qiffang (approved):
Reviewed exact head `9d767ecf287722dbe490f0e1ee3230f6a7c93615` against base `ffe6663c97e0fc1c8ac2b1dafe03a54d32aee77e`.

APPROVE. I did not find a blocking correctness, auth, data-safety, or coverage regression in this refactor.

What I checked:
- `DRIVE9_TENANT_PROVIDER=local` now boots through the production `drive9-server` path while defaulting only local-mode env. The default token/master keys are rejected on wildcard/non-loopback listen, and public `/v1/provision` for `ProviderLocal` rejects credential bodies instead of accepting arbitrary local credentials.
- The local provisioner creates/drops only `drive9_` tenant DBs, initializes the real TiDB app/VECTOR schema, and keeps local DB TLS off instead of inheriting TiDB Cloud credential TLS behavior.
- Removing `mysql_no_embedding` is a net schema-coverage improvement: tests now run against the production TiDB app schema instead of a MySQL-only no-vector fork, and the old `none`/`auto` local embedding modes are rejected explicitly.
- The audio real providers are not removed: `openai` and `qwen-asr` remain in `cmd/drive9-server/main.go`; the new stub is explicit `DRIVE9_AUDIO_EXTRACT_MODE=stub` only, with default still `openai` when audio extraction is enabled.
- Hosted checks on the exact head are green: Code CI, Local E2E, and Python/JS/Rust/Kotlin/Swift SDK checks.

Local validation I ran: `git diff --check`, compile check for the touched server/tenant/backend/datastore/client packages, `go vet ./cmd/drive9-server ./pkg/tenant/local ./pkg/tenant ./pkg/tenant/schema ./pkg/server ./pkg/backend ./pkg/datastore ./pkg/client`, and `go test ./cmd/drive9-server -count=1`. Runtime tenant/local package tests still hit this machine's rootless-Docker/TestMain panic before selected tests can run; hosted CI covers those TiDB-container paths.

Non-blocking should-fix notes:
- Re-home the deleted stub audio env-wiring/transcript test into `cmd/drive9-server`; the real provider paths are covered, but the explicit `stub` mode is currently untested.
- For local provider with custom keys plus wildcard/non-loopback listen and no explicit `DRIVE9_PUBLIC_URL`, the default currently becomes `http://<listen>` such as `http://0.0.0.0:9009`; that is usually not an externally usable presign base. Consider keeping the old fail-fast for wildcard/non-loopback unless `DRIVE9_PUBLIC_URL` is set.
- `embedding_writeback_test.go` now reasonably avoids exact VECTOR string formatting, but it should also assert tail/padding semantics so padding corruption is still caught.
- Unify TiDB versions between testcontainers (`v8.5.6`) and local-e2e (`v8.5.5`) to avoid version-skew-only failures.
- Touched `pkg/leader/*` files still lack final newlines, so `gofmt -l` reports them even though this predates the PR and CI is green.

PR #857, opened 2026-08-26
fix: restore Windows CLI release build

mashenjun:
## Summary

- move `withEnsureCredentialEnv` out of the Unix-only mount supervision file
- make the credential environment wrapper available to Windows CLI builds without changing its behavior
- fix the Windows cross-build failure from [release run 32922898362](https://github.com/mem9-ai/drive9/actions/runs/32922898362/job/98039827621)

## Validation

- `go vet ./cmd/drive9/cli`
- `make build-cli-release DIST_DIR=/private/tmp/drive9-cli-release-test VERSION=dev CLI_TARGETS="linux/amd64 linux/arm64 darwin/amd64 darwin/arm64 windows/amd64 windows/arm64"`

## Summary by CodeRabbit

* **Bug Fixes**
  * Mount operations now reliably apply the configured server and mount credentials.
  * Credential environment settings are preserved and restored after mounting.
  * Token credentials take precedence over API keys, while alternate credential values are cleared to prevent conflicts.
  * Remount and adoption operations no longer unexpectedly alter existing credential settings.

### DIFF
--- a/.github/workflows/code-ci.yml
+++ b/.github/workflows/code-ci.yml
@@ -56,3 +56,6 @@ jobs:
 
       - name: Build (server and cli)
         run: make build
+
+      - name: Build Windows CLI
+        run: CGO_ENABLED=0 GOOS=windows GOARCH=amd64 go build -o /tmp/drive9-windows-amd64.exe ./cmd/drive9
--- a/cmd/drive9/cli/mount_credentials.go
+++ b/cmd/drive9/cli/mount_credentials.go
@@ -1,0 +1,42 @@
+package cli
+
+import "os"
+
+// withEnsureCredentialEnv temporarily installs stored mount credentials into
+// the process environment for a mount operation.
+func withEnsureCredentialEnv(server, apiKey, token string, fn func() error) error {
+	if server == "" && apiKey == "" && token == "" {
+		return fn()
+	}
+	type snap struct {
+		key string
+		val string
+		ok  bool
+	}
+	keys := []string{EnvServer, EnvAPIKey, EnvVaultToken}
+	prev := make([]snap, 0, len(keys))
+	for _, k := range keys {
+		v, ok := os.LookupEnv(k)
+		prev = append(prev, snap{key: k, val: v, ok: ok})
+	}
+	defer func() {
+		for _, s := range prev {
+			if s.ok {
+				_ = os.Setenv(s.key, s.val)
+			} else {
+				_ = os.Unsetenv(s.key)
+			}
+		}
+	}()
+	if server != "" {
+		_ = os.Setenv(EnvServer, server)
+	}
+	if token != "" {
+		_ = os.Setenv(EnvVaultToken, token)
+		_ = os.Unsetenv(EnvAPIKey)
+	} else if apiKey != "" {
+		_ = os.Setenv(EnvAPIKey, apiKey)
+		_ = os.Unsetenv(EnvVaultToken)
+	}
+	return fn()
+}
--- a/cmd/drive9/cli/mount_supervise.go
+++ b/cmd/drive9/cli/mount_supervise.go
@@ -489,45 +489,6 @@ func argsHasFlag(args []string, name string) bool {
 	return false
 }
 
-// withEnsureCredentialEnv temporarily installs stored mount credentials into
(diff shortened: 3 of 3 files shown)


qiffang (changes requested):
Re-reviewed exact head `1f7c0e7b1901c34ab690110791abfe535f381008`.

The implementation itself is correct and behavior-preserving: `withEnsureCredentialEnv` is moved unchanged from the `!windows` file into a platform-neutral file, so the untagged caller in `mount.go` is now available on Windows while Unix keeps the same snapshot/restore and token-over-API-key behavior. I reproduced the regression discriminator: base `ca125063` fails `GOOS=windows GOARCH=amd64 go build ./cmd/drive9` with `undefined: withEnsureCredentialEnv`; this head passes. Native CLI tests/vet, Windows test compilation, all six release binary builds, gofmt, and diff-check also pass (the aggregate Make target only fails afterward on this macOS host because `sha256sum` is unavailable, after every binary was built).

One blocking coverage gap remains: required `Code CI` still only builds the runner's Linux target. The exact base regression passed required PR CI and was discovered only later by the manually dispatched release workflow. This PR adds no committed Windows cross-build check, so the same build-tag class can regress again without any required PR check failing.

Please add a minimal required Windows compile gate, for example `CGO_ENABLED=0 GOOS=windows GOARCH=amd64 go build ./cmd/drive9` in `.github/workflows/code-ci.yml` (or an equivalent required check). That is a discriminating regression: it fails on the base and passes on this head. The credential-wrapper code needs no further changes.

mornyx (approved):
## Review summary

This restores `GOOS=windows` CLI compilation by moving `withEnsureCredentialEnv` out of `mount_supervise.go` (`//go:build !windows`) into a new untagged `mount_credentials.go`. That matches the untagged caller in `mount.go` (object-store foreground mount) without duplicating the helper in a Windows stub. The body is unchanged: snapshot/restore of `DRIVE9_SERVER` / `DRIVE9_API_KEY` / `DRIVE9_VAULT_TOKEN`, token-over-API-key with the other credential unset, and a no-op when all three inputs are empty. Under a strict maintainability bar this is the minimal correct structure; nothing here is merge-blocking.

**What looks good**
- One shared implementation in an untagged file, not a Windows stub that would skip credential install if that path ever lights up.
- The new file has no accidental platform tag; the helper is defined exactly once after the move.
- Validation covers the full release matrix (`windows/amd64` and `windows/arm64`), which is the failure that actually broke the release job.

---

### Nits
- `TestWithEnsureCredentialEnvRestores` still lives in `mount_supervise_test.go` (`//go:build !windows`). The helper is now compiled on all platforms. Not a coverage hole today — the logic is portable `os.Setenv`/`Unsetenv`, Unix CI already runs the test, and Windows never reaches this helper at runtime (object mounts are rejected; supervise is stubbed). Moving the test into an untagged `mount_credentials_test.go` would keep test placement aligned with the code.
- Pre-existing detection gap, not introduced here: PR `code-ci.yml` only `make build`s the host (Linux) CLI. Windows compile is checked in the manually dispatched `release-cli.yml`, which is how this broke. A cheap `GOOS=windows GOARCH=amd64 go build ./cmd/drive9` (and the kubectl plugin, if that matrix ships Windows too) in PR CI would catch the next tag mismatch before merge. Optional follow-up; this PR does not need to take it on.

---

### Recommendation
**Approve.** The untagged move closes the Windows missing-symbol failure with identical credential-env behavior; remaining notes are optional placement and a pre-existing CI smoke gap.

mashenjun:
@qiffang Addressed the blocking coverage gap in `bd7ad094`.

- Added a required `Build Windows CLI` step to `.github/workflows/code-ci.yml` using the requested discriminator: `CGO_ENABLED=0 GOOS=windows GOARCH=amd64 go build -o /tmp/drive9-windows-amd64.exe ./cmd/drive9`.
- Confirmed locally that the base fails this command and the current head passes; all six release targets also build.
- The exact-head required [Code CI run 32924757361](https://github.com/mem9-ai/drive9/actions/runs/32924757361) is now green, including the new Windows compile gate.

The review has no inline thread to resolve because it was submitted as a PR-level `CHANGES_REQUESTED` review. Please re-review and update the review state.

PR #858, opened 2026-08-26
feat: first-wave object-store CLI defaults and --auth=server docs

mornyx:
## Summary

Follow-up to [#845](https://github.com/mem9-ai/drive9/pull/845) after hosted `--auth=server` testing on S3, COS, and TOS.

Customer-facing setup for server mint is in [`docs/design/object-store-cli.md`](docs/design/object-store-cli.md): common `--auth=server` flow, then one section each for **Amazon S3**, **Tencent COS**, and **Volcengine TOS** (what to create in the cloud console, which flags to pass, tenant commands). Schema note now matches the control plane (`tenant_object_namespaces`, not a column on `tenant_tidbcloud_org_bindings`).

CLI data-plane fixes from that testing:

- **COS** — when `?region=` / minted region is set and no endpoint is given, use `https://cos.<region>.myqcloud.com` instead of rclone's AWS default.
- **TOS** — S3-compatible gateway requires virtual-host addressing (`force_path_style=false`). Path-style fails with `InvalidPathAccess`.
- **Mount STS refresh** — `DRIVE9_OBJECT_SESSION_REFRESH_MIN_LEAD` / `MAX_LEAD` override remint lead (hosted tests remint in seconds instead of waiting ~45m of a 1h session). Refresh logs a truncated access key id so two ASIA keys from the same account are distinguishable.

Manual hosted e2e (not wired into local-e2e): `e2e/object-auth-s3-hosted-test.sh`, `object-auth-cos-hosted-test.sh`, `object-auth-tos-hosted-test.sh` — two-tenant `cust` vs `cust-evil` isolation, fail-closed mint, read vs write, drive9↔object copy, FUSE mount, in-place STS refresh.

Companion server PR: [tidbcloud/fs#109](https://github.com/tidbcloud/fs/pull/109).

## Test plan

- [x] `go test ./pkg/objectfs/`
- [x] Hosted `--auth=server` S3 / COS / TOS (9/9 smoke + 44/44 hosted coverage against drive9-dev)
- [ ] Code CI

## Summary by CodeRabbit

- **Documentation**
  - Expanded object-store authentication guidance, including tenant isolation, credential scoping, session renewal, administrative workflows, and provider-specific setup for S3, COS, TOS, OSS, GCS, and Azure.
  - Added configuration examples and validation requirements for local and hosted authentication.

- **Bug Fixes**
  - Improved default endpoint handling for Tencent COS.
  - Corrected Volcengine TOS connection behavior.
  - Made session refresh timing configurable and more reliable, with safer diagnostic logging.

- **Tests**
  - Added hosted smoke coverage for S3, COS, and TOS isolation, mounting, copying, write-back, read-only access, and credential refresh.


qiffang (changes requested):
Reviewed exact head `78a5993ef29fbf122d230801788214ab2dc490b3`.

Blocking isolation-contract gap: this PR tells operators to "Bind each Drive9 tenant to a unique prefix id" and describes minted sessions as tenant-isolated, but the server does not enforce that uniqueness. In `tidbcloud/fs`, `tenant_object_namespaces` is keyed only by `tenant_id`; `namespace_id` has no unique constraint, and `SetTenantObjectNamespaceID` only locks/upserts that tenant row without checking another tenant's namespace. (The adjacent `storage_namespaces` table does enforce `namespace_id` ownership, so the missing guard is especially visible.)

Two tenants in the same organization can therefore be assigned the same non-empty namespace. Mint then derives both sessions from the same org backend plus the same namespace, giving both tenants credentials for the same bucket/prefix and defeating the cross-tenant isolation promised by this document and the hosted tests.

Before publishing this customer workflow, enforce namespace ownership transactionally at the real isolation boundary (at least organization/backend + namespace), and add a deterministic regression: tenant A claims `cust`; tenant B in the same boundary must fail, including a concurrent two-claim one-winner case. Empty/cleared namespaces should remain allowed. Alternatively, remove the isolation/unique-prefix guarantee and explicitly document that the operator owns collision prevention, but that would be a materially weaker security contract.

Also fix the local-auth example: the URI containing `&forcePathStyle=true` must be quoted. As written, a shell treats `&` as background syntax and does not pass that query parameter to `drive9`.

The provider defaults/refresh code itself looks correct: COS derives the official regional endpoint, TOS forces virtual-host style unless explicitly overridden, refresh lead overrides fail safely, and access-key logging is truncated. Local `pkg/objectfs` normal+race tests, JSON examples, shell syntax checks, and diff-check pass; exact-head Code CI, Local E2E, and CodeRabbit are green. Those checks do not cover the cross-tenant namespace collision.

qiffang:
Two additional non-blocking fixes found in the final script/document pass:

1. All three hosted scripts support `DRIVE9_E2E_OBJECT_PREFIX` when registering the backend and building Drive9 URIs, but the raw cloud-provider verification still addresses `${NS_A}/...` and `${NS_B}/...` directly (list/get/put). With a non-empty backend prefix, Drive9 writes under `<backend-prefix>/<namespace>/...` while these assertions probe `<namespace>/...`; the documented prefix mode therefore tests the wrong keys. Use a shared helper that prepends the configured backend prefix for every raw provider key/list prefix and expected key.

2. The blanket statements that minted credentials are "scoped to that tenant's prefix" and that the server returns a "prefix-scoped session" need to be limited to providers where that is true (the first-wave S3/COS/TOS path). The same document correctly says GCS is not downscoped and Azure is container-scoped, so the current global wording is a contradictory security contract.

These do not change the main CHANGES_REQUESTED blocker about duplicate tenant namespace ownership, but should be fixed in the same documentation/test pass.

mornyx:
@qiffang Thanks. Namespace uniqueness is an operator choice — the server still allows two tenants to share an id, and the document now says so (including schema / limits). Also addressed the rest of the review:

- quoted the `--auth=local` MinIO URI so `&forcePathStyle=true` is not backgrounded
- prefix-scoped mint wording is limited to S3/COS/TOS; GCS/Azure stay as they are
- hosted S3/COS/TOS scripts prepend `DRIVE9_E2E_OBJECT_PREFIX` on every native list/get/put key via a shared helper

`449230fc`

qiffang (approved):

qiffang (changes requested):
Re-reviewed exact head `449230fc5d7d2a9c5e343f739e7fbfb418fccffd`, including every unresolved inline thread. The previous namespace false-contract finding is closed: the document now explicitly says namespace uniqueness is operator-managed and equal non-empty IDs intentionally share one prefix. The quoted URI, provider-specific scope wording, and configured backend-prefix handling are also fixed.

I still request changes because the hosted security scripts have current safety and discrimination gaps:

1. **Fixed namespaces make cleanup destructive across runs.** All three scripts bind tenants to `cust` / `cust-evil` and then recursively remove those entire prefixes (`e2e/object-auth-s3-hosted-test.sh:389`, `:317`; COS `:388`, `:315`; TOS `:399`, `:326`). Concurrent runs, a stale previous run, or real data under those generic prefixes can be overwritten or deleted. Generate a run-unique prefix-confusion pair (preserving `A` vs `A-evil`) and clean only that run's owned prefixes/keys.

2. **Several denial checks can pass on unrelated failures.** The admin gate accepts every status except 200/201 (`s3:409-415`, COS `:408-414`, TOS `:419-425`), so 404/5xx also pass; this route's current tenant-bearer contract is 401, so assert the specific denial status/body. COS/TOS `list_keys` discard the command exit status (`cos:248-267`, TOS `:256-278`), allowing a network/tool/endpoint failure with empty output to count as isolation. The read-only PutObject checks accept any nonzero exit (`s3:471-476`, COS/TOS equivalents), so malformed arguments or endpoint failures also count as policy enforcement. Capture every exit code, treat only a verified provider `AccessDenied`/`Forbidden` response as the expected denial, and fail closed on every other error.

3. **S3 native checks ignore the configured endpoint.** `aws_with_session` reads only credentials and region and invokes `aws` without `--endpoint-url` (`e2e/object-auth-s3-hosted-test.sh:230-246`). When `DRIVE9_E2E_OBJECT_ENDPOINT` selects an S3-compatible backend, Drive9 tests that backend while the native isolation/read-only checks go to public AWS. Use the minted endpoint, with the configured endpoint as fallback.

4. **The refresh section does not prove the mount uses refreshed credentials.** `initial_ak` comes from an independent mint that was never supplied to the mount (`s3:529-557`; same pattern in COS/TOS), and post-refresh I/O runs while the mount's original one-hour session is still valid. A regression that logs a remint but fails to apply it to rclone still passes. Invalidate/expire the mount's actual original session before post-refresh I/O, or add an equivalent provider-side/request-signature discriminator proving the new session signs the request.

Validation: exact-head Code CI and Local E2E are green; `git diff --check`, `bash -n` for all three hosted scripts, `go test ./pkg/objectfs`, and `go vet ./pkg/objectfs` pass locally. `shellcheck` is not installed, and I could not execute the manual hosted scripts without the three providers' live credentials.

PR #867, opened 2026-08-28
feat: add embedding config SDK, CLI, and hosted E2E

srstack:
## Summary

- add typed Go SDK GET/PUT support for tenant `embedding-config`
- add `drive9 admin tenant embedding-config get|set` with full-replacement validation and API-key redaction
- expose extract provider `protocol` through the existing Go SDK and CLI
- add manual-only hosted video extraction and embedding processing E2E scripts
- document required provider variables, the MP4 fixture, and the fixed 1024-dimension embedding contract

## Verification

- `go test ./pkg/client -run 'TestAdmin.*Tenant(Extract|Embedding)Config' -count=1`\n- `go test ./cmd/drive9/cli -run 'TestAdminTenant(Extract|Embedding)Config' -count=1`\n- `bash -n e2e/video-extract-config-smoke-test.sh e2e/embedding-config-smoke-test.sh`\n- verified both hosted scripts skip before network access when required environment variables are absent\n- `gofmt -l` on changed Go files and `git diff --check`\n\nThe billable hosted provider paths were not executed locally; they remain manual-only and require control-plane credentials, provider credentials, and an MP4 fixture for the video suite.

## Summary by CodeRabbit

* **New Features**
  * Added administrative commands and SDK support to view and update tenant embedding configurations.
  * Added protocol selection for tenant video extract configurations.
  * Sensitive provider API keys are masked in displayed results.
  * Configuration changes validate required and incompatible options.

* **Documentation**
  * Updated command help and end-to-end testing guidance for embedding and video extraction.

* **Tests**
  * Expanded coverage for configuration operations, validation, secure HTTPS endpoints, masking, and smoke-test workflows.

### DIFF
--- a/.github/workflows/code-ci.yml
+++ b/.github/workflows/code-ci.yml
@@ -14,6 +14,9 @@ jobs:
       - name: Checkout
         uses: actions/checkout@v4
 
+      - name: Test hosted config smoke guards
+        run: bash scripts/test-hosted-config-smoke-guards.sh
+
       - name: Checkout agfs dependency
         run: |
           git clone --depth 1 https://github.com/c4pt0r/agfs ../agfs
--- a/cmd/drive9/cli/admin.go
+++ b/cmd/drive9/cli/admin.go
@@ -51,6 +51,8 @@ func adminTenant(args []string) error {
 		return quotaSet(args[1:])
 	case "extract-config":
 		return adminTenantExtractConfig(args[1:])
+	case "embedding-config":
+		return adminTenantEmbeddingConfig(args[1:])
 	case "object-namespace":
 		return adminTenantObjectNamespace(args[1:])
 	case "pool":
@@ -679,6 +681,7 @@ commands:
   tenant delete --tenant-id ID      delete one tenant
   tenant set-quota --tenant-id ID   set quota for one tenant
   tenant extract-config <get|set>   get or update media extract config
+  tenant embedding-config <get|set>  get or replace embedding config
   tenant object-namespace <get|set|clear>  bind a customer object prefix id
   object-backend <add|get|ls|update|rm>  org object-store credentials
   pool <command> [flags]            manage the tenant pool
@@ -734,12 +737,13 @@ commands:
   delete --tenant-id ID            delete one tenant
   set-quota --tenant-id ID         set quota for one tenant
   extract-config <get|set>         get or update media extract config
+  embedding-config <get|set>       get or replace embedding config
   object-namespace <get|set|clear> bind customer object prefix id
 
 flags:
   --server URL                     server URL (default: active context server)
   --region-code CODE               TiDBCloud Mode region code; ignored when --server is set
-  --tenant-id ID                   drive9 tenant id for get/delete/set-quota/extract-config/object-namespace
+  --tenant-id ID                   drive9 tenant id for get/delete/set-quota/extract-config/embedding-config/object-namespace
   --tidbcloud-public-key KEY       TiDB Cloud public key
   --tidbcloud-private-key KEY      TiDB Cloud private key
   --max-storage-size Mi            set-quota: max confirmed+reserved storage size in Mi
--- a/cmd/drive9/cli/admin_embedding_config.go
+++ b/cmd/drive9/cli/admin_embedding_config.go
@@ -1,0 +1,298 @@
+package cli
+
+import (
+	"context"
+	"encoding/json"
+	"errors"
+	"fmt"
+	"os"
+	"strings"
+	"text/tabwriter"
+
+	"github.com/mem9-ai/drive9/pkg/client"
+)
+
+func adminTenantEmbeddingConfig(args []string) error {
+	if len(args) == 0 {
+		return fmt.Errorf("%s", adminTenantEmbeddingConfigUsage())
+	}
+	switch args[0] {
+	case "-h", "-help", "--help", "help":
+		_, _ = fmt.Fprintln(os.Stdout, adminTenantEmbeddingConfigUsage())
+		return nil
+	case "get":
+		return adminTenantEmbeddingConfigGet(args[1:])
+	case "set":
+		return adminTenantEmbeddingConfigSet(args[1:])
+	default:
+		return fmt.Errorf("unknown admin tenant embedding-config command %q\n%s", args[0], adminTenantEmbeddingConfigUsage())
+	}
+}
+
+type adminTenantEmbeddingConfigFlags struct {
+	serverFlag      string
+	serverGiven     bool
+	regionCodeFlag  string
+	regionCodeGiven bool
+	tenantID        string
+	tenantIDGiven   bool
+	publicKeyFlag   string
+	publicKeyGiven  bool
+	privateKeyFlag  string
+	privateKeyGiven bool
+	asJSON          bool
+	enabled         *bool
+	apiBase         *string
+	apiKey          *string
+	model           *string
+}
+
+func parseAdminTenantEmbeddingConfigFlags(args []string, usage string, allowConfig bool) (adminTenantEmbeddingConfigFlags, error) {
+	var flags adminTenantEmbeddingConfigFlags
+	for i := 0; i < len(args); i++ {
+		switch args[i] {
+		case "-h", "-help", "--help", "help":
+			_, _ = fmt.Fprintln(os.Stdout, usage)
+			return flags, errHelpRequested{}
+		case "--server":
+			value, next, err := nextAdminExtractFlag(args, i, "--server")
+			if err != nil {
+				return flags, err
+			}
+			flags.serverFlag, flags.serverGiven, i = value, true, next
+		case "--region-code":
+			value, next, err := nextAdminExtractFlag(args, i, "--region-code")
+			if err != nil {
+				return flags, err
+			}
+			flags.regionCodeFlag, flags.regionCodeGiven, i = value, true, next
+		case "--tenant-id":
+			value, next, err := nextAdminExtractFlag(args, i, "--tenant-id")
+			if err != nil {
+				return flags, err
+			}
+			flags.tenantID, flags.tenantIDGiven, i = value, true, next
+		case "--tidbcloud-public-key":
+			value, next, err := nextAdminExtractFlag(args, i, "--tidbcloud-public-key")
+			if err != nil {
+				return flags, err
+			}
+			flags.publicKeyFlag, flags.publicKeyGiven, i = value, true, next
+		case "--tidbcloud-private-key":
+			value, next, err := nextAdminExtractFlag(args, i, "--tidbcloud-private-key")
+			if err != nil {
+				return flags, err
+			}
+			flags.privateKeyFlag, flags.privateKeyGiven, i = value, true, next
+		case "--json":
+			flags.asJSON = true
+		case "--enabled":
+			if !allowConfig {
+				return flags, fmt.Errorf("unknown flag %q\n%s", args[i], usage)
+			}
+			value, next, err := nextAdminExtractFlag(args, i, "--enabled")
+			if err != nil {
+				return flags, err
+			}
+			parsed, err := parseAdminExtractBool(value)
+			if err != nil {
+				return flags, err
+			}
+			flags.enabled, i = &parsed, next
+		case "--api-base", "--api-key", "--model":
+			if !allowConfig {
+				return flags, fmt.Errorf("unknown flag %q\n%s", args[i], usage)
+			}
+			name := args[i]
+			value, next, err := nextAdminExtractFlag(args, i, name)
+			if err != nil {
+				return flags, err
+			}
+			switch name {
+			case "--api-base":
+				flags.apiBase = &value
+			case "--api-key":
+				flags.apiKey = &value
+			case "--model":
+				flags.model = &value
+			}
+			i = next
+		default:
+			return flags, fmt.Errorf("unknown flag %q\n%s", args[i], usage)
+		}
+	}
+	if err := rejectEmptyFlag("server", strings.TrimSpace(flags.serverFlag), flags.serverGiven); err != nil {
+		return flags, err
+	}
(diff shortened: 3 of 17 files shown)


qiffang (changes requested):
Reviewed exact head `604ba5cf0b9fcec8f535dd35907ce4049d5e5e0e`. The SDK/CLI contracts match the current `tidbcloud/fs` server, but the new hosted E2E coverage has two merge-blocking problems.

1. **The scripts can transmit real control-plane and provider credentials over cleartext HTTP.** `video-extract-config-smoke-test.sh` no longer requires `DRIVE9_BASE`; when it is omitted, lines 29-30 silently select a public `http://` ELB, then `/v1/provision` and the config PUT send `X-TiDBCloud-Private-Key`, the provisioned owner key, and the provider API key to that endpoint. `embedding-config-smoke-test.sh` requires a base but accepts any scheme, while the repository's hosted-endpoint instructions advertise the same HTTP dev deployment. These billable credential-bearing tests must require an explicit HTTPS base (or allow cleartext only for an explicit loopback/insecure opt-in) and must fail before network access otherwise. Please add a no-network validation regression so the old default cannot return.

2. **The embedding smoke can pass without the tenant-managed provider being used by the worker/query path.** Lines 204-214 accept an initial `source` of either `none` or `default`. After the custom PUT, the only data-plane assertion is semantic recall at lines 278-294. The production resolver intentionally uses the process default when no tenant row is observed. Therefore a regression where pool invalidation, worker resolution, or query resolution ignores the just-written tenant config can still embed both documents and the query with the process default and make this test green. The `source=custom` PUT response proves persistence, not data-plane consumption. Run this case only when the initial source is `none`, or use an instrumented/distinguishable provider and assert that both document and query requests actually hit the configured tenant provider.

Also tighten the video assertion before relying on it as visual-content coverage: the only checked token is injected verbatim into the model prompt (lines 197-204, 239-248), so a generic prompt-following response can pass without proving any fixture-specific visual fact. Use a deterministic fixture and assert a visible fact/token that is not present in the prompt.

The remaining changed surfaces look correct: embedding GET/PUT full-replacement fields, masking/redaction, extract `protocol` plumbing, CLI validation/help, and cookbook method coverage align with the production server. Exact-head Code CI and Local E2E are green; local CLI/cookbook tests, package compile/vet, `bash -n`, gofmt, and diff-check pass. The billable live scripts were not executed and their current assertions are not sufficient substitutes.

mornyx (commented):
## Review summary

Re-review of HEAD `1deea1ca` after the previous request-changes on `604ba5cf`. The SDK/CLI work is in good shape: embedding-config GET/PUT matches the quota control-plane contract (TiDB Cloud headers, no tenant Bearer, full-replacement body, CLI redaction), extract `protocol` is plumbed through the existing types rather than a second path, and cookbook coverage is registered. The live credential leak and the embedding false-green hole are closed on this HEAD. What remains is assertion/protocol hardening on the new billable smokes and docs that still contradict the fail-closed HTTPS bar — strong should-fix, not a merge-blocking product bug.

**What looks good**
- Prior HTTP default is gone: all three hosted config smokes require an explicit `DRIVE9_BASE` and `die` on non-`https://` before any network I/O (embedding:56, video:57, image:64). Skip-before-network on missing env is intact.
- Embedding smoke now requires initial `source=none` (rejects `default` / `database_auto`) and uses a vocabulary-disjoint query, so persistence is not treated as proof of data-plane consumption.
- Video assertion no longer interpolates the marker into the prompt; the fixture token is caller-provided. Disable-then-reupload negative control is present.
- New Go stays in the canonical layer: `setQuotaHeaders`, conversion `AdminTenantEmbeddingConfigGetRequest(identity)` (S1016), Bearer omitted with a test, SET header asserts on enable and disable, cookbook methods registered.

---

**1. Video initiate PUT still sends the fixture body; the large-file contract does not read it**

`e2e/video-extract-config-smoke-test.sh:127-130` does `PUT /v1/fs/...` with `X-Dat9-Part-Checksums` **and** `--data-binary "@$VIDEO_FIXTURE"`. The server’s large-file branch (`pkg/server/server.go` around 2452–2523) answers `202` **without reading the body** once `IsLargeFile(cl)` is true (hosted S3 + size ≥ inline threshold, historically 50KB). The canonical client never does this: `initiateUploadLegacy` (`pkg/client/transfer.go:755-763`) sends a **bodyless** PUT with `X-Dat9-Content-Length` + `X-Dat9-Part-Checksums`. `e2e/api-smoke-test.sh` uses `POST /v1/uploads/initiate` with a JSON body.

Go’s HTTP server will only drain ~256KB of an unread body and then close the connection. On loopback a 5MB unread-body `202` can still surface to curl; across an ELB/proxy it is a race, and the bytes are discarded either way. The multipart commit is the primary video path and was never executed live.

Fix: drop `--data-binary` from the initiate PUT and send `X-Dat9-Content-Length` (or switch to `POST /v1/uploads/initiate` like api-smoke). Then actually run the script once against a hosted HTTPS deployment.

**2. Video smoke never requires `source=none` (image and embedding in this PR do)**

`e2e/image-extract-config-smoke-test.sh:221-231` and `e2e/embedding-config-smoke-test.sh:206-216` GET the initial config and refuse `source=default`. The video script PUTs custom config at line 261 with no prelude. If the hosted process has a default vision provider, later `semantic_text` can come from that default while the just-written tenant row is ignored. The disable-window check is not a substitute (a slow default extractor still leaves the second upload empty). Mirror the image GET/`source=none` block before the video PUT.

**3. Video marker is still not checked against the prompt**

The marker is no longer interpolated (that hole is closed), but the script never `die`s if `"$MARKER"` is a substring of the prompt at lines 268–269 (`video`, `frames`, `visible`, `content`, `English`, …). The assertion is only “`semantic_text` contains `$MARKER` and is longer than it.” Before upload, reject a marker that appears in the prompt (case-insensitive) and prefer a long fixture-specific token.

**4. No no-network regression for the HTTPS fail-closed guards**

The previous review asked for a check so the old `http://` ELB default cannot return. The guards exist; nothing automated pins them. A one-line deletion would reintroduce credential-bearing HTTP. Add a cheap no-network check (dummy keys, `DRIVE9_BASE=http://127.0.0.1:9` and the old ELB host, assert FAIL + SKIP when `DRIVE9_BASE` is unset, no real curl). Wire it into an existing cheap gate.

**5. Docs still fight the new HTTPS contract**

- `e2e/README.md:219-234` and `e2e/AGENTS.md:151-166` copy-paste blocks omit `DRIVE9_TIDBCLOUD_PUBLIC_KEY` / `DRIVE9_TIDBCLOUD_PRIVATE_KEY` while the image example immediately above includes them. A verbatim run SKIP-exits 0.
- `e2e/AGENTS.md:32,35` still advertises cleartext `http://` ELB hosts as the default `DRIVE9_BASE`. These three suites now hard-fail that value. Point credential-bearing suites at an HTTPS base (or explicitly forbid the HTTP dev endpoint for any script that sends control-plane or provider keys).

---

### Nits

- `pkg/client/admin_embedding_config.go:100` bypasses `c.do` on purpose (tested). Add a one-line comment so a later cleanup does not “unify” it back onto Bearer-attaching `c.do`. Aligning extract-config/quota onto the same helper is a follow-up, not this PR.
- `adminTenantEmbeddingConfigSetBody` duplicates `AdminTenantEmbeddingConfigSetRequest` JSON; marshal the request (or keep one private body type, not both).
- Embedding smoke has no post-disable data-plane negative control (image/video re-upload and wait). Optional symmetry: a third document that a disjoint query must not recall.

---

### Recommendation
**Comment.** The previous merge blockers (cleartext credentials, embedding smoke proving only persistence, prompt-parroted video token) are addressed on this HEAD. Remaining items are should-fix on the new billable smokes and docs; they are not an approve.

srstack:
Addressed the latest review in `4a9675a7`:

- switched video upload initiation to the canonical bodyless `POST /v1/uploads/initiate` multipart contract;
- required initial video config `source=none`;
- rejected case-insensitive marker/prompt overlap before network access;
- added a no-network CI regression for missing and insecure `DRIVE9_BASE` across image/video/embedding smokes, including the old HTTP ELB;
- completed hosted command examples with control-plane credentials and documented the HTTP endpoint restriction;
- removed the duplicate embedding SET body type and documented why embedding admin calls bypass the Bearer-attaching helper.

Focused verification: hosted guard regression passed, shell syntax checks passed, `go test ./pkg/client -run 'TestAdmin.*EmbeddingConfig' -count=1` passed, and `git diff --check` passed.\n\nThe optional embedding post-disable billable data-plane flow was not added: initial `source=none` plus the positive vocabulary-disjoint query proves tenant provider use, while image/video already cover disabled worker behavior. The live hosted video smoke was not rerun because the available local E2E environment currently uses an HTTP base and has no `DRIVE9_E2E_VIDEO_EXPECTED_MARKER`; the tightened script rejects that environment before network access.

mornyx (commented):
## Review summary

Re-review of HEAD `4a9675a` after the previous COMMENT on `1deea1ca`. The five should-fix items from that round are closed: video initiate is canonical `POST /v1/uploads/initiate` (JSON plan, no fixture body), video requires initial `source=none`, marker/prompt overlap is rejected before network I/O, `scripts/test-hosted-config-smoke-guards.sh` is wired as the first Code CI step and pins HTTPS fail-closed plus skip-before-network (including the old HTTP ELB), and the hosted copy-paste examples include control-plane keys while forbidding those HTTP bases for credential-bearing suites. SDK/CLI contracts were already in good shape. One new portability bug in the marker check keeps this from an approve.

**What looks good**
- Prior COMMENT C1–C5 landed as real code, not docs-only: bodyless JSON initiate matching `api-smoke` / `handleUploadInitiate`, GET/`source=none` prelude, pre-network marker check, PATH-shadowed fake `curl` in CI, docs restriction on HTTP ELBs.
- Duplicate embedding SET body type is gone; `doAdminTenantEmbeddingConfig` documents why it skips `c.do`.
- Author’s won’t-fix on embedding post-disable data-plane is accepted: `source=none` plus the disjoint query already prove tenant-provider use.

---

**1. Marker/prompt check uses bash 4-only `${var,,}` — stock macOS bash 3.2 dies**

`e2e/video-extract-config-smoke-test.sh:59`:

```bash
[[ "${VIDEO_PROMPT,,}" != *"${MARKER,,}"* ]] || die "DRIVE9_E2E_VIDEO_EXPECTED_MARKER must not appear in the video prompt"
```

`${var,,}` needs bash ≥ 4. This is the only such construct in the e2e tree. Default `/bin/bash` on macOS is 3.2.57; with required env set the script exits `line 59: ${VIDEO_PROMPT,,}: bad substitution` instead of the intended message. The new guard then fails locally with `FAIL: video smoke did not report the marker/prompt conflict` (reproduced on this machine). Ubuntu Code CI is fine (bash 5.x); developers on stock Mac cannot run the video smoke or the guard.

Fix with bash 3.2-safe case-folding, e.g. `shopt -s nocasematch` and `[[ "$VIDEO_PROMPT" == *"$MARKER"* ]]`, or `tr '[:upper:]' '[:lower:]'`. Re-run `bash scripts/test-hosted-config-smoke-guards.sh` under `/bin/bash` 3.2 afterwards.

---

### Nits

- On guard failure, dump `"$output_file"` before `exit 1` (the EXIT trap deletes `$FAKE_BIN`). Wrap the skip-case invocation (line 64) in the same `if … FAIL` shape as the HTTP cases so a non-SKIP crash is not a raw `set -e` abort.
- After video disable PUT, a short propagation wait before uploading `disabled.mp4` would match the embedding script’s outbox note (optional; same gap exists in the image smoke).

---

### Recommendation
**Comment.** Previous should-fix items are verified fixed on this HEAD; the new marker guard should not crash on the stock macOS shell. Not an approve until that one-liner is bash 3.2-safe.

srstack:
Addressed the latest portability review in `67349ab2`:

- replaced Bash 4-only `${var,,}` case-folding with Bash 3.2-compatible `nocasematch`;
- changed the regression marker to mixed-case `ViDeO`, so the guard pins case-insensitive prompt overlap;
- made guard failures print the captured child-script output before cleanup;
- wrapped the missing-`DRIVE9_BASE` path explicitly so a non-SKIP crash reports a useful failure instead of exiting through `set -e`.

Focused verification passed: `bash -n` for the video and guard scripts, the no-network guard suite, and `git diff --check`. This Linux environment has Bash 5.2 and no Bash 3 binary, so the script was not executed under Bash 3.2 locally; the replacement uses `shopt -s nocasematch`, which is supported by Bash 3.2, and the Bash 4-only expansion is gone.

I did not add the optional post-disable propagation sleep. That would change test timing and the same question applies to image smoke; it should be handled consistently rather than only for video.

qiffang (approved):
Reviewed latest head `67349ab290d87d5ea33adbf36ba29ad92be08287`.

The previous blockers are closed for this client/CLI/docs PR:

- credential-bearing hosted smoke scripts now require explicit `https://` `DRIVE9_BASE` and the guard test proves they reject both local HTTP and the old HTTP dev endpoint before calling `curl`;
- image/video/embedding hosted smokes require an initial `source=none` where needed, so they cannot pass by silently using a process default provider;
- the embedding smoke validates bad-key and unreachable-provider failures without persistence, then proves tenant-managed embedding by uploading target/distractor text and using a vocabulary-disjoint semantic query;
- the video smoke uses a caller-supplied fixture marker that is rejected if it appears in the prompt, so the success assertion is tied to fixture content rather than prompt leakage;
- the new embedding-config SDK/CLI request/response shape matches the authoritative server contract in `tidbcloud/fs` main: full replacement PUT, masked keys, `source` values, `generation`, and disabled custom off-marker behavior.

Hosted checks are green on this exact head (`Code CI`, `Local E2E`, CodeRabbit). Local verification: `git diff --check`, touched `gofmt`, shell syntax checks, `scripts/test-hosted-config-smoke-guards.sh`, CLI focused tests, package compile, and `go vet` passed. `pkg/client` runtime tests are still blocked locally by the repo TestMain rootless-Docker bootstrap; the compile check passed.

PR #881, opened 2026-09-02
feat(cli): add fs setmeta command and show description in fs stat

srstack:
## What

Add `drive9 fs setmeta` so tags and description can be updated at any time, not only at upload. Backed by the new server endpoint `POST /v1/fs/{path}?setmeta=1` (tidbcloud/fs#125).

```bash
drive9 fs setmeta :/path --tag owner=bob --tag env=prod      # replace user tag set
drive9 fs setmeta :/path --description "new desc"           # set description
drive9 fs setmeta :/path --clear-tags --clear-description    # clear both
```

## Details

- `pkg/client.SetMetadata(ctx, path, SetMetadataOptions)`: nil `Tags` / nil `Description` leave the field unchanged; empty map / empty string clear it (pointer fields on the wire distinguish absent from empty).
- Tag flags reuse `cp`'s `parseAndMergeTag` validation (key=value, dup-key rejection, tagutil rules); `--tag` vs `--clear-tags` and `--description` vs `--clear-description` are mutually exclusive; at least one operation is required.
- Capability check uses `CapWrite` (metadata write); no layer support (layer entries don't have file tag parity yet).
- `fs stat` now prints `description:` (text output, when non-empty) and includes `description` in `-o json`, matching the server's enriched stat.
- Registered in `runFS`, `fsUsage`, and visual help.

## Tests

- `cmd/drive9/cli/setmeta_test.go` (httptest): wire format (POST /v1/fs/{path}?setmeta=1 JSON body), clear semantics (`tags:{}`, `description:""`), flag validation matrix, server error propagation.
- `go build ./...`, `go test ./cmd/drive9/... ./pkg/client/`, `make lint` all clean.
- End-to-end smoke against a local drive9-server (tidbcloud/fs feat/setmeta) + TiDB: upload with `--tag`/`--description` → `fs setmeta` replace/clear → `fs stat` read-back; revision unchanged throughout.

## Summary by CodeRabbit

* **New Features**
  * Added the `drive9 fs setmeta` command to update file tags and descriptions without re-uploading content.
  * Supports replacing or clearing tags and descriptions with validation and clear error messages.
  * File status output now includes descriptions when available.
  * Added SDK and API support for updating file metadata.
* **Documentation**
  * Updated filesystem and visual help with `setmeta` usage and options.
  * Added metadata update examples to the Go SDK cookbook.

### DIFF
--- a/cmd/drive9/cli/setmeta.go
+++ b/cmd/drive9/cli/setmeta.go
@@ -1,0 +1,107 @@
+package cli
+
+import (
+	"fmt"
+	"strings"
+	"unicode/utf8"
+
+	"github.com/mem9-ai/drive9/pkg/backend"
+	"github.com/mem9-ai/drive9/pkg/client"
+)
+
+// SetMeta updates tags and/or description of an existing remote file without
+// rewriting its content.
+func SetMeta(c *client.Client, args []string) error {
+	authLocal, args, err := peelObjectAuth(args)
+	if err != nil {
+		return err
+	}
+	defer withObjectAuthLocal(authLocal)()
+
+	var tags map[string]string
+	tagsSet := false
+	clearTags := false
+	var description string
+	descriptionSet := false
+	clearDescription := false
+
+	var positional []string
+	for i := 0; i < len(args); i++ {
+		a := args[i]
+		switch {
+		case a == "--":
+			positional = append(positional, args[i+1:]...)
+			i = len(args)
+		case a == "--tag":
+			if i+1 >= len(args) {
+				return fmt.Errorf("--tag requires argument")
+			}
+			i++
+			tags, err = parseAndMergeTag(tags, args[i])
+			if err != nil {
+				return err
+			}
+			tagsSet = true
+		case strings.HasPrefix(a, "--tag="):
+			tags, err = parseAndMergeTag(tags, strings.TrimPrefix(a, "--tag="))
+			if err != nil {
+				return err
+			}
+			tagsSet = true
+		case a == "--clear-tags":
+			clearTags = true
+		case a == "--description":
+			if i+1 >= len(args) {
+				return fmt.Errorf("--description requires argument")
+			}
+			i++
+			description = args[i]
+			descriptionSet = true
+		case strings.HasPrefix(a, "--description="):
+			description = strings.TrimPrefix(a, "--description=")
+			descriptionSet = true
+		case a == "--clear-description":
+			clearDescription = true
+		case strings.HasPrefix(a, "-"):
+			return fmt.Errorf("unknown flag %q", a)
+		default:
+			positional = append(positional, a)
+		}
+	}
+	if tagsSet && clearTags {
+		return fmt.Errorf("--tag and --clear-tags are mutually exclusive")
+	}
+	if descriptionSet && clearDescription {
+		return fmt.Errorf("--description and --clear-description are mutually exclusive")
+	}
+	if !tagsSet && !clearTags && !descriptionSet && !clearDescription {
+		return fmt.Errorf("usage: drive9 fs setmeta [--tag key=value]... [--clear-tags] [--description <text>] [--clear-description] <path>")
+	}
+	if len(positional) != 1 {
+		return fmt.Errorf("usage: drive9 fs setmeta [--tag key=value]... [--clear-tags] [--description <text>] [--clear-description] <path>")
+	}
+	if descriptionSet && utf8.RuneCountInString(description) > backend.MaxDescriptionLen {
+		return fmt.Errorf("description exceeds %d characters", backend.MaxDescriptionLen)
+	}
+
+	h, err := fsHandleForArg(c, positional[0])
+	if err != nil {
+		return err
+	}
+	if err := requireCapOnHandle(h, CapWrite, "setmeta"); err != nil {
+		return err
+	}
+	c, path := h.Client, h.Path
+
+	var opts client.SetMetadataOptions
+	if tagsSet || clearTags {
+		if tags == nil {
+			tags = map[string]string{}
+		}
+		opts.Tags = tags
+	}
+	if descriptionSet || clearDescription {
+		opts.Description = &description
+	}
+	return c.SetMetadata(path, opts)
+}
--- a/cmd/drive9/cli/setmeta.go
+++ b/cmd/drive9/cli/setmeta.go
@@ -1,6 +1,7 @@
 package cli
 
 import (
+	"errors"
 	"fmt"
 	"strings"
 	"unicode/utf8"
@@ -12,6 +13,7 @@ import (
 // SetMeta updates tags and/or description of an existing remote file without
 // rewriting its content.
 func SetMeta(c *client.Client, args []string) error {
+	const usage = "usage: drive9 fs setmeta [--tag key=value]... [--clear-tags] [--description <text>] [--clear-description] <path>"
 	authLocal, args, err := peelObjectAuth(args)
 	if err != nil {
 		return err
@@ -75,10 +77,10 @@ func SetMeta(c *client.Client, args []string) error {
 		return fmt.Errorf("--description and --clear-description are mutually exclusive")
 	}
 	if !tagsSet && !clearTags && !descriptionSet && !clearDescription {
-		return fmt.Errorf("usage: drive9 fs setmeta [--tag key=value]... [--clear-tags] [--description <text>] [--clear-description] <path>")
+		return errors.New(usage)
 	}
 	if len(positional) != 1 {
-		return fmt.Errorf("usage: drive9 fs setmeta [--tag key=value]... [--clear-tags] [--description <text>] [--clear-description] <path>")
+		return errors.New(usage)
 	}
 	if descriptionSet && utf8.RuneCountInString(description) > backend.MaxDescriptionLen {
 		return fmt.Errorf("description exceeds %d characters", backend.MaxDescriptionLen)
@@ -88,8 +90,18 @@ func SetMeta(c *client.Client, args []string) error {
 	if err != nil {
 		return err
 	}
-	if err := requireCapOnHandle(h, CapWrite, "setmeta"); err != nil {
-		return err
+	// Object backends advertise CapWrite but carry no drive9 client and have
+	// no file metadata; gate on the metadata caps so object URIs fail closed
+	// instead of dereferencing a nil client.
+	if tagsSet || clearTags {
+		if err := requireCapOnHandle(h, CapTags, "setmeta"); err != nil {
+			return err
+		}
+	}
+	if descriptionSet || clearDescription {
+		if err := requireCapOnHandle(h, CapDescription, "setmeta"); err != nil {
+			return err
+		}
 	}
 	c, path := h.Client, h.Path
 
--- a/cmd/drive9/cli/setmeta_test.go
+++ b/cmd/drive9/cli/setmeta_test.go
@@ -8,6 +8,7 @@ import (
 	"strings"
 	"testing"
 
+	"github.com/mem9-ai/drive9/pkg/backend"
 	"github.com/mem9-ai/drive9/pkg/client"
 )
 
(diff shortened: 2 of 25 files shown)


mornyx (changes requested):
## Review summary

`drive9 fs setmeta` is a focused CLI+client addition: pointer fields on the wire correctly distinguish leave-unchanged from clear, tag parsing reuses `parseAndMergeTag`, and `fs stat` surfaces `description` in JSON unconditionally (matching `content_type` / `semantic_text`) while hiding empty text. Flag exclusivity, `validateTags(nil)` on description-only updates, URL/`do`/`readError` reuse, and `--layer` fail-closed (unknown flag) all look right.

The merge-blocking problem is the capability/dispatch path. `requireCapOnHandle(h, CapWrite, "setmeta")` is the wrong gate: object backends already advertise `CapWrite`, object handles carry a nil `Client` by design, and `SetMeta` then calls `SetMetadata` on that nil client. `s3://…` (and the other object schemes) therefore panic instead of failing closed.

**What looks good**
- Absent-vs-empty wire contract (`*map[string]string` / `*string` + `omitempty`) actually emits `tags:{}` and `description:""` on clear; the clear test asserts raw JSON tokens, not round-tripped emptiness.
- New client code lives in a small `pkg/client/setmeta.go` and reuses `url` / `do` / `readError` / `validateTags`, so `client.go` only grew by the `StatMetadataResult.Description` field.
- Help is updated in one pass (`fsUsage`, visual help), and unknown flags / `--` / mutual exclusion are stricter than they needed to be.

---

**1. `setmeta` panics on object-store URIs because it gates on `CapWrite`**

`cmd/drive9/cli/setmeta.go:91` calls `requireCapOnHandle(h, CapWrite, "setmeta")`, then `c, path := h.Client, h.Path` and `c.SetMetadata(...)`.

That combination fails open for object locations:

- `fsHandleForArg` builds `KindObject` handles with `Client: nil` (`cmd/drive9/cli/fsbackend.go:87-91`). That invariant is already pinned by `TestFsHandleForArgObjectDoesNotUseClient`.
- `objectBackend.Caps()` includes `CapWrite` (content write) but not `CapTags` / `CapDescription` (`fsbackend.go:644-647`).
- `requireCapOnHandle` only consults caps for `KindObject`; on success the command uses `h.Client`. For object URIs that client is nil, so `drive9 fs setmeta --tag a=b s3://bucket/key` panics in `(*Client).url` once the object backend opens (i.e. in the configurations where object URIs work at all).

`chmod` / `find` / `grep` avoid the same shape only because they require caps object backends actually lack (`CapChmod`, `CapSearch`). `CapTags` and `CapDescription` already exist for this distinction (`fsloc.go:109-110`) and are advertised by `drive9FS`, but no command requires them yet.

Do not “fix” this by falling back to the original `Client`: `h.Path` is the object key, so a POST to `/v1/fs/{key}?setmeta=1` could mutate an unrelated drive9 file.

Concrete fix (either is fine):

- Require the metadata caps instead of `CapWrite`, e.g. `requireCapOnHandle(h, CapTags|CapDescription, "setmeta")` (`Capability.Has` requires all bits, so object backends fail closed with the existing `backend "s3" does not support …` error). If you OR the bits, add a `capName` case so the error does not collapse to `"this operation"`. Even better: require `CapTags` when tags are being written and `CapDescription` when description is being written.
- Or explicitly reject `h.Loc.Kind == KindObject` (and nil `h.Client`) before the HTTP call, same idea as `symlink`: `"object-store URIs are not supported"`.

Add a regression test that `SetMeta(c, []string{"--tag", "a=b", "s3://bucket/k"})` returns an error and does not panic. `peelObjectAuth` is already wired, so this path is first-class, not hypothetical.

---

**2. Tests never pin the “leave unchanged” half of the wire contract**

`TestSetMetaSendsTagsAndDescription` and `TestSetMetaClearSendsEmptyValues` assert presence (`tags` + `description`, or `{}` / `""`). Nothing asserts absence.

A description-only update must omit the `tags` key entirely (and a tags-only update must omit `description`). That path is currently correct — nil `Tags` plus `omitempty` omits the field, and `validateTags(nil)` is a no-op — but a one-line refactor that always sets `reqBody.Tags = &opts.Tags` would silently wipe the tag set on every description-only call, and no existing test would fail.

Please add two httptest cases:

- `--description x` only → body has `description` and no `tags` key
- `--tag a=b` only → body has `tags` and no `description` key

The object-URI fail-closed case belongs with item 1.

---

### Nits

- `TestSetMetaValidation` does not cover the UTF-8 `MaxDescriptionLen` rejection (`setmeta.go:83-85`) or `--description=` empty vs `--clear-description` producing the same `""` body.
- The usage string is duplicated verbatim at `setmeta.go:75` and `:82`; hoist it so the next flag cannot drift the two copies.

---

### Recommendation
**Request changes.** Object-store paths panic because `CapWrite` is treated as “can setmeta”; that crash on a first-class CLI location kind should be fixed (with a fail-closed test) before merge.

qiffang (changes requested):
## Review summary

The client wire encoding for `setmeta` is directionally right: pointer fields preserve the distinction between omitted fields and explicit clear values, and the CLI flag parser handles the basic conflict cases.

This head is not mergeable yet.

### Blocking

1. Code CI fails on the SDK cookbook coverage gate

The hosted `Code CI` check fails on this exact head (`648123c148eb39897f522a518b5b17231686556b`). The PR adds exported `Client` methods, but `examples/go-sdk-cookbook/cookbook_test.go` requires every exported SDK method to have cookbook coverage:

```text
SDK cookbook missing Client method examples: SetMetadata, SetMetadataCtx
```

I reproduced the PR-specific failure locally with:

```sh
go test ./examples/go-sdk-cookbook -run TestClientMethodExampleCoverage -count=1
```

Please add examples and mark `SetMetadata` / `SetMetadataCtx` covered, or intentionally change the coverage policy.

2. `drive9 fs setmeta` panics on object-store URIs instead of failing closed

`SetMeta` accepts the generic fs path parser, checks only `CapWrite`, then dereferences `h.Client`:

- `cmd/drive9/cli/setmeta.go:87-107`
- object handles have no `Client`, and object backends advertise `CapWrite` but not `CapTags` / `CapDescription`: `cmd/drive9/cli/fsbackend.go:72-91`, `cmd/drive9/cli/fsbackend.go:643-648`

For an object URI that opens successfully, `drive9 fs setmeta --tag a=b s3://bucket/key` reaches `c.SetMetadata(...)` with `c == nil`. This should be a deterministic unsupported-path error, not a panic or accidental fallback to `/v1/fs/{object-key}`. Add a regression for object URI fail-closed behavior.

### Cross-PR Dependency

The paired server PR tidbcloud/fs#125 currently has a blocker in generated-tag collision handling. Even with this CLI fixed, `setmeta` will fail on a valid user update when the requested key already exists as a generated tag until the server path is corrected.

Local verification: `git diff --check`, gofmt check, `go test -c ./cmd/drive9 ./cmd/drive9/cli ./pkg/client`, and `go vet ./cmd/drive9 ./cmd/drive9/cli ./pkg/client` pass. `go test ./cmd/drive9 ./cmd/drive9/cli ./examples/go-sdk-cookbook -run 'Test(SetMeta|ClientMethodExampleCoverage|Dispatch)' -count=1` fails only on the cookbook coverage item above; `pkg/client` runtime tests are blocked locally by rootless Docker/TestMain.

srstack:
Review addressed in 77ce345:

**Blocking 1 — object-URI panic:** `SetMeta` now gates on `CapTags` / `CapDescription` per requested operation instead of `CapWrite`. Object backends advertise `CapWrite` but not the metadata caps, so `s3://bucket/key` fails closed with `backend "s3" does not support tags` (or `description`) before the nil `h.Client` is ever dereferenced. Regression: `TestSetMetaRejectsObjectURI` (uses `--auth=local` so the object backend opens without server credentials).

**Blocking 2 — cookbook gate:** added `SetMetadata` / `SetMetadataCtx` examples to `ExampleClient_filesystemCRUDAndMetadata` and marked both covered; `TestClientMethodExampleCoverage` passes.

**Should-fix — absence semantics:** `TestSetMetaOmitsUnsetFields` pins that a description-only update omits the `tags` key entirely and a tags-only update omits `description`, plus `--description=` emits `""`.

**Nits:** usage string hoisted to a const; `MaxDescriptionLen` rejection covered by `TestSetMetaDescriptionTooLong`.

Verified: `make lint` clean; `go test ./cmd/drive9/... ./pkg/client/ ./examples/...` clean.

mornyx (approved):
## Review summary

Round-2 head `77ce3452` addresses the previous request-changes items. `drive9 fs setmeta` now gates each requested operation on `CapTags` / `CapDescription` **before** `h.Client` is read, so object-store URIs fail closed instead of panicking. The SDK cookbook exemplifies and marks `SetMetadata` / `SetMetadataCtx`. The nil-vs-empty wire contract is pinned on JSON key absence, not just empty values.

Under a strict maintainability bar this is a small, consistent CLI+client addition: pointer `omitempty` on the wire, `parseAndMergeTag` reuse, and `fs stat` description output matching existing field conventions. Remaining notes are nits.

- Object-URI panic: `setmeta.go` requires `CapTags` when tags are written and `CapDescription` when description is written. `objectBackend.Caps()` still has `CapWrite` but neither metadata cap, so `s3://bucket/key` returns `backend "s3" does not support tags` (or `description`) and never reaches the nil client. Combined `--tag`+`--description` fails at the first gate. `TestSetMetaRejectsObjectURI` uses `--auth=local`.
- Cookbook CI gate: both methods are called in `ExampleClient_filesystemCRUDAndMetadata` and listed in `coveredClientMethods`.
- Absence-side wire tests: `TestSetMetaOmitsUnsetFields` asserts description-only omits `tags`, tags-only omits `description`, and `--description=` emits `""`.
- Nits from last round: usage string hoisted; `TestSetMetaDescriptionTooLong` covers `MaxDescriptionLen`.

**What looks good**
- Per-operation metadata caps are the right shape: same `requireCapOnHandle` path as `chmod`, no special-case `KindObject` branch, no fallback onto the original client (which would POST the object key into drive9).
- Pointer fields + `omitempty` actually distinguish leave-unchanged from clear, and tests now lock both sides of that contract.
- New client API lives in a small `pkg/client/setmeta.go` and reuses `url` / `do` / `readError` / `validateTags`.

---

### Nits

- `peelObjectAuth` still runs before flag parsing, so a spaced `--description` whose next token is `--auth=local` (or `--auth=server`) never becomes the description. The simple case is a confusing usage error; `--description --auth=local --clear-tags /a.txt` can succeed with description `"--clear-tags"` and skip the tag clear. This is the existing CLI-wide peel-first pattern (`cp` has the same `--description` hazard). `--description=--auth=local` works. Not worth a dedicated setmeta-only patch; a later shared parser fix would cover every fs command.
- The new `fs stat` `description:` text line has no fixture coverage. `TestStatDefaultOutputIncludesMetadataFields` is a two-line addition away from pinning print-when-non-empty.
- This repo’s `handleFS` still 400s `?setmeta=1`, and `handleStatMetadata` still has no `description` field. That matches the PR body’s pairing with tidbcloud/fs#125; both degrade safely here. Merge/deploy with the server change so help text does not advertise a command the current in-repo server cannot serve.

---

### Recommendation
**Approve.** Prior blockers are fixed on current HEAD with regression tests; only nits remain.

qiffang (approved):
Re-reviewed latest head `77ce3452c5b13293845d650bdf576b9cee0f3d81`.

The previous blockers are closed:

- `Client.SetMetadata` / `SetMetadataCtx` now have cookbook coverage; `TestClientMethodExampleCoverage` passes locally and hosted Code CI is green.
- `drive9 fs setmeta` now gates object URIs on metadata capabilities (`CapTags` / `CapDescription`) instead of `CapWrite`, so object backends fail closed instead of dereferencing a nil Drive9 client. `TestSetMetaRejectsObjectURI` covers both tags and description.

Local verification:

- `git diff --check`
- gofmt check on touched files
- `go test ./examples/go-sdk-cookbook -run TestClientMethodExampleCoverage -count=1`
- `go test ./cmd/drive9/cli -run 'TestSetMeta' -count=1`
- `go test ./cmd/drive9/cli -run 'Test(Stat|SetMeta|LsObjectURIDoesNotFallBackToLocalAWSEnv)' -count=1`
- `go test ./cmd/drive9/cli ./examples/go-sdk-cookbook -count=1`
- `go test -c ./cmd/drive9/cli ./pkg/client ./examples/go-sdk-cookbook`
- `go vet ./cmd/drive9/cli ./pkg/client ./examples/go-sdk-cookbook`

`pkg/client` runtime tests still cannot run in this local environment because its `TestMain` requires rootless Docker/testcontainers, but compile/vet and hosted checks are green.

srstack:
Server-side gap addressed in e258e8c: the setmeta endpoint is now ported into **this** repo's server (`pkg/server`), so `fs setmeta` no longer hits `400 unknown POST action` against a server built from this repository (Codex P1s). Adapted to this codebase's generation (no `file_tags.source` provenance → full tag replace consistent with upload-time semantics; single-statement writeback text guard instead of the lease-ownership tx). `GET ?stat=1` now returns `description` (Codex P1 on stat). Includes the dentry delete+recreate fence (409) and failpoint-gated race regressions; `pkg/server`, `pkg/backend`, `pkg/datastore` suites + failpoint suite + lint all green.

The remaining Copilot/Codex nit about `peelObjectAuth` consuming a description value that looks like `--auth=local` is a pre-existing quirk shared by `cp` and every other command that peels auth flags before parsing; deferring rather than changing the shared helper in this PR.

PR #893, opened 2026-09-07
feat: support append-log defaults in mount profiles

mashenjun:
FUSE users currently need explicit append-log flags to optimize SQLite WAL files. Add an `[append-log]` profile section and default `coding-agent` to `**/*-wal`, with parsing, display, and deduplicated merging of profile, environment, and flag rules.

Before mounting or detaching, check the existing backend `append_log_v1` capability. If unavailable or unconfirmed, ignore profile-derived rules and warn the caller while preserving explicit rules and existing FUSE capability guards. Supervised workers recheck without repeating the warning. Existing append-log files retain a conditional PUT fallback for path-based truncation even when profile rules have been filtered out.

Validation:

1. CLI, client, and pathfilter package tests passed; client tests used a temporary local TiDB instance.
2. `make lint` (0 issues), `make build-cli`, gofmt, and `git diff --check` passed. Profile export/reload was verified.
3. Added FUSE regression coverage for empty-match path truncation, size changes, conflicts, and bounded fallback. FUSE runtime validation is deferred to Linux/CI: local macOS FUSE regressions were explicitly excluded after an initial run hit disk free-space guards. The new truncation tests have not been executed.
4. Real directory-bucket backend E2E has not been run. The server contract was checked read-only in the fs repository; no server changes are included.

## Summary by CodeRabbit

* **New Features**
  * Added append-log pattern support to mount profiles, including a default for the coding-agent profile.
  * Append-log rules can be combined from profiles, environment settings, and command-line options.
  * Mounts now detect backend support before enabling append-log optimization.

* **Bug Fixes**
  * Unsupported backends now fall back to standard writes with a warning.
  * Improved handling of truncation and existing append-log files.

* **Documentation**
  * Documented append-log settings, defaults, capability checks, and rule precedence.

### DIFF
--- a/cmd/drive9/cli/mount.go
+++ b/cmd/drive9/cli/mount.go
@@ -61,6 +61,8 @@ type mountBackgroundRequest struct {
 
 var startMountBackground = startMountBackgroundImpl
 var startMountSupervisedBackground = startMountSupervisedBackgroundImpl
+var runSuperviseForeground = runSuperviseForegroundImpl
+var mountProfileAppendLogSupported = probeMountProfileAppendLogSupport
 
 // MountCmd handles the "drive9 mount" command.
 //
@@ -479,6 +481,10 @@ func fsMountCmdWithBackground(args []string, background bool) error {
 		}
 	}
 	*profile = profileCfg.Name
+	if err := validateMountPolicyPatterns(profileCfg.AppendLogPatterns); err != nil {
+		return err
+	}
+	effectiveAppendLogPatterns = mergeProfileValues(profileCfg.AppendLogPatterns, effectiveAppendLogPatterns)
 	effectiveLocalOnlyPatterns := mergeProfileValues(profileCfg.LocalOnlyPatterns, envLocalOnlyPatterns, localOnlyPatterns)
 	effectiveRemoteOnlyPatterns := mergeProfileValues(profileCfg.RemoteOnlyPatterns, envRemoteOnlyPatterns, remoteOnlyPatterns)
 	effectivePackPaths := mergeProfileValues(profileCfg.PackPaths)
@@ -592,6 +598,18 @@ func fsMountCmdWithBackground(args []string, background bool) error {
 	}
 	if err := validateMountProfileFlags(profileCfg.Name, normalizedLocalRoot, effectiveLocalOnlyPatterns, effectiveRemoteOnlyPatterns, effectivePackPaths); err != nil {
 		return err
+	}
+	if resolved == MountModeFUSE && runtime.GOOS != "windows" && len(profileCfg.AppendLogPatterns) > 0 {
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		supported := mountProfileAppendLogSupported(ctx, serverVal, apiKeyVal, tokenVal)
+		cancel()
+		if !supported {
+			// Keep explicit rules; FUSE retains its existing capability and layout guards.
+			effectiveAppendLogPatterns = mergeProfileValues(envAppendLogPatterns, appendLogPatterns)
+			if !*supervised {
+				fmt.Fprintf(os.Stderr, "drive9: warning: profile %q [append-log] ignored: backend append-log support is unavailable or could not be confirmed; using normal writes for profile-selected files\n", profileCfg.Name)
+			}
+		}
 	}
 
 	autoUnpack := overlayProfile && len(effectivePackPaths) > 0 && !*noAutoUnpack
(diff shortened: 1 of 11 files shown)


mornyx (commented):
## Review summary

This adds `[append-log]` to mount profiles (`coding-agent` defaults to `**/*-wal`), merges profile / env / flag rules with dedup, and fail-closes profile-derived rules when a bounded `append_log_v1` probe cannot confirm support. Explicit env/flag rules stay, supervised workers re-probe without repeating the warning, and FUSE reuses the existing conditional-PUT rewrite so an unmatched physical append-log layout remains truncatable. Under a strict maintainability bar the split is sound: CLI does not invent a second capability parser, and FUSE layout/capability guards stay authoritative. No merge-blocking correctness or security issues on current HEAD; one small should-fix remains so the explicit-rule set has a single source of truth.

**What looks good**
- Dropping unsupported profile rules by restoring env+flag rules (not subtracting from the combined list) keeps overlapping explicit patterns.
- The probe uses `client.New` / `NewWithToken` plus `Warm` + `CachedAppendLogSupported` — the same identity and cache as the mount client — with skip-on-empty-profile-rules and warning-before-detach coverage across start modes.
- `rewriteAppendLogPathTruncate` is a local extract reused as a one-shot fallback on `append_log_unsupported`, with revision/size/layout checks and no retry of the rejected generic upload.

---

**1. Bind explicit append-log rules once**

`cmd/drive9/cli/mount.go:277` and `:608` both compute `mergeProfileValues(envAppendLogPatterns, appendLogPatterns)` about 330 lines apart in the same function. The second copy is the definition of “rules that must survive capability filtering,” which is this PR’s core guarantee. If a later change adds a source or changes trim/dedup at only one site, the fallback silently diverges.

Bind once after the env is consumed:

```go
explicitAppendLogPatterns := mergeProfileValues(envAppendLogPatterns, appendLogPatterns)
effectiveAppendLogPatterns := explicitAppendLogPatterns
```

Then, on unsupported:

```go
effectiveAppendLogPatterns = explicitAppendLogPatterns
```

---

### Nits

- `probeMountProfileAppendLogSupport` (`cmd/drive9/cli/mount.go:1002`) is a fourth copy of `if token != "" { NewWithToken } else { New }`. A tiny `newMountClient(server, apiKey, token)` helper would keep credential-kind handling in one place.
- The probe gate (`:602`) still runs for read-only / `--checkpoint` mounts (`*readOnly` is already forced at `:575`). Skipping the probe (or at least the “using normal writes” warning) would avoid a confusing extra `/v1/status` on a restore view.
- The probe uses a magic `5*time.Second` while the same `/v1/status` warm path uses `fsClientWarmTimeout` (3s). Name `mountProfileAppendLogProbeTimeout` (or reuse the existing constant) so the documented five-second bound is not a bare literal.
- Profile `[append-log]` validation (`:484`) goes through `validateMountPolicyPatterns`, whose error still says `invalid local policy pattern %q` (`:1676`). Wrap with section context or generalize the wording now that this helper covers append-log profile rules.

---

### Recommendation
**Comment.** No merge blockers; please bind the explicit-rule set once so the capability fallback cannot drift from the initial merge.

mashenjun:
Addressed the explicit-rule source-of-truth recommendation in ca9173e9.

`explicitAppendLogPatterns` is now calculated once after consuming the environment, then used both for the initial effective rules and the unsupported-capability fallback. This keeps the explicit env/flag rule set from drifting between those paths.

Focused CLI append-log tests and `make lint` pass. I left the remaining items as non-blocking nits to keep this follow-up scoped to the requested refactor.

mornyx (approved):
## Review summary

Follow-up on `ca9173e9`: the previous should-fix is in. Env+flag append-log rules are bound once as `explicitAppendLogPatterns` and reused for both the initial effective set and the unsupported-capability fallback, with unchanged profile → env → flag merge order and fail-closed probing. A full re-check of current HEAD found no new correctness, security, or merge-blocking maintainability issues. Previously deferred nits stay optional.

**What looks good**
- `explicitAppendLogPatterns` is now the single source of truth for “rules that survive capability filtering.”
- Probe still uses `New` / `NewWithToken` plus `Warm` + `CachedAppendLogSupported`, drops only profile-derived rules, and keeps FUSE layout/capability guards.
- Path truncate of an unmatched physical append-log object still goes through one conditional PUT rewrite with no retry of the rejected generic upload.

- Bind explicit append-log rules once (`cmd/drive9/cli/mount.go:277` / `:609`). Verified on current HEAD.

---

### Nits

- The four items from the last review (`newMountClient`, skip probe on read-only/checkpoint, named 5s timeout, “invalid local policy pattern” wording) remain as previously deferred optional polish.
- Optional: compute the final effective rules after the probe instead of merge-then-restore across `fsMountCmdWithBackground`. Correct as written; not required for merge.

---

### Recommendation
**Approve.** The requested source-of-truth fix landed with unchanged semantics, and nothing remaining is merge-blocking.

PR #897, opened 2026-09-07
fix(fuse): rebind stale clean same-path SQLite read candidates

### LINKED ISSUE #894: 10MiB SQLite main DB reads stale after WAL checkpoint (same-path-dirty serves stale sibling handle)
# 10MiB SQLite main DB reads stale after WAL checkpoint (same-path-dirty serves a stale sibling handle)

## Summary

On `main` (CLI `2fd0c88`), with `--durability fsync --gvisor-compat=true --append-log='**/repro.db-wal'`, a fixed-size 10MiB SQLite database in WAL mode fails close-to-open/read-your-writes consistency: after the second `PRAGMA wal_checkpoint(TRUNCATE)` (200 acknowledged commits), the writer's own mount reads the **pre-checkpoint image** of the main DB (100-commit content) while the remote object already holds the correct 200-commit content.

This is a local read-view bug in the FUSE daemon, not remote data loss, gVisor, the CSI driver, or SQLite misuse.

- Severity: P1 (correctness / stale read; no data loss, remote remains correct)
- Area: FUSE read path / SQLite same-path handle coherence
- Fix location: `pkg/fuse/dat9fs.go`, `readSamePathDirtyHandleAllowJournalsOnce` (~20-35 production LoC)

## Affected version / environment

- Drive9 CLI: `2fd0c88` (`2fd0c88023e135f8b991d04a3cd95d5855db8463`), `git_branch=main`
- Server: `https://aws-us-east-1.drive9.ai`
- Cluster: `dev-drive9-eks-us-east-1`, gVisor (`directfs`), CSI image `drive9-2fd0c88-csi-4ec9118`
- Mount (writer):

```text
drive9 mount --supervise-foreground --mode=fuse --direct-mount-strict
  --gvisor-compat=true --server https://aws-us-east-1.drive9.ai --allow-other
  --durability fsync --profile coding-agent
  --append-log=**/repro.db-wal
  --writeback-batch-window 20ms :/ <target>
```

## Repro

1. Pre-create a **remote** 10MiB SQLite DB (`repro.db`, 2560 pages of 4096 bytes, 2552 rows; row 1 holds a 32-char hex counter, other rows hold deterministic 3000-byte payloads). Do not create it through the mount.
2. Open it on the mount, `PRAGMA journal_mode=WAL`, `PRAGMA synchronous=FULL`, `PRAGMA wal_autocheckpoint=0`.
3. Update only `items.id=1` to `{i:032x}` for `i=1..200`, committing each row.
4. Every 100 commits run `PRAGMA wal_checkpoint(TRUNCATE)`, then immediately verify the **main DB file bytes** with `sqlite3 'file:<db>?mode=ro&immutable=1'` (WAL intentionally ignored).
5. After checkpoint01 (100) the read passes. After checkpoint02 (200) the read returns the checkpoint01 image.

Expected: checkpoint02 read shows `value = '000000000000000000000000000000c8'` (200).
Actual: checkpoint02 read shows `value = '00000000000000000000000000000064'` (100) with fingerprint identical to checkpoint01.

This reproduces 3/3 times on the same build.

## Key evidence

Data fingerprints (SHA-256 over `(id, payload, value)` of all rows read via `immutable=1`):

| Phase | updates | pass | fingerprint |
| --- | --- | --- | --- |
| seed | 0 | true | `47d9de5b01b2e4f04929651ba41c88b1b858088907121422fdacbeec147eb259` |
| checkpoint-01 | 100 | true | `a61365ec53baa2eec09457fe6f5f53883ccfd2cf7f166ee98ddd7fd7f368c122` |
| checkpoint-02 | 200 | **false** | `a61365ec53baa2eec09457fe6f5f53883ccfd2cf7f166ee98ddd7fd7f368c122` |
| remote after writer exit | 200 | true | `45998cccfe0b81a64ccd31b0bb5c2572b6185f1ef7022accc6c137554e6a71ad` |
| fresh mount | 200 | true | row1 = `000000000000000000000000000000c8` |

Perf deltas around the two checkpoints (10s `perf.jsonl` samples):

| window | fuse_ops.read | remote_ops.read |
| --- | --- | --- |
| checkpoint-01 + inspect01 | +6731 | **+2** (part loads) |
| checkpoint-02 + inspect02 | +6718 | **+0** (fully local) |

The checkpoint02 read served ~6700 reads with **zero** remote access, while checkpoint01 needed 2 remote part loads. `read_cache_hit` and `prefetch_hit` were unchanged in both windows, so the serving source was neither ReadCache nor prefetch.

Daemon debug timeline (`dat9 debug:` from the mount log; `expected_rev=N` is the CAS base, committed revision is `N+1`):

```text
10:19:20 open repro.db fh=4  flags=0x28002(O_RDWR) base_rev=1          # writer's SQLite connection
10:19:47 main_db_write offset=8192 source=dirty-buffer                  # checkpoint01 writes page 3
10:19:47 open repro.db fh=12 flags=0x28001(O_WRONLY) base_rev=2        # SQLite checkpoint sidecar handle
10:19:47 fsync fh=4 -> flushHandle patch dirty_parts=1 expected_rev=2   # -> committed rev 3 (100)
10:19:48 open repro.db fh=14 flags=0x28000(O_RDONLY) base_rev=3        # inspect01 clean read handle
10:19:48 dirty load part 1 (8388608) + part 2 (2097152)                # fh=12 buffer loaded from remote (100)
10:19:48 read fh=14 ... source=same-path-dirty                         # inspect01 served via fh=12 buffer
10:20:18 main_db_write offset=8192 source=dirty-buffer                  # checkpoint02 writes page 3 (200)
10:20:18 fsync fh=4 -> flushHandle patch dirty_parts=1 expected_rev=3   # -> committed rev 4 (200)
10:20:18 read fh=14 ... source=same-path-dirty                         # inspect02: no remote load, stale 100
10:20:20 flush fh=12 expected_rev=2 -> 14us no-op (superseded by rev 4) # proves fh=12 pinned rev-2/100 view
```

## Root cause

`Read` routes a clean read-only handle for a SQLite main DB through the same-path dirty path:

```text
pkg/fuse/dat9fs.go:11446 Read
  -> fh.Dirty == nil && isSQLiteVisibleSamePathDirtyPath(fh.Path)
  -> readSamePathDirtyHandle(fh.Path, fh, ...)     source="same-path-dirty"
  -> readSamePathDirtyHandleAllowJournalsOnce       pkg/fuse/dat9fs.go:4628
```

`readSamePathDirtyHandleAllowJournalsOnce` selects any same-path open handle with `DirtySeq > 0` and serves reads from that handle's `WriteBuffer` (`src.Dirty.ReadAt` after `EnsureLoaded`). It does **not** check whether that buffer was loaded at an older committed revision.

In this workload SQLite opens a second O_WRONLY main-DB handle (`fh=12`, `base_rev=2`) around the first checkpoint and keeps it open. Its lazy buffer is populated during the first inspect (remote reads at rev 3 = 100-commit content). When the real writer handle (`fh=4`) later commits rev 4 (200-commit content), sibling `fh=12` is neither invalidated nor rebound: its `BaseRev` stays 2 and its already-loaded parts stay at the 100-commit image. The second inspect's clean reads then hit `fh=12` again and serve the stale bytes with zero remote reads.

This is the same "linked SQLite handle must be kept current" class that `2fd0c88` fixed for the `-wal` sidecar zero-truncate (`isSQLitePersistentJournalPath`), but the main-DB sibling-handle read-through path is unfixed.

Why it is not visible in the small/empty-DB cases: those stay below `writeBackThreshold` and are served through the revision-checked whole-file/ReadCache path, and normal WAL-aware reads replay the WAL, masking main-DB staleness.

## Fix plan (recommended)

In `readSamePathDirtyHandleAllowJournalsOnce`, rebind a stale clean sibling buffer before serving it:

```go
// A same-path writable handle can remain open with an already-loaded buffer
// from an older committed revision. Rebind clean (no uncommitted dirty parts)
// candidates so the subsequent EnsureLoaded fetches at the current revision.
if !src.Dirty.HasDirtyParts() {
    fs.adoptCommittedRevisionLocked(src)
}
```

Place this inside the candidate loop after the `src.DirtySeq != c.dirtySeq` guard and before reading `src.Dirty`. `adoptCommittedRevisionLocked` (`pkg/fuse/dat9fs.go:2967`) already handles the stale case: when `latestCommittedRevision(path) > src.BaseRev` it calls `rebindCleanWriteBufferToRemoteLocked`, which drops the old parts so the next `EnsureLoaded` re-fetches at the new revision. The `!HasDirtyParts()` guard ensures uncommitted writes are never rebased away. The caller already holds `src.mu` as required.

Alternative (smaller, lower risk, loses the local fast path): skip clean candidates whose `BaseRev < latestCommittedRevision` so the read falls back to shadowStore/remote.

Production Net LoC: ~20-35 (`size/S`). No schema, API, or lock-model changes.

## Test plan

1. Add a `pkg/fuse` unit test: two same-path writable handles on a multi-part file; handle A writes and fsyncs a new revision; a clean same-path read must return A's content, not sibling handle B's stale buffer.
2. `go test ./pkg/fuse` and `go test ./...` pass; `gofmt`/`git diff --check` clean.
3. Re-run the 10MiB case above: checkpoint-02 fingerprint must equal the 200-update fingerprint `45998cccfe0b81a64ccd31b0bb5c2572b6185f1ef7022accc6c137554e6a71ad` and `pass=true`.

## Non-goals

- No CSI driver change.
- No change to the append-log `-wal` sidecar path unless a test shows the same bug there (file as follow-up).
- No remote/server or protocol change.

## Definition of done

- Root-cause regression test added and green.
- The 10MiB durability case passes checkpoint02 and reads 200 updates in the writer's own mount.

mashenjun:
Fixes https://github.com/mem9-ai/drive9/issues/894

## Summary
A clean SQLite main-DB read on the same mount can be served from an older sibling WriteBuffer after a later fsync-committed WAL checkpoint, producing a stale read.

## Root cause
readSamePathDirtyHandleAllowJournalsOnce selects any same-path handle with DirtySeq > 0 and serves its loaded WriteBuffer without checking whether that buffer is older than the latest committed revision. A long-lived checkpoint sidecar handle keeps the pre-checkpoint image loaded and is selected for subsequent reads.

## Change
Add a read-path-specific rebindStaleCleanSamePathReadCandidateLocked helper. For a clean candidate whose BaseRev is below the latest committed revision, it rebinds the buffer to the new revision while keeping DirtySeq unchanged so the local same-path fast path remains usable. Unsafe candidates (new, zero-base, unlinked, shadow-backed, streamed, or pending-staged) are skipped and fall back to shadowStore/remote.

## Scope
- pkg/fuse/dat9fs.go only (47 production lines added)
- No server, CSI, schema, lock-model, or gVisor change

## Tests
- TestReadSQLiteSamePathDirtyRebindsStaleCleanSibling: three reads over two parts assert two remote part loads and local buffer reuse.
- TestReadSQLiteSamePathDirtySkipsUnsafeStaleSibling: unsafe candidate is skipped and the read falls back to remote.

Tests were not executed locally on macOS. gofmt, git diff --check, go vet ./pkg/fuse, and go build ./cmd/drive9 ./pkg/fuse pass. Unified regression/race and the 10MiB checkpoint02 e2e run are planned after development.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved same-path reads to refresh stale, clean file data after newer changes are committed.
  * Safely skips stale file handles that cannot be refreshed and continues with an available fallback.
  * Prevented unnecessary repeated remote fetches when reading refreshed data.

### DIFF
--- a/pkg/fuse/dat9fs.go
+++ b/pkg/fuse/dat9fs.go
@@ -4632,6 +4632,46 @@ func (fs *Dat9FS) readSQLitePersistentJournalVisibleRange(path string, skip *Fil
 	return nil, 0, false, gofuse.OK, ""
 }
 
+// rebindStaleCleanSamePathReadCandidateLocked refreshes a clean same-path read
+// candidate whose loaded WriteBuffer is older than the latest committed
+// revision. It keeps DirtySeq unchanged so the candidate remains eligible for
+// the local same-path fast path, and returns false when the candidate is stale
+// but unsafe to rebind so the caller can skip it and fall back to
+// shadowStore/remote.
+func (fs *Dat9FS) rebindStaleCleanSamePathReadCandidateLocked(fh *FileHandle) bool {
+	if fs == nil || fh == nil || fh.Dirty == nil || fh.Dirty.HasDirtyParts() {
+		return false
+	}
+	revision, committedSize, hasSize := fs.latestCommittedRevisionWithSize(fh.Path)
+	if revision <= 0 || revision <= fh.BaseRev {
+		return true
+	}
+	if fh.WriteBackSeq != 0 || fh.ShadowCommitReady || fh.ShadowCommitSeq != 0 ||
+		fh.IsNew || fh.ZeroBase || fh.Unlinked || fh.UnlinkedSnapshot ||
+		fh.UnlinkedData != nil || fh.ShadowReady || fh.ShadowSpill ||
+		fh.Streamer != nil {
+		return false
+	}
+	if fs.pendingIndex != nil && fh.PendingIndexGen != 0 &&
+		fs.pendingIndex.Generation(fh.Path) == fh.PendingIndexGen {
+		return false
+	}
+	if fs.shadowStore != nil && fh.ShadowStageGen != 0 &&
+		fs.shadowStore.ActiveGeneration(fh.Path) == fh.ShadowStageGen {
+		return false
+	}
+	if !hasSize {
+		committedSize = fs.committedHandleSizeLocked(fh)
+	}
+	fh.BaseRev = revision
+	if hasSize && fh.Dirty.Size() >= committedSize {
+		fh.OrigSize = committedSize
+		fh.appendLogRebindLayout(revision, committedSize)
+	}
+	fs.rebindCleanWriteBufferToRemoteLocked(fh, committedSize)
+	return true
+}
+
 func (fs *Dat9FS) readSamePathDirtyHandleAllowJournalsOnce(path string, skip *FileHandle, offset int64, reqSize uint32) ([]byte, int, bool, gofuse.Status, bool) {
 	if fs == nil || fs.openHandles == nil || path == "" || reqSize == 0 {
 		return nil, 0, false, gofuse.OK, false
@@ -4683,6 +4723,13 @@ restartLoop:
 				src.Unlock()
 				pending = true
 				continue restartLoop
+			}
+			if !src.Dirty.HasDirtyParts() &&
+				(fs.shadowStore == nil || (!src.ShadowReady && !src.ShadowSpill)) {
+				if !fs.rebindStaleCleanSamePathReadCandidateLocked(src) {
+					src.Unlock()
+					continue
+				}
 			}
 			size := src.Dirty.Size()
 			if offset >= size {
--- a/pkg/fuse/dat9fs_test.go
+++ b/pkg/fuse/dat9fs_test.go
@@ -12837,6 +12837,149 @@ func TestReadSQLiteSamePathDirtyHandleSkipsIncompleteCandidate(t *testing.T) {
 	}
 }
 
+func TestReadSQLiteSamePathDirtyRebindsStaleCleanSibling(t *testing.T) {
+	const (
+		filePath = "/workload.db"
+		partSize = int64(4096)
+		fileSize = 2 * partSize
+	)
+	oldData := bytes.Repeat([]byte{'A'}, int(fileSize))
+	newData := bytes.Repeat([]byte{'B'}, int(fileSize))
+	var remoteGets atomic.Int64
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.Method != http.MethodGet || r.URL.Path != "/v1/fs"+filePath {
+			http.NotFound(w, r)
+			return
+		}
+		remoteGets.Add(1)
+		w.Header().Set("Content-Length", strconv.FormatInt(fileSize, 10))
+		w.WriteHeader(http.StatusOK)
+		_, _ = w.Write(newData)
+	}))
+	defer ts.Close()
+
+	opts := &MountOptions{TrustLocalEvents: true}
+	opts.setDefaults()
+	fs := NewDat9FS(newTestClient(ts.URL), opts)
+	ino := fs.inodes.Lookup(filePath, false, fileSize, time.Now())
+	fs.inodes.UpdateRevision(ino, 2)
+
+	stale := &FileHandle{
+		Ino:      ino,
+		Path:     filePath,
+		BaseRev:  2,
+		OrigSize: fileSize,
+		Dirty:    fs.newWriteBuffer(filePath, maxPreloadSize, partSize),
+	}
+	if _, err := stale.Dirty.Write(0, oldData); err != nil {
+		t.Fatalf("seed stale buffer: %v", err)
+	}
+	stale.Dirty.ClearDirty()
+	stale.DirtySeq = fs.markDirtySize(ino, fileSize)
+	fs.openHandles.Add(stale)
+	defer fs.openHandles.Remove(stale)
+
+	fs.recordCommittedRevisionWithSize(filePath, 3, fileSize)
+	fs.inodes.UpdateRevision(ino, 3)
+	fs.inodes.UpdateSize(ino, fileSize)
+
+	readPart := func(offset int64) []byte {
+		t.Helper()
+		got, n, ok, st := fs.readSamePathDirtyHandle(filePath, nil, offset, uint32(partSize))
+		if !ok || st != gofuse.OK {
+			t.Fatalf("read offset=%d ok=%t status=%v, want true/OK", offset, ok, st)
+		}
+		if n != int(partSize) {
+			t.Fatalf("read offset=%d bytes=%d, want %d", offset, n, partSize)
+		}
+		return append([]byte(nil), got[:n]...)
+	}
+	if got := readPart(0); !bytes.Equal(got, newData[:partSize]) {
+		t.Fatalf("first read of part0 = %q, want newData prefix", got)
+	}
+	if got := readPart(0); !bytes.Equal(got, newData[:partSize]) {
+		t.Fatalf("second read of part0 = %q, want newData prefix", got)
+	}
+	if got := readPart(partSize); !bytes.Equal(got, newData[partSize:]) {
+		t.Fatalf("read of part1 = %q, want newData suffix", got)
+	}
+	if got := remoteGets.Load(); got != 2 {
+		t.Fatalf("remote gets = %d, want 2: one per distinct part, not one per FUSE read", got)
+	}
+	if got := stale.BaseRev; got != 3 {
+		t.Fatalf("stale sibling BaseRev = %d, want 3 after rebind", got)
+	}
+}
+
+func TestReadSQLiteSamePathDirtySkipsUnsafeStaleSibling(t *testing.T) {
+	const (
+		filePath = "/workload.db"
+		partSize = int64(4096)
+		fileSize = 2 * partSize
+	)
+	oldData := bytes.Repeat([]byte{'A'}, int(fileSize))
+	newData := bytes.Repeat([]byte{'B'}, int(fileSize))
+	var remoteGets atomic.Int64
+	ts := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		if r.Method != http.MethodGet || r.URL.Path != "/v1/fs"+filePath {
+			http.NotFound(w, r)
+			return
+		}
+		remoteGets.Add(1)
+		w.Header().Set("Content-Length", strconv.FormatInt(fileSize, 10))
+		w.WriteHeader(http.StatusOK)
+		_, _ = w.Write(newData)
+	}))
+	defer ts.Close()
+
+	opts := &MountOptions{TrustLocalEvents: true}
+	opts.setDefaults()
+	fs := NewDat9FS(newTestClient(ts.URL), opts)
+	ino := fs.inodes.Lookup(filePath, false, fileSize, time.Now())
+	fs.inodes.UpdateRevision(ino, 2)
+
+	stale := &FileHandle{
+		Ino:      ino,
+		Path:     filePath,
+		BaseRev:  0,
+		OrigSize: fileSize,
+		IsNew:    true,
+		Dirty:    fs.newWriteBuffer(filePath, maxPreloadSize, partSize),
+	}
+	if _, err := stale.Dirty.Write(0, oldData); err != nil {
+		t.Fatalf("seed stale buffer: %v", err)
+	}
+	stale.Dirty.ClearDirty()
+	stale.DirtySeq = fs.markDirtySize(ino, fileSize)
+	fs.openHandles.Add(stale)
+	defer fs.openHandles.Remove(stale)
+
+	fs.recordCommittedRevisionWithSize(filePath, 3, fileSize)
+	fs.inodes.UpdateRevision(ino, 3)
+	fs.inodes.UpdateSize(ino, fileSize)
+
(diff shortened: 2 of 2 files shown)


qiffang (approved):

qiffang (approved):
No blocking findings on exact head `40050162e5abc91a97ae06ca8b2905b6a76a7524`.

I re-reviewed the issue #894 path and the adjacent same-path dirty-handle reads. The new guard only rebinds clean, stale same-path SQLite main DB candidates when the handle has no dirty parts and no unsafe writeback/shadow/unlink/stage state, and otherwise skips the candidate so the normal remote read path can return the committed data. That matches the failure mode: a clean but stale writable sibling buffer should not be allowed to satisfy later same-path reads after a newer local commit.

The new regressions are discriminating: one test would return old buffered bytes and avoid remote GETs on the previous code; the unsafe-candidate test proves stale new/zero-base candidates are not rebound and fall back to the committed remote object. I also checked the persistent-journal sidecar path separately; it remains covered by its existing dirty-range behavior and is outside #894's stated non-goal.

Validation: GitHub Code CI, Local E2E, and CodeRabbit are green on this head. Locally I ran `git diff --check`, gofmt check, `go test -c ./pkg/fuse`, `go vet ./pkg/fuse`, focused same-path SQLite read tests including `-count=10` for the new regressions, and focused SQLite persistent-journal tests. Full `go test ./pkg/fuse -count=1` still hits the known local Darwin/rootless mount-permission environment tests only; I did not see a PR-related failure.

qiffang (approved):
Reviewed exact head `40050162e5abc91a97ae06ca8b2905b6a76a7524`.

No blocking finding. The stale clean same-path writable sibling is now rebound only after the committed revision advances, while dirty/writeback/shadow/unlink/staging states remain excluded and fall back to the normal remote path. I independently exercised the actual route with two real SQLite checkpoints: a stale clean sibling was held across a newer handle's real `Fsync`, then a top-level read returned checkpoint B. The test passed 20x normally and 20x under `-race`; removing only the new rebind call deterministically returned checkpoint A, so the fix is discriminating.

Exact-head Code CI, Local E2E, and CodeRabbit are green. Local focused/broader same-path tests, compile, vet, lint, gofmt, and diff-check pass; full `pkg/fuse` only hits the two known macOS mount-permission environment failures.

Non-blocking evidence gap: the committed unit test manually publishes the committed watermark and calls the same-path helper rather than driving the full sibling `Fsync` + top-level `Read` route. The stronger deterministic regression above would be worth upstreaming. The issue DoD's real 10 MiB checkpoint02 / 200-update E2E evidence is also still absent, so attach that evidence before closing the issue or making the release claim.

mornyx (commented):
## Review summary

This PR fixes a real stale-read hole (issue #894): a clean same-path sibling with `DirtySeq > 0` — the SQLite checkpoint sidecar shape — is rejected by every commit-time refresh path, so `readSamePathDirtyHandleAllowJournalsOnce` could keep serving its pre-checkpoint WriteBuffer. Rebinding such candidates onto the latest committed revision at read time, preserving `DirtySeq`, and falling back to shadowStore/remote for unsafe candidates is the right behavior, and I could not find a correctness or locking defect in the mechanism: the rebind runs under the candidate's lock with sound lock ordering, no missed unlocks, and the guard list is complete against today's `FileHandle` field set. Both new tests are deterministic and assert the mechanism (per-part remote-load counts, `BaseRev` advance, real `Read` fallback), and CI is green on the head commit. The merge-relevant concerns are structural and coverage-related, graded below under a strict maintainability bar — none are merge-blocking.

**What looks good**

- Fixing at the candidate-consumption point with an "unsafe → skip → remote fallback" contract, instead of widening the write-side adopt predicate whose `DirtySeq != 0` refusal exists for a real silent-rollback reason.
- `DirtySeq` retention is load-bearing and correct: it keeps the local same-path fast path eligible, and the rebind also eliminates a pre-existing mixed-revision hazard (stale loaded parts + fresh lazily-loaded parts) by discarding all loaded parts.
- The rebind composes existing canonical helpers (`rebindCleanWriteBufferToRemoteLocked`, `appendLogRebindLayout`, `committedHandleSizeLocked`) and mirrors the adopt path's conditional layout rebind rather than inventing new revision math.
- Tests assert mechanism, not just output: `remoteGets == 2` proves per-part caching after rebind (and implicitly `DirtySeq` preservation), and the unsafe-sibling test drives a real `Read` fallback that must hit remote.

---

**1. The unsafe-candidate guard is a second source of truth for "safe to rebase onto a committed revision" — extract a shared predicate**

The guard block at `pkg/fuse/dat9fs.go:4649-4662` re-implements, nearly verbatim, the staged-state predicate already owned by `handleCanAdoptCommittedRevisionLocked` (`WriteBackSeq != 0 || ShadowCommitReady || ShadowCommitSeq != 0`, plus the `PendingIndexGen` and `ShadowStageGen` generation-equality checks), then layers read-specific refusals (`IsNew`/`ZeroBase`/`Unlinked*`/`ShadowReady/Spill`/`Streamer`) on top. Counting adopt, the two `refreshCommittedRevisionForOpenHandles` variants, `refreshCleanCommittedRevisionForHandleLocked`, and this helper, the "set BaseRev (+conditional OrigSize/appendLogRebindLayout) + rebindCleanWriteBufferToRemoteLocked" sequence now exists in about five near-identical shapes — and the copies have already drifted (only `adoptCommittedRevisionLocked` and this new helper call `appendLogRebindLayout`). The next in-flight-commit flag added to `handleCanAdoptCommittedRevisionLocked` will silently not protect this read path, resurrecting exactly the stale-read class this PR fixes.

Suggestion: split the staged-state predicate out — e.g. `handleHasLiveStagedStateLocked(fh)` covering the WriteBack/ShadowCommit flags and both generation checks — shared by `handleCanAdoptCommittedRevisionLocked` and this helper, leaving genuinely divergent policy (`DirtySeq` retention here vs. rejection there, the read-specific refusals) at the call sites. Ideally also collapse the repeated BaseRev/rebind sequence into one shared rebind core.

**2. Guard-matrix test coverage: only one of ~10 unsafe conditions is exercised**

The safety of the whole change rests on the completeness of the refusal list, but the skip-and-fallback test covers only `IsNew`. `WriteBackSeq`, `ShadowCommitReady/Seq`, `ZeroBase`, the `Unlinked*` flags, `ShadowReady/Spill`, `Streamer`, and the two pending-staged generation checks have no coverage; a future regression dropping any one flag would let a stale image be rebound and served with nothing catching it. The `hasSize == false` size-fallback path and the remote-load-failure-after-successful-rebind path (EnsureLoaded error → pending → fallback, sibling left rebound) are also untested.

Suggestion: make the skip test table-driven — one stale clean sibling per unsafe condition, asserting skip + remote fallback for each — and add one case for the `!hasSize` fallback and one for a failing `LoadPart` after rebind.

**3. After a successful rebind, the remote part fill runs while holding the sibling's FileHandle lock**

The rebind drops every loaded part, so the existing `EnsureLoaded` loop that follows does `LoadPart` → `acquireRemoteReadSlot` → HTTP GET while still holding the candidate's lock (`TryLock` at the top of the candidate iteration, unlock after `ReadAt`). Unlock discipline itself is intact and this is not a new lock-ordering hazard — the `EnsureLoaded`-under-lock shape pre-exists — but this PR makes it the common path: every successful stale-clean rebind now stalls same-handle writers for a network RTT, and the commit-time refresh (TryLock-only by design) will skip the handle for the whole fetch.

Suggestion: either unlock before the post-rebind `EnsureLoaded` and revalidate `DirtySeq`/`BaseRev`/buffer identity on re-lock, or skip the stale candidate and let the existing fallback serve this read while a shared refresh (finding 1) repairs the handle without holding it across the GET. If the current shape is kept deliberately, a short comment stating the trade-off would help.

### Nits

- `pkg/fuse/dat9fs.go:4667` — the conditional `fh.OrigSize = committedSize` store is dead: `rebindCleanWriteBufferToRemoteLocked` overwrites `OrigSize` unconditionally on the next line. The block's only live effect is `appendLogRebindLayout`; drop the store and add a one-line comment that the conditional layout rebind mirrors the adopt path (intentionally leaving layout stale, fail-safe, when the committed size is unknown or the file grew).
- `pkg/fuse/dat9fs.go:4663` — when the committed-size map has no entry, the fallback to the inode size can lead the actual remote object during an in-flight same-inode write; combined with zero-filled short reads this can serve a zero-filled tail until the writer commits. Reachability is narrow (a live dirty writer normally wins candidate selection first), but please make a deliberate choice — skip the rebind when `!hasSize`, or document why an inode-size-led size is acceptable here.
- `pkg/fuse/dat9fs.go:4495` — scope completeness: the twin persistent-journal candidate loop (`readSQLitePersistentJournalDirtyRangeOnce`) has no stale-revision check. Reachability of the #894 shape there looks much lower, but please confirm or note in the PR why it is out of scope so the next reader doesn't assume same-path stale candidates are handled everywhere.
- Scope is broader than the PR body states: the rebind also runs on the generic close-sync/write-sync same-path read path (`dat9fs.go:11917-11923`), for any file type — which is correct and beneficial, but currently undocumented and untested for that wiring. Worth a sentence in the PR body or one close-sync test mirroring the existing `TestReadCloseSyncSamePathDirtyWriter*` pattern.

---

### Recommendation

**Comment.** The fix is correct and well-tested for what it asserts, with CI green on head — but the duplicated guard predicate (drift risk on the load-bearing safety mechanism) and the thin guard-matrix coverage are strong should-fix items that shouldn't be waved through as nits.

Issue #895, opened 2026-09-07
FUSE: ReadStreamRange omits first-hop Range, stalling large WAL checkpoints


qiffang:
#### Scope: when can an SQLite WAL legitimately reach 40 MiB?

40 MiB is a reproduction threshold, not an SQLite WAL limit. With a 4 KiB
database page, each WAL frame is a 24-byte frame header plus 4,096 bytes, and
the WAL itself has a 32-byte header. Therefore the first WAL size at or above
40 MiB is:

```text
32 + 10,181 * (24 + 4,096) = 41,945,752 bytes
```

SQLite normally starts a PASSIVE auto-checkpoint at 1,000 pages (about 3.93
MiB at this page size), but that threshold starts work; it is not a hard size
cap. A WAL can exceed 40 MiB when any of these occurs:

1. `wal_autocheckpoint` is disabled or raised, as in applications that move
   checkpoint work to an idle/background window.
2. Long or continuously overlapping read transactions prevent checkpoints
   from completing and resetting the WAL.
3. Writes generate dirty pages faster than a checkpointer can apply them.
4. One large write transaction itself generates more than 40 MiB before the
   post-commit checkpoint can run.
5. Separately, the physical WAL allocation can remain large after a successful
   checkpoint because SQLite normally rewinds/reuses it instead of truncating
   it. That is not what happened in this fresh-path reproduction: all 12,496
   frames were written before the first checkpoint/reset.

See SQLite's documentation on [automatic checkpointing and concurrency](https://www.sqlite.org/wal.html#checkpointing)
and [avoiding excessively large WAL files](https://www.sqlite.org/wal.html#avoiding_excessively_large_wal_files).

#### Confirmed environment

| Component | Value |
|---|---|
| Drive9 server | `https://aws-us-east-1.drive9.ai/` |
| Reproduced client | `2fd0c88023e135f8b991d04a3cd95d5855db8463` |
| Current `main` inspected | `82bb5945565c8fba68d57e60be9ca1d25095470c` |
| Host/runtime | Same-region native Linux EC2, x86_64, direct FUSE mount |
| SQLite | 3.46.1, 4 KiB pages |
| SQLite settings | WAL, `synchronous=FULL`, `wal_autocheckpoint=0`, `mmap_size=0` |
| Drive9 settings | `profile=coding-agent`, `durability=fsync` (strict sync + writeback), append-log enabled, `remote-only=**/tmp/**` |
[...]

PR #899, opened 2026-09-07
fix(client): send first-hop Range on inline range reads (#895)

mashenjun:
Issue: #895

## Problem

`ReadStreamRange` sent the Range header only on the 302/307 second hop.
Append-log WAL files are served inline on the first hop as 200, so every
nonzero-offset page read streamed the body from byte zero and discarded the
prefix in `sliceBody`. A 49 MiB WAL checkpoint turned into tens of GiB of
transfer and exceeded SQLite's 300 s busy timeout.

## Change

- Send `Range: bytes=start-end` on the initial `/v1/fs` GET.
- First-hop 206: validate Content-Range start and bound with `LimitReader`
  (no double slicing).
- First-hop 416: treat as EOF.
- First-hop 200: keep `sliceBody` fallback for small inline files and old
  servers.
- 302/307 redirect path unchanged.

The fs-repo server already honors Range on append-log reads; this fixes the
client-side omission only.

## Tests

- pkg/client: first-hop 206 at offset 0 and nonzero offset, 416, 200
  fallback, bad Content-Range, bounded-transfer byte count.
- pkg/fuse: nonzero-offset 4 KiB WAL read carries Range and the fake server
  writes exactly 4 KiB.

## Deferred

Full regression, FUSE e2e, and the hosted >40 MiB WAL checkpoint acceptance
run (issue criteria 3-6) are executed separately.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved ranged file reads by requesting only the requested byte range in the initial server response.
  * Correctly handles partial-content responses and treats unsatisfiable ranges as empty reads.
  * Safely handles servers that ignore range requests while returning only the requested data.
* **Tests**
  * Added coverage for ranged reads, response validation, fallback behavior, and bounded data transfers.

### DIFF
--- a/pkg/client/transfer.go
+++ b/pkg/client/transfer.go
@@ -1515,13 +1515,15 @@ func (c *Client) ReadStreamRange(ctx context.Context, path string, offset, lengt
 		}
 
 	case resp.StatusCode == http.StatusPartialContent:
-		if err := validatePartialContentRange(resp, offset); err != nil {
+		end, err := validatePartialContentRange(resp, offset, offset+length-1)
+		if err != nil {
 			_ = resp.Body.Close()
 			return nil, fmt.Errorf("read %s: %w", path, err)
 		}
 		// First-hop Range honored. The body already starts at offset, so
-		// bound it to length without skipping the offset a second time.
-		return &limitedReadCloser{r: io.LimitReader(resp.Body, length), c: resp.Body}, nil
+		// bound it to the declared response range without skipping the
+		// offset a second time.
+		return &limitedReadCloser{r: io.LimitReader(resp.Body, end-offset+1), c: resp.Body}, nil
 
 	case resp.StatusCode == http.StatusRequestedRangeNotSatisfiable:
 		_ = resp.Body.Close()
@@ -1601,43 +1603,53 @@ func (c *Client) sliceBody(rc io.ReadCloser, offset, length int64) (io.ReadClose
 }
 
 // validatePartialContentRange checks that a first-hop 206 response carries a
-// well-formed Content-Range starting at the requested offset. The caller
-// bounds the length via io.LimitReader, so a server-clamped end (EOF) is
-// allowed here.
-func validatePartialContentRange(resp *http.Response, offset int64) error {
+// well-formed Content-Range starting at the requested offset and not extending
+// past the requested end. It returns the declared response end so the caller
+// can bound the body to the declared range. A server-clamped end (EOF) is
+// allowed.
+func validatePartialContentRange(resp *http.Response, offset, requestedEnd int64) (int64, error) {
 	contentRange := resp.Header.Get("Content-Range")
 	if contentRange == "" {
-		return fmt.Errorf("206 response missing Content-Range for offset %d", offset)
+		return 0, fmt.Errorf("206 response missing Content-Range for offset %d", offset)
 	}
 	rest, ok := strings.CutPrefix(contentRange, "bytes ")
 	if !ok {
-		return fmt.Errorf("invalid Content-Range %q for offset %d", contentRange, offset)
+		return 0, fmt.Errorf("invalid Content-Range %q for offset %d", contentRange, offset)
 	}
 	byteRange, complete, ok := strings.Cut(rest, "/")
 	if !ok {
-		return fmt.Errorf("invalid Content-Range %q for offset %d", contentRange, offset)
+		return 0, fmt.Errorf("invalid Content-Range %q for offset %d", contentRange, offset)
 	}
+	completeSize := int64(-1)
 	if complete != "*" {
-		if _, err := strconv.ParseInt(complete, 10, 64); err != nil {
-			return fmt.Errorf("invalid Content-Range %q for offset %d", contentRange, offset)
+		var err error
+		completeSize, err = strconv.ParseInt(complete, 10, 64)
+		if err != nil || completeSize < 0 {
+			return 0, fmt.Errorf("invalid Content-Range %q for offset %d", contentRange, offset)
 		}
 	}
 	startStr, endStr, ok := strings.Cut(byteRange, "-")
 	if !ok || startStr == "" || endStr == "" {
-		return fmt.Errorf("invalid Content-Range %q for offset %d", contentRange, offset)
+		return 0, fmt.Errorf("invalid Content-Range %q for offset %d", contentRange, offset)
 	}
 	start, err := strconv.ParseInt(startStr, 10, 64)
 	if err != nil {
-		return fmt.Errorf("invalid Content-Range %q for offset %d", contentRange, offset)
+		return 0, fmt.Errorf("invalid Content-Range %q for offset %d", contentRange, offset)
 	}
 	end, err := strconv.ParseInt(endStr, 10, 64)
 	if err != nil || end < start {
-		return fmt.Errorf("invalid Content-Range %q for offset %d", contentRange, offset)
+		return 0, fmt.Errorf("invalid Content-Range %q for offset %d", contentRange, offset)
 	}
 	if start != offset {
-		return fmt.Errorf("unexpected Content-Range %q for offset %d", contentRange, offset)
+		return 0, fmt.Errorf("unexpected Content-Range %q for offset %d", contentRange, offset)
 	}
-	return nil
+	if end > requestedEnd {
+		return 0, fmt.Errorf("Content-Range %q extends past requested end %d for offset %d", contentRange, requestedEnd, offset)
+	}
+	if completeSize >= 0 && completeSize <= end {
+		return 0, fmt.Errorf("invalid Content-Range %q for offset %d", contentRange, offset)
+	}
+	return end, nil
 }
 
 type limitedReadCloser struct {
--- a/pkg/client/transfer.go
+++ b/pkg/client/transfer.go
@@ -14,6 +14,7 @@ import (
 	"net/url"
 	"os"
 	"runtime"
+	"strconv"
 	"strings"
 	"sync"
 	"sync/atomic"
@@ -1599,15 +1600,41 @@ func (c *Client) sliceBody(rc io.ReadCloser, offset, length int64) (io.ReadClose
 	return &limitedReadCloser{r: io.LimitReader(rc, length), c: rc}, nil
 }
 
-// validatePartialContentRange checks that a first-hop 206 response starts at
-// the requested offset. The caller bounds the length via io.LimitReader, so a
-// server-clamped end (EOF) is allowed here.
+// validatePartialContentRange checks that a first-hop 206 response carries a
+// well-formed Content-Range starting at the requested offset. The caller
+// bounds the length via io.LimitReader, so a server-clamped end (EOF) is
+// allowed here.
 func validatePartialContentRange(resp *http.Response, offset int64) error {
 	contentRange := resp.Header.Get("Content-Range")
 	if contentRange == "" {
 		return fmt.Errorf("206 response missing Content-Range for offset %d", offset)
 	}
-	if !strings.HasPrefix(contentRange, fmt.Sprintf("bytes %d-", offset)) {
+	rest, ok := strings.CutPrefix(contentRange, "bytes ")
+	if !ok {
+		return fmt.Errorf("invalid Content-Range %q for offset %d", contentRange, offset)
+	}
+	byteRange, complete, ok := strings.Cut(rest, "/")
+	if !ok {
+		return fmt.Errorf("invalid Content-Range %q for offset %d", contentRange, offset)
+	}
+	if complete != "*" {
+		if _, err := strconv.ParseInt(complete, 10, 64); err != nil {
+			return fmt.Errorf("invalid Content-Range %q for offset %d", contentRange, offset)
+		}
+	}
+	startStr, endStr, ok := strings.Cut(byteRange, "-")
+	if !ok || startStr == "" || endStr == "" {
+		return fmt.Errorf("invalid Content-Range %q for offset %d", contentRange, offset)
+	}
+	start, err := strconv.ParseInt(startStr, 10, 64)
+	if err != nil {
+		return fmt.Errorf("invalid Content-Range %q for offset %d", contentRange, offset)
+	}
+	end, err := strconv.ParseInt(endStr, 10, 64)
+	if err != nil || end < start {
+		return fmt.Errorf("invalid Content-Range %q for offset %d", contentRange, offset)
+	}
+	if start != offset {
 		return fmt.Errorf("unexpected Content-Range %q for offset %d", contentRange, offset)
 	}
 	return nil
--- a/pkg/client/transfer.go
+++ b/pkg/client/transfer.go
@@ -1604,9 +1604,10 @@ func (c *Client) sliceBody(rc io.ReadCloser, offset, length int64) (io.ReadClose
 
 // validatePartialContentRange checks that a first-hop 206 response carries a
 // well-formed Content-Range starting at the requested offset and not extending
-// past the requested end. It returns the declared response end so the caller
-// can bound the body to the declared range. A server-clamped end (EOF) is
-// allowed.
+// past the requested end. Byte positions and the complete length must be
+// strict decimal digits, with `*` allowed for the complete length. It returns
+// the declared response end so the caller can bound the body to the declared
+// range. A server-clamped end (EOF) is allowed.
 func validatePartialContentRange(resp *http.Response, offset, requestedEnd int64) (int64, error) {
 	contentRange := resp.Header.Get("Content-Range")
 	if contentRange == "" {
@@ -1622,14 +1623,17 @@ func validatePartialContentRange(resp *http.Response, offset, requestedEnd int64
 	}
 	completeSize := int64(-1)
 	if complete != "*" {
+		if !isDecimalDigits(complete) {
+			return 0, fmt.Errorf("invalid Content-Range %q for offset %d", contentRange, offset)
+		}
 		var err error
 		completeSize, err = strconv.ParseInt(complete, 10, 64)
-		if err != nil || completeSize < 0 {
+		if err != nil {
 			return 0, fmt.Errorf("invalid Content-Range %q for offset %d", contentRange, offset)
 		}
 	}
 	startStr, endStr, ok := strings.Cut(byteRange, "-")
-	if !ok || startStr == "" || endStr == "" {
+	if !ok || !isDecimalDigits(startStr) || !isDecimalDigits(endStr) {
 		return 0, fmt.Errorf("invalid Content-Range %q for offset %d", contentRange, offset)
 	}
 	start, err := strconv.ParseInt(startStr, 10, 64)
@@ -1650,6 +1654,21 @@ func validatePartialContentRange(resp *http.Response, offset, requestedEnd int64
 		return 0, fmt.Errorf("invalid Content-Range %q for offset %d", contentRange, offset)
 	}
 	return end, nil
+}
+
+// isDecimalDigits reports whether s is a non-empty sequence of ASCII digits.
+// HTTP byte-range positions and complete lengths are digit sequences; signed
+// values are not valid.
+func isDecimalDigits(s string) bool {
+	if s == "" {
+		return false
+	}
+	for i := 0; i < len(s); i++ {
+		if s[i] < '0' || s[i] > '9' {
+			return false
+		}
+	}
+	return true
 }
 
 type limitedReadCloser struct {
--- a/pkg/client/transfer.go
+++ b/pkg/client/transfer.go
@@ -1345,7 +1345,7 @@ func newMultipartAbortContext(parent context.Context, timeout time.Duration) (co
 
 // ReadStream reads a file, following 302 redirects for large files.
 func (c *Client) ReadStream(ctx context.Context, path string) (io.ReadCloser, error) {
-	resp, err := c.readWithoutRedirect(ctx, path)
+	resp, err := c.readWithoutRedirect(ctx, path, "")
 	if err != nil {
 		return nil, err
 	}
@@ -1382,7 +1382,7 @@ func (c *Client) ReadStream(ctx context.Context, path string) (io.ReadCloser, er
 	}
 }
 
-func (c *Client) readWithoutRedirect(ctx context.Context, path string) (*http.Response, error) {
+func (c *Client) readWithoutRedirect(ctx context.Context, path, rangeHeader string) (*http.Response, error) {
 	// Disable redirect following so we can detect 302
 	noRedirectClient := *c.httpClient
 	noRedirectClient.CheckRedirect = func(req *http.Request, via []*http.Request) error {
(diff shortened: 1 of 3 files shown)


mornyx (commented):
## Review summary

This is the right client-side fix for #895: `ReadStreamRange` now sends `Range` on the first `/v1/fs` GET, so inline WAL page reads no longer stream from byte 0 and discard the prefix. First-hop 206 is not double-sliced (`LimitReader` only), 416 is EOF, 200 still uses `sliceBody`, and the 302/307 hop is unchanged. Under a strict maintainability bar the new 206 guard is still too weak — Content-Range is prefix-matched rather than parsed — and the “bounded transfer” unit test does not actually pin the client-side length cap. Neither is merge-blocking against a well-behaved server.

**What looks good**
- Range is added only on `ReadStreamRange`; `ReadStream` and `resolveReadTarget` stay unranged, so redirect discovery is not turned into a 206.
- The nonzero-offset first-hop 206 test would fail if `sliceBody` were reused — the easy way to “fix” this and return empty pages.
- The FUSE WAL test pins the original failure: omitted Range streams the full body and the server byte-count assertion fails (40 KiB vs 4 KiB).

---

**1. Parse `Content-Range` instead of prefix-matching**

`validatePartialContentRange` (`pkg/client/transfer.go:1610`) only checks `strings.HasPrefix(contentRange, "bytes {offset}-")`. The dash correctly rejects a wrong start (`bytes 5-` does not match `bytes 50-…`), but malformed values that share that prefix are accepted and then treated as file data: `bytes 5-not-a-range`, `bytes 5-4/64` (`end < start`), `bytes 5-`, and trailing junk. `TestReadStreamRangeFirstHop206BadContentRange` still only covers missing and `bytes 0-3/64`.

This is incomplete hardening, not a live corruption path against a spec-compliant server: length is still capped with `io.LimitReader`, and the pre-change code had no first-hop 206 handling at all. Still, this helper’s job is to reject a lying/broken 206, and the prior-thread finding remains unresolved on this HEAD.

Parse `bytes start-end/complete` (tolerate `*` total), require `start == offset` and `end >= start`, reject non-numeric junk, and keep allowing a server-clamped end. Extend the bad-Content-Range table with `bytes 5-not-a-range` and `end < start`.

**2. Bounded-transfer unit test does not pin `LimitReader`**

`TestReadStreamRangeFirstHopBoundedTransfer` (`pkg/client/transfer_test.go:1702`) writes `wal[offset:]`, which is already exactly `length` bytes, then asserts the server wrote `length`. A client that returned `resp.Body` unbounded would pass identically. The test does pin no-double-slice and data equality; the FUSE test pins the #895 transfer blow-up. What it does not pin is the client-side bound the PR body claims.

Have the 206 handler write past the advertised end and assert the client still returns exactly `length`.

---

### Nits

- First-hop 206 is now validated and `LimitReader`-bounded; the redirect-hop 206 still returns raw `resp2.Body`. Harmless against S3 and intentionally unchanged in this PR, but the two hops now have different contracts. A small shared `boundedRangeBody` helper would keep them from drifting later.
- Optionally assert the FUSE Range header as the exact `bytes=%d-%d` rather than a start prefix.

---

### Recommendation
**Comment.** The first-hop Range fix and tests look correct for #895; parse Content-Range and pin the client-side length cap before calling the guard rails done.

qiffang (approved):
No blocking findings on exact head `dbb6d24cb0ba4d8b7ccc431ee7efb6045df63051`.

I verified the #895 root-cause path rather than only the diff shape: `ReadStreamRange` now sends `Range` on the initial authenticated `/v1/fs` GET, handles first-hop `206` without double-slicing after checking the `Content-Range` start, treats first-hop `416` as EOF, preserves the old first-hop `200` slice fallback, and leaves the 302/307 redirected object-range path behavior unchanged.

I also checked the authoritative server side in `tidbcloud/fs` `origin/main`: append-log reads already parse the first-hop `Range` header and return `206 + Content-Range` for partial reads, so this client change matches the production server contract. Normal inline files that ignore Range still go through the compatibility `200` fallback.

The tests are discriminating for the defect being fixed. The client tests cover first-hop 206 at zero and nonzero offsets, 416, 200 fallback, bad Content-Range, and bounded transfer. The FUSE test uses a SQLite WAL sidecar path, which bypasses stable caches and reaches the real `ReadStreamRange` path; old code would omit Range, get a full-body 200, and fail the byte-count assertion.

Validation: GitHub Code CI, Local E2E, and CodeRabbit are green on this head. Locally I ran `git diff --check`, gofmt check, `go test -c ./pkg/client ./pkg/fuse`, `go vet ./pkg/client ./pkg/fuse`, focused FUSE WAL/range regressions, and `TestDat9FSReadSQLiteJournalSendsFirstHopRange -race -count=20`. Full local `go test ./pkg/fuse -count=1` still fails only on the known Darwin/rootless mount-permission environment tests (`TestShouldContinueAfterWaitMountPermissionErrorWhenProbePasses/Fails`), with no PR-related failure. `pkg/client` runtime tests are blocked locally by the package `TestMain` requiring rootless Docker, so I relied on compile/vet plus hosted Code CI for that package's runtime coverage.

Non-blocking note: this PR is a correct client-side fix, but issue #895's larger hosted >40 MiB WAL checkpoint acceptance run is still deferred. I would not close the issue or make an external “full acceptance passed” claim until that run is attached.

qiffang (changes requested):
Reviewed exact head `dbb6d24cb0ba4d8b7ccc431ee7efb6045df63051`.

One merge blocker remains in `pkg/client/transfer.go:1605`: `validatePartialContentRange` does not parse or validate a Content-Range; it only checks the string prefix `bytes <offset>-`. Responses such as `Content-Range: bytes 5-not-a-range`, `bytes 5-4/64`, or a declared one-byte range followed by a longer body are accepted as the requested file slice. I added an independent regression through the exported client API: a 206 response with `bytes 5-not-a-range` and body `old!` is returned as successful data. This is a data-integrity failure at the exact new protocol boundary, and the committed bad-range test covers only missing/wrong-start headers, so it stays green.

Before merge, parse the complete satisfied byte-range response and reject malformed/inconsistent start/end/complete-length values; never expose more bytes than both the declared response range and the requested range. Add malformed numeric/order/extent cases to the regression.

The main fix direction is otherwise correct: first-hop Range is sent, the authoritative append-log server contract returns 206 + Content-Range, 416/200/redirect compatibility remains, and the FUSE test genuinely discriminates prefix transfer. Exact-head Code CI, Local E2E, and CodeRabbit are green; local focused/race FUSE tests, compile, vet, lint, gofmt, and diff-check pass. Client runtime tests are blocked locally by the package TestMain requiring rootless Docker.

The >40 MiB hosted checkpoint acceptance remains a non-blocking evidence gap for closing issue #895 or making the release claim.

qiffang (changes requested):
I am retracting my earlier approval on this same head after rechecking @dat9-dev2's finding. The main #895 fix is still directionally correct, but there is one merge-blocking validation gap in the new first-hop `206` path.

`pkg/client/transfer.go:1516-1523` now trusts a first-hop partial response as long as `validatePartialContentRange` succeeds, and `validatePartialContentRange` at `pkg/client/transfer.go:1605-1613` only checks `strings.HasPrefix(contentRange, fmt.Sprintf("bytes %d-", offset))`. That accepts syntactically invalid or contradictory partial ranges such as `Content-Range: bytes 5-not-a-range` or `Content-Range: bytes 5-4/64`. The caller then returns the response body as file bytes, bounded only by the requested length.

That is a silent data-integrity failure mode: a server/proxy bug that returns malformed `206` with the requested start can be surfaced to FUSE/SDK callers as valid content. The existing bad-range test only covers missing and wrong-start headers, so it does not distinguish this class of malformed same-start responses.

Fix gate: parse and validate the full `Content-Range` for first-hop `206` before returning bytes. At minimum, require `bytes <start>-<end>/<size-or-*>`, `start == requested offset`, `end >= start`, `end <= requested end` unless intentionally documenting a wider server response, and total size either unknown or consistent with `end`. Bound the returned body by the validated response range as well as the requested length. Add discriminating tests for malformed numeric fields, inverted ranges, and ranges extending outside the requested window.

The rest of the review stands: first-hop `Range` is now sent, authoritative `tidbcloud/fs` append-log reads emit valid `206 + Content-Range`, 416/200 fallback/redirect compatibility paths are preserved, and the FUSE WAL test hits the real range-read path. This blocker is limited to accepting malformed first-hop `206` as valid file data.

mashenjun:
@mornyx thanks for the review. Both should-fix items are addressed in `871f784e`:

1. Content-Range is now fully parsed: `validatePartialContentRange` parses `bytes start-end/complete` (tolerating `*` as the complete length), requires `start == offset` and `end >= start`, and rejects non-numeric, open-ended, trailing-junk, and missing-`bytes ` values. `TestReadStreamRangeFirstHop206BadContentRange` gained cases for `bytes 5-x/64`, `bytes 5-4/64`, `bytes 5-/64`, `bytes 5-8/64junk`, `chunks 5-8/64`, and `bytes 5-8/`.

2. The bounded-transfer test now pins the client-side bound: the read offset moved to `walSize-8192` and the handler streams 8192 bytes past the advertised 4096 end; the client must still return exactly 4096, so an unbounded body fails the length check.

On the nits:

- Redirect-hop 206 intentionally keeps the raw `resp2.Body` contract; the 302/307 path is out of this PR's scope. Extracting a shared `boundedRangeBody` helper can be a follow-up.
- The FUSE test keeps the Range prefix assertion to stay minimal. Exact `bytes=%d-%d` matching is also feasible there since that path has no Prefetcher; happy to tighten if you'd like.

This commit also covers CodeRabbit's equivalent Content-Range finding.

qiffang (changes requested):
Fresh re-review of current head `871f784e25d3b9a3e8cb6c2f14f6fab0675a9488`.

The previous malformed `Content-Range` blocker is partially fixed: nonnumeric end values, inverted ranges, missing unit/prefix, and trailing junk are now rejected. The original first-hop Range fix remains correct.

One validation gap still needs to be closed before merge. `validatePartialContentRange` now parses `bytes <start>-<end>/<complete>`, but it still accepts impossible complete lengths such as `Content-Range: bytes 5-8/4` and `Content-Range: bytes 5-8/-1`, because it only checks that `complete` parses as an int. It also still does not receive the requested length, so it cannot reject a `206` response that claims a range outside the request window, e.g. request `bytes=5-8` but response `Content-Range: bytes 5-100/200`. The body is currently bounded by the requested length, which prevents over-read, but the client is still accepting an inconsistent partial-content contract as valid file data.

For this filesystem path, malformed `206` should fail closed. Please validate the full first-hop response: `complete` must be `*` or nonnegative and greater than `end`, `start == requested offset`, `end >= start`, and the declared response range should not extend outside the requested range unless that wider-response behavior is explicitly intended and tested. Add same-start invalid extent cases (`bytes 5-8/4`, negative complete length, and wider-than-requested range) to the bad-range regression.

Validation on this head: local `git diff --check`, gofmt check, `go test -c ./pkg/client ./pkg/fuse`, `go vet ./pkg/client ./pkg/fuse`, and focused FUSE WAL/range tests pass. GitHub Code CI and Local E2E are still in progress on this new head at the time of review.

qiffang (changes requested):
Fresh review of exact head `871f784e25d3b9a3e8cb6c2f14f6fab0675a9488`.

The prior malformed-end finding is only partially closed. `validatePartialContentRange` now parses `start` and `end`, but it still does not validate the numeric complete length against the declared range, does not receive the requested end, and the returned reader is bounded only by the requested length rather than the response's declared range.

I reproduced all four cases through the exported `Client.ReadStreamRange` API on this head:

- request `bytes=5-8`, response `Content-Range: bytes 5-8/4` is accepted even though the complete length is not greater than `end`;
- `bytes 5-8/-1` is accepted because `strconv.ParseInt` accepts a negative value even though the grammar is digits-only;
- `bytes 5-100/200` is accepted although it extends beyond the requested window;
- `bytes 5-5/64` with a four-byte body returns all four bytes, even though the response declares that only one byte is enclosed.

This is still a fail-open data-integrity boundary. RFC 9110 §14.4 defines `complete-length` as digits and marks a satisfied Content-Range invalid when the numeric complete length is less than or equal to the last position. It also requires the Content-Range to describe the bytes enclosed in the 206 body.

Before merge, validate the full response against both its own fields and the request: require a strict nonnegative decimal complete length or `*`, require numeric `complete > end`, require `start == offset`, and reject `end > offset+length-1`. Bound the exposed reader by `min(requested length, end-start+1)` so a valid server-clamped/subset response cannot expose bytes outside its declared range. Add discriminating cases for impossible/negative complete length, wider-than-requested ranges, and a shorter declared range followed by extra body bytes.

The main #895 fix remains correct: the first hop sends Range, authoritative append-log server responses match the valid 206 contract, 416/200/redirect compatibility remains, and the FUSE WAL test hits the real nonzero-offset path. Local focused FUSE tests passed 20x normally and under `-race`; compile, vet, lint, gofmt, and diff-check pass. Exact-head Code CI and Local E2E were still running at this review.

mashenjun:
@qiffang the remaining first-hop 206 gap is closed in `f7688800`:

- `validatePartialContentRange` now takes the requested end and returns the declared response end. It requires `complete` to be `*` or a nonnegative size greater than the declared `end`, and rejects a declared range that extends past the requested end (e.g. `Content-Range: bytes 5-100/200` for a `bytes=5-8` request).
- The returned body is bounded by the declared response range (`end-offset+1`), not just the requested length, so a server streaming past its declared end cannot leak extra bytes.
- `TestReadStreamRangeFirstHop206BadContentRange` gained `bytes 5-8/4`, `bytes 5-8/-1`, and `bytes 5-100/200` cases.

`start == offset`, `end >= start`, and the server-clamped-EOF case (`end < requestedEnd` with a consistent `complete`) all remain accepted. gofmt/vet/test-compile checks pass locally; runtime regression runs stay with the unified CI/e2e flow.

mornyx (approved):
## Review summary

Follow-up on HEAD `f768880`. The original #895 fix still looks right: `ReadStreamRange` sends `Range` on the first `/v1/fs` GET, first-hop 206 is not double-sliced, 416 is EOF, 200 still uses `sliceBody`, and the FUSE WAL test still makes a missing Range observable via byte count. Under a strict maintainability bar, the previous should-fixes and the later first-hop 206 integrity gaps are resolved in code on this HEAD. Remaining notes are optional test/grammar polish, not merge blockers.

- `validatePartialContentRange` now parses `bytes start-end/complete` (`*` allowed), requires `start == offset` and `end >= start`, and rejects non-numeric / open-ended / trailing-junk / missing-unit values.
- Impossible or negative complete lengths (`bytes 5-8/4`, `bytes 5-8/-1`) and a declared window past the request (`bytes 5-100/200` for `bytes=5-8`) are rejected.
- The 206 body is bounded by the declared end (`end-offset+1`), not only the requested length.
- `TestReadStreamRangeFirstHopBoundedTransfer` now over-sends (server writes 2× the advertised range) and asserts the client still returns exactly the requested slice.

**What looks good**
- First-hop 206 validates the declared window, then applies a single `LimitReader` — no second `sliceBody` skip.
- The bad-Content-Range table now covers the malformed and inconsistent cases that previously slipped through prefix matching.
- Redirect-hop 206 and the FUSE Range prefix assertion stay as agreed out of scope for this PR.

---

### Nits

- The declared-end bound only differs from `LimitReader(length)` when the server returns a valid shorter range and extra body bytes. `TestReadStreamRangeFirstHopBoundedTransfer` still declares `end == requestedEnd`, so it would stay green if the bound were reverted to requested length. Optional: request `bytes=5-8`, respond `Content-Range: bytes 5-6/64` with 8 body bytes, and assert exactly 2 bytes returned.
- `strconv.ParseInt` accepts a leading `+` (`bytes +5-8/+64`). Grammar-only; numeric values and the body bound are unchanged.

---

### Recommendation
**Approve.** Prior should-fixes and first-hop 206 integrity items are verified fixed on this HEAD; leftover nits are optional.

qiffang (changes requested):
Reviewed exact head `f76888008e4ec548fc8a5ee166af8e615506142f`.

The three previously reported data-integrity behaviors are now correct in production code:

- numeric complete length must be greater than the declared end;
- the declared end cannot exceed the requested end;
- the returned stream is limited to `end - start + 1`, so a response that declares `bytes 5-5/64` but sends a longer body exposes only one byte.

I verified the last behavior through the exported `ReadStreamRange` API with a real `httptest.Server`; the current head returns exactly the one declared byte. The added invalid-range cases also reject the impossible total, negative total, and wider-than-requested examples from the prior review.

Two small but discriminating test/contract gaps remain before approval:

1. The new response-declared body bound has no committed regression. `TestReadStreamRangeFirstHopBoundedTransfer` still declares the full requested range and therefore passes if the implementation regresses from `end-offset+1` back to the old request `length`. Please add the exact short-declaration/long-body case (`Content-Range: bytes 5-5/64`, four-byte body, request 5..8) and assert that only the first byte is returned. This was one of the concrete prior failures and is the only test that pins the new bound.
2. `validatePartialContentRange` says it accepts a well-formed HTTP byte range, but `strconv.ParseInt` also accepts a leading `+`. I confirmed `Content-Range: bytes +5-+8/+64` is accepted and returned successfully, although RFC byte positions and complete length are `1*DIGIT`. The previous gate asked for strict nonnegative decimal or `*`; reject sign-prefixed fields and add a malformed-header case.

The main first-hop Range fix, 200 fallback, 302/307 path, 416 EOF behavior, and FUSE WAL route remain sound. Exact-head Code CI, Local E2E, and CodeRabbit are green. Local diff/gofmt, client/fuse compilation and vet, plus the FUSE first-hop WAL regression in normal and race mode (20x each) passed. The hosted >40 MiB checkpoint run remains release/issue-closure evidence, not a blocker for this code review.

qiffang (changes requested):
Reviewed current head `f76888008e4ec548fc8a5ee166af8e615506142f`.

The code change closes the previous first-hop `206 Content-Range` behavior issue: the validator now parses start/end/complete, rejects negative or impossible complete lengths, rejects ranges past the requested end, and the first-hop `206` body is now limited to the declared range length.

The remaining blocker is test coverage for the exact regression we agreed on in the last review round. `TestReadStreamRangeFirstHopBoundedTransfer` still declares the same end as the requested end, so it would pass even if the implementation regressed back to `io.LimitReader(resp.Body, length)`. That means the suite does not catch the `Content-Range: bytes 5-5/64` + four-byte body case that previously returned all four bytes.

Please add a discriminating test that requests a larger window but receives a shorter declared `206` range, for example request `offset=5,length=4`, respond with `Content-Range: bytes 5-5/64` and body `old!`, then assert the client exposes exactly one byte. That test should fail on `871f784e` and pass on this head. Once that is added, I expect this to be approvable unless the new delta introduces something unexpected.

Validation on this head: hosted Code CI / Local E2E / CodeRabbit are green; local `git diff --check`, gofmt check, `go test -c ./pkg/client ./pkg/fuse`, `go vet ./pkg/client ./pkg/fuse`, focused FUSE WAL range test `-count=10`, and the same test under `-race -count=3` passed. Local focused `pkg/client` runtime tests still fail before test execution on this machine because package `TestMain` requires rootless Docker.

qiffang (changes requested):
Follow-up on current head `f76888008e4ec548fc8a5ee166af8e615506142f` after rechecking the parser boundary.

In addition to the missing short-declared/long-body regression test, the validator should reject signed byte positions and complete lengths. `strconv.ParseInt` accepts `+5` / `+64`, but HTTP byte-range positions and complete-length are digit sequences or `*`, not signed integers. Today a response like `Content-Range: bytes +5-+8/+64` is accepted as valid and data is exposed.

Please make the parser require strict decimal digits for start/end/complete (with `*` still allowed for complete), and add a malformed signed-field case to the bad Content-Range table. This is the same first-hop `206` contract gate as the earlier malformed-range fixes.

mashenjun:
@qiffang both remaining gaps are closed in `cac255c2`:

1. Strict decimal fields: `validatePartialContentRange` now requires start/end/complete to be digit sequences via `isDecimalDigits` (`*` still allowed for complete), so signed values like `Content-Range: bytes +5-+8/+64` are rejected. The bad-range table gained a `signed-fields` case.

2. Discriminating regression: new `TestReadStreamRangeFirstHop206ShorterDeclaredRange` requests `bytes=5-8`, gets `Content-Range: bytes 5-5/64` with a four-byte body `old!`, and asserts the client exposes exactly one byte. It fails on `871f784e` (which bounded by requested length) and passes on this head.

gofmt/vet/test-compile checks pass locally; runtime runs stay with the unified CI/e2e flow.

qiffang (approved):
Reviewed exact head `cac255c2d66797b4e8519ed0331ace7f00504d77`.

The two remaining gates from the prior review are closed, with discriminating tests:

- `isDecimalDigits` now requires a non-empty ASCII digit sequence for start/end/complete, while complete still permits `*`. The added signed-field case rejects `bytes +5-+8/+64`; oversized digit strings still fail `ParseInt`.
- `TestReadStreamRangeFirstHop206ShorterDeclaredRange` requests bytes 5..8, receives `Content-Range: bytes 5-5/64` with a four-byte body, and asserts that only one byte is exposed. This fails the previous request-length limiter and passes only with the response-declared bound.

I also exercised the exported `ReadStreamRange` API independently: short-declared/long-body returns exactly one byte, while signed fields, an impossible total, and a response wider than the request are rejected. The main first-hop Range behavior remains compatible with first-hop 200 fallback, 302/307 redirects, 416 EOF, and the FUSE SQLite journal path. I found no remaining code-level blocker.

Local `git diff --check`, touched-file gofmt, client/fuse compilation and vet, the FUSE first-hop WAL test (20x), and its race form (10x) passed. CodeRabbit is green on this head; exact-head Code CI and Local E2E were still running at review time, so they remain merge gates rather than code-review findings. The separate >40 MiB hosted checkpoint run remains evidence required before closing #895 or making the release claim, not a blocker for this patch.

qiffang (approved):
Reviewed current head `cac255c2d66797b4e8519ed0331ace7f00504d77`.

No blocking findings from this review seat. The two remaining first-hop `206 Content-Range` gates are now closed:

- `validatePartialContentRange` requires strict non-empty ASCII digits for start/end/numeric complete length, while still allowing `*` for complete length, so signed forms like `bytes +5-+8/+64` are rejected.
- `ReadStreamRange` uses the parsed response end to cap the first-hop `206` body to the declared range. The new shorter-declared-range regression requests `bytes=5-8`, receives `Content-Range: bytes 5-5/64` with a four-byte body, and asserts only the declared one byte is exposed. That would fail on the previous requested-length-only limiter.

The earlier behavior gates remain intact: first-hop GET sends `Range`, legal `206 + Content-Range` is accepted, impossible total length and ranges beyond the requested end are rejected, `416` maps to EOF, first-hop `200` fallback still slices old/inline responses, and the redirect path still repeats the range on the object URL. The FUSE SQLite WAL test still hits the real non-zero-offset `ReadStreamRange` path and verifies only the requested bytes are transferred.

Validation: local `git diff --check`, gofmt check, `go test -c ./pkg/client ./pkg/fuse`, `go vet ./pkg/client ./pkg/fuse`, FUSE WAL range `-count=10`, and FUSE WAL range under `-race -count=3` passed. Local `pkg/client` runtime tests still fail before execution because this machine lacks rootless Docker for the package TestMain. Hosted Code CI and CodeRabbit are green on this exact head; Local E2E was still in progress when I submitted this approval, so merge should still wait for that required check to finish green.

Non-blocking caveat: the >40MiB hosted WAL checkpoint evidence remains a close/release criterion for issue #895, not a code blocker for this PR.

PR #900, opened 2026-09-07
fix(fuse): restore evicted WriteBuffer parts on WAL back-write

### LINKED ISSUE #898: FUSE append-log: large WAL back-write into evicted part corrupts SQLite after 1GiB WAL transaction
## Bug Report

On `--durability interactive` with `--append-log` matching `*.db-wal`, a single ~1GiB SQLite WAL transaction finishes COMMIT, then the same connection gets `database disk image is malformed`. The main DB file grows to ~1.2GiB and a later open can still see table `kv`, but `SELECT count(*)` did not return within 60s. Mount logs show append-log **back-write into an already-evicted WAL part**, with zeros promised for the hole.

This is not the 4KiB-empty-DB CAS loss in #896 (that was DELETE-journal `kvtest init`). It is also not the checkpoint **read** amplification in #895.

#### Environment

| Component | Value |
|---|---|
| Host | Linux x86_64 EC2, real FUSE |
| SQLite | 3.54.0 (Python `sqlite3`) |
| Mount | `--durability interactive --profile none --append-log='**/*-wal' --append-log='**/*-journal' --write-cache-size-mb 8192` |

`--write-cache-size-mb 8192` is only to get past the default 1GiB shadow quota (#896). It does not cause the WAL warning.

On the mount, a unique path:

```python
import os, sqlite3

path = "./d9/plain/kv-wal-1g.db"
conn = sqlite3.connect(path)
conn.isolation_level = None
cur = conn.cursor()
cur.execute("PRAGMA page_size=4096")
assert cur.execute("PRAGMA journal_mode=WAL").fetchone()[0].lower() == "wal"
cur.execute("VACUUM")
cur.execute("BEGIN")
cur.execute("CREATE TABLE kv(k INTEGER PRIMARY KEY, v BLOB)")
cur.execute(
    "WITH RECURSIVE c(x) AS (VALUES(1) UNION ALL SELECT x+1 FROM c WHERE x<262144) "
    "INSERT INTO kv(k,v) SELECT x, randomblob(4096) FROM c"
)
cur.execute("COMMIT")
print("size", os.path.getsize(path), "wal", os.path.getsize(path + "-wal"))
print(cur.execute("SELECT count(*) FROM kv").fetchone())  # fails here
conn.close()
```

This is `kvtest init` scaled to 1GiB, except `PRAGMA journal_mode=WAL` is set **before** the large insert (stock `kvtest init` stays on DELETE journal).

- COMMIT returns success
- same connection `SELECT count(*) FROM kv` = 262144
- `PRAGMA integrity_check` = `ok`
- a new connection sees the same row count
- no zero-fill of WAL holes

During the insert the WAL grew to ~1.2GiB while `main.db` stayed 4096 bytes (normal WAL). COMMIT then copied into the main DB (~1208307712 bytes). Then:

```text
same_conn SELECT count(*): sqlite3.DatabaseError: database disk image is malformed
```

Mount log:

```text
13:02:16 commit queue: successfully uploaded /plain/kv-wal-1g-rerun.db (4096 bytes, rev=1)
13:02:21 WARNING: back-write to evicted part 2 of /plain/kv-wal-1g-rerun.db-wal
         — non-written bytes will be zero (original data already uploaded to S3)
13:10:31 commit queue: successfully uploaded /plain/kv-wal-1g-rerun.db (1208307712 bytes, rev=2)
```

A later open on the same mount:

- `kv-wal-1g-rerun.db` still 1208307712 bytes
- `sqlite_master` lists `kv`
- `SELECT count(*)` did not return in 60s (not measured to completion)

So this is **not** “init succeeds then remote is a 4KiB empty header” (#896). The large main DB is uploaded as rev=2. The failure is WAL-side eviction/zero-fill plus a malformed image / stuck read after checkpoint.

The same `back-write to evicted part of *.db-wal` warning was also seen on `threadtest3 walthread5` (WAL copy/header rewrite hang) in the community.sqlite run.

```text
component: drive9
version: dev
go_version: go1.27.1
```

linux/amd64, `CGO_ENABLED=0`, built from `origin/main`.

## Not these issues

| Issue | Why this is different |
|---|---|
| [#896](https://github.com/mem9-ai/drive9/issues/896) | DELETE-journal `kvtest init`: 409/LWW, durable object is **4KiB empty DB**, `no such table`. WAL 1GiB leaves a **1.2GiB** main DB and table `kv`. |
| [#895](https://github.com/mem9-ai/drive9/issues/895) | Checkpoint **read** path: first-hop GET without `Range`, prefix streaming, stall, **no corruption**; recovery from WAL is clean. This case **writes** into an evicted WAL part and SQLite reports **malformed**. |
| [#875](https://github.com/mem9-ai/drive9/issues/875) | Every WAL fsync **re-uploads the whole growing WAL** (write amplification). Here append-log is on; the defect is **back-write after part eviction**, not full-file PUT on each commit. |
| [#889](https://github.com/mem9-ai/drive9/issues/889) | WAL **truncate** leaves `UnlinkedData` on live handles (closed). This workload never successfully truncated/reset the WAL before the warning; it is growth + evict + write into the hole. |
| [#894](https://github.com/mem9-ai/drive9/issues/894) | Local stale **read** of main DB sibling buffer; remote is correct. |

## Impact

A single large WAL transaction (cache spill / one COMMIT of ~1GiB of blobs) can poison the WAL via zero-filled evicted parts. SQLite then reports a malformed image. `integrity_check` / row counts are not a reliable success signal if the first statement after COMMIT already fails.

Suggested area: FUSE append-log WAL + writeback part eviction (`back-write to evicted part`), not `ReadStreamRange` and not commit-queue 4KiB CAS.

mornyx:
Refs #898

## Summary
SQLite WAL is not pure append at FUSE part granularity. ShadowSpill evicts completed WriteBuffer parts, and a later non-tail write into an evicted part used to recreate that part as zeros. That is the `back-write to evicted part` warning and the Dirty hole mashenjun reproduced.

This does not auto-close #898: later byte-level evidence shows the remote WAL matched local shadow and WAL reads used `source=shadow-spill`. The remaining corruption is on the main DB commit-queue / checkpoint path.

## Change
- `ensurePart` restores an evicted part via `RestorePart` (local shadow) or `LoadPart` before applying the overlay.
- ShadowSpill Create/Open/truncate/WAL generation-reset wires `RestorePart` to read the part from the shadow file (writes already hit shadow first).
- Restore failure returns an error instead of silently zero-filling.
- Streamer-only buffers with no restorer keep the previous zero-fill warning.

Does not change sequential append, eviction, or append-log tail vs rewrite routing (`appendSafe` is still cleared on non-tail writes).

## Tests
- `TestWriteBuffer_RestorePartReloadsEvictedBytes`
- `TestWriteBuffer_RestorePartErrorFailsBackWrite`
- `TestShadowSpillBackWriteRestoresFromShadow`

`go test ./pkg/fuse -run 'TestWriteBuffer_RestorePart|TestWriteBuffer_EvictPart|TestShadowSpillBackWriteRestoresFromShadow|TestWriteBufferCanMaterializeFull|TestAppendLogGenerationReset'` pass.

A 1GiB WAL e2e on the original mount flags is still the field confirmation; this PR closes the zero-fill mechanism with a small-part unit gate.

## Summary by CodeRabbit

- **Bug Fixes**
  - Preserved existing data during back-writes and appends after buffered data was evicted.
  - Restored evicted content before applying updates, preventing previously written bytes from being replaced with zeros.
  - Improved write validation and error handling for oversized writes, failed restores, and incomplete reads.
  - Ensured successful synchronization and commits correctly reset temporary eviction state for subsequent writes.
  - Preserved retained data when files are truncated and later extended.
  - Ensured reads and writes continue to work correctly after an unlinked file is modified.

### DIFF
--- a/pkg/fuse/append_log.go
+++ b/pkg/fuse/append_log.go
@@ -681,12 +681,7 @@ func (fs *Dat9FS) rotateAppendLogGenerationShadowLocked(fh *FileHandle, path str
 	}
 	fh.ShadowReady = true
 	fh.ShadowSpill = true
-	if fh.Dirty != nil {
-		wb := fh.Dirty
-		wb.OnPartFull = func(partIdx int, _ []byte) {
-			wb.EvictPart(partIdx)
-		}
-	}
+	fs.bindShadowSpillEvictionLocked(fh)
 	fs.recordAppendLogGenerationResetShadowReady()
 }
 
--- a/pkg/fuse/dat9fs.go
+++ b/pkg/fuse/dat9fs.go
@@ -11736,6 +11736,42 @@ func (fs *Dat9FS) Read(cancel <-chan struct{}, input *gofuse.ReadIn, buf []byte)
 		bytesRead = n
 		return gofuse.ReadResultData(data), gofuse.OK
 	}
+	// Post-Fsync rebaseline leaves Dirty != nil with LoadPart. Large unlinks
+	// pin UnlinkedShadowGen instead of UnlinkedData; serve that private
+	// backing before falling through to a remote GET of the deleted path.
+	if fh.Unlinked && fh.UnlinkedShadowGen != 0 {
+		offset := int64(input.Offset)
+		unlinkedSize := fh.UnlinkedSize
+		gen := fh.UnlinkedShadowGen
+		if offset >= unlinkedSize {
+			fh.Unlock()
+			source = "unlinked-shadow-eof"
+			bytesRead = 0
+			return gofuse.ReadResultData(nil), gofuse.OK
+		}
+		end := offset + int64(input.Size)
+		if end > unlinkedSize {
+			end = unlinkedSize
+		}
+		fh.Unlock()
+		if fs.shadowStore == nil {
+			source = "unlinked-shadow-missing-store"
+			return nil, gofuse.EIO
+		}
+		result := make([]byte, end-offset)
+		n, err := fs.shadowStore.ReadAtGen(gen, offset, result)
+		if err != nil && (!errors.Is(err, io.EOF) || n != len(result)) {
+			source = "unlinked-shadow-error"
+			return nil, gofuse.EIO
+		}
+		if n != len(result) {
+			source = "unlinked-shadow-short"
+			return nil, gofuse.EIO
+		}
+		source = "unlinked-shadow"
+		bytesRead = n
+		return gofuse.ReadResultData(result), gofuse.OK
+	}
 	fs.refreshCleanCommittedRevisionForHandleLocked(fh)
 
 	if fh.ShadowSpill && fs.shadowStore != nil && fh.Dirty != nil && isSQLitePersistentJournalPath(fh.Path) && fh.Dirty.Size() == 0 && !fh.Dirty.hasDirtyPartMarks() && !fh.ZeroBase && fh.Flags&syscall.O_TRUNC == 0 {
--- a/pkg/fuse/dat9fs.go
+++ b/pkg/fuse/dat9fs.go
@@ -3124,10 +3124,21 @@ func (fs *Dat9FS) markHandleRemoteCommittedLocked(fh *FileHandle, revision int64
 // buffer. A successful Fsync/Flush must not leave uploadedParts/RestorePart
 // pointing at a shadow that the commit path is about to delete (#900).
 func (fs *Dat9FS) rebaselineCommittedDirtyBufferLocked(fh *FileHandle) {
-	if fh == nil || fh.Dirty == nil || fh.Dirty.HasDirtyParts() || !fh.Dirty.needsCommitRebaseline() {
+	if fh == nil || fh.Dirty == nil || fh.Dirty.HasDirtyParts() {
 		return
 	}
-	fs.rebindCleanWriteBufferToRemoteLocked(fh, fh.Dirty.Size())
+	if fh.Dirty.needsCommitRebaseline() {
+		// Evicted parts are gone from memory; the next same-FD write must
+		// load the just-committed remote object instead of a deleted shadow.
+		fs.rebindCleanWriteBufferToRemoteLocked(fh, fh.Dirty.Size())
+		return
+	}
+	// Create installs RestorePart/OnPartFull even for tiny files that never
+	// evicted. Clearing those callbacks is enough: keep the in-memory image
+	// so a later write does not GET a path the commit just deleted (unlink)
+	// or that the test server never served.
+	fh.Dirty.RestorePart = nil
+	fh.Dirty.OnPartFull = nil
 }
 
 func (fs *Dat9FS) seedReadCacheFromShadowGenerationLocked(path string, size int64, revision int64, shadowGen uint64) bool {
--- a/pkg/fuse/dat9fs.go
+++ b/pkg/fuse/dat9fs.go
@@ -213,6 +213,55 @@ func (fs *Dat9FS) newWriteBuffer(path string, maxSize, partSize int64) *WriteBuf
 	return wb
 }
 
+// bindShadowSpillEvictionLocked evicts completed WriteBuffer parts to keep
+// Dirty memory bounded and restores them from the shadow file on back-write.
+// SQLite WAL is not pure append at this granularity; recreating an evicted
+// part as zeros (#898) poisons later Dirty consumers (header reads, rewrite
+// snapshots if spill is dropped, same-path-dirty fallback).
+func (fs *Dat9FS) bindShadowSpillEvictionLocked(fh *FileHandle) {
+	if fh == nil || fh.Dirty == nil || !fh.ShadowSpill {
+		return
+	}
+	wb := fh.Dirty
+	wb.OnPartFull = func(partIdx int, _ []byte) {
+		wb.EvictPart(partIdx)
+	}
+	if fs == nil || fs.shadowStore == nil {
+		wb.RestorePart = nil
+		return
+	}
+	store := fs.shadowStore
+	wb.RestorePart = func(partNum int) ([]byte, error) {
+		return restoreWriteBufferPartFromShadow(store, wb, partNum)
+	}
+}
+
+func restoreWriteBufferPartFromShadow(store *ShadowStore, wb *WriteBuffer, partNum int) ([]byte, error) {
+	if store == nil || wb == nil || partNum < 1 || wb.partSize <= 0 {
+		return nil, fmt.Errorf("invalid shadow part restore")
+	}
+	partStart := int64(partNum-1) * wb.partSize
+	end := partStart + wb.partSize
+	if wb.totalSize > 0 && partStart >= wb.totalSize {
+		return nil, fmt.Errorf("part %d starts at %d beyond size %d", partNum, partStart, wb.totalSize)
+	}
+	if wb.totalSize > 0 && end > wb.totalSize {
+		end = wb.totalSize
+	}
+	if end <= partStart {
+		return nil, fmt.Errorf("empty part %d", partNum)
+	}
+	buf := make([]byte, end-partStart)
+	n, err := store.ReadAt(wb.path, partStart, buf)
+	if err != nil && !errors.Is(err, io.EOF) {
+		return nil, err
+	}
+	if n != len(buf) {
+		return nil, fmt.Errorf("shadow part %d short read n=%d want=%d", partNum, n, len(buf))
+	}
+	return buf, nil
+}
+
 // inlineThreshold returns a small-file cutoff suitable for performance
 // heuristics (read-cache prefetch sizing, debounce flush timing, write-
 // buffer fast-path bounds). It mirrors the server's inline_threshold once
@@ -2188,10 +2237,7 @@ func (fs *Dat9FS) truncateWritableHandleLocked(fh *FileHandle, newSize int64) (f
 	if newSize == 0 {
 		fh.Dirty.ResetStreamingState()
 		if fh.ShadowSpill {
-			wb := fh.Dirty
-			wb.OnPartFull = func(partIdx int, data []byte) {
-				wb.EvictPart(partIdx)
-			}
+			fs.bindShadowSpillEvictionLocked(fh)
 		}
 		if fh.Streamer != nil {
 			streamer := fh.Streamer
@@ -2246,8 +2292,7 @@ func (fs *Dat9FS) applySQLiteZeroTruncateLocked(fh *FileHandle) bool {
 	fh.Dirty.ResetStreamingState()
 	fh.Dirty.ResetSequentialState(0)
 	if fh.ShadowSpill {
-		wb := fh.Dirty
-		wb.OnPartFull = func(partIdx int, _ []byte) { wb.EvictPart(partIdx) }
+		fs.bindShadowSpillEvictionLocked(fh)
 	}
 	fh.OrigSize = 0
 	// The initiating FD owns the truncate mutation. A sibling remains a
@@ -11223,11 +11268,7 @@ func (fs *Dat9FS) Create(cancel <-chan struct{}, input *gofuse.CreateIn, name st
 	}
 
 	if fh.ShadowSpill {
-		// ShadowSpill mode: shadow is the authoritative data source.
-		// OnPartFull evicts memory immediately — no StreamUploader needed.
-		wb.OnPartFull = func(partIdx int, data []byte) {
-			wb.EvictPart(partIdx)
-		}
+		fs.bindShadowSpillEvictionLocked(fh)
 	} else if fh.WritePolicy != WritePolicyWriteSync && !fs.layerEnabled() && !fs.appendLogNewPathActive(childP) {
 		// Normal mode: attach streaming uploader for sequential write streaming.
 		fh.Streamer = NewStreamUploader(fs.client, childP, expectedRevisionForHandle(fh), fs.remoteRoot())
@@ -11442,11 +11483,7 @@ func (fs *Dat9FS) Open(cancel <-chan struct{}, input *gofuse.OpenIn, out *gofuse
 			}
 
 			if fh.ShadowSpill {
-				// ShadowSpill mode: shadow is the authoritative data source.
-				wb := fh.Dirty
-				wb.OnPartFull = func(partIdx int, data []byte) {
-					wb.EvictPart(partIdx)
-				}
+				fs.bindShadowSpillEvictionLocked(fh)
 			} else if fh.WritePolicy != WritePolicyWriteSync && !fs.layerEnabled() && !fs.appendLogPathConfigured(p) {
 				// Normal mode: attach streaming uploader with OnPartFull wiring.
 				fh.Streamer = NewStreamUploader(fs.client, p, expectedRevisionForHandle(fh), fs.remoteRoot())
--- a/pkg/fuse/dat9fs.go
+++ b/pkg/fuse/dat9fs.go
@@ -3111,11 +3111,23 @@ func (fs *Dat9FS) markHandleRemoteCommittedLocked(fh *FileHandle, revision int64
 	if fh.ZeroBase && fh.Dirty != nil && fh.Dirty.Size() > 0 {
 		fh.ZeroBase = false
 	}
+	fs.rebaselineCommittedDirtyBufferLocked(fh)
 	fs.removeHandleStagingLocked(fh)
 	fh.ShadowReady = false
 	fh.ShadowSpill = false
 	fh.ShadowCommitReady = false
 	fh.ShadowCommitSeq = 0
+}
+
+// rebaselineCommittedDirtyBufferLocked replaces a clean live WriteBuffer that
+// still carries ShadowSpill eviction/restore bookkeeping with a remote-backed
+// buffer. A successful Fsync/Flush must not leave uploadedParts/RestorePart
+// pointing at a shadow that the commit path is about to delete (#900).
+func (fs *Dat9FS) rebaselineCommittedDirtyBufferLocked(fh *FileHandle) {
+	if fh == nil || fh.Dirty == nil || fh.Dirty.HasDirtyParts() || !fh.Dirty.needsCommitRebaseline() {
+		return
+	}
+	fs.rebindCleanWriteBufferToRemoteLocked(fh, fh.Dirty.Size())
 }
 
 func (fs *Dat9FS) seedReadCacheFromShadowGenerationLocked(path string, size int64, revision int64, shadowGen uint64) bool {
@@ -3873,6 +3885,7 @@ func (fs *Dat9FS) finalizeHandleFlushLocked(fh *FileHandle, expectedRevision int
 			fh.OrigSize = size
 			if fh.DirtySeq == 0 {
 				fh.appendLogAdoptCommittedBaseline(fh.BaseRev, size)
+				fs.rebaselineCommittedDirtyBufferLocked(fh)
 			} else {
 				fh.appendLogRebindLayout(fh.BaseRev, size)
 			}
@@ -12454,10 +12467,23 @@ func (fs *Dat9FS) Write(cancel <-chan struct{}, input *gofuse.WriteIn, data []by
 	if fh.Flags&uint32(syscall.O_APPEND) != 0 {
 		writeOffset = fh.Dirty.Size()
 	}
+	writeLen := int64(len(data))
+	if fh.Dirty != nil && writeOffset+writeLen > fh.Dirty.maxSize {
+		source = "dirty-write-error"
+		return 0, gofuse.Status(syscall.EFBIG)
+	}
 
 	// ShadowSpill: write shadow FIRST, before Dirty. If shadow fails, return
 	// EIO without touching Dirty — OnPartFull may evict the part, so writing
 	// Dirty first could lose data if shadow then fails.
+	// Restore evicted Dirty parts before mutating shadow so a restore failure
+	// cannot leave authoritative staging bytes from a Write the caller saw fail.
+	if fh.ShadowSpill && fh.ShadowReady && fs.shadowStore != nil && fh.Dirty != nil {
+		if err := fh.Dirty.EnsurePartsForWrite(writeOffset, writeLen); err != nil {
+			source = "dirty-restore-error"
+			return 0, gofuse.EIO
+		}
+	}
 	if fh.ShadowSpill && fh.ShadowReady && fs.shadowStore != nil {
 		shadowStart := time.Now()
 		unlockShadowWrite := fs.lockHandleRemoteCommitPathLocked(fh)
@@ -12486,7 +12512,10 @@ func (fs *Dat9FS) Write(cancel <-chan struct{}, input *gofuse.WriteIn, data []by
 	n, err := fh.Dirty.Write(writeOffset, data)
 	if err != nil {
 		source = "dirty-write-error"
-		return 0, gofuse.Status(syscall.EFBIG)
+		if errors.Is(err, syscall.EFBIG) {
+			return 0, gofuse.Status(syscall.EFBIG)
+		}
+		return 0, gofuse.EIO
 	}
 	fh.appendLogRecordUserWrite(preWriteSize, writeOffset, int64(n))
 	if fs.appendLogPathConfigured(fh.Path) && !fh.IsNew && n > 0 && writeOffset != preWriteSize {
--- a/pkg/fuse/dat9fs.go
+++ b/pkg/fuse/dat9fs.go
@@ -7531,32 +7531,36 @@ func (fs *Dat9FS) loadUnlinkedHandlePart(fh *FileHandle, offset, length int64) (
 	if fh == nil || !fh.Unlinked || length <= 0 {
 		return nil, false, nil
 	}
+	anonSize := fh.UnlinkedSize
+	if anonSize < 0 {
+		anonSize = 0
+	}
+	if offset >= anonSize {
+		return nil, true, nil
+	}
+	want := length
+	if remaining := anonSize - offset; remaining < want {
+		want = remaining
+	}
+	if want <= 0 {
+		return nil, true, nil
+	}
 	if fh.UnlinkedShadowGen != 0 && fs != nil && fs.shadowStore != nil {
-		buf := make([]byte, length)
+		buf := make([]byte, want)
 		n, err := fs.shadowStore.ReadAtGen(fh.UnlinkedShadowGen, offset, buf)
 		if err != nil && !errors.Is(err, io.EOF) {
 			return nil, true, err
 		}
-		if n < 0 {
-			n = 0
+		if int64(n) != want {
+			return nil, true, fmt.Errorf("unlinked shadow gen %d short read n=%d want=%d off=%d", fh.UnlinkedShadowGen, n, want, offset)
 		}
-		if n > len(buf) {
-			n = len(buf)
-		}
-		return buf[:n], true, nil
+		return buf, true, nil
 	}
 	if fh.UnlinkedData != nil {
-		if offset >= int64(len(fh.UnlinkedData)) {
-			return nil, true, nil
+		if offset >= int64(len(fh.UnlinkedData)) || offset+want > int64(len(fh.UnlinkedData)) {
+			return nil, true, fmt.Errorf("unlinked snapshot short read want=%d have=%d off=%d", want, len(fh.UnlinkedData), offset)
 		}
-		end := offset + length
-		if end > int64(len(fh.UnlinkedData)) {
-			end = int64(len(fh.UnlinkedData))
-		}
-		if end <= offset {
-			return nil, true, nil
-		}
-		return append([]byte(nil), fh.UnlinkedData[offset:end]...), true, nil
+		return append([]byte(nil), fh.UnlinkedData[offset:offset+want]...), true, nil
 	}
 	return nil, false, nil
 }
--- a/pkg/fuse/dat9fs.go
+++ b/pkg/fuse/dat9fs.go
@@ -2784,6 +2784,9 @@ func (fs *Dat9FS) rebindCleanWriteBufferToRemoteLocked(fh *FileHandle, committed
 		if length <= 0 {
 			return nil, nil
 		}
+		if data, handled, err := fs.loadUnlinkedHandlePart(fh, offset, length); handled {
+			return data, err
+		}
 
 		lpCtx, lpCf := context.WithTimeout(context.Background(), fuseTimeout)
 		defer lpCf()
@@ -3204,6 +3207,9 @@ func (fs *Dat9FS) preloadWritableHandle(ctx context.Context, fh *FileHandle) gof
 		}
 		if length <= 0 {
 			return nil, nil
+		}
+		if data, handled, err := fs.loadUnlinkedHandlePart(fh, offset, length); handled {
+			return data, err
 		}
 
 		lpCtx, lpCf := context.WithTimeout(context.Background(), fuseTimeout)
@@ -7377,7 +7383,8 @@ func (fs *Dat9FS) markOpenHandlesUnlinked(ctx context.Context, p string, snapsho
 				continue
 			}
 			fh.Lock()
-			handleNeedsSnapshot := cleanRemoteHandleNeedsUnlinkedSnapshotLocked(fh)
+			handleNeedsSnapshot := cleanRemoteHandleNeedsUnlinkedSnapshotLocked(fh) ||
+				remoteBackedDirtyHandleNeedsUnlinkedSnapshotLocked(fh)
 			if handleNeedsSnapshot {
 				needsSnapshot = true
 				snapshotNeedCount++
@@ -7440,6 +7447,8 @@ func (fs *Dat9FS) markOpenHandlesUnlinked(ctx context.Context, p string, snapsho
 					fh.UnlinkedShadowGen = gen
 					fh.UnlinkedSnapshot = true
 				}
+			} else if snapshotOK && remoteBackedDirtyHandleNeedsUnlinkedSnapshotLocked(fh) {
+				attachUnlinkedSnapshotLocked(fh, snapshot, snapshotShadowGen, size)
 			}
 		} else if snapshotOK {
 			if snapshotShadowGen != 0 && cleanRemoteHandleNeedsUnlinkedSnapshotLocked(fh) {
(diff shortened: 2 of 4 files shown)


qiffang (changes requested):
Reviewed exact head `26795b8647767620dfa14433797b24cf6986934b`.

The code change for the zero-fill mechanism looks sound to me: `ShadowSpill` writes hit the shadow store before `Dirty.Write`, `ensurePart` now restores an evicted part before applying a non-tail overlay, restore failures propagate instead of silently recreating zeros, and Create/Open/truncate/generation-reset rebind the ShadowSpill eviction callback through the same helper. The three new tests are discriminating for the old zero-fill behavior, including the real shadow-store path.

I am still requesting changes because the PR body starts with `Fixes #898`, and current issue evidence does not support auto-closing #898 yet. The latest byte-level update on #898 says the reproduced remote WAL object matched the local shadow byte-for-byte, all WAL reads used `source=shadow-spill`, and the corrupted durable object was the main DB object; the missing link is still the main DB commit-queue upload / checkpoint path, not `ensurePart` alone: https://github.com/mem9-ai/drive9/issues/898#issuecomment-5573134939

The PR body itself also says the original 1GiB WAL e2e is still field confirmation. So as written, merging this PR can auto-close #898 while the observed corruption is still not proven fixed. Please either:

1. change `Fixes #898` to `Refs #898` / `Partially addresses #898` and leave the issue open for the main-DB corruption proof, or
2. attach the original-mount 1GiB WAL e2e evidence proving same-connection count, `integrity_check`, reopen count, and the main DB object path are fixed.

Validation: hosted Code CI / Local E2E / CodeRabbit are green; locally I ran `git diff --check`, gofmt check, `go test -c ./pkg/fuse`, `go vet ./pkg/fuse`, the focused restore/ShadowSpill/generation-reset test set, and the new restore tests under `-race -count=3`. Full local `go test ./pkg/fuse -count=1` still fails only on the known Darwin/rootless mount-permission tests (`TestShouldContinueAfterWaitMountPermissionErrorWhenProbePasses/Fails`).

mornyx:
@qiffang Agreed — this should not auto-close #898.

Updated the PR body and commit message from `Fixes #898` to `Refs #898`. The WAL Dirty zero-fill restore stays in this PR; the remaining main-DB commit-queue / checkpoint corruption stays on the issue.

qiffang (changes requested):
PR body is now `Refs #898`, so I withdraw my earlier objection about auto-closing #898.

There is still a handle-lifecycle blocker in the new `RestorePart` path. After a successful large `ShadowSpill` remote commit, the commit paths clear dirty marks and remove the shadow staging, but leave the live `fh.Dirty` object carrying stale eviction state:

- `Fsync`: `pkg/fuse/dat9fs.go:13776-13803`
- `Flush`: `pkg/fuse/dat9fs.go:13397-13423`
- `syncHandleToRemoteWithoutAppendLogLocked`: `pkg/fuse/dat9fs.go:12937-12955`

`WriteBuffer.ClearDirty()` only clears `dirtyParts` / `touched`; it does not clear `uploadedParts`, `RestorePart`, or `OnPartFull` (`pkg/fuse/write.go:951-957`). `markHandleRemoteCommittedLocked` / `removeShadowPendingStagingGenerationLocked` then clear `ShadowReady` / `ShadowSpill` and remove the shadow generation (`pkg/fuse/dat9fs.go:1547-1565`, `3091-3118`). A later write on the same still-open handle into a previously evicted part still enters `ensurePart`'s `uploadedParts` branch (`pkg/fuse/write.go:436-443`) and calls the stale `RestorePart` callback, which now points at a removed shadow file. `Dat9FS.Write` maps that dirty-write error to `EFBIG` (`pkg/fuse/dat9fs.go:12484-12490`). That means a successful `Fsync`/strict `Flush` can make later same-FD back-writes fail.

Fix gate: after a successful `ShadowSpill` remote commit, rebaseline the open handle onto a clean remote-backed `WriteBuffer` before removing/invalidating the shadow, or otherwise clear the eviction/restorer state (`uploadedParts`, `RestorePart`, `OnPartFull`) while setting `remoteSize` / `appendCursor` / `BaseRev` consistently. Please add a deterministic regression: large `ShadowSpill` fills and evicts a part, `Fsync` or strict `Flush` succeeds and removes shadow staging, then the same open handle writes into the evicted earlier part; the write must succeed and the next commit/read must preserve committed baseline bytes plus the overlay. Cover both `Fsync` and `Flush` if they remain separate commit paths.

Validation on current head `0247f8eb0dc48c515309b9505a8027b47b27d2af`: `git diff --check`, gofmt check, `go test -c ./pkg/fuse`, `go vet ./pkg/fuse`, focused restore/ShadowSpill tests, and focused `-race` all pass. The failure above is not covered by the current tests because they restore before the first remote commit, not after shadow cleanup on a still-open handle.

qiffang (changes requested):
Reviewed exact head `0247f8eb0dc48c515309b9505a8027b47b27d2af` (tree `80708c24b38ecfa58f83b37f1dc7365ffff5d15c`). **REQUEST_CHANGES**.

### 1. A successful ShadowSpill Fsync leaves the open handle pointing at deleted backing data

`bindShadowSpillEvictionLocked` installs a `RestorePart` closure over the active shadow and `EvictPart` records completed parts in `uploadedParts` (`pkg/fuse/dat9fs.go:221-236`, `pkg/fuse/write.go:989-1009`). After a strict remote-durable ShadowSpill commit, the Fsync/Flush/sync paths only call `ClearDirty`; then staging cleanup deletes the captured shadow generation and clears `ShadowSpill` (`pkg/fuse/dat9fs.go:13776-13800`, `13397-13422`, `12937-12955`, `1543-1564`). They do not rebase/reset the live `WriteBuffer`: its evicted parts are still absent, `uploadedParts` still marks them, `RestorePart` still reads the removed shadow, and a newly-created buffer still has `remoteSize == 0`.

I reproduced both consequences deterministically through the real strict `Fsync` path with 64-byte parts and a three-part new ShadowSpill file:

* first `Fsync` succeeds and removes the shadow; a same-handle back-write at offset 8 then returns `EFBIG` because restoration reads the deleted shadow;
* if the same handle instead appends `tail` and calls `Fsync` again, the second uploaded body is 192 zero bytes plus `tail`, silently replacing the already-committed prefix. `CanMaterializeFull` treats `remoteSize == 0` as complete, while `Bytes` zero-fills the missing evicted parts.

This is a WAL lifecycle regression inside the scope of this PR: SQLite keeps the WAL handle open across multiple sync/write cycles. A successful remote commit must either rebuild/rebase the live `WriteBuffer` against the committed revision (including a usable remote loader and cleared spill bookkeeping) or retain generation-backed restore data until the handle is released. Please add discriminating `Fsync -> same-handle back-write` and `Fsync -> same-handle append -> Fsync` tests with actual eviction; the current helper tests stop before the shadow is retired.

### 2. Restore failure mutates the authoritative shadow even though Write reports failure

`Dat9FS.Write` writes ShadowSpill storage first and only then calls `fh.Dirty.Write` (`pkg/fuse/dat9fs.go:12458-12489`). This PR makes `Dirty.Write` able to fail on restore I/O or malformed restored data. In that case the shadow already contains the caller's bytes, but FUSE returns failure; a later flush can persist bytes from a syscall that the application was told did not happen. It also maps every restore failure to `EFBIG`, including ordinary shadow I/O failures.

A FUSE-level injected-restorer test passes 10/10 on the current code while proving the bad state: `fs.Write` fails, yet reading the shadow shows the failed overlay at the requested offset. The isolated `TestWriteBuffer_RestorePartErrorFailsBackWrite` cannot detect this cross-object partial mutation.

Please make restore/validation happen before mutating the authoritative shadow, or roll the shadow back on failure; preserve dirty/shadow consistency and map restore I/O failure to `EIO`. Add a FUSE-level failure regression that checks both returned errno and unchanged shadow contents.

I accept the commit-message change from `Fixes #898` to `Refs #898`: the latest byte-level evidence on #898 says this zero-fill path did not cause the observed main-DB corruption, so I am not treating that separate root cause as this PR's blocker. But the narrower claim that WAL back-write is resolved is still premature until finding 1 is closed.

Existing focused/full `pkg/fuse` tests, race tests, compile, vet, gofmt, CLI build, and hosted checks pass; none currently exercises these post-commit or restore-failure state transitions.

qiffang (changes requested):
Additional review on current head `0247f8eb0dc48c515309b9505a8027b47b27d2af`: I independently confirm the second failure path @dat9-dev2 reported.

`Dat9FS.Write` mutates the authoritative ShadowSpill file before it verifies that the `WriteBuffer` can accept the write (`pkg/fuse/dat9fs.go:12458-12490`). If `fh.Dirty.Write` then fails while restoring an evicted part (`pkg/fuse/write.go:436-443`, `477-486`), the syscall returns failure, but the shadow already contains those bytes. The write-sync snapshot/rollback path does not cover this branch because the function returns immediately from the Dirty write error before `syncWriteHandleToRemoteLocked` and `restoreFailedWriteSyncLocked` run. A later Flush/Fsync can stage/upload that mutated shadow and commit bytes from a failed write.

This is an atomicity violation at the FUSE write boundary: a failed user-visible Write must not leave authoritative staging state that can be committed later. It also misreports restore/read failures as `EFBIG`, masking I/O/corruption causes as a size-limit failure.

Fix gate: for ShadowSpill back-writes into evicted parts, restore/validate the target part before mutating the shadow, or make the shadow mutation rollback generation-scoped on any Dirty failure. Add a FUSE-level regression where the shadow write succeeds, Dirty restore fails, the write returns an error, and a subsequent Flush/Fsync proves the failed bytes were not committed. The stale-restorer-after-successful-Fsync blocker from my previous review remains open as well.

mornyx:
@qiffang Thanks — both post-commit lifecycle issues were real.

Pushed `4400bfe3` on this PR:

1. **Same-FD after successful ShadowSpill commit.** `markHandleRemoteCommittedLocked` / `finalizeHandleFlushLocked` now rebaseline a clean live `WriteBuffer` onto a remote-backed buffer (`remoteSize`, `LoadPart`, cleared `uploadedParts` / `RestorePart` / `OnPartFull`) *before* staging cleanup deletes the shadow. Same-handle back-write loads the committed part instead of hitting a stale restorer; a later append+Fsync keeps the committed prefix.

2. **Restore failure vs shadow mutation.** `EnsurePartsForWrite` restores evicted Dirty parts *before* `shadow.WriteAt`. Restore I/O returns `EIO` rather than `EFBIG`. A failed Write no longer leaves overlay bytes in the authoritative shadow for a later Fsync to commit.

Tests:
- `TestMarkHandleRemoteCommittedRebaselinesEvictedSpillBuffer` (shared Fsync/Flush helper)
- `TestShadowSpillFsyncRebaselineAllowsSameHandleBackWrite`
- `TestShadowSpillFsyncThenAppendDoesNotZeroPrefix`
- `TestShadowSpillRestoreFailureDoesNotMutateShadow`

Still `Refs #898`; this does not claim the main-DB commit-queue corruption is fixed.

qiffang (changes requested):
Current head `90b3108386df282b96be176b1e09be132f7fe60d` is still not mergeable.

The two previously reported ShadowSpill issues are addressed in the large/evicted-part path: successful remote commit now attempts to rebaseline stale spill bookkeeping, and `Write` restores overlapping evicted parts before mutating the shadow. The new regressions for those paths are useful and pass locally.

However, the new rebaseline is too broad and now breaks existing writable-handle behavior. `markHandleRemoteCommittedLocked` unconditionally calls `rebaselineCommittedDirtyBufferLocked` after a successful commit (`pkg/fuse/dat9fs.go:3114`), and `needsCommitRebaseline` returns true whenever `RestorePart` or `OnPartFull` is non-nil (`pkg/fuse/write.go:498-505`). That means even a clean, fully materialized small ShadowSpill buffer with no evicted/uploaded parts is replaced by a remote-backed lazy buffer (`pkg/fuse/dat9fs.go:2748-2815`). The next same-open-handle write at offset 0 then has to perform a remote range GET; if that GET is unavailable or returns 404, the write fails with `EIO` even though the handle already had the committed bytes locally.

This is already caught by existing tests on this head:

```text
$ go test ./pkg/fuse -run '^TestFsyncShadowSpillCommitRefreshesRevisionAndClearsPending$' -count=1 -v
=== RUN   TestFsyncShadowSpillCommitRefreshesRevisionAndClearsPending
    writeback_test.go:3628: second Write: 5=input/output error
--- FAIL: TestFsyncShadowSpillCommitRefreshesRevisionAndClearsPending (0.01s)
```

Hosted Code CI is failing the same package, including this failure, and also `TestWriteSyncWriteAfterUnlinkDoesNotResurrectPath`:

```text
--- FAIL: TestWriteSyncWriteAfterUnlinkDoesNotResurrectPath
    dat9fs_test.go:24634: Write after unlink: 5=input/output error
--- FAIL: TestFsyncShadowSpillCommitRefreshesRevisionAndClearsPending
    writeback_test.go:3628: second Write: 5=input/output error
FAIL github.com/mem9-ai/drive9/pkg/fuse
```

Fix gate: separate "drop stale spill callbacks/eviction state" from "replace the whole clean buffer with a remote-backed lazy buffer". When a clean committed buffer can still materialize the full data and has no evicted/uploaded parts, preserve that in-memory buffer and clear only `RestorePart`/`OnPartFull`/uploaded bookkeeping. Only rebind to remote when the committed clean buffer cannot safely provide the bytes. Keep the existing failing regressions green, and retain/add a no-eviction same-handle post-Fsync rewrite/append case so this does not regress again.

Validation I ran on the current head:
- gofmt check on touched fuse files passed
- `go test -c ./pkg/fuse` passed
- `go vet ./pkg/fuse` passed
- new focused #900 tests and adjacent restore/race tests passed
- broader adjacent ShadowSpill run fails on `TestFsyncShadowSpillCommitRefreshesRevisionAndClearsPending`

mornyx:
@qiffang Agreed — rebasing every ShadowSpill handle just because Create installed `RestorePart`/`OnPartFull` was too broad.

`d346d4d2` splits that: we only replace the live buffer with a remote-backed loader when `uploadedParts` is non-empty (actual eviction). Otherwise we keep the in-memory image and only clear `RestorePart`/`OnPartFull`. That should green `TestFsyncShadowSpillCommitRefreshesRevisionAndClearsPending` and `TestWriteSyncWriteAfterUnlinkDoesNotResurrectPath`, without dropping the evicted-part Fsync→same-handle tests.

The large/evicted path and restore-before-shadow changes are unchanged.

qiffang (approved):
Re-reviewed exact head `d346d4d2f31f78b2a4e5d686c1bb27fca5200955`. The two ShadowSpill correctness blockers are closed.

- Successful remote commit now remote-rebinds only buffers that actually evicted parts (`uploadedParts` non-empty), so their next same-FD write loads the committed object after shadow cleanup. Buffers that retained a complete local image keep those bytes and only drop stale `RestorePart`/`OnPartFull` callbacks, preserving write-after-Fsync and write-after-unlink semantics.
- ShadowSpill back-write now restores every overlapping evicted part before mutating the authoritative shadow; restore failures return EIO and cannot leave bytes from a failed syscall staged for a later flush.
- The new state-machine regressions cover Fsync -> same-FD back-write, Fsync -> append -> Fsync without zeroing the committed prefix, committed eviction rebaseline, and restore-failure atomicity. The two pre-existing tests that failed on `90b3108` are green again.

Local validation: focused tests 20x, focused race tests 10x, package compile, vet, gofmt, and diff-check pass. Full `pkg/fuse` only hits the two known macOS mount-permission environment failures. CodeRabbit is green; exact-head Code CI and Local E2E are still running and remain merge gates. No remaining code-level blocker found.

qiffang (changes requested):
I reviewed exact head `d346d4d2f31f78b2a4e5d686c1bb27fca5200955`. The prior restore-order bug and the over-broad rebaseline regression are fixed for the covered paths, but there is one remaining open-FD data contract blocker.

**Blocking: evicted ShadowSpill handles lose anonymous-fd semantics after `Fsync` then `Unlink`.**

For a clean buffer that actually evicted parts, `rebaselineCommittedDirtyBufferLocked` necessarily calls `rebindCleanWriteBufferToRemoteLocked` (`pkg/fuse/dat9fs.go:3126-3134`). That replaces the handle's clean local/shadow-backed state with a lazy `LoadPart` that reads the committed pathname from remote (`2748-2814`), and the successful commit then clears `fh.ShadowSpill`/`fh.ShadowReady` and removes shadow staging (`3090-3120`).

If the path is then unlinked while the fd remains open, `markOpenHandlesUnlinked` does not snapshot this handle: the remote snapshot path only considers `cleanRemoteHandleNeedsUnlinkedSnapshotLocked`, which explicitly requires `fh.Dirty == nil` (`7351-7421`, `7464-7472`), and the dirty-handle branch only pins shadow when `fh.ShadowSpill` is still true (`7428-7443`). After this PR's rebaseline, the handle is `Dirty != nil`, `ShadowSpill == false`, and has no shadow to pin. The remote DELETE succeeds, then the same open fd writes back into an old part; `Dirty.Write` calls the lazy `LoadPart`, which now GETs the deleted path and returns `EIO` (`12487-12530`).

I reproduced this deterministically with a review-only test: create a large small-part ShadowSpill file, `Fsync` it so the clean buffer is remote-rebound, `Unlink` the path, then write to the same fd at offset 8. Current head returns `5=input/output error`; POSIX anonymous fd semantics require that write to succeed and stay local, and existing tests already enforce the same contract for write-sync open fds. This means the required large/evicted path still has the same class of post-commit rebaseline problem when unlink enters the state machine.

Fix gate: when a clean remote-backed Dirty handle is about to be unlinked, snapshot/pin the committed data before DELETE as an anonymous backing, including the large case that may need a shadow snapshot; or otherwise keep a durable private backing for the open fd before removing the pathname. Add a regression for: large ShadowSpill with evicted parts -> `Fsync` success -> `Unlink` while fd open -> same-fd back-write/read succeeds and does not resurrect the remote path.

Validation I ran: hosted Code CI / Local E2E / CodeRabbit are green on this head; local `git diff --check`, gofmt check, `go test -c ./pkg/fuse`, `go vet ./pkg/fuse`, focused #900 state-machine tests under `-race -count=3`, and the same focused suite under `-count=20` passed. The review-only unlink regression above fails on the current head; the temporary test file was removed afterward.

qiffang (changes requested):
Reviewed exact head `d346d4d2f31f78b2a4e5d686c1bb27fca5200955`. **REQUEST_CHANGES / blocking**.

The main fixes in this PR look directionally right: evicted `ShadowSpill` parts are restored before overlay, post-commit live buffers are rebaselined when they actually evicted parts, restore failures return `EIO` before mutating the authoritative shadow, and the previously reported same-FD Fsync/append cases are covered. Hosted Code CI / Local E2E / CodeRabbit are green, and my local targeted/race plus `go test ./pkg/fuse -skip WaitMount -count=1` passed.

But there is still a restore boundary that can make valid writes fail after a non-zero shrink/truncate. `Dat9FS.Write` now calls `fh.Dirty.EnsurePartsForWrite()` before mutating the `ShadowSpill` shadow (`pkg/fuse/dat9fs.go` around the ShadowSpill write path). If a large spill handle had evicted part N, then shrinks to a non-zero size exactly at the boundary before part N, a later append at that EOF first tries to restore part N from the current logical `wb.totalSize`. `restoreWriteBufferPartFromShadow()` currently treats `partStart >= wb.totalSize` as an error, so the FUSE write returns `EIO` even though the part is beyond the retained file size and should be treated as a new zero-filled part.

I reproduced it with a small-part FUSE regression shape:

```text
ShadowSpill write 3 parts -> part 1 evicted
truncate/shrink to 1 part (non-zero)
append "tail" at the new EOF (offset == partSize)
actual: Write returns EIO
expected: append succeeds; no old truncated bytes are restored
```

This matches the latest CodeRabbit observation, but I verified it through the actual `Dat9FS.Write` ordering, not only static analysis. Fix gate: split “part beyond current logical size” from “restore failure”. A beyond-size evicted part after shrink should return “not restored” so `ensurePart()` can create the new zero-filled part, while real shadow I/O/short-read errors must still fail safe. Please add a regression covering `ShadowSpill` non-zero shrink/truncate followed by append at the evicted boundary (and preferably the partial-retained last-part case to prove retained bytes are preserved).

This does **not** require closing the remaining #898 main-DB checkpoint corruption in this PR; the PR is correctly scoped as `Refs #898`. The blocker is inside the narrower WAL/ShadowSpill restore semantics this PR changes.

qiffang (changes requested):
Re-reviewed exact head `d346d4d2f31f78b2a4e5d686c1bb27fca5200955`; this supersedes my earlier approval. The two original ShadowSpill blockers and the intermediate no-eviction regression are fixed, but one same-FD lifecycle blocker remains.

After an evicted ShadowSpill buffer is successfully committed, `rebaselineCommittedDirtyBufferLocked` replaces it with a clean remote-backed `Dirty` buffer and removes the shadow. For files larger than `maxPathTruncateInMemoryBytes` (64 MiB), `snapshotOpenHandlesBeforeUnlink` intentionally skips the in-memory snapshot. The later `markOpenHandlesUnlinked(..., snapshotRemote=true)` also does not snapshot this handle because `cleanRemoteHandleNeedsUnlinkedSnapshotLocked` requires `Dirty == nil`, while the dirty-handle branch only pins a backing when `ShadowSpill` is still true. This post-Fsync handle is exactly `Dirty != nil && ShadowSpill == false`, so Unlink deletes the only remote baseline without creating anonymous backing. A subsequent same-FD back-write/read lazy-loads the deleted path and returns EIO, violating open-unlink semantics.

I reproduced this deterministically 10/10 with a review-only test that constructs the exact post-Fsync evicted remote-backed state above 64 MiB, calls real `Unlink`, then writes offset 8 through the still-open handle; current code returns `5=input/output error`. The temporary test was removed and the worktree is clean.

Merge gate: before deleting the path, provide generation-private anonymous backing for clean remote-backed `Dirty` handles too (small files may use memory; large files need a shadow snapshot/pin or equivalent), and add a discriminating regression: evicted ShadowSpill -> Fsync -> Unlink while FD remains open -> same-FD back-write/read succeeds, while the remote pathname remains absent. Exact-head Code CI, Local E2E, and CodeRabbit are green but do not cover this state transition.

mornyx:
@qiffang Both remaining blockers should be closed on this push.

1. **Fsync → Unlink → same-FD write.** `markOpenHandlesUnlinked` now snapshots clean remote-backed Dirty handles (`LoadPart` + `remoteSize > 0`, no `ShadowSpill`) before DELETE. `LoadPart` prefers that anonymous backing (`UnlinkedData` / pinned shadow gen) so a later overlay does not GET the deleted path. Regression: `TestShadowSpillFsyncUnlinkSameHandleWriteStaysLocal`.

2. **Non-zero shrink then append at an evicted boundary.** Truncate forgets `uploadedParts` whose start is `>=` the new size, and `ensurePart` treats a beyond-size evicted part as a new zero part instead of a restore error. Regression: `TestWriteBuffer_ShrinkDropsEvictedPartsBeyondSize` and `TestShadowSpillShrinkThenAppendAtEvictedBoundary`.

Still `Refs #898`.

qiffang (changes requested):
Blocking finding on current head `b44e071724443fc98570fa101020f4faa84ace14`:

`Fsync -> Unlink -> Read` is still broken for large evicted `ShadowSpill` handles before any post-unlink write. The new unlink fix snapshots remote-backed dirty handles into `UnlinkedShadowGen` (`markOpenHandlesUnlinked`), and `Dirty.LoadPart` knows how to read that private backing. But the FUSE `Read` path does not use it for the clean remote-backed dirty state: `readUnlinkedData` only handles `UnlinkedData`, and the later unlinked-shadow branch is gated on `fh.Unlinked && fh.Dirty == nil`. For the post-Fsync rebaseline state this handle is `Dirty != nil`, `DirtySeq == 0`, `HasDirtyParts == false`, `UnlinkedShadowGen != 0`, so `Read` falls through to the remote path after DELETE and returns ENOENT/EIO instead of serving the anonymous fd bytes.

I reproduced this with a review-only test on the exact head: write a file larger than `maxPathTruncateInMemoryBytes` with small ShadowSpill parts, `Fsync`, `Unlink` while the fd stays open, then `Read` the same fd before any new write. Current result is `Read after unlink before write: 2=no such file or directory`. The existing new regression only writes first and then inspects `fh.Dirty` directly, so it does not cover this syscall path.

Fix gate: make clean remote-backed dirty open-unlinked handles serve `UnlinkedShadowGen`/`UnlinkedData` from FUSE `Read` as well as from `WriteBuffer.LoadPart`, and add a regression for `large ShadowSpill evicted -> Fsync -> Unlink -> same-FD read before write succeeds`, plus keep the no remote resurrection assertion.

Validation: hosted Code CI and CodeRabbit are green; Local E2E was still running when I checked. Locally passed `git diff --check`, gofmt check, `go test -c ./pkg/fuse`, `go vet ./pkg/fuse`, and the existing focused ShadowSpill tests under normal and race runs. The failing review-only test was removed; worktree is clean.

qiffang (changes requested):
Fresh review of exact head `b44e071724443fc98570fa101020f4faa84ace14`.

The previous post-Fsync large ShadowSpill write-back failure is substantially fixed: a clean remote-backed `WriteBuffer` can now receive an unlink snapshot, `LoadPart` can read that private backing, and the shrink bookkeeping is corrected. The existing focused/race tests all pass.

Two adjacent open-unlink gaps remain blocking:

1. A clean open-unlinked remote-backed dirty handle cannot read its new large snapshot before another write loads a part. The early FUSE `Read` snapshot path only handles `UnlinkedData`; the later `UnlinkedShadowGen` branch requires `fh.Dirty == nil`. The exact post-Fsync state is `Dirty != nil`, no dirty parts, and `UnlinkedShadowGen != 0`, so it falls through to the deleted remote path and returns ENOENT/EIO (`pkg/fuse/dat9fs.go:11733`, `pkg/fuse/dat9fs.go:11914`, `pkg/fuse/dat9fs.go:11961`). The new regression writes first and then directly inspects `Dirty`, so it never exercises top-level `Read` in this state. Require `Fsync -> Unlink -> same-FD Read before any new write` to return the anonymous snapshot without a remote GET.

2. The large-snapshot attachment has a mutation TOCTOU. `markOpenHandlesUnlinked` counts eligible clean handles, releases every `fh` lock, streams the entire remote object, then re-locks each handle and attaches only if `remoteBackedDirtyHandleNeedsUnlinkedSnapshotLocked` is still true (`pkg/fuse/dat9fs.go:7381`, `pkg/fuse/dat9fs.go:7394`, `pkg/fuse/dat9fs.go:7431`, `pkg/fuse/dat9fs.go:7450`). A concurrent same-FD write can complete during that network window and make the predicate false. The handle is still marked unlinked, but receives neither `UnlinkedData` nor `UnlinkedShadowGen`; after DELETE, any untouched unloaded part still depends on the removed pathname. I reproduced this deterministically 3/3 with a >64 MiB remote-backed dirty handle: block the unlink snapshot GET, complete an offset-8 write, release the snapshot, let Unlink return OK, then read an untouched part. The read returns EIO. The pre-counted snapshot pin also has no handle owner in this interleaving, so the retired generation leaks.

Unlink must not complete unless every open handle that depends on the remote baseline has generation-private backing consistent with its current mutation state. Fence/revalidate mutations around snapshot creation and retry or fail closed when state changes; attaching the baseline must preserve concurrent dirty overlays. Add a blocking-snapshot/concurrent-write regression that asserts the write remains visible, untouched baseline reads succeed after DELETE, the path is not resurrected, and the snapshot generation is released on final `Release`.

Validation: PR tests passed 20x, race passed 5x, the existing large-open-unlink shadow test passed 3x, fuse compile/vet/gofmt/diff-check passed, and the review-only tests were removed; worktree is clean. Exact-head Code CI, Local E2E, and CodeRabbit are green but miss both transitions.

mornyx:
@qiffang The remaining hole was FUSE `Read`, not `Write`/`LoadPart`.

Post-Fsync rebaseline keeps `Dirty != nil`. Large Unlink pins `UnlinkedShadowGen`, but `Read` only served that backing when `Dirty == nil`, so a same-FD read *before* any write fell through to GET-after-DELETE (`ENOENT`).

`Read` now serves `UnlinkedShadowGen` for unlinked remote-backed dirty handles before the remote path. Regressions: `TestUnlinkedRemoteBackedDirtyReadUsesPinnedShadowGen` (pinned-gen / no remote GET) and `TestShadowSpillFsyncUnlinkSameHandleWriteStaysLocal` now also `Read`s before the overlay write.

Still `Refs #898`.

mornyx:
@qiffang The 17:55 TOCTOU is addressed on this push.

A concurrent same-FD write during the unlink snapshot GET used to fail `remoteBackedDirtyHandleNeedsUnlinkedSnapshotLocked` (`HasDirtyParts` / `DirtySeq`), so the handle was marked `Unlinked` with neither `UnlinkedData` nor `UnlinkedShadowGen`. Attach now follows “still depends on the committed pathname via `LoadPart`”, not “still clean”. Dirty overlays stay in `Dirty`; `Read` prefers them over the anonymous snapshot when present. Unused large-snapshot pins are dropped if no handle takes the generation.

Regression: `TestUnlinkSnapshotConcurrentWriteKeepsOverlayAndBaseline` (block snapshot GET, write offset 8, then assert overlay + untouched part, no remote resurrection).

The earlier `Read`-before-write hole remains covered by `a93e4203`. Still `Refs #898`.

qiffang (approved):
Reviewed exact head `fdba7a663f580acd123885779e12840b1010b96f`.

No blocking findings from my seat. The previous `Fsync -> Unlink` open-FD bug is closed: clean remote-backed dirty handles now serve `UnlinkedShadowGen`/anonymous backing on read, and the snapshot attach path revalidates enough to keep baseline + overlay when a same-FD write races with the remote snapshot. The large `snapshotUnlinkedRemoteToShadow` path also passed my review-only reproducer for blocked snapshot + concurrent write; temp file was removed and the worktree is clean.

Validation: GitHub Code CI, Local E2E, and CodeRabbit are green on this exact head. Locally I reran `git diff --check`, gofmt check, `go test -c ./pkg/fuse`, `go vet ./pkg/fuse`, focused ShadowSpill/unlink tests `-count=10`, and the same focused set under `-race -count=3`; all passed.

Scope note remains: PR correctly uses `Refs #898`; it should not be treated as proof that the broader main-db/#896 checkpoint problem is solved.

qiffang (changes requested):
Adversarial review on exact head `fdba7a663f580acd123885779e12840b1010b96f`: **REQUEST_CHANGES**.

The two previous open-unlink blockers are addressed in the main state machine:

- `Read` now serves `UnlinkedShadowGen` for unlinked remote-backed dirty handles before falling through to the deleted remote pathname, so `Fsync -> Unlink -> same-FD Read before write` is covered.
- `markOpenHandlesUnlinked` now attaches the unlink snapshot based on whether the handle still depends on the committed pathname via `LoadPart`, even if a concurrent same-FD write dirtied the buffer during the snapshot GET. The new concurrent-write regression covers overlay + untouched baseline + no remote resurrection.

However, there is still a data-integrity blocker in the new private-backing loader.

`loadUnlinkedHandlePart()` treats a short `ReadAtGen` from `UnlinkedShadowGen` as success and returns `buf[:n]` with `nil` error (`pkg/fuse/dat9fs.go:7534-7546`). That path is used by `WriteBuffer.LoadPart` when `EnsureLoaded()` or a post-unlink write needs an unloaded baseline part. If the private backing generation is shorter than the requested part, the short slice is accepted as a loaded part; later `WriteBuffer.ReadAt()` zero-fills the missing bytes inside the logical file. This converts an incomplete/corrupt anonymous backing into silent zero data instead of failing closed.

I reproduced this on the exact head with a review-only unit test: create a pinned unlinked shadow generation containing 3 bytes, set the remote-backed dirty buffer size/remoteSize to 8, call `EnsureLoaded(0)`, then read the buffer. Current result accepts the short backing and returns `6162630000000000` instead of returning an error. The top-level `Read` path already treats short `ReadAtGen` as `EIO`; the `LoadPart` path must enforce the same invariant.

Fix gate: make `loadUnlinkedHandlePart()` fail when `UnlinkedShadowGen` cannot return the full requested `length` for a range that is still inside the anonymous file size. Do not let `WriteBuffer` store a short private-backing part that will be zero-filled later. Add a regression for short/corrupt `UnlinkedShadowGen` backing through the `LoadPart` / `EnsureLoaded` path, while keeping the current `Read`-before-write and concurrent unlink snapshot regressions green.

Validation: exact-head hosted Code CI / Local E2E are green; locally, the focused ShadowSpill/unlink tests, focused race runs, diff/gofmt checks, and broader `pkg/fuse` run all pass. The failing review-only short-backing test was removed and the worktree is clean.

mornyx:
@qiffang Agreed — that short-read hole is in the new private-backing loader, not a pre-existing path.

`loadUnlinkedHandlePart` now fails if `UnlinkedShadowGen` (or `UnlinkedData`) cannot supply the full requested range that still lies inside the anonymous file size, so `EnsureLoaded`/`LoadPart` cannot store a truncated part that `ReadAt` would later zero-fill. Top-level `Read` already treated short `ReadAtGen` as EIO; this matches that invariant.

Regression: `TestLoadUnlinkedHandlePartShortShadowGenFailsClosed`. The Read-before-write and concurrent unlink-snapshot tests stay green.

Still `Refs #898`.

qiffang (approved):
Reviewed exact head `c7c6c50e8798490a10cf2a407e8059eba064634d`.

The previous private-backing short-read blocker is closed. `loadUnlinkedHandlePart()` now computes the requested range inside `UnlinkedSize` and fails closed when `UnlinkedShadowGen`/`UnlinkedData` cannot provide the full requested bytes, so `WriteBuffer.LoadPart` / `EnsureLoaded` can no longer store a truncated baseline that later reads as zero-filled data. The new `TestLoadUnlinkedHandlePartShortShadowGenFailsClosed` covers the `UnlinkedShadowGen` path, and the prior `Fsync -> Unlink -> same-FD Read`, concurrent unlink-snapshot, shrink/append, and ShadowSpill restore regressions stay green.

Validation on this exact head: GitHub Code CI, Local E2E, and CodeRabbit are green. Locally I ran `git diff --check`, gofmt check on changed Go files, `go vet ./pkg/fuse`, `make lint`, the focused ShadowSpill/unlinked tests, focused `-race -count=3`, and `go test ./pkg/fuse -skip WaitMount -count=1`; all passed.

Scope note: this PR still correctly uses `Refs #898`; it fixes the WAL/ShadowSpill evicted-part and open-unlink safety issues covered here, but should not be treated as proof that the broader main-DB/#896 checkpoint corruption root cause is fully closed.

qiffang (approved):
Adversarial re-review on exact head `c7c6c50e8798490a10cf2a407e8059eba064634d`: **APPROVE**.

The previous private-backing short-read blocker is closed. `loadUnlinkedHandlePart()` now derives the requested range from `UnlinkedSize`, requires `UnlinkedShadowGen`/`UnlinkedData` to provide the full in-file range, and returns an error instead of storing a short part that `WriteBuffer.ReadAt()` would later zero-fill. The new `TestLoadUnlinkedHandlePartShortShadowGenFailsClosed` covers the `LoadPart` / `EnsureLoaded` path that was missing before.

I also rechecked the adjacent open-unlink fixes: `Read` serves pinned `UnlinkedShadowGen` for remote-backed dirty handles before falling back to the deleted remote path; the concurrent unlink-snapshot case now attaches private backing even if a same-FD write dirties the handle during the snapshot GET, while preserving the overlay and avoiding remote resurrection. The shrink/truncate boundary append fix remains covered.

Validation: exact-head Code CI and Local E2E are green. Locally, diff/gofmt/compile/vet checks, the focused ShadowSpill/unlink tests, focused race runs, and broader `pkg/fuse` coverage all pass. I found no remaining blocker in the PR-scoped WAL/ShadowSpill restore semantics. This remains correctly scoped as `Refs #898`; it does not claim to close the separate main-DB commit-queue/checkpoint corruption investigation.

qiffang (changes requested):
Re-reviewed exact head `c7c6c50e8798490a10cf2a407e8059eba064634d`. The prior immediate-read and concurrent-Write-only unlink failures are fixed, and the new short private-backing check correctly fails closed. One adjacent data-integrity race remains:

**A successful concurrent Fsync during the unlink snapshot window is lost from the open file description.** `markOpenHandlesUnlinked` captures the remote snapshot while holding neither `fh.mu` nor `remoteCommitLock` (`pkg/fuse/dat9fs.go:7375-7421`), then later re-locks the handle and unconditionally attaches that captured snapshot when the handle is still remote-backed (`pkg/fuse/dat9fs.go:7431-7455`). The relaxed predicate intentionally accepts state changes during the GET (`pkg/fuse/dat9fs.go:7493-7511`). A same-FD `Write` followed by strict `Fsync` can therefore commit while the snapshot GET is blocked: Fsync lands the new bytes and rebaselines the handle, then unlink attaches the older pre-Fsync snapshot and deletes the newly committed path. Subsequent reads from the still-open FD return the old bytes, losing data whose Fsync already returned success.

I reproduced this deterministically 3/3 on the current head: seed and Fsync 192 bytes, block unlink's snapshot GET after it copied the old body, write a two-byte overlay and Fsync it successfully, release the old snapshot, let unlink return OK, then read the same FD. Remote state before DELETE contained the overlay, but the FD returned the original bytes (`8888...` instead of `...aabb...`). The committed `TestUnlinkSnapshotConcurrentWriteKeepsOverlayAndBaseline` only performs `Write` during the window and leaves the dirty overlay in memory; it does not cover a concurrent Fsync that clears/rebaselines that overlay before attachment.

Merge gate: fence/revalidate the remote snapshot against handle mutation plus committed revision before attaching it. If a concurrent Fsync lands during the snapshot, retry from that committed generation (or fail unlink closed); never attach a pre-Fsync image after Fsync succeeded. Add a regression: blocked snapshot GET -> same-FD Write+Fsync -> resume Unlink -> open-FD read must contain the fsynced overlay, with no path resurrection and correct snapshot cleanup. Cover the large pinned-shadow path as well because the same preflight/attach split exists there.

Validation on this exact head: committed focused unlink/ShadowSpill tests pass 10x, focused race tests pass 3x, compile/vet/gofmt/diff-check pass. Code CI, Local E2E, and CodeRabbit are green. The review-only regression was removed and the worktree is clean.

Issue #876, opened 2026-08-31
FUSE: SQLite WAL workload acknowledges 1000 commits but persists only 991 rows


mashenjun:
#### Confirmed environment

| Component | Value |
|---|---|
| CSI image | `ghcr.io/drive9-ai/drive9-csi:drive9-452ed07-csi-667d57f` |
| Drive9 CLI/FUSE | `452ed0770feddabc7add48a3444085134a1276a7` |
| Runtime | gVisor / `runsc`, `directfs=true` |
| Architecture | linux/arm64 |
| SQLite | 3.40.1; reproduced again with 3.46.1 |
| SQLite settings | WAL, `synchronous=FULL`, `wal_autocheckpoint=1000` |
| Endpoint | `https://aws-us-east-1.drive9.ai` |
| Mount | reproduced with interactive durability and explicit `durability=fsync` |

Use a fresh Drive9 PVC and a unique DB path:

```python
import os
import sqlite3

db_path = "/workspace/sqlite-checkpoint-unique/repro.db"
os.makedirs(os.path.dirname(db_path), exist_ok=False)

conn = sqlite3.connect(db_path)
assert conn.execute("PRAGMA journal_mode=WAL;").fetchone()[0].lower() == "wal"
assert conn.execute("PRAGMA synchronous;").fetchone()[0] == 2
assert conn.execute("PRAGMA wal_autocheckpoint;").fetchone()[0] == 1000

conn.execute("CREATE TABLE items (id INTEGER PRIMARY KEY, val TEXT)")
conn.commit()

for i in range(1000):
    conn.execute("INSERT INTO items (val) VALUES (?)", (f"val-{i}",))
    conn.commit()

conn.close()

conn2 = sqlite3.connect(db_path)
print(conn2.execute("SELECT COUNT(*), MIN(id), MAX(id) FROM items").fetchone())
print(conn2.execute("PRAGMA integrity_check").fetchone()[0])
conn2.close()
```

After Drive9 asynchronous work settles, unmount and verify through a fresh read-only consumer of the same PVC:

```python
import sqlite3

db_path = "/workspace/sqlite-checkpoint-unique/repro.db"
conn = sqlite3.connect("file:" + db_path + "?mode=ro&immutable=1", uri=True)
print(conn.execute("SELECT COUNT(*), MIN(id), MAX(id) FROM items").fetchone())
print(conn.execute("PRAGMA integrity_check").fetchone()[0])
conn.close()
```

[...]

PR #901, opened 2026-09-07
fix(fuse): rebase grown same-path commits onto landed revision (#896)

### LINKED ISSUE #896: FUSE interactive: 1GiB SQLite kvtest init acknowledges success then persists a 4KiB empty DB (revision conflict)
## Bug Report

Interactive FUSE writeback can acknowledge a ~1GiB SQLite `kvtest init` and then persist only a 4096-byte empty database. The next process sees `no such table: kv`. The same 1GiB workload completes on `--durability fsync`.

This is remote data loss on the writer mount after a successful local init, not a stale local read of a correct remote object.

#### Environment

| Component | Value |
|---|---|
| Host | Linux x86_64 EC2, real FUSE (`/dev/fuse`) |
| SQLite | official `kvtest` from sqlite 3.54.0 amalgamation |
| Durability | `--durability interactive` |

Default `--write-cache-size-mb 1024` rejects the 1GiB init with:

```text
write-back quota rejected: ... required_bytes=1208307712 quota_bytes=1073741824 reason=byte_quota_exceeded
ERROR: database create failed: disk I/O error
```

That quota is a documented limit. Raise it so the init is allowed (this does not fix the loss):

```bash
drive9 mount --mode=fuse --foreground --durability interactive --profile none \
  --append-log='**/*-wal' --append-log='**/*-journal' \
  --write-cache-size-mb 8192 \
  :/kvbench ./d9
```

On the mount, run official kvtest against a unique path (262144 × 4096-byte blobs ≈ 1GiB):

```bash
./kvtest init ./d9/kv/kv-1GiB.db --count 262144 --size 4096
# init exits 0 in ~50s

./kvtest run ./d9/kv/kv-1GiB.db --count 3000 --max-id 262144 --random --jmode wal --multitrans
```

Then stat the main DB and open it with a new SQLite connection (or `kvtest run` as above).

Control: the same `kvtest init` + `kvtest run` on the same server with `--durability fsync` (and default 1GiB write-cache quota) completes: init 42s, random read ~142ms/blob, update ~135ms/blob.

- `kvtest init` exit 0
- `kv-1GiB.db` size ≈ 1GiB
- `kvtest run` / a new `sqlite3` connection sees table `kv` and 262144 rows
- Remote object after commit-queue drain matches that image

`kvtest init` returned 0. The immediately following `kvtest run` failed:

```text
ERROR: cannot prepare query: no such table: kv
```

Visible files after init:

```text
kv-1GiB.db      4096 bytes
kv-1GiB.db-wal     0 bytes
sqlite_master      empty
```

Mount log (commit queue):

```text
12:11:38 commit queue: successfully uploaded /kv/kv-1GiB.db (4096 bytes, rev=1)
12:12:33 commit queue: conflict committing /kv/kv-1GiB.db at base revision 1, attempting auto-resolve
12:12:33 commit queue: auto-resolving conflict for /kv/kv-1GiB.db via LWW (base rev 1 → server rev 1)
12:12:34 commit queue: auto-resolve LWW re-upload failed for /kv/kv-1GiB.db: revision conflict
12:12:34 commit queue: terminal failure for /kv/kv-1GiB.db — local data preserved for manual recovery
12:12:34 commit queue: successfully uploaded /kv/kv-1GiB.db (4096 bytes, rev=2)
```

SQLite creates a one-page (4096-byte) empty DB, then grows the same path to ~1GiB inside one `BEGIN`/`COMMIT`. Interactive fsync only stages the shadow; the commit queue asynchronously uploads generations of that path. The empty header lands as rev=1. The 1GiB payload then CAS-fails at expected revision 1. LWW also 409s. A later 4096-byte upload becomes rev=2. The durable object is the empty header, not the init payload.

The same 100MiB interactive kvtest on this mount also logged repeated `conflict committing /kv/kv-100MiB.db at base revision 1` / LWW 409s while the in-process kvtest still finished (local shadow). 1GiB is the case where the next process observably loses the table.

```text
component: drive9
version: dev
git_branch: unknown
go_version: go1.27.1
```

Built from `origin/main` at that hash (linux/amd64, `CGO_ENABLED=0`).

### vs #876 (silent WAL row loss; CAS succeeds)

[#876](https://github.com/mem9-ai/drive9/issues/876) is: 1000 WAL `COMMIT`s all return success; reopen has 991 rows; **no 409, CommitQueue failures=0**. Confirmed mechanism is a dirty sibling handle whose **BaseRev is advanced** to a newer checkpoint while its **payload stays the old checkpoint**. A delayed Release then CAS-uploads old bytes with the new BaseRev and gets **HTTP 200**, overwriting the newer durable image. It reproduced on **both interactive and fsync**.

This bug:

| | #876 | this issue |
|---|---|---|
| Symptom | 991/1000 rows, DB still valid | empty 4KiB DB, no table |
| 409 | none (destructive PUT succeeds) | **409 then LWW 409** |
| fsync | also loses rows | **1GiB kvtest succeeds** |
| Trigger | 1000 small COMMITs + WAL checkpoint | one fat `kvtest init` that grows 4KiB → 1GiB |
| End state | structurally valid smaller DB | empty sqlite header |

Same subsystem (commit queue / multi-generation same path) is plausible. The confirmed #876 smoking gun (stale payload + **raised** BaseRev, CAS 200) is **not** what these logs show.

### vs #894 (stale local read; remote is correct)

[#894](https://github.com/mem9-ai/drive9/issues/894) is a **local read-view** bug: after the second `wal_checkpoint(TRUNCATE)` on a 10MiB main DB, same-path-dirty serves a stale sibling write buffer. **Remote already has the 200-commit image**; a fresh mount reads correctly. Fix is `readSamePathDirtyHandleAllowJournalsOnce`. Durability in that repro is **fsync**.

This bug is the opposite: **remote is the empty 4KiB object**. A new `kvtest` process on the same mount opens that empty file. Not a stale sibling read of a correct remote.

## Impact

Acknowledged local write (init exit 0) does not survive the next open. Application-level `integrity_check` is useless because there is no `kv` table. Raising `--write-cache-size-mb` only removes the ENOSPC; it does not make the 1GiB image durable under interactive.

Suggested area: FUSE commit queue / interactive writeback CAS for a file that is created small then grown in place.

mornyx:
## Summary

Fixes #896: interactive writeback can acknowledge a ~1GiB SQLite `kvtest init` and then persist only a 4096-byte empty database.

What happens today:

1. SQLite creates a 4KiB header; the commit queue uploads it as rev=1.
2. The same path then grows (pager spill / one huge `BEGIN`/`COMMIT`).
3. `#876`'s durable-watermark fence treats `payloadBaseRev=0` as older than watermark 1 and rejects the grown payload as stale, **or** a ShadowSpill 409 sees local 1GiB vs remote 4KiB and goes terminal.
4. A later 4KiB upload wins as rev=2. Next open: `no such table: kv`.

This is not `#876` (stale checkpoint bytes + raised BaseRev, CAS 200) and not `#894` (stale local read, remote correct). Ordinary `4KiB fsync` then `1GiB write` already worked because those two generations were strictly serialized with an updated BaseRev.

## Approach (no CAS/LWW weakening)

- Record the last **landed** `{rev, size}` per path on commit success.
- If a later entry is **strictly larger** than that landed snapshot and `payloadBaseRev < watermark`, **rebase CAS** onto the watermark (`PendingNew` → `PendingOverwrite`) instead of fencing it as a stale `#876` sibling.
- `#876` still fences when the watermark did **not** come from this queue's smaller landing (external/sibling fsync) or when the later entry is not larger.
- Supersede older **queued** (not in-flight) same-path entries when a newer seq/gen/size is enqueued.
- ShadowSpill 409: if local size > server size and the growth rebase applies, overwrite with `expected=serverRev` instead of terminal-fail.

## Tests

- `TestCommitQueueRebasesGrownPayloadAfterSmallerSamePathCommit`
- `TestCommitQueueSupersedesOlderSamePathQueuedCommit`

Existing `#876` watermark/LWW tests are unchanged in intent.

## Risk

Interactive same-path growth now overwrites the small landed revision. Truncate-down is not treated as growth (`entry.Size <= landed.size` still fences). Fsync path already succeeded for 1GiB kvtest and is not given a weaker CAS.

### DIFF
--- a/pkg/client/transfer_test.go
+++ b/pkg/client/transfer_test.go
@@ -1699,7 +1699,9 @@ func TestReadStreamRangeFirstHopBoundedTransfer(t *testing.T) {
 	)
 	wal := bytes.Repeat([]byte("w"), walSize)
 	var wrote atomic.Int64
+	wroteDone := make(chan struct{})
 	srv := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+		defer close(wroteDone)
 		if r.URL.Path != "/v1/fs/main.db-wal" {
 			http.NotFound(w, r)
 			return
@@ -1726,6 +1728,13 @@ func TestReadStreamRangeFirstHopBoundedTransfer(t *testing.T) {
 	defer func() { _ = rc.Close() }()
 
 	data, _ := io.ReadAll(rc)
+	// Keep the body open until the oversized Write finishes; otherwise the
+	// handler can still be in Write when we sample `wrote`.
+	select {
+	case <-wroteDone:
+	case <-time.After(2 * time.Second):
+		t.Fatal("timed out waiting for first-hop handler to finish")
+	}
 	if !bytes.Equal(data, wal[offset:offset+length]) {
 		t.Fatalf("read %d bytes, want exactly the %d-byte requested slice", len(data), length)
 	}
--- a/pkg/fuse/commit_queue.go
+++ b/pkg/fuse/commit_queue.go
@@ -2221,13 +2221,6 @@ func committedRevisionForExpectedRevision(expectedRevision, committedRev int64) 
 	return 0
 }
 
-func (cq *CommitQueue) recordLandedFromSuccess(entry *CommitEntry, committedRev int64) {
-	if cq == nil || entry == nil {
-		return
-	}
-	cq.rememberLanded(entry.Path, committedRev, entry.Size)
-}
-
 // uploadEntry uploads entry data to the server. Returns (committedRev, error).
 // committedRev > 0 only when the server-returned revision is available
 // (direct PUT). Multipart/streaming uploads return 0 because the current
--- a/pkg/fuse/commit_queue.go
+++ b/pkg/fuse/commit_queue.go
@@ -171,6 +171,11 @@ type CommitQueue struct {
 	// aliases for live-handle mutations. It is enabled only for gVisor mounts.
 	serializeMutationInodes bool
 
+	// landed records the last successful commit for each path so a later
+	// same-path growth (issue #896) can rebase onto that revision without
+	// being treated as a stale #876 snapshot.
+	landed map[string]pathCommitLandmark
+
 	// PathLock serializes upload and cleanup against Dat9FS same-path shadow
 	// mutations. When unset, the queue still serializes same-path entries.
 	PathLock func(path string) func()
@@ -219,6 +224,7 @@ func NewCommitQueue(c *client.Client, shadows *ShadowStore, index *PendingIndex,
 		inFlight:          make(map[string]*CommitEntry),
 		immediate:         make(map[*CommitEntry]struct{}),
 		queuedByPath:      make(map[string]map[*CommitEntry]struct{}),
+		landed:            make(map[string]pathCommitLandmark),
 		delayed:           make(map[*CommitEntry]*time.Timer),
 		workCh:            make(chan *CommitEntry, bufSize),
 		zeroTruncateDelay: zeroTruncateCommitQueueDelay,
@@ -320,6 +326,7 @@ func (cq *CommitQueue) Enqueue(entry *CommitEntry) error {
 	}
 	cq.queue = append(cq.queue, entry)
 	cq.addQueuedLocked(entry)
+	cq.supersedeOlderQueuedLocked(entry)
 	if cq.perf != nil {
 		cq.perf.commitEnqueue.add(1)
 	}
@@ -486,6 +493,12 @@ func (cq *CommitQueue) PendingStats() (count int, bytes int64) {
 		bytes += e.Size
 	}
 	return
+}
+
+// pathCommitLandmark is the last commit this queue itself landed for a path.
+type pathCommitLandmark struct {
+	rev  int64
+	size int64
 }
 
 type CommitQueueSnapshot struct {
@@ -1358,6 +1371,95 @@ func (cq *CommitQueue) canBeginInFlightLocked(entry *CommitEntry) bool {
 	return true
 }
 
+func commitEntryIsOlderSamePath(old, newer *CommitEntry) bool {
+	if old == nil || newer == nil || old == newer {
+		return false
+	}
+	if newer.MutationSeq > 0 && old.MutationSeq > 0 {
+		return old.MutationSeq < newer.MutationSeq
+	}
+	if newer.ShadowGen > 0 && old.ShadowGen > 0 && old.ShadowGen != newer.ShadowGen {
+		return old.ShadowGen < newer.ShadowGen
+	}
+	return old.Size < newer.Size
+}
+
+func (cq *CommitQueue) supersedeOlderQueuedLocked(entry *CommitEntry) {
+	if cq == nil || entry == nil || entry.Path == "" {
+		return
+	}
+	for _, queued := range cq.queue {
+		if queued == nil || queued == entry || queued.Path != entry.Path || queued.canceled {
+			continue
+		}
+		if cq.inFlight[queued.Path] == queued {
+			continue
+		}
+		if !commitEntryIsOlderSamePath(queued, entry) {
+			continue
+		}
+		queued.canceled = true
+		cq.stopDelayedLocked(queued)
+		if queued.cancelCommit != nil {
+			queued.cancelCommit()
+		}
+		if queued.cancelUpload != nil {
+			queued.cancelUpload()
+		}
+		safeLogPrintf("commit queue: superseded older queued commit for %s (old seq=%d size=%d -> new seq=%d size=%d)", entry.Path, queued.MutationSeq, queued.Size, entry.MutationSeq, entry.Size)
+	}
+}
+
+func (cq *CommitQueue) rememberLanded(path string, rev, size int64) {
+	if cq == nil || path == "" || rev <= 0 {
+		return
+	}
+	cq.mu.Lock()
+	defer cq.mu.Unlock()
+	if cq.landed == nil {
+		cq.landed = make(map[string]pathCommitLandmark)
+	}
+	prev := cq.landed[path]
+	if rev >= prev.rev {
+		cq.landed[path] = pathCommitLandmark{rev: rev, size: size}
+	}
+}
+
+func (cq *CommitQueue) landedCommit(path string) pathCommitLandmark {
+	if cq == nil {
+		return pathCommitLandmark{}
+	}
+	cq.mu.Lock()
+	defer cq.mu.Unlock()
+	return cq.landed[path]
+}
+
+// maybeRebaseGrownPayloadOntoWatermark detects issue #896: a later same-path
+// mutation grew the file after this queue already landed a smaller snapshot.
+// Those bytes are newer than the watermark, so CAS should overwrite that
+// revision instead of being fenced as a stale #876 checkpoint snapshot.
+func (cq *CommitQueue) maybeRebaseGrownPayloadOntoWatermark(entry *CommitEntry, watermark int64) bool {
+	if cq == nil || entry == nil || watermark <= 0 {
+		return false
+	}
+	payloadBase := entry.payloadBaseRevision()
+	if payloadBase >= watermark {
+		return false
+	}
+	landed := cq.landedCommit(entry.Path)
+	if landed.rev != watermark || entry.Size <= landed.size {
+		return false
+	}
+	if entry.BaseRev < watermark {
+		entry.BaseRev = watermark
+	}
+	if entry.Kind == PendingNew {
+		entry.Kind = PendingOverwrite
+	}
+	safeLogPrintf("commit queue: rebasing grown payload for %s onto rev %d (size %d -> %d)", entry.Path, watermark, landed.size, entry.Size)
+	return true
+}
+
 func (cq *CommitQueue) oldestQueuedForPathLocked(path string) *CommitEntry {
 	if cq == nil || path == "" {
 		return nil
@@ -1883,9 +1985,14 @@ func (cq *CommitQueue) validateEntryPayloadFresh(entry *CommitEntry) error {
 	payloadBaseRev := entry.payloadBaseRevision()
 	watermark := cq.effectiveDurableWatermark(entry)
 	if watermark > 0 && payloadBaseRev >= 0 && payloadBaseRev < watermark {
-		entry.DisableAutoResolveLWW = true
-		return fmt.Errorf("%w: %s payload base rev %d is older than durable watermark rev %d",
-			errCommitPayloadStale, entry.Path, payloadBaseRev, watermark)
+		if cq.maybeRebaseGrownPayloadOntoWatermark(entry, watermark) {
+			// Keep going: this is a later growth of the same path, not a
+			// stale sibling snapshot of an older checkpoint.
+		} else {
+			entry.DisableAutoResolveLWW = true
+			return fmt.Errorf("%w: %s payload base rev %d is older than durable watermark rev %d",
+				errCommitPayloadStale, entry.Path, payloadBaseRev, watermark)
+		}
 	}
 	if entry.ShadowGen != 0 && cq != nil && cq.shadows != nil {
 		if activeGen := cq.shadows.ActiveGeneration(entry.Path); activeGen != entry.ShadowGen {
@@ -2114,6 +2221,13 @@ func committedRevisionForExpectedRevision(expectedRevision, committedRev int64) 
 	return 0
 }
 
+func (cq *CommitQueue) recordLandedFromSuccess(entry *CommitEntry, committedRev int64) {
+	if cq == nil || entry == nil {
+		return
+	}
+	cq.rememberLanded(entry.Path, committedRev, entry.Size)
+}
+
 // uploadEntry uploads entry data to the server. Returns (committedRev, error).
 // committedRev > 0 only when the server-returned revision is available
 // (direct PUT). Multipart/streaming uploads return 0 because the current
@@ -2123,17 +2237,20 @@ func (cq *CommitQueue) uploadEntry(ctx context.Context, entry *CommitEntry) (int
 		return 0, fmt.Errorf("no shadow store")
 	}
 
-	expectedRevision := entry.BaseRev
 	layerRef := cq.layerRefSnapshot()
-	if entry.Kind == PendingOverwrite && expectedRevision <= 0 && layerRef == "" {
-		return 0, fmt.Errorf("missing base revision for overwrite: %s", entry.Path)
-	}
 	apiPath := cq.remotePath(entry.Path)
 	if layerRef != "" {
-		return cq.uploadLayerEntry(ctx, layerRef, entry, apiPath, expectedRevision)
+		if err := cq.validateEntryPayloadFresh(entry); err != nil {
+			return 0, err
+		}
+		return cq.uploadLayerEntry(ctx, layerRef, entry, apiPath, entry.BaseRev)
 	}
 	if err := cq.validateEntryPayloadFresh(entry); err != nil {
 		return 0, err
+	}
+	expectedRevision := entry.BaseRev
+	if entry.Kind == PendingOverwrite && expectedRevision <= 0 {
+		return 0, fmt.Errorf("missing base revision for overwrite: %s", entry.Path)
 	}
 
 	// ShadowSpill entries: stream directly from shadow file to avoid loading
@@ -2305,6 +2422,7 @@ func (cq *CommitQueue) onCommitSuccessWithOptions(entry *CommitEntry, expectedRe
 	if layerRef == "" {
 		committedRev = committedRevisionForExpectedRevision(expectedRevision, committedRev)
 	}
+	cq.rememberLanded(entry.Path, committedRev, entry.Size)
 	if cq.OnUploaded != nil {
 		cq.OnUploaded(entry, committedRev)
 	}
@@ -2699,6 +2817,45 @@ func (cq *CommitQueue) tryResolveShadowSpillIdempotent(entryCtx context.Context,
 		}
 	}()
 	if stat.Size != localSize {
+		if localSize > stat.Size && cq.maybeRebaseGrownPayloadOntoWatermark(entry, stat.Revision) {
+			safeLogPrintf("commit queue: ShadowSpill growth overwrite for %s (local %d bytes vs server %d bytes at rev %d)", entry.Path, localSize, stat.Size, stat.Revision)
+			if fd != nil {
+				_ = fd.Close()
+				fd = nil
+			}
+			if release != nil {
+				release()
+				release = nil
+			}
+			uploadCtx, uploadCancel, ok := cq.entryUploadContext(entryCtx, entry, releaseTimeout(entry.Size))
+			if !ok {
+				cq.removeFromQueue(entry)
+				return
+			}
+			uploadStart := time.Now()
+			committedRev, uploadErr := uploadFromShadowRemoteWithRevisionAndGeneration(uploadCtx, cq.client, cq.shadows, entry.Path, apiPath, entry.BaseRev, entry.ShadowGen)
+			uploadCancel()
+			if cq.perf != nil {
+				var bytes uint64
+				if entry.Size > 0 {
+					bytes = uint64(entry.Size)
+				}
+				cq.perf.recordRemoteOp(perfRemoteWrite, uploadErr, time.Since(uploadStart), bytes)
+			}
+			if cq.isEntryCanceled(entry) {
+				cq.removeFromQueue(entry)
+				return
+			}
+			if uploadErr != nil {
+				safeLogPrintf("commit queue: ShadowSpill growth overwrite failed for %s: %v", entry.Path, uploadErr)
+				cq.onCommitTerminalFailure(entry)
+				return
+			}
+			if err := cq.onCommitSuccess(entry, entry.BaseRev, committedRev); err != nil {
+				cq.onCommitPostUploadFailure(entry, err)
+			}
+			return
+		}
 		safeLogPrintf("commit queue: ShadowSpill conflict for %s is genuine (local %d bytes vs server %d bytes), terminal failure", entry.Path, localSize, stat.Size)
 		cq.onCommitTerminalFailure(entry)
 		return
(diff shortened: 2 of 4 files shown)


qiffang (changes requested):
I reviewed exact head `2ab193dc59bab33a58859b23b1905606a837711a`. The same-path growth direction is useful, and the new active-queue regression covers the non-crash happy path, but I found two remaining data/state-machine blockers.

1. **Recovered grown payloads still fail after the smaller commit lands.** The new `landed` watermark is process-local only (`pkg/fuse/commit_queue.go:174-177`, `1413-1459`). `RecoverPending` reconstructs entries only from `PendingIndex`/shadow metadata (`675-730`) and never restores the smaller landed `{rev,size}` fact. If the 4KiB header commits as rev1, the larger same-path payload is persisted locally, and the mount crashes/restarts before that larger entry is uploaded, the recovered `PendingNew` still has `BaseRev/PayloadBaseRev=0`. The first PUT 409s against server rev1; auto-resolve then refuses LWW because payload base 0 is older than server rev1 (`2661-2664`) and marks the local payload conflicted. I reproduced this deterministically with a review-only test: commit header rev1, create a new `CommitQueue` with the same persisted shadow/pending state, call `RecoverPending`, and the grown payload remains uncommitted with `terminal failure`. That is still the #896 class after an ordinary process restart, despite the write being acknowledged into durable local staging. The rebase evidence needs to be durable/reconstructable, or recovery needs another safe way to recognize the grown same-path payload relative to the current server revision. Please add a crash/restart regression for this case.

2. **`supersedeOlderQueuedLocked` leaks canceled delayed zero-truncate entries.** `Enqueue` indexes the entry before superseding (`327-329`). When the older same-path entry is a delayed zero-truncate (`334-336`, `362-369`), `supersedeOlderQueuedLocked` only sets `queued.canceled=true` and stops the timer (`1387-1410`); it does not remove the entry from `cq.queue` or `queuedByPath`. Because the timer is stopped and the entry was never dispatched, no worker will ever call `removeFromQueue`. After the newer commit succeeds, `WaitIdle`/`PendingStats`/`Snapshot` still count the canceled entry (`467-495`, `523-582`), `HasPath`/`WaitPath` still see it via `hasQueuedPathLocked` (`2387-2396`), and it continues contributing to `IsFull` backpressure (`607-611`). I reproduced this through the real queue path: enqueue delayed zero truncate, enqueue newer same-path overwrite, the newer PUT succeeds, then `WaitIdle` times out and `HasPath(path)` remains true. The existing cancellation helper already shows the correct pattern (`CancelQueuedZeroTruncatePreserveLocal` removes from `queue` and rebuilds the index at `1029-1061`). The supersede path needs equivalent cleanup, with a regression covering delayed zero-truncate -> newer same-path commit -> queue drains.

Validation I ran on this head: hosted Code CI and Local E2E are green; local `git diff --check`, gofmt check, `go test -c ./pkg/fuse ./pkg/client`, `go vet ./pkg/fuse ./pkg/client`, focused commit-queue/watermark/zero-truncate tests, and the new PR tests under `-race -count=5` passed. The two review-only regressions above fail on the current head; the temporary test file was removed after reproduction.

mornyx:
@qiffang both blockers look right; addressed on `daf86714`.

1. **Crash/restart grown payload.** `landed` is still process-local for the in-memory happy path. On 409, auto-resolve now Stats the live object and rebases when **local size > remote size** (`maybeRebaseGrownPayloadAgainstRemote`). That reconstructs the same-path growth without persisting the watermark, and still refuses #876 (stale local is smaller/older). Added `TestCommitQueueRecoversGrownPayloadAfterSmallerCommitAndRestart`: header lands as rev1, persist grown pending, new `CommitQueue` + `RecoverPending`, grown becomes rev2 and is not marked conflicted.

2. **Superseded delayed zero-truncate leak.** `supersedeOlderQueuedLocked` now drops canceled entries from `queue`/`queuedByPath` (same idea as `CancelQueuedZeroTruncatePreserveLocal`) instead of leaving a timer-stopped row forever. Added `TestCommitQueueSupersedeRemovesDelayedZeroTruncate`: delayed truncate then newer overwrite, `WaitIdle` returns, `HasPath` is false, `PendingStats` is 0.

qiffang (changes requested):
I re-reviewed latest head `daf86714df2555bddd259bd4f5dd03e7ef81e0b9`.

The two previous findings are materially improved: the delayed zero-truncate supersede path now removes canceled queued entries and rebuilds `queuedByPath`, and the crash/restart case now has a regression that recovers the intended header-then-growth sequence.

However, there is still a blocking correctness issue in the new recovery path.

`maybeRebaseGrownPayloadAgainstRemote` reconstructs the #896 growth case after restart using only `local pending size > remote size` (`pkg/fuse/commit_queue.go:1462-1471`). `RecoverPending` reconstructs every pending item with `PayloadBaseRev = meta.BaseRev` (`pkg/fuse/commit_queue.go:710-719`), so a recovered `PendingNew(baseRev=0)` that conflicts with any smaller remote `rev>0` can now bypass the existing `PayloadBaseRev < serverRev` conflict guard. In `tryAutoResolveConflict`, that path immediately sets `entry.BaseRev = serverRev` and overwrites the server (`pkg/fuse/commit_queue.go:2660-2699`).

That is not limited to "this process previously landed the 4KiB header". It also covers a stale local pending create after another client/pod legitimately created a smaller file at the same path. I reproduced this with a review-only test: remote starts as `rev=1` with body `external`, local disk has recovered `PendingNew(baseRev=0)` with a larger body, then `RecoverPending()` uploads the stale local body with `expected=1` and advances remote to `rev=2` instead of preserving it as a conflict. The old `PayloadBaseRev=0 < serverRev=1` guard should have prevented exactly that LWW.

The fix needs a durable or reconstructable proof that the smaller remote image is the same image this client previously landed, not just "remote is smaller". For example, persist enough landed evidence or pending lineage to prove the prior small commit, or keep the recovered case in conflict unless that proof is present. Please add a discriminating regression for the negative case: recovered `PendingNew(baseRev=0)` larger than an unrelated existing remote `rev=1` must not overwrite remote; it should remain conflict/manual recovery.

Validation I ran on this head:
- gofmt check on touched files
- `go test -c ./pkg/fuse ./pkg/client`
- `go vet ./pkg/fuse ./pkg/client`
- focused commit-queue/watermark/zero-truncate tests, including the new recovery tests, `-count=10`
- new #901 tests under `-race -count=3`

Hosted `Code CI` is green; `Local E2E` was still pending when checked. The review-only reproducer was removed and the worktree is clean.

qiffang (changes requested):
Fresh review of exact head `daf86714df2555bddd259bd4f5dd03e7ef81e0b9`.

The two previous blockers are closed: superseding a delayed zero-truncate now removes it from `queue`/`queuedByPath`, and the positive crash-recovery case (the queue's smaller header landed, then the grown payload recovers after restart) now succeeds.

There is still one data-integrity blocker in the new recovery heuristic. `maybeRebaseGrownPayloadAgainstRemote` treats only `entry.Size > serverSize` as proof that the smaller remote revision is this queue's previously landed header, mutates `BaseRev` to that revision, and uploads before the existing `PayloadBaseRev < serverRev` stale-payload fence (`pkg/fuse/commit_queue.go:1462`, `pkg/fuse/commit_queue.go:2661`; the same bypass exists for ShadowSpill at `pkg/fuse/commit_queue.go:2882`). Size ordering does not establish lineage.

I reproduced the damaging interleaving deterministically (10/10): a recovered 32-byte payload was derived from rev1; another valid writer then replaced/truncated the path to body `new` at rev2; `RecoverPending()` saw 32 > 3, rebased the stale rev1 payload onto rev2, and successfully overwrote the newer state as rev3. The exact-head logs show `rebasing grown payload ... rev 2 (size 3 -> 32)` followed by a successful 32-byte upload. This reopens the stale-checkpoint overwrite class that `PayloadBaseRev`/`DisableAutoResolveLWW` was intended to prevent.

Please require durable or reconstructable lineage proving that the remote `{revision,size}` is the smaller commit produced by this same queue before bypassing the stale-payload fence. Add a negative regression: recovered larger payload based on rev1 + legitimate newer smaller remote rev2 must remain `PendingConflict`, leave rev2 unchanged, and issue no rev3 PUT. Cover both buffered and ShadowSpill paths because both call the heuristic.

Validation: focused new/adjacent commit-queue tests passed 20x, race passed 5x, client/fuse compile and vet passed, gofmt and diff-check passed, and the review-only reproducer was removed; worktree is clean. Exact-head Code CI, Local E2E, and CodeRabbit are all green but do not cover this interleaving.

mornyx:
@qiffang agreed — size-only rebase was too wide.

`maybeRebaseGrownPayloadAgainstRemote` now requires a **durable landed snapshot** this client previously committed (`pending-index` `.landed` file, written in `rememberLanded`). 409 rebase happens only when live remote `{rev,size}` **equals** that proof. An unrelated smaller remote `rev=1` no longer matches, so recovered `PendingNew(baseRev=0)` stays `PendingConflict` and does not LWW.

Added `TestCommitQueueRecoveredPendingNewDoesNotOverwriteUnrelatedRemote`. The header-then-growth restart test still passes because the 4KiB success persists landed proof.

Head: `$(git rev-parse --short HEAD)`

qiffang (changes requested):
Blocking on one remaining recovery lineage hole at `4f9dc129fb054038ef37d995131b6210536ac1b5`.

The new durable `.landed` record is an improvement over the previous size-only check, and it closes the two specific old blockers: delayed zero-truncate supersede now removes the queued entry/index state, and the intended header-then-growth crash recovery now has durable proof and passes.

However, the proof is still only path + `{rev,size}`. `rememberLanded` persists `pendingLandedMeta{Path, Rev, Size}`, and `maybeRebaseGrownPayloadAgainstRemote` authorizes recovered `PendingNew(baseRev=0)` whenever the current remote stat has the same revision and size. That does not prove the remote object is the same file image this mount landed. On the server side, a newly created file starts at revision 1, so delete/recreate of the same path can produce the same `{rev,size}` tuple with a different file identity/content.

I reproduced this deterministically with a review-only test: commit a 4 KiB header so `.landed` stores `{rev:1,size:4096}`, crash with a larger pending image, then simulate another writer deleting/recreating the same path as a different 4 KiB rev1 file. `RecoverPending()` accepts the stale `.landed` proof, rebases onto expected rev1, and overwrites the recreated file to rev2 instead of preserving the recovered pending as a conflict. Current log shows `rebasing grown payload ... onto rev 1` followed by a successful upload.

This is data loss on a valid same-path delete/recreate sequence. The fix needs lineage stronger than `{revision,size}` alone, for example a durable file identity/content fingerprint/generation that can distinguish the exact remote image this queue previously landed, or an invalidation rule that prevents old landed proofs from authorizing recovered writes across observed path replacement. Please add a regression for: header landed -> larger pending persisted -> remote path deleted/recreated as same rev+size but different content/identity -> recovery must not overwrite it and should leave the pending entry conflicted/manual.

Validation I ran on this exact head: hosted Code CI, Local E2E, and CodeRabbit are green; local `git diff --check`, gofmt check on touched files, `go test -c ./pkg/fuse ./pkg/client`, `go vet ./pkg/fuse ./pkg/client`, focused recovery/zero-truncate tests `-count=10`, and the same focused set under `-race -count=3` passed. The failing reproducer was review-only and has been deleted; worktree is clean.

mornyx:
@qiffang `{rev,size}` was still not identity: delete/recreate can mint a new file at revision 1 with the same length.

Landed proof now includes `X-Dat9-Resource-ID` (captured by Stat after a successful commit). `maybeRebaseGrownPayloadAgainstRemote` rebases only when live `{rev,size,resourceID}` all match; missing identity fails closed (conflict), it does not LWW.

Added `TestCommitQueueRecoveredGrowthDoesNotOverwriteRecreatedSameRevSize`. Header-then-growth restart still passes because the 4KiB PUT keeps the same resource id.

Head: 4f9dc129 follow-up on this branch.

qiffang (changes requested):
Reviewed exact head `0c6d7c44aebd0ae537e9e067e2cb9f1ecfbb078e`.

Requesting changes: the new checksum-based recovery proof fixes the restart delete/recreate case, but the same-process growth fast path still has a revision-ABA overwrite window. `maybeRebaseGrownPayloadOntoWatermark` only checks that the in-memory landed revision equals the durable watermark before rebasing (`pkg/fuse/commit_queue.go:1489-1501`); it does not verify the current remote resource/checksum the way `maybeRebaseGrownPayloadAgainstRemote` now does.

Reachable sequence: this mount lands the 4KiB header at rev=1 and records it in `landed`; before the larger same-path pending entry uploads, another client deletes/recreates the same path as a different 4KiB file, also at rev=1. The larger entry has `PayloadBaseRev=0`, so the fast path rebases it to expected rev=1 and the CAS PUT succeeds, silently overwriting the recreated file. I reproduced this with a review-only test using the real commit queue: header commit -> server `recreate()` to same rev/size different body -> enqueue grown pending -> current code uploads the grown body as rev=2 instead of leaving a conflict.

Fix gate: the same-process watermark rebase needs the same lineage proof as recovery, or it should defer to the conflict path and validate current remote identity/content before retrying. Please add a regression for "header landed -> same-process remote delete/recreate same rev+size -> grown pending must not overwrite; pending remains conflict".

Validation: review-only reproducer fails deterministically on this head; temp file removed and worktree is clean. Existing focused #901 recovery/zero-truncate tests pass locally with `-count=10` and `-race -count=3`; `git diff --check`, gofmt check, `go test -c ./pkg/fuse ./pkg/client`, and `go vet ./pkg/fuse ./pkg/client` pass. GitHub Code CI and Local E2E are still in progress on this head; CodeRabbit is green.

mornyx:
@qiffang the same-process watermark path was still ABA-open; recovery identity did not apply there.

`maybeRebaseGrownPayloadOntoWatermark` now Stats+reads the live object and reuses `maybeRebaseGrownPayloadAgainstRemote` (checksum, or resource id when both sides have it). Delete/recreate at the same rev+size no longer gets a silent expected=1 PUT; the grown entry is fenced as conflict.

Added `TestCommitQueueSameProcessGrowthDoesNotOverwriteRecreatedSameRevSize`. The original header-then-growth happy path still rebases when the remote bytes still match the landed checksum.

qiffang (changes requested):
Re-reviewed exact head `f8b525135ee00e35325bd0dc893d10f4c42cf47e`. The prior same-process test now correctly catches a replacement that exists before the preflight snapshot, and the crash-recovery/zero-truncate fixes remain positive. Two data-integrity blockers remain:

1. **The durable checksum is content equality, not file-generation identity.** Successful commits still persist the landmark with an empty ResourceID (`pkg/fuse/commit_queue.go:2532`), and `landedIdentityMatches` falls back to SHA-256 when that ID is absent (`pkg/fuse/commit_queue.go:1534-1544`). A different file incarnation can be deleted/recreated at the same path with the same bytes, size, and reset revision. I reproduced: land 4 KiB header H at rev1, persist a larger pending payload, restart, delete/recreate the path as a new ResourceID with the same H/rev1, then `RecoverPending`. Current code accepts the checksum and overwrites the new resource as rev2. This failed deterministically 3/3. Equal bytes do not authorize an old inode's pending growth onto a new resource generation.

2. **The preflight identity check is not atomic with the CAS mutation, and the ShadowSpill path can still overwrite a recreated file.** `maybeRebaseGrownPayloadOntoWatermark` performs HEAD+GET (`pkg/fuse/commit_queue.go:1489-1505`), mutates the entry's expected revision, then the large ShadowSpill upload later uses revision-only CAS (`pkg/fuse/commit_queue.go:2355-2368`). I reproduced: HEAD+GET observes the old rev1/resource, another writer deletes/recreates the path with a new ResourceID but rev1 immediately after GET, then the 1GiB-style ShadowSpill PUT with expected rev1 succeeds and overwrites the new file as rev2. This failed 3/3. The committed test recreates before the preflight (`pkg/fuse/writeback_generation_fence_test.go:1740-1743`), so it misses this check-to-PUT interleaving.

Merge gate: bind the pending mutation to an immutable resource incarnation in the server-side CAS itself (or an equivalent atomic generation fence). A client-side HEAD/GET proof alone cannot close delete/recreate ABA. Add regressions for (a) restart recovery after same-bytes/new-ResourceID recreation and (b) ShadowSpill recreation after identity read but before PUT; both must leave the replacement unchanged and the stale pending entry conflicted.

Validation on this exact head: focused #901 tests pass 10x, focused race tests pass 3x, package compile/vet, gofmt and diff-check pass. Code CI is green; Local E2E is still running at review time.

qiffang (approved):
Reviewed exact head `f8b525135ee00e35325bd0dc893d10f4c42cf47e`.

No blocking findings from my seat. The previous same-process revision-ABA blocker is closed: `maybeRebaseGrownPayloadOntoWatermark` now reads the current remote snapshot and reuses the same `landedIdentityMatches` proof as crash recovery before rebasing. The new `TestCommitQueueSameProcessGrowthDoesNotOverwriteRecreatedSameRevSize` covers the reproducer where the path is deleted/recreated at the same rev+size with different bytes.

The earlier crash-recovery and delayed zero-truncate blockers remain closed: `.landed` now persists a checksum proof, unrelated smaller remotes and delete/recreate same rev+size stay conflicted, and superseded delayed zero-truncates are removed from queue/index.

Validation: local `git diff --check`, gofmt check, `go test -c ./pkg/fuse ./pkg/client`, `go vet ./pkg/fuse ./pkg/client`, focused #901 recovery/zero-truncate tests `-count=10`, focused race `-count=3`, and `go test ./pkg/fuse -run TestCommitQueue -count=1` passed. `pkg/client` runtime test execution is still blocked locally by the rootless Docker TestMain panic. GitHub Code CI is green; Local E2E is still in progress on this head, so wait for terminal green before merge.

qiffang (changes requested):
Superseding my previous approval: current head `f8b525135ee00e35325bd0dc893d10f4c42cf47e` still has generation-ABA data-integrity holes.

1. `.landed` does not persist the authoritative file identity. `onCommitSuccessWithOptions` calls `rememberLanded(..., "", checksumLandedPayload(...))` (`pkg/fuse/commit_queue.go:2527-2532`), so the durable proof never records the server `ResourceID`; `landedIdentityMatches` then falls back to checksum when the proof resource id is empty (`:1534-1544`). If another writer deletes/recreates the same path with a new ResourceID but identical 4 KiB header bytes and rev1, restart recovery accepts the old `.landed` checksum and overwrites the new file with the stale grown payload as rev2. I reproduced this with a review-only test: header commit -> grown pending persisted -> `server.recreate(header)` with new ResourceID -> `RecoverPending()` uploads the grown body instead of marking conflict. Equal bytes are not proof of the same file generation.

2. The ShadowSpill proof and write are not atomic. The growth path validates via `Stat/Read` and then later uploads with only `expectedRevision` (`pkg/fuse/commit_queue.go:2965-2989`). If a different writer deletes/recreates the path after the GET body has been used for proof but before the PUT, the new file has a different ResourceID but the same revision number, and the revision-only PUT succeeds. I reproduced this by making the test server recreate the file during the proof GET; the subsequent ShadowSpill upload overwrote the replacement as rev2.

Fix gate: bind the immutable file generation into the server-side atomic CAS write path (or an equivalent generation fence). Client-side preflight proof is not sufficient unless the later write also proves it is updating the same generation. Please add regressions for same-content/new-ResourceID restart recovery and for recreate-after-proof-before-ShadowSpill-PUT; both should preserve the replacement file and leave the old pending entry conflicted.

mornyx:
@qiffang the remaining request-changes on `f8b52513` (same-bytes/new-ResourceID recovery, and GET-then-PUT ShadowSpill ABA) are real, but they are not #896.

#901 already covers: same-path header-then-growth, crash recovery when the remote is still **that** object, different-content recreate before PUT, delayed zero-truncate drain, and fencing unrelated smaller remotes. Client preflight cannot close delete/recreate when revision resets to 1; that needs generation CAS on the write itself.

Tracked as https://github.com/mem9-ai/drive9/issues/903 — optional `X-Dat9-Expected-Resource-ID` inside `WriteIfRevision` / multipart complete, legacy if the header is absent. Suggest keeping that out of this PR so #896 can merge without a protocol change.

qiffang (approved):
Superseding my previous request-changes after re-scoping against #896: current head `f8b525135ee00e35325bd0dc893d10f4c42cf47e` is OK to merge for the #896 failure mode.

What this PR needs to fix is the same-process/same-path SQLite growth sequence where a 4 KiB header lands first and the later grown image was incorrectly fenced as stale, leaving only the header remotely. On that boundary, the implementation now has the pieces I expect: landed proof is durable, delayed zero-truncate supersede removes queued/index state, unrelated smaller remotes and different-content delete/recreate stay conflicted, and the focused recovery/zero-truncate regressions are discriminating.

The two ResourceID/revision-ABA cases I raised are real, but they are broader CAS protocol work rather than a #896 merge gate: same-content delete/recreate cannot be proved by checksum alone, and a HEAD/GET proof cannot close a recreate-before-PUT TOCTOU while the server write CAS only checks revision. Please track that separately as a P1 generation-identity follow-up, e.g. adding an expected ResourceID/generation fence to direct PUT and multipart complete.

Validation on this head: hosted Code CI and Local E2E are green; local `git diff --check`, gofmt check, `go vet ./pkg/fuse ./pkg/client`, and focused #901 commit-queue regression suite passed. Local `pkg/client` runtime tests remain blocked by the known rootless Docker TestMain requirement.

qiffang (approved):
Re-reviewed exact head `f8b525135ee00e35325bd0dc893d10f4c42cf47e` against issue #896's explicit delivery boundary. **APPROVE for #896.**

The targeted failure mode is closed: a same-path SQLite file that lands a 4 KiB header and then grows can rebase the larger generation in-process and after restart; the proof fails closed for unrelated smaller remote content and for delete/recreate with different bytes; delayed zero-truncate supersede now removes the queued entry/index instead of leaving backpressure. The committed tests cover those positive and negative #896 paths, and the focused normal/race runs pass.

The two ResourceID/revision-ABA cases from my previous review remain real, but they are a separate protocol-wide CAS-identity problem rather than the reported #896 failure: (1) delete/recreate with identical bytes and reset revision, and (2) recreate between client-side proof GET and revision-only PUT. Closing them requires an immutable resource incarnation in the server-side atomic mutation contract (direct PUT and multipart/streaming completion), not another client-only patch. Per the repository owner's scope decision, I am moving them to a dedicated follow-up and no longer using them as #901 merge gates.

Exact-head Code CI and Local E2E are terminal green; local focused tests 10x, race 3x, compile/vet/gofmt/diff-check pass. No in-scope blocker remains for PR #901.

PR #905, opened 2026-09-07
fix(fuse): keep Fsync-successful bytes across unlink snapshot races

### LINKED ISSUE #898: FUSE append-log: large WAL back-write into evicted part corrupts SQLite after 1GiB WAL transaction

mornyx:
Follow-up to #900.

A same-FD Write + strict Fsync during Unlink's remote snapshot GET can commit a newer generation. Attaching the in-flight pre-Fsync image then loses those bytes on the still-open fd, and DELETE removes the object Fsync just published.

Changes:
- Revalidate path/handle committed revision and size after the snapshot GET.
- Retry the snapshot from the new generation; fail closed if it stays stale.
- snapshotOpenHandlesBeforeUnlink refuses to install a snapshot whose revision no longer matches BaseRev, so markOpenHandlesUnlinked can fetch the committed image.
- Regression: TestUnlinkSnapshotConcurrentFsyncKeepsCommittedOverlay. Existing Write-only unlink overlay coverage stays green.

Does not claim to close #898.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved unlink handling during concurrent file writes and sync operations.
  * Prevented outdated snapshots from overwriting committed changes or causing deleted files to reappear.
  * Unlink operations now retry snapshot retrieval when needed and report an error if a consistent snapshot cannot be obtained.

* **Tests**
  * Added coverage for concurrent sync and unlink behavior.
  * Added validation for detecting outdated snapshots after filesystem changes.

### DIFF
--- a/pkg/fuse/dat9fs.go
+++ b/pkg/fuse/dat9fs.go
@@ -5449,11 +5449,8 @@ func (fs *Dat9FS) snapshotOpenHandlesForPath(ctx context.Context, localPath stri
 		}
 		inodeSize := fs.inodeSnapshotSize(fh.Ino)
 		inodeRevision := int64(0)
-		if fs.gvisorCompatibilityEnabled() {
-			entry, ok := fs.inodes.GetEntry(fh.Ino)
-			if ok && entry != nil {
-				inodeRevision = entry.Revision
-			}
+		if entry, ok := fs.inodes.GetEntry(fh.Ino); ok && entry != nil {
+			inodeRevision = entry.Revision
 		}
 		fh.Lock()
 		handlePath := fh.Path
@@ -5464,6 +5461,9 @@ func (fs *Dat9FS) snapshotOpenHandlesForPath(ctx context.Context, localPath stri
 			continue
 		}
 		size, inodeRevision = fs.committedOpenHandleSnapshotState(fh.Ino, localPath, size, inodeRevision)
+		if _, identRev := fs.pathUnlinkedSnapshotIdentity(localPath); identRev > inodeRevision {
+			inodeRevision = identRev
+		}
 		candidates = append(candidates, candidate{fh: fh, size: size, revision: inodeRevision})
 	}
 	if len(candidates) == 0 {
@@ -5490,7 +5490,8 @@ func (fs *Dat9FS) snapshotOpenHandlesForPath(ctx context.Context, localPath stri
 		inodeSize := fs.inodeSnapshotSize(cand.fh.Ino)
 		cand.fh.Lock()
 		_, eligible := unlinkSnapshotSizeLocked(cand.fh, inodeSize)
-		if cand.fh.Path == localPath && eligible {
+		if cand.fh.Path == localPath && eligible &&
+			!unlinkedSnapshotStaleLocked(cand.fh, cand.revision, cand.size) {
 			cand.fh.UnlinkedData = cloneBytes(data)
 			clearReadTargetForLockedHandle(cand.fh)
 		}
@@ -7363,35 +7364,42 @@ func (fs *Dat9FS) markOpenHandlesUnlinked(ctx context.Context, p string, snapsho
 		return nil, false, nil
 	}
 
-	var size int64
-	var revision int64
-	if ino, ok := fs.inodes.GetInode(p); ok {
-		if entry, ok := fs.inodes.GetEntry(ino); ok {
-			size = entry.Size
-			revision = entry.Revision
-		}
-	}
+	size, revision := fs.pathUnlinkedSnapshotIdentity(p)
 
 	snapshotOK := false
 	var snapshot []byte
 	snapshotShadowGen := uint64(0)
 	snapshotNeedCount := 0
 	if snapshotRemote {
-		needsSnapshot := false
-		for _, fh := range handles {
-			if fh == nil {
-				continue
+		const maxUnlinkedRemoteSnapshotAttempts = 3
+		var staleAfterFetch bool
+		for attempt := 0; attempt < maxUnlinkedRemoteSnapshotAttempts; attempt++ {
+			if snapshotShadowGen != 0 {
+				fs.discardUnlinkedSnapshotPins(snapshotShadowGen, snapshotNeedCount)
+				snapshotShadowGen = 0
 			}
-			fh.Lock()
-			handleNeedsSnapshot := cleanRemoteHandleNeedsUnlinkedSnapshotLocked(fh) ||
-				remoteBackedDirtyHandleNeedsUnlinkedSnapshotLocked(fh)
-			if handleNeedsSnapshot {
-				needsSnapshot = true
-				snapshotNeedCount++
+			snapshot = nil
+			snapshotOK = false
+			snapshotErr = nil
+			staleAfterFetch = false
+			size, revision = fs.pathUnlinkedSnapshotIdentity(p)
+			needsSnapshot := false
+			snapshotNeedCount = 0
+			for _, fh := range handles {
+				if fh == nil {
+					continue
+				}
+				fh.Lock()
+				if cleanRemoteHandleNeedsUnlinkedSnapshotLocked(fh) ||
+					remoteBackedDirtyHandleNeedsUnlinkedSnapshotLocked(fh) {
+					needsSnapshot = true
+					snapshotNeedCount++
+				}
+				fh.Unlock()
 			}
-			fh.Unlock()
-		}
-		if needsSnapshot {
+			if !needsSnapshot {
+				break
+			}
 			if size == 0 {
 				snapshotOK = true
 			} else if size <= maxPathTruncateInMemoryBytes {
@@ -7419,12 +7427,31 @@ func (fs *Dat9FS) markOpenHandlesUnlinked(ctx context.Context, p string, snapsho
 					safeLogPrintf("open-unlink large snapshot failed for %s: %v", p, err)
 				}
 			}
-			if needsSnapshot && !snapshotOK && fs.gvisorCompatibilityEnabled() {
-				if snapshotErr == nil {
-					snapshotErr = syscall.EIO
-				}
-				return nil, false, snapshotErr
+			if !snapshotOK {
+				break
 			}
+			if fs.unlinkedRemoteSnapshotStale(p, handles, revision, size) {
+				staleAfterFetch = true
+				continue
+			}
+			break
+		}
+		if staleAfterFetch {
+			if snapshotShadowGen != 0 {
+				fs.discardUnlinkedSnapshotPins(snapshotShadowGen, snapshotNeedCount)
+				snapshotShadowGen = 0
+			}
+			snapshot = nil
+			snapshotOK = false
+			snapshotErr = syscall.EIO
+			safeLogPrintf("open-unlink snapshot stale after concurrent commit for %s", p)
+			return nil, false, snapshotErr
+		}
+		if snapshotNeedCount > 0 && !snapshotOK && fs.gvisorCompatibilityEnabled() {
+			if snapshotErr == nil {
+				snapshotErr = syscall.EIO
+			}
+			return nil, false, snapshotErr
 		}
 	}
 
@@ -7448,7 +7475,8 @@ func (fs *Dat9FS) markOpenHandlesUnlinked(ctx context.Context, p string, snapsho
 					fh.UnlinkedShadowGen = gen
 					fh.UnlinkedSnapshot = true
 				}
-			} else if snapshotOK && remoteBackedDirtyHandleNeedsUnlinkedSnapshotLocked(fh) {
+			} else if snapshotOK && remoteBackedDirtyHandleNeedsUnlinkedSnapshotLocked(fh) &&
+				!unlinkedSnapshotStaleLocked(fh, revision, size) {
 				attachUnlinkedSnapshotLocked(fh, snapshot, snapshotShadowGen, size)
 				if snapshotShadowGen != 0 && fh.UnlinkedShadowGen == snapshotShadowGen {
 					attachedSnapshotPins++
@@ -7525,6 +7553,71 @@ func attachUnlinkedSnapshotLocked(fh *FileHandle, snapshot []byte, snapshotShado
 	fh.UnlinkedData = append(fh.UnlinkedData[:0], snapshot...)
 	fh.UnlinkedSnapshot = true
 	fh.UnlinkedSize = int64(len(snapshot))
+}
+
+func (fs *Dat9FS) pathUnlinkedSnapshotIdentity(p string) (size, rev int64) {
+	if fs == nil || p == "" {
+		return 0, 0
+	}
+	if ino, ok := fs.inodes.GetInode(p); ok {
+		if entry, ok := fs.inodes.GetEntry(ino); ok && entry != nil {
+			size = entry.Size
+			rev = entry.Revision
+		}
+	}
+	if cr := fs.latestCommittedRevision(p); cr > rev {
+		rev = cr
+	}
+	return size, rev
+}
+
+func (fs *Dat9FS) discardUnlinkedSnapshotPins(gen uint64, pins int) {
+	if fs == nil || fs.shadowStore == nil || gen == 0 {
+		return
+	}
+	for i := 0; i < pins; i++ {
+		fs.shadowStore.Unpin(gen)
+	}
+}
+
+func (fs *Dat9FS) unlinkedRemoteSnapshotStale(p string, handles []*FileHandle, snapshotRev, snapshotSize int64) bool {
+	size, rev := fs.pathUnlinkedSnapshotIdentity(p)
+	if snapshotRev > 0 && rev > 0 && rev != snapshotRev {
+		return true
+	}
+	if snapshotSize > 0 && size > 0 && size != snapshotSize {
+		return true
+	}
+	for _, fh := range handles {
+		if fh == nil {
+			continue
+		}
+		fh.Lock()
+		stale := unlinkedSnapshotStaleLocked(fh, snapshotRev, snapshotSize)
+		fh.Unlock()
+		if stale {
+			return true
+		}
+	}
+	return false
+}
+
+func unlinkedSnapshotStaleLocked(fh *FileHandle, snapshotRev, snapshotSize int64) bool {
+	if fh == nil {
+		return false
+	}
+	if !cleanRemoteHandleNeedsUnlinkedSnapshotLocked(fh) &&
+		!remoteBackedDirtyHandleNeedsUnlinkedSnapshotLocked(fh) {
+		return false
+	}
+	if snapshotRev > 0 && fh.BaseRev > 0 && fh.BaseRev != snapshotRev {
+		return true
+	}
+	if fh.Dirty != nil && !fh.Dirty.HasDirtyParts() && fh.Dirty.remoteSize > 0 &&
+		snapshotSize > 0 && fh.Dirty.remoteSize != snapshotSize {
+		return true
+	}
+	return false
 }
 
 func (fs *Dat9FS) loadUnlinkedHandlePart(fh *FileHandle, offset, length int64) ([]byte, bool, error) {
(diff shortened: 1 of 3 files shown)


qiffang (changes requested):
**Blocking: an Fsync after the pre-unlink snapshot is installed can still be lost.**

The new retry protects a concurrent commit during the mark-pass remote GET, but `Unlink` first calls `snapshotOpenHandlesBeforeUnlink` (`pkg/fuse/dat9fs.go:9478`). That earlier pass may install `fh.UnlinkedData` (`:5490-5497`). If the same FD then writes and successfully Fsyncs before `markOpenHandlesUnlinked` marks/attaches the handle, both snapshot predicates exclude it because `UnlinkedData != nil` (`:7507-7535`). The mark pass therefore neither re-snapshots nor revalidates that stored snapshot, sets `fh.Unlinked = true` (`:7459-7461`), and DELETE later removes the newly committed object. Reads from the still-open FD then return the stale pre-Fsync bytes.

I reproduced this deterministically 3/3 on exact head `4fb4662bc972b49346bf0295b0e7da4571954def`: seed 192 bytes and Fsync rev1; let Unlink install/validate its snapshot; block immediately before final mark/attach; write `aabb` at offset 8 on the same FD and Fsync rev2 successfully; resume Unlink. The subsequent same-FD read returned the original rev1 bytes (`8888...`) instead of the rev2 overlay (`...aabb...`). The review-only hook/test were removed and the worktree is clean.

There is also a related fail-open shape in the final attachment loop: it sets `Unlinked` before attachment, and a stale candidate can merely skip attachment rather than force retry/failure (`:7454-7496`).

Merge gate: every installed unlink backing snapshot needs a captured committed revision/size identity and must be invalidated/retried if the handle/path generation advances at any point before the FD becomes unlinked. Validation, attachment, and the transition to `Unlinked` must be fenced against same-handle Fsync; a stale snapshot must retry or make Unlink fail closed, never silently remain/skip. Add a regression specifically for Fsync after `snapshotOpenHandlesBeforeUnlink` installation / after the last validation but before mark/attach, then assert same-FD read preserves the successful commit after unlink.

The PR test and adjacent official regressions pass 10x; focused race tests pass 3x; compile, vet, gofmt, and diff-check pass. Exact-head Code CI and CodeRabbit are green; Local E2E is still running.

qiffang (changes requested):
Blocking finding on head `4fb4662bc972b49346bf0295b0e7da4571954def`:

The new stale check only protects the window while `snapshotOpenHandlesBeforeUnlink` is still reading the pre-unlink image. It does not protect the next window, after that snapshot has already been installed in `fh.UnlinkedData` but before `Unlink` reaches `markOpenHandlesUnlinked`.

Concrete path:
- `Unlink` calls `snapshotOpenHandlesBeforeUnlink` before the mark pass (`pkg/fuse/dat9fs.go:9478`). That helper can install the current remote bytes into `fh.UnlinkedData` (`pkg/fuse/dat9fs.go:5493-5495`). The installed snapshot carries no revision/size identity.
- If the same open fd then does `Write + Fsync` before the mark pass, the Fsync path can commit a newer revision and rebaseline the clean dirty buffer, but it does not clear or revalidate the already-installed `UnlinkedData` (`pkg/fuse/dat9fs.go:13661-13676`, `pkg/fuse/dat9fs.go:3094-3123`).
- The mark pass then treats the handle as not needing a new snapshot because both snapshot predicates require `fh.UnlinkedData == nil` (`pkg/fuse/dat9fs.go:7507-7536`), marks it unlinked, and deletes the newly committed remote object.
- After that, `Read` on the still-open fd returns `readUnlinkedData` first when there are no dirty overlay parts (`pkg/fuse/dat9fs.go:11838-11844`), so it serves the pre-Fsync bytes and loses data that `Fsync` had already reported durable.

I reproduced this deterministically with a review-only test by preinstalling the unlink snapshot, then doing same-fd `Write + Fsync`, confirming the remote object contains the overlay, then calling `Unlink`; the post-unlink read returned the old all-`0x88` prefix instead of the fsynced `aabb` overlay. The current regression blocks the snapshot GET and exercises staleness before attachment; it misses this already-attached stale snapshot state.

Fix gate: `UnlinkedData`/unlinked snapshot state needs a committed revision/size identity and final validation in the mark/commit path, or `Fsync` must invalidate/rebuild any preinstalled unlink snapshot before `Unlink` can mark/delete. Add a regression for: early unlink snapshot installed -> same fd `Write+Fsync` succeeds -> `Unlink` returns -> same fd read preserves the fsynced bytes and does not resurrect the remote path.

Validation I ran: `git diff --check`, gofmt check on touched files, `go test -c ./pkg/fuse`, `go vet ./pkg/fuse`, focused unlink/ShadowSpill tests `-count=10`, and the focused set under `-race -count=3` all pass. Hosted Code CI / Local E2E / CodeRabbit are green on this head. The blocker is a state-machine gap not covered by those tests.

mornyx:
@qiffang The remaining hole was after `snapshotOpenHandlesBeforeUnlink` had already installed `UnlinkedData`.

A later same-FD Fsync advanced `BaseRev` but left that snapshot in place, so `markOpenHandlesUnlinked` skipped a new fetch (`UnlinkedData != nil`) and DELETE dropped the just-committed object.

Fsync/Flush now drops a preinstalled unlink snapshot while the fd is still linked. Installed snapshots also carry revision/size identity; the mark pass discards stale ones and re-snapshots before setting `Unlinked`. Attachment happens before `Unlinked = true`.

Regression: `TestUnlinkAfterPreinstalledSnapshotKeepsLaterFsync`.

qiffang (changes requested):
Reviewed exact head `3d89d2edd477d3c82d1dac489de8b6d7a56bf717`.

The new snapshot revision/size identity and commit-time invalidation correctly close the previously reported case where `Fsync` completes before the final mark pass begins. One adjacent lost-write race remains in `markOpenHandlesUnlinked`, so I am keeping this blocked.

`markOpenHandlesUnlinked` validates every candidate in one loop (`dat9fs.go:7470-7489`), releases all handle locks, and only afterward reacquires each handle to attach the snapshot and set `fh.Unlinked = true` (`dat9fs.go:7494-7533`). A same-FD `Write+Fsync` can commit a newer revision in that gap. In the final loop the old snapshot then fails `unlinkedSnapshotStaleLocked` and is not attached, but the function still unconditionally marks the handle unlinked and proceeds to DELETE the newly committed pathname. The accepted snapshot identity and the unlinked transition are therefore not atomic.

I added a review-only hook exactly between those two loops and reproduced this deterministically 5/5 on this head: seed rev1, start Unlink, after validation write `aabb` at offset 8 and `Fsync` successfully to rev2, then resume attachment/DELETE. The PUT count proves both commits completed, but the still-open FD reads the original rev1 bytes after Unlink instead of the Fsync-successful overlay. The temporary hook/test were removed and the worktree is clean.

Merge gate: make final validation + snapshot attachment + `Unlinked` transition one fenced operation for the captured handle generation, or restart/fail closed whenever any handle's committed revision/size changes before attachment. Never set `Unlinked` when the snapshot needed by that handle became stale or disappeared. Add a discriminating regression for `snapshot validated -> same-FD Write+Fsync -> attach/mark -> DELETE -> same-FD Read`, and cover both in-memory and large pinned-shadow backing.

The new committed tests cover Fsync during the remote GET and Fsync before a later Unlink invocation, but not this validation-to-attachment window. Exact-head Code CI, Local E2E, and CodeRabbit are green; local compile, vet, diff/gofmt checks, and the focused new/adjacent tests (normal 10x, race 3x) pass.

qiffang (changes requested):
Blocking finding on exact head `3d89d2edd477d3c82d1dac489de8b6d7a56bf717`:

`markOpenHandlesUnlinked` still has a split identity/attach transition. The first pass checks for stale snapshots (`dat9fs.go:7468-7487`), but then it releases all handle locks and the final loop later conditionally attaches the snapshot while still unconditionally setting `fh.Unlinked = true` (`dat9fs.go:7490-7528`). If the same open fd does `Write+Fsync` after the stale-check pass but before the attach pass, the handle's `BaseRev`/remote-backed size advances. The attach branch skips because the snapshot is now stale (`dat9fs.go:7510-7512`), but the handle is still marked unlinked. The subsequent DELETE removes the just-fsynced revision, and that open fd no longer has a valid anonymous backing for its committed bytes.

This is the same correctness contract as the prior blocker, just one window later: once unlink starts snapshotting an open fd, any later committed same-fd bytes must either be included in the anonymous backing or the unlink must retry/fail closed. It cannot mark the handle unlinked with no matching snapshot.

Required gate before merge:
- Make the final identity check, snapshot attach, and `Unlinked` transition one fenced operation, or retry/fail closed if identity changes before attach.
- Add a deterministic regression for `snapshot validation -> same-fd Write+Fsync -> attach/delete -> Read`, covering both in-memory snapshot and large pinned-shadow/remote-backed cases.

I rechecked that the PR's existing focused unlink/ShadowSpill tests pass on this head, which is exactly why this needs a new regression: the current tests hit the snapshot-fetch/preinstall windows, not the final mark window.

mornyx:
@qiffang The remaining window was the split between the stale-check loop and the attach/Unlinked loop.

Identity check, snapshot attach, and `Unlinked = true` are now one `fh.mu` critical section, so a same-FD Fsync cannot commit between validation and mark. If a test hook (or a retry after Fsync) sees a stale/missing snapshot, we do not set `Unlinked`; we unmark anything already marked and refetch. Never DELETE a generation the open fd does not have backing for.

Regression: `TestUnlinkMarkWindowConcurrentFsyncKeepsOverlay` (hook between attach and Unlinked: Write+Fsync rev2, then Unlink must keep the overlay). `TestAttachUnlinkedHandleSnapshotDoesNotMarkWhenStale` covers the in-memory and remote-backed attach predicate used by the large pinned-shadow path.

qiffang (approved):
Code review on exact head `d4c1cd0adb02e39f4331b52fccd35e438508db18`: APPROVE.

The previous blocker is closed. `markOpenHandlesUnlinked` now re-runs `attachUnlinkedHandleSnapshotLocked` under the same `fh.mu` hold immediately before `fh.Unlinked = true`; if the handle identity changed, the attach helper fails closed and the outer loop retries instead of marking an fd with a stale/missing anonymous backing. That closes the final validation -> attach/mark race reported on the prior head.

The new `TestUnlinkMarkWindowConcurrentFsyncKeepsOverlay` is discriminating for that window: the test hook performs same-fd `Write+Fsync` between the first attach attempt and the final unlinked transition, then verifies post-unlink reads preserve the fsynced overlay. The existing preinstalled snapshot and large/remote-backed unlink tests still cover the adjacent paths.

Local validation I ran:
- `git diff --check 1a1d8737827d9420fe62d8828fb7e7e0f33dd959...HEAD`
- `gofmt -l pkg/fuse/dat9fs.go pkg/fuse/handle.go pkg/fuse/fuse_test.go`
- focused unlink/ShadowSpill suite `-count=10`
- focused critical unlink suite under `-race -count=3`
- `go test -c ./pkg/fuse`
- `go vet ./pkg/fuse`

A full local `go test ./pkg/fuse -count=1` still fails only in the known Darwin mount-wait permission tests (`TestShouldContinueAfterWaitMountPermissionErrorWhenProbePasses/Fails`), not in this PR's touched state-machine paths.

Merge caveat: at the time of this review, GitHub Code CI and CodeRabbit are green on this head, but Local E2E is still in progress. Do not merge until that required hosted check is terminal green.

qiffang (changes requested):
I need to move this back to REQUEST_CHANGES on exact head `d4c1cd0adb02e39f4331b52fccd35e438508db18`.

The previous single-handle stale-snapshot window is closed, but the new implementation still exposes a partially-unlinked handle state before the unlink operation reaches its commit point. `markOpenHandlesUnlinked` marks handles one at a time: after a handle passes `attachUnlinkedHandleSnapshotLocked`, it immediately sets `fh.Unlinked = true` and unlocks that handle before the rest of the handle set has been fenced and before the remote DELETE (`pkg/fuse/dat9fs.go:7458-7493`).

That state is observable by concurrent operations on the already-marked fd. `Fsync` on such a handle returns `OK` through the `unlinked-discard` branch (`pkg/fuse/dat9fs.go:13908-13914`), and `discardUnlinkedHandleStateLocked` clears `Dirty`, `DirtySeq`, and staged state (`pkg/fuse/dat9fs.go:1479-1484`). If another open handle on the same path then changes the revision/size and forces the mark pass to retry, rollback only clears `Unlinked`/snapshot state for the already-marked handles; it cannot restore the dirty bytes that the successful `Fsync` discarded. A later successful retry + DELETE leaves the first fd without data that the caller wrote and fsynced successfully.

Fix gate: make final identity check, snapshot attach, and `Unlinked` transition atomic for the whole open-handle set, or otherwise prevent any handle from taking the unlinked-discard path until the unlink commit point is irrevocable. If any handle's identity changes, the whole pass must retry/fail without having made another handle discard writes. Please add a deterministic two-fd regression: open the same path twice, let the mark pass mark fd A, then before fd B is finalized do `Write+Fsync` on fd A and mutate/fync fd B to force retry; after unlink succeeds, fd A must read back the fsynced bytes and the remote pathname must not be resurrected.

qiffang (changes requested):
Blocking: multi-handle partial marking exposes a transient `Unlinked` state and can discard a successful same-FD Write+Fsync.

At `pkg/fuse/dat9fs.go:7458-7483`, handles are attached/marked one at a time. After handle A is set `Unlinked=true` and unlocked, the loop has not yet validated/marked handle B. A concurrent Write+Fsync on A is therefore accepted, but Fsync takes the `fh.Unlinked` branch at `pkg/fuse/dat9fs.go:13911-13914`; `discardUnlinkedHandleStateLocked` clears its dirty generation and returns OK. If B then advances the committed revision while its hook/attach is in progress, the pass retries and can ultimately complete Unlink, but A has already lost the bytes whose Write and Fsync both returned success.

I reproduced this deterministically on exact head `d4c1cd0adb02e39f4331b52fccd35e438508db18`: open the same 192-byte file through two writable FDs; during the second per-handle transition, Write+Fsync `aabb` through the already-marked FD, then Write+Fsync `ccdd` through the still-unmarked FD so the current snapshot becomes stale. Unlink returns OK, but the first open FD reads the original bytes at offset 8 instead of `aabb`. The regression failed 10/10 normally and 3/3 under `-race`. The current single-handle test cannot exercise this partial-pass state.

Merge gate: no handle may become observably unlinked until the whole captured handle set has a valid snapshot/anonymous backing and the transition can commit, or rollback must preserve every successful operation performed during a partial pass. Add a two-writable-handle regression covering partial mark -> successful Write+Fsync on the already-marked FD -> later-handle stale/retry -> Unlink -> reads from both open FDs retain their successful writes.

Existing focused unlink/ShadowSpill tests pass 10x and key race tests pass 3x; compile-only package test, vet, gofmt, and diff-check also pass. Hosted Code CI, Local E2E, and CodeRabbit are green, but they do not cover this inter-handle window.

PR #908, opened 2026-09-08
fix(fuse): sync pending-new rename on close-sync mounts

mornyx:
## Summary

Close-sync/write-sync mounts still treated ordinary pending-new `rename(2)` as best-effort. Agent `write tmp; mv tmp final` returned success as soon as the local shadow moved, while the final-path PUT sat in `CommitQueue`. After ~1 minute of upload failure, `onCommitTerminalFailure` dequeued the entry and marked conflict. Sandbox TTL / undrained umount then deleted the shadow. Remote kept the old file: silent rollback.

Git loose-object `tmp_obj_* -> <sha>` already committed synchronously. This extends that contract to close-sync/write-sync using the **mount** write policy (rename has no open-file WritePolicy).

Writeback stays asynchronous. Server CAS / multi-mount policy is unchanged. This does not address sandbox `rm -rf` of whole trees.

## Change

- `renamePendingNewCommit` calls `commitNowPathLocked` before returning when `WritePolicyCloseSync` or `WritePolicyWriteSync` (same path as git loose objects).
- Sync upload failure returns `EAGAIN`; local shadow/pending stay recoverable. It does **not** go through `onCommitTerminalFailure`.
- Enqueue-failure fallback to sync commit is unchanged for writeback.
- Existing-target pending-new rename still uses remote rename fallback (no mistaken create).
- `onCommitTerminalFailure` now takes the causing error and logs a stable FUSE-local alert line:

  `drive9: error event=commit_terminal_failure path=... err=...`

  Collect FUSE stderr or `~/.cache/drive9/mount-logs/` **outside** the sandbox. Retry `upload attempt N/5` lines are not `drive9: error`. The drive9 server never sees this log.
- Perf snapshot gains `commit_terminal_failure` (writeback async drops still need this). `commit_failure` is unchanged.
- `drive9 mount drain` already fails when `commit_queue_conflicts != 0`. Close-sync rename failure no longer depends on drain to become visible.

## Tests

- `TestRenamePendingNewCommitCloseSyncUploadsBeforeReturning`
- `TestRenamePendingNewCommitWriteSyncUploadsBeforeReturning`
- `TestRenamePendingNewCommitWriteBackAllowsAsyncUploadAfterReturning`
- `TestRenamePendingNewCommitCloseSyncFailureReturnsErrorAndKeepsShadow`
- `TestRenamePendingNewCommitWriteSyncFailureReturnsErrorAndKeepsShadow`
- Existing git-loose / target-exists rename tests
- `TestCommitQueueNonConflictErrorUnchanged` asserts the alertable log line
- `TestFlush_CloseSyncUploadsBeforeReturning`

```bash
go test ./pkg/fuse/ -count=1   -run 'TestRenamePendingNewCommitCloseSyncUploadsBeforeReturning|TestRenamePendingNewCommitCloseSyncFailureReturnsErrorAndKeepsShadow|TestRenamePendingNewCommitWriteSyncUploadsBeforeReturning|TestRenamePendingNewCommitSyncCommitsGitLooseObjectFinalPath|TestRenamePendingNewCommitGitLooseObjectSyncFailureKeepsRecoverableShadow|TestRenamePendingNewCommitFallsBackWhenFinalTargetExists|TestFlush_CloseSyncUploadsBeforeReturning|TestCommitQueueNonConflictErrorUnchanged'
```

## Out of scope

- kimi `rm -rf output/app` / whole-tree DELETE at 09-04 02:24, 09-05 07:36, 09-05 08:07
- actionCard zero PUT/rename (dropped before the request)
- Changing writeback default to sync

## Summary by CodeRabbit

* **New Features**
  * Renaming newly created files now completes remote uploads before returning when durable write settings are enabled.
  * Write-back settings continue to defer uploads until queued work is processed.
  * Failed synchronous uploads return a retryable error while preserving local data for recovery.
  * Added monitoring for terminal commit failures, including counters and alertable diagnostics.

* **Bug Fixes**
  * Terminal commit failures now retain the triggering error for clearer troubleshooting.
  * Stale-generation failures no longer generate misleading alerts or affect newer pending data.

### DIFF
--- a/docs/design/fuse-durability-policy.md
+++ b/docs/design/fuse-durability-policy.md
@@ -117,6 +117,10 @@ Therefore:
 
 - `close-sync` is primarily enforced in `Flush`;
 - `write-sync` is enforced at the end of `Write`;
+- pending-new `rename(2)` (`write tmp; mv tmp final`) is also remote-durable
+  on `close-sync` and `write-sync` before rename returns. Upload failure
+  surfaces as `EAGAIN` and keeps the local shadow. Writeback/`auto`/
+  `interactive`/`fsync` mounts still enqueue that rename asynchronously;
 - `Release` handles cleanup and fallback synchronization, but it is not the
   primary close-sync error propagation point.
 
--- a/pkg/fuse/commit_queue.go
+++ b/pkg/fuse/commit_queue.go
@@ -1697,9 +1697,9 @@ func (cq *CommitQueue) commitBatch(entries []*CommitEntry) {
 				if entry == nil {
 					continue
 				}
-				if errors.Is(cq.validateEntryPayloadFresh(entry), errCommitPayloadStale) {
-					safeLogPrintf("commit queue: stale batched payload rejected for %s: %v", entry.Path, err)
-					cq.onCommitTerminalFailure(entry, err)
+				if verr := cq.validateEntryPayloadFresh(entry); errors.Is(verr, errCommitPayloadStale) {
+					safeLogPrintf("commit queue: stale batched payload rejected for %s: %v", entry.Path, verr)
+					cq.onCommitTerminalFailure(entry, verr)
 					cq.endInFlight(entry)
 					continue
 				}
@@ -2768,10 +2768,6 @@ func (cq *CommitQueue) onCommitPostUploadFailure(entry *CommitEntry, err error) 
 }
 
 func (cq *CommitQueue) onCommitTerminalFailure(entry *CommitEntry, lastErr error) {
-	if cq.perf != nil {
-		cq.perf.commitFailure.add(1)
-		cq.perf.commitTerminalFailure.add(1)
-	}
 	// Mark the entry as conflicted in the pending index so that crash
 	// recovery (RecoverPending) skips it instead of retrying forever.
 	// Preserve both the shadow file and the pending metadata so the user
@@ -2786,6 +2782,7 @@ func (cq *CommitQueue) onCommitTerminalFailure(entry *CommitEntry, lastErr error
 			if err != nil {
 				safeLogPrintf("commit queue: failed to mark conflict for %s: %v (entry remains queued)", entry.Path, err)
 				cq.removeFromQueue(entry)
+				cq.recordCommitTerminalFailure(entry, lastErr)
 				return
 			}
 			if !marked {
@@ -2799,6 +2796,7 @@ func (cq *CommitQueue) onCommitTerminalFailure(entry *CommitEntry, lastErr error
 			// silently dropping it.
 			safeLogPrintf("commit queue: failed to mark conflict for %s: %v (entry remains queued)", entry.Path, err)
 			cq.removeFromQueue(entry)
+			cq.recordCommitTerminalFailure(entry, lastErr)
 			return
 		}
 	}
@@ -2814,7 +2812,14 @@ func (cq *CommitQueue) onCommitTerminalFailure(entry *CommitEntry, lastErr error
 	// Remove from queue AFTER all cleanup so WaitPath sees the entry
 	// until bookkeeping is complete.
 	cq.removeFromQueue(entry)
+	cq.recordCommitTerminalFailure(entry, lastErr)
+}
 
+func (cq *CommitQueue) recordCommitTerminalFailure(entry *CommitEntry, lastErr error) {
+	if cq.perf != nil {
+		cq.perf.commitFailure.add(1)
+		cq.perf.commitTerminalFailure.add(1)
+	}
 	// Alertable FUSE-local error. Collectors should match
 	// "drive9: error event=commit_terminal_failure" on stderr / mount-log
 	// files (typically ~/.cache/drive9/mount-logs/). The drive9 server never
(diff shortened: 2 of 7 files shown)


srstack (commented):
Reviewed the full diff, ran the PR's test list plus `TestCommitQueue|TestOnCommit|TestRename` locally (all pass, pkg/fuse 15s), and cross-checked against an end-to-end reproduction of the 2026-09-04 silent-rollback incident on the production build (41503f74): `write .tmp + mv` returned 0 in ~100ms while the final-path PUT sat in the queue, 5 retries exhausted → `giving up … after 5 retries` → terminal failure with data only in the local shadow, the server saw nothing but HEAD/GET 404 probes, `mount drain` correctly reported `commit_queue_conflicts=2` (exit 1), and a fresh-cache remount (= sandbox recycle) made the file vanish everywhere. This PR converts exactly that `mv` into a synchronous commit or an immediate `EAGAIN`, which is the right contract.

**[MAJOR] (deployment, not code)** The sync-rename path only activates for `WritePolicyCloseSync`/`WritePolicyWriteSync`. Default `-durability auto` resolves to `WritePolicyWriteBack` (`parseFuseDurability`), which stays async by design. So this PR changes nothing for mounts that don't explicitly opt into close-sync/write-sync. The incident sandbox config needs a companion change (mount with `-durability close-sync` or a profile that maps to it) for the fix to actually protect warden workloads — could you confirm that's tracked (runbook/release note/config PR), not just implied? The new `drive9: error event=commit_terminal_failure` line plus `commit_terminal_failure` counter covers the writeback remainder, which is a good stopgap.

**[MINOR]** `recordCommitTerminalFailure` emits `err=<nil>` for the genuine-conflict terminal paths (`DisableAutoResolveLWW`, payload-base-older-than-server, ShadowSpill genuine content/size mismatch, layer-mount auto-resolve unsupported). Collectors matching `event=commit_terminal_failure` will treat these as upload failures even though they're concurrency conflicts. Consider adding a `reason=upload_failure|conflict|abandoned` field so alert severity can distinguish "remote never got the data (retryable class)" from "genuine conflict (needs human merge)".

**[MINOR]** `commit_failure` counter semantics change slightly: the stale-generation path (`MarkConflictIfGeneration` returning `!marked`) used to increment `commit_failure` and now increments nothing. Intentional per `TestOnCommitTerminalFailureStaleGenerationSkipsAlert`, but anyone dashboarding `commit_failure` should expect a small drop.

**[NIT]** Pre-existing, not introduced here: the two "failed to mark conflict … (entry remains queued)" log lines immediately call `removeFromQueue(entry)`, so the parenthetical is misleading. Fine to leave for a follow-up.

**[NIT]** The sync rename path is single-attempt (`commitNowPathLocked` → one `uploadEntry`, no retry). A transient blip now surfaces as `EAGAIN` on `rename(2)`. That's consistent with the existing git-loose-object contract and callers can retry, but worth a conscious yes.

What looks good: EAGAIN-on-failure with shadow/pending preserved and verified recoverable through remount (the failure test re-Lookup/Open/Reads the recovered file — nice); ShadowSpill getting a size-aware budget instead of the 30s request deadline; the stable, documented alert line with retry attempts explicitly kept out of the `drive9: error` namespace; stale-generation correctly suppressing both the alert and the conflict marker; and the drain.go comment making conflict visibility expectations explicit.

Approving the approach; the only thing I'd like answered before merge is the MAJOR deployment-coverage question above.

srstack (approved):

mornyx:
@srstack thanks — addressing the review here.

**MAJOR (deployment):** Agreed this PR does not change `auto`/`interactive`/`fsync` (still writeback + async rename). The 2026-09-04 silent-rollback incident was already on **close-sync** (`WritePolicyCloseSync` covered Flush/Release but not pending-new `mv`). This PR is the client fix for that contract; writeback default stays async on purpose.

Warden/sandbox mounts that need the same guarantee still have to opt into `--durability=close-sync` (or `write-sync` / a profile that maps to it). I documented the rename barrier in `docs/design/fuse-durability-policy.md`. Config/runbook for the sandbox launcher is outside this client PR.

**MINOR (`reason=`):** Done in `10b32236`. The alert line is now:

`drive9: error event=commit_terminal_failure path=... reason=upload_failure|conflict|abandoned err=...`

Collectors can keep matching `event=commit_terminal_failure` and split severity on `reason=`.

**MINOR (`commit_failure` drop on stale generation):** Intentional. That path is not a terminal failure of current data (newer pending generation exists), so neither the counter nor the alert fires. Dashboards on `commit_failure` may see a small drop.

**NIT (remains queued):** Fixed those two logs: pending/shadow preserved, in-memory queue entry removed.

**NIT (single-attempt sync rename):** Conscious yes — same as git loose-object rename. Caller sees `EAGAIN` and can retry; we do not re-enqueue (that would recreate silent success / double-upload).

srstack (commented):
Follow-up on `10b32236` — all review points are addressed:

- `reason=upload_failure|conflict|abandoned` with a unit-tested classifier, and the `errLayerRolledBack` wrap on the abandoned path so classification actually fires — exactly what I had in mind.
- The two misleading "(entry remains queued)" lines now state what actually happens (pending/shadow preserved, in-memory entry removed).
- `docs/design/fuse-durability-policy.md` documents the close-sync/write-sync rename barrier, and good to have confirmation that the incident mount was already close-sync — that makes this PR the direct fix for the incident path rather than a partial one.

Re-ran locally on `10b32236`: `TestClassifyCommitTerminalReason`, `TestCommitQueueNonConflictErrorUnchanged`, `TestOnCommitTerminalFailureStaleGenerationSkipsAlert`, `TestRenamePendingNewCommit*` — all pass. CI is still pending at the time of writing; nothing left blocking from my side.

Issue #907, opened 2026-09-08
FUSE: support multi-connection SQLite WAL (writer + checkpoint + readers) — tracking issue


mashenjun:
### Summary

Track the work to support multi-connection SQLite WAL workloads on Drive9 FUSE
with `durability=fsync` and append-log enabled: one connection writing, one
connection running `PRAGMA wal_checkpoint(TRUNCATE)`, and concurrent readers on
the same mount. SQLite itself serializes mutators via POSIX locks, so "multi
connection" here means turn-taking mutations across handles, not byte-level
concurrent writers.

This extends the accepted append-log scope. The current spec explicitly marks
multi-writer coordination out of scope (`docs/design/s3-express-append-log-fuse.md`,
"Out of scope", item 5, and section 3.2). The proposal below narrows that line
to "same-mount multi-connection" and keeps multi-mount / multi-process
concurrent writers out of scope.

### Background

#902 reports three failure modes in one scale/concurrency matrix (4 MiB WAL,
TRUNCATE checkpoint, main.db at 1 MiB / 2 MiB / 2 GiB):

A. A fresh checkpoint fd does ~400 serial 4 KiB remote WAL reads. SQLite
   sidecars bypass the generic whole-file cache by design, and ordinary
   append-log commits do not maintain an equivalent reusable committed view,
   so an ~12 ms ext4 checkpoint takes ~38 s.
B. After the checkpoint truncates the WAL and publishes a new revision, the
   pre-opened writer handle writes the new 32-byte generation header but its
   fsync returns EEXIST -> `SQLITE_IOERR_FSYNC`.
C. With a 2 GiB main.db, growing it by only ~1 MiB disables PATCH
   (`size <= OrigSize`), hydrates all 257 old parts, builds full-size in-memory
   copies, hits an already-expired 30 s fsync context at multipart initiation,
   and returns `SQLITE_IOERR_FSYNC`; close then re-publishes the whole image
   under a size-aware timeout.

The bugs are tracked in #902; this issue is the implementation umbrella that
decomposes the fix into bounded workstreams and records the PATCH decision.

### Decision: growth goes through PATCH

The server already supports growing an S3-backed `single`-layout file through
the patch plan: `appendDirtyPartNumbers` computes the dirty tail parts, and the
plan presigns parts beyond the original object without a `read_url`
(`tidbcloud/fs pkg/backend/patch.go`). The only blocker is client-side: two
[...]

PR #910, opened 2026-09-08
fuse: support multi-connection SQLite WAL with growth PATCH (#907)

mashenjun:
## Summary

Implements #907: same-mount multi-connection SQLite WAL (writer +
`PRAGMA wal_checkpoint(TRUNCATE)` + concurrent readers) on FUSE with
`durability=fsync` and append-log. Fixes the three #902 failure modes.

Design: `docs/design/multi-connection-sqlite-wal-fuse.md` (two review rounds,
verdict READY). Scope is unchanged from #907: no persistent schema, no new
lock, no new worker, no second commit state machine.

### Changes

1. W1: rebind the append-log baseline when a clean sibling adopts a
   zero-truncate commit, and admit a 0-byte baseline to the generation-reset
   route (fixes the post-checkpoint EEXIST / `SQLITE_IOERR_FSYNC`).
2. W2: serve checkpoint sidecar reads from a clean sibling at the latest
   committed revision (revision-gated; the `-shm` short-read guard is
   preserved). The sub-64 MiB write-through shadow leg was dropped as dead
   code in `c61ef5d1`.
3. W3: allow PATCH when `main.db` grows, behind a growth-region
   dirty+loaded coverage guard; reroute strict-fsync growth through the
   patch-capable flush path; full-rewrite fallback on patch-plan rejection.
4. W4: size-aware timeout for the strict-sync writeback upload (T4.1 deferred
   after a scope audit).

Tests: 17 targeted white-box tests in `pkg/fuse` plus the existing
`pkg/client` patch tests; all pass. `gofmt`, `go vet`, and `go build ./...`
are clean. Full regression / e2e / FUSE-mount suites were not run locally.

## Validation

Environment:

1. Cluster `dev-drive9-eks-us-east-1`, service `drive9-tidbcloud/drive9-server-lb`.
2. Endpoint `http://k8s-drive9ti-drive9se-a6c8a08117-cc4e5d0b88a22b64.elb.us-east-1.amazonaws.com` (HTTP only).
3. Existing dev tenant, status `active`, `max_upload_bytes=10737418240`,
   `storage_capabilities.append_log_v1=true`.
4. EC2 `54.87.177.246`, ext4 workdir, `/dev/fuse` present; binary
   `drive9-linux-amd64` built from commit `42919475`.
5. Harness `wal4m_scale.py` from #902, sha256
   `d9f7ba10a59d6c24258a1dca8dcbe8c73d2998283ab02350df92e9ac11c0aaa5`; SQLite
   3.46.1, page size 4096, 1000-frame / 4,120,032-byte WAL.

Mount flags: `--mode=fuse --durability fsync --profile coding-agent
--append-log '**/*wal' --append-log '**/main.db-wal'
--writeback-batch-window 20ms --debug --perf-dir ...`. `--direct-mount-strict`
is not permitted for the EC2 user; the run fell back to fusermount.

Results:

| Metric | 1m | 2g |
| --- | --- | --- |
| status | ok | ok |
| primary checkpoint | 2.54 s (was ~38 s) | 16.14 s (was ~91 s) |
| final checkpoint | 0.72 s | 12.22 s |
| checkpoint result | [0,0,0] | [0,0,0] |
| EEXIST / SQLITE_IOERR_FSYNC | none | none |
| live verification | ok, state/sequence exact, 0 mismatches | ok, state/sequence exact, 0 mismatches |
| growth commit | flushHandle PATCH, 1 dirty part | flushHandle PATCH, 18 dirty parts |
| mount peak RSS | - | 660 MB |

Mechanism evidence from the mount debug log:

1. `sibling_truncate_rebind` x8 with the expected field transitions.
2. Checkpoint reads: `sqlite-sidecar-clean-sibling` x2747 (~50 us) vs
   `sqlite-sidecar-clean-remote` x236 (~18 ms).
3. `perf append_log success=411 conflict=0 error=0 full_rewrite_count=4
   generation_reset_count=0`.
4. `generation_reset_count=0` because the harness overlap writer opened the
   existing 4 MiB WAL (per-handle H0 unproven), so the post-checkpoint header
   took the tail-append route after the W1 rebind; the generation-reset route
   for the proven-H0 topology is covered by
   `TestW1TruncateCommitThenWriterHeaderTakesGenerationReset`.

Evidence bundle: results JSONL, mount debug log, RSS samples
(`issue-907-validation.tgz`, retained locally).

## Not in this PR

1. T3.5: skip `read_url` download in `uploadPatchPart` (optional).
2. T4.1: derive the lazy part loader context from the parent (deferred after
   scope audit: ~26 call sites for a bounded, marginal benefit).
3. W6.3: server-side growth patch-plan contract test (lives in the
   `tidbcloud/fs` repo).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved SQLite WAL handling across multiple connections, including safer truncate and generation-reset behavior.
  - Prevented stale clean-sibling reads by revalidating data before returning it.
  - Improved sparse file growth handling, routing eligible updates through PATCH while failing closed on conflicts.
  - Removed unsafe full-upload and fsync rerouting paths.

- **Documentation**
  - Added and updated design reviews covering the multi-connection SQLite WAL proposal, resolved findings, and remaining concerns.

- **Tests**
  - Expanded coverage for truncate rebinding, generation resets, checkpoint reads, sparse growth, PATCH uploads, and revision conflicts.


qiffang (changes requested):
结论：REQUEST_CHANGES / 暂不放行。

我按 exact head `e7ef4508994c4af29836324ecdc45ccdb076fcf6` 复核了 diff、PR comments、#907/#902 语义、本地测试和 hosted checks。方向上 W1/W2/W3 的目标是对的，但当前还有这些必须先闭合的 correctness/CI blocker：

1. **Required Code CI 仍是红的**  
   hosted Code CI 和我本地 `make lint` 都在 `pkg/fuse/w3_growth_patch_test.go:241` 复现 staticcheck S1021：`var ts *httptest.Server` 应与下一行赋值合并。required check 不绿不能放行。

2. **append-log legacy fallback 被 growth PATCH 误改写**  
   现有回归 `TestAppendLogLegacyUploadPlanUnsupportedReroutesOnceToFullPUT` 在当前 head 失败：服务端收到 `PATCH /v1/fs/db-wal`，而测试期望 legacy/upload-plan unsupported 后只做一次 full PUT/full rewrite fallback。PR 放宽 `growthPatchSafeLocked()` 后，append-log-configured WAL 在 known single layout 路径会绕进 generic PATCH；这会改变 WAL append-log fallback 的传输语义。请收窄 `usePatch/canPatchExisting`，或在 append-log fallback/legacy unsupported 状态下强制 full upload，并补对应回归。

3. **lazy-buffer growth fallback 也被 PATCH 改坏**  
   `TestFlushGrowthLazyBufferFetchesMissingParts` 当前 head 失败：`flushHandle` 和 `write-sync` 子用例都收到意外的 `PATCH /v1/fs/growth-lazy-fetch.dat`。这个用例本来证明 full materialization 会通过 lazy loader 补齐缺失 part；现在直接走 PATCH 会跳过该不变量。请明确 lazy-backed growth 的 PATCH 适用边界，并补测试证明需要 materialize/fetch missing parts 的场景不会被 generic PATCH 误吃掉。

4. **write-sync 大顺序写 streaming 语义被 PATCH 改坏**  
   `TestWriteSyncDisablesStreamingUploaderForLargeSequentialWrites` 当前 head 失败：write-sync 路径出现意外 `PATCH /v1/fs/stream_sync.bin` 并导致第二次 write 返回 `EIO`。write-sync 下原有“禁用 streaming uploader / 同步上传”的边界不能被 growth PATCH 旁路；请收窄 PATCH 条件并补回归。

5. **PATCH plan initiate error 分类过宽**  
   `PatchFile` 现在把所有非 202 plan 响应都包装成 `ErrPatchPlanInitiate`；FUSE 侧据此设置 sticky `patchPlanRejected` 并 materialize/full-upload。这样 409 revision conflict、401/403、429、5xx 都会被误当成“可恢复的 plan rejection”，导致多 GiB SQLite 文件做不必要全量上传并屏蔽真实故障。请只对明确的 capability/plan-validation rejection 允许 fallback；CAS/鉴权/限流/5xx 必须 fail closed，并补 409/5xx 与合法 plan rejection 的单测。

6. **strict fsync 的 PATCH reroute 没有和 active writeback uploader 串行**  
   W3 分支里 `RemoveIfGeneration` 只能移除 cache entry；如果 `WriteBackUploader.uploadOne` 已经拷出同 generation bytes 并正在上传，它不会被取消。随后 `flushHandle` PATCH 会和这个 full upload 用同一 base revision 并发提交，strict fsync 可能非确定性 conflict/EEXIST/IOERR。请在丢弃 snapshot/走 PATCH 前等待同 path in-flight uploader，补“background 已拷出并阻塞，fsync reroute 必须等待/不双提交”的回归。

7. **W2 clean-sibling latest revision gate 不是原子边界**  
   `readSQLitePersistentJournalVisibleRange` 捕获 `latestCommittedRevision` 后再复制 sibling；如果同路径 commit 在复制期间推进 revision，而 refresh 因 `TryLock` 跳过被 reader 锁住的旧 sibling，当前 read 仍可能返回 stale BaseRev 的 WAL bytes。请在拷贝后重查 latest revision 或持有 per-path fence，并补确定性并发回归。

另外 CodeRabbit 指出的 `TestW2CleanSiblingShadowServesSidecarRead` fixture 问题也建议一起修：现在 shadow bytes 和 `writer.Dirty` 相同，不能证明真的读了 shadow leg。

本地验证：
- `git diff --check origin/main...HEAD` 通过；
- touched Go files `gofmt -l` 无输出；
- `make lint` 失败，同 hosted Code CI；
- `go test ./pkg/fuse -run '^(TestW1|TestW2|TestW3|TestFlushHandle_Path2_GapGrowthBypassesPatch)$' -count=1` 通过；
- `go test ./pkg/fuse -run '^TestAppendLogLegacyUploadPlanUnsupportedReroutesOnceToFullPUT$' -count=1` 失败如上；
- `go test ./pkg/fuse -skip 'WaitMount' -count=1` 失败，包含上述 append-log fallback、lazy-buffer growth fallback 和 write-sync streaming 回归。

qiffang (changes requested):
结论：**REQUEST_CHANGES / 不放行**。我按 exact head `e7ef4508994c4af29836324ecdc45ccdb076fcf6` 复核了 diff、现有 comments/checks，以及 WAL/ShadowSpill/growth PATCH 的关键路径。方向上 W1/W3 的主修复意图是对的，但当前实现仍有下面阻塞点；这些闭合前不能合并。

### 1. strict fsync 的 growth PATCH reroute 没有等待已在飞的 writeback upload

位置：`pkg/fuse/dat9fs.go:14160-14171`，对照 `pkg/fuse/writeback_uploader.go:380-468` / `:486-488`。

`Fsync` 发现 grown S3-backed large file 后，直接 `RemoveIfGeneration()` 丢 writeback snapshot，然后调用 `flushHandle()` 走 PATCH。问题是：如果 background `WriteBackUploader.uploadOne()` 已经在 `GetMetaAndViewWithCallback()` 后把同一 generation 的 bytes 拷到局部变量并开始上传，`RemoveIfGeneration()` 只能删 cache 条目，**不能取消这次 in-flight full upload**。这时 fsync 的 PATCH 和 background full upload 会用同一个 `expectedRevision` 并发提交，输的一方变成 409；如果 background 先赢，fsync 可能向 SQLite 返回 `EEXIST`/`SQLITE_IOERR_FSYNC`。这正好打破 PR 要修的 2GiB fsync 稳定性。

必须补：reroute 前和 active writeback uploader 的 same-path in-flight 状态形成明确串行边界，并加回归覆盖“background 已 snapshot/上传中时 strict fsync reroute 不冲突”。

### 2. W2 clean-sibling read 的 latest-revision gate 不是原子边界，且 shadow 分支读的是 path 而不是稳定 generation

位置：`pkg/fuse/dat9fs.go:4739-4785`。

当前先读 `latestCommittedRevision(path)`，再去扫描/拷贝 clean sibling。这个检查没有和 candidate snapshot 原子化：如果另一个 same-path commit 在 `latest` 被捕获后推进全局 revision，而 `refreshCommittedRevisionForOpenHandlesWithSize()` 因为 `TryLock()` 跳过了被 reader 锁住的旧 sibling，这里仍可能返回旧 `BaseRev == capturedLatest` 的 WAL bytes。多连接 checkpoint 里，shared `-shm` 可能已经指向新 revision，此时会再次出现 stale/short WAL read。

同一个函数的 shadow leg 还有额外风险：`src` eligibility 在 `src.mu` 下判断完后立刻 unlock，然后用 `fs.shadowStore.ReadAt(path, ...)` 读当前 active path。解锁后 same-handle write 或 same-path shadow cleanup/recreate 可以改变 path 对应的 shadow；这会让 clean-sibling 快路径读到不属于已验证 revision 的 bytes，甚至读到未提交的新 shadow。已有 generation-safe shadow API 没有在这里使用。

必须补：读取前后要证明 revision 仍是 latest，且 shadow 数据源必须绑定到验证过的 stable generation/handle state；同时补确定性并发回归覆盖 latest advance、TryLock skip、shadow path 被同 handle写入/替换的窗口。

### 3. `ErrPatchPlanInitiate` 过宽，把 operational/冲突错误当成可 full-upload fallback 的 plan rejection

位置：`pkg/client/patch.go:110-116`，调用方在 `pkg/fuse/dat9fs.go:13022-13047`、`:15603-15614`。

现在任何 PATCH plan 阶段非 202 都 wrap 成 `ErrPatchPlanInitiate`，包括 409 revision conflict、401/403 auth、429 throttling、5xx、以及 transient network/server error。FUSE 调用方随后会设置 sticky `fh.patchPlanRejected=true`，materialize 整个 multi-GiB 文件，并重试 full upload；这会把本应 fail-closed/可重试的操作错误变成大内存/大网络 fallback，并且让同一 handle 后续永久避开 PATCH。

必须补：只有明确的 capability/plan-validation rejection 才能触发 full-upload fallback；409、auth、throttle、5xx、decode/transport 等失败必须保持原错误语义并 fail closed。需要覆盖 409 和 5xx 不设置 `patchPlanRejected`、不触发 full upload 的回归。

### 4. 测试/CI 仍未达到质量门

- GitHub Code CI 当前红：`pkg/fuse/w3_growth_patch_test.go:241` staticcheck S1021。
- `pkg/fuse/w2_checkpoint_read_test.go:103-129` 的 shadow fixture 把 `shadowStore` 和 `writer.Dirty` 写成同一内容，不能证明真的走了 shadow leg；如果代码退回 buffer leg 也会过。这个测试必须把两者内容区分开。
- 缺第 1/2/3 点对应的并发和错误分类回归；现有 Local E2E 绿不能覆盖这些 failure windows。

修完后请重新跑 exact-head Code CI / Local E2E，并贴关键回归说明后再复审。

mornyx (changes requested):
## Review summary

This PR lands same-mount multi-connection SQLite WAL support: W1 rebinds the append-log baseline after a sibling zero-truncate, W2 serves checkpoint sidecar reads from a revision-gated clean sibling, W3 relaxes PATCH for `main.db` growth behind a dirty+loaded coverage guard, and W4 makes the remaining strict-sync upload timeout size-aware. W1 is in the path that actually runs (before the shadow-removed continue, one-shot `sqliteWALTruncated` latch, H0 preserved), and `growthPatchSafeLocked` is the right fail-closed shape for sparse / evicted growth. Under a strict maintainability and correctness bar this is **not merge-ready**: the new strict-fsync PATCH shortcut races an in-flight writeback upload into `EEXIST` / `SQLITE_IOERR_FSYNC`, growth PATCH now intercepts WAL and several existing full-PUT tests so `make test` is red, and `make lint` fails staticcheck S1021.

**What looks good**
- W1 rebind sits in `refreshCommittedRevisionForOpenHandlesWithSize` before the shadow-removed early-continue; the 0-byte generation-reset gate keeps the upstream fences; the H0-after-truncate test covers the topology the live harness did not.
- `growthPatchSafeLocked` encodes the server contract (beyond-original parts are client-constructed even when absent from the dirty set) and ships with a falsifying matrix (contiguous / sparse / evicted).
- W4's size-aware `releaseTimeout` on the remaining writeback `UploadSyncWithRevision` path is the right deadline for the 2 GiB case when PATCH is not taken.

---

**1. Strict-fsync PATCH reroute races an in-flight writeback upload (`pkg/fuse/dat9fs.go:14164`)**

The new W3 branch discards the snapshot with `RemoveIfGeneration` and immediately `flushHandle`s a PATCH on the same `expectedRevision`. That does **not** cancel a worker that already copied the generation: `WriteBackUploader.uploadOne` (`pkg/fuse/writeback_uploader.go:380-419`) snapshots meta+bytes first, then does a conditional full upload. `UploadSyncWithRevision`, the path this branch replaces, starts with `WaitPath` (`writeback_uploader.go:488-489`). `uploadOne` does not take `remoteCommitLock`, so `flushHandle`'s commit fence does not serialize with it.

If the worker has already copied this generation, both commits CAS the same base. The PATCH loser returns conflict → SQLite `EEXIST` / `SQLITE_IOERR_FSYNC` even when the bytes are already durable (`WriteBackSeq == DirtySeq` means the in-flight payload **is** this fsync's dirty image). After a successful in-flight upload, `OnSuccess` refreshes siblings with `TryLock` and skips this handle because Fsync holds `fh.mu`, so a follow-on PATCH still uses the stale base. The comment ("cannot re-upload a stale full image") is false for an upload already in flight. `TestW3FsyncGrowthReroutesToPatch` never `Submit`s, so it cannot catch this.

Reachability: write-back snapshots exist only below `writeBackThreshold` (10 MiB). The 2 GiB `main.db` case already PATCHes via the fallthrough `flushHandleDebounced` without this shortcut. The new branch is the <10 MiB grown-file shortcut, and it is the one that races.

Fix (prefer the simpler one):

- **Delete the reroute.** For ≥10 MiB there is no snapshot and `flushHandle` already PATCHes under the relaxed `usePatch` gate. For <10 MiB, `UploadSyncWithRevision` (which `WaitPath`s) is a bounded full upload and is race-free. That deletes the special case instead of hardening it.
- If the shortcut stays: `WaitPath` first (deadlock-safe: `OnSuccess` uses `TryLock`). If the generation is gone after the wait, adopt the recorded revision, clear dirty, return OK. Only if it is still present, then `RemoveIfGeneration` and PATCH. Do not zero `WriteBackSeq`/`WriteBackGen` until the commit succeeds.

**2. `make test` is red: growth PATCH intercepts full-PUT paths, including WAL (`pkg/fuse/dat9fs.go:12922`, `15395`)**

Verified on this HEAD (`e7ef450`). All three fail here and expect PUT, not PATCH:

| Test | What happens now |
| --- | --- |
| `TestFlushGrowthLazyBufferFetchesMissingParts` | growth with a lazy loader selects PATCH; fixture rejects PATCH |
| `TestWriteSyncDisablesStreamingUploaderForLargeSequentialWrites` | second 8 MiB flush PATCHes → EIO |
| `TestAppendLogLegacyUploadPlanUnsupportedReroutesOnceToFullPUT` | `PATCH /v1/fs/db-wal` before the legacy full-PUT reroute |

Cases 1–2 may be *intended* under the new generic growth-PATCH contract — then update those tests deliberately, do not leave them pinning the old B11 bypass. Case 3 is a **policy leak**, not a stale pin: `tryAppendLogFullRewriteLocked` declines `ContentLayoutSingle` (`pkg/fuse/append_log.go:874-877`) and the new `usePatch` predicate has no SQLite-sidecar / append-log-layout exclusion. The design's transport policy is "append-log remains the transport for the WAL itself; PATCH applies only to `main.db` growth." A legacy-layout `db-wal` flush now issues a PATCH plan.

Gate growth PATCH off `isSQLitePersistentJournalPath` (or require append-log layout to have already taken the rewrite/tail route), update tests 1–2 to the new generic contract, and re-run `make test`.

**3. `make lint` fails staticcheck S1021 (`pkg/fuse/w3_growth_patch_test.go:241`)**

```go
var ts *httptest.Server
ts = httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
```

The handler never closes over `ts`. CI runs `make lint` before tests. Merge into `ts := httptest.NewServer(...)` (the same file already does this at line 70).

---

**4. `ErrPatchPlanInitiate` is far too broad for a sticky full-rewrite (`pkg/client/patch.go:111`)**

Every non-202 from plan initiation — including 409 conflict, 401/403, 429, and 5xx — is wrapped as `ErrPatchPlanInitiate`. Both FUSE callers (`syncWriteHandleToRemoteLocked` ~13022, `flushHandle` ~15603) then set sticky `patchPlanRejected`, `materializeFullForUploadLocked`, and full-upload. Transport errors from `c.do` are *not* wrapped; only the status-code funnel is.

A 409 still uses the same `expectedRevision`, so the fallback cannot win the CAS; it only burns a full-file copy and then fails closed, and wrongly disables PATCH for the rest of the handle life even after a later re-base. A 429/5xx on a multi-GiB `main.db` recreates the memory/network blow-up W3 exists to avoid.

Wrap the sentinel only for known plan-validation / capability 400s (part-count over the S3 limit, invalid plan). Check `errors.Is(err, client.ErrConflict)` before `IsPatchPlanInitiateErr` and fail closed. Do not set `patchPlanRejected` for 5xx/429.

**5. W2 latest-revision gate is not atomic with the sibling copy (`pkg/fuse/dat9fs.go:4739`)**

`latestCommittedRevision` is sampled, then `readSQLitePersistentJournalCleanSiblingRange` `TryLock`s a handle whose `BaseRev` still equals that snapshot. A same-path commit can publish R+1 in between; its refresh also uses `TryLock` and skips the handle the reader is holding, so the copy returns revision-R bytes after R+1 committed. The shadow leg unlocks before `shadowStore.ReadAt` (`dat9fs.go:4776-4778`), and `ReadAt` is path-keyed with no generation token, so a concurrent generation-reset can rotate the file under the read.

This is **not** the #902 short-read matrix by itself (TRUNCATE is not supposed to run with live WAL readers; salt/checksum mismatches treat foreign-generation frames as end-of-log), but the gate does not provide the "equals latest committed revision" property the comment and design risk table claim. Re-read `latestCommittedRevision` after the copy (and after a shadow `ReadAt`) and drop the candidate if it advanced. For the shadow leg, hold the handle across the read or use generation-scoped `ReadAtGen`. No new lock required.

**6. Duplicated PATCH eligibility and fallback blocks in `dat9fs.go`**

The plan-rejected fallback duplicates the adjacent unsupported-storage fallback (~25 lines: materialize → `bytesView` → `WriteStreamConditional`, differing only in the flag vs `StorageClass` side effects). The Fsync reroute inlines a third copy of the PATCH predicate already expressed at `dat9fs.go:12922` and `15395`. Extract `patchRouteEligibleLocked(fh, size)` and one full-upload fallback helper parameterized by the side effects. If the Fsync shortcut is deleted (item 1), that third copy goes away for free.

---

### Nits

- `TestW2CleanSiblingShadowServesSidecarRead` (`pkg/fuse/w2_checkpoint_read_test.go:115`) writes the same bytes into `writer.Dirty` and `shadowStore`, and both legs report `sqlite-sidecar-clean-sibling`. A regression that skips the shadow read still passes. Put distinct bytes in the buffer and keep the shadow bytes as the expected result.
- `growthPatchSafeLocked` mixes 1-based `DirtyPartNumbers` with 0-based `IsPartLoaded` (`dat9fs.go:3916`). It is correct, but the helper doc does not say so. One comment line stating the convention would prevent a future "fix" that breaks the guard.

---

### Recommendation
**Request changes.** The fsync PATCH shortcut can recreate nondeterministic `SQLITE_IOERR_FSYNC`, `make test` is red (including a WAL PATCH leak the design forbids), and `make lint` is red. The patch-plan sentinel breadth and W2 gate atomicity should land in the same pass if practical, but they are not additional merge blockers.

qiffang (changes requested):
#910 @ `e7ef4508994c4af29836324ecdc45ccdb076fcf6` 继续 REQUEST_CHANGES。

主方向（W1 zero-truncate generation reset、W3 growth PATCH safe guard / strict fsync reroute）看起来是对的，targeted FUSE 测试也能跑过；但 W2 新增的 clean-sibling shadow fast path 还有一个合前数据一致性 blocker：

1. `readSQLitePersistentJournalCleanSiblingRange` 先在 `pkg/fuse/dat9fs.go:4768-4778` 判定 sibling 是 `DirtySeq == 0` / no dirty parts / `BaseRev == latest`，随后释放 `src` 锁，再用 `fs.shadowStore.ReadAt(path, ...)` 读 path-global active shadow。释放锁后，同一个 writer handle 可以进入 `Write`。`Write` 对 ShadowSpill 会先在 `pkg/fuse/dat9fs.go:12718-12735` 把新字节写进 shadow store，最后到 `pkg/fuse/dat9fs.go:12799` 才 bump `DirtySeq`；`ShadowStore.ReadAt` 又只是按 path 读 active fd（`pkg/fuse/shadow.go:852-858`），没有 generation pin 或回读后的 DirtySeq/BaseRev revalidate。结果是 checkpoint/reader fd 可在“已确认 clean/latest”的窗口里读到同 writer 尚未 fsync/commit 的新 WAL 字节，违反 W2 的 committed-revision-only 语义。现有 W2 测试只覆盖 shadow happy path和 behind-revision skip，没有覆盖“clean 判定后 same-handle write 修改 shadow”的 interleaving。

Fix gate：shadow 分支需要把最终身份校验、shadow generation、read 做成同一个 fence。可在 handle lock 下捕获并 pin active shadow generation 后用 generation-bound read，并在返回前 revalidate `DirtySeq/BaseRev`；或保持等价的 handle/path/shadow 锁直到读完成。补一个判别测：clean ShadowSpill sibling at latest rev -> reader 判定 eligible 后阻塞 -> same writer 写新 WAL bytes -> reader 不能返回这批未提交 bytes（应等待/回退/失败 closed）。

另外当前 required hosted Code CI 是红的：`make lint` / staticcheck S1021 指到 `pkg/fuse/w3_growth_patch_test.go:241`，`var ts *httptest.Server` 需要和下一行赋值合并。这个是独立的合并门。

我本地验证：`git diff --check`、gofmt check、`go test -c ./pkg/fuse ./pkg/client`、`go vet ./pkg/fuse ./pkg/client`、focused W1/W2/W3/SQLite same-path FUSE tests，以及同一 focused 集合 `-race -count=3` 通过；`pkg/client` runtime focused tests 被本机 rootless Docker TestMain 挡住，没执行到目标测试。

qiffang (changes requested):
Reviewed exact head `e7ef4508994c4af29836324ecdc45ccdb076fcf6`. The W1 baseline rebind and the growth coverage predicate are directionally sound, but this head is not mergeable yet.

1. **W2 can return bytes from the wrong shadow generation and from a no-longer-latest revision.** `readSQLitePersistentJournalCleanSiblingRange` validates `DirtySeq == 0` / `BaseRev == revision` under `src.mu`, then releases that lock and reads the path-global active shadow (`pkg/fuse/dat9fs.go:4768-4785`). A same-path write can replace/mutate the shadow after eligibility but before `ReadAt`; my deterministic review-only test replaced the active shadow in that window and current code returned uncommitted bytes as revision 6 in 10/10 runs. Separately, `readSQLitePersistentJournalVisibleRange` samples `latestCommittedRevision` once (`:4739`) and never rechecks it after copying the candidate. I forced a newer commit to publish while refresh skipped the locked sibling; the reader still returned the old revision-5 buffer as `sqlite-sidecar-clean-sibling` in 10/10 runs. The read must be bound to both a specific shadow generation and a still-current committed watermark through the final copy, with regressions for both interleavings. The committed W2 shadow test uses identical buffer/shadow bytes, so it cannot distinguish the shadow source or either race.

2. **Strict growth `Fsync` races an already-running background full upload.** The reroute only calls `writeBack.RemoveIfGeneration` (`pkg/fuse/dat9fs.go:14155-14171`); that removes the cache record but cannot cancel `WriteBackUploader.uploadOne` after it has copied the payload. Unlike `UploadSyncWithRevision`, this path never waits on the uploader's per-path in-flight fence. I submitted the cached generation, blocked its conditional PUT in the server, then invoked `Fsync`; PATCH started before the PUT was released in 5/5 runs. Both carry the same expected revision, so one CAS loses and SQLite can receive `IOERR_FSYNC`. The new test at `pkg/fuse/w3_growth_patch_test.go:182-225` never calls `Submit`, so it cannot detect this. Fence/cancel the active uploader before PATCH and add this exact blocked-upload interleaving.

3. **`ErrPatchPlanInitiate` incorrectly treats outcome-unknown and semantic failures as safe full-rewrite fallbacks.** `pkg/client/patch.go:110-116` wraps every non-202 response (including 409, auth failures, 429 and 5xx) and also a malformed/truncated 202 response; the FUSE paths set sticky `patchPlanRejected` and recursively full-upload. On current `tidbcloud/fs` main (`5f2eb89`), patch initiation has already created/copied the S3 MPU, reserved quota, superseded any active upload, and inserted pending file/upload rows before the handler emits 202 JSON. Thus a 202 decode failure is explicitly not “before any side effects.” My review-only test returned 202 with a truncated plan and current code immediately performed a full rewrite and returned OK in 5/5 runs. Under tight quota the replacement full-upload reserve can fail while the unseen patch reservation remains active, and the sticky flag prevents re-entering PATCH. Only explicitly classified, pre-side-effect capability/validation rejections may fall back; outcome-unknown/conflict/auth/rate/server errors must remain fail-closed. Add status-by-status and truncated-202 regressions.

4. **The growth PATCH predicate leaks into the WAL/append-log fallback path despite the design saying PATCH applies only to `main.db`.** The existing regression `TestAppendLogLegacyUploadPlanUnsupportedReroutesOnceToFullPUT` now fails because `/db-wal` unexpectedly receives PATCH instead of the legacy append-log/full-rewrite sequence. Gate W3 to the intended main-database route and keep the existing append-log test green.

Machine evidence: the three review-only races/failure cases above reproduced 10/10, 5/5 and 5/5 respectively; temporary hooks/tests were removed and the worktree is clean. PR W1/W2/W3 tests pass 3x and under `-race`, compile/vet/diff-check pass. Existing focused tests also fail for `TestFlushGrowthLazyBufferFetchesMissingParts`, `TestWriteSyncDisablesStreamingUploaderForLargeSequentialWrites`, and the append-log test above. Required Code CI is red at `pkg/fuse/w3_growth_patch_test.go:241` (`staticcheck` S1021).

qiffang (changes requested):
Additional review on exact head `e7ef4508994c4af29836324ecdc45ccdb076fcf6`; my earlier W2 REQUEST_CHANGES still applies, and I independently confirmed the additional blockers @dat9-dev2 surfaced.

1. W2 clean-sibling reads are still not fenced to the data source identity. `readSQLitePersistentJournalCleanSiblingRange` decides eligibility under `src` lock (`dat9fs.go:4768-4770`) but then releases the lock and reads the path-global active shadow via `shadowStore.ReadAt(path, ...)` (`dat9fs.go:4776-4785`). It also captures `latestCommittedRevision(path)` once in the caller (`dat9fs.go:4739-4741`) and does no final latest-revision recheck before returning old clean sibling bytes. A same-path writer can swap/mutate the active shadow generation or another commit can advance the watermark in that window, while this reader still returns bytes as if they were the latest committed revision. This needs a generation/identity fence plus final revision validation, and a deterministic same-path concurrent-write/commit regression.

2. Strict growth `Fsync` races an already in-flight background full PUT. The new reroute only calls `RemoveIfGeneration` and clears `WriteBackGen/WriteBackSeq` before `flushHandle` (`dat9fs.go:14155-14171`). That does not wait for a background uploader already past `acquirePath` and `GetMetaAndViewWithCallback` (`writeback_uploader.go:355-420`). The older synchronous path explicitly waits with `WaitPath` (`writeback_uploader.go:486-489`); this new PATCH reroute skips that serialization, so foreground PATCH and background full PUT can race with the same expected revision. The existing W3 test seeds cache but never calls `Submit`, so it cannot cover this state.

3. `ErrPatchPlanInitiate` is broader than the server contract allows. The client marks every non-202 response and every 202 JSON decode failure as safe for full-rewrite fallback (`pkg/client/patch.go:16-19`, `100-116`), and `flushHandle` / write-sync then sticky-reroute to full upload (`dat9fs.go:15603-15612`, `13022-13047`). On authoritative tidbcloud/fs main `5f2eb89`, `handlePatch` writes `202` only after `InitiatePatchUploadIfRevision` returns (`server.go:3251`, `3299-3304`), and the backend has already created an MPU, generated presigned URLs, copied retained parts, reserved quota, superseded existing active uploads, and inserted pending file/upload rows before returning the plan (`backend/patch.go:224-229`, `307-324`, `336-364`, `367-411`). A truncated/undecodable 202 is outcome-unknown, not no-side-effect. 409/auth/429/5xx also should not become a sticky “never PATCH this handle again” decision.

4. Growth PATCH is taking the append-log/WAL legacy fallback path. `flushHandle` tries append-log routing first, but if `routeAppendLogLocked` returns not-handled for a configured existing single-layout append-log file, generic upload selection still allows `usePatch` (`dat9fs.go:15076`, `15395-15401`; `append_log.go:1023-1061`). The repo’s existing `TestAppendLogLegacyUploadPlanUnsupportedReroutesOnceToFullPUT` now fails on this exact head with `unexpected request PATCH /v1/fs/db-wal`, which means the design’s legacy full-rewrite path for append-log/WAL is bypassed by generic growth PATCH.

Required Code CI is also red on this head (`make lint`, staticcheck S1021 at `pkg/fuse/w3_growth_patch_test.go:241`).

Local verification for this addendum: current checkout is clean at the exact head; `go test ./pkg/fuse -run TestAppendLogLegacyUploadPlanUnsupportedReroutesOnceToFullPUT -count=1` fails with the unexpected `PATCH /v1/fs/db-wal`; server-side PATCH initiate evidence was checked against tidbcloud/fs `origin/main` at `5f2eb89efd992b606abe4c2d6b5c5c9833cca70a`.

qiffang (commented):
本轮复审固定在 `6bf4134a8ec1cac8b136213e602643a9772dbf7d`，对照上一轮评论与 `e7ef450` 之后的修复，并沿实际调用链核对。以下为协审汇总结论；本次提交 COMMENT，不是合并批准。

### 已核实的修复
- W2 不再从路径级 shadow 读取缺少快照身份约束的数据；保留的 loaded/dirty buffer 路径在句柄锁内复制，复制后再次校验最新 revision。不同 revision 的回归覆盖了结果作废条件。
- 不安全的 strict-fsync 分流已经撤回；不再用前轮的错误分流作为当前 head 的阻塞依据。
- 宽泛的 PATCH plan-error 回退及其持久标志已经撤回，`client/patch.go` 与本 PR merge-base 一致；不能再把任意冲突、鉴权或服务错误笼统当成 full-upload 回退条件。
- growth 路由排除了 SQLite 持久 journal，旧 append-log 回归通过。更新后的普通文件 PATCH 用例也检查保留前缀、新尾部和未修改分片，而不只是检查调用次数。

### 验证与边界
独立执行的 W1/W2/W3 及相关既有回归竞态测试、AppendLog/SQLite 相关测试和 `go vet ./pkg/fuse` 通过；当前 head 的 Code CI、Local E2E 也已核实通过。

本轮没有足够证据提交新的阻塞项，但以下不能算作已验证：
1. W1 的“截断已被句柄消费、提交时句柄忙、之后先 Read 再 Write”延迟重绑交错尚无有效动态结论。初版探针没有建立已提交 revision 的前置状态，不能把其失败当成产品缺陷，也不能当作该边界通过。
2. 放大上传/租户限制后，增长跨越 PATCH 分片上限的边界尚未完成动态验证；默认 10 GiB 限制下不能用该候选声称线上回归。full upload 按最终大小调整分片，与既有 PATCH 布局的限制不能直接等同。
3. PR 描述中的旧版本 EC2/性能数据不属于本次 `6bf4134`；本轮不据此宣称当前 head 已完成现场或性能验收。

因此，本次关闭上述已核实的旧问题，不新增未经证明的 inline finding；也不将 CI 绿色或这份限定范围 COMMENT 解释为所有并发边界均已验证。

pkg/fuse/w1_truncate_rebind_test.go
```diff
@@ -0,0 +1,390 @@
...
+		sqliteWALConfirmed:       true,
+		sqliteWALCommittedHeader: oldHeader,
+	}
+	// The sibling zero-truncate event is applied on the writer.
+	writer.appendLogRecordTruncate()
+	fs.openHandles.Add(writer)
+	defer fs.openHandles.Remove(writer)
+
+	// The checkpointer's truncate commit publishes revision 6, size 0.
+	fs.refreshCommittedRevisionForOpenHandlesWithSize(writer.Path, 6, checkpointer, 0)
```
qiffang, line 353: [P2] 补齐提交刷新跳过忙句柄后的 Read→Write 回归

这里同步调用 refresh 且 writer 可立即加锁，只验证了新增加的直接 adopt 分支。尚未覆盖：writer 已消费 sibling 的 zero-truncate，随后在 checkpoint commit 时持锁导致 refresh 的 TryLock 跳过；提交完成后 writer 先 Read，再写新 WAL header 并 fsync。现有 BusySibling 测试是在 truncate 发布之前持锁，事件尚未消费，是不同路径。

这不是任意扩展的并发排列：`refreshCleanCommittedRevisionForHandleLocked` 会先更新 BaseRev，却不更新 append-log rewrite baseline；之后 `adoptCommittedRevisionLocked` 的 `advanced` 已为 false，可能不再进入 baseline rebind，而 `appendLogCommittedBaseline` 仍优先选择 rewriteBase。当前用例无法检出这一状态分离。

请增加有前置断言/同步的真实 Read、Write、fsync 回归：确认 committed revision 已发布、truncate 已消费、直接 refresh 确实跳过；再断言 header 提交使用新 baseline、服务端内容正确，并以不先 Read 的顺序作对照。只有初版缺失 committed-revision 前置条件的失败探针，不足以接受或否定这条路径。

qiffang (changes requested):
补充明确的质量门结论：REQUEST_CHANGES，暂不批准合并。前一份 COMMENT 中认可的旧问题修复和测试结果保持不变。

阻塞范围仅为下方 W1 延迟重绑的有效回归缺口，不包含未经动态验证的 W3 极端上传限制，也不将旧版本 EC2 性能结果作为代码阻塞。这里没有声称初版失败探针证明了数据错误：该探针前置状态不成立，已明确作废。请补齐能实际进入所述路径的回归；若证明当前实现已正确处理即可关闭，若暴露问题则一并修复。

这是 adversary-2 的独立质量门判断；协审方对是否可直接批准有不同意见，不将其表述成一致批准。

pkg/fuse/w1_truncate_rebind_test.go
qiffang: 补充新证据（仍固定 `6bf4134`）：开发方已修正探针，实际发布 `(revision=6,size=0)` 到 committed-revision 状态；我核对后使用同一份探针独立复跑竞态测试 3 次，结果一致：

- 不先 Read：真实 Write 后进入生产 append-log 提交路由，成功提交，最终 `BaseRev=7`、rewrite baseline 清除。
- 先 Read：读后 `BaseRev=6`，但 `rewriteBaseRevision=5` 且 truncate 标记仍保留；随后真实 Write 再进入相同提交路由，返回 EIO，没有成功 PUT，状态仍为 `BaseRev=6/rewriteBase=5`。

因此该项不再只有覆盖不足的推测，已确认此状态分离能使新 header 的提交路由失败；请修复并保留回归。P2/REQUEST_CHANGES 不变。原先缺少 committed-revision 前置条件的探针继续作废。

证据范围：受控句柄/提交状态 + 真实 Read、Write 和 `routeAppendLogLocked` + HTTP 替身；末尾调用的是生产提交路由，不是完整 Fsync 方法，也不是云端或真实挂载测试。未声称已复现数据丢失。

qiffang (approved):
**Review round 3 — exact head `e2f0ffe0`. Verdict: APPROVED.**

The single blocker left open by round 2 (head `6bf4134a`) — the missing valid regression for the W1 deferred-rebind / generation-reset interleaving — is now genuinely closed, and the other round-2 items are verified fixed against production code. Built, vetted, and ran the relevant suites locally.

### W1 deferred re-bind regression gap → CLOSED (`e44c2cf`, `e2f0ffe`)
The prior probe was voided because it never established a committed-revision precondition (opened an existing WAL, H0 unproven → header write took tail-append instead of generation reset). The new `TestW1TruncateCommitThenWriterHeaderTakesGenerationReset` (`pkg/fuse/w1_truncate_rebind_test.go`) fixes exactly that: it sets `sqliteWALConfirmed`+`sqliteWALCommittedHeader` at rev 5 (proven H0), applies a real sibling zero-truncate, drives a genuine committed-revision transition (rev 6) through `appendLogAdoptCommittedBaseline`, and asserts the new-salt 32-byte header write routes through the **generation reset** (PUT with `X-Dat9-Expected-Revision: 6`, body == new header, `appendCalls == 0`) — the semantic route, not a call-count delta. Reversion-verified: forcing the generation-reset gate to NotApplicable makes it fall through to tail-append and fail with "unexpected tail append request".

The exact "truncate consumed → handle busy at commit → deferred → Read-before-Write" ordering is additionally covered by the pre-existing `TestSQLiteZeroTruncateWaiting…` / `tryAppendLogGenerationResetLocked` tests (SetAttr truncate holds the path fence while an Fsync waits, asserting no stale payload upload). Structurally, a truncated handle carries `appendSafe=false`, so tail-append is unreachable — the only reachable path is the CAS-guarded rewrite/generation reset, which fails closed (EEXIST/conflict) on a revision mismatch rather than appending to a rotated WAL. No stale-generation path reaches wrong remote state.

### Other round-2 items verified
1. **Strict-fsync PATCH reroute removed** — grown large files fall through `flushHandleDebounced` to PATCH; sub-threshold growth keeps the race-free `UploadSyncWithRevision` (WaitPath). Removes the nondeterministic CAS race.
2. **Growth PATCH kept off SQLite sidecars** — gated `size <= OrigSize || !isSQLitePersistentJournalPath`; append-log stays the WAL transport, legacy full-PUT chain preserved.
3. **Over-broad `ErrPatchPlanInitiate` fallback removed** — the sentinel and `IsPatchPlanInitiateErr` are deleted (no remaining refs anywhere); plan-initiate 4xx/5xx now return the raw error and fail closed. The only full-upload fallback trigger is `client.IsUnsupportedStorageTargetErr` (2 sites, dat9fs.go:13000/15553), which sets `StorageClass=db9` and bounds the retry at depth 1 so it cannot re-enter the patch branch. Same-path serialization + revision recording (B3/B10 fences) preserved.
4. **W2** — non-atomic path-keyed shadow leg dropped; latest committed revision revalidated after copying a clean sibling, with a deterministic regression.

### Verification
- `go build ./pkg/fuse/ ./pkg/client/` clean; `go vet ./pkg/fuse/` clean.
- `go test ./pkg/fuse/ -run 'W1|W2|W3|SQLiteZeroTruncate|GenerationReset|TruncateRebind'` → ok.
- W1 regression reversion-confirmed to catch broken routing.
- Code CI / Local E2E / CodeRabbit all green on `e2f0ffe0`.

No remaining blocker on this head. LGTM.

PR #906, opened 2026-09-07
Revert #901 (grown same-path commit rebase) to unblock WAL

mornyx:
## Summary

Revert squash merge #901 (`1a1d8737`) from `main`.

#900 and #901 each pass a 10MiB WAL single-transaction insert on fsync + `--append-log`. Combined `main` (`1a1d8737`) fails with `database disk image is malformed` (~123s, ~11MiB db+WAL). Pre-#900/#901 (`1bc5c2c8`) and #900-only (`fdadeea4`) are OK at 10MiB and 100MiB.

This restores #900-only so WAL/append-log is usable while the interaction is investigated. It **reopens #896** (interactive 1GiB DELETE-journal `kvtest init` leaving a 4KiB empty DB).

## Test plan

- [x] Local revert is the inverse of #901 (commit_queue / pending_index / fence tests / transfer_test flake fix)
- [ ] EC2: fsync + append-log, 1GiB WAL `INSERT` (this PR binary)
- Do not merge until 1GiB WAL result is posted

Issue #903, opened 2026-09-07
FUSE CAS: revision-only overwrite can replace a same-path delete/recreate (resource generation ABA)


mornyx:
## Bug Report

Revision-only compare-and-set (`X-Dat9-Expected-Revision`) does not bind a write to a file generation. After delete/recreate, a new object at the same path often starts at revision 1 again. A pending overwrite that still holds `expected=1` can land on the new incarnation.

This is not #896 (same-path SQLite header then growth leaving a 4KiB empty DB). #896 is about commit-queue stale payloadBaseRev vs a later generation of the **same** object. This issue tracks the remaining inode/generation ABA called out on #901.

### 1. Minimal reproduce step (Required)

Two sequences (review-only on #901 head `f8b52513`; both need a test server that can recreate the path with a new `X-Dat9-Resource-ID` while keeping rev=1).

**A. Same bytes, new ResourceID, after restart**

1. Interactive FUSE commit queue lands a 4KiB header at rev=1; persist a larger pending payload for the same path.
2. Crash / new CommitQueue + RecoverPending.
3. Another writer deletes and recreates the path as a **new** ResourceID, **identical** 4KiB bytes, revision 1.
4. Recovery rebases on checksum / expected=1 and PUTs the grown body as rev=2.

**B. Proof GET then PUT (ShadowSpill TOCTOU)**

1. Same-process growth: Stat/GET observes the old rev=1 object and treats it as the landed header.
2. Before the ShadowSpill/direct PUT, another writer recreates the path (new ResourceID, same revision number).
3. PUT with only `X-Dat9-Expected-Revision: 1` succeeds and overwrites the replacement.

Client-side preflight (HEAD/GET + checksum) cannot close B: the check is not in the same atomic write as the mutation.

### 2. What did you expect to see? (Required)

- Recreate is a new file generation. Stale pending growth must 409 / stay PendingConflict.
- Replacement content and ResourceID unchanged; no rev=2 PUT of the old payload.
- CAS should match revision **and** generation (for example `X-Dat9-Expected-Resource-ID` checked inside WriteIfRevision / multipart complete), not revision alone.

### 3. What did you see instead (Required)

- A: checksum of the 4KiB header matches the new object, so recovery uploads the grown payload as rev=2.
- B: GET-based proof then revision-only PUT wins the ABA window.

[...]

PR #911, opened 2026-09-08
fix(fuse): prove lineage before rebasing grown creates

### LINKED ISSUE #896: FUSE interactive: 1GiB SQLite kvtest init acknowledges success then persists a 4KiB empty DB (revision conflict)

qiffang:
## Summary

Fixes #896 and replaces the size-based growth heuristic from #901 (reverted by #906).

Interactive SQLite can stage a 4 KiB database header as a base-zero create, then stage the same file's much larger full image while the header is being committed. Once the header lands as rev1, the `#876` stale-payload fence correctly rejects an arbitrary base-zero payload, but it also rejected this legitimate direct child and could leave only the empty header remotely.

PR #901 used sequence/size ordering to rebase the larger payload. After #900, that is not a sufficient causal proof: a stale larger SQLite checkpoint or an older, larger WAL generation can satisfy the same size test and overwrite newer bytes.

## Approach

- Give each live full-file staging generation a random process-local `SnapshotID` and record the exact `ParentSnapshotID` from which it was derived.
- Rebase only a trusted, non-recovered `PendingNew(baseRev=0)` whose parent is exactly the last snapshot this queue landed, whose size strictly grew, and whose remote revision/size/content still match that landed parent.
- Make growth authorization one-shot and retain the original `PayloadBaseRev` for the existing stale-payload audit/fence.
- Coalesce an unlanded queued parent only when the child's parent ID names that exact queue entry; reparenting is guarded by the pending-index generation.
- Never persist causal lineage. Shadow bytes and JSON metadata are not atomically replaced across a crash, so recovered entries deliberately lose lineage and fail closed as `PendingConflict`.
- Fail staging when the pending index cannot publish the metadata corresponding to the staged shadow, allowing the existing synchronous fallback instead of publishing unverifiable lineage.
- Preserve #900's WAL evicted-part restoration, #905's unlink-generation guards, and #908's durable rename behavior. Rename now binds the exact metadata returned by `PrepareRename` rather than rereading a potentially newer path generation.

No `.landed` file or recovery-side shadow is added, so this fix adds no persistent cleanup/storage surface.

## Correctness coverage

The regressions cover:

- positive 4 KiB-header-to-grown-image direct-child rebase;
- upload completion after a newer same-path shadow has replaced the path-local generation;
- stale larger SQLite checkpoint rejection;
- old larger WAL generation rejection after a reset/header generation;
- sibling/fork rejection;
- one-shot validation;
- exact queued-parent coalescing only;
- restart/recovery dropping process-local lineage;
- recovered growth failing closed;
- delete/recreate with the same revision and size but different content failing closed;
- path-wide shadow preload dropping lineage trust;
- rename binding the exact prepared generation and lineage.

Validation:

```text
go test -race ./pkg/fuse -run '^TestCommitQueue' -count=1
ok github.com/mem9-ai/drive9/pkg/fuse 16.326s

targeted lineage/#900/#905/#908 regressions under -race
ok github.com/mem9-ai/drive9/pkg/fuse 2.208s

go vet ./pkg/fuse
git diff --check
```

The complete `pkg/fuse` suite still has three environment/baseline failures that reproduce unchanged on clean `origin/main@788ec769`: the two wait-mount permission-probe tests, plus the Python SQLite checkpoint test failing to open its temporary database.

## Real FUSE smoke tests

Built Linux/amd64 from this commit and tested on EC2 against `https://aws-us-east-1.drive9.ai` with debug and standard FUSE perf capture enabled.

1. Interactive, SQLite DELETE journal, one 10 MiB transaction:
   - final main DB: 11,804,672 bytes;
   - commit: 384.98 ms;
   - 2560/2560 rows and `integrity_check=ok`;
   - mounted final image and post-unmount authoritative object had identical SHA-256.
2. `durability=fsync`, SQLite WAL, `synchronous=FULL`, WAL/journal append-log enabled:
   - main DB: 11,804,672 bytes; WAL before close: 11,877,992 bytes;
   - commit: 1749.30 ms;
   - 2560/2560 rows, `quick_check=ok`, `integrity_check=ok`;
   - before-close, after-close, and post-unmount authoritative main SHA-256 were identical;
   - no `malformed`, disk I/O error, terminal conflict, or row loss.

The low-latency live cases serialized the header before the grown entry was enqueued, so the new rebase branch is exercised deterministically by the commit-queue race tests rather than claimed from those smoke logs.

## Deliberate boundaries

- This closes #896's observed single 4 KiB parent -> 1 GiB child transition. A multi-hop growth chain whose immediately landed parent exceeds the bounded identity-check size fails closed rather than weakening CAS; it is not claimed as generic sustained-growth support here.
- Resource-generation ABA between client preflight and revision-only PUT/multipart complete remains #903. Closing that needs an expected resource-generation value in the server-side atomic mutation contract.
- A process restart discards live lineage and preserves the payload for manual recovery instead of guessing ancestry.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved reliability when multiple edits are made rapidly to the same file.
  - Reduced incorrect rejection of valid, newer changes as stale.
  - Improved handling of file growth, queued updates, renames, and interrupted recovery.
  - Strengthened safeguards against overwriting newer remote content when versions diverge.
  - Preserved update relationships across staging, shadow storage, and reopened handles.
  - Improved consistency when transferring staged changes between file paths.
  - Improved cancellation handling during conflict checks and remote reads.
  - Improved recovery when staged content changes during upload.
  - Improved handling of batch conflicts and large staged payloads.


qiffang:
CI follow-up on head 6e9a69e1. The original hosted failure was the known synchronization race in TestReadStreamRangeFirstHopBoundedTransfer: the client had returned, but the handler had not yet published its byte count. The exact synchronization fix previously carried by #901 was restored in ba0a6bba. Follow-up hardening fixes stage-fallback generation capture, caller-cancellation propagation, bounded and invalidation-safe landed lineage metadata, and independent retry contexts for failed batch items. Local validation passed: focused lineage and batch tests x20, race x5, the full CommitQueue test group, go vet for pkg/fuse and pkg/client, and make lint. [Code CI](https://github.com/mem9-ai/drive9/actions/runs/34244900068) passed in 7m11s. [Local E2E](https://github.com/mem9-ai/drive9/actions/runs/34244900066) passed in 17m06s, including FUSE crash recovery, supervision, and write-perf budget. Issue #903 resource-generation ABA remains deliberately out of scope: it requires server-side atomic ResourceID plus revision CAS and a distinct non-LWW generation-conflict outcome.

mornyx (changes requested):
## Review summary

This replaces the reverted size/seq growth heuristic with process-local SnapshotID/ParentSnapshotID, one-shot rebase authorization, exact queued-parent coalescing, and fail-closed recovery. The causal model is the right fix for #896, and the negative matrix (fork, stale checkpoint, older WAL generation, delete/recreate, recovered growth) is aimed at the real sibling-mutation cases. Under a strict maintainability bar the new lineage machinery still lands as another subsystem inside already-huge files, but that is not what blocks merge.

Prior bot notes on cancellation-vs-stale, bounded landed landmarks, staged-fallback generation capture, test defer order, and independent batch-retry contexts look fixed on current HEAD. Issue #903 resource-generation ABA remains out of scope, as stated. The merge blocker is a sibling-path miss in the layer upload that this PR itself taught to authorize a rebase.

**What looks good**
- Exact parent snapshot + remote identity, instead of bringing back the size heuristic that #906 reverted; PayloadBaseRev stays on the entry for the #876 audit fence.
- Recovery and restart deliberately drop lineage and fail closed; no persistent `.landed` surface.
- Landmark lifecycle is thought through: reference-aware prune, invalidation on unverifiable/revision-reset successors, one-shot `growthRebaseRev`.
- Cancellation is kept orthogonal to `PendingConflict` on the validate-path proof (`ctx.Err()` rather than stale).

---

**1. Layer upload sends the pre-rebase revision**
`uploadEntry` snapshots `entry.BaseRev` and passes it by value into `uploadLayerEntry` *before* that helper runs `validateEntryPayloadFreshCtx`, which is the call that authorizes the #896 rebase and rewrites `entry.BaseRev` (0 → landed parent rev) plus `DisableAutoResolveLWW`. Both layer send sites then use the stale copy (`UploadFSLayerFile(..., expectedRevision)` and `FSLayerEntryRequest.BaseRevision`).

On a layer/gVisor mount the headline case (4 KiB header landed at rev1, base-zero grown child) therefore issues CAS at revision 0 against a server at rev1, 409s, and because authorization already disabled LWW auto-resolve, `commitOne` routes it to terminal `PendingConflict` — the symptom this PR claims to close. The non-layer branch is correctly ordered (validate at 2494, then read `entry.BaseRev` at 2499), which is why non-layer FUSE smoke tests would not catch this.

Fix: validate before reading the revision in both branches (hoist `validateEntryPayloadFreshCtx` above the `layerRef` early-return, or re-read `entry.BaseRev` after validation inside `uploadLayerEntry`). Add one layer-mode regression: landed parent + grown base-zero child must send the authorized revision.

```go
if layerRef != "" {
    return cq.uploadLayerEntry(ctx, layerRef, entry, apiPath)
}

func (cq *CommitQueue) uploadLayerEntry(...) (int64, error) {
    if err := cq.validateEntryPayloadFreshCtx(ctx, entry); err != nil {
        return 0, err
    }
    expectedRevision := entry.BaseRev
    // ...
}
```

---

**2. Production Create is ShadowSpill; overlapping landed-proof test is not**
Async Create sets `ShadowSpill`. Spill upload streams via `uploadFromShadowRemoteWithRevisionAndGeneration` and never `bindPayload`, even though the small-file branch already reads those bytes into a local slice and drops them. `checksumLandedPayload` then only recovers identity from `payloadBound` or a still-active `ReadAllIfGeneration`. After a newer same-path `WriteFull`, generation no longer matches, checksum is empty, `rememberLanded` deletes/skips the landmark, and a legitimate grown child fails closed as stale.

`TestCommitQueueLandedProofUsesUploadedPayloadAfterShadowReplacement` is the contract for that race, but it uses a non-spill entry so `payloadBound` saves it. FUSE `WaitPath` usually serializes Flush, which is why live smoke tests often never hit this; it is still the production Create flag on the path this PR exists to protect. Capture checksum/bind on spill upload for size ≤ 1 MiB (do not persist `.landed` files), and add a ShadowSpill clone of that overlapping test.

**3. Landed identity still advertises a dead ResourceID channel**
The only production `rememberLanded` call passes `resourceID=""`. `pathCommitLandmark.resourceID` and the ResourceID branch of `landedIdentityMatches` are unreachable; every live proof is SHA-256 of a ≤1 MiB parent, and larger parents self-invalidate (the documented fail-closed bound — do not weaken CAS to “fix” that). Delete the unused field/parameter until something can actually populate it, or wire a real PUT/Stat identity and test it. As written it overstates proof strength.

**4. ShadowSpill conflict-path growth is a second, weaker copy of the proof**
`tryResolveShadowSpillIdempotent` does the lineage-proof read with `context.WithTimeout(context.Background(), 10s)` and swallows the error, then has its own upload+finalize sequence. That reintroduces the caller-cancellation gap this PR fixed on `readRemoteSnapshot`, and there is no positive end-to-end test that a 409 on a spill create actually commits via this branch (the happy rebase tests go through `uploadEntry`). Reuse `readRemoteSnapshot(entryCtx, ...)` and extract one grown-payload finalize helper; add a positive spill-409 → rebase commit test.

**5. Extract the lineage unit rather than growing the two oversized files again**
`commit_queue.go` 2842→3302 and `writeback_generation_fence_test.go` 1568→2544 were already well past a healthy size; this is not a new crossing of ~1k, so it is not a merge block. The added landmark/coalesce/rebase block is a cohesive ~300–460 line unit with two touchpoints (`validateEntryPayloadFreshCtx`, `onCommitSuccessWithOptions`). Move it to `commit_queue_lineage.go` and the new regressions to `commit_queue_lineage_test.go` (or equivalent) now, while the boundary is still clean. A small `snapshotLineage` value would also collapse the 9-argument AndLineage plumbing.

---

### Nits
- Two new tests (`TestStageShadowPendingFailureKeepsCurrentGenerationForSyncFallback`, `TestOpenHandlePathWideShadowPreloadDropsLineageTrust`) use quota-enforcing `NewShadowStore(t.TempDir())`; sibling lineage tests use `NewShadowStoreWithQuota(..., 0, 0)` and do not depend on host free-ratio.
- `coalesceDirectParentQueuedLocked` scans the whole queue under `mu` per parented Enqueue; filter via `queuedByPath` first.
- `TestReadStreamRangeFirstHopBoundedTransfer` `defer close(wroteDone)` runs on every handler invocation, including 404; a second request panics. Guard with `sync.Once`.
- Pin the `HasMode` coalesce-refusal and a rename-commit-then-grow rebase at the new path; both are only covered at the binding layer today.

---

### Recommendation
**Request changes.**
Layer mounts authorize the #896 rebase and then send the pre-authorization revision, so legitimate grown children still land in terminal `PendingConflict` on that path.

qiffang:
已根据本轮 review 推送修复：`d88af4e`，另以 `9b4e1ea` 补空 parent 回归。

### 修改
- layer 在 freshness validation **之后**读取 BaseRev，inline/stream 两个发送点使用同一份当前 claim。需要澄清：overlay upsert 不会自动产生 main rev1；保留其 rev0/无 main landmark 契约，测试分别覆盖 overlay parent 与已有 main parent 证明的发送边界。
- ≤1MiB 的 ShadowSpill 按 generation 和实际尺寸绑定私有 payload，上传与 landed checksum 同源；更大文件继续流式，不新增持久化证明。覆盖真实 upload 返回→shadow 替换→success 窗口，以及小文件 multipart。
- 冲突 stat、growth read、等长 read 均继承调用方 context；buffered/spill 共享增长重传和等长成功收尾，取消后不发布、不清理保留的 staging。
- 移除未被生产调用填充的 ResourceID 证明通道；landed proof 仍限于小 parent 的 SHA-256，不扩展 #903 的 ABA 范围。
- 拆出 lineage 实现/回归，按同路径索引筛选 coalescing；补 HasMode 拒绝、rename→commit→growth、实际尺寸上限、空 parent 等用例，并修正 quota fixture 与 first-hop 测试的通知关闭方式。

### 验证
- CommitQueue 全组 race 连续三轮、相关静态检查及 lint 通过。分别恢复旧 CAS 捕获、旧 Spill 上传、Background 读取时，新回归均能识别故障；mutation 已全部还原。
- **当前 head `9b4e1ea` 在 EC2 Linux 的 FUSE 全包测试及定向三轮均通过**。本地 macOS 全包仅有两项既有 mount-probe 平台差异；HTTP first-hop 回归 race 十轮通过。常规 client DB 测试受本机容器条件限制，未宣称通过。
- 真实 FUSE 挂载、冷读和 API 哈希/SQLite 场景已准备，仍等待授权测试 tenant 凭证，**尚未执行，不计为通过**。新 head 的 GitHub CI 正在运行。

未合并 PR。

qiffang:
已补推 `f4bacb7390e820dc8d1f474011c9e09614fe2494`，处理本轮最后两处建议：补齐 lineage 测试的队列清理（包括提前失败路径），并在一次加锁淘汰中复用引用集合，新增 queue/in-flight/immediate 三类引用的多次淘汰保护回归。

### 最终验证
- 当前 head 的 Code CI（普通测试、failpoint、lint、build）与 Local E2E 均通过；CommitQueue 全组竞态回归及 EC2 Linux 全 Fuse 单测通过。
- 使用该 head 构建的 CLI，在授权 EC2/现有服务上运行普通 Linux FUSE 与 `--gvisor-compat` 参数两组；每组覆盖 4KiB→12KiB/8MiB/64MiB 增长、rename→增长、8 个不同文件并发写及 SQLite WAL 64 行/完整性。两组共 26 项 API 全量大小/SHA256 校验通过。
- 重验时明确给每次挂载独立 `--cache-dir` 和 `--local-root`，启动前断言目录不存在/为空；两组各 13 个文件的冷读通过。此前仅替换 local-root 的结果不作为冷缓存证据。
- 独立 layer 用例覆盖三种大小增长、空缓存 layer 冷读、提交到 main 后的 3 项 API 大小/SHA256 校验，均通过。

### 仍保留的边界
- 扩展测试并非全绿：匿名 ShadowSpill 句柄 `write(before) → unlink → append(-after) → fsync → read` 仍只读到 `before`。PR merge-base `b5d35e9`、修复前 `6e9a69e` 和修复后版本均复现，已另行跟进；本 PR 不将此旧问题伪装成通过。
- 真实增长测试属于功能 smoke，不宣称必然命中 direct-parent landmark 竞争窗口；该窗口由明确同步的回归及反向变更验证覆盖。`--gvisor-compat` 参数测试也不等同于真实 gVisor runtime 测试。

测试只使用隔离路径，未修改服务部署；自建测试目录已清理。未合并 PR。

qiffang:
补充新审查后的修复：`e75c92b286e107215a8b3b77080b31099185a69a` 已推送。

- 修复 `directParentPendingNew` 过早读取上传线程所拥有字段的 data race：coalescer 先排除 nil/canceled/in-flight，再执行 lineage/payload 判断。
- 新增真实 production coalescer 与 `bindPayload` 并发的回归，覆盖 `queuedByPath` 与 fallback queue 两入口。同一回归在隔离的 f4 旧实现明确检出 `commit_queue.go:124` 写 / `commit_queue_lineage.go:49` 读；修复后重复竞态测试、CommitQueue 全组及静态检查通过，独立 reviewer 已复核。
- e75 精确版本的 EC2 Linux 全 Fuse 单测已通过；当前 head 的 Code CI 与 Local E2E 均已通过。此前完整 EC2 API/冷读/SQLite/layer 现场矩阵属于 f4，不改写成 e75 的现场结果。

**取消保证范围修正**：目前解决的是读取过程中与收尾入口已观察到的取消传播/保留；`CancelPathPreserveLocal` 与 success/terminal 收尾之间仍没有原子决胜边界，历史残余另行跟进，不声称完整线性化 cancel-preserve 已实现。rename 调用方还有 WaitPath/远端重探，未将该窗口描述为已证实的线上 rename 丢失。匿名 ShadowSpill 旧读也继续独立保留。未合并。

mornyx (approved):
## Review summary

Round-2 pass on `e75c92b2`. The previous merge blocker is gone: layer uploads now take the CAS revision only after freshness validation, on both the object and inline send sites, with overlay-vs-main-parent coverage. The rest of the last review also landed structurally rather than as drive-by patches — small ShadowSpill parents bind a private payload shared by upload and checksum, the unused ResourceID channel is deleted, conflict-path reads inherit the caller context, and lineage/spill live in dedicated files. Under a strict maintainability bar this is in good shape to merge.

- Layer `uploadLayerEntry` validates, then reads `entry.BaseRev`; tests pin overlay (no main landmark) vs existing main parent.
- ≤1 MiB ShadowSpill `bindPayload` with generation and actual-size checks; overlapping replacement-before-success, small multipart, and oversize-metadata cases.
- ResourceID proof plumbing removed; identity is checksum-only for small parents.
- Spill conflict stat/growth/equal reads use `entryCtx`; cancel is not turned into `PendingConflict`.
- Lineage extracted (`commit_queue_lineage.go`, `commit_queue_spill.go`, dedicated tests). `commit_queue.go` is 2935 (was 3302); fence tests 1659 (was 2544).
- Coalescer skips in-flight entries before reading worker-owned fields (`e75c92b2`), with a race regression.
- Prior nits (quota fixtures, `wroteDone` once, HasMode coalesce, rename-then-grow, prune referenced set, `DrainAll`) look addressed.

**What looks good**
- `commitGrownPayload` / `finishResolveFailure` / `finishIdempotentCommit` collapse duplicated upload/finalize/cancel tails so cancel-vs-conflict is one decision.
- New tests target the windows that were actually wrong last round, not only the happy rebase.
- Documented fail-closed boundaries (#903, overlay not inventing main rev1, no persisted `.landed`) are still honored.

---

### Nits
- A transport error from `readRemoteSnapshot` currently returns false and, unless the context is canceled, becomes terminal `PendingConflict`. Distinguishing “identity unreadable” (retryable) from “identity mismatch” (fail closed) would avoid turning a blip into manual recovery; similarly, type the deterministic spill size-mismatch as `errCommitPayloadStale` so it does not burn the full retry backoff. Optional, and consistent with this PR’s fail-closed stance if left as-is.

---

### Recommendation
**Approve.**
Prior blocking and should-fix items are verified on current HEAD; only optional availability polish remains.

PR #916, opened 2026-09-10
feat(cli): opt-in privacy-preserving command telemetry

mornyx:
> **Aligned with [tidbcloud/ti-cli](https://github.com/tidbcloud/ti-cli).** This is not a new telemetry system: it implements the same privacy contract as ti-cli and posts to the same product-owned ingestion service ([`internal/telemetrybackend`](https://github.com/tidbcloud/ti-cli/tree/main/internal/telemetrybackend)) and TiDB cluster, so both CLIs land in one `telemetry_events` table. The client mirrors ti-cli's [`internal/telemetry/telemetry.go`](https://github.com/tidbcloud/ti-cli/blob/main/internal/telemetry/telemetry.go) — same enablement precedence shape (environment override → persisted preference → default, with drive9 choosing the stricter `off` default), same `~/.<product>/` identity file, same schema v2 payload, same best-effort delivery rules — and this repo runs no second backend. The only backend work required is the small allowlist widening for `drive9` listed under [Backend prerequisite](#backend-prerequisite-needed-before-events-can-be-stored-not-before-merging).

## Summary

Adds **opt-in, privacy-preserving command telemetry** to the `drive9` CLI. When a user opts in, each eligible command posts exactly one `drive9.command.finished` event to the telemetry ingestion service shared with `ti-cli` (same TiDB cluster, no second backend in this repo).

Telemetry is **disabled by default**: nothing is sent, and no telemetry state is read or written, unless the user explicitly enables it. No build type, install source, or CI heuristic can turn it on.

## How users opt in / out

| Priority | Switch | Meaning |
| --- | --- | --- |
| 1 | `DRIVE9_TELEMETRY=on\|true\|yes\|1` / `off\|false\|no\|0` | per-process override; empty value = "not set"; an invalid value fails closed |
| 2 | `"telemetry": {"enabled": true\|false}` in `~/.drive9/config` | persisted choice, next to the existing contexts |
| 3 | — | **disabled** |

```jsonc
// ~/.drive9/config
{ "current_context": "...", "contexts": { "...": {} }, "telemetry": { "enabled": true } }
```

The preference intentionally lives in the existing `~/.drive9/config` (not a new file), and stays a field on `cli.Config` so `saveConfig` round-trips it — otherwise any `ctx` mutation would silently drop the user's choice. `TestTelemetryPreferenceSurvivesConfigRewrites` guards that.

## Event payload

Allowlisted fields only:

- `command_path` — built from a hardcoded command tree (`drive9 fs layer create`), never from argv
- `flag_names` — names only, deduplicated, `^[a-z][a-z0-9-]{0,63}$`, max 64
- `exit_code`, `duration_ms`
- `cli_version`, `os`, `arch`, `install_source`
- `cloud_provider` / `region_code` — allowlisted values read from the command's own context; unknown values collapse to `unknown`, and a command addressing another context (`prod:/file`) reports neither
- `profile_source` — `default` / `explicit` / `env` / `unknown`
- `anonymous_installation_id` — `drive9_` + 22 base64url chars, generated once at `~/.drive9/.telemetry-installation-id` (mode `0600`)
- plus envelope fields `schema_version` (2), `sent_at`, `event_id`, `occurred_at`

**Never collected:** flag values, credentials or tokens, SQL text, file paths, file contents, command output, API payloads, context names, tenant IDs, hostnames. There is no free-text escape hatch: the earlier `DRIVE9_TELEMETRY_TAG` / `_EXTRA` channel was removed entirely, because a key deny-list could not make it safe.

### Flag names can never be values

Two independent defenses keep user data out of `flag_names`:

1. **A dash token that directly follows another dash token is treated as that flag's value and is not recorded.** Flag names come from argv without a per-command schema, and a value-taking flag consumes the next argv entry unconditionally, so this is the only sound reading. `--flag=value` forms are unaffected. The cost is under-reporting flags written back-to-back (`-l --json` records `l` only); the alternative — guessing arity — would let a value be recorded as a name.
2. **A closed allowlist of flag names the CLI actually defines** (`cmd/drive9/telemetry_flags.go`). A dash-prefixed token that is not one of them is user input — a mistyped or unknown flag, or a value the command would reject, e.g. `drive9 fs grep -my-pattern` — and is dropped. Dropping an unknown name can only lose analytics, never leak a value. `TestTelemetryKnownFlagNamesCoverEveryDefinedFlag` re-derives the set from source with `go/ast` (both `flag.FlagSet` definitions and hand-rolled parsers that compare argv against literals) and fails when a flag is missing from the allowlist or stale in it, so the list cannot silently rot.

Tests cover `--subject -secret-project`, `--description --private-note`, the inline forms, and unknown flag-shaped tokens.

## Exclusions

No event is emitted, and no telemetry state is touched, for:

- help / version in any form, commandless usage, unknown commands
- a bare `help` operand on the trees whose parsers treat it as help (`drive9 admin tenant create help`), while `drive9 fs grep help` stays a real operand
- every `drive9 update` form
- internal mount machinery: `mount supervise`, `--supervised`, `--supervise-foreground`
- every process drive9 spawns itself (background mount, supervisor and its workers, detached `git hydrate`, generated systemd units) — they get `DRIVE9_TELEMETRY=off` via `telemetryDisabledEnv`, and the generated unit carries `Environment=DRIVE9_TELEMETRY=off`

One user command therefore produces **at most one** event, and mount commands no longer report a second event with a mount-lifetime duration.

## Delivery contract

One `POST` per eligible command, `1s` total budget (client timeout and request context), `202 Accepted` is the only success status, no retries, no local queue, redirects refused, response body drained and discarded. Delivery runs on the exit path, so the budget is also the worst-case latency added to a command; a dropped event is preferable to a visibly slow CLI. Failures never change stdout, stderr, or the exit status.

Client-side validation mirrors the ingestion contract (`event_id`, `command_path`, `cli_version`, `flag_names`, `os`/`arch` allowlists) so an event that would be rejected as a whole is dropped locally instead of costing a guaranteed-400 request.

## Exit paths

`main()` now runs inside the telemetry wrapper, and command handlers exit through `cli.SetExitHook` / `exitProcess` instead of `os.Exit`. The three `flag.ExitOnError` flag sets became `flag.ContinueOnError` returning `cli.UsageError`, which keeps the previous behavior exactly (same message + usage printed once, same exit code 2) while making the exit observable — `drive9 mount --bogus-flag` and `drive9 mount vault` now report `exit_code: 2` instead of silently bypassing telemetry.

## Command tree drift guard

`cmd/drive9/telemetry.go` mirrors the CLI's dispatch tables by hand. `cmd/drive9/telemetry_drift_test.go` extracts the `case` labels from the real dispatchers with `go/ast` and fails when the tree does not know a command or subcommand. Verified by temporarily adding a fake `fs frobnicate` case, which the test caught.

## Build / release wiring

- `make build-cli-release` bakes `INSTALL_SOURCE=github-release` and the ingestion endpoint (`https://tdc-telemetry.tidbcloud.com/v1/telemetry/batch`); `TELEMETRY_ENDPOINT=` can override it (e.g. staging).
- Development builds (`make build`, `make build-cli`) carry **no** endpoint, so a dev binary can never post to production.
- Release artifacts ignore `DRIVE9_ALLOW_TEST_ENDPOINTS` / `DRIVE9_TEST_TELEMETRY_ENDPOINT`; those two exist only so source builds can exercise the delivery path against a local sink.

## Backend prerequisite (needed before events can be stored, not before merging)

The shared ingestion service currently accepts only `ti` / `tdc`, so drive9 events are rejected with `400` and dropped silently by the CLI (visible with `DRIVE9_TELEMETRY_DEBUG=1` as `telemetry delivery was rejected`). Four gates must be widened in `tidbcloud/ti-cli` — none of them requires a schema migration:

1. [`internal/telemetrybackend/server.go`](https://github.com/tidbcloud/ti-cli/blob/main/internal/telemetrybackend/server.go) — `validUserAgent`: accept the `drive9/` prefix (today only `ti/` and `tdc/` are allowed, so every drive9 request is rejected before any field is looked at).
2. [`internal/telemetrybackend/event.go`](https://github.com/tidbcloud/ti-cli/blob/main/internal/telemetrybackend/event.go) — `installationIDPattern`: `^(?:ti|tdc|drive9)_[A-Za-z0-9_-]{16,96}$`.
3. Same file — event-name check: allow `drive9.command.finished` alongside `ti.command.finished` / `tdc.command.finished`.
4. Same file — `commandPathPattern`: allow the `drive9` prefix and widen the subcommand depth from `{0,2}` to `{0,3}` — drive9 flattens deep paths into up to four tokens (`drive9 admin tenant pool-create`).

Because validation is whole-batch and all-or-nothing, any single gate still closed means 100% loss. This is safe to merge before that change lands: with telemetry off by default nobody sends anything, and opted-in users simply see events rejected until the backend is updated.

## Tests and verification

Unit tests (`pkg/telemetry`, `cmd/drive9`, `cmd/drive9/cli`) cover enablement precedence (including invalid values and unreadable preferences failing closed), identity creation/permissions/concurrency, the exact payload key set and normalization, contract-rejection paths, delivery silence and no-redirect behavior, resolver eligibility, flag-name soundness and the flag allowlist, config round-trip, the exit hook, and both drift guards (command tree, flag names).

End-to-end with a real release-style binary against a local sink:

| Scenario | Events |
| --- | --- |
| release build, no opt-in | 0 (and no state file written) |
| `DRIVE9_TELEMETRY=on` / config `enabled: true` | 1 |
| config `enabled: true` + `DRIVE9_TELEMETRY=off` | 0 |
| `--help`, `version`, unknown command, `update --check` | 0 |
| `mount --foreground --supervised`, `mount --supervise-foreground` | 0 |
| `mount --bogus-flag`, `mount vault` (usage errors) | 1 each, `exit_code: 2` |
| `token issue --subject -secret-project` | 1, `flag_names: ["subject"]` |
| `fs grep -my-private-pattern` | 1, `flag_names: []` (unknown token = user input, dropped) |
| `mount --bogus-flag` | 1, `exit_code: 2`, `flag_names: []` |
| `fs find -name version`, `fs grep help` | 1 each (previously suppressed) |

| `admin tenant create help` | 0 |
| `fs cat prod:/file` | 1, no provider/region, `profile_source: unknown` |

Latency: against an endpoint that accepts and never answers, command time is a constant ~1.02s (previously ~3.0s); with telemetry off it is unchanged.

Notes for reviewers: `golangci-lint` v2.5.0 as installed in `bin/` was built with go1.26 and panics against this repo's go1.27 toolchain, so the local gate here was `gofmt` + `go vet` + the package tests; CI lint is the real check. `TestProcessMatchesIdentitySelf` fails in my local sandbox only because `/bin/ps` is blocked there — it is unrelated to this change.

## Follow-ups (not in this PR)

- The ti-cli change above (its own PR: code + [`client_contract_test.go`](https://github.com/tidbcloud/ti-cli/blob/main/internal/telemetrybackend/client_contract_test.go) fixtures + [`docs/telemetry-backend-design.md`](https://github.com/tidbcloud/ti-cli/blob/main/docs/telemetry-backend-design.md)); optionally a `product` column so drive9 and ti traffic can be separated in dashboards (`DisallowUnknownFields` means the payload cannot carry a discriminator field before the backend accepts one).
- Shared-service rate limits (per-IP 60/min, burst 120, global 10k-event buffer, whole-batch drops) are a capacity question for that service; the client deliberately does not retry or spool locally.

## Summary by CodeRabbit

- **New Features**
  - Added optional CLI telemetry, disabled by default.
  - Enable telemetry with `DRIVE9_TELEMETRY=on` or the CLI configuration file.
  - Reports limited command completion information, such as command path, platform, exit status, and approved environment metadata.
  - Excludes sensitive data, credentials, context names, flag values, help, version, and internal commands.
  - Telemetry failures never affect command results.

- **Documentation**
  - Added configuration, privacy, debugging, and endpoint guidance for CLI telemetry.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -169,15 +169,25 @@ to keep issue structure and required context consistent.
 - Help, version, commandless usage, unknown commands, every `drive9 update`
   form, and internal mount processes (`mount supervise`, `--supervised`,
   `--supervise-foreground`) are excluded before telemetry reads
-  `~/.drive9/config` or `.telemetry-installation-id`.
+  `~/.drive9/config` or `.telemetry-installation-id`. The `admin` subtree also
+  opts out of a bare `help` operand (`drive9 admin tenant create help` is a help
+  request there) via `bareHelp`; `fs grep help` is a real operand and still
+  reports.
+- Placement fields describe the command's own context. A command that addresses
+  a named context (`drive9 fs cat prod:/file`) reports no provider/region and
+  `profile_source: unknown`, because the active context does not describe it.
 - One user command must produce at most one event. Every process drive9 spawns
   itself — background mounts, supervisors, detached hydration — must go through
   `telemetryDisabledEnv`, and the generated systemd unit sets
   `Environment=DRIVE9_TELEMETRY=off`.
 - Flag names come from argv, so a flag value must never be recorded as a name:
-  a dash token directly after another dash token is treated as a value.
-  `telemetryFlagNames` documents the tradeoff; do not "fix" the resulting
-  under-reporting by guessing arity.
+  a dash token directly after another dash token is treated as a value, and
+  `telemetryKnownFlagNames` (guarded by
+  `TestTelemetryKnownFlagNamesCoverEveryDefinedFlag`) drops anything that is not
+  a flag the CLI defines. A dash-shaped operand that happens to spell a real flag
+  is still reported; that is an analytics-fidelity limit, not a privacy one,
+  because the recorded string is always a known flag name. Do not "fix" the
+  remaining under-reporting by guessing arity.
 - `cmd/drive9/telemetry.go` holds a hand-maintained copy of the command tree, and
   `cmd/drive9/telemetry_drift_test.go` compares it against the real dispatchers.
   Adding a command or subcommand means updating the tree; the test fails
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -132,6 +132,7 @@ pkg/
   tenant/               Tenant schema management
   pathutil/             Path canonicalization and validation
   semantic/             Durable background task types
+  telemetry/            Privacy-preserving CLI command telemetry client
   traceid/              Trace ID helpers
 internal/
   testtidb/             TiDB test helpers (shared across packages)
@@ -152,6 +153,50 @@ to keep issue structure and required context consistent.
 - `drive9 fs find ... -tag key=value` is an exact key/value match.
 - `drive9 fs find ... -tag key` means tag-key existence match.
 - `-tag` does not support fuzzy, prefix, contains, or regex matching.
+
+### CLI telemetry
+
+- Telemetry is opt-in and off by default. Nothing is sent, and no telemetry
+  state file is read or written, unless the user sets `DRIVE9_TELEMETRY=on` or
+  `"telemetry": {"enabled": true}` in `~/.drive9/config`. `cli.TelemetryPreference`
+  is the only reader of that preference; the field must stay on `cli.Config` so
+  `saveConfig` round-trips it. Do not add a build, install-source, or CI
+  heuristic that enables it implicitly, and do not add a telemetry management
+  command.
+- Opted-in commands post one best-effort `drive9.command.finished` event to the
+  shared ti-cli telemetry ingestion endpoint (`POST /v1/telemetry/batch`, `202`
+  is the only success status, 1s timeout, no retries, no local queue).
+- Help, version, commandless usage, unknown commands, every `drive9 update`
+  form, and internal mount processes (`mount supervise`, `--supervised`,
+  `--supervise-foreground`) are excluded before telemetry reads
+  `~/.drive9/config` or `.telemetry-installation-id`.
+- One user command must produce at most one event. Every process drive9 spawns
+  itself — background mounts, supervisors, detached hydration — must go through
+  `telemetryDisabledEnv`, and the generated systemd unit sets
+  `Environment=DRIVE9_TELEMETRY=off`.
+- Flag names come from argv, so a flag value must never be recorded as a name:
+  a dash token directly after another dash token is treated as a value.
+  `telemetryFlagNames` documents the tradeoff; do not "fix" the resulting
+  under-reporting by guessing arity.
+- `cmd/drive9/telemetry.go` holds a hand-maintained copy of the command tree, and
+  `cmd/drive9/telemetry_drift_test.go` compares it against the real dispatchers.
+  Adding a command or subcommand means updating the tree; the test fails
+  otherwise.
+- The shared ingestion service must allow `drive9.command.finished`,
+  `drive9_[A-Za-z0-9_-]{22}` installation IDs, `drive9` command paths of up to
+  four segments, the `drive9/` User-Agent, and must widen its `command_path`
+  pattern from `{0,2}` to `{0,3}` subcommand segments. It validates with
+  `DisallowUnknownFields`, so no product-discriminator field can be added to the
+  payload until the backend accepts one.
+- Events must never include flag values, credentials, paths, file contents,
+  command output, API payloads, context names, or tenant IDs. `pkg/telemetry`
+  sends no caller-supplied free text: `DRIVE9_TELEMETRY_TAG`/`_EXTRA` were
+  removed for that reason and must not come back without a key allowlist.
+- The shared ingestion service rate-limits per client IP (60/min, burst 120) and
+  drops whole batches when its in-memory buffer is full. The CLI sends one event
+  per command and never retries, so bursts (agent loops, many users behind one
+  NAT) lose events by design; that is a backend capacity question, not something
+  to fix with client-side retries or a local queue.
 
 ---
 
--- a/Makefile
+++ b/Makefile
@@ -29,6 +29,12 @@ VERSION ?=
 GIT_HASH ?= $(shell git rev-parse HEAD 2>/dev/null || echo unknown)
 GIT_BRANCH ?= $(shell git rev-parse --abbrev-ref HEAD 2>/dev/null || echo unknown)
 BUILD_TIME ?= $(shell date -u '+%Y-%m-%dT%H:%M:%SZ')
+INSTALL_SOURCE ?=
+# Empty for development builds on purpose: only release artifacts carry an
+# ingestion endpoint, so a dev binary can never post to production. Override it
+# to package a release build against another endpoint (e.g. staging).
+TELEMETRY_ENDPOINT ?=
+RELEASE_TELEMETRY_ENDPOINT = https://tdc-telemetry.tidbcloud.com/v1/telemetry/batch
 
 CLI_TARGETS ?= linux/amd64 linux/arm64 darwin/amd64 darwin/arm64 windows/amd64 windows/arm64
 MIGRATION_TARGETS ?= linux/amd64 linux/arm64
@@ -53,7 +59,9 @@ TEST_PKGS ?= ./...
 BUILDINFO_LDFLAGS = -X github.com/mem9-ai/drive9/pkg/buildinfo.Version=$(if $(VERSION),$(VERSION),dev) \
 	-X github.com/mem9-ai/drive9/pkg/buildinfo.GitHash=$(GIT_HASH) \
 	-X github.com/mem9-ai/drive9/pkg/buildinfo.GitBranch=$(GIT_BRANCH) \
-	-X github.com/mem9-ai/drive9/pkg/buildinfo.BuildTime=$(BUILD_TIME)
+	-X github.com/mem9-ai/drive9/pkg/buildinfo.BuildTime=$(BUILD_TIME) \
+	-X github.com/mem9-ai/drive9/pkg/buildinfo.InstallSource=$(INSTALL_SOURCE) \
+	-X github.com/mem9-ai/drive9/pkg/buildinfo.TelemetryEndpoint=$(TELEMETRY_ENDPOINT)
 
 .PHONY: mod test test-failpoint test-podman fmt lint install-lint build build-server build-cli build-cli-release build-migration build-migration-release build-migration-kube-plugin build-migration-kube-plugin-release run-server-local e2e-local sdk-integration-tests docker-build docker-build-migration docker-push-migration-multi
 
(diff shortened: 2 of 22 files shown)


mornyx:
Addressed the flag-name finding (thank you — it was a real gap, not just a theoretical one).

**The gap:** `telemetryFlagNames` accepted any dash-prefixed token in flag position. The pairwise rule only protected tokens that *follow* another flag. So a token that is user data rather than a flag — e.g. `drive9 fs grep -my-pattern`, where `grep`'s flag parser rejects the token but telemetry still reported it — could reach `flag_names`.

**The fix (2b0c69ad):**

- `cmd/drive9/telemetry_flags.go` adds `telemetryKnownFlagNames`, a closed allowlist of every flag the CLI defines, and `telemetryFlagNames` now drops anything outside it. Dropping an unknown name can only lose analytics, never leak a value.
- `TestTelemetryKnownFlagNamesCoverEveryDefinedFlag` re-derives that set from source with `go/ast` — both `flag.FlagSet` definitions (`fs.String("json", ...)`, `fs.DurationVar(&d, "timeout", ...)`, `fs.Var(v, "tag", ...)`, matched by method name, arity and a variable receiver so `zap.String("command", ...)` is not mistaken for a flag) and hand-rolled parsers (`case "--flag":`, `arg == "-x"`, `HasPrefix(arg, "--x=")`). It fails if a flag is missing from the allowlist (its name would be dropped silently) or stale in it (the entry could record user input), so the list cannot rot as the CLI grows.

Verified end-to-end against a local capture endpoint with a release-style binary:

| command | before | after |
| --- | --- | --- |
| `fs ls -l :/` | `["l"]` | `["l"]` |
| `fs find -name report :/` | `["name"]` | `["name"]` |
| `token issue --subject -secret-project` | `["subject"]` | `["subject"]` |
| `fs grep -my-private-pattern :/` | `["my-private-pattern"]` | `[]` |
| `mount --bogus-flag` | `["bogus-flag"]` | `[]` (event still sent, `exit_code: 2`) |

The PR description's "Flag names can never be values" section now documents both defenses.

mashenjun:
@codex review

srstack (approved):
LGTM — reviewed the full diff and verified locally on the PR head (2b0c69a).

What I checked:
- Privacy posture is solid: closed payload key set, flag names restricted to a closed allowlist re-derived from source by the AST drift test, dash-token-after-dash-token treated as a value, region/provider/version all collapse to allowlisted values, and enablement fails closed on invalid env values and unreadable preferences. No free-text channel.
- Exit-path wiring is complete: no remaining os.Exit in cmd/drive9/cli handlers outside the new exit.go hook; the three ExitOnError→ContinueOnError conversions preserve the old behavior (same message + usage, exit code 2) while making exits observable.
- Spawned-process suppression is doubly covered (DRIVE9_TELEMETRY=off via telemetryDisabledEnv + hasInternalProcessFlag argv check), including the generated systemd unit.
- -h/-v/--help/--version exclusion has no conflicts in this CLI: every such token before "--" genuinely means help/version today (worst case is under-reporting when one is a non-bool flag value — the safe direction).
- Command tree matches the real dispatchers, and both drift tests (command tree, flag allowlist) make future drift loud.
- Local verification: go test ./pkg/telemetry/... and the telemetry/exit-hook tests in ./cmd/drive9 and ./cmd/drive9/cli all pass; go vet and gofmt clean. CI was still pending at review time, so lint/e2e confirmation is on the bots.

Two non-blocking notes for follow-up:
1. fatal() in main.go uses direct type assertions for Reported()/ExitCode() — if a UsageError ever gets wrapped with %w on its way to fatal, the usage text would print twice and the exit code would degrade from 2 to 1. errors.As would make this robust; nothing wraps it today.
2. In loadOrCreateInstallationID, the rename fallback on filesystems without hard links is last-writer-wins, so two racing processes there can briefly diverge on identity. Negligible for best-effort analytics, and the read-back converges the next process.

Nice work — the design doc discipline in the PR description (contract mirroring, rejection table, latency budget) matches what's actually in the code.

mornyx:
Codex review follow-ups (b89decb7). Four of the five were actionable; one I am consciously leaving as a documented fidelity limit rather than "fixing" it with a per-command schema.

| Finding | Verdict | Action |
| --- | --- | --- |
| `cmd/drive9/telemetry.go:469` — placement read from the active context | Valid | **Fixed.** A context-scoped operand (`drive9 fs cat prod:/file`) is served by that context's client, so the event now reports no `cloud_provider`/`region_code` and `profile_source: unknown` rather than the active context's placement. `s3://…` and `:/path` keep the active context. |
| `cmd/drive9/telemetry.go:277` — bare `help` on leaf handlers | Valid | **Fixed** for the tree that needs it: `admin` leaves treat a bare `help` operand as a help request (`drive9 admin tenant create help` exits 0), so the `admin` subtree now carries `bareHelp` and such invocations are excluded. `fs grep help` / `fs find help` remain operands (their parsers have no bare-help branch) and still report, and `--server help` is not mistaken for help. |
| `pkg/telemetry/telemetry.go:377` — rename fallback replaces a concurrent identity | Valid | **Fixed.** The hard-link-free path now claims the destination with `O_CREATE`+`O_EXCL` first and only renames over its own placeholder, so the loser of a race adopts the winner's identity instead of replacing it. `TestClaimInstallationIDPathIsExclusive` covers the exclusivity. |
| `README.md:276` — incomplete field disclosure | Valid | **Fixed.** The collected list now names install source, the active context's provider/region (including the "absent for another context" rule), and the credential source. |
| `cmd/drive9/telemetry.go:375` — dash-shaped operand recorded as a flag | Acknowledged, not fixed | This is an analytics-fidelity limit, not a privacy one: since the closed flag allowlist landed, a recorded name is always a flag the CLI defines, so an operand's text is never transmitted — only the fact that a token spelling a real flag name appeared. Separating `drive9 fs cp :/src --output` (operand) from `drive9 fs find :/dir -name x` (flag after an operand) needs a per-command operand schema, which is exactly the kind of hand-maintained table that drifts; flag names are the wrong place to pay that cost. Documented in `AGENTS.md` and in the `telemetryFlagNames` comment instead. |

Verified end-to-end with a release-style binary against a capture endpoint: `admin tenant create help` gives 0 events, `admin tenant create --json` gives 1, `fs cat prod:/file` gives 1 with empty placement and `profile_source: unknown`, `fs cat :/file` gives 1 with `aws` / `aws-ap-southeast-1`.

PR #922, opened 2026-09-14
test(e2e,blackbox): add Node.js fs coverage across e2e PR gate and blackbox

mornyx:
## What

Node.js `fs` coverage for Drive9 mounts, in three layers. Up to now our compatibility story rested on pjdfstests and LTP, which validate single-syscall POSIX semantics — but real Node tooling drives the filesystem through libuv (threadpool interleaving of open/write/fsync/close, streams with tunable highWaterMark, `err.code`-based error handling, `readdir` dirent types, `statfs` space checks). None of that was exercised. All three layers below were verified end-to-end against a local `drive9-server` (`provider=local`, TiDB, real FUSE mount).

### 1. e2e PR gate — `e2e/fuse-nodefs-smoke-test.sh`

Basic-operations gate for the PR set, wired into `e2e/fuse-release-gate.sh` (default on, `RUN_FUSE_NODEFS_SMOKE=0` to skip), so it reaches `local-e2e.yml` with the rest of the release gate. The workload (`e2e/tools/nodefs_smoke.js`, Node stdlib only, requires node >= 18.17) covers:

- sync/promises read-write roundtrips, append semantics
- `createWriteStream`/`createReadStream` with 4KiB and 64KiB highWaterMarks, 9MiB payload crossing the storage-tier boundary, SHA-256 verified
- `FileHandle` positional read/write including a sparse hole, truncate, `datasync`/`sync`, `'a'` append
- `mkdir`/`readdir({withFileTypes, recursive})` (dirent types), `mkdtemp`, `rm` recursive
- rename incl. atomic replace and directory rename; the write-temp-then-rename publish pattern
- symlink (`readlink`/`lstat`/`stat` follows), hardlink (`nlink`, survives origin unlink)
- stat/bigint stat/`statfs` sanity, `realpath(.native)`, `copyFile` incl. `COPYFILE_FICLONE`, `chmod`/`utimes`
- the `ENOENT`/`EEXIST`/`ENOTEMPTY`/`EISDIR` error-code matrix (ecosystem tools switch on `err.code`)
- bidirectional cross-channel checks: CLI `fs cat` reads a Node-written file; mounted Node reads a CLI-uploaded fixture with matching checksum
- unmount → remount → re-verify the whole tree against a checksum manifest

Verification run: 41 in-mount assertions + 25 remount assertions, all green.

### 2. blackbox `community.node_fs` — official Node.js core fs suite

[...]

PR #929, opened 2026-09-15
fix(fuse): gate umount success on kernel mountinfo entry

### LINKED ISSUE #928: umount returns success while the kernel mountinfo entry outlives the (cleanly exited) foreground mount by 100s+

On a Linux host with FUSE (as observed on `ubuntu-latest` GitHub runners with fuse3, `DRIVE9_TENANT_PROVIDER=local`, TiDB at `127.0.0.1:4000`, local mock S3 via `scripts/local-minio.sh`):

1. Build CLI + server from the repo (`make build-cli build-server`), start `drive9-server` with `DRIVE9_TENANT_PROVIDER=local`.
2. Run the Node.js fs e2e smoke, which mounts twice and unmounts twice with `--foreground`:

   ```bash
   bash e2e/fuse-nodefs-smoke-test.sh
   ```

   The script mounts the tenant root at a fresh mountpoint (`drive9 mount --mode=fuse --foreground <mp>`), writes a fixture tree through the mount (its mount log shows commit-queue uploads still completing at unmount time), calls `drive9 umount --timeout 60s <mp>`, waits for the mount process to exit, then verifies the mountpoint is gone from `/proc/self/mountinfo`.

3. The check is timing-sensitive and nondeterministic — the surest reproducer is simply running the Local E2E workflow on a branch containing `e2e/fuse-nodefs-smoke-test.sh` repeatedly; it failed 7 consecutive runs before the test-side workaround landed (PR #922 runs [34948565339](https://github.com/mem9-ai/drive9/actions/runs/34948565339), [34949566188](https://github.com/mem9-ai/drive9/actions/runs/34949566188), [34951422197](https://github.com/mem9-ai/drive9/actions/runs/34951422197), [34952497377](https://github.com/mem9-ai/drive9/actions/runs/34952497377), [34953915244](https://github.com/mem9-ai/drive9/actions/runs/34953915244), [34955088031](https://github.com/mem9-ai/drive9/actions/runs/34955088031), [34956415537](https://github.com/mem9-ai/drive9/actions/runs/34956415537)).

A focused repro loop (closer to the unit of failure), unverified but directly implied by the evidence below:

```bash
drive9 mount --mode=fuse --foreground "$MP" & MPID=$!
# write enough through the mount that the commit queue is still draining
dd if=/dev/urandom of="$MP/big.bin" bs=1M count=16
drive9 umount --timeout 60s "$MP"
wait "$MPID"   # mount process exits cleanly, code 0
# immediately and repeatedly poll the kernel table:
grep -F " $MP " /proc/self/mountinfo
```

Once `drive9 umount` returns success and the foreground mount process has exited with code 0, the kernel mount table entry for the mountpoint should be gone from `/proc/self/mountinfo` — promptly and deterministically (well under a second for an idle mount; bounded and predictable when the commit queue is still draining).

Three independent observations from the CI forensics (full guard diagnostics are embedded in the linked run logs; the instrumented script is `e2e/fuse-nodefs-smoke-test.sh` on PR #922):

- **The kernel mountinfo entry outlives a cleanly-exited mount by 80–100+ seconds, nondeterministically.** In the failing runs, after the final `drive9 umount` returned 0 and the mount process logged `mount lifecycle event=exit code=0 reason=signal`, polling `awk '$5==mp' /proc/self/mountinfo` kept reporting the entry as present for over 80 seconds (a 60s settle window plus force-unmount attempts plus a final 20s wait all saw it), then it disappeared on its own. In other runs the exact same sequence cleared in under 1 second. The mount log shows `commit queue: successfully uploaded ...` lines completing right up to unmount time, so a drain/connection-teardown race is the leading hypothesis.
- **`/etc/mtab` and the kernel table disagree during the linger.** During the lingering window, `fusermount3 <mp>` fails with `entry for <mp> not found in /etc/mtab` — the userspace table is already clean while `/proc/self/mountinfo` still lists the mount. After the mtab entry is gone, fusermount can no longer be used to clear the leftover kernel entry from an unprivileged context.
- **Secondary trap for whoever debugs this (do not chase it as the bug):** `mountpoint(1)` from util-linux 2.39.3 (ubuntu 24.04) is an unreliable probe here — it exits **32**, not 1, for a plain non-mountpoint directory, and its `-q` mode reported the unmounted path as *mounted* for the entire 80s+ window while mountinfo/mtab both said otherwise. macOS never sees any of this (no `mountpoint` binary; different unmount path). Use `/proc/self/mountinfo` as the source of truth when investigating.

Context that was already ruled out while narrowing this down (each fix is in PR #922, commits `4c8e51b2`..`6fb959f5`):

- Deleting the remote backing tree under a live mount (causes a different umount failure — fixed by unmounting before remote cleanup).
- Supervised background mounts (`drive9 mount` without `--foreground` daemonizes a supervisor whose heal path re-mounted after a verified umount — fixed by mounting with `--foreground`; the linger was still observed *after* this, with the mount process verifiably dead).

Impact beyond the e2e script: any consumer that checks mount state after a successful `drive9 umount` sees a stale mount. A recursive `rm -rf` near the mountpoint would traverse into the half-dead mount (the e2e cleanup guard exists precisely to prevent this); mounting the same mountpoint again can stack entries; supervision/heal logic keying off mount presence can misfire.

```
component: drive9 mount
version: dev
git_hash: 1dd5ac04c706488d0a5d060ef34cdbf4e7ef40bf (and eb4651405f08930ec9dba6f05393c8762db840e8, e5445463d522c28aa77e497fac0eef20be9852d0 in later CI runs — all main @ 2026-09-14/15)
go_version: go1.27.1 / go1.25.1
platform: linux/amd64, ubuntu-latest (24.04), fuse3
```

### 5. Suspected area and fix directions

- `cmd/drive9/cli/mount.go` `runUmount` (~line 1855): the graceful-stop path SIGTERMs the process, then runs `fusermount`/`umount` via `umountArgv`, with `mountStillActiveAfterUmount` / `forceUnmountAfterUmount` only as supervisor-failure fallbacks. If the mtab entry is removed before the kernel entry is reaped (or the FUSE connection is still draining commit-queue uploads), the cleanup commands become no-ops that report "not found" while the kernel entry lingers until the connection finally closes.
- `drive9 umount` should not report success while `/proc/self/mountinfo` still lists the mountpoint: either wait for the kernel entry to disappear (bounded by `--timeout`), or make the teardown ordering such that the kernel entry cannot outlive the successful return (e.g. ensure umount2/fusermount completes only after the connection is released, or explicitly lazy/force-unmount and verify against mountinfo).
- The commit queue (visible in mount logs: `commit queue: successfully uploaded ...`) appears to still be draining at unmount time; check whether an in-flight upload holds the FUSE session open past `serve_end`.

**Acceptance criterion:** after `drive9 umount` returns 0 for a `--foreground` mount, the mountpoint is absent from `/proc/self/mountinfo` within a small bound (e.g. ≤5s), deterministically, including when the commit queue is draining at unmount time. The test-side workaround currently in `e2e/fuse-nodefs-smoke-test.sh` (180s settle + `fusermount -uz`/`umount -l` fallback in `wait_mount_settled`/`force_unmount`) can be tightened once this is fixed; its guard diagnostics (`/proc/self/mountinfo` dump + `IS_MOUNTED_TRACE` per-call verdict log, printed when the guard fires) will capture a recurrence automatically.

mornyx:
Fixes #928

## Problem

`drive9 umount` returned success (and the foreground mount process exited 0) while the kernel mount table (`/proc/self/mountinfo`) still listed the mountpoint for 80–100+s. During that window `/etc/mtab` was already clean, so `fusermount3 <mp>` refused to help with `entry for <mp> not found in /etc/mtab`, and the entry only disappeared on its own once the last reference was reaped.

## Root cause

Two compounding issues:

1. **Lazy (MNT_DETACH) fallbacks everywhere.** Teardown paths fall back to `fusermount3 -uz` / `umount -l` (worker signal handler retry exhaustion, supervisor `shutdownClean`, CLI umount failure forgiveness). A lazy detach updates `/etc/mtab` (and utab) immediately, but the kernel keeps the mount listed until the mount's last reference drops — the draining commit queue / live `/dev/fuse` fd / open files hold references well past unmount time.
2. **Stat-based success decisions.** Every "is it unmounted yet" check used `activeMountPoint` (device/inode comparison). After a lazy detach, path resolution falls through to the underlying directory, so the stat probe reports *inactive* while the kernel entry lingers — the exact blind spot that let umount (and the worker's exit-0 path) forgive a still-listed mount.

## Fix

- **`pkg/fuse`: authoritative kernel-table probe** — new `KernelMountTableHas` / `WaitMountTableClear` / `UnmountSyscall`. On Linux, parse `/proc/self/mountinfo` field 5 with the kernel's octal escaping (`\040` etc.), matching both the absolute and symlink-resolved spellings; other platforms keep the bounded stat probe (darwin umount is non-lazy, where they are equivalent). `UnmountSyscall` calls `umount2(2)` directly, which still works for the mount owner once the mtab entry is gone and fusermount refuses the leftover.
- **CLI `drive9 umount`** (`cmd/drive9/cli/mount.go`): the post-unmount probe is now mountinfo-authoritative, and success is **gated on the kernel entry actually disappearing**, bounded by `--timeout` (semantics documented in the flag help). While the entry lingers, it escalates: non-lazy fusermount retries (transient EBUSY) → direct `umount2` → one lazy detach, then keeps polling until the entry is reaped. Only a cleared table (or, unchanged from before, a plain fusermount failure for a mount we didn't stop) yields success. The whole gate runs inside the generation-gated cleanup closure, so it can never act on a successor's fresh mount.
- **FUSE worker** (`pkg/fuse/mount.go`): `Mount` now waits (≤10s) for the Serve goroutine to finish so the `/dev/fuse` fd is closed before returning, and the clean exit path waits (≤5s) for the kernel entry to clear, emitting a `mount_entry_linger` lifecycle event instead of silently exiting while the mount is still listed.

## Acceptance criterion (#928)

After `drive9 umount` returns 0 for a `--foreground` mount, the mountpoint is absent from `/proc/self/mountinfo` within the `--timeout` budget, deterministically, including when the commit queue is draining at unmount time (the daemon is stopped and waited for first; the entry is reaped when its references drop, and umount only then reports success). If a foreign holder keeps the entry alive past `--timeout`, umount now fails with `kernel mount-table entry ... still present after unmount` instead of a false success.

The 180s settle + `fusermount -uz`/`umount -l` workaround in `e2e/fuse-nodefs-smoke-test.sh` (PR #922) can be tightened on top of this; its guard diagnostics will still capture any recurrence.

## Testing

- `pkg/fuse/mount_table_test.go`: mountinfo field decoding (octal escapes, malformed lines), mount-point parsing, candidate matching (abs + symlink-resolved), Linux-gated fixture tests overriding `/proc/self/mountinfo` (listed / unlisted / unreadable-table fallback).
- `cmd/drive9/cli/mount_test.go`: umount fails while the kernel entry lingers past `--timeout`; umount escalates (one lazy detach) and succeeds once the entry clears within `--timeout`; a stopped-supervisor umount with a clean table invokes no unmount helper.
- `go vet` clean; `golangci-lint` 0 issues (with the repo's Go 1.25.1 toolchain; the local 1.27.1 toolchain cannot be read by golangci-lint v2.5.0 — pre-existing environment issue); full `cmd/drive9/cli` suite passes; `pkg/fuse` suite passes except two pre-existing darwin-host failures (`TestShouldContinueAfterWaitMountPermissionError*` fail on unmodified main too — the tested helper is Linux-gated; CI runs Linux).

The Linux FUSE e2e suites (`e2e/fuse-*.sh`, local-e2e workflow) all use `drive9 umount --timeout 60s`, which now also bounds the entry-clear wait — behavior-compatible: previously-fast clean unmounts still return immediately (clean table short-circuits the gate).

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved unmount reliability by confirming kernel mount entries are cleared before reporting success.
  * Added bounded retries and fallback cleanup for mounts that remain active.
  * Prevented unmount helpers and mount-status checks from blocking indefinitely.
  * Improved shutdown handling by waiting for FUSE services and mount entries to close.
  * Added conservative handling when mount status cannot be determined.

* **Tests**
  * Added coverage for lingering mounts, retry escalation, mount detection, and cleanup timeouts.


mornyx (commented):
## Review summary

This PR correctly diagnoses #928 (lazy `MNT_DETACH` + stat-based “inactive”) and puts the right primitive in the right layer: Linux `/proc/self/mountinfo` parsing, CLI success gated on the kernel entry actually disappearing, and a FUSE worker wait for Serve to close the `/dev/fuse` fd. Under a strict maintainability bar the direction is sound, and the new probe lives in small per-OS files rather than growing the two already-huge modules.

Prior review comments on this head are still open (no author follow-up). None of them recreate a common-path false umount success while mountinfo is readable, so they are not merge-blocking — but several still violate the PR’s own contract (`--timeout` bounds the wait; escalate non-lazy → `umount2` → one lazy; never forgive an unreadable table as “cleared”). Please address the should-fix items before treating this as done.

**What looks good**
- Mountinfo parsing is a real module (`pkg/fuse/mount_table*.go`) with octal-escape decoding, malformed-line tolerance, and abs + symlink-resolved candidate matching, plus fixture tests that override `procMountInfoPath`.
- The CLI gate runs inside `finishOwnedGenerationCleanup` (supervisor flock + generation re-check), so the new retries cannot fusermount a successor’s mount.
- Worker `serveDone` / `waitServeClosed` targets the actual go-fuse gap (`Server.Wait()` returns after the read loops; the fd closes later in `Serve()`), and `mount_entry_linger` replaces a silent exit while the entry is still listed.

---

**1. `KernelMountTableHas` fail-opens when the Linux table cannot be read**

`pkg/fuse/exports.go:133` maps `os.IsNotExist` to “not listed”. On Linux, `kernelMountTableHas` never stats the mountpoint — the only ENOENT is `os.ReadFile(procMountInfoPath)` — so a missing `/proc/self/mountinfo` (chroot / missing proc) reports the mount cleared. Any other table-read error then falls through to `activeMountPointBounded`, which is the MNT_DETACH blind spot this PR exists to replace: after a lazy detach, path resolution hits the underlying directory and the stat probe reports inactive while the kernel entry can still be listed.

That contradicts the comment immediately below (“stay conservative on unknowns so callers never forgive a failed unmount as success”). `TestKernelMountTableHasFixture` (`pkg/fuse/mount_table_test.go:114-117`) locks the unsafe ENOENT behavior in, and names a fallback it never reaches.

Fix: on Linux, any `kernelMountTableHas` error is indeterminate — treat as still listed (`true`) so umount fails closed. Do not special-case ENOENT for the table file, and do not use the stat probe to answer “is the kernel entry gone?” Keep the stat probe only on `!linux`. Flip the fixture to require conservative behavior when the table is missing/unreadable.

**2. Lazy detach runs on the first linger iteration, killing the documented ladder**

Comments, the PR body, and `waitForKernelMountEntryClear` all specify: non-lazy fusermount retries → direct `umount2` → one lazy detach, then poll. The loop does one helper, then immediately `forceUnmountAfterUmount` (`cmd/drive9/cli/mount.go:2407-2411`). `unmountSyscallCLI` only runs on attempt 4+.

After `MNT_DETACH`, path lookup falls through to the underlying directory, so later `fusermount3 -u` hits “entry for \<mp\> not found in /etc/mtab” and path-based `umount2(mp, 0)` cannot target the detached superblock. The new syscall rung is dead on the linger path it was added for. The outer closure can also lazy-detach at line 2041 before the wait starts, so the EBUSY path is not “one lazy detach”. A transient EBUSY that non-lazy retries would clear becomes a detached linger that only polls until `--timeout`.

Current tests assert `lazyCalls == 1` and never assert order; the linger-timeout test’s fake clock never reaches the `umount2` rung.

Fix: lazy-escalate only after helper retries and a non-lazy `umount2` (`if attempt > fusermountClearRetries && !lazyTried`). After lazy, poll; do not keep issuing path-based fusermount/`umount2`. Drop the closure’s extra force-detach (or make `lazyTried` see it). Give the ladder a single owner so helper + lazy are not split across the cleanup closure and this function. Assert that order in tests.

**3. `--timeout` does not actually bound the gate**

Flag help now promises one budget for process exit **and** kernel-entry clear. `defaultUmountDeps().run` is `exec.Command(...).Run()` with no context (`cmd/drive9/cli/mount.go:1862-1866`); `waitForKernelMountEntryClear` only checks the deadline after the command returns (`:2416`), and `deps.sleep(500ms)` is not capped to remaining time (`:2419`). A stuck helper (the domain this PR operates in) holds the wait — and the supervisor flock around it — past `--timeout`. `pkg/fuse` already has the bounded pattern: `runUnmountCmd` + `exec.CommandContext`.

The probe has the same hang: `mountTableCandidates` calls unbounded `filepath.EvalSymlinks` (`pkg/fuse/mount_table.go:23`), which lstats the FUSE path. That is the failure `activeMountPointBounded` (2s) was written to avoid, and both the CLI gate and the worker’s 5s exit wait go through it.

Separately, supervisor stop, worker stop, and entry-clear each get a full `--timeout`, so a 60s flag can run ~3× that.

Fix: one deadline from the start of `runUmount`. Run helpers via `CommandContext` (or reuse `runUnmountCmd`). Check the deadline before each attempt; cap sleep to remaining. Bound or skip EvalSymlinks of the mountpoint itself (parent-only resolve, or match cleaned abs path against mountinfo without statting the FUSE root).

**4. Non-Linux syscall rung overwrites the real failure**

On darwin, `attempt > fusermountClearRetries` calls `unmountSyscallCLI` → `unmountSyscall` in `pkg/fuse/mount_table_other.go:20-23`, which always returns `errUnmountSyscallUnsupported`. From attempt 4 the loop stops retrying the binary helper and `lastErr` is overwritten (`cmd/drive9/cli/mount.go:2399-2403`). A macOS user whose `umount` failed EBUSY then sees “syscall unmount not supported on this platform” after the full default timeout. The comment (“the binary-based unmount ladder still applies”) describes the intent, not the code.

Fix: skip the syscall rung unless it is actually supported; keep retrying the helper and preserve the original error. Inject the rung on `umountDeps` so the ladder is testable on Linux CI.

---

### Nits

- `--timeout 0` is documented as “disables waiting” / “verifies once”, but the loop still performs one helper re-issue **and** the lazy detach before the `timeout == 0` break (`cmd/drive9/cli/mount.go:2416`). Probe once and return, or fix the docs.
- If the per-phase budgets stay, say so in the flag help rather than implying a single wall-clock budget (e2e uses `--timeout 60s`).

---

### Recommendation
**Comment.** The #928 success gate is the right design and the common readable-mountinfo path no longer false-succeeds, but the table-unavailable fail-open, inverted escalation order, and unbounded `--timeout` still need to be fixed on this PR.

mornyx:
@mornyx Thanks for the thorough review — all four should-fix items and both nits are addressed in 26b70ade and 4cbf2d0c.

**1. Fail-open on unreadable table** — `KernelMountTableHas` now delegates probe failures to a per-platform `mountTableProbeFailed`: on Linux the kernel table is the only authoritative source and never stats the mountpoint, so any read failure (missing `/proc`, chroot, permission) is indeterminate and reports *still listed* — fail closed. The ENOENT → not-mounted mapping and the bounded stat fallback now exist only on platforms without a kernel-table file, preserving the previous darwin/windows semantics. `TestKernelMountTableHasFixture` is flipped to require the conservative behavior when the table is missing.

**2. Ladder order / single owner** — the lazy detach no longer fires on the first linger iteration and the cleanup closure's extra force-detach is gone; `waitForKernelMountEntryClear` is the single owner of the whole ladder: non-lazy helper retries → one direct `umount2` → a single lazy detach → poll-only (path-based unmounts can no longer target a detached superblock, so the loop stops issuing them). `TestRunUmountLadderOrderHelperThenSyscallThenLazy` pins the exact event sequence (`helper×4 → syscall → lazy`, nothing after), and `TestRunUmountNoSyscallRungPreservesHelperError` asserts poll-only after the lazy detach.

**3. `--timeout` bounding** — helper runs go through `exec.CommandContext` (5s per command, same bound `pkg/fuse` uses), the deadline is checked before every attempt, and sleeps are capped to the remaining budget, so the gate — and the supervisor flock around it — can no longer be held past the budget by a stuck helper (worst case overshoot is one in-flight command timeout, documented). The probe-side hang is also fixed: `mountTableCandidates` resolves symlinks through a 2s-bounded wrapper so a wedged FUSE endpoint cannot stall the CLI gate or the worker's 5s exit wait. On the single wall-clock deadline across supervisor-stop/worker-stop/entry-clear: I kept the existing per-phase budget rather than restructuring the pre-existing `terminateProcessGraceful` flow, and documented that explicitly in the flag help per your nit — happy to move to a shared budget as a follow-up if you prefer.

**4. Non-Linux syscall rung** — the rung is now injected as `umountDeps.unmountSyscall` (wired only on Linux); platforms without it take one final binary-helper attempt whose real error ("Resource busy", not a "not supported" placeholder) stays the reported one. Covered by a test on any host.

**Nits** — `--timeout 0` now verifies once and returns immediately without attempting any unmount; the flag help states the per-phase budget semantics.

Also folded in while re-testing: the two umount tests that spawn a real child now reap it promptly, so they no longer stall the real terminators for their full budget (16s/130s → ~0.1s).

Full `cmd/drive9/cli` suite passes; `pkg/fuse` passes except the two pre-existing darwin-host `TestShouldContinueAfterWaitMountPermissionError*` failures (present on unmodified main; the tested helper is Linux-gated); lint clean.

srstack (commented):
Reviewed this head against `fe581b9`. `go build`/`go vet` clean; full `./pkg/fuse` and `./cmd/drive9/cli` suites pass. The design is right: the authoritative mountinfo probe, the gate running inside `finishOwnedGenerationCleanup`, and the `serveDone` wait that closes the `/dev/fuse` fd before `Mount` returns all address the real #928 mechanics. Two findings remain.

**[P2] Fail closed when the mountpoint spelling cannot be resolved — `pkg/fuse/mount_table.go:31`**

`mountTableCandidates` keeps the resolved spelling only when `boundedEvalSymlinks` returns non-empty; on a timeout or any error it silently falls back to `filepath.Abs(mountPoint)` alone. `kernelMountTableHas` then returns `(false, nil)` when `/proc/self/mountinfo` lists the post-symlink dentry path (`mount_table_linux.go:26-30`), so `KernelMountTableHas` reports the entry cleared, `waitForKernelMountEntryClear` returns before running a single rung, and `drive9 umount /link` exits 0 with the mount still listed — the #928 false success, triggered in the exact wedged-endpoint case the 2s bound exists for (the CLI deliberately avoids `EvalSymlinks` on mountpoints for this reason: `cmd/drive9/cli/mount.go:1359`). Reproduced on this head: with a mountinfo fixture listing `/…/real-target` and a symlink mountpoint whose resolution errors, `kernelMountTableHas` returned `false, <nil>`. The worker's `WaitMountTableClear` uses the same probe.

Fix: propagate resolution failure and treat it as still listed (fail closed), or stop statting the FUSE root (resolve only the parent, or match the cleaned abs path plus parent-resolved spellings).

**[P3] Do not re-run the bounded probe after a timeout — `pkg/fuse/mount_table_other.go:27`**

A timeout from the first `activeMountPointBounded` is not `os.IsNotExist`, so `mountTableProbeFailed` launches the same bounded probe again: up to 2× `cleanupActiveMountTimeout` and a second abandoned goroutine per probe, and the retry can return `(false, nil)` if the endpoint recovers, flipping the conservative answer to a false clearance. Return `true` on the timeout without re-probing.

Test gap: `TestMountTableCandidates` only covers successful resolution, so the P2 path has no coverage.

Residual (not a regression here): the supervisor still decides via the stat probe (`pkg/mountsupervisor/supervisor.go:371,851`, `EnsureCleanMountpoint`/`mountStillNeedsClean`), so supervisor-side "clean" decisions keep the old blind spot.

mornyx:
@srstack Both findings are fixed — thanks for the fixture-level repro, it made the P2 path easy to pin down. Landed in 2db957fd3 and 5eba40f14.

**[P2] Inconclusive spelling now fails closed** — `mountTableCandidates` returns `(candidates, conclusive)`; `boundedEvalSymlinks` reports a timeout or non-ENOENT error as inconclusive, and `kernelMountTableHas` on Linux treats a non-match with an inconclusive resolution as **still listed** instead of `(false, nil)` — your exact scenario (kernel lists the post-symlink dentry, the symlink mountpoint's resolution errors) now fails closed, and `drive9 umount /link` can no longer exit 0 with the entry listed. ENOENT stays conclusive, since a missing path cannot hide a different kernel spelling behind a symlink. Covered by `TestKernelMountTableHasUnresolvedSpellingFailsClosed`, which reproduces your fixture setup (mountinfo lists only the target; resolution errors → still listed; ENOENT → conclusive no-match) and `TestMountTableCandidatesResolutionFailureIsInconclusive` — both verified on a real Linux host, plus the rest of the mount-table tests.

**[P3] No re-probe after a timeout** — `activeMountPointBounded` timeout errors now carry an `errors.Is`-able sentinel (`errActiveMountProbeTimeout`), and `mountTableProbeFailed` on the stat-probe platforms returns still-listed immediately on it — no second bounded wait, no second abandoned goroutine, and no recovery-window flip of the conservative answer. `TestMountTableProbeFailedTimeoutFailsClosedWithoutReprobe` asserts zero re-probes. The documented single retry is preserved for other transient errors.

On the residual you flagged (supervisor-side clean decisions via `EnsureCleanMountpoint`/`mountStillNeedsClean` still using the stat probe at `supervisor.go:371,851`): agreed it's not a regression from this PR — the supervisor path stops the worker first and the CLI gate now owns the authoritative verification, but switching those call sites to the mount-table probe looks like a natural follow-up. Happy to do it here if you'd rather have it in this change.

srstack (commented):
Re-reviewed at `5eba40f`. The previous findings are fixed and verified:

- **[P2] inconclusive spelling** — `mountTableCandidates` returns `(candidates, conclusive)` and Linux now fails closed on a non-match with inconclusive resolution. The fixture-level repro (`TestKernelMountTableHasUnresolvedSpellingFailsClosed`) matches what I could not resolve before.
- **[P3] probe timeout re-probe** — `errActiveMountProbeTimeout` carries `errors.Is`, and `mountTableProbeFailed` returns still-listed immediately instead of re-probing.
- The CodeRabbit items are also in: `exec.CommandContext` + `umountCommandTimeout` + deadline checks + capped sleeps, the ladder order (helper retries → `umount2` → one lazy detach → poll-only, pinned by `TestRunUmountLadderOrderHelperThenSyscallThenLazy`), and the unreadable-table fail-closed path.

Two notes remain.

**`[MINOR]` The `--timeout` overshoot bound is understated.**
The doc says a helper started just before the deadline can overshoot by at most one command timeout, but each poll also runs `KernelMountTableHas` → `mountTableCandidates` → `boundedEvalSymlinks`, which can block for `candidateResolveTimeout` (2s) when the endpoint is wedged. So a gate iteration can exceed `--timeout` by ~7s, not 5s (and `WaitMountTableClear` can add a poll interval). Either state both bounds or cache the symlink resolution for the lifetime of one `waitForKernelMountEntryClear` call — the parent chain does not change while the same mountpoint is polled, so only the mountinfo read needs to repeat. That would also cut the worst-case poll cost from 2s to ~0.

**`[MINOR]` The gate now holds the supervisor flock for up to `--timeout`.**
`finishOwnedGenerationCleanup` runs `underLock` under `tryExclusiveReadyTimeout`, and the closure can now poll for 60s by default (plus the overshoot above). This is defensible — it keeps a successor from mounting over a lingering entry — but it is a new lock-hold duration: a concurrent `drive9 mount` on the same mountpoint used to get the lock back after a helper invocation and now may wait out the whole budget. One line in the flag help or the closure comment would set expectations.

**`[NIT]` Matching is path-based, not mount-identity-based.** Any mountinfo entry at that path counts, so an unrelated filesystem mounted over a detached drive9 superblock (manual `mount`, or a successor that somehow slips past the generation gate) makes the gate fail closed after the timeout instead of reporting the drive9 unmount complete. Safe direction, but worth a sentence in `kernelMountTableHas`.

I also confirmed the blast radius of switching `mountPointStillActiveImpl` to the kernel-table probe: every caller (`mount.go` 2047/2072/2081/2084 and the gate itself) is on the umount path; mount-time readiness uses `probeMountPointReadyCLI`, so no other decision changed. The acceptance criterion holds structurally too: for non-Windows, `runErr` is the only success signal and every path that sets it nil goes through `waitForKernelMountEntryClear`.

Verification on `5eba40f`: `go build ./...`, `go vet ./pkg/fuse/... ./cmd/...`, `go test ./pkg/fuse/ ./cmd/drive9/cli/ -count=1` all pass.

srstack (commented):
Re-reviewed at `5eba40f`. The previous findings are fixed and verified:

- **[P2] inconclusive spelling** — `mountTableCandidates` returns `(candidates, conclusive)` and Linux now fails closed on a non-match with inconclusive resolution. The fixture-level repro (`TestKernelMountTableHasUnresolvedSpellingFailsClosed`) matches what I could not resolve before.
- **[P3] probe timeout re-probe** — `errActiveMountProbeTimeout` carries `errors.Is`, and `mountTableProbeFailed` returns still-listed immediately instead of re-probing.
- The CodeRabbit items are also in: `exec.CommandContext` + `umountCommandTimeout` + deadline checks + capped sleeps, the ladder order (helper retries → `umount2` → one lazy detach → poll-only, pinned by `TestRunUmountLadderOrderHelperThenSyscallThenLazy`), and the unreadable-table fail-closed path.

Two notes remain.

**`[MINOR]` The `--timeout` overshoot bound is understated.**
The doc says a helper started just before the deadline can overshoot by at most one command timeout, but each poll also runs `KernelMountTableHas` → `mountTableCandidates` → `boundedEvalSymlinks`, which can block for `candidateResolveTimeout` (2s) when the endpoint is wedged. So a gate iteration can exceed `--timeout` by ~7s, not 5s (and `WaitMountTableClear` can add a poll interval). Either state both bounds or cache the symlink resolution for the lifetime of one `waitForKernelMountEntryClear` call — the parent chain does not change while the same mountpoint is polled, so only the mountinfo read needs to repeat. That would also cut the worst-case poll cost from 2s to ~0.

**`[MINOR]` The gate now holds the supervisor flock for up to `--timeout`.**
`finishOwnedGenerationCleanup` runs `underLock` under `tryExclusiveReadyTimeout`, and the closure can now poll for 60s by default (plus the overshoot above). This is defensible — it keeps a successor from mounting over a lingering entry — but it is a new lock-hold duration: a concurrent `drive9 mount` on the same mountpoint used to get the lock back after a helper invocation and now may wait out the whole budget. One line in the flag help or the closure comment would set expectations.

**`[NIT]` Matching is path-based, not mount-identity-based.** Any mountinfo entry at that path counts, so an unrelated filesystem mounted over a detached drive9 superblock (manual `mount`, or a successor that somehow slips past the generation gate) makes the gate fail closed after the timeout instead of reporting the drive9 unmount complete. Safe direction, but worth a sentence in `kernelMountTableHas`.

I also confirmed the blast radius of switching `mountPointStillActiveImpl` to the kernel-table probe: every caller (`mount.go` 2047/2072/2081/2084 and the gate itself) is on the umount path; mount-time readiness uses `probeMountPointReadyCLI`, so no other decision changed. The acceptance criterion holds structurally too: for non-Windows, `runErr` is the only success signal and every path that sets it nil goes through `waitForKernelMountEntryClear`.

Verification on `5eba40f`: `go build ./...`, `go vet ./pkg/fuse/... ./cmd/...`, `go test ./pkg/fuse/ ./cmd/drive9/cli/ -count=1` all pass.

mornyx:
@srstack Thanks for the re-review and the blast-radius confirmation. All three notes are addressed in 2344f3da and df1de2d6 (documentation-only on top of the code you verified):

**[MINOR] Overshoot bound** — went with your "state both bounds" option: `waitForKernelMountEntryClear` now documents the overshoot per in-flight element rather than a single aggregate (each helper command up to `umountCommandTimeout` = 5s; each probe up to one bounded symlink resolution = 2s on a wedged endpoint; an iteration can run a helper plus two probes, and the post-loop verification adds one more probe), and `WaitMountTableClear` documents its own probe + poll-interval overshoot. I skipped the per-call resolution cache: it would bypass the injectable `mountStillActiveAfterUmount` probe the gate's tests stub, and the wedged-endpoint case it optimizes ends in the same bounded failure either way. If that worst-case poll cost ever matters in practice, the cache is a clean follow-up on top of the now-documented bounds.

**[MINOR] Flock hold** — the closure comment now states the gate may hold the supervisor flock for the whole `--timeout` deliberately (so a successor cannot mount over a lingering entry mid-cleanup), and that a concurrent `drive9 mount` on the mountpoint fails fast with lock contention instead of waiting out the budget — matching the `LOCK_EX|LOCK_NB` behavior you traced.

**[NIT] Path-based matching** — `KernelMountTableHas` now says so explicitly: any entry at a candidate path counts as listed, so an unrelated filesystem over the path reads as still-listed — the fail-closed direction — and identity-based matching would need fstype/owner correlation.

srstack:
Re-reviewed at `df1de2d`. All three notes are addressed, and the docs now match the code:

- The overshoot accounting is right per in-flight element (helper ≤5s, each probe ≤ one bounded symlink resolution; an iteration can run a helper plus two probes, plus the post-loop probe) — that mirrors `waitForKernelMountEntryClear` exactly.
- The flock note is accurate: I checked `mountsupervisor.TryExclusive` (`pkg/mountsupervisor/lock_unix.go:40-53`) — non-blocking acquire, `acquired=false` on contention — so a concurrent `drive9 mount` really does fail fast instead of blocking for the budget.
- `KernelMountTableHas` is the right place for the path-based-matching caveat.

Nothing outstanding from my side. Verified on `df1de2d`: build/vet clean, `go test ./pkg/fuse/ ./cmd/drive9/cli/ -count=1` passes.

srstack (approved):
LGTM. Re-reviewed at `df1de2d`: the authoritative mountinfo probe with fail-closed resolution and probe-failure handling, the ordered escalation ladder under one `--timeout` budget, and the `serveDone` wait that closes the /dev/fuse fd before `Mount` returns all address #928. Verified locally: build/vet clean, `go test ./pkg/fuse/ ./cmd/drive9/cli/ -count=1` passes.

PR #933, opened 2026-09-17
fix(fuse): return EACCES for list-only content reads

qiffang:
## Summary

- preserve per-path 403 content-read denial discovered through authorized directory listings
- reject read-capable FUSE opens with `EACCES` before Linux can translate a denied `READ` reply into `EIO`
- keep per-path 401 fail-closed by resetting the mount view, while leaving unrelated per-path failures unchanged

## Root cause

Drive9 already mapped the server 403 to FUSE status 13 (`EACCES`), but returned it from the `READ` operation after `open()` succeeded. Linux surfaced that sequence to the caller as `EIO`. Carrying the list-only authorization state into the inode/cache and refusing the read-capable `open()` produces the required POSIX `Permission denied` result without issuing a content read.

## Regression coverage

- batch-stat per-path 403: entry remains list-visible; read-capable open returns `EACCES`; no file handle or content request is created
- Lookup stat 403 + authorized parent list: same `EACCES` behavior
- batch-stat per-path 401: mount view resets and fails closed
- existing direct 401/403 content-read behavior remains covered

## Verification

- targeted regressions pass, including race detection
- `pkg/client` + `pkg/fuse` pass with the two pre-existing macOS-incompatible Linux-only mount-wait cases skipped
- patched Linux CLI replay of `e2e/fuse-supervision-test.sh`: **87/87 PASS**, including the original list-only permission assertion; exact mount/process/local/remote residual checks are zero

## Summary by CodeRabbit

- **Bug Fixes**
  - Enforced access restrictions when files can be listed but their content or metadata cannot be read.
  - Opening a denied file now returns an access-denied error without creating a file handle or attempting content reads.
  - Denied access is correctly isolated per file path, including hardlink aliases.
  - Authorization failures now clear stale mount-view information and refresh access state appropriately.
  - Directory listings now fail cleanly when authorization checks cannot be completed, without publishing incomplete entries.


qiffang:
停止当前实现，禁止合并。原 403→EIO P1 与 exact reproducer 保留；本 PR 引入的 hardlink/path-scoped authorization、kernel dentry revocation、extent alias isolation 需要独立架构设计与完整证明，后续另行处理。

Issue #936, opened 2026-09-18
P1: write-sync concurrent writes deadlock in kernel-cache notifyWrite boundary


qiffang:
## Summary

`--durability write-sync` loses forward progress under the existing deterministic FUSE concurrency workload. Four independent writer threads all hang after their first completed file; the FUSE syscalls do not return until the connection is forcibly aborted.

This is a **product failure / release blocker for write-sync**, not a harness failure.

## Frozen environment

- Drive9 commit: `11d9356e4509f90e838d61c47d650228ab376fd1`
- Engineering E2E run: `e06bc45b-8686-4962-a5ec-cc4fe977f319`
- Workload: existing `e2e/fuse-concurrency-stress.sh`
- Host: Linux FUSE
- Exact mount authority:

```text
drive9 mount --mode=fuse --foreground --durability write-sync \
  --cache-dir <run-scoped-cache> <run-scoped-mountpoint>
```

- Workload shape: 4 writers × 8 files, 2 concurrent readers; temp write + `fsync` + atomic rename, append + `fsync`, unlink churn, directory rename, and open-handle rename/read.
- Timeout: 120 seconds.

## Minimal reproduction

1. Build the above Drive9 SHA.
2. Start a fresh FUSE mount with the exact authority above.
3. Run `e2e/fuse-concurrency-stress.sh`'s deterministic workload (4 writers, 8 files/writer, 2 readers, 32 KiB payloads) against that mount.
4. Wait 120 seconds.

Equivalent script change versus the existing suite is only the frozen mount mode/authority:

```diff
-drive9 mount --mode=fuse "$MOUNT_POINT"
+drive9 mount --mode=fuse --foreground --durability write-sync \
+  --cache-dir "$RUN_ROOT/cache" "$MOUNT_POINT"
```

## Expected

- All writer threads finish within 120 seconds.
- Mounted and post-unmount remote manifests exactly match the deterministic expected manifest.
- No FUSE requests remain blocked.

## Actual

- All 4 writer threads time out after completing only `final/w{0..3}/file-000.txt`.
- Four append files exist as zero bytes in the mounted snapshot while their write syscalls remain blocked.
- Kernel shows four FUSE writes waiting indefinitely.
- Drive9 goroutines converge on a mutex wait in the kernel-cache notification boundary for more than five minutes.
- Result: **FAIL 10/12**.

Raw workload excerpt:

```text
writer thread timed out: Thread-3 (writer)
[...]

PR #941, opened 2026-09-20
test(e2e): run concurrency gate with write-sync

qiffang:
## Summary
- make the existing FUSE concurrency stress gate mount with explicit `--durability=write-sync`
- record the exact mount argv in the suite artifact
- add a regression test and document the #936 coverage boundary

## Verification
- `bash -n e2e/fuse-concurrency-stress.sh`
- `go test ./e2e`

## Scope
This is the minimal write-sync coverage gap closure requested for the Drive9 Autopilot four-mode campaign; it does not add a new runner or framework.


qiffang:
Blocking evidence gap: `mount_argv` is written only to `$MOUNT_LOG` under `$RUN_ROOT`. On a successful run, the EXIT trap removes `$RUN_ROOT` when `FUSE_CONCURRENCY_KEEP_ARTIFACTS != 1`, and Drive9 Autopilot only collects the suite stdout/stderr/result artifacts. Therefore a green acceptance run can lose the exact argv and cannot prove it actually used `--durability=write-sync`.

Please emit the already-built argv to a collected artifact (the suite stdout is sufficient) without executing a second mount, and extend the contract test to gate that collected-evidence path. The mount behavior itself and array quoting look correct; this request is specifically required to satisfy the acceptance criterion “actual argv contains `--durability write-sync`.”

qiffang:
Follow-up review at head `0d66735e`:

- The blocking evidence gap is fixed: the exact argv is now emitted through `tee` to suite stdout (collected by Autopilot) as well as `mount.log`.
- The logged argv and executed argv share the same `mount_args` array, so the evidence cannot silently diverge from the invocation.
- The actual invocation is explicit `mount --mode=fuse --durability=write-sync` and still preserves the optional profile.
- Regression test locks the write-sync argument, collected stdout evidence path, and execution through the array.
- `bash -n e2e/fuse-concurrency-stress.sh`, `go test ./e2e`, and `git diff --check` pass locally.

No remaining review blocker in this diff. Merge should still wait for the two required CI checks, then the independent Autopilot run must prove the collected stdout contains `--durability=write-sync` and report the real PASS/FAIL.

PR #946, opened 2026-09-20
test(e2e): emit role-bound mount evidence

qiffang:
## Summary
- emit machine-readable mount role plus same-source argv before every catalog mount
- preserve the legacy mount_argv evidence line and standalone durability defaults
- freeze stable roles for SQLite, perf, concurrency, crash recovery, supervision, and Git suites
- reconcile supervision PASS evidence against the branch-specific final role plan

## Tests
- `bash e2e/durability-contract-test.sh`
- `bash -n` for all modified E2E scripts
- `python3 -m py_compile e2e/durability-argv-test.py`
- `git diff --check`

Do not auto-merge; Drive9 E2E protocol change requires independent review.

## Summary by CodeRabbit

- **Tests**
  - Expanded end-to-end coverage for mount evidence, including role labels, argument matching, ordering, and completeness.
  - Added validation for invalid, missing, duplicate, or conflicting evidence and malformed mount arguments.
  - Updated mount scenarios to identify initial, recovery, remount, supervised, denied, and Git-related mounts.
  - Added checks ensuring observed mount roles match expected execution plans.
  - Added safeguards confirming mounts do not start when evidence recording fails.
  - Added coverage for CLI overrides, argument quoting, cleanup behavior, and crash recovery.


mashenjun:
@codex review

mornyx (changes requested):
## Review summary

This PR adds a role-bound `mount_evidence` record (paired with same-source `mount_argv`) before every catalog mount, gates evidence-write failure at status 70, freezes explicit roles on the SQLite/perf/concurrency/crash/supervision/Git suites, and reconciles supervision observed roles against a branch-specific plan. The producer contract is largely sound: one argv array feeds both evidence lines, a failed write cannot proceed to mount, and HEAD closes the earlier ensure-recovery gap (`record_ensure_recovery_mount` runs immediately before `drive9 mount ensure`, and the plan includes `ensure-recovery` iff `RUN_ENSURE_SMOKE=1`).

Under a strict maintainability bar the protocol still is not merge-ready. The new role validator is locale-collation dependent, so the PR’s own listed test command fails on the repo’s documented macOS e2e environment. The Python harness also jumped well past a healthy size with duplicated role catalogs, and the new CLI-override canonicalization rejects the natural macOS `$TMPDIR` spelling.

**What looks good**
- Same-source `mount_evidence` + `mount_argv` from one rendered argv array, with a `shlex` roundtrip guard and a 70 gate before any real `drive9` spawn.
- Supervision plan is actually branch-aware (`RUN_ENSURE_SMOKE` / `RUN_FOREGROUND_SMOKE` / token-management 200), and ensure-recovery evidence now matches that branch.
- CLI override ownership is reset on a later invalid request so cleanup cannot unlink a trusted external binary.
- The meta-harness executes the real builders (and, for crash recovery, a sourced real `main`) rather than restating argv by hand, including rejection of self-consistent invented evidence pairs.

---

**1. Role validator is locale-collation dependent; the listed contract test fails on macOS UTF-8**

`drive9_e2e_validate_mount_role` uses `case "$1" in ""|*[!a-z0-9-]*|-*|*-)`. Bash bracket ranges follow `LC_COLLATE`, not codepoints. Under `en_US.UTF-8` (macOS default, and this machine), `[a-z]` also matches uppercase and some non-ASCII, so `Initial` and `É` are accepted.

Verified on HEAD `2fe56c8`:
- `LC_ALL=en_US.UTF-8 bash e2e/durability-contract-test.sh` → `invalid mount role unexpectedly accepted: Initial` (exit 1)
- `LC_ALL=C bash e2e/durability-contract-test.sh` → all three PASS lines (exit 0)
- Isolated: `drive9_e2e_validate_mount_role Initial` / `"É"` accept under `en_US.UTF-8` and reject under `C`

CI with C/C.UTF-8 collation stays green while the PR body’s own test command is red on macOS, which this repo treats as a first-class e2e platform. The sibling `drive9_e2e_validate_mount_argv` already pins `local LC_ALL=C` (line 100) for exactly this class of glob.

Fix: add `local LC_ALL=C` at the top of `drive9_e2e_validate_mount_role` (all three role entry points funnel through it), or enumerate an explicit charset without a range. Then run the role-validation block of `durability-contract-test.sh` twice (ambient UTF-8 locale and `LC_ALL=C`) so this cannot regress behind a C-only CI.

---

**2. `durability-argv-test.py` grew 328 → 1177 lines, with the role catalog copied in several places**

This PR takes a 328-line argv harness to 1177 lines by absorbing CLI-override, cleanup ownership, callsite scraping, failure gates, exact transcripts, crash-recovery `main`, and supervision plan suites as more inline `f"""…"""` bash shims. That crosses the ~1k line bar without a module boundary. The same role lists are re-encoded at least in `callsite_role_contract_suite`, `exact_role_sequence_execution_suite`, git expected-role constants, `build_mount_role_plan`, and the grep block in `durability-contract-test.sh` (~206–225). A role rename now needs coordinated edits in two languages.

Split along the seams that already exist (protocol/CLI-override vs catalog role execution vs supervision plan), keep a thin runner, and make one `EXPECTED_ROLES` table the source of truth. Delete the weaker bash grep catalog once the Python callsite suite owns order and uniqueness.

**3. CLI override canonicalization rejects valid macOS `$TMPDIR` paths**

`drive9_e2e_use_cli_override` requires `pwd -P` parent + basename to equal the input string. On macOS, `$TMPDIR` is `/var/folders/…` → `/private/var/folders/…`, so a regular executable at `"$TMPDIR/probe-drive9"` is rejected with 64 and “must be an absolute canonical path”. Verified on HEAD. The Python suite never sees this because it builds fixtures with `Path(temp).resolve()`.

The same join also produces `//drive9` for a binary at filesystem root, so `/drive9` is rejected even though it is already canonical.

`e2e/AGENTS.md` documents `DRIVE9_CLI_BIN` with no “must already be `pwd -P`” caveat. Prefer canonicalize-then-use: keep the relative/symlink/non-exec checks, then set `CLI_BIN` from the `pwd -P` parent (and join with `"${parent%/}/…"`). Add a contract case that passes an unresolved `$TMPDIR`-style path.

**4. Evidence-then-gate snippet is copy-pasted across every catalog builder**

The `print_mount_evidence | tee -a "$log"; echo …; return 70` block is repeated at sqlite, perf, concurrency, write-perf, crash-recovery, git, and four supervision helpers. Launch style differs; the record+fail-closed part does not. One forgotten `tee`, a redirect of evidence off stdout, or `return 1` instead of 70 will drift.

Add a single helper, e.g. `drive9_e2e_emit_mount_evidence "$role" "$log" "$CLI_BIN" "${args[@]}"`, that owns validate/render/tee and returns 70. Leave process launch in each suite.

**5. `%q` + Python `shlex` is locale- and bash-version-sensitive for non-ASCII argv**

`drive9_e2e_render_mount_argv` round-trips Bash `%q` through `shlex.split`. Under `LC_ALL=C`, non-ASCII tokens become ANSI-C `$'…'` quoting that `shlex` does not decode; on macOS bash 3.2 the mismatch also appears under UTF-8. The helper then returns 64 and aborts the mount. This is fail-closed (no silent missed mount) and today’s catalog paths are ASCII, but Drive9 paths are UTF-8 and every catalog mount now goes through this helper.

Quote with a `shlex`-stable encoder (or pin rendering locale and document ASCII-only argv with a dedicated error). Add one C-locale and one UTF-8 unicode roundtrip in `durability-contract-test.sh`.

---

### Nits

- `fuse-supervision-test.sh:1169` expands `"${MOUNT_ROLES_OBSERVED[@]}"` unguarded under `set -u`. On bash 3.2 an empty array is an unbound variable (verified). Use the file’s own `${arr[@]+"${arr[@]}"}` idiom; this path is the one the reconciliation exists to diagnose.
- `evidence_failure_gate_suite` checks `rc==70` before `test ! -e "$CAPTURE_FILE"`. At HEAD the gate precedes any background `drive9`, so this is not a false green. `wait || true` before the negative assert would still make a future gate regression deterministic for the five builders that background the CLI.

---

### Recommendation

**Request changes.** The role grammar does not hold on macOS UTF-8, and the PR’s own contract test fails there; fix that before merge. The file-size, CLI canonicalization, and duplicated role/evidence helpers should land in the same pass if practical, but they are not additional correctness stoppers.

